PR #25, opened 2018-01-21
Error reporting like rustc

kdy1:
Error reporting is implemented using rustc 

-----
Edit: grammar


kdy1:
Now this shows error message like

```
error: An async function cannot be generator
 --> a.js:1:15
  |
1 | async function* a() {
  |               ^

thread 'main' panicked at 'Failed to parse module', src/main.rs:81:17


```

Issue #18, opened 2018-01-16
Plugin system


kdy1:

Michael-F-Bryan:
I'd be keen to help out with this! We can probably reuse a lot of the logic from the [Dynamic Loading and Plugins][1] chapter from my FFI guide.

The main thing we'd need is a stable-ish interface for doing transformations (i.e. the `Fold<T>` trait) and then figure out a set of conventions for how users register their plugin(s).

[1]: https://s3.amazonaws.com/temp.michaelfbryan.com/dynamic-loading/index.html

kdy1:
Note: All core transformations ported from babel and closure compiler will be statically compiled.

stefanpenner:
I have been doing some exploration re: work to get swc powering ember applications.
Currently, we rely heavily on babel plugins. Some of these plugins may make good core functionality, others maybe should remain as part of a plugin ecosystem.

My tracking issue: https://github.com/stefanpenner/broccoli-swc/issues/10

* [babel-plugin-ember-modules-api-polyfill](https://github.com/ember-cli/babel-plugin-ember-modules-api-polyfill) rewrite `imports` to globals
* [babel-plugin-debug-macros](https://www.npmjs.com/package/babel-plugin-debug-macros) a macro system for build time code stripping
*  [ember-compatibility-helpers](https://github.com/pzuraq/ember-compatibility-helpers/blob/master/comparision-plugin.js) a plugin to aide transitioning consumers from 1 system to another.
* [babel-plugin-htmlbars-inline-precompile](https://github.com/ember-cli/babel-plugin-htmlbars-inline-precompile) a plugin which rewrites tagged template literals containing HBS templates into their compiled output

kdy1:
@stefanpenner I have some questions

- babel-plugin-debug-macros

swc currently has `inline_globals` pass and `simplifier` pass which performs dead code elimination.


```js
if (__DEBUG__) {
    console.log('Foo')
}
```

Isn't this enough? If then, is importing flag from somewhere important?

(See https://github.com/swc-project/node-swc/blob/06845c25c64ccfc626cfdac53987f17b94ef95d8/__tests__/optimizer_test.js#L4-L22 for usage)

stefanpenner:
@kdy1 it is very close, the exception is our API accomplishes this not via globals but rather via module imports.

so where you have `__DEBUG__` global, we have:

```
import { DEBUG } from '@ember/env-flags';
```

and

```
[...]

PR #42, opened 2018-11-02
WIP: A Rudimentary Plugin System

Michael-F-Bryan:
This adds the basic infrastructure for a binary plugin system. We essentially use `LibLoading` or `dlopen` to load a DLL at runtime and execute a couple well known symbols to register a plugin.

Later on we may want to create a small utility which can download a plugin repo (e.g. from GitHub), compile it, then copy the DLL to some user-specific plugins directory. We'd also need some sort of plugin discovery mechanism (e.g. a `SWC_PLUGIN_PATH` environment variable which defaults to the plugin directory) to know where to look when loading plugins as part of the build process. 

Outstanding questions:

- What should a plugin look like? Does it implement a special `Plugin` trait, is it just an implementation of `Fold<T>`, etc? This probably ties in with #41.
- `rustc` doesn't guarantee ABI compatibility between versions, meaning each plugin would have to be compiled with *exactly* the same version of `rustc` as `swc`
- How do you declare a plugin? Should we have an equivalent of `.babelrc`, or maybe take a different approach?
- Do we want to support writing plugins in other languages?
- How will the rest of the system want to use loaded plugins?

This works towards #18.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ swc_atoms = { path = "./atoms" }
 swc_common = { path = "./common" }
 swc_ecmascript = { path = "./ecmascript" }
 swc_macros = { path = "./macros" }
+swc_plugins = { path = "./plugins" }
 rayon = "0.9"
 slog = "2"
 slog-envlogger = "2.1"
--- a/plugins/Cargo.toml
+++ b/plugins/Cargo.toml
@@ -7,3 +7,6 @@ authors = ["Michael Bryan <michaelfbryan@gmail.com>"]
 swc_common = { path = "../common" }
 anymap = "0.12.1"
 libloading = "0.5.0"
+
+[dev-dependencies]
+dummy_plugin = { path = "tests/dummy_plugin" }
--- a/plugins/Cargo.toml
+++ b/plugins/Cargo.toml
@@ -6,3 +6,4 @@ authors = ["Michael Bryan <michaelfbryan@gmail.com>"]
 [dependencies]
 swc_common = { path = "../common" }
 anymap = "0.12.1"
+libloading = "0.5.0"
--- a/plugins/Cargo.toml
+++ b/plugins/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "swc_plugins"
+version = "0.1.0"
+authors = ["Michael Bryan <michaelfbryan@gmail.com>"]
+
+[dependencies]
+swc_common = { path = "../common" }
+anymap = "0.12.1"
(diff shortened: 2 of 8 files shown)


kdy1:
Is loading dll safe? I'm not sure about it.

Michael-F-Bryan:
> Is loading dll safe? I'm not sure about it.

That's a bit of an open question and depends on how you implement it. You normally create a safe interface which everyone needs to go through in order to register a plugin (e.g. using a `Plugin` trait object). In this case, we use a `register_plugin!()` macro to generate functions with known symbol names that the plugin loader can use. 

Because we would have complete control over the exported functions, we can make sure everything stays type safe. ABI problems can be avoided by also exporting functions which return the `rustc` and plugin library versions and checking for compatibility before starting the main registration procedure.

From a security standpoint it's no less safe than any other plugin mechanism. At some point we're letting 3rd party code be executed, and that code can do whatever it wants.

> It exposes plugin api from binding and plugin can be written is node js. I also think plugins should be implemented in javascript (considering rust's learning curve).
(Best is being drop-in-replacement of the babel, but I don't think it's possible)

Considering we're targeting the JavaScript community, I agree that it makes sense to write plugins in JavaScript. That said, you'd need to write loads of boilerplate code to allow NodeJS to access Rust data, which is arguably a lot less safe than using a Rust DLL which imports and uses the `swc` crates internally.

Considering code transformations are normally small functions called in tight loops and that each FFI call adds a lot of overhead (due to inhibiting optimisations, marshalling data back and forth, etc.), I wouldn't be surprised if we actually get *worse* performance than Babel by using JavaScript plugins...

kdy1:
I'm not sure which feature should be implemented as a plugin. I mean, almost features of babel / closure compiler / webpack will be statically linked.

I want to make swc as a final step of compilation. (for initial release)


e.g.)

**v0.1**:
    - jsx  -> (babel) -> ecmascript -> (swc) -> bundle.js

**v0.2**:
    - jsx / stage 0 ~ 3  -> (swc) -> bundle.js

memchk:
@Michael-F-Bryan If you are using macros to generate the plugin interface, could you not also mark those interfaces as `#[repr(C)]`, and generate the translation code? This would allow the plugin and swc to be built with differing rustc versions.

ishitatsuyuki:
Personally, I'm not in favor of dynamic linking based plugin approaches. Version or ABI mismatch can easily occur, plus it's simply too hard to get unloading right.

Making things too dynamic can degrade performance; before deciding on the interface, what are the things that we want to be pluggable?

kdy1:
I don't want compiler passes to be pluggable.
I prefer something like javascript plugin or https://github.com/kentcdodds/babel-plugin-macros

Most preferable way I think is implementing https://github.com/kentcdodds/babel-plugin-macros as a static plugin and encourage user to use macros instead of developing plugin.

PR #44, opened 2018-11-03
Fix tests

kdy1:


kdy1:
I think codegen tests should be updated.. I'll do it soon.

PR #76, opened 2018-11-23
[WIP] Node binding

kdy1:
We should provide prebuilt binary.
Compiling swc takes 3~4 minutes.


kdy1:
This will be developed in an separate repository

PR #205, opened 2019-02-13
Fix Bugs

### LINKED ISSUE #196: swc typescript parser silently fails when a type is exported
This fails silently:
```ts
export type Link = { key: string; text: string };
console.log("Foo");
```

This does not, but the output is incorrect:
```ts
type Link = { key: string; text: string };
console.log("Foo");
export { Link };
```

transpiles to: 
```js
console.log('Foo');
export { Link }
```

but should be 
```js
console.log('Foo');
```

To cover another edge case, this:
```ts
type Link = { key: string; text: string };
const Link = "Boo";
console.log("Foo");
export { Link };
```

should transpile to:
```js
const Link = "Boo";
console.log("Foo");
export { Link };
```

### LINKED ISSUE #201: Failure using named exports as object keys
### Input
```js
import * as ns from './namespace';

let outer;
outer = {
  inner: {
    [ns.EXPORT1]: true,
    [ns.EXPORT2]: true
  }
};
```

### Output
```
$ swc test.js
    if (key in obj) {
            value: value,
            enumerable: true,
            configurable: true,
            writable: true
        });
    } else {
        obj[key] = value;
    }
    return obj;
}
import * as ns from './namespace';
var outer;
var _obj;
outer = {
    inner: _obj = {
    }, _defineProperty(_obj, ns.EXPORT1, true), _defineProperty(_obj, ns.EXPORT2, true), _obj
};
```

### Error demonstrated at run-time

```
$ swc test.js | node -r esm
    }, _defineProperty(_obj, ns.EXPORT1, true), _defineProperty(_obj, ns.EXPORT2, true), _obj
                               ^

SyntaxError: Invalid or unexpected token
```

### LINKED ISSUE #203: Failure using sibling object values with duplicate keys
### Input

```js
const obj = {};

obj.prop = {
  alpha: {
    charlie: true
  },
  beta: {
    charlie: true,
    delta: true
  }
};
```

### Test run
```
$ swc test.js | node
[stdin]:22
    }, _defineProperty(_obj, 'charlie', true), _defineProperty(_obj, 'delta', true), _obj
                             ^^^^^^^^^

SyntaxError: Unexpected string
```

### LINKED ISSUE #204: String concatenation with carriage return broken
### Input
```js
const a = fn() + '\r\n';
```

### Output
```js
\n';a = fn() + '
```

### LINKED ISSUE #206: access object properties using bracket notation produce wrong result
### Input

```javascript
const foo = {
  bar: {
    baz: 1
  }
}

const bar1 = 'bar'
const baz1 = 'baz'

const number = foo[bar1][baz1]

console.log(number)
```

### Output

```javascript
var foo = {
    bar: {
        baz: 1
    }
};
var bar1 = 'bar';
var baz1 = 'baz';
var number = foo.bar1.baz1;
console.log(number);
```

### LINKED ISSUE #207: Parentheses incorrectly removed (2 cases)
Tested on 1.0.9

Input:
```
export const foo = a => ({x: 'xxx', y: {a}});
```

Output:
```
'use strict';
Object.defineProperty(exports, '__esModule', {
    value: true
});
exports.foo = void 0;
var foo = (a)=>{
        x: 'xxx',
        y: {
            a
        }
    }
;
exports.foo = foo;
```

The problem is it removes the `()` around the object literal, which causes it to be treated as a code block: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Returning_object_literals

(Poc is a little bigger just so it gives a parse error that clearly shows the problem and not silent wrong code.)

kdy1:
Fix #201, fix #196, fix #204, fix #203, fix #207, fix #206


kdy1:
I've run test locally, and all tests except doc test passes.

PR #247, opened 2019-02-20
Implement deserialize for ast nodes

kdy1:
This allows 
 - interop between nodejs / rust
 - plugin system

**Note**: current ast nodes does not match EsTree spec


kdy1:
I'll create a new pr after make ast nodes to follow estree spec

PR #352, opened 2019-03-19
Update util.rs

LuoZijun:


ecmascript/parser/src/lexer/util.rs
```diff
@@ -268,7 +268,7 @@ pub trait CharExt: Copy {
             None => return false,
         };
         // TODO: Use Unicode ID instead of XID.
-        c == '$' || c == '_' || c == '\u{200c}' || c == '\u{200d}' || UnicodeXID::is_xid_continue(c)
```
kdy1: Why did you remove _?

ecmascript/parser/src/lexer/util.rs
```diff
@@ -291,17 +291,14 @@ pub trait CharExt: Copy {
...
-                // Any other Unicode “Space_Separator” code point
-                true
-            }
-            _ => false,
+            _ => {
+                if self.is_line_break() {
+                    // NOTE: Line terminator is not whitespace.
+                    false
+                } else {
+                    c.is_whitespace()
```
kdy1: Why did you changed this?

kdy1 (changes requested):

ecmascript/parser/src/lexer/util.rs
LuoZijun: Beacuse `UnicodeXID::is_xid_continue('_')` will return `true`.

ecmascript/parser/src/lexer/util.rs
LuoZijun: ECMAScript WhiteSpace:
```rust
// White Space Code Points
// https://www.ecma-international.org/ecma-262/9.0/index.html#sec-white-space
pub const TAB: char    = '\u{0009}';
pub const VT: char     = '\u{000B}';
pub const FF: char     = '\u{000C}';
pub const SP: char     = '\u{0020}';
pub const NBSP: char   = '\u{00A0}';
// https://www.ecma-international.org/ecma-262/9.0/index.html#sec-unicode-format-control-characters
pub const ZWNBSP: char = '\u{FEFF}';

// https://github.com/open-i18n/rust-unic/blob/master/unic/ucd/common/tables/white_space.rsv
OTHER_UNICODE_WHITE_SPACE; // But not include `LineTerminator`
```

The char [`\u{180e}`](https://github.com/open-i18n/rust-unic/blob/master/unic/ucd/category/tables/general_category.rsv#L1112) is not belong `White_Space/Space_Separator` category.

So this is a mistake.

**Why use std method ?**

std impl: [char::is_whitespace](https://doc.rust-lang.org/nightly/std/primitive.char.html#method.is_whitespace)

Since rust std lib have impl `is_whitespace` method for `char` type, so we do not need maintain a `WhiteSpace` database ( And considering that in the future, as the version of Unicode is upgraded, new definitions may be added ).

But in ECMAScript definitions, `WhiteSpace` not include `LineTerminator`.

So we must ensure the `char` is not an `LineTerminator` first.

kdy1 (approved):

PR #363, opened 2019-04-09
Don't emit octal escape for null character followed by digit

### LINKED ISSUE #361: Invalid octal escape emitted for null character followed by number
```sh
> npx swc --version
@swc/cli: 0.1.18
@swc/core: 1.0.39
```

Input:
```js
'use strict';
'\x001'; // === '\0' + '1'
'\x009'; // === '\0' + '9'
```

Expected output:
```js
'use strict';
'\x001';
'\x009';
```

Actual output:
```js
'use strict';
'\01'; // === '\x01' in sloppy mode
'\09'; // === '\x009' in sloppy mode
```

Both strings are syntax errors in strict mode, even though `9` is not a valid octal digit.

The spec is weird about octal escape sequences in strict mode (see [section 7.8.4 of ES5](https://www.ecma-international.org/ecma-262/5.1/index.html#sec-7.8.4)).
The `EscapeSequence` production (without sloppy mode extensions) is defined as:
```
EscapeSequence = CharacterEscapeSequence
               | 0 [lookahead ∉ DecimalDigit]
               | HexEscapeSequence
               | UnicodeEscapeSequence
```
...which forbids null-character escapes if `\0` is followed by any `DecimalDigit`, not any `OctalDigit` like one might expect. So even though `'\09'` in sloppy mode is a valid string (equal to `'\0' + '9'`) which does not contain an octal escape, in strict mode it is a syntax error due to containing an octal escape!

erikdesjardins:
Fixes #361

### DIFF
--- a/ecmascript/codegen/src/lib.rs
+++ b/ecmascript/codegen/src/lib.rs
@@ -358,6 +358,16 @@ impl<'a> Emitter<'a> {
             .replace("\r", "\\r")
             .replace("\t", "\\t")
             .replace('\u{000B}', "\\v")
+            .replace("\00", "\\x000")
+            .replace("\01", "\\x001")
+            .replace("\02", "\\x002")
+            .replace("\03", "\\x003")
+            .replace("\04", "\\x004")
+            .replace("\05", "\\x005")
+            .replace("\06", "\\x006")
+            .replace("\07", "\\x007")
+            .replace("\08", "\\x008")
+            .replace("\09", "\\x009")
             .replace("\0", "\\0");
         // let value = node.value.replace("\n", "\\n");
 
--- a/ecmascript/codegen/src/tests.rs
+++ b/ecmascript/codegen/src/tests.rs
@@ -158,6 +158,20 @@ a;",
     );
 }
 
+#[test]
+fn no_octal_escape() {
+    test_from_to(
+        r#"'\x00a';
+'\x000';
+'\x001';
+'\x009'"#,
+        r#"'\0a';
+'\x000';
+'\x001';
+'\x009';"#,
+    );
+}
+
 #[derive(Debug, Clone)]
 struct Buf(Arc<RwLock<Vec<u8>>>);
 impl Write for Buf {


ecmascript/codegen/src/lib.rs
```diff
@@ -358,6 +358,16 @@ impl<'a> Emitter<'a> {
...
+            .replace("\00", "\\x000")
+            .replace("\01", "\\x001")
+            .replace("\02", "\\x002")
+            .replace("\03", "\\x003")
+            .replace("\04", "\\x004")
+            .replace("\05", "\\x005")
+            .replace("\06", "\\x006")
+            .replace("\07", "\\x007")
+            .replace("\08", "\\x008")
+            .replace("\09", "\\x009")
```
erikdesjardins: I didn't use regex for this because it seems the `#[emitter]` attibute can't handle `lazy_static!`

kdy1:
Is this correct?

I think `\09` should be printed as `\09`, not `\x009`

erikdesjardins:
Yes.

In Rust, `"\07"` means `<null character><the digit "7">`, but in JS it means `<the character with codepoint 0o07>`. So it's not sufficient to just replace the null character with `\0` when writing a string, because it might be followed by a number, and then JS will treat it as an octal escape.

Currently, SWC changes string literals: if you input `"\x007"` (which means `<null character><the digit "7">` in JS), it outputs `"\07"`, which is different. Try running `"\x007" === "\07"` in your browser console.

As for `"\09"` specifically, it ought to be equivalent to `<null character><the digit "9">` in JS, because `9` is not an octal digit. It is, in sloppy mode, but in strict mode it's a syntax error (see #361). So to be safe for 8/9 it's best to always output a full hex-escaped null character (`\x00`) whenever a null character is followed by a number.

erikdesjardins:
And I just remembered, [I fixed the same bug in Uglify a few years ago](https://github.com/mishoo/UglifyJS2/pull/2273), if you want to see more discussion

kdy1:
I got it.
bors r+

PR #384, opened 2019-05-01
Fix sourcemap.

### LINKED ISSUE #349: sourceMaps support
Now sourceMaps don't seem to be generated correctly: 
```js
var swc = require('@swc/core')
var fs = require('fs')
var inlineSourcemap = require('inline-source-map-comment')

swc.transform(
`
(() => { const a = 1; });



xxx;
`, {
	sourceMaps: true
}).then(function(result) {
	fs.writeFileSync('./result.js', result.code + inlineSourcemap(result.map))
})
```

Running this code in browser shows that sourcemap wasn't applied:
image

Babel handles code better, keeping line numbers.

Affected packages:
- https://github.com/swc-project/cli/issues/10
- https://github.com/swc-project/swcify/issues/1
- https://github.com/mattdesl/budo-multi-bundle (see related tweet https://twitter.com/DimaYv/status/1104825198430380032)

kdy1:
Closes #349

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "swc_common"
-version = "0.3.0"
+version = "0.3.1"
 authors = ["강동윤 <kdy1@outlook.kr>"]
 license = "Apache-2.0/MIT"
 repository = "https://github.com/swc-project/swc.git"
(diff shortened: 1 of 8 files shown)


kdy1:
@dy I fixed sourcemap issue. Does this seems correct?

https://sokra.github.io/source-map-visualization/

Example files: https://gist.github.com/kdy1/e8a4ae870192f7f46653e213e1650a6f

PR #391, opened 2019-05-16
Simplify: `left.rhs * right` is only safe when operators are the same

erikdesjardins:
Some counterexamples:
```js
var x = 2**32 - 1;
var y = x & ~3;

(x * 2) & 2; // 2
x * (2 & 2); // 8589934590

(y & 1) | 2; // 2
y & (1 | 2); // 0

(y | 1) & 2; // 0
y | (1 & 2); // -4
```


kdy1:
bors r+
sorry for late respone

PR #493, opened 2019-12-10
Fixes warning: unused import: `SpanExt`

anurbol:


kdy1:
It seems like SpanExt is used

anurbol:
So this seems to be a Rust's quirk. The trait is used somewhere, but Rust complains that it's not 😸

PR #494, opened 2019-12-10
Fixing the Wasm blocker for Parser

anurbol:
Rust has a [bug](https://github.com/rust-lang/rust/issues/64450) when compiling a project for Wasm (with `--target wasm32-unknown-unknown` flag), these changes are a workaround.

These changes are only supposed to fix the SWC Parser, not the whole SWC project. This is because I only need the Parser at the moment, also fixing the whole SWC Project would be a nightmare for me, a newbie in SWC. Maybe later.

### DIFF
--- a/ecmascript/parser/Cargo.toml
+++ b/ecmascript/parser/Cargo.toml
@@ -11,15 +11,15 @@ edition = "2018"
 [features]
 default = []
 # Requires nightly.
-fold = ["swc_common/fold", "ast/fold"]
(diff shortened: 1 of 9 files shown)


anurbol:
I can't understand what's wrong there for the linter...

kdy1:
Can you fix tests too?

anurbol:
@kdy Yes, I one second, God how did I overlooked that...

PR #479, opened 2019-12-02
use 2018 edition syntax

zimond:
Hey, this PR changes every crate in swc to actually adopt to Rust's 2018 syntax.

PR #495, opened 2019-12-10
Fixing the Wasm blocker for Parser - try 2

anurbol:
Rust has a [bug](https://github.com/rust-lang/rust/issues/64450) when compiling a project for Wasm (with `--target wasm32-unknown-unknown` flag), these changes are a workaround.

These changes are only supposed to fix the SWC Parser, not the whole SWC project. This is because I only need the Parser at the moment, also fixing the whole SWC Project would be a nightmare for me, a newbie in SWC. Maybe later.

### DIFF
--- a/ecmascript/parser/Cargo.toml
+++ b/ecmascript/parser/Cargo.toml
@@ -11,15 +11,15 @@ edition = "2018"
 [features]
 default = []
 # Requires nightly.
-fold = ["swc_common/fold", "ast/fold"]
+fold = ["swc_common/fold", "swc_ecma_ast/fold"]
 # Verify that expression is valid. Requires nightly.
 verify = ["fold"]
 
 [dependencies]
 swc_atoms = { version = "0.2", path ="../../atoms" }
 swc_common = { version = "0.4.0", path ="../../common" }
-ast = { package = "swc_ecma_ast", version = "0.10.0", path ="../ast" }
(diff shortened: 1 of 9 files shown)


kdy1:
It's great enough. Note that parser is special in swc, as it should be usable from stable rust, and as an ecmascript parser. Some rust projects uses it already.

anurbol:
@kdy1 that's good to hear! I suspected that this should be the case, because the parser alone seems to be of incredible usefulness!

kdy1:
Can you run cargo fmt --all?

kdy1:
I think the rustfmt error is related to alphabetical sorting.

anurbol:
Yep, it was the case. Now fixed. Do you by any chance know, is there an easier way in such situations than creating a new pull request every time?

kdy1:
Pushing a new commit to the pr branch would update the pr.

anurbol:
Oh I think it will fail again because of CRLF thing...

kdy1:
@anurbol Thank you for the pr!! I merged it.

anurbol:
Great!!! Thanks for merging, Dong Yoon (I hope it's correct)!

kdy1:
It's correct :)

anurbol:
@kdy1 Lol, I now see https://github.com/swc-project/swc/pull/479/files#diff-b052edc95cbba2d88ad873bdb3551146 was the "culprit" :D (no blame intended)

PR #518, opened 2019-12-20
Fix #517

### LINKED ISSUE #517: White space inserted after text nodes in JSX
Hey!

I'm looking to move a codebase from using `babel-register` to `swc-register` - and it's all looking smooth excluding one issue.

The JSX parser seems to be off for text nodes.

`<div>Hello World</div>` becomes `<div>Hello World </div>` - notice the extra space after `World`

Added a failing test case here -> https://github.com/swc-project/swc/pull/518

chinedufn:
This PR has a failing test for #517 - and will hopefully fix it once I find the cause of the problem.

If you see this before I find it I'd appreciate if you pointed me in the right direction!

To run the failing test:

```
git remote add tw issue-517 https://github.com/tailwind/swc
git checkout tw/fix/jsx-whitespace
cd ecmascript/transforms
cargo test 517
```

### DIFF
--- a/ecmascript/transforms/src/react/jsx/tests.rs
+++ b/ecmascript/transforms/src/react/jsx/tests.rs
@@ -1080,3 +1080,27 @@ test!(
     "<span> {foo}</span>;",
     "React.createElement('span', null, ' ', foo);"
 );
+
+// https://github.com/swc-project/swc/issues/517
+test!(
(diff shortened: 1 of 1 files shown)


chinedufn:
Actually - instead of diving in I'll just pause here since you're more likely to just immediately recognize the issue.

Otherwise I'd be happy to look further!

PR #534, opened 2019-12-25
Comment and TokenAndSpan - Implement Spanned

dsherret:
As title. These structures both have a `span` property, but didn't implement the `Spanned` trait.


kdy1:
Thank you for the pr!

Issue #535, opened 2019-12-26
ExportAll - String literal hi position extends beyond end of string literal token


dsherret:
Right now in this scenario...

```ts
export  *  from   "test"  ;
```

...the string literal span will go from the opening quote to the beginning of the semi-colon.

This is inconsistent with scenarios such as the following:

```ts
type Test = "test"    ;
```

My workaround is to get the `Token::Str` then use the span for that.

I believe this happens with any node that uses the `parse_from_clause_and_semi` function.

PR #540, opened 2019-12-28
Fix string literal span on import/export specifiers to only include string literal

dsherret:
Expanding on fix for #535. Took a look at the PR for that to figure this out.

Currently there's definitely going to be a failing test here. I still can't figure out how to compile the tests locally so I'm going to use the CI to test... (and to help me fill in the missing info in the json file that I know I'll get wrong the first time).


kdy1 (approved):
LGTM
I'll test it locally when I go home and merge it if the test successes

ecmascript/parser/tests/jsx/basic/custom/unary-paren/input.js.json
```diff
@@ -10,7 +10,7 @@
       "type": "ImportDeclaration",
       "span": {
         "start": 0,
-        "end": 26,
+        "end": 27,
```
dsherret: Import declaration should include the last semi-colon. This test and some others seem to be failing on master.

kdy1: Currently travis ci chokes on other tests for parser.
(It seems like a rustc version issue)


So I'll merge the pr after testing locally.

dsherret:
@kdy1 thanks and no rush! I am building off the source anyway so this is not blocking me.

What's the best way to run only one of these tests? I found I can run `cargo +nightly test --color always --all --all-features` on the root directory ok. I do get a lot of failing tests specific to windows I bet:

```
---- issue_528 stdout ----
[/bar
a,foo
b/bar
];
thread 'issue_528' panicked at 'assertion failed: `(left == right)`
  left: `"//bar\r[\n//foo\ra,\n//baz\r//bar\rb\n];"`,
 right: `"//bar\n[\n//foo\na,\n//baz\n//bar\nb\n];"`', tests\projects.rs:247:5
```

ecmascript/parser/tests/jsx/basic/custom/unary-paren/input.js.json
dsherret: Oh I see. By the way, feel free to make any edits to fix any mistakes I made. I am going a bit blind here because II'm not sure how to run the tests for this. (I tried `cargo +nightly test --color always --all --all-features`, but it doesn't seem to run the input.json-like tests)

kdy1: cargo +nightly test --color always --all --all-features

invokes input.ts based reference testing, but it updates reference file when comparison failed. This behavior was introduced to avoid updating test one by one.

kdy1:
The test issue_528 is declared on top level crate (swc), so you can just remove --all.
Otherwise, you can append issue_528 to the last of cargo test invokation.

kdy1:
I tested it locally, and got some failures.
Can you run `cargo test --all-features` in swc/ecmascript/paser directory? It will update related reference files.


Related tests:
```
/tests/jsx/basic/custom/unary-paren/input.js.json
/tests/jsx/basic/custom/unary/input.js.json
/tests/typescript/custom/issue-535/input.ts.json
/tests/typescript/custom/tsx-unary/input.tsx.json
/tests/typescript/import/not-top-level/input.ts.json
```

dsherret:
@kdy1 Oops... I accidentally looked at the 1-indexed "column" in my text editor instead of a 0-indexed value. I was correct originally then I changed it to be incorrect once I reviewed it.

I fixed the other span issues. It's helpful to be able to run the tests, but there are a lot of changed files on my machine after running that (it changes them to all go from \r\n to \n). The tests should probably be changed to bring in everything as \n newlines from the file system... or I think perhaps I have a setting in git that's changing them to \r\n on Windows.

kdy1:
>  Oops... I accidentally looked at the 1-indexed "column" in my text editor instead of a 0-indexed value. I was correct originally then I changed it to be incorrect once I reviewed it.

Can you revert it?



> I fixed the other span issues. It's helpful to be able to run the tests, but there are a lot of changed files on my machine after running that (it changes them to all go from \r\n to \n). The tests should probably be changed to bring in everything as \n newlines from the file system... or I think perhaps I have a setting in git that's changing them to \r\n on Windows.

Then I'll invearigate.

dsherret:
I changed the code to get the `Str`, then parse the semi-colon, then return the declaration. Let me know if that makes sense or if I should revert that to what I had yesterday.

I think everything is good now! My "oops" was for a past mistake where I did it correctly, rechecked it in my editor and thought it was wrong (because the column value was 1-indexed), then made the numbers wrong based on that 1-index, then reverted it back.

kdy1:
I think it makes sense and you don't need to revert it.

I'll merge pr soon.

kdy1:
Merged it. Thank you for the pr!!

PR #544, opened 2019-12-29
Fix comment line's end to be on same line

dsherret:
The comment text and span should not include the newline.

Previously on windows it would actually only include half of a `\r\n` and only include the `\r`.

This change also seems to improve the compiled JS indentation.


dsherret:
I might have missed updating some tests because when I run the test command it outputs the following:

```
thread 'test262::golden::6b5e7e125097d439.js' has overflowed its stack
test test262::golden::6b63d36394b0ffb3.js ... error: test failed, to rerun pass '--test test262'
```

I was able to get past this by skipping the following tests:

```rust
if file_name.contains("6b5e7e125097d439.js")
    || file_name.contains("714be6d28082eaa7.js")
    || file_name.contains("882910de7dd1aef9.js")
    || file_name.contains("dd3c63403db5c06e.js")
{
    continue;
}
```

kdy1:
Thank you for the pr!!

PR #547, opened 2019-12-31
Fix parsing class properties as readonly

dsherret:
The `readonly` property was always false.


kdy1:
Thank you for the pr!!

PR #551, opened 2020-01-02
Fix span of await token in for/forin/forof stmts

dsherret:
As title.

I didn't see any tests this affects, which I was kind of surprised by. I'm probably missing something?


kdy1:
@dsherret It means test262 does not have a test related to for of statement with await. Anyway, chnage seems correct so I'll merge it.

PR #552, opened 2020-01-02
Proposal: Comments.take_all?

dsherret:
Sorry, usually for enhancements I would open an issue first for discussion, but this is a very small change.

I would like the ability to be able to get all the comments stored in the `Comments` struct. This will allow me to work with comments in a way that works well for my situation (allow me to get all comments before and after nodes regardless of definition of trailing comments in swc and ensure I don't miss handling anything).

I believe the current way to do this would would be to iterate all the tokens then call `take_leading_comments` and `take_trailing_comments` on the `lo` and `hi` of each one, but it would be nice to just have a single `take_all` method. I also think this single method allows for a more efficient drainage of the concurrent hash maps to avoid acquiring a lock so many times.

Thoughts?

### DIFF
--- a/common/src/comments.rs
+++ b/common/src/comments.rs
@@ -3,6 +3,7 @@ use crate::{
     syntax_pos::{BytePos, Span},
 };
 use chashmap::{CHashMap, ReadGuard};
+use std::collections::HashMap;
 
 type CommentMap = CHashMap<BytePos, Vec<Comment>>;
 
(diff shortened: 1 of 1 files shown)


kdy1:
It seems like a good changed. Merged it.

dsherret:
@kdy1 could I get a new release of `swc_common` for this? Thanks in advance!

kdy1:
@dsherret I published `swc_common@0.4.5`.

dsherret:
@kdy1 awesome! Thanks so much. It's working :)

PR #553, opened 2020-01-02
Fix switch case span to include entire switch case

dsherret:
Previously the switch case span would not include the entire text.

It would be nice to add some very basic tests for the node positions for this (similar to the .ts and .t.json files). I would add it to the tests directory, but running the tests on windows is pretty annoying so I'm going to not bother... plus I have my own test suite in my project so I'll catch it quickly if this ever breaks in the future.


kdy1:
I reviewed it, invoked `cargo test --all` locally and merged it.

PR #554, opened 2020-01-02
Regenerator

kdy1:
swc_ecma_transforms:
 - implement regenerator pass
 - add regenerator pass to es2015 preset

swc_ecma_preset_env:
 - add regenerator


benjamn:
@kdy1 As the author/maintainer of https://github.com/facebook/regenerator, I am truly amazed/impressed that you were able to convert that highly dynamic code to Rust!

The reason I decided to check out `swc` again today was that I was waiting for Babel take 105 seconds (all by itself!) to compile a mere 1100 files in one of my larger projects.

The last time I looked at `swc`, you were just getting started, but it's come a LONG way since then. The existence of a regenerator transform has completely changed my mind about the possibility of using `swc` instead of Babel within https://github.com/meteor/meteor (another project I'm heavily involved in).

If there's anything I can do to help with this effort, please don't hesitate to let me know. I'm still learning Rust, but I can definitely answer questions about why things work(ed) the way they do in the JS implementation.

kdy1:
@benjamn Glad to hear that! Porting regenerator was quite hard as rust does not allow two &mut-s.


> The reason I decided to check out swc again today was that I was waiting for Babel take 105 seconds (all by itself!) to compile a mere 1100 files in one of my larger projects.

I think if you don't use babel plugins, swc can be an option. swc is 67x faster on 4core 8HT cpu (intel zeon e3-1275 v2). It uses all core automatically if you use async api. (like transform / transformFile)



> The last time I looked at swc, you were just getting started, but it's come a LONG way since then. Going through the list of supported features, I was sure a regenerator would be one of the transforms that weren't implemented yet, but I'm glad I was wrong about that. The existence of this transform has completely changed my mind about the possibility of using swc instead of Babel within https://github.com/meteor/meteor (another project I'm heavily involved in).

`swc` got improved over time, and it becomes more mature (although I think it's not enough)


> If there's anything I can do to help with this effort, please don't hesitate to let me know. I'm still learning Rust, but I can definitely answer questions about why things work(ed) the way they do in the JS implementation.

Great. I'll ask you when I need help. Thank you very much!

PR #557, opened 2020-01-03
GetterProp - Add type_ann

dsherret:
Adds parsing of a type annotation to `GetterProp`.

This is valid typescript:

```ts
const obj = {
    get test() : string {
        return "";
    }
};
```

https://www.typescriptlang.org/play/?ssl=5&ssc=3&pln=1&pc=1#code/MYewdgzgLgBCBGArGBeGBvAsAKBnmA5gKaxRHQAUAlDAFwzQBOAlmARjvlzIyQK6MwMAETCA3J3wBfHFLFA


ecmascript/parser/tests/typescript/object/getter-prop/input.ts
```diff
@@ -0,0 +1,8 @@
+const obj = {
+    get prop1(): string {
+        return "";
+    },
+    get prop2() {
+        return 5;
+    }
+};
```
dsherret: I should add a newline here.

dsherret: Done.

kdy1:
The pr seems good to me, but I expect swc_ecma_transforms would be broken.  

Can you give me a permission to push to your fork? I'm asking this as I don't like breaking master branch. (Although tracis ci fails, I always check that all tests passes locally)

dsherret:
@kdy1 that would be great! I believe you already have permission. to push to my fork. Thanks and no rush!

kdy1:
@dsherret I didn't know that I have push permission on fork. I fixed `swc_ecma_transforms` and verified that all tests pass.

PR #558, opened 2020-01-03
Fix span for patterns with a type annotation

dsherret:
Fixes the span for patterns with a type annotation.

@kdy1 some tests need to still be run as part of this PR. The problems with me running the tests are because I'm on Windows... for the tests, could we load the files from the disk and replace any \r\n newlines with \n? On a quick glance I couldn't figure out how to do that in `fn with_parser` in tests\typescript.rs (I'm not getting that great code completion on certain things in VSCode with rust sadly).


ecmascript/ast/src/pat.rs
```diff
@@ -63,12 +63,12 @@ pub struct AssignPat {
 /// EsTree `RestElement`
 #[ast_node("RestElement")]
 pub struct RestPat {
+    pub span: Span,
```
dsherret: Added this because the "hi" position depends on the existence of a type annotation.

Issue #548, opened 2019-12-31
TsExprWithTypeArgs should have type_args property instead of type_params


dsherret:
This library is very good at distinguishing type parameters from type arguments ([unlike babel](https://github.com/babel/babel/issues/9418)), but the declaration for `TsExprWithTypeArgs` should have a `type_args` property rather than `type_params` at it currently does:

```rust
#[ast_node("TsExpressionWithTypeArguments")]
pub struct TsExprWithTypeArgs {
    pub span: Span,
    #[serde(rename = "expression")]
    pub expr: TsEntityName,
    #[serde(default)]
    pub type_params: Option<TsTypeParamInstantiation>,
}
```

This is a breaking change of course... looks like it was done on `NewExpr` in the past:

```
#[ast_node("NewExpression")]
pub struct NewExpr {
    pub span: Span,

    pub callee: Box<Expr>,

    #[serde(default, rename = "arguments")]
    pub args: Option<Vec<ExprOrSpread>>,

    #[serde(default, rename = "typeArguments")]
    pub type_args: Option<TsTypeParamInstantiation>,
    // pub type_params: Option<TsTypeParamInstantiation>,
}
```

kdy1:
Although this is breaking change, I fixed it as it is a correctness issue.

dsherret:
Awesome! Happy new year, btw!

kdy1:
@dsherret I fixed it. Happy new year!

PR #560, opened 2020-01-03
[Merged by Bors] - Import type should have type arguments rather than type params

dsherret:
Just found another occurrence of this. Similar to #548. 

image

I'm currently going through all the types right now so you might want to hold off on merging this or doing a breaking change release until I finish that.


kdy1:
I prefer reducing # of commits of master. I'll just wait. Please ping me when you are done.

kdy1:
@dsherret Can you rebase it? I added some github actions to verify that code works.

PR #565, opened 2020-01-04
Add failing test for simplify transform removing necessary braces around nested if-statements

erikdesjardins:
For this input:

```js
if (x) {
  if (y) log(1);
  else if (z) log(2);
} else log(3);
```

swc outputs:

```js
if (x) if (y) log(1);
else if (z) log(2);
else log(3);

// which is equivalent to:

if (x) {
  if (y) log(1);
  else if (z) log(2);
  else log(3);
}
```

It should preserve the braces from the input, since they are necessary to keep the `else` "attached" to the outer `if` statement.

I don't expect this to be merged as-is of course, but hopefully it's enough for someone to fix the issue. Otherwise I'll look into it later when I have time.


kdy1:
Nice catch! I'll fix it.

PR #566, opened 2020-01-04
[Merged by Bors] - Fix class and interface spans

dsherret:
The interface span for something like `interface Test {}` was `Test {}`. Also, for stuff like `export default abstract class Test{}` it was `class Test{}` instead of `abstract class Test {}`.

### DIFF
--- a/ecmascript/parser/src/parser/class_and_fn.rs
+++ b/ecmascript/parser/src/parser/class_and_fn.rs
@@ -45,31 +45,36 @@ impl<'a, I: Tokens> Parser<'a, I> {
         self.parse_fn(None, decorators)
     }
 
-    pub(super) fn parse_class_decl(&mut self, decorators: Vec<Decorator>) -> PResult<'a, Decl> {
-        self.parse_class(decorators)
+    pub(super) fn parse_class_decl(
+        &mut self,
+        start: BytePos,
+        decorators: Vec<Decorator>,
+    ) -> PResult<'a, Decl> {
+        self.parse_class(start, decorators)
     }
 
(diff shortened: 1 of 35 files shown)


ecmascript/parser/src/parser/class_and_fn.rs
```diff
@@ -45,31 +45,43 @@ impl<'a, I: Tokens> Parser<'a, I> {
         self.parse_fn(None, decorators)
     }
 
-    pub(super) fn parse_class_decl(&mut self, decorators: Vec<Decorator>) -> PResult<'a, Decl> {
-        self.parse_class(decorators)
+    pub(super) fn parse_class_decl(
+        &mut self,
+        start: BytePos,
```
dsherret: I copied what the rest of the code was doing by passing in byte positions, but I don't think this is ideal. I think it would be better to get the position within the functions then do assertions for... for example, functions like `parse_interface_decl` could start off with `let start = cur_pos!(); assert_and_bump!("interface")`. Anyway, doing that seemed like too big of a change for this PR.

dsherret:
Here's an example of one of the failing tests I get on my local machine.

image

So I get a lot of noise with failing tests, which makes it difficult to sort through.

kdy1:
@dsherret You need to set autocrlf to input. Tests fail because \r\n is two byte while \n is one byte.

```
git config core.autocrlf input
```

dsherret:
@kdy1 thanks! That worked for me :)

I noticed some problems with my PR now and will fix them tomorrow.

dsherret:
Ok, this is good now. Thanks for the tip about setting the git config!

PR #567, opened 2020-01-05
TypePredicate - Parse the asserts modifier

dsherret:
Start of adding support for the asserts modifier:

```ts
function isString(x: unknown): asserts x is string {
    if (typeof x !== "string")
        throw new Error("Not a string.");
}
```

Start of adding the asserts modifier, but this doesn't compile.

More info: https://devblogs.microsoft.com/typescript/announcing-typescript-3-7/#assertion-functions

### DIFF
--- a/ecmascript/ast/src/typescript.rs
+++ b/ecmascript/ast/src/typescript.rs
@@ -397,6 +397,7 @@ pub struct TsTypeRef {
 #[ast_node("TsTypePredicate")]
 pub struct TsTypePredicate {
     pub span: Span,
+    pub asserts: bool,
     pub param_name: TsThisTypeOrIdent,
     #[serde(rename = "typeAnnotation")]
     pub type_ann: TsTypeAnn,
--- a/ecmascript/parser/src/parser/typescript.rs
+++ b/ecmascript/parser/src/parser/typescript.rs
@@ -1627,6 +1627,17 @@ impl<'a, I: Tokens> Parser<'a, I> {
 
         match *cur!(true)? {
             Token::Word(Word::Ident(..)) | tok!("void") | tok!("null") => {
+                if is!("asserts") && peeked_is!("this") {
+                    let start = cur_pos!();
+                    assert_and_bump!("this");
+                    assert_and_bump!("is");
+                    let this_keyword = self.parse_ts_this_type_node()?;
+                    return self
+                        .parse_ts_this_type_predicate(start, true, this_keyword)
+                        .map(TsType::from)
+                        .map(Box::new);
+                }
+
                 let kind = if is!("void") {
                     Some(TsKeywordTypeKind::TsVoidKeyword)
                 } else if is!("null") {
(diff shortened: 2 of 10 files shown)


ecmascript/parser/src/parser/typescript.rs
```diff
@@ -256,11 +256,15 @@ impl<'a, I: Tokens> Parser<'a, I> {
...
+    fn parse_ts_this_type_predicate(
+        &mut self,
+        start: BytePos,
+        has_asserts_keyword: bool,
+        lhs: TsThisType,
+    ) -> PResult<'a, TsTypePredicate> {
         debug_assert!(self.input.syntax().typescript());
 
         assert_and_bump!("is");
-        let start = cur_pos!();
```
dsherret: Also, will fix this span issue.

ecmascript/parser/src/parser/typescript.rs
```diff
@@ -1687,11 +1698,13 @@ impl<'a, I: Tokens> Parser<'a, I> {
                 return self.parse_ts_import_type().map(TsType::from).map(Box::new);
             }
 
-            tok!("this") => {
+            tok!("this") | tok!("asserts") if is!("this") || peek_is!("this") => {
```
dsherret: How would you recommend checking for "asserts" here and above for `is!("asserts")`? This doesn't compile.

ecmascript/ast/src/typescript.rs
```diff
@@ -397,6 +397,7 @@ pub struct TsTypeRef {
 #[ast_node("TsTypePredicate")]
 pub struct TsTypePredicate {
     pub span: Span,
+    pub asserts: bool,
```
dsherret: Is doing this boolean flag ok?

ecmascript/parser/src/parser/typescript.rs
kdy1: Great

ecmascript/parser/src/parser/typescript.rs
kdy1: You need to add "asserts" to tok! macro.

ecmascript/ast/src/typescript.rs
kdy1: Yes. I think it's enough.

ecmascript/parser/src/parser/typescript.rs
```diff
@@ -400,14 +405,19 @@ impl<'a, I: Tokens> Parser<'a, I> {
         debug_assert!(self.input.syntax().typescript());
 
         self.in_type().parse_with(|p| {
-            let start = cur_pos!();
+            let start = cur_pos!(); // todo: the starts below should not include the return_token
```
dsherret: I believe everything is good in this PR, but it shouldn't use this start for type predicates in functions. I believe this should go after the if statement below, but might be good to fix that in a separate PR.

ecmascript/parser/tests/typescript/function/predicate-types/input.ts.json
```diff
@@ -222,6 +224,113 @@
...
+        "type": "TsTypeAnnotation",
+        "span": {
+          "start": 90,
+          "end": 112,
+          "ctxt": 0
+        },
+        "typeAnnotation": {
+          "type": "TsTypePredicate",
+          "span": {
+            "start": 90,
```
dsherret: As pointed out above, this should be 92, but fixing it causes too many other changes for this PR.

ecmascript/parser/src/parser/typescript.rs
kdy1: I agree with you.

kdy1:
Thank you for the pr!

bors r+

dsherret:
@kdy1 I'm not sure what's going on with the release build test failing so if you have any ideas feel free to push a fix.

PR #577, opened 2020-01-09
Make parser work identically in release mode

kdy1:
This pr makes parser work identically on debug mode and release mode. Although assertions are removed, asssert_and_bump! can be invoked solely.

I also removed a release mode test because it is not required anymore.


kdy1:
I got success when I run tests on my machine.

bors r+ p=100

PR #581, opened 2020-01-11
Fix span of nodes with decorators, parameter properties, and class members with accessibility

dsherret:
* Spans of class members and classes should include decorators.
* Fix hi position for class parameter properties.
* Fix span of members with accessibility.

### DIFF
--- a/ecmascript/parser/src/parser/class_and_fn.rs
+++ b/ecmascript/parser/src/parser/class_and_fn.rs
@@ -295,9 +295,8 @@ impl<'a, I: Tokens> Parser<'a, I> {
     }
 
     fn parse_class_member(&mut self) -> PResult<'a, ClassMember> {
-        let decorators = self.parse_decorators(false)?;
-
         let start = cur_pos!();
+        let decorators = self.parse_decorators(false)?;
 
         if eat!("declare") {
             self.emit_err(self.input.prev_span(), SyntaxError::TS1031);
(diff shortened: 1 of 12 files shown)


dsherret:
I think the spans of function nodes with decorators might be still incorrect (not going to bother with them for this PR as they're not supported in typescript).

ecmascript/parser/src/parser/class_and_fn.rs
```diff
@@ -364,19 +363,18 @@ impl<'a, I: Tokens> Parser<'a, I> {
...
     #[allow(clippy::cognitive_complexity)]
     fn parse_class_member_with_is_static(
         &mut self,
+        start: BytePos,
         accessibility: Option<Accessibility>,
         static_token: Option<Span>,
         decorators: Vec<Decorator>,
     ) -> PResult<'a, ClassMember> {
         let is_static = static_token.is_some();
-        let start = static_token.map(|s| s.lo()).unwrap_or(cur_pos!());
```
dsherret: This start did not include decorators or the accessibility modifier start.

kdy1: You are right. I think I forgot to update it when I added a typescript parser and decorators.

dsherret: Makes sense. Yeah, it's no problem—very easy to fix 🙂

PR #593, opened 2020-01-18
Fix spans for "as" and const assertions

dsherret:
As title.


kdy1 (approved):

kdy1:
Thank you for the pr!!

Issue #106, opened 2018-12-30
wasm?


jhpratt:
Could this potentially be compiled to WebAssembly? This would allow full interop with the existing Babel packages, and would truly thrust this project into the spotlight.

dennythecoder:
You may want to check out https://github.com/swc-project/swc/issues/87

anurbol:
@kdy1 Continuing https://github.com/swc-project/swc/issues/87#issuecomment-563009276... So what are those `cfg`s to build a wasm binary? I've tried to build only the Parser for WASM and got this [error](https://github.com/rust-lang/rust/issues/67170) I am not sure if it relates to SWC or to Rust. Do you have any clues?

kdy1:
@anurbol 

> So what are those cfgs to build a wasm binary? 
`#[cfg]` should be placed on some structs in swc_common. Other codes do not use os-specific features.


> got this error I am not sure if it relates to SWC or to Rust. Do you have any clues?
I think it's rustc's bug related to wasm target. I'm not sure about the way to fix it.

anurbol:
@kdy1 Hey, I've managed to compile the Parser (nothing more) with `--target wasm32-unknown-unknown`! It turns out that dependency renaming is [broken](https://github.com/rust-lang/rust/issues/64450) for the Wasm target. Therefore I did this...
In ecmascript\parser\Cargo.toml 
Changed this:
```
ast = { package = "swc_ecma_ast", version = "0.10.0", path ="../ast" }
parser_macros = { package = "swc_ecma_parser_macros", version = "0.4", path ="./macros" }
```
...to this...
```
swc_ecma_ast = { version = "0.10.0", path ="../ast" }
swc_ecma_parser_macros = { package = "swc_ecma_parser_macros", version = "0.4", path ="./macros" }
```
...and minor things, including replacements throughout the `swc_ecma_parser` crate.

Is it OK? Are the new names OK? Can I make a pull request?

BTW this project is incredible, I am your fan, are you even a human? Because humans can't do such amount of work 😄

kdy1:
@anurbol I'll be happy with a such PR. Thank you for the interest on swc!

anurbol:
[...]

PR #621, opened 2020-02-03
Remove dependency renaming to enable WASM

alubbe:
@anurbol is this what you had in mind for the follow-up PR? https://github.com/swc-project/swc/issues/106#issuecomment-581282353

### DIFF
--- a/ecmascript/parser/Cargo.toml
+++ b/ecmascript/parser/Cargo.toml
@@ -19,7 +19,7 @@ verify = ["fold"]
 swc_atoms = { version = "0.2", path ="../../atoms" }
 swc_common = { version = "0.5.0", path ="../../common" }
 swc_ecma_ast = { version = "0.16.0", path ="../ast" }
-swc_ecma_parser_macros = { package = "swc_ecma_parser_macros", version = "0.4", path ="./macros" }
+swc_ecma_parser_macros = { version = "0.4", path ="./macros" }
 enum_kind = { version = "0.2", path ="../../macros/enum_kind" }
 unicode-xid = "0.2"
 log = { version = "0.4", features = ["release_max_level_debug"] }
--- a/ecmascript/src/lib.rs
+++ b/ecmascript/src/lib.rs
@@ -1,5 +1,5 @@
-pub use ast;
-pub use codegen;
-pub use parser;
-pub use preset_env;
-pub use transforms;
+pub use swc_ecma_ast;
+pub use swc_ecma_codegen;
+pub use swc_ecma_parser;
+pub use swc_ecma_preset_env;
+pub use swc_ecma_transforms;
--- a/ecmascript/src/lib.rs
+++ b/ecmascript/src/lib.rs
@@ -1,5 +1,5 @@
-pub use swc_ecma_ast;
-pub use swc_ecma_codegen;
-pub use swc_ecma_parser;
-pub use swc_ecma_preset_env;
-pub use swc_ecma_transforms;
+pub use swc_ecma_ast as ast;
+pub use swc_ecma_codegen as codegen;
+pub use swc_ecma_parser as parser;
+pub use swc_ecma_preset_env as preset_env;
+pub use swc_ecma_transforms as transforms;
--- a/ecmascript/transforms/Cargo.toml
+++ b/ecmascript/transforms/Cargo.toml
@@ -11,8 +11,8 @@ edition = "2018"
 [dependencies]
 swc_atoms = { version = "0.2.0", path ="../../atoms" }
 swc_common = { version = "0.5.0", path ="../../common" }
-ast = { package = "swc_ecma_ast", version = "0.16.0", path ="../ast" }
-utils = { package = "swc_ecma_utils", version = "0.3.0", path ="../utils" }
+swc_ecma_ast = { version = "0.16.0", path ="../ast" }
+swc_ecma_utils = { version = "0.3.0", path ="../utils" }
 swc_ecma_parser = { version = "0.18", path ="../parser", features = ["verify"] }
 dashmap = "3.2.2"
 either = "1.5"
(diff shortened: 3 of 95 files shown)


ecmascript/src/lib.rs
```diff
@@ -1,5 +1,5 @@
-pub use ast;
-pub use codegen;
-pub use parser;
-pub use preset_env;
-pub use transforms;
+pub use swc_ecma_ast;
+pub use swc_ecma_codegen;
+pub use swc_ecma_parser;
```
kdy1: Please change these to ast, codegen, parser, preset_env, trnasforms, like
`pub use swc_ecma_ast as ast;`

ecmascript/parser/Cargo.toml
```diff
@@ -42,4 +42,4 @@ pretty_assertions = "0.6"
 name = "lexer"
 
 [[example]]
-name = "typescript"
\ No newline at end of file
+name = "typescript"
```
kdy1: If you are using windows, please set autocrlf and change line endings to lf

alubbe: no, it's vs code on a mac

alubbe:
Regarding the failing `fmt` test - I'm using this command, why are the outputs different from what CI would like?

```
cargo +nightly fmt -- --unstable-features --skip-children
```

ecmascript/transforms/Cargo.toml
```diff
@@ -36,4 +36,4 @@ testing = { version = "0.5", path ="../../testing" }
 swc_ecma_codegen = { version = "0.15.0", path ="../codegen" }
 tempfile = "3"
 pretty_assertions = "0.6"
-sourcemap = "4.1.1"
\ No newline at end of file
+sourcemap = "4.1.1"
```
kdy1: Same as above

ecmascript/parser/Cargo.toml
alubbe: it just added a new line to a file that didn't have an empty last line

kdy1:
> Regarding the failing `fmt` test - I'm using this command, why are the outputs different from what CI would like?
> 
> ```
> cargo +nightly fmt -- --unstable-features --skip-children
> ```

You need to add `--all` to run cargo fmt on all crates.

ecmascript/parser/Cargo.toml
kdy1: I got it.

anurbol:
Wow, great work. As kdy already mentioned, yep it seems that you've forgotten to add aliases to the imported names (just like in javascript - you can't just rename imported variable, you should assign an alias in order for things to keep working)

kdy1 (approved):

kdy1:
It seems like `cargo fmt --all` is required

kdy1:
Btw, it's a breaking change as wasm module cannot be loaded synchronously, so transformSync and transformFileSync cannot be used.
How do you think about publishing it with a different name?

tests/serde.rs
```diff
@@ -1,6 +1,6 @@
 use std::path::Path;
 use swc::{
-    ecmascript::{
+    swc_ecmascript::{
```
kdy1: This should be `ecmascript`

alubbe:
I think `swc-wasm` should be fine

examples/usage.rs
```diff
@@ -2,11 +2,11 @@ use swc;
 
 use std::{path::Path, sync::Arc};
 use swc::{
-    common::{
+    config::Options,
+    swc_common::{
```
kdy1: This should be `common`

kdy1 (changes requested):

src/lib.rs
```diff
@@ -1,9 +1,9 @@
 #![feature(box_syntax, box_patterns)]
 
-pub use atoms;
-pub use common;
-pub use ecmascript;
 pub use sourcemap;
+pub use swc_atoms;
```
kdy1: This should be `pub use swc_atoms as atoms`

anurbol:
>wasm module cannot be loaded synchronously

@kdy1 Hi, are you sure? `new WebAssembly.Module()` is synchronous.

kdy1:
> > wasm module cannot be loaded synchronously
> 
> @kdy1 Hi, are you sure? `new WebAssembly.Module()` is synchronous.

Oh.. i didn't know that. Then if wasm's performance is not bad, I'll change @swc/core to use wasm

anurbol:
I don't know if you are currently making and distributing binaries, but for me that's hell. Wasm is so much easier, it's breeze. That said, wasm is 2-3 times slower than native (at least for me, on Windows), but still incredibly faster than JS (in some tasks, e.g. processing strings, x30 faster).

anurbol:
>I've searched about it, and it seems like the size of the synchronous wasm module is limited to 4kb.  Could you share the link to that info? Seems weird. I am loading a wasm module that's a few Mb (if we talk about .wasm file size) and it works fine. 

If you do this, how are you going to sync changes and files? Is there and easy way I am not aware of?

kdy1:
@anurbol https://developers.google.com/web/updates/2018/04/loading-wasm
Hmm... Does the limitation exist only for browser?

anurbol:
@kdy1 I guess the limitation is only for browsers. For node.js that limitation just doesn't make sense.

anurbol:
@kdy1 Also, I didn't even touch the WebAssembly API, I am just using [Wasm-Bindgen](https://github.com/rustwasm/wasm-bindgen) and it generates the bindings, and they are 100% synchronous. And their [book](https://rustwasm.github.io/book/) didn't mention anything about size limit.

kdy1:
@anurbol It sounds great.

alubbe:
It was easier for me to work all these changes into a new PR - so I'm closing this in favour of https://github.com/swc-project/swc/pull/625

PR #625, opened 2020-02-04
[Merged by Bors] - Remove dependency renaming to enable WASM

alubbe:
This PR supercedes https://github.com/swc-project/swc/pull/621 and tries to remove dependency renaming to enable WASM, as discussed in https://github.com/swc-project/swc/issues/106

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,7 +1,7 @@
 use crate::config::{GlobalPassOption, JscTarget, ModuleConfig};
-use atoms::JsWord;
-use common::{chain, errors::Handler, fold::and_then::AndThen, SourceMap};
-use ecmascript::{
+use swc_atoms::JsWord;
+use swc_common::{chain, errors::Handler, fold::and_then::AndThen, SourceMap};
+use swc_ecmascript::{
     parser::Syntax,
     preset_env,
     transforms::{
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,4 +1,7 @@
 use crate::config::{GlobalPassOption, JscTarget, ModuleConfig};
+use either::Either;
+use hashbrown::hash_map::HashMap;
+use std::sync::Arc;
 use swc_atoms::JsWord;
 use swc_common::{chain, errors::Handler, fold::and_then::AndThen, SourceMap};
 use swc_ecmascript::{
@@ -10,9 +13,6 @@ use swc_ecmascript::{
         typescript,
     },
 };
-use either::Either;
-use hashbrown::hash_map::HashMap;
-use std::sync::Arc;
 
 /// Builder is used to create a high performance `Compiler`.
 pub struct PassBuilder<'a, 'b, P: Pass> {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,8 +1,8 @@
 #![feature(box_syntax, box_patterns)]
 
-pub use atoms;
-pub use common;
-pub use ecmascript;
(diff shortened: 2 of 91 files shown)


src/lib.rs
```diff
@@ -28,13 +31,10 @@ use ecmascript::{
         util::COMMENTS,
     },
 };
-pub use ecmascript::{
+pub use swc_ecmascript::{
```
kdy1: Is this required? I think ecmascript:: would work because `pub use swc_ecmascript as ecmascript` exists

src/builder.rs
```diff
@@ -1,7 +1,10 @@
 use crate::config::{GlobalPassOption, JscTarget, ModuleConfig};
-use atoms::JsWord;
-use common::{chain, errors::Handler, fold::and_then::AndThen, SourceMap};
-use ecmascript::{
+use either::Either;
+use hashbrown::hash_map::HashMap;
+use std::sync::Arc;
+use swc_atoms::JsWord;
+use swc_common::{chain, errors::Handler, fold::and_then::AndThen, SourceMap};
+use swc_ecmascript::{
```
kdy1: I want swc to reexport subcrate and export it like `swc::common`, `swc;;ecmascript::ast`. Also, I prefer common over swc_common.

kdy1:
Btw, thank you very much for such pr1

src/lib.rs
alubbe: you're right, it's not needed

src/builder.rs
alubbe: sure thing - how can that be done? :D

alubbe:
Regarding your re-export comment, how about we merge the PR as it is right now and then you can apply the clean up you have in mind? I admit I'm not super familiar with the best practices around that

kdy1:
> Regarding your re-export comment, how about we merge the PR as it is right now and then you can apply the clean up you have in mind? I admit I'm not super familiar with the best practices around that

I'll add some commits to this pr and merge it.

alubbe:
sounds great!

src/builder.rs
kdy1: Oh.. Renaming is not possible at here without extern crate. Sorry for my mistake.

I thought it is possible because `extern crate`-s exist before changing to toml-based renaming

kdy1:
It was my misunderstanding. With es2018 rust, the change I requested cannot be done.I'm merging the pr as-is.

bors r+

kdy1:
I started working on a wasm version of swc. Thanks!

kdy1:
@anurbol  Can you tell me how did you published a wasm module, and benchmakred it?

anurbol:
Hi, by "published" do you mean "built"? If so, I used some tutorial on the web, can't remember which one, but my current build process is as simple as: 
```
cargo build --target wasm32-unknown-unknown
// For below command to work run `cargo install wasm-bindgen-cli`
wasm-bindgen target/wasm32-unknown-unknown/release/my_app.wasm --out-dir ../build-wasm --nodejs
```

Then in the `build-wasm` directory you'll find the bindings, which are simply ready-to-use JS functions.

kdy1:
@anurbol I built it with wasm-pack and wasm-bindgen, but generated file contains`import * as wasm from './swc_wasm_bg.wasm';` thus I can't require it. 

I'll try your command. Thanks!!

anurbol:
@kdy1 See the last line to my prev comment (just added it)

PR #628, opened 2020-02-05
[Merged by Bors] - Fix start position on class declarations with decorators

### LINKED ISSUE #626: Support top level await
**Describe the feature**

Would be nice to now support top level await:

```ts
async function f() {
}
await f();
```

I believe this would be a feature that isn't compiled and is only supported when targeting `es2017`.

https://devblogs.microsoft.com/typescript/announcing-typescript-3-8-beta/#top-level-await

dsherret:
Fixes: https://github.com/swc-project/swc/issues/626#issuecomment-582567305

See #581

### DIFF
--- a/ecmascript/parser/src/parser/stmt.rs
+++ b/ecmascript/parser/src/parser/stmt.rs
@@ -92,7 +92,6 @@ impl<'a, I: Tokens> Parser<'a, I> {
         top_level: bool,
         decorators: Vec<Decorator>,
     ) -> PResult<'a, Stmt> {
-        let start = cur_pos!();
         if top_level && is!("await") {
             let valid = self.target() >= JscTarget::Es2017 && self.syntax().top_level_await();
 
--- a/ecmascript/parser/tests/typescript/class/decorators/input.ts.json
+++ b/ecmascript/parser/tests/typescript/class/decorators/input.ts.json
@@ -21,7 +21,7 @@
       },
       "declare": false,
       "span": {
-        "start": 5,
+        "start": 0,
         "end": 196,
         "ctxt": 0
       },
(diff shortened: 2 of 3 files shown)


ecmascript/parser/src/parser/stmt.rs
```diff
@@ -92,7 +92,6 @@ impl<'a, I: Tokens> Parser<'a, I> {
         top_level: bool,
         decorators: Vec<Decorator>,
     ) -> PResult<'a, Stmt> {
-        let start = cur_pos!();
```
dsherret: See line 90.

kdy1:
Thanks!

dsherret:
Thanks @kdy1 and that's awesome you implemented top level await so fast!

Would it be possible to get a release for swc_ecma_parser sometime you find the chance?

kdy1:
@dsherret I publised it. Note that you need to update swc_common too.

dsherret:
Thanks a lot @kdy1 !

I'm wondering how to get the byte position and comment vector pairs out of the `RefMulti` in the `DashMap` though. Any ideas? If not, I will sadly clone for now.

```rust
let (leading_comments, trailing_comments) = comments.take_all();
// previously did something like this:
let leading_comments: HashMap<BytePos, Vec<Comment>>
    = leading_comments.into_iter().collect();
```

kdy1:
@dsherret 
DashMap does not support taking all value, but you can use `iter_mut` with `std::mem::replace`.
Iterate over a dashmap with iter_mut(), and replace the value with
```rust
replace( item..value_mut(), Default::default())
```

dsherret:
@kdy1 awesome idea. Thanks! 🙂

dsherret:
Hmmm... couldn't figure it out then tried to just use `CommentMap` in my code to see how it performs, but ran into some issues. I just went back to cloning for now :(

Only slightly slower so no big deal! I'll look into this later.

PR #630, opened 2020-02-06
[Merged by Bors] - Remove empty expression in expression statement with top level await

dsherret:
Eats the semi-colon.

This doesn't compile (edit: doesn't compile the code in the tests) though because the target of that folder is ES2015. Also, this is kind of strange for me to put the tests in the "typescript" folder. Should we create something more general?

You can take over this PR if you'd like!


dsherret:
@kdy1 do you have any suggestions for resolving this failed test:

```
========== Running reference test await-expression/input.ts
Source:
await    test   ;


/home/runner/work/swc/swc/ecmascript/parser/tests/typescript/await-expression/input.ts.json:/home/runner/work/swc/swc/ecmascript/parser/target/swc-test-results/ui/tests/typescript/await-expression/input.ts.json
thread 'typescript::reference::await-expression/input.ts' panicked at 'called `Result::unwrap()` on an `Err` value: error: top level await requires target to es2017 or higher and topLevelAwait:true for ecmascript
 --> $DIR/tests/typescript/await-expression/input.ts:1:1
  |
1 | await    test   ;
  | ^^^^^

', ecmascript/parser/tests/typescript.rs:111:17
```

I feel like these AST parsing tests should not be dependent on the specified target and any target specific tests should live elsewhere. If I remove this requirement though then another test fails that checks for this error.

kdy1:
Hmm... 🤔 
I can move this to top crate (swc), to use swcrc
Do you mind me adding some commits to your pr?

dsherret:
@kdy1 for sure. Please!

PR #641, opened 2020-02-09
[Merged by Bors] - Strip type-only imports

### LINKED ISSUE #640: Typescript is not properly stripped out
**Input code**
```ts
import { Handler } from 'aws-lambda';
export const handler: Handler = async (event, context) => {};
```

**Config**
```json
{
  "test": ".*.(m?js|ts)?$",
  "env": {
    "targets": {
      "node": "12"
    }
  },
  "jsc": {
    "target": "es2016",
    "parser": {
      "syntax": "typescript"
    }
  }
}
```

**Expected behavior**

Types should be stripped out.

kdy1:
Fixes #640

### DIFF
--- a/ecmascript/transforms/src/typescript.rs
+++ b/ecmascript/transforms/src/typescript.rs
@@ -2,12 +2,11 @@ use crate::{
     pass::Pass,
     util::{prepend_stmts, var::VarCollector, ExprFactory},
 };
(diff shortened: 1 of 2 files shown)


kdy1:
It seems like rustfmt is not included in the current nightly.

bors r+

PR #642, opened 2020-02-09
Support ecmascript private fields when parsing TypeScript

dsherret:
As title.


dsherret:
Some background: https://devblogs.microsoft.com/typescript/announcing-typescript-3-8-rc/

simnalamburt:
I'm not the project maintainer, but it seems obvious and a good fix!

kdy1:
bors retry

PR #643, opened 2020-02-09
Fix 'fmt' GitHub Actions workflow

simnalamburt:


simnalamburt:
Oops Looks like that fix wasn't enough. I'll fix it

PR #662, opened 2020-02-13
type-only import, exports

### LINKED ISSUE #649: Support type only imports and exports (TS 3.8)
**Describe the feature**

The TS 3.8 release will happen in a couple of weeks. It has a new type only imports and exports feature:

https://devblogs.microsoft.com/typescript/announcing-typescript-3-8-rc/#type-only-imports-exports

Probably:

* For parsing, the only trouble might be differentiating between a type alias and a type-only export (ex. `export type `). I think it should be fairly straightforward for import declarations.
* For transforming, these should just be stripped from the output.

kdy1:
Closes #649.


kdy1:
@dsherret Can you check if this syntax is correct? I searched test case from babel but I failed to found one.

PR #667, opened 2020-02-15
Support type predicate "asserts" with no "is"

dsherret:
Supports parsing code that looks like the following:

```ts
class Test {
    test(): asserts this { // apparently valid typescript code...
    }
}

function assert(condition: any, msg?: string): asserts condition {
  if (!condition) {
    throw new Error(msg);
  }
}
```

https://ts-ast-viewer.com/#code/MYGwhgzhAEAqCmEAu0DeAoaXpMUgFAJQBc0kE8ATkjEgBYCWMG20Avuh+gGYCuAdsCQMA9vzJQqBYGIAmDYWNJh+ATwA00ALYQA5gH5SySg366SEitRgz+8xeJbQG3aPgCEt+6P6E0mbHpKEQB3aH54MIBRSmDKfB1zAG4AjjYgA

https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=10&pc=2#code/MYGwhgzhAEAqCmEAu0DeBYAUNH0mKQAoBKALmkgngCckYkALASxg21wF8svMsAzAK4A7YEiYB7IRSg0iwSQBMmYyeTBCAngBpoAWwgBzAPzlk1JkINlpVWjHlClKqWxxM+0QgEIHTiUOI0LFw8BmpxAHdoIXgogFFqcOpCfSsAbmDoHg4gA

### DIFF
--- a/ecmascript/ast/src/typescript.rs
+++ b/ecmascript/ast/src/typescript.rs
@@ -400,7 +400,7 @@ pub struct TsTypePredicate {
     pub asserts: bool,
     pub param_name: TsThisTypeOrIdent,
     #[serde(rename = "typeAnnotation")]
-    pub type_ann: TsTypeAnn,
+    pub type_ann: Option<TsTypeAnn>,
 }
 
 #[ast_node]
--- a/ecmascript/parser/src/parser/typescript.rs
+++ b/ecmascript/parser/src/parser/typescript.rs
@@ -421,7 +421,9 @@ impl<'a, I: Tokens> Parser<'a, I> {
                 cur!(false)?;
             }
 
-            let has_type_pred_is = is!(IdentRef) && peeked_is!("is") && !p.input.has_linebreak_between_cur_and_peeked();
+            let has_type_pred_is = is!(IdentRef)
+                && peeked_is!("is")
+                && !p.input.has_linebreak_between_cur_and_peeked();
             let is_type_predicate = has_type_pred_asserts || has_type_pred_is;
             if !is_type_predicate {
                 return p.parse_ts_type_ann(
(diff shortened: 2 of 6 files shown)


ecmascript/ast/src/typescript.rs
```diff
@@ -400,7 +400,7 @@ pub struct TsTypePredicate {
     pub asserts: bool,
     pub param_name: TsThisTypeOrIdent,
     #[serde(rename = "typeAnnotation")]
-    pub type_ann: TsTypeAnn,
+    pub type_ann: Option<TsTypeAnn>,
```
dsherret: Copied from compiler API:

https://github.com/microsoft/TypeScript/blob/7cc4a8df9482ffdfa6b3500a009c0454681d5f4b/lib/typescript.d.ts#L794

ecmascript/parser/src/parser/typescript.rs
```diff
@@ -456,19 +459,6 @@ impl<'a, I: Tokens> Parser<'a, I> {
         })
     }
 
-    fn parse_ts_type_predicate_prefix(&mut self) -> PResult<'a, Option<Ident>> {
```
dsherret: I ended up removing this because I needed to check for the `is` keyword above and ended up changing the code to clearly return when it's not a type predicate.

kdy1:
ThNk you for the pr!
Can you run cergo fmt --all ?

dsherret:
Hmmm... works on my machine. Seems like there's some version mismatch on the CI.

kdy1:
@dsherret  Oh.. you are right.

kdy1:
I've published a new version.

dsherret:
Exciting!! Thanks @kdy1!

PR #677, opened 2020-02-19
Fix tagged template span to include tag

dsherret:
As title.


kdy1:
I've published a new version

dsherret:
Oh, sweet. Thanks @kdy1!

PR #691, opened 2020-02-26
Wasm

kdy1:
I've managed to implement `transformSync`and `trasnformFileSync` within a wasm module.
Currently, it does not have same api as `@swc/core` and only two methods are implemented.

## Benchmark result

(Note: es2015 = es5 as there's no compiler pass to convert es2015 code to es5 code)
```
[transform]
  swc (es3) x 391 ops/sec ±2.03% (62 runs sampled)
  swc-wasm (es3) x 94.30 ops/sec ±11.77% (65 runs sampled)
  swc (es2015) x 623 ops/sec ±0.65% (86 runs sampled)
  swc-wasm (es2015) x 151 ops/sec ±0.45% (81 runs sampled)
  swc (es2016) x 1,768 ops/sec ±0.96% (88 runs sampled)
  swc-wasm (es2016) x 526 ops/sec ±0.60% (88 runs sampled)
  swc (es2017) x 1,778 ops/sec ±0.57% (89 runs sampled)
  swc-wasm (es2017) x 535 ops/sec ±0.54% (86 runs sampled)
  swc (es2018) x 2,363 ops/sec ±1.34% (89 runs sampled)
  swc-wasm (es2018) x 745 ops/sec ±0.62% (87 runs sampled)
  swc-optimize (es3) x 543 ops/sec ±0.59% (87 runs sampled)
  babel (es5) x 35.56 ops/sec ±5.80% (53 runs sampled)
```

It's 3 ~ 4 times slower than the prebuilt binary.

Btw, I'm not sure about the way to implement async apis.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -7,3 +7,9 @@ rustflags = [
 rustdocflags = [
     "--cfg", "procmacro2_semver_exempt",
 ]
+
+[target.wasm32-unknown-unknown]
+rustflags = [
+    "--cfg", "procmacro2_semver_exempt",
+    "-C", "link-args=-z stack-size=16777216",
+]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["ecmascript/visit"]
+members = ["ecmascript/visit", "wasm"]
 
 [package]
 name = "swc"
--- a/ecmascript/transforms/tests/es2015_parameters.rs
+++ b/ecmascript/transforms/tests/es2015_parameters.rs
@@ -6,37 +6,37 @@
 use swc_common::{chain, Mark};
 use swc_ecma_parser::Syntax;
 use swc_ecma_transforms::{
-  compat::{
-    es2015::{arrow, block_scoping, destructuring, parameters, Classes},
-    es2017::async_to_generator,
-  },
-  modules::common_js::common_js,
-  pass::Pass,
-  resolver,
+    compat::{
+        es2015::{arrow, block_scoping, destructuring, parameters, Classes},
+        es2017::async_to_generator,
+    },
+    modules::common_js::common_js,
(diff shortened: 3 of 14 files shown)


vovaguguiev:
Please land this PR, it's such a game-changer!

kdy1:
@vovaguguiev Can you tell me your usecase? I think x3 slowdown is not a simple issue.

Note: As typical usage of `swc` involves multiple promises, swc is 60+ times faster than babel.
However, wasm version does not support such threading. So it's only 6x times faster than babel.

vovaguguiev:
I would like to use it in a Cloudflare Worker, the options I have there are either JS or WASM. Even 5x speed up comparing to Babel is a great improvement for my use case because of workers’ limits on CPU time per request (50ms)

vovaguguiev:
The alternative I am exploring right now is to try and use swc directly from rust code that I compile to wasm, I could write a custom transformation, that I need, directly in Rust similar to how transform-react and other are implemented. Do you think it’s a viable option?

kdy1:
@vovaguguiev Oh... Fair point.
I forgot environments where native binaries are restricted.
I'll start working on this right now.

kdy1:
@vovaguguiev I'm pinging you because it's almost done.

I only implemented `transformSync`, `parseSync` and `printSync` at the moment. Is it enough for your use case?

I can't make apis related to file because wasm does **not** support it and I'm not sure if threading is supported in wasm.

vovaguguiev:
@kdy1 Thanks a lot! I think sync versions are enough for my case. Look forward to trying it!

kdy1:
Published `@swc/wasm@0.1.0`

whatnickcodes:
@vovacodes did you ever see success with this in a Cloudflare worker? Curious to know what kind of performance / use case you had?

PR #693, opened 2020-02-29
Fix assignment pattern start when there is a type annotation

dsherret:
The assignment pattern start didn't include the left expression start sometimes.

Also, upgraded dashmap because there is apparently a way to into_iter().collect() now, so I will be able to eliminate my comment cloning 🕺. Let me know if you'd like a separate PR for this though.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ once_cell = "1"
 hashbrown = "0.6"
 regex = "1"
 either = "1"
-dashmap = "=3.4.0"
+dashmap = "=3.5.1"
 sourcemap = "5"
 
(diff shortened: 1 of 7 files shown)


ecmascript/parser/tests/test262.rs
```diff
@@ -321,15 +321,13 @@ where
     output
 }
 
-#[test]
```
dsherret: Oops... my local change because I'm too lazy to set this up properly. Will remove.

kdy1:
@dsherret Thank you for the pr! I've published a new version

dsherret:
Thanks @kdy1!!

PR #712, opened 2020-03-10
Fix arrow start in conditional expression

dsherret:
Also, fixed `optional` being serialized as `"elements"`


ecmascript/ast/src/pat.rs
```diff
@@ -35,8 +35,8 @@ pub struct ArrayPat {
     #[serde(rename = "elements")]
     pub elems: Vec<Option<Pat>>,
 
-    /// Only in .d.ts file
```
dsherret: Changed this comment because it doesn't necessarily have to be in a .d.ts file.

kdy1:
Thanks!! I were too busy and I couldn't review my own patch throughtfully

PR #714, opened 2020-03-10
Fix tagged template start to include tag

dsherret:
Regression of https://github.com/swc-project/swc/pull/677/files

### DIFF
--- a/ecmascript/parser/src/parser/expr.rs
+++ b/ecmascript/parser/src/parser/expr.rs
@@ -776,9 +776,8 @@ impl<'a, I: Tokens> Parser<'a, I> {
         tag: Box<Expr>,
         type_params: Option<TsTypeParamInstantiation>,
     ) -> PResult<'a, TaggedTpl> {
-        let start = tag.span().lo();
+        let tagged_tpl_start = tag.span().lo();
         trace_cur!(parse_tagged_tpl);
-        let start = cur_pos!();
 
         assert_and_bump!('`');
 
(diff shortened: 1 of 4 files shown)


ecmascript/parser/src/parser/expr.rs
```diff
@@ -776,17 +776,16 @@ impl<'a, I: Tokens> Parser<'a, I> {
         tag: Box<Expr>,
         type_params: Option<TsTypeParamInstantiation>,
     ) -> PResult<'a, TaggedTpl> {
-        let start = tag.span().lo();
+        let tagged_tpl_start = tag.span().lo();
```
dsherret: I'm thinking the key is that we should start giving these more descriptive names to prevent these regressions. `start` is probably too generic.

dsherret:
Sorry, I was trying to test the changes earlier, but I had to do a bit of a refactoring because `TokenAndSpan` no longer implemented `Spanned` (no big deal, I ended up refactoring everything to use `SpanData`, which is probably more efficient).

Anyway, everything looks good now after this PR!

ecmascript/parser/src/parser/expr.rs
kdy1: I agree. It's hard to know which position is pointed by `start`.
I'll use some explanatory name in the future.

kdy1:
I've published a new version. Thanks for the pr!

dsherret:
Thanks!

PR #715, opened 2020-03-10
Allow optional binding pattern in ambient context

### LINKED ISSUE #709: Error parsing optional binding pattern parameter in an ambient context
**Input code**

```ts
declare namespace __event {
    export class EventInit {
        constructor({
            bubbles,
            cancelable,
            composed
        }?: {
            bubbles?: boolean | undefined;
            cancelable?: boolean | undefined;
            composed?: boolean | undefined;
        });
    }
}
```

**Expected behavior**

No error, as this is not an implementation signature since this is an ambient context.

**Actual behavior**

Has diagnostic: `A binding pattern parameter cannot be optional in an implementation signature.`

**Version**

kdy1:
Fix #709, really

### DIFF
--- a/ecmascript/parser/src/parser/object.rs
+++ b/ecmascript/parser/src/parser/object.rs
@@ -373,7 +373,7 @@ impl<'a, I: Tokens> ParseObject<'a, Pat> for Parser<'a, I> {
             }
         }
 
-        let optional = self.input.syntax().dts() && eat!('?');
+        let optional = (self.input.syntax().dts() || self.ctx().in_declare) && eat!('?');
 
         Ok(Pat::Object(ObjectPat {
             span,
--- a/ecmascript/parser/src/parser/pat.rs
+++ b/ecmascript/parser/src/parser/pat.rs
@@ -111,7 +111,7 @@ impl<'a, I: Tokens> Parser<'a, I> {
         }
 
         expect!(']');
-        let optional = self.input.syntax().dts() && eat!('?');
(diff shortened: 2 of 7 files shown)


ecmascript/parser/src/parser/pat.rs
```diff
@@ -111,7 +111,7 @@ impl<'a, I: Tokens> Parser<'a, I> {
         }
 
         expect!(']');
-        let optional = self.input.syntax().dts() && eat!('?');
+        let optional = (self.input.syntax().dts() || self.ctx().in_declare) && eat!('?');
```
dsherret: Does around line 162 need to be updated to get rid of the syntax error? I think same in parser/expr.rs around 1230?

ecmascript/parser/src/parser/object.rs
```diff
@@ -373,7 +373,7 @@ impl<'a, I: Tokens> ParseObject<'a, Pat> for Parser<'a, I> {
             }
         }
 
-        let optional = self.input.syntax().dts() && eat!('?');
+        let optional = (self.input.syntax().dts() || self.ctx().in_declare) && eat!('?');
```
dsherret: +1. Thanks! 🙂

ecmascript/parser/src/parser/pat.rs
kdy1: Oh.. Thanks. I added bunch of error recovery logic, but I completely forgot about it.
(I'm too busy due to coronavirus...)

PR #735, opened 2020-03-28
Fix deeply nested optional chaining

### LINKED ISSUE #732: Optional chaining bug
Optional chaining doesn't catch the scenario where a parent key is optional, but child keys are required.

**Example:**

```ts
type NestedType = {
  a?: {
    b: {
      c: {
        d: string;
      };
    };
  };
};

const test: NestedType = {};

console.log(test.a?.b.c.d);
```

**Expected output**

This is what Typescript 3.8 produces (based on a default config with no specific target):

```js
"use strict";
var _a;
const test = {};
console.log((_a = test.a) === null || _a === void 0 ? void 0 : _a.b.c.d);
```

**But `swc` produces:**

```js
var ref;
var test = {
};
console.log(((ref = test.a) === null || ref === void 0 ? void 0 : ref.b.c).d);
```

Running the output, I'd expect to see:

```
undefined
```

Instead, `swc`, produces:

```
console.log(((ref = test.a) === null || ref === void 0 ? void 0 : ref.b.c).d);
                                                                           ^

TypeError: Cannot read property 'd' of undefined
    at Object.<anonymous> (<file>:4:76)
```

kdy1:
- Handle deeply nested optional chaining expression correctly (Closes #732)

### DIFF
--- a/ecmascript/parser/src/lib.rs
+++ b/ecmascript/parser/src/lib.rs
@@ -112,7 +112,7 @@ mod parser;
 pub mod token;
 
 #[derive(Debug, Clone, Copy, Deserialize, Serialize)]
-#[serde(tag = "syntax")]
+#[serde(deny_unknown_fields, tag = "syntax")]
 pub enum Syntax {
     /// Standard
     #[serde(rename = "ecmascript")]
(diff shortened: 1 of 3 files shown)


leebenson:
thank you @kdy1 !

PR #744, opened 2020-04-02
Do not depend on hashbrown

### LINKED ISSUE #703: swc_common depends on two versions of ahash
swc_common 0.5.8 depends on dashmap and hashbrown each depend on different versions of ahash.

kdy1:
Closes #703


kdy1 (commented):
I verified that there's no breaking change

Issue #690, opened 2020-02-25
Preserve copyright comments on minify


ptman:
**Describe the feature**

/*! Copyright (c) 2020 Foo */ should not be removed when minifying

**Babel plugin or link to the feature description**

https://stackoverflow.com/questions/11248363/the-purpose-of-starting-an-initial-comment-with-in-javascript-and-css-files

PR #753, opened 2020-04-21
Preserve commits starting with "!" when minifying (#690)

FabianWolff:
Hi! I'm a first-time contributor to this project, so I decided to try to tackle one of the easy issues currently open; specifically, #690.

Comments are now always lexed by `Compiler::process_js`, but when the `minify` configuration option is set, all comments except the ones starting with "!" are discarded.

### DIFF
--- a/common/src/comments.rs
+++ b/common/src/comments.rs
@@ -73,6 +73,14 @@ impl Comments {
         }
     }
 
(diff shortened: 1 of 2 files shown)


kdy1:
Thank you very much!!

Issue #750, opened 2020-04-18
Support decorators on method parameters


dsherret:
It seems like swc does not support decorators on parameters in methods and I forgot about this as well.

The following is valid TypeScript code:

```ts
class MyClass {
    myMethod(@myDecorator myParameter: string) {}
}
```

But the following error happens:

```
Line 2, column 14: Unexpected token Some(At)
```

Perhaps it's time for a `Parameter` AST node? https://ts-ast-viewer.com/#code/MYGwhgzhAECyCeBhcVoG8BQ1vQLb1gFMAXACwHsATACgAF8ARQ4cgJzGLb3gAUx3cJQqwBc0CMVYBLAHYBzAJToAvhmVA

kdy1:
Is this valid in ecmascript? I failed to find any reference... It seems like valid only in typescript.

dsherret:
It’s only valid TypeScript at the moment.

PR #765, opened 2020-05-09
Support for parsing parameter decorators

dsherret:
I took a quick stab at an initial implementation in hopes that we could start working from here. Would really like to get this in swc.

Note: This is a parsing only commit. I have not looked at transformations at all.

Do you have any suggestion for the AST?

**TypeScript**

image

**ESTree**

image

For #750

### DIFF
--- a/ecmascript/ast/src/class.rs
+++ b/ecmascript/ast/src/class.rs
@@ -1,6 +1,6 @@
 use crate::{
     expr::Expr,
-    function::{Function, PatOrTsParamProp},
+    function::{Function, ParamOrTsParamProp},
     ident::PrivateName,
     prop::PropName,
     stmt::BlockStmt,
@@ -151,7 +151,7 @@ pub struct Constructor {
 
     pub key: PropName,
 
-    pub params: Vec<PatOrTsParamProp>,
+    pub params: Vec<ParamOrTsParamProp>,
 
     #[serde(default)]
     pub body: Option<BlockStmt>,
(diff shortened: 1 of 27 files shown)


yharaskrik:
Any status on this? I discovered SWC and played around building our Nest app but we obviously have parameters in a lot of function calls (in Controllers).

kdy1:
@yharaskrik
Hmm... I'll do this at the weekend. It's not a hard issue, but I postponed it mainly because I didn't know any use case of annotations on parameters.

yharaskrik:
I'm not aware of anywhere else that uses it other that Nest and Angular. Both use decorators inside function params and constructors.

kdy1:
Hmm... It's hard to determine where to store decorators.
I think creating a new type `Param` would be better than adding decorators to ident / array pat / etc...

kdy1:
Closing in favor of #775

dsherret:
@kdy1 thanks for doing the transformation work on this! Would you be able to add me on that commit as a co-author for doing the parsing and investigation work? It seems github didn’t keep track of it https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

kdy1:
@dsherret I did it and [seems like it's applied](https://github.com/swc-project/swc/commits/master)

(I don't think force pushing is a good idea, but I think it is a correct thing to do)

dsherret:
@kdy1 awesome. Thanks so much!

yharaskrik:
Amazing! Thank you! Once a release is cut with it I'll give it another try!

PR #772, opened 2020-05-15
Directly map the box contents without unsafe

95th:
We can directly map box's contents without using unsafe with no difference in generated assembly.


kdy1:
Great! Thank you.

PR #777, opened 2020-05-17
Rename Import and Export Node names

### LINKED ISSUE #536: Import and export specifier node names
I wanted to make the following recommendations on the AST node names for import specifiers:

* `ImportDefault` -> `ImportDefaultSpecifier`
* `ImportStarAs` -> `ImportStarAsSpecifier` (or `ImportNamespaceSpecifier` as babel calls it?)
* `ImportSpecific` -> `ImportNamedSpecifier`

The comments for these should also probably be updated (ex. on `ImportDefault` it should go from `/// e.g. `import foo from 'mod.js'`` to `/// e.g. `foo` in `import foo from 'mod.js'``).

Note that the export specifier names should probably be switch to be consistent as well if we decide to do this:

* `NamespaceExportSpecifier` -> `ExportNamespaceSpecifier`
* `DefaultExportSpecifier` -> `ExportDefaultSpecifier`
* `NamedExportSpecifier` -> `ExportNamedSpecifier`

95th:
Closes #536


kdy1:
May I add some commit to your branch?
As it's breaking change, I need to bump versions.

95th:
Sure.. Please go ahead.

PR #795, opened 2020-05-23
Remove comma separators in benchmark presentation

Veetaha:
By first looking at this table I was confused of how could you advertise 2.1-2.9 ops/sec performance, only time later I figured out that the comma is in fact the separator for thousands.
I'd suggest not putting comma here in order to prevent confusion especially for people from countries where commas are used as floating point separators.


kdy1:
I didn't think that in some countries a comma is used to represent float. Thank you.

PR #805, opened 2020-05-28
Fix new reserved_words pass

### LINKED ISSUE #804: Object property default renamed to _default
**Describe the bug**

If an item has a property named default, it is renamed to _default when code is compiled.

**Input code**
repo with example https://github.com/kyle-mccarthy/swc-default-rename

```typescript
export const config = {
  someKey: {
    default: 1,
    arg: 'someKey',
  },
};
```

**Config**
```json
{
  "test": ".*.ts$",
  "jsc": {
    "parser": {
      "syntax": "typescript"
    }
  },
  "module": {
    "type": "commonjs",
    "noInterop": false
  }
}
```

**Output code**

```javascript
'use strict';
Object.defineProperty(exports, '__esModule', {
    value: true
});
exports.config = void 0;
var config = {
    someKey: {
        _default: 1,
        arg: 'someKey'
    }
};
exports.config = config;
```

**Expected behavior**

Object property/key with name default should not be renamed

**Version**

kdy1:
Closes  #804

### DIFF
--- a/ecmascript/transforms/src/compat/es3/reserved_word.rs
+++ b/ecmascript/transforms/src/compat/es3/reserved_word.rs
@@ -59,12 +59,6 @@ impl Fold<ImportNamedSpecifier> for ReservedWord {
     }
 }
 
-impl Fold<Module> for ReservedWord {
-    fn fold(&mut self, node: Module) -> Module {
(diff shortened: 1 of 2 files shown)


kdy1:
Merging as local test passed.

PR #810, opened 2020-05-30
Remove unused swc_ecma_codegen Handlers

peterwmwong:
While using `swc_ecma_codegen` to generate JS source, I noticed the `Emitter` required a `swc_ecma_codegen::Handlers` param that didn't seemed to be used (only referenced in commented out code from the [original codegen commit](https://github.com/swc-project/swc/pull/40/files#diff-016536858b54a2765fd0e5894ef94bba)).

Removal will make using `swc_ecma_codegen` easier (no need to define/impl an empty `Handler`, less cognitive load) for consumers of `swc_ecma_codgen` and reduce API surface area to maintain for swc developers.


kdy1:
I'll merge this pr when I make a breaking change.

kdy1:
Closing as I will use handlers for testing

PR #815, opened 2020-06-02
Implement codegen of `export foo as Foo from foo;`

vovacodes:
I'm a complete newbie in Rust so please take a close look to these changes and let me know if the code is not written in an idiomatic way 😀


kdy1:
Thanks for the pr!

kdy1:
Published a new version

PR #817, opened 2020-06-03
Fix type-only import declaration with empty named imports being parsed as having a default import specifier

dsherret:
As title.

Sorry, I missed this until now.

### DIFF
--- a/ecmascript/parser/src/parser/stmt/module_item.rs
+++ b/ecmascript/parser/src/parser/stmt/module_item.rs
@@ -60,8 +60,13 @@ impl<'a, I: Tokens> Parser<'a, I> {
             .map(ModuleItem::from);
         }
 
-        let type_only_span = self.input.cur_span();
-        let type_only = self.syntax().typescript() && eat!("type");
+        let type_only = self.input.syntax().typescript()
+            && is!("type")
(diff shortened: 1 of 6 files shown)


ecmascript/parser/src/parser/stmt/module_item.rs
```diff
@@ -60,8 +60,13 @@ impl<'a, I: Tokens> Parser<'a, I> {
             .map(ModuleItem::from);
         }
 
-        let type_only_span = self.input.cur_span();
-        let type_only = self.syntax().typescript() && eat!("type");
+        let type_only = self.input.syntax().typescript()
+            && is!("type")
+            && (peeked_is!('{') || !peeked_is!("from") && !peeked_is!(','));
```
dsherret, line 65: Wait... `import type from from "./a"` 🤯

That's actually a bug in the typescript compiler.

dsherret: I'm not sure about a way to do a peek twice or what would be best to fix this. Overall it's not too important and such an edge case... going to sleep now.

kdy1: I think the best solution is just ignoring it.
I don't expect a user to write code like `import type from 'foo';`

kdy1 (approved):
Great! Thanks!

PR #829, opened 2020-06-09
Add a jsdoc parser

### LINKED ISSUE #827: JSDoc Parser
It would be useful if swc could be used to parse JSDocs similarly to how the TS compiler parses them.

Use cases:

1. Type checking
2. Automatically formatting JS docs.
3. Analyzing JS docs for information.

I think it should be lazily parsed as parsing and storing this information is not so useful when compiling.

kdy1:
It seems more complex than I initially thought.

Especially, I'm not sure if it's ok to change argument number based on the tag.
For example, `param` is used like `@param {type} name description`, so it can be parsed as an item with 3 parameters. But I don't know well about jsdoc spec.

cc @dsherret @mattd3v

Closes #827.


ecmascript/ast/src/jsdoc.rs
```diff
@@ -0,0 +1,322 @@
...
+pub struct JsDoc {
+    pub span: Span,
+
+    pub description: Str,
+    pub tags: Vec<JsDocTagItem>,
+}
+
+/// JsDoc tag with comments.
+#[ast_node]
+pub struct JsDocTagItem {
```
dsherret, line 14: What about just `JsDocTag` to copy what's done in the TS compiler (and it's shorter 🙂)?

image

dsherret: Oh, I should scroll down to see the enum. Sorry! That is awesome you've implemented all those!!

PR #844, opened 2020-06-14
store declare as start for functions declaration

### LINKED ISSUE #843: declare span is not stored for functions
**Describe the bug**
When using `declare` before a function declaration in a typescript file, the `span` field of the `Function` struct in `FunctionDecl` only starts at the start of the word `function`. Compared to the class declaration where the span field starts at the beginning of the `declare` keyword. It is problematic to find the start of the declaration (see [this issue](https://github.com/denoland/deno/issues/6117#issuecomment-642643810)).

**Input code**
```typescript
declare function example(): undefined;
declare class Example {};
```
**Config**
Using swc as a crate

**Expected behavior**
`FunctionDecl.Function.span.lo()` should return the start of the `declare` keyword as it is done with the class declaration.

**Version**
The version of @swc/core: `master`

Pangoraw:
Hope this fixes the problem as intended :+1: 

Fixes https://github.com/swc-project/swc/issues/843

### DIFF
--- a/ecmascript/parser/src/parser/class_and_fn.rs
+++ b/ecmascript/parser/src/parser/class_and_fn.rs
@@ -25,6 +25,14 @@ impl<'a, I: Tokens> Parser<'a, I> {
         self.parse_fn(Some(start), decorators)
     }
 
+    pub(super) fn parse_declare_fn_decl(
+        &mut self,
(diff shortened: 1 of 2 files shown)


kdy1:
Closing as inactive

PR #868, opened 2020-06-28
Fixed wasm compilation error. 

rickevry:
Cwd references default_cwd which is excluded in the wasm32 build.


rickevry:
Found more problems

PR #869, opened 2020-06-28
Fixed wasm compilation errors.

rickevry:
Cwd references default_cwd which is excluded in the wasm32 build.

Disable_hygiene must be defined for the code to compile.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -46,6 +46,10 @@ pub struct ParseOptions {
     pub target: JscTarget,
 }
 
+fn default_as_true() -> bool {
+    true
+}
+
 #[derive(Debug, Clone, Default, Deserialize)]
 #[serde(deny_unknown_fields, rename_all = "camelCase")]
 pub struct Options {
@@ -54,6 +58,10 @@ pub struct Options {
 
     #[cfg(not(target_arch = "wasm32"))]
     #[serde(skip_deserializing, default)]
+    pub disable_hygiene: bool,
+
+    #[cfg(target_arch = "wasm32")]
+    #[serde(default="default_as_true")]
     pub disable_hygiene: bool,
 
     #[serde(skip_deserializing, default)]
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -56,12 +60,17 @@ pub struct Options {
     #[serde(skip_deserializing, default)]
     pub disable_hygiene: bool,
 
+    #[cfg(target_arch = "wasm32")]
```
kdy1, line 63: Why did you do this?

src/config.rs
```diff
@@ -56,12 +60,17 @@ pub struct Options {
...
+    #[serde(default = "default_as_true")]
+    pub disable_hygiene: bool,
+
     #[serde(skip_deserializing, default)]
     pub disable_fixer: bool,
 
     #[serde(skip_deserializing, default)]
     pub global_mark: Option<Mark>,
 
+    #[cfg(not(target_arch = "wasm32"))]
```
kdy1, line 73: If I remember correctly, there's two version of `default_cwd `. One for wasm, one for non-wasm. Can you delete the wasm version?

kdy1 (changes requested):

src/config.rs
rickevry: The real problem is line 59 that removes the property:

    #[cfg(not(target_arch = "wasm32"))]
    #[serde(skip_deserializing, default)]
    pub disable_hygiene: bool

it breaks line 250:
         .hygiene(!self.disable_hygiene)

My fix only added the property back when compiling for wasm

src/config.rs
rickevry: I can not find default_cwd for wasm anywhere. The only way to get the code to compile was to remove it.

src/config.rs
kdy1: I checked it!

kdy1:
Thanks!

PR #870, opened 2020-06-28
Added build scripts without --debug for web and nodejs

rickevry:
I am not sure you need this, but these scripts show how to build wasm without debug information. Also how to use different targets.


rickevry:
New PR with more stuff coming soon

PR #871, opened 2020-06-29
Example that shows how to run swc in the browser. 

rickevry:
Instructions can be found in swc/wasm/example/readme.txt


kdy1:
Thanks!

PR #877, opened 2020-07-06
Fix span of MethodProp

dsherret:
Needs to include stuff like the `async` modifier.

There is actually a test for this ("method-async-generator"), but the output was already correct, which I don't understand. I'm probably missing something or perhaps the CI will shed some light.


kdy1:
Thanks a lot!

PR #881, opened 2020-07-17
spack: remove_item is deprecated

95th:
`Vec::remove_item` is deprecated in https://github.com/rust-lang/rust/pull/73539


kdy1:
Great! Thanks!

PR #887, opened 2020-07-22
refractor: `native` refractors

littledivy:
Refracts the native (neon) source code into submodules for every method


kdy1:
Great! Thanks!

PR #891, opened 2020-07-24
Add brief explanation of project to README

clbarnes:
I know it's in the github tagline and on the website, but they're not part of the repo. IMO the purpose of a README is saying "read me!": if someone reads one thing about the project, it'll be that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,8 @@
    </a>
 </p>
 
+swc is a super-fast typescript / javascript compiler written in rust.
+
 <h2 align="center">Supporting swc</h2>
 
 <p align="center">


clbarnes:
Pretty sure the failure is unrelated to the PR.

kdy1:
Thanks!

hi2u:
Additionally, I don't think it would hurt to be a little bit more specific about what it means by "compiler"...

> swc is a super-fast typescript / javascript compiler written in rust.

This could be taken as either:

* a) Something that compiles to binaries, i.e. an alternative to https://github.com/vercel/pkg
* b) Something that transpiles JS/TS to JS, i.e. an alternative to `tsc`

Maybe obvious to some, but it wasn't to me. I initially thought it was (a) - took a bit of research to figure out that it's actually (b).

clbarnes:
I suppose strictly it's a transpiler but I feel that term is falling out of favour a bit:

> swc is a super-fast compiler written in rust; producing widely-supported javascript from modern standards and typescript.

kdy1:
Hmm... I think 

> swc is a super-fast compiler written in rust; producing widely-supported javascript from modern standards and typescript.

this one is better. I'll change it.

kdy1:
Btw, swc will be a compiler in future.
I named it compiler because my blueprint of swc was not just transpiler. Transpiling is one of core feature of the compiler I was implementing.

> The design of swc was for an all-in-one package. Bundler / web assets processor (js, css, html, svg, etc..) / linter. That's why I created lots of macros and created the span hygiene.

This is what I said while talking

Issue #894, opened 2020-07-25
Support declare modifier on class properties


dsherret:
The declare modifier is valid on class properties as of TS 3.7: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html

Right now, the following parses as if it didn't have the `declare` or `public` modifiers.

```ts
class Test {
    declare public prop1: string;
    declare prop2: string;
}
```

https://www.typescriptlang.org/play?ts=4.0.0-beta#code/MYGwhgzhAEAqCmEAu0DeBYAUNH0Am8oYATvNAA4CuARiAJbAXED25AjAFzTLF0B2AcwDcWXPkLhSTVgCYuPfsKwBfIA

It's important to note that this should generate a diagnostic:

```ts
class Test {
    declare #other: string;
}
```

https://www.typescriptlang.org/play?ts=4.0.0-beta#code/MYGwhgzhAEAqCmEAu0DeBYAUNH0Am8oYATvNAMQD2SAFvMQFzTLECWAdgOYDcWAvkA

dsherret:
I'm working on the parsing for this right now.

PR #897, opened 2020-07-25
Support parsing declare properties

dsherret:
This is a parsing only implementation for #894.

These properties need to be ignored in the output when transforming.

### DIFF
--- a/ecmascript/ast/src/class.rs
+++ b/ecmascript/ast/src/class.rs
@@ -103,6 +103,9 @@ macro_rules! property {
             pub readonly: bool,
 
             #[serde(default)]
+            pub declare: bool,
+
+            #[serde(default)]
             pub definite: bool,
         }
     };
--- a/ecmascript/ast/src/class.rs
+++ b/ecmascript/ast/src/class.rs
@@ -63,56 +63,90 @@ pub enum ClassMember {
     Empty(EmptyStmt),
 }
 
-macro_rules! property {
-    ($name:ident, $ty:literal, $KEY:ty) => {
-        #[ast_node($ty)]
-        #[derive(Eq, Hash)]
-        pub struct $name {
-            #[serde(default)]
-            pub span: Span,
+#[ast_node("ClassProperty")]
+#[derive(Eq, Hash)]
+pub struct ClassProp {
+    #[serde(default)]
+    pub span: Span,
 
-            pub key: $KEY,
+    pub key: Box<Expr>,
 
-            #[serde(default)]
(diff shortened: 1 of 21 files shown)


ecmascript/ast/src/class.rs
```diff
@@ -102,6 +102,9 @@ macro_rules! property {
             #[serde(default)]
             pub readonly: bool,
 
+            #[serde(default)]
+            pub declare: bool,
```
dsherret, line 106: Should we get rid of this macro so this property doesn't appear on `PrivateProp`?

ecmascript/parser/src/parser/class_and_fn.rs
```diff
@@ -736,6 +734,9 @@ impl<'a, I: Tokens> Parser<I> {
         if is_constructor(&key) {
             syntax_error!(key.span(), SyntaxError::PropertyNamedConstructor);
         }
+        if declare && key.is_left() {
+            syntax_error!(key.span(), SyntaxError::DeclarePrivateIdentifier);
```
dsherret, line 738: What's the recommended way to add a test for this?

ecmascript/ast/src/class.rs
kdy1: I think it is right way to go

dsherret: Ok, sounds good. I'll do this now.

Also, how do I get the tests to automatically update the .json files? Previously it would just happen, but now I'm not sure how.

ecmascript/parser/src/parser/class_and_fn.rs
kdy1: I think creating a file at ecmascript/parser/tests/typescript/issue-894/input.ts would be the best way

ecmascript/ast/src/class.rs
kdy1: You can do `UPDATE=1 cargo test`

dsherret: Sweet. Thanks!

ecmascript/parser/src/parser/class_and_fn.rs
dsherret: I ended up putting it in the typescript-errors folder because I couldn't get it working in the typescript folder after trying a bit.

kdy1: I'm sorry. I thought you was talking about correct case.
(I'm sleepy as it's over 02 am.)

kdy1 (approved):

PR #901, opened 2020-07-27
Fix bugs

### LINKED ISSUE #879: Property decorators shouldn't duplicate class
**Describe the bug**

Using a property decorator causes the class definition to be duplicated in generated code. This makes it impossible to edit the prototype itself in the decorator, since the given target object refers to a duplicate of the class.

I didn't test with non-legacy decorators, since I need Typescript anyway.

**Input code**

```ts
export default class X {
    @networked
    prop: string = "";
}
```

**Config**

(I'm using Rust directly)
```rust
c.process_js_file(
            fm,
            &Options {
                is_module: true,
                config: Some(Config {
                    env: Some(EnvConfig {
                        ..Default::default()
                    }),
                    module: Some(ModuleConfig::CommonJs(
                        Default::default()
                    )),
                    jsc: JscConfig {
                        syntax: Some(Syntax::Typescript(TsConfig {
                            tsx: true,
                            decorators: true,
                            ..Default::default()
                        })),
                        transform: Some(TransformConfig {
                            legacy_decorator: true,
                            ..Default::default()
                        }),
                        ..Default::default()
                    },
                    ..Default::default()
                }),
                ..Default::default()
            },
        )
```

**SWC-generated code**
```js
// ..bunch of boilerplate
var X = (function() {
    var X = function X() {
        'use strict';
        _classCallCheck(this, X);
        _initializerDefineProperty(this, 'prop', _descriptor, this);
    };
    return X;
}(), _descriptor = _applyDecoratedDescriptor(function() {
    var X = function X() {
        'use strict';
        _classCallCheck(this, X);
        _initializerDefineProperty(this, 'prop', _descriptor, this);
    };
    return X;
}().prototype, 'prop', [
    networked
], {
    configurable: true,
    enumerable: true,
    writable: true,
    initializer: function initializer() {
        return '';
    }
}), function() {
    var X = function X() {
        'use strict';
        _classCallCheck(this, X);
        _initializerDefineProperty(this, 'prop', _descriptor, this);
    };
    return X;
}());

```

**Expected behavior**

`_applyDecoratedDescriptor` should receive the final prototype as the first argument. That's what TSC does:
```js
var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {
    var c = arguments.length, r = c < 3 ? target : desc === null ? desc = Object.getOwnPropertyDescriptor(target, key) : desc, d;
    if (typeof Reflect === "object" && typeof Reflect.decorate === "function") r = Reflect.decorate(decorators, target, key, desc);
    else for (var i = decorators.length - 1; i >= 0; i--) if (d = decorators[i]) r = (c < 3 ? d(r) : c > 3 ? d(target, key, r) : d(target, key)) || r;
    return c > 3 && r && Object.defineProperty(target, key, r), r;
};
let X = /** @class */ (() => {
    class X {
        constructor() {
            this.prop = "";
        }
    }
    __decorate([
        networked
    ], X.prototype, "prop", void 0);
    return X;
})();
export default X;
```

**Version**
Latest github master

### LINKED ISSUE #896: Does not support type only import/export syntax
**Describe the feature**

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export

**Behavior in current version**

```
error: Expected LBrace, got Some(BinOp(Mul))
  --> packages/types/src/index.ts:13:13
   |
13 | export type * from '../typings'

```

kdy1:
swc_ecma_parser:
 - Accept `export type * from '../typings'` (Closes #896)

swc_ecma_transforms:
 - Ensure that swc does not duplicate classes (Closes #879)

### DIFF
--- a/ecmascript/parser/tests/typescript/issue-896/input.ts.json
+++ b/ecmascript/parser/tests/typescript/issue-896/input.ts.json
@@ -1,0 +1,29 @@
+{
+  "type": "Module",
+  "span": {
+    "start": 14,
+    "end": 45,
+    "ctxt": 0
+  },
+  "body": [
+    {
+      "type": "ExportAllDeclaration",
+      "span": {
+        "start": 14,
+        "end": 45,
+        "ctxt": 0
+      },
+      "source": {
+        "type": "StringLiteral",
+        "span": {
+          "start": 33,
+          "end": 45,
+          "ctxt": 0
+        },
+        "value": "../typings",
+        "hasEscape": false
+      }
+    }
+  ],
+  "interpreter": null
+}
--- a/tests/projects.rs
+++ b/tests/projects.rs
@@ -540,7 +540,7 @@ fn issue_879() {
     .unwrap();
     println!("{}", f);
 
-    assert!(f.find("function X"), "swc should compile class");
+    assert!(f.find("function X").is_some(), "swc should compile class");
     assert_eq!(
         f.find("function X"),
         f.rfind("function X"),
--- a/tests/projects.rs
+++ b/tests/projects.rs
@@ -1,10 +1,13 @@
 use rayon::prelude::*;
 use std::{path::Path, sync::Arc};
 use swc::{
-    config::{Config, Options, SourceMapsConfig},
+    config::{Config, JscConfig, ModuleConfig, Options, SourceMapsConfig, TransformConfig},
     Compiler,
 };
-use swc_ecmascript::preset_env;
+use swc_ecmascript::{
+    parser::{Syntax, TsConfig},
+    preset_env,
+};
 use testing::{NormalizedOutput, StdErr, Tester};
 use walkdir::WalkDir;
 
@@ -39,7 +42,7 @@ fn file_with_opt(f: &str, options: Options) -> Result<NormalizedOutput, StdErr> 
                     Ok(v.code.into())
                 }
             }
-            Err(err) => panic!("Error: {}", err),
+            Err(err) => panic!("Error: {:?}", err),
         }
     })
 }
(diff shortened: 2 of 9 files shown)


ecmascript/parser/tests/typescript/issue-896/input.ts.json
```diff
@@ -0,0 +1,29 @@
...
+        "span": {
+          "start": 33,
+          "end": 45,
+          "ctxt": 0
+        },
+        "value": "../typings",
+        "hasEscape": false
+      }
+    }
+  ],
```
bartlomieju, line 27: Looks like it's the same AST node as for `export * from ".."` - so I guess there will be no way of telling if this is type only import/export. IMO it'd be very useful to have that information, in Deno we got `--no-check` option which emits type stripped code and elides type-only imports/exports. Currently Deno uses TSC to do that stripping but I intended to rewrite that functionality in SWC (https://github.com/denoland/deno/issues/6864), but I think it might not be possible without knowing that this statement can be safely elided.

bartlomieju: Unless I'm out of my way and `typescript::strip()` handles that automatically?

kdy1: I got it.

@bartlomieju 
How do you think about emitting error when user input is 
```ts
export type * from '../typings' 
```
It's invalid syntax, but error can be ignored using `@ts-ignore`

bartlomieju: To be honest I don't know, CCing @kitsonk for his expertise

bartlomieju: On the other hand, probably the least surprising behavior would be to duplicate what TSC does

kdy1: I'll create another pr for ast changes

kitsonk: Babel and typescript-eslint set `exportKind` to `"type"` instead of `"value"`: https://astexplorer.net/#/gist/408ad6e07e866fd7743725e9073bdf5b/74e0f25919bf3f997a3d83dee2a3635fc701156a.  TypeScript `transpileOnly()` will always strip `export type` and `import type` statements.  The main reason the syntax was introduced was to make type stripping possible (as `export ... from ...` cannot be statically determined to be a type or not).

Erroring I suspect would be a challenge, since it is valid TypeScript and we wouldn't want to `@ts-ignore` it.

kdy1: tsc emits `TS1383: Only named exports may use 'export type'` for

```ts
export type * from '../typings'
```

tests/projects.rs
```diff
@@ -507,3 +510,40 @@ fn issue_801() {
...
+                env: Some(Default::default()),
+                module: Some(ModuleConfig::CommonJs(Default::default())),
+                jsc: JscConfig {
+                    syntax: Some(Syntax::Typescript(TsConfig {
+                        tsx: true,
+                        decorators: true,
+                        ..Default::default()
+                    })),
+                    transform: Some(TransformConfig {
+                        legacy_decorator: true,
```
wyozi, line 530: @kdy1 I think I might have accidentally given you invalid config in the issue #879, since this one doesn't even seem to enable optimizations if I read the code correctly (`optimizer` would default to `None` as is defined at https://github.com/swc-project/swc/blob/master/src/config.rs#L572)

I tested HEAD transpiler locally, and this issue still seems to happen, although my transpiler config is not the same as it was in the issue anymore.

kdy1: I reopened it.

wyozi: Nice. Really appreciate all the work and active effort you put in

PR #916, opened 2020-07-29
fix: readonly is not stripped from private prop

### LINKED ISSUE #915: typescript::strip() leaves "readonly" modifier on private props
**Describe the bug**
Using `typescript::strip()` pass leaves "readonly" TS modifiers on private props (`#prop`).

**Input code**
https://github.com/denoland/deno/blob/cde4dbb3/std/log/logger.ts

Output:
```js
// Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
import { LogLevels, getLevelByName, getLevelName } from "./levels.ts";
export class LogRecord {
    #args;
    #datetime;
    constructor(options){
        this.msg = options.msg;
        this.#args = [
            ...options.args
        ];
        this.level = options.level;
        this.loggerName = options.loggerName;
        this.#datetime = new Date();
        this.levelName = getLevelName(options.level);
    }
    get args() {
        return [
            ...this.#args
        ];
    }
    get datetime() {
        return new Date(this.#datetime.getTime());
    }
}
export class Logger {
    #level;
    #handlers;
    readonly #loggerName;  // <---- offending line
    constructor(loggerName, levelName, options = {
    }){
        this.#loggerName = loggerName;
        this.#level = getLevelByName(levelName);
        this.#handlers = options.handlers || [];
    }
    ... snip ....
}
```

**Please notice that all but one `readonly` modifier is removed.**
Namely:
```ts
export class Logger {
  #level: LogLevels;
  #handlers: BaseHandler[];
  readonly #loggerName: string;
   ...snip...
}
```

Fails to remove `readonly`, but following class is properly stripped:

```
export class LogRecord {
  readonly msg: string;
  #args: unknown[];
  #datetime: Date;
  readonly level: number;
  readonly levelName: string;
  readonly loggerName: string;
  ... snip ...
}
```

**Expected behavior**
All `readonly` modifiers are removed.

bartlomieju:
Closes #915


ecmascript/codegen/src/lib.rs
```diff
@@ -901,10 +901,10 @@ impl<'a> Emitter<'a> {
...
         self.emit_accesibility(n.accessibility)?;
 
-        if n.readonly {
-            keyword!("readonly");
-            space!();
-        }
+        // if n.readonly {
+        //     keyword!("readonly");
+        //     space!();
+        // }
```
bartlomieju, line 907: This is probably not right, @kdy1 could you give me a short explanation on how emitter works? `readonly` should be emitted if we emit TypeScript, but should be stripped when emitting JavaScript. I can't find any place where that's decided.

kdy1: @bartlomieju It's determined by the data of node.

If you want special behavior only for typescript, you can change swc_emca_transforms::typescript::strip. In this case, you can modify `fold_private_prop` to remove readonly flag.

kdy1: @bartlomieju If you are ok with it, I'll add some commits to the pr to fix the bug.

bartlomieju: @kdy1 się, however it seems more bugs were found: 
- It's not just private props: target/debug/deno cache --unstable --no-check -r std/mime/multipart.ts. (The source line is even more haywire here, it's caused by readonly left on public MultipartReader fields.)

- I think there are way more bugs than with readonly.
- target/debug/deno cache --unstable --no-check -r std/path/win32.ts: generates with messed up backslash escape in template literal.
- target/debug/deno cache --unstable --no-check -r std/node/module.ts: generates with messed up hexadecimal escape sequence.

- - target/debug/deno cache --unstable --no-check -r std/path/_globrex.ts: backslash problem described earlier.
- target/debug/deno cache --unstable --no-check -r std/wasi/snapshot_preview1.ts: same.
- target/debug/deno cache --unstable --no-check -r std/ws/test.ts: doesn't strip enum.
That's all the errors I could yield in std, some might be blocked by others.

I'll add tests for all of them in a couple hours, and your commits are more than welcome

kdy1: I fixed private property issue and there are some known - although no issue exists on github - escape issues with codegen.
An enum not being stripped is bit surprising.

Test codes are in the deno repository, is it correct?

kdy1: @bartlomieju  There was a bug in the parser and codegen. I send a [pr](https://github.com/bartlomieju/swc/pull/1)

bartlomieju: Thanks for explanation @kdy1, I'll add more tests to codegen for the escape sequences.

ecmascript/codegen/src/lib.rs
```diff
@@ -2308,15 +2308,10 @@ fn unescape(s: &str) -> String {
         }
     }
 
-    // let s = s.replace("\\\\", "\\");
-
     let mut result = String::with_capacity(s.len() * 6 / 5);
     let mut chars = s.chars();
 
-    println!("S: {:?}", s);
```
kdy1: Oh... Nice catch.
I completely forgot about it as tests passed and nothing was shown.

bartlomieju:
@kdy1 I found a few more problems:
```ts
// Input:
export class MultipartReader {
  readonly newLine = encoder.encode("\r\n");
  ... snip ...
}

// Output:
export class MultipartReader {
  readonly newLine = encoder.encode("\r\n"); <--- doesn't remove readonly
  ... snip ...
}
```

**Already fixed the one above**

```ts
// Input:
relResolveCacheIdentifier = `${parent.path}\x00${request}`;

// Output:
relResolveCacheIdentifier = `${parent.path}\x0${request}`;  // <--- \x00 transformed into \x0
```

```ts
// Input:
const data = text.encode(`${arg}\0`);

// Output:
const data = text.encode(`${arg}\`); // <--- \0 transformed into \
```

```ts
// Input:
export class MultipartReader {
  readonly newLine = encoder.encode("\r\n");
  readonly newLineDashBoundary = encoder.encode(`\r\n--${this.boundary}`);
  readonly dashBoundaryDash = encoder.encode(`--${this.boundary}--`);
 ... snip ...
}

// Output:
export class MultipartReader {
    newLine = encoder.encode("\r\n");
    newLineDashBoundary = encoder.encode(`\
\
--${this.boundary}`);   // <--- completely messed up \r\n replaced with actual newline token
    dashBoundaryDash = encoder.encode(`--${this.boundary}--`);
  ... snip ...
}
```

kdy1:
I'll take a look

bartlomieju:
It's strange cause you added exact same tests for the escape problems 😕

kdy1:
@bartlomieju I think I fixed it. I'm not sure if it is correct fix, though.

bartlomieju:
@kdy1 I just pulled the code and tested again, those two things are still failing:
```ts
const data = text.encode(`${arg}\0`);
// strips to:
const data = text.encode(`${arg}\`);
```

```ts
export class MultipartReader {
  readonly newLine = encoder.encode("\r\n");
  readonly newLineDashBoundary = encoder.encode(`\r\n--${this.boundary}`);
  readonly dashBoundaryDash = encoder.encode(`--${this.boundary}--`);
 ... snip ...
}

// strips to
export class MultipartReader {
    newLine = encoder.encode("\r\n");
    newLineDashBoundary = encoder.encode(`\
\
--${this.boundary}`); 
```

kdy1:
I'll try using raw files instead of string literals in test. I think the issue is related to it.

bartlomieju:
@kdy1 this test fails:
```
#[test]
fn tpl_escape_5() {
    test_from_to(
        r#"const data = text.encode(`${arg}\0`);"#,
        r#"const data = text.encode(`${arg}\0`);"#,
    );
}
```

```
---- tests::tpl_escape_5 stdout ----
--------------------
Source:
const data = text.encode(`${arg}\0`);
Pos: BytePos(0) ~ BytePos(37)

thread 'tests::tpl_escape_5' panicked at 'assertion failed: `(left == right)`
  left: `const data = text.encode(`${arg}\`);`,
 right: `const data = text.encode(`${arg}\0`);`', ecmascript/codegen/src/tests.rs:94:5


failures:
    tests::tpl_escape_5
```

kdy1:
@bartlomieju I fixed \0 problem

bartlomieju:
> @bartlomieju I fixed \0 problem

Thank you, I confirmed that problematic files are now properly stripped. How you seen my change to new line handling? https://github.com/swc-project/swc/pull/916/commits/a9bd9446f273e12f99dd8533e2606ec9248e249d

kdy1:
I think it is great. That's what I should do after removing `s = s.replace("\\\\", "\\")`

bartlomieju:
Awesome 👍  we'll be releasing new version of Deno tomorrow so I expect more bugs will be uncovered by users 😅  hopefully we'll be able to fix them as we go

PR #943, opened 2020-08-07
swc_bundler

kdy1:
# TODOs
 - [x] Non-path modules

Currently it works only if there's no common module nor dynamic imports.

 - [x] Circular imports
   - [x] Merge
   - [x] Hygiene
     - [x] Hygiene in cyclic merging
     - [x] Hygiene while importing merged module
   - [x] Order of execution
     - [x] Class inheritance
     - [x] console.log
     - [x] Hygiene

Basically it works, but I think it should handle much more cases.

 - [x] Option to skip require analysis

 - [x] bug of load()

Currently, if a depends on b and b depends on a, 

```
a -> b -> a
       ^^ This part is broken. Due to circular dependency detection, swc_bundler skips analyzing imports from a to b. 
``` 

 - [x] Remove logging (hygiene info is printed to stdout)

 - [x] Concurrency bug

 - [x] DCE bugs

 - [x] Parser bugs


kdy1:
cc @ry @bartlomieju

bartlomieju:
@kdy1 could you point me where should I start going through the code to get some kind of end-to-end bundling functionality?

kdy1:
I'll create an example for it soon.

kdy1:
I added an example, but currently non-file module does not work because last pass uses path information to update imports.


Also I expect some bugs from tree shaker, especially when it's combined with typescript.

kdy1:
It's almost done, but I need a way to distinguish global and top level binding.

Maybe making resolver accept two marker - one for top level items, one for references to global - can be a solution

kdy1:
cc @ry @bartlomieju It's merged!

PR #948, opened 2020-08-09
Fix metadata emitted in parameters decorator

Brooooooklyn:
`@Session() session: express.Express.Session` will emitted as `typeof express.Express.Session === 'undefined' ? Object : express.Express.Session` which caused runtime error.


kdy1:
Thanks!

PR #954, opened 2020-08-10
Fix typescript strip in class properties pass

Brooooooklyn:


ecmascript/transforms/src/compat/es2020/class_properties.rs
```diff
@@ -586,7 +586,10 @@ impl ClassProperties {
                         c.params = c.params.move_map(|mut param| match &mut param {
                             ParamOrTsParamProp::TsParamProp(p) => match &p.param {
                                 TsParamPropParam::Ident(i) => {
-                                    typescript_constructor_properties.push(store(i));
+                                    typescript_constructor_properties.push(store(&Ident {
+                                        type_ann: None,
```
kdy1, line 590: I think `type_ann` will be dropped by typescript_strip pass, isn't it?

ecmascript/transforms/src/compat/es2020/class_properties.rs
```diff
@@ -595,11 +598,19 @@ impl ClassProperties {
                                 }
                                 TsParamPropParam::Assign(pat) => match &*pat.left {
                                     Pat::Ident(i) => {
-                                        typescript_constructor_properties.push(store(i));
+                                        typescript_constructor_properties.push(store(&Ident {
+                                            type_ann: None,
```
kdy1, line 602: Same as above

kdy1:
Anyway, the change looks harmless to me. Thanks

ecmascript/transforms/src/compat/es2020/class_properties.rs
Brooooooklyn: No, seems like `type_ann` has not been dropped

kdy1: Ok, I got it.

kdy1 (approved):

ecmascript/transforms/src/compat/es2020/class_properties.rs
Brooooooklyn: I got this output:

```js
class BadRequestError extends Error {
    constructor(message){
        super(message);
        this.message: string = message;
    }
}
```

from

```ts
class BadRequestError extends Error {
  constructor(public readonly message: string) {
    super(message)
  }
}
```

PR #957, opened 2020-08-11
Never rename exports ident

Brooooooklyn:
Now `export default isBuffer; ` is transformed into `exports1.default = _default;`

### DIFF
--- a/ecmascript/transforms/src/hygiene.rs
+++ b/ecmascript/transforms/src/hygiene.rs
@@ -583,7 +583,10 @@ impl<'a> Fold for Hygiene<'a> {
 
     /// Invoked for `IdetifierRefrence` / `BindingIdentifier`
     fn fold_ident(&mut self, i: Ident) -> Ident {
-        if i.sym == js_word!("arguments") || i.sym == js_word!("undefined") {
+        if i.sym == js_word!("arguments")
+            || i.sym == js_word!("undefined")
+            || i.sym == js_word!("exports")
+        {
             return i;
         }
 
--- a/ecmascript/transforms/tests/modules_common_js.rs
+++ b/ecmascript/transforms/tests/modules_common_js.rs
@@ -4185,3 +4185,45 @@ function setup(url: string, obj: any) {
     return _url1;
 }"
 );
+
+test!(
+    ts_syntax(),
+    |_| tr(Config {
+        ..Default::default()
+    }),
+    lodash_es,
+    "import root from './_root.js';
+  import stubFalse from './stubFalse.js';
+  
+  var freeExports = typeof exports == 'object' && exports && !exports.nodeType && exports;
+
+  var freeModule = freeExports && typeof module == 'object' && module && !module.nodeType && \
+     module;
+  
+  var moduleExports = freeModule && freeModule.exports === freeExports;
+  
+  var Buffer = moduleExports ? root.Buffer : undefined;
+  
+  var nativeIsBuffer = Buffer ? Buffer.isBuffer : undefined;
+  
+  var isBuffer = nativeIsBuffer || stubFalse;
+  
+  export default isBuffer;
+",
+    "'use strict';
+Object.defineProperty(exports, '__esModule', {
+    value: true
+});
+exports.default = void 0;
+var _rootJs = _interopRequireDefault(require('./_root.js'));
+var _stubFalseJs = _interopRequireDefault(require('./stubFalse.js'));
+var freeExports = typeof exports == 'object' && exports && !exports.nodeType && exports;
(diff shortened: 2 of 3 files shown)


ecmascript/transforms/tests/modules_common_js.rs
```diff
@@ -4185,3 +4185,45 @@ function setup(url: string, obj: any) {
...
+  
+  var moduleExports = freeModule && freeModule.exports === freeExports;
+  
+  var Buffer = moduleExports ? root.Buffer : undefined;
+  
+  var nativeIsBuffer = Buffer ? Buffer.isBuffer : undefined;
+  
+  var isBuffer = nativeIsBuffer || stubFalse;
+  
+  export default isBuffer;
```
kdy1, line 4211: Given this line, the file is treated as an es module so `exports` is just an identifier.

ecmascript/transforms/src/hygiene.rs
```diff
@@ -583,7 +583,10 @@ impl<'a> Fold for Hygiene<'a> {
 
     /// Invoked for `IdetifierRefrence` / `BindingIdentifier`
     fn fold_ident(&mut self, i: Ident) -> Ident {
-        if i.sym == js_word!("arguments") || i.sym == js_word!("undefined") {
+        if i.sym == js_word!("arguments")
+            || i.sym == js_word!("undefined")
+            || i.sym == js_word!("exports")
```
kdy1, line 588: `exports` is not a special keyword in javascript, so I'm not sure if preserving it from hygiene pass is correct.

ecmascript/transforms/tests/modules_common_js.rs
kdy1: As a side note: Renaming was intentional. See: https://github.com/babel/babel/blob/e498bee10f0123bb208baa228ce6417542a2c3c4/packages/babel-plugin-transform-modules-commonjs/src/index.js#L149-L153

ecmascript/transforms/src/hygiene.rs
Brooooooklyn: > exports is not a special keyword in javascript

hmmm, yes I agree, seems like I need to find another way to solve my problem.

Currently I got `exports1 is not defined` when I use `swc` with `lodash-es`

kdy1: If it's correct thing to do regarding the comment above, you can pass mark passed to resolver_with_mark to modules pass and apply those marks to exports and others.
However I'm not sure if it's correct way to go. I think lodash-es will break babel too if you apply common js module plugin to it.

Brooooooklyn: >  think lodash-es will break babel too if you apply common js module plugin to it

I will try later.

And I think maybe follow the `TypeScript` behavior would take less risk. And I will figure out how `TypeScript` handle this situation.

ecmascript/transforms/tests/modules_common_js.rs
Brooooooklyn: works fine in `babel + babel-plugin-transform-modules-commonjs`, [Babel repl](https://babeljs.io/en/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&spec=false&loose=false&code_lz=JYWwDg9gTgLgBFCF4DNEjgcgHQHoD6iy2AVgM6YDcAUKJLHGTAK4BGAYgIYA2ZApnDQQMOXEzZdefUhRrUAbpyiCofPgFEAHvRhk4AXjgwAnmD4QUcPtui6DhzBFYk-AYxiY4AMi9WbsPR84AEJrHTJsADsIABM-ABVTASCw2zIaRWU0NQBZWOZuAUNsjX87IJMzCzgQfML7LCcXd08g2piC5N9g9s6o2ISk719qODHewrlMmrrS8IMVXNnhxb48jsLsVID7YtU5tKmlOAAhZhQUPmVDCYOdgH4EJBhsM4uruAAuOGZIuJRgJE-DEjspIpwYMB5HwAJJkN6Xa6nc6IuCPBFXbDAeEoj7fX7_QHA0FwbEYpHgyHQuHkuAAHzpjBYHB4_Dk23g_04BXgZNxUGoQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-3%2Cenv&prettier=false&targets=&version=7.11.1&externalPlugins=%40babel%2Fplugin-transform-modules-commonjs%407.10.4)

Brooooooklyn:
Close because of incorrect approach. I will open an issue to track this problem.

PR #972, opened 2020-08-15
Improve spack

### LINKED ISSUE #965: Spack should use cache properly
Currently, a file is loaded multiple times when the bundler is in concurrent mode and multiple file depends on it.
I think the best solution is to prefetch all dependencies.

I will try to fix this at next version

kdy1:
swd_bundler:
 - Handle reexports

spack:
 - Do not transform files from node_modules

--- 

Closes #965


lastmjs:
> Do not transform files from node_modules

Why are files from node_modules not transformed? I'm running into an issue where spack is not parsing an import as TypeScript because it is in node_modules. I would really like it to just parse it as it is, this works fine in development.

Issue #583, opened 2020-01-14
Add test for node text?


dsherret:
There have been a decent amount of bugs related to the AST node start and end being incorrect. I think we could help prevent regressions and make future node positions a little more testable if we develop a new test for this.

Perhaps beyond the current test that uses numbers it would be useful to create two typescript files (one jsx, one non-jsx) that hold all the different possible AST node combinations. Then a json test file that has all the AST nodes, but only contains the type, text, and any children. Maybe something like the following:

```json
// perhaps we'd skip the root module text because that would just be long and not so useful
[{
    "type": "FunctionDeclaration",
    "text": "function test {}",
    "name": {
        "type": "Identifier",
        "text": "test"
    },
    "params": [],
    // etc...
}, {
    // ... other statements go here ....
}]
```

The file would be auto-generated similar to the `input.ts.json` files, but this would make it extremely easy to visually identify problems with node span lengths.

Thoughts?

kdy1:
It seems like a good idea. I'll implement it after doing another project which will be done I guess before the weekend.

kdy1:
Postponing as it requires lots of work. I'll do this when I have an idea to reduce work. 
(Adding a field to nested json is really hard in rust)

PR #974, opened 2020-08-17
remove codegen::Handlers

bartlomieju:
`codegen::Handler` is a noop trait. It's used everytime `codegen::Emitter` must be created and there were comments that it should be removed. So I went ahead and did that. @kdy1 if you want to use that trait in the future feel free to close this PR.


kdy1:
Thanks for the pr!

I previously closed a similar pr because I thought I will use it while testing, but I found better solution for #583, and I think the solution for #583 also can be used to test type engine

PR #1005, opened 2020-08-26
Improve error messages for AST node proc macros.

teymour-aldridge:
Not sure if this is actually a useful change, but I thought it might help?


kdy1:
Sorry but I'm not sure if it is required.

PR #1010, opened 2020-08-29
Add dependency analyzer

bartlomieju:


bartlomieju:
@kdy1 PTAL

ecmascript/dep-graph/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+    ImportType,
+    Export,
+    ExportType,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct DependencyDescriptor {
+    pub kind: DependencyKind,
+    /// A flag indicating if the import is dynamic or not.
+    pub is_dynamic: bool,
```
bartlomieju, line 35: Debatable - maybe better to use separate kind for dynamic imports as well as `require()`?

ecmascript/dep-graph/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+    /// Any leading comments associated with the dependency.  This is used for
+    /// further processing of supported pragma that impact the dependency.
+    pub leading_comments: Vec<Comment>,
```
bartlomieju, line 38: Not very handy in tests to check spans, but oh well

ecmascript/dep-graph/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+    /// The location of the import/export statement.
+    pub col: usize,
+    pub line: usize,
+    /// The text specifier associated with the import/export statement.
+    pub specifier: String,
```
kdy1, line 43: Is this type intentional?

I think it will not impact performance anyway, but just to be sure.

ecmascript/dep-graph/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+}
+
+struct DependencyCollector<'a> {
+    comments: &'a SingleThreadedComments,
```
kdy1, line 47: SingleThreadedComments seems bit opinionated. How about using `&'a dyn Comments?`

ecmascript/dep-graph/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+                        specifier,
+                    });
+                }
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
kdy1, line 170: Thanks for the test!

ecmascript/dep-graph/src/lib.rs
bartlomieju: Yeah, but I'm open to changing it, what's your suggestion?

ecmascript/dep-graph/src/lib.rs
bartlomieju: Sure!

bartlomieju: Minor inconvenience is that `Comments` has no `with_leading` method - I can work around it by taking a comment, cloning it, and putting it back into the map. Does that work for you?

ecmascript/dep-graph/src/lib.rs
kdy1: I think flag is right way to go as `require`  can be both dynamic or static.
swc_bundler actually handle requires as a static library if they are called at top level.
e.g.
```js
const { join } = require('path');
```
This is an unconditional import with specifier 'join'.

bartlomieju: Right, so I guess we can just add another `DependencyKind::Require` to handle them?

ecmascript/dep-graph/src/lib.rs
kdy1: I think `JsWord`, which is interned version of `Box<str>`, would be better.
Same identifier is typically used widely across modules.

ecmascript/dep-graph/src/lib.rs
kdy1: It works for me but I think using `SingleThreadedComments` is better than it.

ecmascript/dep-graph/src/lib.rs
kdy1: Yes, it will work.

bartlomieju: Done

kdy1:
Thanks a lot!

PR #1025, opened 2020-09-03
Upgrade to napi@0.5

Brooooooklyn:
Some apis changed.

I have tried migrate `get_deserialized` to `from_js_value` (which implemented by `serde`), but performance got a little decrease. I think it's because call `nodejs napi`  (create_string/create_boolean etc....) has heavy overhead, `serde_json.from_slice` could avoid them.


kdy1 (approved):

kdy1:
Thanks!

PR #1028, opened 2020-09-04
Handle dynamic imports in non-es6 modules

### LINKED ISSUE #1018: dynamic import should be transformed if module is not es6
config:

```js
{
  target: 'es2018',
  module: 'commonjs',
  dynamicImport: true
}
```

Source code:

```ts
describe('foo', () => {
  it('should import', async () => {
    expect.assertions(1);

    const { FOO } = await import('./foo')

    expect(FOO).toBe(true)
  })
})
```

swc output:

```js
var regeneratorRuntime = require("regenerator-runtime");
describe('foo', function() {
    it('should import', _asyncToGenerator(regeneratorRuntime.mark(function _callee() {
        var ref, FOO;
                case 0:
                    expect.assertions(1);
                    _ctx.next = 3;
                    return import('./foo');
                case 3:
                    ref = _ctx.sent;
                    FOO = ref.FOO;
                    void 0;
                    expect(FOO).toBe(true);
                case 7:
                case "end":
                    return _ctx.stop();
            }
        }, _callee);
    })));
});
```

TypeScript output:

```js
describe('foo', function () {
    it('should import', function () { return __awaiter(_this, void 0, void 0, function () {
        var FOO;
        return __generator(this, function (_a) {
            switch (_a.label) {
                case 0:
                    expect.assertions(1);
                    return [4 /*yield*/, Promise.resolve().then(function () { return require('./foo'); })];
                case 1:
                    FOO = (_a.sent()).FOO;
                    expect(FOO).toBe(true);
                    return [2 /*return*/];
            }
        });
    }); });
});
```

BTW, I don't think using `regenerator-runtime` is a good idea, the `TypeScript` async/await helper is smaller and faster. Maybe use `tslib` is better?

kdy1:
Closes #1018.


ecmascript/transforms/tests/modules_common_js.rs
```diff
@@ -4226,3 +4229,22 @@ test!(
...
+        ..Default::default()
+    }),
+    issue_1018_1,
+    "async function foo() {
+      await import('foo');
+    }",
+    "
+    \"use strict\";
+    async function foo() {
+        await Promise.resolve().then(function() {
```
Brooooooklyn, line 4245: I think just `await require("foo")` is ok

PR #1060, opened 2020-09-11
typescript_strip: option to remove completely unused imports

kdy1:
Current behavior is intended for multiple reason.

 1. At the time of implementing it, type-only imports without `import type` are only stripped out if it's used as type.

[ts 3.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html)  introduced type-only imports, because the problem I wrote above.
(Note that typescript had a breaking change after it)

 2. `swc` works on file-by-file basis, so it cannot know if it's a type-only.

---

I'm considering two options.

 1. Implement emit-only tsc which also handles some special stuffs like const enum within **swc_bundler**, as it already works by loading multiple file.

 2. Add config to allow user select which behavior should be used.

---

See: https://github.com/denoland/deno/pull/7413#issuecomment-690801361


Brooooooklyn:
I feel this behavior is  `unsafe`.

The import statement `import { Type } from './types';` , we are not sure if `./types.ts` file is side effect free.

For example:

```ts
// types.ts

windows.a = 1
```

And in other file:

```
const b = window.a + 1;
```

kdy1:
I'm not sure about this. 

> I feel this behavior is unsafe.
>
> The import statement import { Type } from './types'; , we are not sure if ./types.ts file is side effect free.

I agree with you. That's why I implemented it to preserve side effect. But I'm now confused because typescript compiler removes it.

Brooooooklyn:
> But I'm now confused because typescript compiler removes it

I think because `TypeScript` know this is type only import.

lucacasonato:
The typescript default is `--importsNotUsedAsValue=remove` though.

nayeemrmn:
Yeah I guess it makes less sense for swc to be opinionated about that. Can it be a configuration option?

nayeemrmn:
@kdy1 As for this PR, it would be ideal to keep the behaviour of `--importsNotUsedAsValues=remove` for now and exclusively address #1065. The way forward should be more clear.

kdy1:
@nayeemrmn I'll add a option to switch behavior, while preserving current semantic as default.

ecmascript/transforms/src/typescript/strip.rs
```diff
@@ -13,13 +14,45 @@ use swc_ecma_visit::{as_folder, Fold, Node, Visit, VisitMut, VisitMutWith, Visit
 /// Value does not contain TsLit::Bool
 type EnumValues = FxHashMap<Id, TsLit>;
 
+#[derive(Debug, Serialize, Deserialize)]
+#[non_exhaustive]
+pub enum ImportNotUsedAsValues {
+    #[serde(rename = "rename")]
```
Brooooooklyn, line 20: `rename = "remove"` ?

kdy1: Oh, thanks!

ecmascript/transforms/src/typescript/strip.rs
```diff
@@ -680,12 +713,22 @@ impl VisitMut for Strip {
...
+                        _ => true,
+                    },
+                    ImportNotUsedAsValues::Preserve => match entry {
+                        Some(&DeclInfo {
+                            has_type: true,
+                            has_concrete: false,
+                            ..
+                        }) => false,
+                        _ => true,
+                    },
```
nayeemrmn, line 731: @kdy1 This is not the behaviour of `tsc` under `"importsNotUsedAsValues": "preserve"`. The correct behaviour is to _never_ preserve completely unused import identifiers. The above branch should always be used.

kdy1: I'll make a breaking change to fix it.

PR #1074, opened 2020-09-13
Fix musl config

Brooooooklyn:


kdy1 (approved):
Thanks!

kdy1:
Seems like it fixes the problem.

Issue #906, opened 2020-07-27
Visitor plugin - visitObjectPatternProperties throws an error when walking through AST


adamuso:
**Summary:**
Object destructuring throws an error when plugin is provided in transform options.

**I tried this:**
```js
const { transformSync } = require("@swc/core");
const Visitor = require("@swc/core/Visitor").default;

class TestPlugin extends Visitor {

}

const out = transformSync(
  `const obj = { a: 1, b: 2 };
   const { a, b } = obj;`, {
  plugin: m => new TestPlugin().visitProgram(m)
});

console.log(out.code);
```

**I expected to see this happen:** successful, transformed output from swc

**Instead, this happened:** 
```
$ node test.js

C:\swc-bug\node_modules\@swc\core\Visitor.js:1248
        return nodes.map(this.visitObjectPatternProperty.bind(this));
                     ^

TypeError: Cannot read property 'map' of undefined
    at TestPlugin.visitObjectPatternProperties (C:\swc-bug\node_modules\@swc\core\Visitor.js:1248:22)
    at TestPlugin.visitObjectPattern (C:\swc-bug\node_modules\@swc\core\Visitor.js:1243:24)
    at TestPlugin.visitPattern (C:\swc-bug\node_modules\@swc\core\Visitor.js:1222:29)
    at TestPlugin.visitVariableDeclarator (C:\swc-bug\node_modules\@swc\core\Visitor.js:404:21)
    at Array.map (<anonymous>)
    at TestPlugin.visitVariableDeclarators (C:\swc-bug\node_modules\@swc\core\Visitor.js:401:22)
    at TestPlugin.visitVariableDeclaration (C:\swc-bug\node_modules\@swc\core\Visitor.js:397:31)
    at TestPlugin.visitDeclaration (C:\swc-bug\node_modules\@swc\core\Visitor.js:393:29)
    at TestPlugin.visitStatement (C:\swc-bug\node_modules\@swc\core\Visitor.js:218:29)
    at TestPlugin.visitModuleItem (C:\swc-bug\node_modules\@swc\core\Visitor.js:36:29)
```

**I'm using:** 
```
$ npx swc --version
@swc/cli: 0.1.26
@swc/core: 1.2.9
```

**Tip**:
[...]

PR #1079, opened 2020-09-17
fix: 🐛 fix https://github.com/swc-project/swc/issues/906

IWANABETHATGUY:


kdy1:
Thanks for the pr!
But ast nodes are changed and it isn't reflected to typescript types yet.

IWANABETHATGUY:
i change the type defination because the defination is not match with the rust implementation, 
here is my debugging shortcut , you can see runtime ObjectPattern has the **prop** `properties` instead of `props`, 
so i change the `type defination`

IWANABETHATGUY:
image

kdy1:
I meant there are many changes to rust ast, which is not reflected in typescript ast.

PR #1081, opened 2020-09-18
Extension for jest

### LINKED ISSUE #1048: Extension packages with hidden option.
`@swc-node/jest` seems like a good idea and I think such support for some libraries worth it.

But supporting library means a maintenance hazard. What should happen if version of a library is bumped, or abandoned?
It will require bumping major version of swc, which is not a good idea.

So I think the hidden option, intended to be used only by packages in the `@swc` namespace, can be a solution.

kdy1:
cc @Brooooooklyn 
Because definitions of ast nodes change over time, I think this is the easiest way for end-users.
(TypeScript introduces new feature, ecmascript spec evolves)

---

Closes #1048.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,43 +2,44 @@
 members = ["ecmascript", "ecmascript/jsdoc", "native", "spack", "wasm"]
 
 [package]
-name = "swc"
-version = "0.1.0"
 authors = ["강동윤 <kdy1997.dev@gmail.com>"]
-license = "Apache-2.0/MIT"
-repository = "https://github.com/swc-project/swc.git"
-documentation = "https://swc-project.github.io/rustdoc/swc/"
 description = "Speedy web compiler"
+documentation = "https://swc-project.github.io/rustdoc/swc/"
 edition = "2018"
+license = "Apache-2.0/MIT"
+name = "swc"
+repository = "https://github.com/swc-project/swc.git"
+version = "0.0.0"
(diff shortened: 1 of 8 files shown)


Brooooooklyn:
Just a note, `ts-jest` has added `path-mapping` transformer plugin for `jest` users: https://github.com/kulshekhar/ts-jest/blob/master/src/transformers/path-mapping.ts. 
We should also implement it if `{ hidden: { jest: true } }`.

kdy1:
Hmm... I looked at it, but it seems like cjson parser is required.
I'll make `@swc/jest` faster alternative for babel-jest and improve it over time.

bartlomieju:
FYI @dsherret wrote [`jsonc-parser`](https://crates.io/crates/jsonc-parser) which we use successfully in Deno

kdy1:
@bartlomieju Thanks! I think I can add support for path soon.

Brooooooklyn:
So is there any opportunity to have an independent `tsconfig.json` parse which be shared for `deno` and `swc`?

bartlomieju:
> So is there any opportunity to have an independent `tsconfig.json` parse which be shared for `deno` and `swc`?
> 
> Seems like deno already have one? @bartlomieju

@Brooooooklyn the actual parsing of `tsconfig.json` is just a few lines - in Deno in process that config to exclude a lot of settings that we ignore.

PR #1087, opened 2020-09-20
Reduce npm package size

Brooooooklyn:


kdy1:
Thanks! I was putting off thinking I should do it.

PR #1096, opened 2020-09-21
BREAKING(strip): Remove incorrect config

### LINKED ISSUE #1065: TypeScript type imports are only stripped when they are used as a type at least once
**Describe the bug**

TypeScript type imports are only stripped when they are used as a type at least once, even if they are not used as a value.

**Input code**

Working correctly:

```ts
// test2.ts
export interface MyType {
  my_property: string;
}
```
```ts
// test1.ts
import { MyType } from "./test2.ts";

interface MySecondaryType extends MyType {
  
}
```

Not working correctly:

```ts
// test2.ts
export interface MyType {
  my_property: string;
}
```
```ts
// test1.ts
import { MyType } from "./test2.ts";
```

Also not working correctly:

```ts
// test2.ts
export interface MyType {
  my_property: string;
}
```
```ts
// test1.ts
import { MyType } from "./test2.ts";
export type { MyType };
```

**Expected behavior**
I expect that `import { MyType } from "./test2.ts";` is stripped away in both examples.

### LINKED ISSUE #1094: importsNotUsedAsValues is wrong
@kdy1 This is not the behaviour of `tsc` under `"importsNotUsedAsValues": "preserve"`. The correct behaviour is to _never_ preserve completely unused import identifiers. The above branch should always be used.

_Originally posted by @nayeemrmn in https://github.com/swc-project/swc/pull/1060#discussion_r491990064_

It's a breaking change.

nayeemrmn:
Fixes #1065.
Fixes #1094.

### DIFF
--- a/ecmascript/transforms/src/typescript/strip.rs
+++ b/ecmascript/transforms/src/typescript/strip.rs
@@ -3,7 +3,6 @@ use crate::{
     util::{prepend_stmts, var::VarCollector, ExprFactory},
 };
 use fxhash::FxHashMap;
-use serde::{Deserialize, Serialize};
 use std::mem::take;
 use swc_atoms::{js_word, JsWord};
 use swc_common::{util::move_map::MoveMap, Span, Spanned, SyntaxContext, DUMMY_SP};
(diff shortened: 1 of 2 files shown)


kdy1:
I don't like removing it. I want to fix it.

PR #1101, opened 2020-09-23
fix(strip): Fix Config::import_not_used_as_values

### LINKED ISSUE #1065: TypeScript type imports are only stripped when they are used as a type at least once

### LINKED ISSUE #1094: importsNotUsedAsValues is wrong

nayeemrmn:
Closes #1065.
Closes #1094.

@kdy1 Could you cut a release after this?

### DIFF
--- a/ecmascript/transforms/src/typescript/strip.rs
+++ b/ecmascript/transforms/src/typescript/strip.rs
@@ -25,10 +25,10 @@ pub enum ImportNotUsedAsValues {
     Preserve,
 }
 
-/// This value defaults to `Preserve`
+/// This value defaults to `Remove`
 impl Default for ImportNotUsedAsValues {
     fn default() -> Self {
-        Self::Preserve
+        Self::Remove
     }
 }
 
--- a/ecmascript/transforms/src/typescript/strip.rs
+++ b/ecmascript/transforms/src/typescript/strip.rs
@@ -16,6 +16,8 @@ type EnumValues = FxHashMap<Id, TsLit>;
 
 #[derive(Debug, Serialize, Deserialize)]
 #[non_exhaustive]
+// TODO(nayeemrmn): The name should be `ImportsNotUsedAsValues`. Rename as a
+// breaking change.
 pub enum ImportNotUsedAsValues {
     #[serde(rename = "remove")]
     Remove,
@@ -23,10 +25,10 @@ pub enum ImportNotUsedAsValues {
     Preserve,
 }
 
-/// This value defaults to `Preserve`
+/// This value defaults to `Remove`
 impl Default for ImportNotUsedAsValues {
     fn default() -> Self {
-        Self::Preserve
+        Self::Remove
     }
 }
 
@@ -41,7 +43,7 @@ pub fn strip_with_config(config: Config) -> impl Fold {
         config,
         non_top_level: Default::default(),
         scope: Default::default(),
-        was_side_effect_import: Default::default(),
+        is_side_effect_import: Default::default(),
     })
 }
 
@@ -55,8 +57,7 @@ struct Strip {
     config: Config,
     non_top_level: bool,
     scope: Scope,
-
-    was_side_effect_import: bool,
+    is_side_effect_import: bool,
 }
 
 #[derive(Default)]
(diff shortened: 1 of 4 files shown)


ecmascript/transforms/src/typescript/strip.rs
```diff
@@ -16,17 +16,19 @@ type EnumValues = FxHashMap<Id, TsLit>;
...
     #[serde(rename = "preserve")]
     Preserve,
 }
 
-/// This value defaults to `Preserve`
+/// This value defaults to `Remove`
 impl Default for ImportNotUsedAsValues {
     fn default() -> Self {
-        Self::Preserve
+        Self::Remove
```
kdy1, line 31: Is this safe?
Even if the default of tsc is remove, `swc` preserved those.

kdy1: I meant `breaking change` for rust side.
I'll not make a breaking change for js apis. Sorry if you are confused.

ecmascript/transforms/src/typescript/strip.rs
```diff
@@ -713,26 +714,23 @@ impl VisitMut for Strip {
...
-                    },
-                    ImportNotUsedAsValues::Preserve => match entry {
-                        Some(&DeclInfo {
-                            has_type: true,
-                            has_concrete: false,
-                            ..
-                        }) => false,
-                        _ => true,
-                    },
+                match entry {
```
kdy1, line 717: Is this safe? 
It seems like semantic is changed from current behavior to `remove`

kdy1 (changes requested):
Of course, I will publish a new version after this.

ecmascript/transforms/src/typescript/strip.rs
nayeemrmn: This is the actual fix. The semantics for this bit should always be "remove". The meaning of `importsNotUsedAsValues` in `tsc` is to do with whether to keep an only-type import as a side-effect import, or remove it.

nayeemrmn: > Is this safe?

Always removing completely unused import identifiers won't cause any breakages, it's safe.

ecmascript/transforms/src/typescript/strip.rs
nayeemrmn: Reverted 👍

ecmascript/transforms/src/typescript/strip.rs
kdy1: You are right. Thanks!

kdy1:
@nayeemrmn Test fails

nayeemrmn:
Yeah I forgot to adapt the tests for the reverted default. Also I imagine a lot of other tests are dependent on the previous incorrect behaviour.

ecmascript/transforms/src/typescript/strip.rs
nayeemrmn: Wait @kdy1 the effect of `ImportNotUsedAsValues` is completely different from before. Defaulting it to remove would actually align with the current behaviour. Do you mind if I change the default to "remove"? It would be safe after all.

kdy1: No, I don't mind it.
I think it's safe anyway as they are not used.

kdy1: Sorry, I was confused about it.

kdy1 (approved):
Thanks! I'll merge it and publish a new version if the tests pass.

nayeemrmn:
@kdy1 Do you have an idea why this test is failing? It doesn't appear related to this change.
image
It happened in the other PR, too.

kdy1:
Currently output of spack is not deterministic if it's invoked in concurrent mode.
(Semantics are preserved, though)

kdy1:
I'll test locally

PR #1106, opened 2020-09-24
Fix jest extension

kdy1:


ecmascript/ext-transforms/src/jest.rs
```diff
@@ -46,7 +46,7 @@ impl Jest {
                                     computed: false, ..
                                 },
                             ) => match &callee.obj {
-                                ExprOrSuper::Super(_) => {}
+                                ExprOrSuper::Super(_) => new.push(T::from_stmt(stmt)),
```
Brooooooklyn, line 49: I remember this line was added for reasons....

kdy1: I added a test for it. (on js side, because ext-transforms does not have a good test system)

Issue #1012, opened 2020-08-29
Suggestion: TaggedTpl should have Tpl?


dsherret:
The tagged template expression has the following structure:

```rust
#[ast_node("TemplateLiteral")]
#[derive(Eq, Hash)]
pub struct Tpl {
    pub span: Span,

    #[serde(rename = "expressions")]
    pub exprs: Vec<Box<Expr>>,

    pub quasis: Vec<TplElement>,
}

#[ast_node("TaggedTemplateExpression")]
#[derive(Eq, Hash)]
pub struct TaggedTpl {
    pub span: Span,

    pub tag: Box<Expr>,

    #[serde(rename = "expressions")]
    pub exprs: Vec<Box<Expr>>,
    pub quasis: Vec<TplElement>,

    #[serde(default, rename = "typeParameters")]
    pub type_params: Option<TsTypeParamInstantiation>,
}
```

Would it make more sense to have a `tpl` property like so?

```rust
#[ast_node("TaggedTemplateExpression")]
#[derive(Eq, Hash)]
pub struct TaggedTpl {
    pub span: Span,

    pub tag: Box<Expr>,

    #[serde(default, rename = "typeParameters")]
    pub type_params: Option<TsTypeParamInstantiation>,

    #[serde(rename = "template")]
    pub tpl: Tpl,
}
```

Right now, the template of the tagged template is not really represented in the AST as it's skipped in order to go to the contents of the template.

kdy1:
Seems reasonable to me.

dsherret:
Implemented.

PR #1114, opened 2020-09-27
refactor: Change TaggedTpl to have a Tpl

dsherret:
Implements #1012

Obviously a breaking change, but this is better.

Most likely this will fail the CI as developing in the swc codebase on windows has some issues.


kdy1:
I tried this, but it is complex  than just changing ast because it triggers transform to work on it.
Maybe catching them with TaggedTpl will work.

dsherret:
Ok, I will try developing with wsl to see if I can get the tests working and look further into it.

dsherret:
@kdy1 I've finally got these tests to pass and this is ready now.

PR #1121, opened 2020-10-01
fix(jsx): Drop extra spaces around jsx attributes

bebraw:
I didn't get the tests to run on my system so please validate this separately.


ecmascript/codegen/src/jsx.rs
```diff
@@ -51,9 +51,7 @@ impl<'a> Emitter<'a> {
         emit!(node.name);
 
         if let Some(ref value) = node.value {
-            space!();
```
kdy1, line 54: If I remember correctly, the space here was intentional.
What's the point of removing space completely?

bebraw: Based on the test, I would expect the same output for the input. If I'm reading right, we shouldn't have a space in front of a `=`.

Can you try running the test I adjusted when you have time? I think that would validate the behavior.

kdy1: Is spaces disallowed in front of `=`?
jsx spec does not say a lot about whitespace.
https://facebook.github.io/jsx/

kdy1: I looked at the test you adjusted, but I'm not sure if it's correct.

bebraw: Likely the best thing to do would be to have multiple tests. It's entirely possible having space in front/after is valid. However, what I'm looking at is output that's matching the input. In my use case, I'm extracting a piece of existing source by first choosing it from the AST and then by printing it.

Consider the following cases:

1. `<div class="demo">demo</div>`
1. `<div class = "demo">demo</div>`
1. `<div class    =    "demo">demo</div>`
1. `<div class="demo"/>`
1. `<div class = "demo" />`

These all should remain the same in the output.

I think it's less about spaces and more about retaining the formatting now that I think of it.

kdy1: Seems reasonable. Thanks for the explanation.

PR #1142, opened 2020-10-06
export * as ns is part of ES2020

### LINKED ISSUE #1137: export * as ns is part of ES2020 and should not be a proposal pass
**Describe the bug**

Currently swc requires the `swc_ecmascript::transforms::proposals::export` to support _export * as ns_.  This should not be required as the proposal was ratified and is [part of ES2020](https://tc39.es/ecma262/2020/#table-42).

**Expected behavior**

That `swc_ecmascript::transforms::proposals::export` is not required to properly handle code such as:

```js
export * as a from "./a.js";
```

**Version**
The version of @swc/core: 1.2.35

kdy1:
swc_bundler:
 - Add direct support for `export * as ns from 'foo';`. (Closes #1137)

swc_ecma_transforms:
 - Split `export` pass to `export_namespace_from` and `export_default_from`.

swc:
 - Enable `export * as ns` for targets lower than es2020.


kdy1:
@kitsonk I fixed it.

spack/tests/pass/pr-1105/example-2/output/entry.js
```diff
@@ -1,13 +1,13 @@
...
     class C {
     }
     return {
-        c,
+        c: c1,
         default: C
     };
 }();
-const c = _c;
+const c1 = c;
```
kdy1, line 10: This `c1` has different hygiene id from c1 above.

Issue #1139, opened 2020-10-05
swc_bundler: does not properly bundle circular dependencies


kitsonk:
**Describe the bug**

swc_bundler does not properly bundle circular dependencies.

**Input code**

*circular1.ts*

```ts
import * as circular2 from "./circular2.ts";

export function f1(): void {
  console.log("f1");
}

circular2.f2();
```

*circular2.ts*

```ts
import * as circular1 from "./circular1.ts";

export function f2(): void {
  console.log("f2");
}

circular1.f1();
```

**Expected behavior**

An emit something like:

```js
function f2() {
    console.log("f2");
}
export function f1() {
    console.log("f1");
}
f1();
f2();
```

Current bundling outputs:

```js
export function f1() {
    console.log("f1");
}
circular2.f2();
circular1.f1();
```

**Version**

Version: swc_bundler 0.9.1, swc_ecmascript 0.8.4, swc_common 0.10.3

**Additional context**

Working to integrate swc_bundler into Deno.  This is integration test "`bundle_circular`".

kitsonk:
The emit is now:

```js
export function f1() {
    console.log("f1");
}
f2();
function f2() {
    console.log("f2");
}
f1();
```

Which is technically slightly incorrect as this would result in:

```
f2
f1
```

Where as if the modules were loaded separately, it would result in:

```
f1
f2
```

Issue #1138, opened 2020-10-05
swc_bundler: dropped nodes on transient renamed exports


kitsonk:
**Describe the bug**

swc_bundler drops nodes in some transient renamed exports cases

**Input code**

*example11.ts*

```ts
import { a as defaultA } from "./l.ts";

const o: { a?: string } = {};

const { a = defaultA } = o;

console.log(a);
```

*l.ts*

```ts
export { a } from "./a.ts";
```

*a.ts*

```ts
export const a = "a";
```

**Expected behavior**

An emit something like:

```js
const a1 = "a";
const o = {
};
const { a = a1  } = o;
console.log(a);
```

Current bundling outputs:

```js
const o = {
};
const { a =defaultA  } = o;
console.log(a);
```

**Version**

Version: swc_bundler 0.9.1, swc_ecmascript 0.8.4, swc_common 0.10.3

**Additional context**

Working to integrate swc_bundler into Deno.

kitsonk:
Using swc master, it now emits:

```js
const a = "a";
const o = {
};
const { a: a1  } = o;
console.log(a1);
```

Which results in the following when executed:

```
undefined
```

Instead of an expected:

```
a
```

The emit is "renaming" the destructuring, versus setting a default value, which is incorrect, given the source code of:

```js
const { a = defaultA } = o;
```

PR #1144, opened 2020-10-07
Fix hygiene and bundler

kdy1:
swc_bundler:
 - Change order of execution of modules with circular imports. (#1139)

swc_ecma_transforms:
 - hygiene: Fix handling of default in object patterns. (#1138)


kdy1:
@kitsonk I fixed them

PR #1147, opened 2020-10-07
Fix Visitor.ts properties

mattinsler:
It seems that the types for `TsEnumDeclaration` and `ObjectPattern` are slightly off. I found this when parsing a large TypeScript codebase and then debugging the issue. This PR fixes the types and implementation files.

### DIFF
--- a/node-swc/src/Visitor.ts
+++ b/node-swc/src/Visitor.ts
@@ -722,7 +722,7 @@ export default class Visitor {
 
   visitTsEnumDeclaration(n: TsEnumDeclaration): Declaration {
     n.id = this.visitIdentifier(n.id);
-    n.member = this.visitTsEnumMembers(n.member);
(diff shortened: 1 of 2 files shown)


kdy1:
Thanks, but types itself is out of sync with real data.

mattinsler:
Yes exactly. The changes I made to the types are to match the actual data coming from the parser. Is that not the intention?

kdy1:
I meant that ast types declared in `types.ts` are outdated.

mattinsler:
Ah, gotcha. Is there a plan to fix them? I'm trying to use the visitor and it throws an error on these lines on the latest version published to npm.

kdy1:
Of course, but currently I'm too busy porting typescript type system.

mattinsler:
OK, then I'll patch locally for now. Thanks.

PR #1159, opened 2020-10-12
bundler: Handle computed accesses correctly

kdy1:
swc_bundler:
 - Handle computed accesses to imports correctly.
 - Make planning deterministic.
 - Prefer exports over imports while planning.
 - Fix https://deno.land/x/oak@v6.3.1/examples/server.ts

### DIFF
--- a/bundler/Cargo.toml
+++ b/bundler/Cargo.toml
@@ -36,6 +36,7 @@ swc_ecma_visit = {version = "0.19.0", path = "../ecmascript/visit"}
 
 [dev-dependencies]
 reqwest = {version = "0.10.8", features = ["blocking"]}
+tempfile = "3.1.0"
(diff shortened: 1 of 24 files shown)


kdy1:
@kitsonk I think I fixed it. Can you try this branch?

kdy1:
@kitsonk

Currently when I run the generated bundle, `await this.signal` at https://deno.land/std@0.74.0/async/mux_async_iterator.ts#L45 does not resolve.

kdy1:
I fixed it and verified that 127.0.0.1 returns `Hello, world!`

kitsonk:
🎉 🎉 🎉 🎉 

Yes, I can confirm master properly builds that bundle!!!

Thank you @kdy1 great work.

PR #1174, opened 2020-10-18
Remove unnecessary question mark

braddunbar:
`c` can never be a question mark at this point since the match arm doesn't include it.


kdy1:
Thanks!

PR #1175, opened 2020-10-19
Invalid template escape in ES2018+

### LINKED ISSUE #78: tagged template literal can contain invalid escape sequence
parser currently does not support parsing code like

```js
latex`\unicode`
```
Note: this is valid only for **es2018+**
Implementor should check for the `JscTarget`.

This can be done by modifying code at

https://github.com/swc-project/swc/blob/53b09aa35652a84f731a765cbb23ee272c157376/ecmascript/parser/src/lexer/mod.rs#L927-L991

braddunbar:
Fixes #78.

When there's an error reading escaped characters, we return `None` for the cooked value and continue reading.

I think I'm on the right track here, but this is my first pull of any size so please let me know if I should do things differently. I'm happy to change it up.

### DIFF
--- a/bundler/Cargo.toml
+++ b/bundler/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2018"
 license = "Apache-2.0/MIT"
 name = "swc_bundler"
 repository = "https://github.com/swc-project/swc.git"
-version = "0.11.3"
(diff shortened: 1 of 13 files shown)


kdy1 (approved):
Thanks a lot!

PR #1183, opened 2020-10-23
Fix: _typeof helper throws in environments where Symbol is not defined

### LINKED ISSUE #1181: _typeof helper throws in environments where Symbol is not defined
**Describe the bug**

swc rewrites `typeof` expressions into a form that uses an internal `_typeof` helper, much like babel does it to make the `typeof` operator support polyfilled `Symbols`. However, the `_typeof` helper doesn't check for the existence of Symbol before it compares it with the object. This leads to errors in environments that do not support Babel, for example when transpiling ES5 to ES3 or when evaluating the `_typeof` helper _before_ a Symbol polyfill is applied.

Here's the declaration of the `_typeof` helper:

```js
var _typeof = function(obj) {
    return obj && obj.constructor === Symbol ? "symbol" : typeof obj;
};
```

**Input code**
[Please see this repro](https://repl.it/@wessberg/swc-typeof-helper#index.js).

In it, you'll find that `output.js` is the result of running `swc index.js -o output.js`.

**Config**
```json
{}
```

**Expected behavior**

The `_typeof` helper should check for the existence of `Symbol` before comparing it with `obj`.
One way to do so could be:

```js
var _typeof = function(obj) {
    return obj && typeof Symbol !== "undefined" && obj.constructor === Symbol ? "symbol" : typeof obj;
};
```

wessberg:
Fixes #1181

### DIFF
--- a/ecmascript/transforms/src/helpers/_type_of.js
+++ b/ecmascript/transforms/src/helpers/_type_of.js
@@ -1,3 +1,3 @@
-var _typeof = function (obj) {
-    return obj && obj.constructor === Symbol ? "symbol" : typeof obj;
+var _typeof = function(obj) {
+    return obj && typeof Symbol !== "undefined" && obj.constructor === Symbol ? "symbol" : typeof obj;
 };


kdy1 (approved):
Thanks a lot!

PR #1185, opened 2020-10-24
Preserve space for postfix unary operators

### LINKED ISSUE #1180: Minification can generate invalid JavaScript when combining IfStatements with PostfixUnaryExpressions
**Describe the bug**

swc can generate invalid JavaScript when minifying constructs like this:

```js
if (...) {}
else foo++;
```

since it discards the whitespace between `else` and the following `UnaryExpression`, leading to:

```js
elsefoo++;
```

Which is invalid JavaScript. This would have worked for PrefixUnaryExpressions, and I'm guessing that's the case you've optimized for:

```js
else++foo;
```

So to fix this bug we'll have to make sure to preserve at least one whitespace character when the unary operator is at the right side of the expression.

**Input code**
[Please see this repro](https://repl.it/@wessberg/swc-minification-problem#index.js).

In it, you'll find that `output.js` is the result of running `swc index.js -o output.js` with `minify` in the swc options.

**Config**
```json
{
  "minify": true
}
```

**Expected behavior**
At least one whitespace character should separate `else` from the `UnaryExpression`. I would expect the minified output to look like this:

```js
var foo=0;if(true){}else foo++;
```

**Version**
The version of @swc/core: 1.2.36

braddunbar:
Fixes #1180. Easy peasy!

### DIFF
--- a/ecmascript/Cargo.toml
+++ b/ecmascript/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2018"
 license = "Apache-2.0/MIT"
 name = "swc_ecmascript"
 repository = "https://github.com/swc-project/swc.git"
-version = "0.11.1"
+version = "0.11.2"
 
 [features]
 codegen = ["swc_ecma_codegen"]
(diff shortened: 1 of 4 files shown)


kdy1 (approved):
Thanks!

Issue #1187, opened 2020-10-26
Not all Unicode Escape Sequences are escaped correctly, leading to "Unterminated string constant" errors


wessberg:
Please see [this repro](https://repl.it/@wessberg/swc-unicode-bug#index.js).
In it, you'll see an input file such as this:

`export const unicodeWhitespaceCharacters = '\u0009\u000A\u000B\u000C\u000D\u0020\u00A0\u1680\u2000\u2001\u2002\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200A\u202F\u205F\u3000\u2028\u2029\uFEFF';`

`swc` transforms Unicode Escape Sequences into UTF-8 characters, and this is the compiled output (`output.js` in the repro) with default settings:

<img width="656" alt="Screen Shot 2020-10-26 at 11 23 18 AM" src="https://user-images.githubusercontent.com/20454213/97161136-a9ad7480-177d-11eb-87f6-12ea9c300225.png">

Notice the "?" in the end of that string. That's where `\u2028`, `\u2029`, and `\uFEFF` used to be.

Here's the output I would have expected, which is also what tools like `terser` produces, and `esbuild` given the `--charset=utf8` produces:

```js
export var '\t\n\v\f\r                　\u2028\u2029\ufeff';
```

The current behavior of `swc` leads to `Unterminated string constant` exceptions in browser such as IE 11 - even with `<meta charset="utf-8">` in the HTML document. On that note, it is not always possible for the website author to instruct IE to use UTF-8. [There's a bunch of ways IE can choose to ignore that](https://docs.microsoft.com/en-us/troubleshoot/browsers/wrong-character-set-for-html-page).

I ran into this problem specifically while bundling core-js (the declaration from the example is located within `core-js/internals/whitespaces.js` with swc and testing it across browsers.

On a related note, I've also created https://github.com/swc-project/swc/issues/1188

QGB:
python regex whitespace \s not match "\uFEFF"

PR #1189, opened 2020-10-27
Fix issues

### LINKED ISSUE #1164: Automatically detect if file is script or module when parsing
**Describe the feature**
Add ability to `swc_ecmascript`'s `Parser` to automatically detect if parsed source file 
is an ES module or plain old script. 

Currently it's up to implementor to decide if file is a script of a module and based on that
decision should call either `Parser::parse_script` or `Parser::parse_module`.

It makes writing libraries working on both scripts and modules a bit more cumbersome,
because of need to handle both `swc_ecma_ast::Script` and `swc_ecma_ast::Module`.

Proposed API: `Parser::parse_program() -> PResult<Program>`

**Additional context**
Came up in discussion regarding https://github.com/denoland/deno_lint/issues/366

### LINKED ISSUE #1170: panic: Parser panics on malformed TypeScript code
Source code:
```ts
const toString: (local)(this: Function) => string) = undefined;
```

Backtrace:
```
▶ RUST_BACKTRACE=1 deno run foo.ts
thread 'main' panicked at 'internal error: entered unreachable code: invalid pattern: Span { lo: BytePos(24), hi: BytePos(28), ctxt: #0 }', /Users/runner/work/deno/deno/.cargo_home/registry/src/github.com-1ecc6299db9ec823/swc_ecma_parser-0.38.0/src/parser/expr.rs:1355:44
stack backtrace:
   0: _rust_begin_unwind
   1: std::panicking::begin_panic_fmt
   2: swc_ecma_parser::parser::expr::<impl swc_ecma_parser::parser::Parser<I>>::parse_paren_expr_or_arrow_fn
   3: swc_ecma_parser::parser::expr::<impl swc_ecma_parser::parser::Parser<I>>::parse_primary_expr
   4: swc_ecma_parser::parser::expr::<impl swc_ecma_parser::parser::Parser<I>>::parse_member_expr_or_new_expr
   5: swc_ecma_parser::parser::expr::<impl swc_ecma_parser::parser::Parser<I>>::parse_lhs_expr
   6: swc_ecma_parser::parser::expr::ops::<impl swc_ecma_parser::parser::Parser<I>>::parse_unary_expr
   7: swc_ecma_parser::parser::expr::ops::<impl swc_ecma_parser::parser::Parser<I>>::parse_bin_expr
   8: swc_ecma_parser::parser::expr::<impl swc_ecma_parser::parser::Parser<I>>::parse_assignment_expr_base
   9: swc_ecma_parser::parser::expr::<impl swc_ecma_parser::parser::Parser<I>>::parse_assignment_expr
  10: swc_ecma_parser::parser::expr::<impl swc_ecma_parser::parser::Parser<I>>::parse_expr
  11: swc_ecma_parser::parser::stmt::<impl swc_ecma_parser::parser::Parser<I>>::parse_var_stmt
  12: swc_ecma_parser::parser::stmt::<impl swc_ecma_parser::parser::Parser<I>>::parse_stmt_internal
  13: swc_ecma_parser::parser::stmt::<impl swc_ecma_parser::parser::Parser<I>>::parse_block_body
  14: swc_ecma_parser::parser::Parser<I>::parse_module
  15: deno::ast::parse
  16: deno::tsc::pre_process_file
  17: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  18: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  19: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  20: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  21: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  22: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  23: tokio::runtime::Runtime::block_on
  24: deno::main
```

### LINKED ISSUE #1184: swc_bundler: support outputting an IIFE
**Describe the feature**

Currently, swc_bundler can output an ES module, but it does support output as an IIFE.

For outputting the bundle, instead of `export`ing entries from the entry module in the bundle, `swc_bundler` would output an IIFE where the return value of the IIFE is the value of the exported module.  For example if the entry module was:

_a.js_

```js
export const a = "a";
```

The output of the bundle would be something like:

_a.bundle.js_

```js
(function () {
  "use strict";
  var __exports = {};
  Object.defineProperty(__exports, "__esModule", {
    value: true
  });
  __exports.a = "a";
  return __exports;
})();
```

**Babel plugin or link to the feature description**

There are [Babel plugins](https://github.com/MindTouch/babel-plugin-transform-modules-iife) that in theory can do this, but IIFE output is one of the builtin output formats of Rollup (and other bundlers).

**Additional context**

In particular with Deno, we would like to have an option for `Deno.bundle()` to be able to output a single file ES module or an IIFE.

### LINKED ISSUE #1188: Suggestion: Don't rewrite unicode escape sequences to unicode
**Describe the feature**

Over at `esbuild` (which is often used in conjunction with `swc` to target ES5), the decision was recently made [to switch over to producing ASCII-only output by default](https://github.com/evanw/esbuild/blob/v0.7.21/CHANGELOG.md#0720).

The decision was triggered specifically by the discussion [in this issue](https://github.com/evanw/esbuild/issues/485) (and originally [this one](https://github.com/evanw/esbuild/issues/485)) due to a combination of interoperability and performance concerns.

First, in terms of interoperability, this is in line with other bundlers such as Webpack, Parcel, and Rollup, and also in line with babel that also preserves unicode escape sequences.

Parcel and Webpack generates minified bundles by default, and they use `terser` for that. Terser applies an optimization that transforms unicode escape sequences into utf-8 characters, so you might end out with unicode-characters that weren't there in your source code in your output bundle(s) from these tools - but the default behavior of them before terser is in the picture is to respect whatever was written in the source code.

From a performance perspective, however, it get's really interesting. It turns out that many browsers [have an optimized scanning path for ASCII code](https://github.com/evanw/esbuild/issues/485#issuecomment-716058090) which seems to indicate that parsing an ASCII string is around 1.7x faster in V8 than parsing an UTF-8 string. So there's potential performance gains as well.

kdy1:
swc_bundler:
 - Support emitting iife. (Closes #1184)

swc_ecma_parser:
 - Auto-detect script / module (Closes #1164)
 - lexer: Error recovery for strict mode and module mode.
 - More error recovery logic for strict mode violation.
 - Fix panic on invalid input. (Closes #1170)

---

# Postponed tasks

swc_ecma_codegen:
 - Print escapes correctly. (#1187)

swc_ecma_parser:
 - Preserve unicode escape sequences as-is. (#1188)

### DIFF
--- a/bundler/Cargo.toml
+++ b/bundler/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2018"
 license = "Apache-2.0/MIT"
 name = "swc_bundler"
 repository = "https://github.com/swc-project/swc.git"
-version = "0.12.0"
+version = "0.13.0"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [features]
--- a/bundler/Cargo.toml
+++ b/bundler/Cargo.toml
@@ -29,9 +29,9 @@ retain_mut = "=0.1.1"
 swc_atoms = {version = "0.2.4", path = "../atoms"}
 swc_common = {version = "0.10.0", path = "../common"}
 swc_ecma_ast = {version = "0.33.0", path = "../ecmascript/ast"}
-swc_ecma_codegen = {version = "0.38.0", path = "../ecmascript/codegen"}
-swc_ecma_parser = {version = "0.40.0", path = "../ecmascript/parser"}
-swc_ecma_transforms = {version = "0.27.0", path = "../ecmascript/transforms"}
+swc_ecma_codegen = {version = "0.39.0", path = "../ecmascript/codegen"}
+swc_ecma_parser = {version = "0.41.0", path = "../ecmascript/parser"}
+swc_ecma_transforms = {version = "0.28.0", path = "../ecmascript/transforms"}
 swc_ecma_utils = {version = "0.23.0", path = "../ecmascript/utils"}
 swc_ecma_visit = {version = "0.19.0", path = "../ecmascript/visit"}
 
--- a/bundler/src/bundler/finalize.rs
+++ b/bundler/src/bundler/finalize.rs
@@ -129,7 +129,7 @@ where
         // Properties of returned object
         let mut props = vec![];
 
-        let body = BlockStmt {
+        let mut body = BlockStmt {
             span: module.span,
             stmts: module
                 .body
@@ -281,6 +281,13 @@ where
                 })
                 .collect(),
         };
+        body.stmts.push(Stmt::Return(ReturnStmt {
+            span: DUMMY_SP,
+            arg: Some(Box::new(Expr::Object(ObjectLit {
+                span: DUMMY_SP,
+                props,
+            }))),
+        }));
 
         let f = Function {
             is_generator: false,
--- a/bundler/src/bundler/finalize.rs
+++ b/bundler/src/bundler/finalize.rs
@@ -1,4 +1,4 @@
-use crate::{hash::calc_hash, Bundle, BundleKind, Bundler, Load, Resolve};
+use crate::{hash::calc_hash, Bundle, BundleKind, Bundler, Load, ModuleType, Resolve};
 use anyhow::Error;
 use relative_path::RelativePath;
 use std::{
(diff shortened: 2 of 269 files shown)


kdy1:
i decided to split this pr into two parts.

PR #1192, opened 2020-10-28
BigInt literal keys

### LINKED ISSUE #1173: Bigint literal as a key in object literals cannot be parsed
**Describe the bug**

If a property key in object literal is a bigint literal, swc emits an error.

```
SwcDiagnosticBuffer {
  "diagnostics": [
    "Unexpected token `bigint literal`. Expected identifier, string literal, numeric literal or [ for the computed key at deno_lint_test.tsx:1:12"
  ]
}
```

**Input code**

```ts
const x = { 1n: 42 };
```

**Config**

nothing
(executed via deno_lint)

**Expected behavior**

Successfully parsed

**Version**

```toml
swc_atoms = "0.2"
swc_common = "=0.10.4"
swc_ecmascript = { version = "=0.10.1", features = ["parser", "transforms", "utils", "visit"] }
```

**Additional context**

The input code comes from https://github.com/eslint/eslint/blob/v7.11.0/tests/lib/rules/no-dupe-keys.js#L49
Looking at this ESLint's code, it seems like using bigint literal as object key is available on ES2020.
On the other hand, tsc v4.0.2 cannot parse this either as you can see https://www.typescriptlang.org/play?target=99&ts=4.0.2#code/MYewdgzgLgBCBGArGBeGBvAUDGBGMAXDACwBMANJgL4DcQA

So I'm not sure if swc should parse this code. If you decide this won't be fixed for the time being, that's no problem 👍 

Relates to https://github.com/denoland/deno_lint/pull/406#discussion_r507111858

braddunbar:
Fixes #1173

I'd like to add a few more transform tests, but I thought I'd send this to make sure I'm on the right track!


kdy1 (approved):
LGTM. Thanks!

braddunbar:
Actually, I think this is fine with the tests it has. Let me know if you want something different!

kdy1:
@braddunbar
I'm fine with the pr itself. Sorry for the delay. I thought it was not a new feature of es2020 because typescript compiler does not support it.

But when I looked at es2020 spec to be sure, I found that the definition of [`NumericLiteral`](https://www.ecma-international.org/ecma-262/#prod-NumericLiteral) contains [`DecimalBigIntegerLiteral`](https://www.ecma-international.org/ecma-262/#prod-DecimalBigIntegerLiteral), which is bigint literal.

I'll merge it soon.

magurotuna:
Thanks for the fix!
Seems like the bumped version has not been published to crates.io, as we can see https://crates.io/crates/swc_ecmascript which still shows v0.13.3.
Any plan to publish? @kdy1

kdy1:
@magurotuna
Published it :)

magurotuna:
Thanks! 👍

PR #1194, opened 2020-10-29
Fix bundler

kdy1:
swc_bundler:
 - [x] Remove wrong assertion.

Fixes https://github.com/denoland/deno/issues/8148

 - [x] Remove wrong variables created on reexports.

Fixes https://github.com/denoland/deno/issues/8188

swc_ecma_transforms:
 - [x] inlining: Treat usage as an argument as writes.

Fixes https://github.com/denoland/deno/issues/8180.

 - [x] inlining: Visit rhs of `||` and `&&`.

Fixes https://github.com/denoland/deno/issues/8189


kdy1:
@kitsonk I verified previous deno examples (I mean http://127.0.0.1:8000) works properly.

PR #1199, opened 2020-11-02
bundler: Handle swc helpers

kdy1:
swc_bundler:
 - Handle helpers from `swc_ecma_transforms`.

swc_ecma_transforms:
 - dce: Remove unused self-referential functions.


kdy1:
I found that self-referential functions like 

```js
function foo() {
    if (Math.random() > 0.5) {
        foo()
    }
}
```

is not dropped. I will merge this issue after making dce handle to case above.

kdy1:
@kitsonk I noticed that including only required helpers using dead code elimination requires lots of work, so I'll add explicit handling of helpers to `swc_bundler`.

PR #1195, opened 2020-10-30
BREAKING(bundler): Change Hook API to get_import_meta_props()

nayeemrmn:
Addresses https://github.com/denoland/deno/issues/8088.

PR #1201, opened 2020-11-02
fix(bundler): Use a local variable for import.meta

nayeemrmn:
Addresses https://github.com/swc-project/swc/pull/1195#discussion_r515356646. Supports mutations on `import.meta`.

### DIFF
--- a/bundler/src/bundler/chunk/merge.rs
+++ b/bundler/src/bundler/chunk/merge.rs
@@ -15,7 +15,7 @@ use std::{borrow::Cow, mem::take};
 use swc_atoms::js_word;
 use swc_common::{FileName, SyntaxContext, DUMMY_SP};
 use swc_ecma_ast::*;
-use swc_ecma_utils::prepend_stmts;
+use swc_ecma_utils::{prepend_stmts, private_ident};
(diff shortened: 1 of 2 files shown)


bundler/src/bundler/chunk/merge.rs
```diff
@@ -633,10 +635,51 @@ struct ImportMetaHandler<'a, 'b> {
...
+        if self.occurred {
+            match self.hook.get_import_meta_props(
+                n.span,
+                &ModuleRecord {
+                    file_name: self.file.to_owned(),
+                    is_entry: self.is_entry,
+                },
+            ) {
+                Ok(key_value_props) => {
+                    n.body.insert(
```
kdy1, line 656: Can you use `prepend_stmt` from `swc_ecm_utils` instead of `.insert(0, stmt)`?
The utility method exists to prevent inserting statements before `'use strict'`;

kdy1 (changes requested):
Other things are good.
Thanks!

bundler/src/bundler/chunk/merge.rs
nayeemrmn: Done.

kdy1 (approved):
Thanks a lot!

PR #1220, opened 2020-11-16
fix: extienstion

skywalker512:
```js
import { a } from './a.a';

console.log(a);
```

./a.a.js
```js
export const a = 1;
```

The following error occurs
```
Caused by:
    0: failed to analyze module
    1: failed to resolve ./a.a from xxx/src/index.js
    2: index not found: xxx/src/./a.a
```

I don't understand why the following test failed
```
spack/tests/pass/dynmaic-imports/issue-1112
```


kdy1:
Closing as inactive

PR #1229, opened 2020-11-24
feat(node-swc): Macro to keep typescript types in sync

### LINKED ISSUE #1239: how to walk swc ast
**Question**

```js
const { parseFileSync } = require('@swc/core')
const path = require('path')

const result = parseFileSync(path.resolve(__dirname, './sample-js.js'))
```

how to walk result? like acron-walk

kdy1:
ts_macros:
 - [x] Create an interface from named struct.
 - [ ] Create `? (something)` from tuple struct.
 - [ ] Create a type alias from enum.
 - [x] Support `Option<T>`
 - [x] Support `Box<T>`
 - [x] Support `Vec<T>`
 - [ ] Support `#[interface(name = "renamed")]`
 - [x] Support `#[serde(rename = "renamed")]`
 - [x] Support `#[serde(rename_all = "foo")]`
 - [ ] Support `#[serde(default)]`
 - [ ] Support `#[serde(flatten)]`
 - [ ] Preserve comments.

---

Closes #1239


jiby-gh:
@kdy1 was this merged in, I was playing with the web assembly version, and seemingly the ts types and the Visitor seems to be not matching the AST structure. I was going to try to fix by hand the types and visitor, but seeing this i wanted to ask if there is a better automated way ?

kdy1:
It is possible, but I'm not sure if it is better.
Lots of work is still remaining.

PR #1237, opened 2020-11-28
fix(source map): fix inline source map comment slicing

pyrocat101:
This bug results in an out-of-bound error when using `inputSourceMap: true`.


kdy1:
Nice catch, thanks!

Issue #1261, opened 2020-12-09
Ability to borrow leading and trailing comments from SingleThreadedComments


dsherret:
**Describe the feature**

Right now `SingleThreadedComments` is defined as such:

```rs
pub type SingleThreadedCommentsMap = Rc<RefCell<FxHashMap<BytePos, Vec<Comment>>>>;

/// Single-threaded storage for comments.
#[derive(Debug, Clone, Default)]
pub struct SingleThreadedComments {
    leading: SingleThreadedCommentsMap,
    trailing: SingleThreadedCommentsMap,
}
```

For the "ast view" of swc I'm working on, I would like a way to get a reference to both the leading and trailing inner maps (`&FxHashMap<BytePos, Vec<Comment>>>`).

Perhaps similar to `take_all()` it could be something like `borrow_all()`:

```rust
let (leading_comments, trailing_comments) = comments.borrow_all();

// use both of these here
```

This will allow me to only have to borrow once from the `RefCell` and not have to call `take_all()` on the `SingleThreadedComments` (I'd rather people provide a reference to that).

Let me know your thoughts and I will open up a PR to implement this. Another idea might be separate `.borrow_leading()` and `.borrow_trailing()` methods.

PR #1262, opened 2020-12-09
Add `borrow_all()` to `SingleThreadedComments`

dsherret:
I decided to go ahead and work on this because it's very simple. Let me know about any changes you'd like or if you even want something like this. Thanks!

From #1261

### DIFF
--- a/common/src/comments.rs
+++ b/common/src/comments.rs
@@ -3,7 +3,11 @@ use crate::{
     syntax_pos::{BytePos, Span},
 };
 use fxhash::FxHashMap;
-use std::{cell::RefCell, rc::Rc, sync::Arc};
+use std::{
+    cell::{Ref, RefCell},
+    rc::Rc,
+    sync::Arc,
+};
 
 /// Stores comment.
 ///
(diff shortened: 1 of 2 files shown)


common/src/comments.rs
```diff
@@ -107,7 +111,8 @@ where
     delegate!();
 }
 
-pub type SingleThreadedCommentsMap = Rc<RefCell<FxHashMap<BytePos, Vec<Comment>>>>;
+pub type SingleThreadedCommentsMapInner = FxHashMap<BytePos, Vec<Comment>>;
+pub type SingleThreadedCommentsMap = Rc<RefCell<SingleThreadedCommentsMapInner>>;
```
dsherret, line 115: I'm not sure about this naming, but I didn't want to cause a breaking change.

kdy1: It seems ok to me.

dsherret:
Thanks for the quick merge! I upgraded and will start adding comment related functionality now.

kdy1:
No problem. Thanks for the pr.

PR #1267, opened 2020-12-11
Remove duplicate type and correct name

mfeckie:
In the type file, there was a typo (`ImporSpecifier`) and the type duplicated `ImportSpecifier`.

So I've remove the duplicate and updated the reference.

The file was autoformatted with prettier too, but if you want I can avoid that?

### DIFF
--- a/node-swc/src/types.ts
+++ b/node-swc/src/types.ts
@@ -306,7 +306,7 @@ export interface EsParserConfig {
   jsx?: boolean;
   /**
    * Defaults to `false`.
-   * 
+   *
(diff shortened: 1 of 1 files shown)


kdy1:
Thank you! Prettier is ok. It's in my todolist.

But I think publishing a new version right now would not make big difference because there are lots of inconsistencies between rust type definition and typescript definitions.
(I'm working on a code generator to keep types in sync.)

mfeckie:
OK, so should I just close this PR?

kdy1:
I don't think so. I'll merge it. Thanks for the pr!

PR #1283, opened 2020-12-19
Fix #1277: publish @swc/wasm-web package

### LINKED ISSUE #1277: Publish wasm module for web
**Describe the feature**
`@swc/wasm` is deployed automatically to npm but targetting nodejs runtime.
https://github.com/swc-project/swc/blob/25856f230c604868aea7a957ce5f29fe82d2c8b5/.github/workflows/publish-wasm.yml#L39

It would be helpful to have the browser bindings deployed to npm as well.

punkeel:
The @swc/wasm package is meant for use in Node.js projects. Web pages cannot
use this package as it relies on Node.js specific modules (fs, in particular).
A Web-based project that would benefit from having a Web compatible @swc/wasm
version is astexplorer.net, which displays the AST emitted by various parsers.

When wasm-pack adds support for an "all" target exposing Web+Node support, the
build step will be simpler: everything will go to @swc/wasm. For now, we have
to change the package name in the generated package.json file.

### DIFF
--- a/.github/workflows/publish-wasm-web.yml
+++ b/.github/workflows/publish-wasm-web.yml
@@ -1,0 +1,52 @@
+name: Publish (wasm for web)
+
+env:
+  CARGO_INCREMENTAL: 0
+  CI: "1"
+
+on:
+  create:
+    tags:
+      - v*
+
+jobs:
+  build:
+    name: Build - wasm for web
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


punkeel:
CI failure also happens on master.

kdy1:
Thank you!
I'll merge this and publish it after fixing test with another pr.

punkeel:
Thank you!

PR #1285, opened 2020-12-20
fix(ci): Inline parser tests.

kdy1:
swc_ecma_parser:
 - Inline parser tests.


kdy1:
+4,538,299 −2 
This is too much. Closing...

PR #1286, opened 2020-12-21
fix(ecmascript/lexer): Normalize \r\n and \r to \n in template literals

Liamolucko:
According to https://tc39.es/ecma262/#sec-static-semantics-tv-and-trv, '`<CR><LF>` and `<CR>` LineTerminatorSequences are normalized to `<LF>` for both TV and TRV.' This PR implements that.

This is the cause of denoland/deno#8746.


kdy1:
I'll merge this after inlining parser tsc tests.
It's currently `git pull`-ed from ci, but it makes checking pr too hard.

kdy1:
Thanks!

kdy1 (approved):

Issue #50, opened 2018-11-14
Online repl


kdy1:
ratel-rust have one, and it looks great

kdy1:
It would be really hard as swc depends on rustc-ap crates.
Some rustc-ap crates can't be compiled as wasm.

```
cargo build --release --target wasm32-unknown-emscripten
   Compiling darling_core v0.7.0
   Compiling pmutil v0.2.0
   Compiling swc_atoms v0.1.0 (https://github.com/swc-project/swc.git#67daa279)
   Compiling sourcemap v2.2.1
   Compiling rustc-ap-rustc_data_structures v297.0.0
error[E0433]: failed to resolve. Use of undeclared type or module `os`
   --> /Users/kdy1/.cargo/registry/src/github.com-1ecc6299db9ec823/rustc-ap-rustc_data_structures-297.0.0/flock.rs:195:21
    |
195 |                     os::F_WRLCK
    |                     ^^ Use of undeclared type or module `os`

error[E0433]: failed to resolve. Use of undeclared type or module `os`
   --> /Users/kdy1/.cargo/registry/src/github.com-1ecc6299db9ec823/rustc-ap-rustc_data_structures-297.0.0/flock.rs:197:21
    |
197 |                     os::F_RDLCK
    |                     ^^ Use of undeclared type or module `os`

error[E0433]: failed to resolve. Use of undeclared type or module `os`
   --> /Users/kdy1/.cargo/registry/src/github.com-1ecc6299db9ec823/rustc-ap-rustc_data_structures-297.0.0/flock.rs:200:29
    |
200 |                 let flock = os::flock {
    |                             ^^ Use of undeclared type or module `os`

error[E0433]: failed to resolve. Use of undeclared type or module `os`
   --> /Users/kdy1/.cargo/registry/src/github.com-1ecc6299db9ec823/rustc-ap-rustc_data_structures-297.0.0/flock.rs:208:37
    |
208 |                 let cmd = if wait { os::F_SETLKW } else { os::F_SETLK };
    |                                     ^^ Use of undeclared type or module `os`

error[E0433]: failed to resolve. Use of undeclared type or module `os`
   --> /Users/kdy1/.cargo/registry/src/github.com-1ecc6299db9ec823/rustc-ap-rustc_data_structures-297.0.0/flock.rs:208:59
    |
208 |                 let cmd = if wait { os::F_SETLKW } else { os::F_SETLK };
    |                                                           ^^ Use of undeclared type or module `os`

[...]

PR #1293, opened 2020-12-26
fix(ci): insert *_bg files in @swc/wasm-web (#1291)

### LINKED ISSUE #1291: @swc/wasm-web references wasm_bg.js but it does not get published together with it
It's not about @swc/core, but it's more of just publishing problem, I guess.

Screen Shot 2020-12-26 at 6 18 43 PM

For some reason it exports `wasm_bg.js`, but it does not come with `wasm_bg.js` anyways (see left side).

Where is wasm_bg.js supposed to come from?

## How to reproduce

Just install @swc/wasm-web with `yarn add @swc-wasm-web` and see the actual package content inside `node_modules`.

9oelM:
As described by #1291, right now `@swc/wasm-web` cannot be used after installation because `@swc/wasm-web` does not include files that wasm-bindgen produces when it gets published.

This happens due to a bug in wasm-pack (https://github.com/rustwasm/wasm-pack/issues/837).

Therefore, the simplest solution would just be including the `files` manually in `package.json` before publishing.
Initially I thought the only file that needs to be included would be `wasm_bg.js`, but `wasm_bg.wasm.d.ts` comes along with the build output too, so there's no reason not to include it as a part of API.

This would allow #50 to get going too.

Would close #1291 if it gets merged

cc @kdy1

### DIFF
--- a/.github/workflows/publish-wasm-web.yml
+++ b/.github/workflows/publish-wasm-web.yml
@@ -43,6 +43,10 @@ jobs:
           sed -i'' -e 's/"name": "@swc\/wasm"/"name": "@swc\/wasm-web"/g' package.json
         working-directory: wasm/pkg
 
+      - name: Include *_bg files
+        run: jq '.files += ["wasm_bg.js", "wasm_bg.wasm.d.ts"]' package.json
+        working-directory: wasm/pkg
+
       - name: Publish
         run: |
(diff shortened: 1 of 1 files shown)


kdy1 (approved):
LGTM. Thanks!

punkeel:
Thanks!

PR #1298, opened 2020-12-28
feat(ecmascript/codegen): fix and use omit_trailing_semi

punkeel:
swc_ecma_codegen:
- Use omit_trailing_semi in 'assert_min' tests, to identify bugs.
  Before this, no code was using this method - and it had some serious bugs.
- Omit semicolons when writing punctuation, as it's permitted to do so.
- Use two macros ('semi', 'formatting_semi') to distinguish between required
  semicolons, as used in for(...), and optional ones that are used between
  statements.

This commit does not enforce the use of omit_trailing_semi when cfg.minify is used (except in tests using assert_min), and as such should not modify the behavior of existing apps. This can be confirmed by the `test262` suite still passing.


kdy1 (approved):
LGTM. Thanks!

PR #1300, opened 2020-12-29
fix(swc): Fix bugs

### LINKED ISSUE #1227: Unicode Code Point Escape syntax is not transpiled
**Describe the bug**

[Unicode Code Point Escape syntax](https://exploringjs.com/es6/ch_unicode.html) is not being transformed to ES5-compatible syntax when `es5` is the target.

It would seem that swc is always assuming traditional unicode escape sequences of on the form `\u` followed by four digits and simply doesn't parse code point escape syntax.

**Input code**
[Please see this repro](https://repl.it/@wessberg/swcunicodecodepointescapebug#index.js).

In it, you'll find that `output.js` is the result of running `swc index.js -o output.js` with `es5` as the target for the following input:

```js
const foo = "\u{a0}";
```

The output is:

```js
var foo = "\u";
```

Which is incorrect.

**Config**
```json
{
  "jsc": {
    "target": "es5"
  }
}
```

**Expected behavior**
Unicode Code Point Escape syntax should be transpiled into syntax that legacy environments can understand.

Here's how babel would have rewritten the example from the repro:

```js
var foo = "\xA0";
```

**Version**
The version of @swc/core: 1.2.39

### LINKED ISSUE #1306: Incorrect renaming of private members when a property is accessed
**Describe the bug**

A property access on a private member causes code generation to malfunction and emit the name directly instead of using the `_classPrivateFieldGet` guard. This results in a "Private field 'XXX' must be declared in an enclosing class" error.

**Input code**
```typescript
class Animal {
	readonly #name: string

	constructor(name: string) {
		this.#name = name
	}

	public noise() {
		return this.#name.toUpperCase()
	}
}
```

Compiling and running this code results in:
```
example.js:45
                return this.#name.toUpperCase();
                           ^

SyntaxError: Private field '#name' must be declared in an enclosing class
    at wrapSafe (internal/modules/cjs/loader.js:915:16)
    at Module._compile (internal/modules/cjs/loader.js:963:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:60:12)
    at internal/main/run_main_module.js:17:47
```

**Expected behavior**

Instead the following code should be emitted for that line: `return _classPrivateFieldGet(this, _name).toUpperCase();`

**Version**
The version of @swc/core: https://github.com/swc-project/swc/commit/d638fc32cc9d5ed6a1e779ac6abea5381b98360c

### LINKED ISSUE #1307: panic: re-export alias
<s>Probably a data race, `--sync` fixes it.</s>

1. Source code
```ts
export * as foo from "./foo";
```

2. `.swcrc` file
```json
{
  "exclude": ".*.test.ts$",
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "decorators": false,
      "dynamicImport": true
    },
    "target": "es2020",
    "loose": true
  },
  "module": {
    "type": "commonjs",
    "strict": false,
    "strictMode": true,
    "lazy": false,
    "noInterop": false
  },
  "minify": false
}
```

3. Error message gained from `swc --sync <input.js>`
```
thread '<unnamed>' panicked at 'internal error: entered unreachable code: export default from 'foo'; should be removed by previous pass', ecmascript/transforms/src/modules/import_analysis.rs:106:18
```

4. Backtrace

```
Backtrace:    0: node::init::{{closure}}
   1: std::panicking::rust_panic_with_hook
             at /rustc/d9a105fdd46c926ae606777a46dd90e5b838f92f/library/std/src/panicking.rs:595:17
   2: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/d9a105fdd46c926ae606777a46dd90e5b838f92f/library/std/src/panicking.rs:497:13
   3: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/d9a105fdd46c926ae606777a46dd90e5b838f92f/library/std/src/sys_common/backtrace.rs:141:18
   4: rust_begin_unwind
             at /rustc/d9a105fdd46c926ae606777a46dd90e5b838f92f/library/std/src/panicking.rs:493:5
   5: std::panicking::begin_panic_fmt
             at /rustc/d9a105fdd46c926ae606777a46dd90e5b838f92f/library/std/src/panicking.rs:435:5
   6: <swc_ecma_transforms::modules::import_analysis::ImportAnalyzer as swc_ecma_visit::Fold>::fold_module
   7: swc_ecma_visit::fold_program
   8: swc::Compiler::process_js_file
   9: napi::async_work::execute
  10: worker
             at /home/iojs/build/ws/out/../deps/uv/src/threadpool.c:122:5
  11: start_thread
  12: __clone
```

### LINKED ISSUE #1314: transform template literal error
## dependency
1. @swc/cli@0.1.27
2. @swc/core@1.2.43
## code
```js
const a = {
  method() {
    const string = `
width: ${Math.abs(this.currentX - this.startX - left)}px;
height: ${Math.abs(this.currentY - this.realtimeStartY - top)}px;
top: ${Math.min(this.currentY - top, this.realtimeStartY) + this.scrollTop}px;
left: ${Math.min(this.currentX - left, this.startX) + this.scrollLeft}px
`;

  }
}
```
## .swcrc
```json
{
  "jsc": {
    "target": "es2015",
    "parser": {
      "syntax": "ecmascript",
      "privateMethod": false,
      "functionBind": false,
      "exportDefaultFrom": false,
      "exportNamespaceFrom": false,
      "decorators": false,
      "topLevelAwait": false,
      "importMeta": false
    }
  }
}
```
## output 
```js
var a = {
    method: function() {
\n");t: ").concat(Math.min(this.currentX - left, this.startX) + this.scrollLeft, "px, "px;
    }
};

```

## expected
```js
var a = {
  method: function method() {
    var string = "\nwidth: ".concat(Math.abs(this.currentX - this.startX - left), "px;\nheight: ").concat(Math.abs(this.currentY - this.realtimeStartY - top), "px;\ntop: ").concat(Math.min(this.currentY - top, this.realtimeStartY) + this.scrollTop, "px;\nleft: ").concat(Math.min(this.currentX - left, this.startX) + this.scrollLeft, "px\n");
  }
};
```

## babel playground
https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAhjAvDA3gKBjAtgUygCxABMAKASlQ0xlElmgCcBLMAcyRgAM0B3JoggC4YAEhQBZOAQB0cAEYQSBJhGnAArgwY4wUABowAtDGWrocBvqMwANjgBmUMgF8ADgA8A3Gnw4mrfChhMUkZeUVTNU1tXQBNa0jtOBsoJlwAZSgLKHjjKBBXFw9vfNdgiSl8aSwWJXwVKK0dHISCgBoTetUklLScTOzYigBqToaIYAYQGxsAFQK3LzQ7R3LQqpqwOoaNJt0DYxWoDsjzSz0RsbNJ6ZsAGQcoRbRObypnNGcgA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Ces2015&prettier=false&targets=&version=7.12.12&externalPlugins=

### LINKED ISSUE #1326: transform anuglar compilerjs faild
```shell
ERROR in ../node_modules/@angular/compiler/fesm2015/compiler.js 9871:19
Module parse failed: Unterminated string constant (9871:19)
File was processed with these loaders:
 * ../builder-angular/node_modules/happypack/loader.js
You may need an additional loader to handle the result of these loaders.
|             return '\\r';
|         } else {
>             return "\\\".concat(match[0]);
|         }
|     });

webpack compiled with 1 error and 31 warnings
```

seems a inner error,these code not found in compiler.js

kdy1:
# Task

swc_ecma_codegen:
 - [x] Handle unicode escape sequences when target is es5. (Closes #1227, Closes #1326)

swc_ecma_transforms_compat:
 - [x] Handle template literals correctly. (Closes #1314)
 - [x] Handle private class properties correctly. (Closes #1306)

swc:
 - [x] Don't panic on `export * as foo from 'foo'`. (Closes #1307)


patrick91:
Hi @kdy1 does this also fix this case?

```
export default from './CampaignPackagePage';
```

I'm getting this error when using the parser directly:

```
error: Expected ;, got string literal (./CampaignPackagePage)
 --> something.../index.js:1:21
  |
1 | export default from './CampaignPackagePage';
```

my config is:

```
Syntax::Es(EsConfig {
    jsx: true,
    num_sep: true,
    class_private_props: true,
    class_private_methods: true,
    class_props: true,
    fn_bind: false,
    decorators: true,
    decorators_before_export: true,
    export_default_from: true,
    export_namespace_from: true,
    dynamic_import: true,
    nullish_coalescing: true,
    optional_chaining: true,
    import_meta: true,
    top_level_await: true,
    import_assertions: true,
})
```

kdy1:
@patrick91 I'm not sure.
If you got the error using the latest version, it's not.

patrick91:
yes, I used the version from git :)

do you want me to create a small reproduction case?

kdy1:
It's not required. I remember the parsing rule and I expect `from` of `export default from` parsed as an expression to create a default export and the parser seems to want a semicolon to end the statement.

Btw, are you are using swc at the work? I'm asking this because I want to fill the list at https://swc.rs/users/ :)

kdy1:
If not, it's okay :)

patrick91:
> It's not required. I remember the parsing rule and I expect `from` of `export default from` parsed as an expression to create a default export and the parser seems to want a semicolon to end the statement.
Ah nice!

> Btw, are you are using swc at the work? I'm asking this because I want to fill the list at [swc.rs/users](https://swc.rs/users/) :)
I'm using the rust libraries to extract GraphQL queries from our code to do some static analysis. Might make the code opensource at some point 😊

kdy1:
@patrick91 May I add your company to the list :) ?

kdy1:
(In the case you are using swc's parser. Also deno uses swc from rust side and it's on the list)

patrick91:
> @patrick91 May I add your company to the list :) ?

I'll ask, just in case 😊

kdy1:
Oh, I found that this syntax is invalid.

patrick91:
Oh, I think it comes from this proposal -> https://github.com/tc39/proposal-export-default-from

which we use, but it's ok to ignore, for my use case I can ignore the errors 😊

Also just got permission to add my company (team.pollen.co) to the list, shall I make a PR?

kdy1:
I'll add to the list, thanks! And if you are using ecmascript (not typescript), you can use `exportDefaultFrom ` option. See: https://swc.rs/docs/configuring-swc/#jsc

patrick91:
> I'll add to the list, thanks! And if you are using ecmascript (not typescript), you can use `exportDefaultFrom ` option. See: [swc.rs/docs/configuring-swc/#jsc](https://swc.rs/docs/configuring-swc/#jsc)

Oh, nice! It works now, I had an issue with the configuration (it was using TS instead of ES)

PR #1301, opened 2020-12-30
fix(ecmascript/transforms): Fix dce

kdy1:
#  Issues

 - [x] denoland/deno#8952
 - [x] denoland/deno#8985
 - [x] denoland/deno#8988

---

# Tasks

swc_bunder:
 - [x] Fix `keywords` pass.

swc_ecma_codegen:
 - [x] Ensure that the code generator handles unicode characters properly. (denoland/deno#8925)

swc_ecma_parser:
 - [x]  Ensure that the parser handles unicode characters properly. (denoland/deno#8925)

swc_ecma_transforms:
 - [x] Fix dce.

---

I'll just use ast to perform dce. See #1310 for optimizer.


ecmascript/transforms/src/optimization/simplify/dce/mod.rs
```diff
@@ -804,7 +985,7 @@ impl Dce<'_> {
     //            dropped: false,
     //        };
     //
-    //        let ret = op(&mut child);
+    //        let ret = op(&mut  child);
```
soilSpoon, line 988:
```suggestion
    //        let ret = op(&mut child);
```
why use double space?

soilSpoon: 963line and 975line too. 😆

PR #1302, opened 2020-12-30
feat(es/minifier): Implement minifier

### LINKED ISSUE #704: minify + compression improvement ?
I understand the idea that gzip should do the heavy lifting of compressing js files.
However the end result is still kind of "too heavy":
- concatenated bundle: > 500ko
- uglified and gzipped: 73ko
- uglified and brotli: 61ko
- swc+minify and gzipped: 130ko
- swc+minify and brotli: 100ko

I suppose it's not a high priority to improve on this, hence the request ;)

### LINKED ISSUE #1162: Implement JavaScript optimizer using ecmascript lexer/parser of swc
**Question**
I can see that swc have a very impressive performance.
There is any plan to implement a new js optimizer to replace exist js optimizer like terser, uglifyjs?

kdy1:
# Postponed

To simplify management, I'll merge this PR as incomplete state and implement more rules with follow-up PRs.

swc_bundler:
 - [ ] Add option for minification.

swc_ecma_minifier:
 - [ ] Implement.
 - [ ] Provide js api.


soilSpoon:
Are you thinking of implementing it using UglifyJS?

kdy1:
@soilSpoon Yes, but mainly because I don't know well about internals of minifiers.
Is there something better than uglifyjs?

Or I can add some passes after implementing it using uglifyjs.

soilSpoon:
[uglifyjs](https://github.com/mishoo/UglifyJS) doesn't seem to be able to minify ECMAScript 2015 or above. So, to use this, you need to use it after transpile it to ECMAScript 2015 or lower. I think this can adversely affect performance. So I look better to use [terser](https://terser.org/) that can minify ECMAScript 2015 or above.

kdy1:
@soilSpoon Oh. I got it. Thanks!
I'll port terser instead.

divmain:
You may also want to look at [babel-minify](https://github.com/babel/minify) for inspiration.  From what I recall, it doesn't achieve quite the same size wins as Uglify or Terser.  But the port may be significantly more straightforward - the minifier is split into [discrete operations](https://github.com/babel/minify/tree/master/packages) for the different types of transformations that it applies.

ecmascript/minifier/src/compress/optimize/dead_code.rs
```diff
@@ -0,0 +1,75 @@
...
+    ///
+    /// can be optimized as
+    ///
+    /// ```ts
+    /// function foo(){
+    ///     var a; // Will be dropped in next pass.
+    ///     throw x();
+    /// }
+    /// ```
+    pub(super) fn optimize_in_fn_termiation(&mut self, e: &mut Expr) {
```
pickfire, line 32: Typo
```suggestion
    pub(super) fn optimize_in_fn_termination(&mut self, e: &mut Expr) {
```

kdy1: Oh, thanks!

fabiosantoscode:
Hey, I maintain Terser, and it's helplessly slow. Recently I started a port to rust (in a very early stage) but I don't think it's useful to keep two separate projects that are essentially the same. How can I help?

kdy1:
@fabiosantoscode I implemented many rules but there are many rules left.
If you don't mind implementing it as a part of swc, I will add lots of comment how the minifier works and about some design decisions.

Note that being in-tree would make version management much simpler, as bundler depends on minifier and minifier depends on ast / visitor while those crates live in a momorepo.

fabiosantoscode:
Ok, I'll go on your fork and send PRs when I have the time.

I'm interested in how to implement `reduce_vars`, `unused`, and `inline`, because in Terser they're kind of the same for speed reasons. They're also very relevant for tree shaking.

Austaras:
Any possible to expose as webpack plugin like [terser](https://github.com/webpack-contrib/terser-webpack-plugin)

kdy1:
@Austaras Of course. I will do so.

kdy1:
@fabiosantoscode

How do you think about merging this into master and improving it gradually, with multiple PRs?

fabiosantoscode:
Sure! Especially if it's documented as "beta"

fabiosantoscode:
Eventually I'd like to test this against every single Terser test. Not all of them require exact output. Some of them run the resulting code using node's VM module and compare the result.

kdy1:
> Eventually I'd like to test this against every single Terser test. Not all of them require exact output. Some of them run the resulting code using node's VM module and compare the result.

👍  That's also my goal.
Terser has strength in the output and I want to make it faster without degrading output.

SoloJiang:
However, I see this PR had beed merged, but how can I try it out? At now, it only removes comment and space in my code.

kdy1:
@SoloJiang It's not published yet because it's not done

SoloJiang:
Okay, I'm working on use swc instead of Babel and Terser in our project github.com/alibaba/ice. In our situation, bundle size is more important than bundler speed. Without minify, swc result is bigger 10% than Babel. In minify, it's 200% than Terser or esbuild which optimized by comfused code.

fabiosantoscode:
So much excitement for this port, I love it 💖

I need to try to reduce my hours in my side gig so I can give this more attention.

There's some transforms Terser itself doesn't do that are feasible here (such as a dictionary for common immutable constants or cross-module intelligence used in combination with spack) that can reduce this port's output bytes to a point that makes it a viable alternative to Terser.

kdy1:
I'm glad to hear that and thank you!


> There's some transforms Terser itself doesn't do that are feasible here (such as a dictionary for common immutable constants or cross-module intelligence used in combination with spack) that can reduce this port's output bytes to a point that makes it a viable alternative to Terser.

Sounds interesting. I expect such cross-module transforms to reduce final output by a margin because libraries used a lot along with UI frameworks like react encourages such code.

PR #1305, opened 2020-12-30
fix(ci): insert *_bg files in @swc/wasm-web

punkeel:
Write the modified JSON blob to disk, so it's applied. It'll work … at some point

(cc @9oelM)


9oelM:
oops, sorry. Thanks for fixing it.

PR #1376, opened 2021-02-03
Add es2020 to JscTarget in types.ts

wycats:
This target is supported in the Rust code, but the TS types haven't been updated yet.


kdy1:
Thanks!

PR #1390, opened 2021-02-11
avoid removing plugin

JoviDeCroock:
When trying out the rollup swc-plugin I got confused why my `options.plugin` would dissapear during concurrent runs


kdy1:
Thanks!

PR #1396, opened 2021-02-13
Ability to look at the kind of an error

dsherret:
Right now there doesn't seem to be a way to take a look at the error kind without converting the `Error` into a `SyntaxError`.

Thoughts on this? (I did not run tests because so this may fail on the CI)


kdy1:
It seems like a good idea to me.

dsherret:
This said, I just realized I think I'm able to do everything I need to do using only an `error::Error` and not bothering to deal with `DiagnosticBuilder`. This might be a good change anyway.

PR #1401, opened 2021-02-15
Default Path as root for resolver without allocate

pickfire:
Happy chinese new year!


kdy1:
Thanks! But I'm a Korean :) and we have our own holiday called seollal.

I'll merge this if ci passes.

pickfire:
> Thanks! But I'm a Korean :) and we have our own holiday called seollal.

Wow, it's the same day. :D

PR #1416, opened 2021-02-19
feat(node-swc): Babel ast translator

### LINKED ISSUE #1392: Babel compatible AST
Heya,

thanks for this amazingly fast parser! I've been playing with replacing it with Babel's parser, but I also used a number of babel libraries which expect a Babel AST. Namely `span` to be in-lined (collapsing `LogicalExpression` into `BinaryExpression` does not seem to be causing problems). Now I was wondering whether the divergence from Babel's AST is wanted? I can imagine integration into all kinds of related libs to be simpler when conforming to their AST more. Or is there a perf cost associated with these differences?
Would be curious to learn more and I did not find a past issues that made me understand.

Cheers!

kdy1:
# Tasks

# Issues

Closes #1392.


dwoznicki:
I can probably do this, btw.

kdy1:
@dwoznicki If then, it would be really helpful 👍

dwoznicki:
I'll open a PR once I've made a bit more progress.

kdy1:
Great! Thanks

kdy1:
Closing in favor of https://github.com/swc-project/swc/pull/1465

PR #1418, opened 2021-02-20
Reduce Node.js addon binary size

Brooooooklyn:


Cargo.toml
```diff
@@ -1,3 +1,5 @@
+cargo-features = ["strip"]
```
kdy1, line 1: Wow. I didn't know this. Thanks!

kdy1 (approved):
LGTM. Thank you!

Cargo.toml
Brooooooklyn: It's a nightly feature in Cargo: https://github.com/rust-lang/cargo/pull/8246

PR #1419, opened 2021-02-20
try to fix windows tests

### LINKED ISSUE #1417: tests not passing on windows
I didn't see anything mentioned about Windows development in the [contritbuting.md](https://github.com/swc-project/swc/blob/master/CONTRIBUTING.md) file so I assumed it was ok to try

First of all the fixture macro is broken on Windows:
https://github.com/swc-project/swc/blob/eecdca4e866f470781806ed287fd8480aefa8dce/testing/macros/src/fixture.rs#L136

so I added a check for `\\`, but then I still get errors while running just ecmascript/parser: https://gist.github.com/JakubKoralewski/5c66c60998e630663641e2d03446a3c8

`cargo test --all --all-features`: 

1. First run
   https://gist.github.com/JakubKoralewski/8a2c2158a69d3c392348379cd231c4ab
2. Second run
   https://gist.github.com/JakubKoralewski/ad0794768000e010d909789343955cc0
I think they're the same but I was hoping the allocation error would be accidental.

So I just checked on WSL and I'm getting same sort of output too but like 100x slower IO I think

JakubKoralewski:
Fixes https://github.com/swc-project/swc/issues/1417

1. *.debug files were somehow not converted by git to LF file endings so I added those to .gitattributes
   I had to remove all files and git reset for the changes to show
2. spack had very linux specific path handling which I attempted to fix but not all tests are passing:
   https://gist.github.com/JakubKoralewski/f6368a554073f14240184b40fefde888

I'd appreciate some help, as I don't know what to do with the above test failures (fixture::pass::issue_1139\example_1 doesnt seem to be path related, but fixture::pass::pr_1105\example_7 seems to be :confused:)


kdy1:
Can you please change this into a draft PR so I can know when the task is done?

JakubKoralewski:
> Can you please change this into a draft PR so I can know when the task is done?

Sure.

Could you take a look at the `fixture::pass::issue_1139\example_1` test output? I can try to debug `fixture::pass::pr_1105\example_7` as it seems to be something wrong with windows paths, but i have no idea why the former would fail, seems to be unrelated to this PR. Could you confirm?

kdy1:
@JakubKoralewski I think it's a bug of swc and it will be fixed by https://github.com/swc-project/swc/pull/1382

JakubKoralewski:
On further inspection it looks like in the `fixture::pass::pr_1105\example_7` test the created windows path:
```
<    url: "$DIR/tests/pass/pr-1105/example-7/input/f.js",
>    url: "\\\\?\\D:\\projects\\swc\\swc\\spack\\tests\\pass\\pr-1105\\example-7\\input\\f.js",
```

is actually correct, it's just that the way the tests compare the outputs does not take into account Windows paths. I'm not sure if the [verbatim `\\?\` syntax](https://doc.rust-lang.org/std/path/enum.Prefix.html) will be interpreted correctly though.

@kdy1 What do you think? My idea is that it's an issue with tests not with the current implementation, so I believe it is ready for your review. Considering there is no CI for Windows I don't think modifying the tests logic for a single incorrect test is worth it, I'd add it as a bug and move on.

kdy1:
> My idea is that it's an issue with tests not with the current implementation, so I believe it is ready for your review

I agree. I'll take a look after taking some rest.

spack/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![cfg_attr(test, feature(test))]
+#![feature(or_patterns)]
```
kdy1, line 2: Is this really required?

kdy1: swc uses nightly for a while because of some rustfmt options, but it does not rely on nightly features from non-test codes.

kdy1 (commented):
Mostly LGTM.
I'm a bit worried about the nightly feature, though.

spack/src/lib.rs
JakubKoralewski: not required, I just noticed that nightly was used anyway so I wanted to use some terser syntax, i'm not familiar with the consequences of using nightly features. looking at rust-lang/rust/issues/54883 looks to be out in next stable

should I remove it?

kdy1: It's okay if it's going to be stabilized. Thanks!

PR #1426, opened 2021-02-22
Support using swc_common with MIRI

playXE:
swc_common uses `is_x86_feature_detected!("sse2")` inside which makes it impossible to use swc_common with MIRI together. This is needed for example in [starlight](https://github.com/Starlight-JS/starlight) to verify JS interpreter,GC and other stuff.


kdy1:
I published a new version.

PR #1439, opened 2021-02-28
Port @babel/preset-modules to SWC

devongovett:
This is a port of [@babel/preset-modules](https://github.com/babel/preset-modules) to SWC. As with the original, this produces smaller output when targeting browsers supporting esmodules that just works around browser bugs as opposed to compiling to ES5. This was later [integrated](https://babeljs.io/blog/2020/03/16/7.9.0#babelpreset-envs-bugfixes-option-11083httpsgithubcombabelbabelpull11083) into `@babel/preset-env`, and I think that would be a good idea for SWC's preset-env as well.

This is just an initial PR to get feedback on the approach and whether this would even be accepted into SWC. Also left some questions below.

- [x] **transform-async-arrows-in-class**
- [x] **transform-edge-default-parameters**
- [ ] **~~transform-edge-function-name~~** - I think this is unnecessary since SWC already does a lot less than Babel's function-name plugin so it's already basically equivalent to the bugfixes option.
- [x] **transform-jsx-spread** - Implemented the `use_spread` option of the existing JSX pass rather than building a separate pass. 
- [ ] **~~transform-safari-block-shadowing~~** - not needed, SWC always renames shadowed variable declarations, even with let
- [ ] **~~transform-safari-for-shadowing~~** - same as above
- [x] **transform-tagged-template-caching**

cc. @developit

### DIFF
--- a/ecmascript/preset_env/src/transform_data.rs
+++ b/ecmascript/preset_env/src/transform_data.rs
@@ -169,16 +169,12 @@ pub(crate) static BUGFIX_FEATURES: Lazy<HashMap<Feature, BrowserData<Option<Vers
         FEATURES
             .clone()
             .into_iter()
-            .chain(
-                map.into_iter()
-                    .map(|(feature, version)| {
-                        (
-                            feature,
-                            version.map_value(|version| version.map(|v| v.parse().unwrap())),
-                        )
-                    })
-                    .collect::<HashMap<Feature, BrowserData<Option<Version>>>>(),
-            )
+            .chain(map.into_iter().map(|(feature, version)| {
+                (
+                    feature,
+                    version.map_value(|version| version.map(|v| v.parse().unwrap())),
+                )
+            }))
             .collect()
     });
 
--- a/ecmascript/preset_env/src/transform_data.rs
+++ b/ecmascript/preset_env/src/transform_data.rs
@@ -4,8 +4,15 @@ use std::collections::HashMap;
 use string_enum::StringEnum;
 
 impl Feature {
-    pub fn should_enable(self, target: Versions, default: bool) -> bool {
-        let f = &FEATURES[&self];
+    pub fn should_enable(self, target: Versions, bugfixes: bool, default: bool) -> bool {
+        let f = if bugfixes {
+            &BUGFIX_FEATURES[&self]
+        } else {
+            if !FEATURES.contains_key(&self) {
+                return false;
+            }
+            &FEATURES[&self]
+        };
 
         should_enable(target, *f, default)
     }
@@ -126,6 +133,15 @@ pub enum Feature {
 
     /// `transform-unicode-escapes`
     UnicodeEscapes,
+
+    /// `bugfix/transform-async-arrows-in-class`
+    BugfixAsyncArrowsInClass,
+
+    /// `bugfix/transform-edge-default-parameters`
+    BugfixEdgeDefaultParam,
+
+    /// `bugfix/transform-tagged-template-caching`
+    BugfixTaggedTemplateCaching,
 }
 
 pub(crate) static FEATURES: Lazy<HashMap<Feature, BrowserData<Option<Version>>>> =
@@ -144,6 +160,28 @@ pub(crate) static FEATURES: Lazy<HashMap<Feature, BrowserData<Option<Version>>>>
             .collect()
     });
 
+pub(crate) static BUGFIX_FEATURES: Lazy<HashMap<Feature, BrowserData<Option<Version>>>> =
+    Lazy::new(|| {
+        let map: HashMap<Feature, BrowserData<Option<String>>> =
+            serde_json::from_str(include_str!("transform_data_bugfixes.json"))
+                .expect("failed to parse json");
+
+        FEATURES
+            .clone()
+            .into_iter()
+            .chain(
+                map.into_iter()
+                    .map(|(feature, version)| {
+                        (
+                            feature,
+                            version.map_value(|version| version.map(|v| v.parse().unwrap())),
+                        )
+                    })
+                    .collect::<HashMap<Feature, BrowserData<Option<Version>>>>(),
+            )
+            .collect()
+    });
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -155,6 +193,7 @@ mod tests {
                 ie: Some("11.0.0".parse().unwrap()),
                 ..Default::default()
             },
+            false,
             false
         ));
     }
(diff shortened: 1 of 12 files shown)


kdy1:
I think it's a good change. Thank you for sending the PR and I will accept it.
I'll take a look at the changes.

ecmascript/transforms/compat/src/bugfixes/edge_default_param.rs
```diff
@@ -0,0 +1,102 @@
...
+
+impl Fold for EdgeDefaultParam {
+    noop_fold_type!();
+
+    fn fold_arrow_expr(&mut self, n: ArrowExpr) -> ArrowExpr {
+        self.in_arrow = true;
+        let params = n.params.fold_with(self);
+        self.in_arrow = false;
+
+        let body = match n.body {
```
kdy1, line 24: I think this can be simplified to `n.body.fold_with(self)` or `n.body.fold_children_with(self)`.
Later would have exactly same effect.

kdy1 (commented):
LGTM. Thanks!

src/config.rs
```diff
@@ -824,7 +824,9 @@ impl Merge for GlobalPassOption {
 
 impl Merge for react::Options {
     fn merge(&mut self, from: &Self) {
-        *self = from.clone();
+        if *from != react::Options::default() {
```
devongovett, line 827: @kdy1 I'm not really sure about this, but I noticed when testing via the CLI that the React options in my `.swcrc` file were not being used at all. This appears to be because `from` here are the default options, which overwrite the options from the config. Not sure if the merging order should be different, or perhaps the API should be passed an optional in case no options are given rather than using the defaults. Let me know what you want to do.

devongovett:
I think this is ready for review! Added a `bugfixes` option to the `env` config, which applies the bugfix passes and overrides the versions for other passes. Data copied from Babel. Tested manually via the CLI as well and it appears to work.

Also, while `@babel/preset-modules` has a JSX spread transformer, `@babel/preset-react` instead has a `useSpread` option. This makes sense because JSX transformation is not included in preset-env, so I've replicated that here as well. It does mean that it isn't turned on automatically based on env versions though. Not sure if we want to do that somehow. Also, not quite sure why it's in preset-modules to begin with since object spread isn't even supported until Edge 79 I think... @developit do you remember?

src/config.rs
kdy1: >  I noticed when testing via the CLI that the React options in my .swcrc file were not being used at all

This seems like a bug. Maybe I need to add some more tests for `.swcrc`.
I think it should respect `.swcrc`, while preferring the option passed via javascript api.

ecmascript/transforms/compat/src/bugfixes/template_literal_caching.rs
```diff
@@ -0,0 +1,308 @@
...
+                // Use the identity function helper to get a reference to the template's
+                // Strings. We replace all expressions with `0` ensure Strings has
+                // the same shape.   identity`a${0}`
+                let template = TaggedTpl {
+                    span: DUMMY_SP,
+                    tag: Box::new(Expr::Ident(helper_ident.clone())),
+                    quasis: n.quasis,
+                    exprs: n
+                        .exprs
+                        .clone()
```
kdy1, line 96: It seems like the value is ignored on next call to `.map`.
If then, we don't need to clone the vector I guess.

ecmascript/transforms/compat/src/bugfixes/template_literal_caching.rs
```diff
@@ -0,0 +1,308 @@
...
+                Expr::Call(CallExpr {
+                    span: DUMMY_SP,
+                    callee: ExprOrSuper::Expr(n.tag),
+                    args: vec![ExprOrSpread {
+                        expr: Box::new(inline_cache),
+                        spread: None,
+                    }]
+                    .into_iter()
+                    .chain(n.exprs.into_iter().map(|expr| ExprOrSpread {
+                        expr: expr.clone(),
```
kdy1, line 137: `.clone()` seems not necessary.

ecmascript/transforms/compat/src/bugfixes/template_literal_caching.rs
```diff
@@ -0,0 +1,308 @@
...
+                })
+            }
+            _ => n,
+        }
+    }
+
+    fn fold_module(&mut self, n: Module) -> Module {
+        let mut body = n.body.fold_children_with(self);
+        if let Some(var) = self.create_var_decl() {
+            prepend_stmts(&mut body, vec![ModuleItem::Stmt(var)].drain(..))
```
kdy1, line 151: There's a function named `prepend_stmt` which only adds one statement.

ecmascript/transforms/compat/src/bugfixes/template_literal_caching.rs
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        Module { body, ..n }
+    }
+
+    fn fold_script(&mut self, n: Script) -> Script {
+        let mut body = n.body.fold_children_with(self);
+        if let Some(var) = self.create_var_decl() {
+            prepend_stmts(&mut body, vec![var].drain(..))
```
kdy1, line 160: Ditto.

ecmascript/preset_env/src/transform_data.rs
```diff
@@ -144,6 +160,28 @@ pub(crate) static FEATURES: Lazy<HashMap<Feature, BrowserData<Option<Version>>>>
...
+            .into_iter()
+            .chain(
+                map.into_iter()
+                    .map(|(feature, version)| {
+                        (
+                            feature,
+                            version.map_value(|version| version.map(|v| v.parse().unwrap())),
+                        )
+                    })
+                    .collect::<HashMap<Feature, BrowserData<Option<Version>>>>(),
```
kdy1, line 180: Is this `.collect()` required?

kdy1 (changes requested):
Mostly looks good, only small changes/

src/config.rs
devongovett: I didn't look too far, but it could be here where the defaults are being used in case no options are passed via the API: https://github.com/swc-project/swc/blob/a3105428ba2ff9abd7d0e5464868da3ca9b66294/native/src/transform.rs#L71

kdy1: Yeah, you are right.
But I'm not sure about the fix at the moment.
If you want, you can just leave it and I'll fix it with another pr.

devongovett: Sounds good

kdy1 (approved):
LGTM. Thank you a lot.

PR #1453, opened 2021-03-05
fix(swc): Fix bugs

### LINKED ISSUE #1472: strip: Define class fields in constructor by default
Example input:
```ts
class A extends Object {
  a = 1;
  constructor(public b = 2) {
    super();
  }
}
```
TSC actually only preserves class fields under specific conditions:
1) `"target": "ESNext", "useDefineForClassFields": true` - Preserve class fields. Current default behaviour (roughly) of swc.

  ```js
  class A extends Object {
      b;
      a = 1;
      constructor(b = 2) {
          super();
          this.b = b;
      }
  }
  ```
2) `"target": "ESNext", "useDefineForClassFields": false` - Define class fields in constructor. Current behaviour of swc with `typescript_class_properties()` pass. Proposed default.

  ```js
  class A extends Object {
      constructor(b = 2) {
          super();
          this.b = b;
          this.a = 1;
      }
  }
  ```
3) `"target": "ES6-ES2020", "useDefineForClassFields": true` - Define class fields in constructor using `Object.defineProperty()`.

  ```js
  class A extends Object {
      constructor(b = 2) {
          super();
          Object.defineProperty(this, "b", {
              enumerable: true,
              configurable: true,
              writable: true,
              value: b
          });
          Object.defineProperty(this, "a", {
              enumerable: true,
              configurable: true,
              writable: true,
              value: 1
          });
      }
  }
  ```
4) `"target": "ES6-ES2020", "useDefineForClassFields": false` - Same as (2).

So swc should have a similar API. I propose adding `strip::Config::use_define_for_class_fields: bool`, defaulting to `false`. If set to `true`, class fields preserved. Otherwise they are put into the constructor. The `typescript_class_properties()` pass should instead transform class fields to be constructor-defined using `Object.defineProperty()` -- we suggest using it when targeting ES2020 and earlier with `"useDefineForClassFields": true`.

Ref #949.

kdy1:
# Tasks

swc_ecma_transforms_typescript:
  - [x] Add option to use define propert for `strip` pass. (Closes #1472)

node-swc:
  - [x] Fix `keepClassNames`. (Brooooooklyn/swc-node#335)


ecmascript/transforms/typescript/tests/strip.rs
```diff
@@ -3488,3 +3504,58 @@ to!(
...
+    "
+    class A extends Object {
+        a = 1;
+        constructor(b = 2){
+            super();
+            this.b = b;
+        }
+    }
+    "
+);
```
nayeemrmn, line 3561: This is incorrect. See the examples in #1472. `use_define_for_class_fields: true` should give:
```js
class A extends Object {
    b;
    a = 1;
    constructor(b = 2){
        super();
        this.b = b;
    }
}
```
And `use_define_for_class_fields: false` should give:
```js
class A extends Object {
    constructor(b = 2){
        super();
        this.b = b;
        this.a = 1;
    }
}
```
`Object.defineProperty()` is for ES2020 and earlier.

kdy1: It's intentional. It's because `strip` does not care about the target es version.

The caller can use `use_define_for_class_fields: true` to get

```ts
class A extends Object {
    constructor(b = 2) {
        super();
        this.b = b;
        Object.defineProperty(this, "a", {
            enumerable: true,
            configurable: true,
            writable: true,
            value: 1
        });
    }
}
```

where `this.b = b` can be transformed using another pass.


And for, `use_define_for_class_fields: false `, caller should use `class_properties` pass to convert class properties to `this.foo = value`.


But well, I guess I should add some integration tests.

kdy1: Oh, I found that this pass eats all properties, and mixing passes does not work. Thanks for pointing it out!

PR #1463, opened 2021-03-11
Fix codegen for JSX spread props

alvra:
Codegen for JSX spread props currently produces invalid code.

```
<a ...x />
```

This PR fixes that so it outputs:

```
<a {...x} />
```


kdy1:
Great! Thanks!

PR #1465, opened 2021-03-11
feat(node-swc): Babel ast translator

dwoznicki:
This pull request is a WIP.

- [x] Translate babel AST definitions to rust
- [x] Implement convert functions for swc --> babel
- [x] Write tests

Let me know if there's anything missing from the todo list.


kdy1:
Wow, thank you for such enormous work!

kdy1:
Will you implement the conversion from swc ast to babel ast?
If not, I think this can be merged :)

dwoznicki:
Sorry, yes, I am working on the conversions. It's slow going for.. well, a variety of reasons. Perhaps it would be helpful if you gave me a generic example of how you expect the `Babelify` trait to be implemented? Seeing how you defined the `BaseNode` struct was really helpful when writing the Babel AST definitions.

kdy1:
@dwoznicki Good to hear. I'll add some examples for it tomorrow. (It's time to go sleep in my country)

kdy1:
I did some work. Please feel free to ping me if you need ant help.

See: https://github.com/kdy1/swc/tree/d872d001bf387dd7ebe1c27ce87ab438ab42c556/native/babel-compat/src

dwoznicki:
Cool, thanks. I'll let you know if I have any trouble.

dwoznicki:
@kdy1 Hey, a couple questions for you.

1. How can I get the `Span` for nodes that use the `#[span]` macro? E.g. `pat::KeyValuePatProp`.
2. When I have more questions, where should I ask them? Here? Discussions?

kdy1:
@dwoznicki 

1. You can use `.span()` from the trait `swc_common::Spanned` instead.
2. Anywhere is ok for me.

dwoznicki:
Cool, thanks. Okay, another question.

I've found several places where the swc node does not convert cleanly into a babel node, and I need to throw/return an error if the tree is invalid. How should I handle this? Should I modify the `babelify()` function to return a `Result`?

kdy1:
@dwoznicki I think so. Panic would also work, but it doesn't look good.

dwoznicki:
@kdy1 Hey, quick question. The swc `Module` and `Script` shebang field is a JsWord, but Babel's corresponding interpreter field is an `InterpreterDirective` which expects a base node. Is there any way to get a span from a JsWord?

kdy1:
@dwoznicki Currently there's no way to do it.
I think the only way to get span without changing ast would be using span (from `Module` / `Script`) with https://docs.rs/swc_common/0.10.13/swc_common/struct.SourceMap.html#method.span_take_while.

Taking lower `BytePos` from span, creating a new span from lower, and extending it until newline characters would work.

dwoznicki:
Great, yeah, that appears to work. Thanks.

kdy1:
I'll extract ast definitions from the pr so that I can use them to implement the babel plugin runner.

dwoznicki:
Oh, I've fixed a couple things since. Let me make a push right quick. I suspect there will be more problems with my definitions that'll need fixing.

kdy1:
@dwoznicki Oh, thanks!

dwoznicki:
It's up. Are you implementing a babel plugin runner in Rust?

kdy1:
@devongovett I'd like to, but I need to do some experiments to see if it's possible.

kdy1:
And thanks!

dwoznicki:
@kdy1 If you manage to get it working, can you let me know if you solve this issue? https://github.com/napi-rs/napi-rs/issues/436#issue-792708371

I got stuck here while trying to implement Webpack plugins.

kdy1:
Ok, I'll try workarouding it and I'll cc you when I make a success.

dwoznicki:
First pass at conversion is done. It's pretty rough, and completely untested. I'll fix it up as I write tests.

kdy1:
Sounds great!
But I'm also blocked by https://github.com/napi-rs/napi-rs/issues/436

Brooooooklyn:
I'll take a look tonight

dwoznicki:
@kdy1 Hey, I want to get your thoughts on a couple things.

1. How rigid should the swc --> babel AST conversion be? That is, do we want to be able to `babelify` an swc AST and have it produce the exact same output that we would get if we parsed it with babel directly? I'm leaning towards an exact AST translation, btw, but just want to check.
2. On a related note, there appear to be some places where the babel node definitions from [here](https://raw.githubusercontent.com/babel/babel/main/packages/babel-types/src/ast-types/generated/index.ts) don't match the actual output of `@babel/parse`.  For example, the parsed babel AST has an `"errors"` field in the `File` node, but the `File` definition does not mention errors. Should I modify our Rust babel node definition to match the actual output?

kdy1:
@dwoznicki

For 1, I think the safety line would be babel-plugin compatibility. I expect the plugin system to work even though some ast differs (e.g. `Pat::Expr` in `PatOrExpr::Pat`), and if not, it can be fixed in the future.

For 2, I don't think so. Almost, if not all, babel plugins would not depend on some undocumented fields. And same as 1, we can fit it in the future if it causes a problem.

dwoznicki:
Cool thanks. Okay, another more specific question.

Babel wraps the top level `Program` in a `File` node, which swc of course doesn't bother with. So when I `babelify` an swc `Program`, if it starts at bytepos 0, can I assume that this is the top level `Program`?

kdy1:
Yes, but not vice versa.

The `swc_common::SourceMap` is shared and as a result, secondly parsed ast would have non-zero starting byte position.

dwoznicki:
Hmm, I see. Yeah, that sounds like it could pose a problem when parsing multiple files. Is there a better way to figure out of the Program is top level in a file?

kdy1:
> So maybe I can just always assume that programs are top level in files?

Yes, I think it's the case.
Swc only creates `Program` only once for each file.

kdy1:
Wow... Is this done?
I was surprised while reviewing the changes. It's really enormous amount of work. Thank you a ton!

dwoznicki:
It's not done yet, no. I'm still testing, and finding lots of bugs haha.

dwoznicki:
@kdy1 Hey, random question. I want to create a derive macro to automatically implement `Serialize` for enums that are standing in for TS union types. It looks like I should probably make a new crate for it.

1. Where should I put the new crate? `swc/macros/`? `swc/native/babel-compat/`?
2. When I try to add a new crate, I get error `error: current package believes it's in a workspace when it's not`. It looks like I could either remove my new crate from the root workspace altogether by adding `[workspace]` to Cargo.toml, or manually add to to the workspace members in `swc/Cargo.toml`, both of which make the error go away. But I'm unsure whether this is the right way to do it. I noticed that almost every crate within the project is neither listed in the root workspace members not has an empty `[workspace]` in its Cargo.toml. Is there a particular way I should create a new crate?

kdy1:
@dwoznicki For 1, I think `swc/native/babel-compat` would be preferable. `/macros` exists because some macros are used by multiple crates. (Some of them should be moved, but it works)

And workspace error will go away if another crate depends on it. You can add `swc_babel_compat_macros = { path = "macros" }` to `Cargo.toml` and error will go away.

dwoznicki:
@kdy1 Hey, I need some advice.

The way I'm testing is with fixtures. I take an `input.js` file, parse it using `@babel/parser`, then write the result to an `output.json` file. With those two files in place, I read and parse the JS file using swc, `babelify` the AST, then read in and deserialize the `output.json` file to create the expected babel AST (in Rust, using the nodes on this branch), and compare the two ASTs.

The problem I'm running into is that there a lots of really minor differences between the `babelify` ed tree and the babel tree. It's mostly little stuff, like one tree will have `None` and the other will have `Some(false)` or `Some([])`. But it's making it difficult to write tests. I can think of a couple ways forward, and want to get your thoughts.

1. Implement PartialEq for a bunch of these nodes. Time consuming, but fairly certain to be "correct".
2. Create a CloseEnoughPartialEq trait + derive macro that just calls `None == Some(false)` good enough for any field. I think this is fine, but there could maybe be some nodes where we don't want this behavior.
3. Instead of comparing parsed struct trees, write a class to walk and compare JSON trees, which could then implement the close enough comparison without PartialEq. Not sure what this buys, but I've considered it.
4. Forget writing tests altogether, and just do some rough eyeball checks to make sure the conversion is working more or less as expected.

kdy1:
@dwoznicki 
How about normalizing them before comparison? It will certainly take less time.
swc already does many normalizations while testing.

You can do normalization in any of js and rust.
If you want to do it in js, you can use babel's ast traverser and if you want to do it in rust, you can make a visitor traits by doing a similar thing as swc_ecma_visit (https://github.com/swc-project/swc/blob/92bbde3b5377e281525828de426e481fbc70029f/ecmascript/visit/src/lib.rs#L414-L1677).

Gregoor:
Thanks for tackling this @dwoznicki! As mentioned in #1392 I am very interested in this, if there is a piece of work left that can be cut-off without too much coordination cost, I am more than happy to support.

dwoznicki:
@kdy1 Thanks for the idea, that sounds good.

@Gregoor Thank! I'll let you know. I think once I've figured out how to do a bit of normalization, having someone else to help write tests and fix bugs would be very helpful. I suspect there are tons of edge cases that I've gotten wrong in the conversion. Also, I don't really know TypeScript, so if you do know it and would be willing to do TS testing, that would be really helpful as well.

dwoznicki:
@kdy1 Hey, I'm trying to implement visitor traits using the `define!` macro, and I'm getting some compilation errors. All the errors appear to be related to Boxes.

```
error[E0308]: mismatched types
   --> native/babel-compat/src/visit.rs:759:13
    |
759 |         pub super_class: Option<Box<Expression>>,
    |             ^^^^^^^^^^^ expected enum `ast::expr::Expression`, found struct `Box`
    |
    = note: expected enum `std::option::Option<ast::expr::Expression>`
               found enum `std::option::Option<Box<ast::expr::Expression>>`
```

The `define!` call is in `visit.rs`. Here's the commit: 2ff8c5b3c371ce1c8237679c025159dd84824d41
Any idea what I'm doing wrong?

kdy1:
Hmm... You may need to change `Expresion` to `Box<Expression>` in fields.
But I'm not sure about it.

I'll tell you once I found the workaround.

kdy1:
It will be fixed by https://github.com/swc-project/swc/pull/1574.

kdy1:
I found that #1574 is the wrong way because of conflicts of `fn visit_opt_expr(&mut self, v: Option<Box<Expr>>)` and `fn visit_opt_expr(&mut self, v: Option<Expr>)`.

I think wrapping all `Expression`s with box like `Box<Expression>` is the right way to go.

Gregoor:
So I started an attempt at pulling in [@babel/parser's testing](https://github.com/babel/babel/tree/main/packages/babel-parser). Atm I'm wading through a mountain of non-descript errors. I have lost some faith in that approach, as I am not sure if all that they test is meant to be supported over here anyway. Though I do still think it'd be cool if swc would (at some point) be up-to-date and compatible with all the things babel is testing their parser for.
So maybe I'll just abandon it and glue some of their TS fixtures together and send that as a PR to this PR.

dwoznicki:
@kdy1 Sounds good. I'll go hunt down my Box-less expressions and fix them.

@Gregoor Yeah, I'd also like to pass `@babel/parser` tests as well.. But this PR is really far from that atm. I mean, I'm running into tons of errors just trying to convert stuff like

```
const n = 1;
```

and

```
class Rectangle {
    constructor() {
        this.height = 0;
    }
}
```

I'm hoping it'll get better once I can start normalizing the trees. But you could probably start with some *extremely simple* TypeScript fixture tests and find that there's plenty to fix.

dwoznicki:
@kdy1 Okay, I Boxed all the expressions, which seems to have helped. Now I'm getting compilation errors on the `type_annotation` field in various nodes. Nearly all of the errors look like this:

```
error[E0308]: mismatched types
    --> native/babel-compat/src/visit.rs:2021:13
     |
2021 |         pub type_annotation: Option<Box<TSTypeAnnotation>>,
     |             ^^^^^^^^^^^^^^^ expected struct `ast::typescript::TSTypeAnnotation`, found struct `Box`
     |
     = note: expected mutable reference `&mut std::option::Option<ast::typescript::TSTypeAnnotation>`
                found mutable reference `&mut std::option::Option<Box<ast::typescript::TSTypeAnnotation>>`
```

I'm not sure why `define!` expects an un-Boxed version, but I can't remove the `Box` because it results in a recursive type error:

```
error[E0072]: recursive type `ast::typescript::TSTypeAnnotation` has infinite size
   --> native/babel-compat/src/ast/typescript.rs:145:1
    |
145 | pub struct TSTypeAnnotation {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^ recursive type has infinite size
...
148 |     pub type_annotation: TSType,
    |                          ------ recursive without indirection
    |
help: insert some indirection (e.g., a `Box`, `Rc`, or `&`) to make `ast::typescript::TSTypeAnnotation` representable
```

Any ideas?

kdy1:
@dwoznicki I think the error message is misleading. I looked at the visitor code, and I found that `Option<Box<TSTypeAnnotation>> ` is mixed with `Option<TSTypeAnnotation>`.

Seems like the same issue as boxing expression. So... changing `Option<TSTypeAnnotation>` to `Option<Box<TSTypeAnnotation>>` will work.

dwoznicki:
@kdy1 Hey, I have the `define!` call working. Now I just need a little help figuring out how the visitor API works. I started using https://swc.rs/blog/2020/01/04/pmutil/#fold-visit-swc as a guide, but I suspect this post is outdated.

I have my attempt here:
https://github.com/swc-project/swc/blob/044d769c6db1dc07f53fae2f7f39ec2c0f459e55/native/babel-compat/src/main.rs

Could you help me figure out how to run a visitor?

kdy1:
@dwoznicki Good to hear!

You have to implement `Node` for one invalid node. (`swc_ecma_ast::Invalid` would work)
(It's a design flaw, but fixing it would break really lots of deno code)
Then, you can use it as a first argument to `visit_with`.

e.g. `ast_node.visit_with(&Invalid { span: swc_common::DUMMY_SP }, &mut your_visitor);`

dwoznicki:
Success! Okay, time to start normalizing.

As an aside, this change bumps by build time to well above 1 minute, often closer to 2 minutes. I'll probably make visitor code `#[cfg(test)]` only.

kdy1:
Another way would be using multiple crates. If you decide to do so, you should split ast, too.

It would be like
 - `swc_babel_ast`: AST definitions.
 - `swc_babel_visit`: Depends on ast crate and invokes `define!`
 - `swc_babel_compat`: Depends on ast and visitor crate.


It's even more work, but it worths it in my experience.

dwoznicki:
@kdy1 What are the benefits of splitting up the project?

kdy1:
@dwoznicki Splitting crates reduces build time drastically.

dwoznicki:
Okay, I have a few normalizers up, and some of the (very basic) tests are finally passing. See e6ba942 for current implementation. It's quite similar to [swc_ecma_ast::VisitMut](https://rustdoc.swc.rs/swc_ecma_visit/trait.VisitMut.html) but replace the swc node names with babel node names.

The normalizer structure might change if I can figure out how to walk a tree with multiple visitors at once.. @kdy1 Any ideas? I tried making a struct with multiple `visit_mut_*` functions in the `VisitMut` impl, but the functions did not appear to be called. I might have messed up though, so I'll need to double check later when it's not so late.

@Gregoor I think everything should be mostly set up to start testing in earnest. Feel free to continue with TypeScript or do anything else. I'm going to work on testing classes for now, I think.

kdy1:
@dwoznicki It's impossible to use multi visitor at once because it's recursive.
`A.visit_mut_expr` calls `A.visit_mut_call_expr` and it can't be overridden because rust does not have overrides.

If visitor's method is not called, it's very likely that you forgot calling `node.visit_mut_children_with(self)` in visitor function for node.

For example, 
```
fn visit_mut_expr(&mut self, e: &mut Expr) {
    e.visit_mut_chilren_with(self); // This is required to children visitor get called.
}
```

dwoznicki:
@kdy1 Oh really? That's odd, because defining one visitor struct per node type ([like this](https://github.com/swc-project/swc/blob/e6ba9421218d708ad7318ec3469382bf814868c6/native/babel-compat/src/normalize/class.rs)) and then calling `ast.visit_mut_with` for each visitor struct ([like this](https://github.com/swc-project/swc/blob/e6ba9421218d708ad7318ec3469382bf814868c6/native/babel-compat/src/normalize/mod.rs)) seems to work fine without running `visit_mut_chilren_with`. Maybe I've just gotten lucky?

What didn't work was creating a single `Normalizer` super struct and throwing all the `visit_mut_*` functions in there. For example,

```rust
struct Normalizer;

impl VisitMut for Normalizer {
    fn visit_mut_class_declaration(&mut self, node: &mut ClassDeclaration) {
        // Do something
    }

    fn visit_mut_identifier(&mut self, node: &mut Identifier) {
        // Do something
    }
}
```

In this case, neither `visit_mut_class_declaration` nor `visit_mut_identifier` got called, but it sounds like that's expected since it's unsupported behavior.

kdy1:
>  That's odd, because defining one visitor struct per node type (like this) and then calling ast.visit_mut_with for each visitor struct (like this) seems to work fine without running visit_mut_chilren_with. Maybe I've just gotten lucky?

It works because the default implementation of visitor methods actually calls `visit_mut_children_with`.


> In this case, neither visit_mut_class_declaration nor visit_mut_identifier got called, but it sounds like that's expected since it's unsupported behavior.

Methods are expected to be called at least once, even though there's no code to visit children nodes.
Maybe a bug of swc_visit? Or maybe you forgot adding definitions to `define!` call.
(I made such a mistake a long time ago)

dwoznicki:
> Maybe a bug of swc_visit? Or maybe you forgot adding definitions to define! call.

Haha, that sounds like something I'd do. But no, this time I think the issue was that I needed to call `visit_mut_children_with`. I think one `visit_mut_*` method was getting called in fact, and I was just missing it. In any case, this snippet appears to work properly.

```rust
struct Normalizer;
    
impl VisitMut for Normalizer {
    fn visit_mut_variable_declarator(&mut self, node: &mut VariableDeclarator) {
        if node.definite == None {
            node.definite = Some(false);
        }
        node.visit_mut_children_with(self);
    }       
            
    fn visit_mut_variable_declaration(&mut self, node: &mut VariableDeclaration) {
        if node.declare == None {
            node.declare = Some(false);
        }   
        node.visit_mut_children_with(self);
    }       
            
    fn visit_mut_identifier(&mut self, node: &mut Identifier) {
        if node.optional == None {
            node.optional = Some(false);
        }
        if node.decorators == None {
            node.decorators = Some(vec![]);
        }
        node.visit_mut_children_with(self);
    }
}
```

@kdy1 Thanks again for your help!

dwoznicki:
@kdy1 Hey, I realized I have a problem serializing/deserializing nested enum types, and wanted to make sure you don't already have a system before I go spend a bunch of time on it.

Basically the issue is

```rust
// Structs representing AST nodes
struct A;
struct B;
struct C;

// Enum representing TypeScript style union types
enum AB {
    A(A),
    B(B),
}
enum ABC {
    AB(AB),
    C(C),
}

// How do you get ABC(AB(A)) from this JSON?
let json = r#"{"type": "A"}"#;
let a: ABC = serde_json::from_str(&json).unwrap(); // panics
```

I'm imagining a custom Deserialize derive macro that flattens out nested enums, but I'm not sure if that's the way to go.

kdy1:
Actually, there's a system for it.

`#[ast_node]` derives correct `Deserialize`, based on annotations.
Basically, each struct has its own `name`, and you can annotated variants of enums with `#[tag]` to help deserializer.

If multiple ast nodes deserializes to one variant, you can use `#[tag]` multiple time like

https://github.com/swc-project/swc/blob/dbec753ca1b3253c9cea6b6e650606287a7562f2/ecmascript/ast/src/expr.rs#L84-L91

and if you need a wildcard, `#[tag("*")]` will work. See: https://github.com/swc-project/swc/blob/dbec753ca1b3253c9cea6b6e650606287a7562f2/ecmascript/ast/src/pat.rs#L35-L38

dwoznicki:
Aha! Glad I asked. I'll try it out and let you know if I have any more questions.

dwoznicki:
@kdy1 So adding `#[ast_node]` might not be so simple since babel AST nodes don't implement Spanned. Would it be easy to create a similar macro without Spanned?

kdy1:
`Spanned` is not important, but I think creating an alternative macro like e.g. `#[ast_serde]` using the source code of `#[ast_node]` and making `#[ast_node]` expands to `#[ast_serde]` would be easier.
If you want, I can do it for you.

dwoznicki:
Hrmm, yeah go for it please. I think I could eventually figure this out, but it'll probably be faster if you take care of it.

kdy1:
I added a macro named `#[ast_serde]` and documented it at https://rustdoc.swc.rs/ast_node/attr.ast_serde.html

dwoznicki:
Cool, just tested it and everything is working 👌

kdy1:
@dwoznicki Do you mean the pr is done?
Wow! Nice work! Thanks!

dwoznicki:
Oh no, I meant that `#[ast_serde]` is working, haha. Still working on the PR.

kdy1:
Ah, I see. But it's cool anyway!

dwoznicki:
Yeah, sorry it's taking so long. I don't have a ton of time to work on it each day, and it seriously seems like every single test I write uncovers another serious issue.

kdy1:
It's not yout fault, so you should not be sorry for it.

> every single test I write uncovers another serious issue.

It's really common if you play with javascript ast. I was the same when I worked on js transforms. :)

dwoznicki:
@kdy1 Hey, another implementation question. Babel collects all the comments in a file and puts them in a single file ([source](https://github.com/babel/babel/blob/368bf893fad7cc6691840139295a8a55ff10199b/packages/babel-types/src/ast-types/generated/index.ts#L457)), but I don't see a way to do the same with `swc_common::comments::Comments`.

I may regret this confidence later, but it doesn't look that hard to implement: just concat the `SingleThreadedCommentsMapInner` values for leading and trailing. It'll probably require cloning the comments, which isn't great, but I'm not sure if there's a way around that. So I could add the `get_all` or whatever function as a stub to `Comments` and make `SingleThreadedComments` implement it, or maybe downcast `Comments` and get the values with `borrow_all` maybe?

kdy1:
I think adding stub methods is better as swc does not use `SingleThreadedComments ` (it uses `swc::SwcComments`), but I'm not sure about the way to get all comments of a module. I expect a huge change if we want not to degrade performance a lot.

Maybe we can just use multiple instance of `swc_common::Comments`. It would be much easier I guess.
I think this is correct way to go.

dwoznicki:
> Maybe we can just use multiple instance of swc_common::Comments. It would be much easier I guess.

Can you expand on what you mean by this?

kdy1:
I meant creating one `swc_common::Comments` per a file (module / script).
You may use `SingleThreadedComments`.

dwoznicki:
Ah, I see. You're talking about getting all the comments for a single module in a multi-module compilation, but I was just talking about collecting all the comments for a single script. For example, if I have source code

```javascript
function x() {
    // some comment
    const a = false; // another comment
}
```

I just need to collect the AST nodes for `// some comment` and `// another comment`, and put them into a `vec`. It looks like I could get that by flattening and concatenating the values of `leading` and `trailing` in `SwcComments`.

I hadn't thought about how `babelify` ing a multi-module compilation would work, but I guess I could see it being a problem if the `compiler.comments` instance contains comments for every module in the compilation, which it sounds like it does. I'm a little iffy about changing the core structure of swc's comments just to get the correct comments array in the babel AST for multi-module compilations. That is, unless you expect `babel-compat` to be an important feature. Then it might be worth it.

kdy1:
Yes, `swc_common::Comments` stores comments for all modules.
But there's a way to get all comments of a specific module. (I forgot about it)

See: https://rustdoc.swc.rs/src/swc_ecma_transforms_react/refresh/mod.rs.html#534-562

dwoznicki:
Thanks, I got a visitor working for collecting comments.

dwoznicki:
@kdy1 Hey, I think this PR is getting close. There are almost certainly more inconsistencies I've missed, but I'm running out of ideas for tests to write. With one big exception... TypeScript.

I've never written TypeScript, so I only understand it conceptually. Writing TS tests is going to take me a while without help. So the question is, do you want to wait for TS compatibility testing before merging?

kdy1:
@dwoznicki Wow. Nice work! Thank you so much!

I don't think typescript conversion is important at the moment. It's because using type information is very complex, and to use type information, what you need the ast of tsc, not babel. So no one would try to use type ast nodes of babel.

dwoznicki:
Great! Well, I think this PR is actually ready in the case. I suspect there will be a number of bugs at first, so feel free to @ me if you like.

kdy1:
Great Thanks!
I'll try implementing babel plugin runner asap :)

PR #1467, opened 2021-03-12
Add note about required dev tools

arendjr:
I had most of the tooling installed, but `make` was missing, causing me a bit of head-scratching. Let's be nicer to the next person :)


kdy1:
Thanks!

PR #1470, opened 2021-03-14
Fix column positions in generated sourcemaps

devongovett:
I noticed that all mappings were pointing to column zero in the original source. This was because the result of `calc_extra_bytes` was used rather than subtracting from the original position. I copied this logic from the `bytepos_to_file_charpos_with` function in the same file.

Before:
image

After:
Screen Shot 2021-03-14 at 1 52 00 PM

Notice the red mappings point to the corresponding source in the original rather than a single mapping for the whole line.

I'm happy to add tests, but I didn't see an existing one to base it off of. Let me know.


kdy1:
Thanks!

PR #1478, opened 2021-03-19
fix(strip): Combine typescript_class_properties() into strip()

nayeemrmn:
@kdy1 This implements `use_define_for_class_fields: true` as correctly specified in #1472. It also adds the behaviour of the `typescript_class_properties()` pass directly into `strip()` in case of `use_define_for_class_fields: false`. This aligns with tsc. The `typescript_class_properties()` is redundant and should be removed.

However, I can't figure out how to copy the static class field handling from `typescript_class_fields()` to `strip()`. So I left some TODOs and haven't removed it yet. Help would be appreciated.

### DIFF
--- a/ecmascript/transforms/typescript/src/strip.rs
+++ b/ecmascript/transforms/typescript/src/strip.rs
@@ -5,13 +5,10 @@ use swc_atoms::{js_word, JsWord};
 use swc_common::{util::move_map::MoveMap, Span, Spanned, SyntaxContext, DUMMY_SP};
 use swc_ecma_ast::*;
 use swc_ecma_transforms_base::ext::MapWithMut;
-use swc_ecma_utils::default_constructor;
-use swc_ecma_utils::member_expr;
-use swc_ecma_utils::prepend_stmts;
 use swc_ecma_utils::private_ident;
-use swc_ecma_utils::quote_ident;
 use swc_ecma_utils::var::VarCollector;
 use swc_ecma_utils::ExprFactory;
+use swc_ecma_utils::{constructor::inject_after_super, default_constructor};
 use swc_ecma_utils::{ident::IdentLike, Id, StmtLike};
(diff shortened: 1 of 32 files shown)


ecmascript/transforms/typescript/src/strip.rs
```diff
@@ -145,141 +153,100 @@ impl Strip {
...
-                        expr: define_property_call,
-                    });
-
-                    define_property_stmts.push(stmt);
+                    assign_exprs.push(assign_expr);
+                }
+                ClassMember::ClassProp(_) => {
+                    // TODO(nayeemrmn): Static fields should also be moved to
+                    // assignments after the class. Figure out how. They are
+                    // preserved for now.
```
nayeemrmn, line 225: Implementation.

ecmascript/transforms/typescript/tests/strip.rs
```diff
@@ -746,6 +748,8 @@ test!(
 
 test!(
     ::swc_ecma_parser::Syntax::Typescript(Default::default()),
+    // TODO(nayeemrmn): This output should be achieved without
+    // `typescript_class_properties()`.
```
nayeemrmn, line 752: Test.

nayeemrmn (commented):
TODOs for static class field handling when `use_define_for_class_fields: false`.

kdy1:
Sorry for late response. I was too busy because of other projects.

The changes seems ok to me. But I have a question.
Doesn't this change the default behavior of swc even when target < es2020 ?

nayeemrmn:
> Doesn't this change the default behavior of swc even when target < es2020 ?

Not totally sure what you mean by "when target < es2020",  but any change in behaviour is non-breaking and aligns with tsc. The behaviour of `chain!(strip(), typescript_class_properties())` should be exactly the same.

kdy1:
@nayeemrmn Thank you very much for the pr and clarification :)

kdy1 (approved):
LGTM.
I'll merge it today.

PR #1479, opened 2021-03-19
Add failing test for xhtml code escape in tsx

vovacodes:
This PR adds a breaking test illustrating a bug in printing of transformed tsx code where html escape sequenced in source tsx are not unescaped in the compiled code when compilation target is > ES5.

It looks like the difference in behaviour for target <= ES5 and the newer versions is here: https://github.com/swc-project/swc/blob/master/ecmascript/codegen/src/lib.rs#L2497
I'm not sure I really understand this logic so maybe @kdy1 can clarify it and advise what would be the right way to fix it. I'm happy to include the fix into this PR if so

### DIFF
--- a/tests/simple.rs
+++ b/tests/simple.rs
@@ -1,10 +1,11 @@
 use std::sync::Arc;
+use swc::config::JscTarget;
 use swc::{
     config::{Config, JscConfig, Options},
     Compiler,
 };
 use swc_common::FileName;
-use swc_ecma_parser::{EsConfig, Syntax};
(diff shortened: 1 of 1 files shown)


vovacodes:
@kdy1 sorry for bothering you but this bug affects us quite severily, I'd love to contribute a fix, just need a bit guidance here

kdy1:
I fixed it with https://github.com/swc-project/swc/pull/1501 :)

vovacodes:
Thanks a ton, you are a great human! 🙌

PR #1487, opened 2021-03-23
fix(strip): Transform static class fields to assignments

### LINKED ISSUE #1483: strip: Static class fields should be transformed by default
Tracking issue for TODO: https://github.com/swc-project/swc/blob/9bc074ed469a7b3e85a0ef23df27ff857fe44b84/ecmascript/transforms/typescript/src/strip.rs#L232-L234 This  only applies when `use_define_for_class_fields` is `false`. When it's `true`, the current behaviour of preserving all class fields is correct.

Input:
```ts
class A {
  static a = 1;
}
```

Actual output:
```js
class A {
    static a = 1;
}
```

Expected output:
```js
class A {
}
A.a = 1;
```

This can be achieved by additionally applying the `typescript_class_properties()` pass but it should be the default behaviour of `strip()`. ~~Once this is implemented, `typescript_class_properties()` would be redundant and should be removed.~~ Edit: No... of course it would still be useful as a TS -> TS transform.

nayeemrmn:
Fixes #1483.

Also ensures that every test applies the `decorator()` pass before `strip()`. This is necessary now, and it makes sense because there shouldn't be decorators in JS.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2018"
 license = "Apache-2.0/MIT"
 name = "swc"
 repository = "https://github.com/swc-project/swc.git"
-version = "0.12.0"
+version = "0.12.1"
 
 [lib]
 name = "swc"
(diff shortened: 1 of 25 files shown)


nayeemrmn:
@kdy1 Any idea what causes this error? 
image

kdy1:
@nayeemrmn I don't have one, but I now have some time to investigate it. I'll fix it and then merge it.

PR #1498, opened 2021-03-26
fix(strip): Precompute class field keys

### LINKED ISSUE #1497: strip: Class field keys are not precomputed
Note that `class_properties()` already has the correct behaviour, but I didn't transfer this to `strip()` properly.

Input:
```ts
class A {
  [(console.log(1), "a")] = 1;
}
```

Actual output:
```js
class A {
  constructor() {
    this[(console.log(1), "a")] = 1;
  }
}
```

Expected:
```js
var _key;
class A {
  constructor() {
    this[_key] = 1;
  }
}
_key = (console.log(1), "a");
```
This would correctly output `1` on class declaration, rather than once per construction.

nayeemrmn:
Fixes #1497.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,25 +1,2 @@
 {
-  "files.exclude": {
-    "**/.git": true,
-    "**/.svn": true,
-    "**/.hg": true,
-    "**/CVS": true,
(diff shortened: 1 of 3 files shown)


kdy1:
Thanks!

PR #1510, opened 2021-03-28
fix(strip): Expand class expressions as sequences

### LINKED ISSUE #1502: strip: Sequence expressions in class field initialisers missing parentheses
Input:
```ts
class A {
  a = (console.log(1), 2);
}
```

Actual output:
```js
class A {
  a = console.log(1), 2;
}
```

Expected output: Same as input.

### LINKED ISSUE #1509: strip: Use sequence expressions for class expression expansion
For TODO: https://github.com/swc-project/swc/blob/8eae00900f88b09cf04dc1a991f815124a07d08f/ecmascript/transforms/typescript/src/strip.rs#L450-L454

TSC uses sequence expressions here. We can as well.

nayeemrmn:
Fixes #1509.
Fixes #1502.

### DIFF
--- a/bundler/tests/fixture/deno-9591/output/entry.inlined.ts
+++ b/bundler/tests/fixture/deno-9591/output/entry.inlined.ts
@@ -2256,10 +2256,10 @@ class ConsoleHandler extends BaseHandler {
     }
 }
 class WriterHandler extends BaseHandler {
-    #encoder=new TextEncoder();
+    #encoder = new TextEncoder();
 }
 class FileHandler extends WriterHandler {
-    #unloadCallback=()=>this.destroy()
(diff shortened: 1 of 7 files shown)


nayeemrmn:
@kdy1 I hit an assertion error, the actual and expected codes match but the output ast is different from the one passed from the expected string: 
image

nayeemrmn:
Nevermind, I think I fixed it by changing
```rust
PatOrExpr::Expr(Box::new(Expr::Ident(ident)));
```
to
```rust
PatOrExpr::Pat(Box::new(Pat::Ident(BindingIdent {
    id: ident,
    type_ann: None,
})))
```
Not sure what the difference is.

PR #1521, opened 2021-03-30
Fix TS import elision with shadowed declarations

devongovett:
This fixes a few issues with import elision in TypeScript to better match TSC output.

When there is a local concrete declaration of the same name as an import, TSC assumes the import is a type and removes it. For example:

```ts
import { Test } from 'test';
const Test = 2;
console.log(Test);
```

However, if the local declaration is a type, and the name is referenced concretely, the import should not be removed.

```ts
import { Test } from 'test';
interface Test {}
console.log(Test);
```

In addition, imports are hoisted, so when an import is referenced before it is declared, the import should not be removed. This requires tracking all referenced identifiers when traversing, not just ones we've seen an import for already.

```ts
console.log(Test);
import { Test } from 'test';
```

See added tests for more cases.


ecmascript/transforms/typescript/src/strip.rs
```diff
@@ -1113,26 +1113,50 @@ impl Strip {
...
     fn visit_decl(&mut self, n: &Decl, _: &dyn Node) {
         self.handle_decl(n);
 
         let old = self.non_top_level;
         self.non_top_level = true;
-        n.visit_children_with(self);
+        // Visit the declaration body and not the identifier. Otherwise
+        // type declarations could be unintentionally marked as concrete
+        // by the visit_ident function above.
+        match n {
```
kdy1, line 1134: Is there a reason to handle this in `visit_decl` instead of `visit_class_decl`?
(It's not important, just curous)

ecmascript/transforms/typescript/src/strip.rs
```diff
@@ -1113,26 +1113,50 @@ impl Strip {
...
+            Decl::Class(class) => {
+                self.decl_names.insert(class.ident.to_id());
+                class.class.visit_with(class, self);
+            }
+            Decl::Fn(f) => f.function.visit_with(f, self),
+            Decl::Var(ref var) => {
+                for decl in &var.decls {
+                    decl.init.visit_with(decl, self);
```
kdy1, line 1142: This can break code if it contains a default value for a pattern.

For example, 

```ts
import { b } from './foo';
const [a = b] = []
```

`b` will not be visited.

kdy1: I typically workaround this by storing a flag (in_var_pat or something) in the struct and ignoring `Pat::Ident` if the flag is set.

kdy1: Note that you have to visit children nodes of other patterns, because they may contain some expressions as a default value.

ecmascript/transforms/typescript/src/strip.rs
```diff
@@ -1334,10 +1361,24 @@ impl VisitMut for Strip {
...
             ImportSpecifier::Default(ImportDefaultSpecifier { ref local, .. })
-            | ImportSpecifier::Named(ImportNamedSpecifier { ref local, .. }) => {
+            | ImportSpecifier::Named(ImportNamedSpecifier { ref local, .. })
+            | ImportSpecifier::Namespace(ImportStarAsSpecifier { ref local, .. }) => {
+                // If the import is shadowed by a concrete local declaration, TSC
+                // assumes the import is a type and removes it.
+                let decl = self
+                    .scope
+                    .decls
+                    .get(&(local.sym.clone(), local.span.ctxt()));
```
kdy1, line 1371: You can do `&local.to_id()` instead.

ecmascript/transforms/typescript/tests/strip.rs
```diff
@@ -3706,3 +3706,123 @@ to!(
     ",
     ""
 );
+
```
kdy1, line 3709: Thanks for lots of tests!!

kdy1 (changes requested):
Looks like a good change to me, but I think there are some edge cases.

ecmascript/transforms/typescript/src/strip.rs
devongovett: You mean create individual visitor functions for each declaration type? I could do that if it's preferable. Using `match` does verify that we're handling all types of declarations if one is ever added I guess. 🤷

ecmascript/transforms/typescript/src/strip.rs
devongovett: Good catch.

kdy1 (approved):

PR #1524, opened 2021-03-30
feat(es/transforms/react): Support fast refresh

Austaras:
Continue on @vovacodes 's great work to port react fast refresh to swc. Due to fundamental difference between babel and swc it's not feasible to produce exact same output, but I managed to pass all original tests.

I'll figure out option tomorrow. There's still one thing need help @kdy1 : How to iterate over every comment in a single file like in [original code](https://github.com/facebook/react/blob/634cc52e6174b506dee79d908f50ce1ee19f4fbf/packages/react-refresh/src/ReactFreshBabelPlugin.js#L263)? Maybe you could add an extra method in `Comments` trait?


kdy1:
Do you mean literally all comments of a file?
If then, I'll add a method to `Comments`.

Austaras:
Yes I do mean every comment. It's  quite confusing why react team chooses to do so though.

kdy1:
Hmm.. I found that it's very hard to get all comments without knowing spans.
Also, because rust does not have a gc, so... storing same comment multiple times is not an option.

So I used a bit different approach. `CommentCollector` can be used to check all comments.

Austaras:
I have tested your apporach, however comment in following code doesn't seem like to be detected
``` js
    function Bar() {
      const [foo, setFoo] = useState(0);
      React.useEffect(() => {
        // @refresh reset
      });
      return <h1>{foo}</h1>;
    }
```
seems like any comment in a empty `BlockStmt` will be ignored, could you fix it?

kdy1:
Nice catch! I'll fix it with another pr asap.

kdy1:
I fixed the issue, but I guess you should get comments by getting leading comments of `Span.hi`.

Austaras:
Updated. Another problem: why does swc generate
```js
    const bar = _hooks.useFancyState();
```
instead of
```js
    const bar = (0, _hooks).useFancyState();
```
when targeting commonjs?

kdy1 (approved):
LGTM. Thank you a lot!

Austaras:
Thanks for taking your time to review this! I'll continue to work on this branch to catch up with the update to the original implementation this morning, and provide a rich config options to end user

PR #1536, opened 2021-04-03
Fix inlining nested block statements in branch simplifier

devongovett:
This fixes an issue where the branch simplifier would inline single statement blocks too aggressively, resulting in broken code. For example:

```js
if (Date.now() < 0) {
    for(let i = 0; i < 10; i++){
        if (Date.now() < 0) {
            console.log(1);
        }
    }
} else {
    console.log(2);
}
```

gets turned into

```js
if (Date.now() < 0) for(let i = 0; i < 10; i++) if (Date.now() < 0) console.log(1);
else console.log(2);
```

This results in the `else` being applied to the second `if` rather than the first.

I've attempted to solve this by computing the depth of statements in the block's first statement until hitting another block statement. If the depth is greater than 1, then it's not safe to inline the block.

Perhaps there's a better way to solve this? i.e. should the printer be handling this based on precedence?


kdy1 (approved):
Seems like a reasonable approach to me. Thanks!

PR #1537, opened 2021-04-03
support TS 4.3 static index signature in class

g-plane:
Reference: https://devblogs.microsoft.com/typescript/announcing-typescript-4-3-beta/#static-index-signatures


kdy1:
Thanks you a lot!!

PR #1539, opened 2021-04-04
Fix function hoisting with early return branch simplification

devongovett:
This fixes an issue in the branch simplifier where functions declared after a return statement were removed. These functions need to be left in place and are accessible via hoisting. Adjusted to move the declarations before the return statement along with `var` declarations.


ecmascript/transforms/optimization/src/simplify/branch/mod.rs
```diff
@@ -969,15 +969,29 @@ impl Remover {
...
+                                            VarDeclarator {
+                                                span: i.span,
+                                                name: Pat::Ident(i.into()),
+                                                init: None,
+                                                definite: false,
+                                            }
+                                        });
+                                        decls.extend(ids);
+                                    }
+                                    _ => {}
```
kdy1, line 991: Doesn't this drop module items?
I'll see the whole code.

kdy1: I think it should be 
`Err(item) => buf.push(item)`

kdy1 (changes requested):
Thank you a lot!
I found one issue related to usage of generics

kdy1 (approved):
LGTM. Thanks!

PR #1540, opened 2021-04-04
Fix named function expression handling in resolver

devongovett:
Code like the following was broken, because the function expression name was added to the outer scope rather than the inner scope.

```js
export default function foo() {
    foo = function foo(x) {
        return x === 0 ? 1 : 1 + foo(x - 1);
    };
    return foo(10);
}
```

resulted in

```js
export default function foo() {
    foo = function foo1(x) {
        return x === 0 ? 1 : 1 + foo1(x - 1);
    };
    return foo1(10); // <- this is wrong
};
```

Fixed by applying the function expression name to the inner scope instead.

(As a side note, I found it weird that `export default function foo() {}` parses as an `FnExpr` instead of a `FnDecl`. Same with classes. Babel parses them as declarations rather than expressions. Is this intentional?)

### DIFF
--- a/ecmascript/transforms/base/src/resolver/mod.rs
+++ b/ecmascript/transforms/base/src/resolver/mod.rs
@@ -511,12 +511,23 @@ impl<'a> VisitMut for Resolver<'a> {
     }
 
     fn visit_mut_class_expr(&mut self, n: &mut ClassExpr) {
-        let old = self.ident_type;
-        self.ident_type = IdentType::Binding;
-        n.ident.visit_mut_with(self);
-        self.ident_type = old;
+        // Create a child scope. The class name is only accessible within the class.
+        let child_mark = Mark::fresh(self.mark);
 
-        n.class.visit_mut_with(self);
+        let mut folder = Resolver::new(
+            child_mark,
+            Scope::new(ScopeKind::Fn, Some(&self.current)),
+            self.cur_defining.take(),
+            self.handle_types,
+        );
+
+        folder.ident_type = IdentType::Binding;
+        n.ident.visit_mut_with(&mut folder);
+        folder.ident_type = IdentType::Ref;
+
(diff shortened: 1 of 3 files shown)


devongovett:
Actually, I think this will break some things unless the parser is changed to treat export default declarations as declarations rather than expressions.

```js
export default function foo() {
    foo = function foo(x) {
        return x === 0 ? 1 : 1 + foo(x - 1);
    };
    return foo(10);
}

foo = 2;
```

now becomes:

```js
export default function foo1() {
    foo1 = function foo1(x) {
        return x === 0 ? 1 : 1 + foo1(x - 1);
    };
    return foo1(10);
};
foo = 2;
```

Would you like me to make that change?

kdy1:
> I found it weird that export default function foo() {} parses as an FnExpr instead of a FnDecl.

Actually it emits `DefaultDecl`, which has FnExpr as a variant. It's intentional, mainly because I didn't want to make `FnDecl` have `Option<Ident>` for its name.
In this way, the user can override the handling of 'function with optional ident' by implementing `visit_default_decl`.


Ideally, I think it should be


```
export default function foo() {
    foo = function foo(x) {
        return x === 0 ? 1 : 1 + foo(x - 1);
    };
    return foo(10);
};
foo = 2;
```

Using `visit_default_decl` may allow this. Although I'm not sure.

devongovett:
I guess anonymous functions could be parsed as `ExportDefaultExpr` rather than `DefaultDecl`?

I could try to solve it by implementing `visit_default_decl` in the resolver though, which seems like a smaller change.

kdy1:
@devongovett It's impossible because typescript allows exporting an interface as a default item.

> I could try to solve it by implementing visit_default_decl in the resolver though, which seems like a smaller change.

Thanks!

PR #1541, opened 2021-04-05
support TS 4.3 `override` syntax in class

g-plane:
Reference: https://devblogs.microsoft.com/typescript/announcing-typescript-4-3-beta/#override-and-the-noimplicitoverride-flag


Skalman:
@g-plane, just wanted to double check one thing. I saw a commit message _"disallow `override` to precede accessibilities"_. Not really sure what you mean, but does the following still work as expected?

```ts
class B {}

class C extends B {
    override private constructor() {
        super();
    }
}
```

[TS playground](https://www.typescriptlang.org/play?ts=4.3.0-dev.20210405&ssl=1&ssc=1&pln=7&pc=2#code/MYGwhgzhAEBC0G8C+AoFpIwMLQKYA8AXXAOwBMZ4EVpboB7AN1wCcWBLM3aABw8bDFoweiQiEWAV2CF6LABQBKRDTpqIknqyUBuVbVRIgA)

g-plane:
@Skalman I did it based on this: https://github.com/microsoft/TypeScript/blob/2f0c8b2bea3554f9371e1d2076df49498d12c5a7/tests/baselines/reference/override5.errors.txt#L5 .

g-plane:
Let's see how TypeScript team answer: https://github.com/microsoft/TypeScript/issues/43533 . If I'm wrong, I'll update this PR.

g-plane:
As this PR of TypeScript: https://github.com/microsoft/TypeScript/pull/43544 , accessibility modifiers must precede `override`, so there's no need to change in this PR, but maybe I need to add another check: https://github.com/microsoft/TypeScript/issues/43533#issuecomment-813569390 .

g-plane:
https://github.com/microsoft/TypeScript/issues/43533 is closed because https://github.com/microsoft/TypeScript/pull/43544 is merged, so we can go ahead.

kdy1 (approved):
LGTM. Thank you a lot!

PR #1544, opened 2021-04-05
Fix handling of tagged template expressions in Visitor

### LINKED ISSUE #1543: Type mismatch in tagged template expression
**Describe the bug**

I'm trying to write an SWC plugin. Unfortunately, it seems the default `Visitor` implementation in TypeScript doesn't match the structure of the AST it receives, resulting in an exception on `TaggedTemplateExpression` nodes:

```
Module build failed (from ./node_modules/swc-loader/src/index.js):
TypeError: Cannot read property 'map' of undefined
    at LinariaProcessor.visitTaggedTemplateExpression (/home/arendjr/Projects/linaria-swc/node_modules/@swc/core/Visitor.js:837:39)
```

It seems `Visitor.ts` expects `TaggedTemplateExpression` nodes to have an `expressions` array, whereas upon inspection it appears this array is inside a `template` object on the node.

**Input code**

This is the JS I'm processing:

```js
import { css } from "@linaria/core";

const exampleClassName = css`
  background-color: green;
`;
```

This is the (so far unimplemented) plugin:

```ts
import type { Program } from "@swc/core";
import Visitor from "@swc/core/Visitor";

class LinariaProcessor extends Visitor {}

export default function plugin(n: Program): Program {
  return new LinariaProcessor().visitProgram(n);
}
```

**Config**

**.swcrc**

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": true,
      "dynamicImport": false,
      "privateMethod": false,
      "functionBind": false,
      "exportDefaultFrom": false,
      "exportNamespaceFrom": false,
      "decorators": false,
      "topLevelAwait": false,
      "importMeta": false
    },
    "transform": null,
    "target": "es2020",
    "loose": false,
    "externalHelpers": false,
    "keepClassNames": false
  }
}
```

**webpack.config.js**

```js
const path = require("path");

const linariaPlugin = require("../lib").default;

module.exports = {
  entry: "./src/index.js",
  output: {
    filename: "main.js",
    path: path.resolve(__dirname, "dist"),
  },
  module: {
    rules: [
      {
        test: /\.m?js$/,
        exclude: /(node_modules|bower_components)/,
        use: {
          loader: "swc-loader",
          options: {
            plugin: linariaPlugin,
            jsc: {
              parser: {
                jsx: true,
              },
              target: "es2020",
            },
          },
        },
      },
    ],
  },
};
```

**Expected behavior**

No exception should be thrown.

**Version**
The version of @swc/core: **1.2.51**

**Additional context**
On a related note, while I was attempting to implement some parts of the plugin, I also noticed the `imported` property of `NamedImportSpecifier` nodes may be `null`, though the TypeScript definitions do not indicate this. It would be helpful if nullability were properly indicated as well.

arendjr:
Fixed #1543.

Fingers crossed this is going to pass the tests, as I still have trouble getting them all to run locally...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2018"
 license = "Apache-2.0/MIT"
 name = "swc"
 repository = "https://github.com/swc-project/swc.git"
-version = "0.13.0"
+version = "0.14.0"
 
 [lib]
 name = "swc"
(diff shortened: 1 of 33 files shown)


kdy1 (approved):
Thank you!

PR #1552, opened 2021-04-07
fix(es): disallow `override` in non-subclass

g-plane:
With this PR, the code below will be invalid:

```ts
class C {
  override t() {}
}
```


kdy1 (approved):
Thanks!

PR #1559, opened 2021-04-09
Fix nested function scopes with duplicate var declarations

devongovett:
Currently, `var` declarations (and hoisted function declarations) are treated differently by the resolver than block scoped declarations (`let`, `const`, `class`, etc.) in regards to nested declarations of the same name (e.g. within a function expression). Specifically, this block of code may decide to reuse an existing name/mark rather than define a new one: https://github.com/swc-project/swc/blob/2211a9908a9a1c2e01d5f483ad91ac396336dc08/ecmascript/transforms/base/src/resolver/mod.rs#L280-L294.

This currently causes some problems. For example, this code:

```js
var Test = (function () {
  var Test = (function () {
      var Test = 2;
      return Test;
  })();

  return Test;
})();
```

compiles to:

```js
var Test = function() {
    var Test = function() {
        var Test1 = 2;
        return Test; // <- this is wrong
    }();
    return Test;
}();
```

This results in broken code when there is more than two levels of nesting (I'm not sure why this is the case). It also only happens with hoisted declarations, not with block scoped declarations, which declares a separate name for each level rather than reusing.

I've attempted to fix this by not propagating `cur_defining` into child function scopes, but I'm not sure this is correct. A bunch of tests fail because of this, and I'm not sure if changing the expected output to match would be desirable. I'm opening this PR to get feedback on the approach.

Questions:

1. What was the reasoning for making nested hoisted declarations different from block scoped declarations? My understanding was that the role of the resolver is to resolve references to declarations, but if multiple declarations in different scopes have the same context then doesn't this defeat the purpose?
2. Is `cur_defining` used for anything else than this? If we move forward with this approach, should we remove it completely?
3. If not, I see a bunch of places where `cur_defining` is propagated out of the child scope into the parent (e.g. `self.cur_defining = child.cur_defining`). This seems like it could also be problematic and doesn't really make sense to me. Could you explain how this is meant to work?


devongovett:
@kdy1 any thoughts on this one?

kdy1:
> 1. What was the reasoning for making nested hoisted declarations different from block scoped declarations? My understanding was that the role of the resolver is to resolve references to declarations, but if multiple declarations in different scopes have the same context then doesn't this defeat the purpose?

It seems like my mistake. Resolver should treat

```js
var Test = (function () {
  var Test = (function () {
      var Test = 2;
      return Test;
  })();

  return Test;
})();
```

as

```js
var Test1 = (function () {
  var Test2 = (function () {
      var Test3 = 2;
      return Test3;
  })();

  return Test2;
})();
```



> 2. Is cur_defining used for anything else than this? If we move forward with this approach, should we remove it completely?

`cur_defining` is a hack for 

```js
var f = function f() {}
```

and 

```js
var Test = (function () {
  var Test = (function () {
      var Test = 2;
      return Test;
  })();

  return Test;
})();
```

as

```js
var Test1 = (function () {
  var Test2 = (function () {
      var Test3 = 2;
      return Test3;
  })();

  return Test2;
})();
```



> 2. Is cur_defining used for anything else than this? If we move forward with this approach, should we remove it completely?

`cur_defining` is for 

```js
var f = function () {
    f()
}
```

If I remember correctly.
At the time of evaluating function body, `f` is not declared.
So we need to track WIP-decls.

But if it can be removed, it would be nice.

> 3.If not, I see a bunch of places where cur_defining is propagated out of the child scope into the parent (e.g. self.cur_defining = child.cur_defining). This seems like it could also be problematic and doesn't really make sense to me. Could you explain how this is meant to work?

It should not be propagated I think. But my intention of copying them is to preserve `cur_defining` passed from the parent.

https://github.com/swc-project/swc/blob/2211a9908a9a1c2e01d5f483ad91ac396336dc08/ecmascript/transforms/base/src/resolver/mod.rs#L439

But cloning sould be much better I guess.

ecmascript/transforms/base/src/resolver/tests.rs
```diff
@@ -542,11 +539,11 @@ to!(
         }(Bar);
     "#,
     r#"
-    var Foo = function(_Bar) {
-            _inherits(Foo, _Bar);
-            function Foo() {
+    var Foo1 = function(_Bar) {
```
devongovett, line 542: Changes like this are due to an issue in the hygiene pass that already existed: if a variable is referenced before it is defined, it will be renamed. For example:

```js
var Foo;
function test() {
  test(Foo);
  function Foo() {}
}
```

is compiled to:

```js
var Foo1;
function test() {
    test(Foo2);
    function Foo2() {
    }
}
```

That's due to this:

https://github.com/swc-project/swc/blob/8f5daa3bbbcc05997a71dd23159347ad3622238c/ecmascript/transforms/base/src/hygiene/mod.rs#L116-L118

The code works but more vars are renamed than really needed. Removing those lines fixes this issue, but breaks some other tests which seem to rely on this behavior so I didn't do it. Perhaps could be tackled separately?

devongovett:
Ok, I've removed `cur_defining` and updated all the tests. I think the output is now more correct, though perhaps less optimal. I think improvements to what gets renamed or not should be done in the hygiene pass rather than the resolver. It's important for passes that rely on the resolver to have correct information. For example:

```js
var f = function f() {
  console.log(f)
};
```

now results in:

```js
var f = function f1() {
    console.log(f1);
};
```

This is more correct as the console log is really referencing the function and not the outer variable declaration. The variable could be re-assigned but the console log always refers to the function. Some optimization pass might rely on this information, so I think they should have two separate syntax contexts rather than only a single one. The actual renaming could perhaps be optimized in the hygiene pass though.

ecmascript/transforms/base/src/resolver/tests.rs
kdy1:
```js
var Foo1;
function test() {
    test(Foo2);
    function Foo2() {
    }
}
```


Isn't this correct? I thought the function `Foo` is hoisted, so `Foo` in `test` refers to the function, not a variable from the parent scope.


If this is wrong, then the implementation is wrong and should be fixed.

kdy1:
> The variable could be re-assigned but the console log always refers to the function.

Oh. I didn't know that. Thanks!

kdy1 (approved):
LGTM. Thanks!

ecmascript/transforms/base/src/resolver/tests.rs
devongovett: Yeah the output is "correct", but the outer `Foo` was unnecessarily renamed to `Foo1`. If you reorder the statements such that the call to `test` is after the declaration, it is not renamed.

kdy1: I See.
Then it seems like an issue that should be fixed separately.
Thanks for the clarification!.

tests/fixture/issue-1555/output/index.js
```diff
@@ -1,5 +1,5 @@
 export default function _setPrototypeOf(o, p) {
-    _setPrototypeOf = Object.setPrototypeOf || function _setPrototypeOf(o1, p1) {
+    _setPrototypeOf = Object.setPrototypeOf || function _setPrototypeOf1(o1, p1) {
```
kdy1, line 2: I found that this is a new bug.
It should be `_setPrototypeOf` I think.

kdy1 (changes requested):
Sorry. I missed it.

tests/fixture/issue-1555/output/index.js
devongovett: I think this is the same case as described above. The inner function name is a separate binding from the outer one which is re-assigned. Assigning to the outer one doesn't re-assign the inner one. If the inner function was referenced, it would still refer to the inner function even if the outer binding is re-assigned.

I think this is correct, though again, perhaps the hygiene pass could avoid renaming in cases like this where a binding is shadowed in an inner scope and the outer binding is not referenced within that scope. That would also improve the output for cases like:

```js
let x = 2;
{
  let x = 3;
}
```

which currently results in:

```js
let x = 2;
{
    let x1 = 3;
}
```

The resolver is correct - the inner x is a different binding from the outer one, but it didn't really need to be renamed to `x1` since the outer binding is not referenced in the inner scope (or any of its children).

devongovett: Here's an example:

```js
let test = function test() {
  console.log(test);
};

setTimeout(test, 0);
test = 2;
console.log(test);
```

If you run that, you'll get:

```
2
[Function: test]
```

Within the function, `test` always refers to the function, not the outer variable. If you change the code to:

```js
let test = function() {
  console.log(test);
};

setTimeout(test, 0);
test = 2;
console.log(test);
```

Then you'll get `2` twice since there's no inner binding.

kdy1: Sorry. I misread it. (I was in hurry because of the midterm exam and now it's over)

Yes, it's the same as the case above.

kdy1: And thank you for the example :)

kdy1 (approved):

PR #1561, opened 2021-04-09
fix(es/transforms/module): transform esmodule import to sequence expression

### LINKED ISSUE #1556: Transform esmodule without break `this`
**Describe the bug**
As we have discussed in #1524, when swc transform esmodule to `commonjs|amd|umd` it will generate code like
```js
const a = require('./a')
a.foo()
```
However, that will make `this` in `foo` points to module `a` instead of `globalThis`. Although few people still use commjs target today and even fewer people use `this` to refer global, this still should be fixed. swc should instead generate code like
```js
const a = require('./a')
(0, a).foo()
```
**Version**
latest

**Additional context**

Austaras:
Fixes #1556

### DIFF
--- a/ecmascript/transforms/react/src/refresh/mod.rs
+++ b/ecmascript/transforms/react/src/refresh/mod.rs
@@ -752,7 +752,8 @@ impl<C: Comments> Fold for Refresh<C> {
         self.scope_binding.truncate(orig_bindinga);
 
         if curr_hook.len() > 0 {
-            let stmts = mem::replace(&mut n.stmts, Vec::new());
+            let stmt_count = n.stmts.len();
+            let stmts = mem::replace(&mut n.stmts, Vec::with_capacity(stmt_count));
             n.stmts.push(self.gen_hook_handle(&curr_hook));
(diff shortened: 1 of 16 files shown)


ecmascript/transforms/react/src/refresh/mod.rs
```diff
@@ -790,15 +794,15 @@ impl<C: Comments> Fold for Refresh<C> {
             return module_items;
         }
 
-        module_items.visit_with(&Invalid { span: DUMMY_SP } as _, self);
+        self.visit_module_items(&module_items, &Invalid { span: DUMMY_SP } as _);
```
kdy1, line 797: Why did you change this? Seems identical to me.

kdy1 (commented):
Mostly LGTM.
I found one tiny problem, though.

ecmascript/transforms/react/src/refresh/mod.rs
Austaras: To save one import. I believe these two are identical.

kdy1: I see. Thank you!

PR #1564, opened 2021-04-10
Add pure annotation comments for React functions + JSX

devongovett:
This ports the `/*#__PURE__*/` annotation support for JSX and React methods from Babel to SWC. They are added prior to compiled JSX elements and fragments, and to known pure React methods. This allows minifiers like Terser to remove unused expressions which improves bundle size. I originally implemented this in https://github.com/babel/babel/pull/11428 - see there for more context on why this is useful.

I was a little unsure about where to put the `add_pure_comment` helper function. I started adding it to the `Comments` type but then realized that there are several implementations and it ended up being duplicated. Is there a better place to put this?


ecmascript/transforms/react/src/pure_annotations/tests.rs
```diff
@@ -0,0 +1,309 @@
...
+            cm: source_map.clone(),
+            wr: writer,
+        };
+        emitter.emit_module(&program).unwrap();
+    }
+
+    return String::from_utf8(buf).unwrap();
+}
+
+fn run_test(input: &str, expected: &str) {
```
devongovett, line 64: I didn't find a way to get the default Tester to pass through comments correctly, so ended up duplicating some of the logic here. Is there a better way?

kdy1: Adding a field to `Tester` might work.

common/src/comments.rs
```diff
@@ -37,6 +37,8 @@ pub trait Comments {
     fn has_trailing(&self, pos: BytePos) -> bool;
     fn move_trailing(&self, from: BytePos, to: BytePos);
     fn take_trailing(&self, pos: BytePos) -> Option<Vec<Comment>>;
+
+    fn add_pure_comment(&self, pos: BytePos);
```
kdy1, line 41: Can you add a default implementation using `take_leading` and `add_leading`?
In this way the pr would be non-breaking change.

kdy1 (changes requested):
Need some changes, but mostly LGTM. Thanks!

ecmascript/transforms/react/src/pure_annotations/tests.rs
devongovett: Yeah, I tried that. There ended up being too many places to change. e.g. Tester only has one comments and source map instance, but really the actual and expected code need separate instances. Plus the `DropSpan` pass that's included in the tester breaks comments because it removes the spans that they are attached to. It just got too complicated with too many changes IMO so I bailed. Perhaps some of this logic I copied could be made reusable though. Just wondered if you had ideas or other places I didn't know about that did something similar.

kdy1: I see.

I used integration test for such tests, but it takes long time to compile.

DropSpan issue can be workarounded by using `eq_ignore_span`, which is added much later than `DropSpan`, but it may break lots of tests.

So your approach seems reasonable to me.

Thanks!

common/src/comments.rs
devongovett: Good idea. I'm still a Rust noob and didn't know this was possible. Thanks!

PR #1567, opened 2021-04-11
Don't replace missing object members with undefined

mischnic:
In general, `{}.foo` shouldn't be replaced with `undefined` because
1. the properties on the object prototype aren't undefined (`{}.constructor` itself, `{}.hasOwnProperty`, ...) and
2. hypothetically, the object's prototype could have been modified: `({}.__proto__).foo = 2; console.log({}.foo)`

Instead, only member accesses for properties that actually exists should be replaced

The fix seems to work, but I'm not so sure about the added test: reverting the change to the simplifier, this is the test output:
```
---- simplify::branch::tests::test_object_literal stdout ----
----- Actual -----
----- Actual -----
----- Actual -----
----- Actual -----
----- Actual -----
>>>>> Orig <<<<<
({}).foo
>>>>> Code <<<<<

thread 'simplify::branch::tests::test_object_literal' panicked at 'assertion failed: `(left == right)`
            ({
}).foo;
```
shouldn't it say something like `Expected: '({}).foo', Actual: 'undefined'`?

### DIFF
--- a/ecmascript/transforms/optimization/Cargo.toml
+++ b/ecmascript/transforms/optimization/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2018"
 license = "Apache-2.0/MIT"
 name = "swc_ecma_transforms_optimization"
 repository = "https://github.com/swc-project/swc.git"
-version = "0.15.2"
+version = "0.15.3"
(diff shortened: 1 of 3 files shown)


kdy1 (approved):
LGTM. Thanks!

PR #1577, opened 2021-04-15
Fix precedence of yield expression inside ternary

devongovett:
The `yield` expression has a lower precedence than the conditional operator, so needs parentheses.


kdy1 (approved):
Nice catch! Thank you.

Issue #1557, opened 2021-04-08
Wrong helper function names used


grimly:
**Describe the bug**

When using the `externalHelpers` option, the emitted code uses functions with their internal name (ie: `_typeof`) instead of their exported name from `@swc/helpers` (ie: `typeof`) making the exported code invalid.

**Input code**

```javascript
const a = typeof 0;
```

**Config**

The 3 below has this flaw. I am myself the most interested with the last.

```json
{
  "jsc": {
    "externalHelpers": true,
    "target": "es3"
  }
}
```

```json
{
  "jsc": {
    "externalHelpers": true,
    "target": "es5"
  }
}
```

```json
{
  "jsc": {
    "externalHelpers": true,
    "target": "es2015"
  }
}
```

**Observed code**

```javascript
import * as swcHelpers from "@swc/helpers";
var a = swcHelpers._typeof(0);
```

**Expected code**

```javascript
import * as swcHelpers from "@swc/helpers";
var a = swcHelpers.typeof(0);
```

**Version**
The version of @swc/core: `1.2.51`

grimly:
After further searches, it looks like this is a duplicate of #1458

PR #1578, opened 2021-04-15
@swc/helpers exports typeOf instead of typeof

grimly:
Hello,

I've made a typo in #1557 in the expected output since `@swc/helpers` exports a `typeOf` function (note the capital `O`).
Also the `_extends` does not exists. It is exposed as `extends` instead.

I can't find what source code triggers the use of `extends` ... I've tried object spread (uses `objectSpread`), Object.assign (unchanged), class inheritance doesn't seem to fit what the code was written for.

I cannot run the tests as of now to make sure nothing broke. I'll have to do that in a later day.

### DIFF
--- a/ecmascript/transforms/base/src/quote.rs
+++ b/ecmascript/transforms/base/src/quote.rs
@@ -2,14 +2,14 @@
 #[doc(hidden)]
 #[macro_export]
 macro_rules! external_name {
+    ("typeof") => {
+        "typeOf"
+    };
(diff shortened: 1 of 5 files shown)


grimly:
I discovered the `extends` function was used for destructuring therefore added a test.
I checked against a revert of `d0b7690` and indeed my new test fails.

kdy1 (approved):
Thanks!

PR #1580, opened 2021-04-16
Fix regenerator transform for yield* expressions

devongovett:
This fixes an issue in the regenerator transform, where the context variable member expression was passed to `delegateYield` rather than the name of that variable. This broke yield expressions where the result was used. See the added test for details: previously it output `[5, NaN]` instead of `[5, 8]`.

### DIFF
--- a/ecmascript/transforms/compat/src/es2015/regenerator/case.rs
+++ b/ecmascript/transforms/compat/src/es2015/regenerator/case.rs
@@ -712,7 +712,7 @@ impl CaseHandler<'_> {
                     let result = self.make_var();
                     let name = match &result {
                         Expr::Member(m) => match &*m.prop {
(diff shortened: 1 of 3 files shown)


ecmascript/transforms/compat/src/es2015/regenerator/case.rs
```diff
@@ -710,6 +710,13 @@ impl CaseHandler<'_> {
 
                 if arg.is_some() && e.delegate {
                     let result = self.make_var();
+                    let name = match &result {
+                        Expr::Member(m) => match &*m.prop {
```
kdy1, line 714: I think check for `m.computed` is required.

kdy1 (changes requested):
Thanks!

kdy1 (approved):
LGTM. Thanks!

PR #1585, opened 2021-04-17
Handle arguments references inside arrow functions and block scoped loops

devongovett:
This transforms `arguments` references inside arrow functions within function declarations/expressions, as well as when block scoped loops get transformed into functions.

```js
function test() {
  return () => arguments[0];
}
```

is transformed into:

```js
function test() {
    return (function(_arguments) {
      return _arguments[0];
    }).bind(this, arguments)
}
```

This is slightly different than how Babel transforms it, but babel also doesn't use `bind` for `this`.

As for block scoped loops:

```js
function test() {
    for (var i = 0; i < arguments.length; i++) {
      var arg = arguments[i];
      console.log((() => arg)());
    }
}
```

is now transformed to:

```js
function test() {
    var _arguments = arguments, _loop = function(i) {
        var arg = _arguments[i];
        console.log(function() {
            return arg;
        }());
    };
    for(var i = 0; i < arguments.length; i++)_loop(i);
}
```


ecmascript/transforms/compat/src/es2015/arrow.rs
```diff
@@ -112,20 +148,85 @@ impl Fold for Arrow {
...
+    arguments: Ident,
+    found: bool,
+}
+
+impl Fold for ArgumentsReplacer {
+    noop_fold_type!();
+
+    fn fold_expr(&mut self, e: Expr) -> Expr {
+        match e {
+            Expr::Ident(id) if id.sym == js_word!("arguments") => {
```
devongovett, line 198: I think ideally we'd also check whether the syntax context is "global" here, but we don't have a reference to the global mark in this transform and adding one would be a breaking API change. This is only a problem in non-strict mode, because declarations with the name `arguments` are prohibited in strict mode. However, I couldn't figure out how to get SWC to parse in anything but strict mode (or at least, the error for `'eval' and 'arguments' cannot be used as a binding identifier in strict mode` always occurred). Is this something we need to worry about?

kdy1: A breaking change is ok I think. (Transforms need some breaking changes anyway).


It's strict mode because it's module code. 
https://swc.rs/docs/usage-core#transform 
This document is outdated, but it has `isModule: false,`.

ecmascript/transforms/compat/src/es2015/block_scoping.rs
```diff
@@ -936,14 +956,18 @@ impl VisitMut for MutationHandler<'_> {
 
     fn visit_mut_function(&mut self, n: &mut Function) {
         let old = self.in_function;
+        let arguments = self.arguments.clone();
```
kdy1, line 959: This can be changed to `self.arguments.take()`.
`Option.take()` takes the value (make target variable None) and returns the original value without cloning.

ecmascript/utils/src/lib.rs
```diff
@@ -99,6 +99,49 @@ impl Visit for IdentFinder<'_> {
...
+
+    /// Don't recurse into member expression prop if not computed
+    fn visit_member_expr(&mut self, m: &MemberExpr, _: &dyn Node) {
+        m.obj.visit_with(m, self);
+        match &*m.prop {
+            Expr::Ident(_) if !m.computed => {}
+            _ => m.prop.visit_with(m, self),
+        }
+    }
+
```
kdy1, line 132: Can you add `visit_prop` which checks for `{ arguments }`?
`Prop::Shorthand(Ident)` may contain arguments, too.

kdy1 (changes requested):
Thanks!

kdy1:
May I make the changes myself?
I'm asking this because I want to publish a new version.

devongovett:
Sure, thanks!

PR #1586, opened 2021-04-17
Fix syntax context of super in class transform

devongovett:
When transforming classes, the super class is passed as a parameter to the iife wrapper for the class. However, this parameter should have a separate syntax context from the original identifier of the super class passed to the call. Currently, re-assigning the super binding has no effect within the class body because of this. For example:

```js
class Foo {}
class Test extends Foo {
  foo() {
    console.log(Foo);
  }
}
Foo = 3;
new Foo().foo();
```

should log `3` but it currently logs the original `Foo` function. This is now transformed to:

```js
let Test = function(Foo1) {
  'use strict';
  _inherits(Test, Foo1);
  function Test() {
      _classCallCheck(this, Test);
      return _possibleConstructorReturn(this, _getPrototypeOf(Test).apply(this, arguments));
  }
  _createClass(Test, [
      {
          key: 'foo',
          value: function foo() {
              console.log(Foo);
          }
      }
  ]);
  return Test;
}(Foo);
Foo = 3;
new Foo().foo();
```

note that the parameter is renamed to `Foo1` (and passed to `_inherits`), but the `console.log` still references `Foo`. This preserves the correct semantics.

### DIFF
--- a/ecmascript/transforms/compat/src/es2015.rs
+++ b/ecmascript/transforms/compat/src/es2015.rs
@@ -254,9 +254,9 @@ class B extends A {
         }]);
     return A;
 }();
-var B = function(A) {
+var B = function(A1) {
     'use strict';
-    _inherits(B, A);
+    _inherits(B, A1);
     function B(num) {
         _classCallCheck(this, B);
         var _this;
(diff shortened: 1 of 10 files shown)


kdy1 (approved):
LGTM. Thanks!
I'll modify the failing test.

PR #1597, opened 2021-04-21
feat(es/preset-env) Make android targets fall back to chrome

wbinnssmith:
Like `@babel/preset-env` does [0], fall back to using Chrome versions if a minimum Android browser version is missing from the feature requirements. It appears that Android browser has aligned its versioning around Chrome/Chromium so this should be safe. swc will also align with Babel's behavior here.

Previously, this resulted in aggressive downleveling of features like classes and generator functions if an Android target was provided to swc preset-env.

Test Plan: Added a unit test.

[0] https://github.com/babel/babel/blob/672a58660f0b15691c44582f1f3fdcdac0fa0d2f/packages/babel-helper-compilation-targets/src/utils.js#L67-L68


wbinnssmith:
I'm hoping to have the CLA signed soon. Sorry for the delay!

wbinnssmith:
Signed the CLA. Thanks for your patience!

kdy1 (approved):
LGTM. Thanks!

PR #1618, opened 2021-04-27
feat(es/transforms) simplify: do not fold do-while loops with conditional stoppers

wbinnssmith:
This prevents the simplify pass from folding do-while loops with constant conditions when the body has a conditional stoppers. Previously, this folded the loop but left behind `break;` and `continue;` statements, creating invalid JavaScript.

I'm not sure it's possible to rewrite all possible loops to behave identically as when they had these stoppers, so this skips folding entirely.

Test Plan: Added unit test.

### DIFF
--- a/ecmascript/transforms/optimization/src/simplify/branch/mod.rs
+++ b/ecmascript/transforms/optimization/src/simplify/branch/mod.rs
@@ -851,6 +851,10 @@ impl Fold for Remover {
             }
 
             Stmt::DoWhile(s) => {
+                if has_conditional_stopper(&[Stmt::DoWhile(s.clone())]) {
+                    return Stmt::DoWhile(s);
+                }
+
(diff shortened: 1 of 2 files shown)


ecmascript/transforms/optimization/src/simplify/branch/mod.rs
```diff
@@ -851,6 +851,10 @@ impl Fold for Remover {
             }
 
             Stmt::DoWhile(s) => {
+                if has_conditional_stopper(&[Stmt::DoWhile(s.clone())]) {
```
kdy1, line 854: This seems like an optimization is possible, but I'll invert the pr with https://github.com/swc-project/swc/pull/1613. Thanks!

kdy1 (approved):

Issue #1644, opened 2021-05-05
Transpiling classes into functions should add pure comments


mischnic:
**Describe the feature**

https://github.com/swc-project/swc/blob/fe107a1223bf8c2361b3ac50cc7280afa32c988c/ecmascript/transforms/compat/src/es2015/classes/mod.rs#L54-L66

should instead become (as Babel does)

```js
var Test = /*#__PURE__*/function () {
  ...
}();
```

PR #1646, opened 2021-05-05
Add pure comment when transpiling classes

mischnic:
Closes https://github.com/swc-project/swc/issues/1644

Is this the correct approach? This also requires adding a comments parameter to preset-env, is that a problem?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2018"
 license = "Apache-2.0/MIT"
 name = "swc"
 repository = "https://github.com/swc-project/swc.git"
-version = "0.16.0"
+version = "0.17.0"
 
 [lib]
 name = "swc"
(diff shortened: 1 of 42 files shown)


kdy1:
Yeah, it's a correct approach I think.

mischnic:
Not sure why so many tests are failing like this...

Bildschirmfoto 2021-05-05 um 20 27 02

kdy1:
Hmm... I'm not sure about the reason, but seems like a sequence expression with a single element is created.

mischnic:
It also seems like `test!` completely ignores comments, otherwise some tests that already contain the pure comment in the expected code would already failed
https://github.com/swc-project/swc/blob/28bb61fb8d2b2ef8940354e7fc2c5c64244ab579/ecmascript/transforms/compat/tests/es2015_classes.rs#L370-L403

I can't quite figure out how to make that happen in the test helpers: [`10bcdce` (#1646)](https://github.com/swc-project/swc/pull/1646/commits/10bcdce633874308e314019a9f03edceb88757db)
```
mismatched types
expected enum `std::option::Option<&dyn Comments>`
   found enum `std::option::Option<Arc<C>>`
```

kdy1:
https://github.com/swc-project/swc/pull/1564/files#diff-9fa4b09c003b4e10b0c37f6c503310cd71dce5869e4d38b4e41e4e28e77ada7fR94-R309

I think this approach can be used for classes, too.

kdy1:
I can take on this if you want.

mischnic:
> I can take on this if you want.

Sure, thanks

PR #1663, opened 2021-05-09
fix(es/parser/typescript): allow `override` with `static`

g-plane:
Previously, TypeScript didn't allow `override` with `static`, but it seems that it was changed later: [TypeScript playground](https://www.typescriptlang.org/play?ts=4.3.0-dev.20210508#code/MYGwhgzhAEBC0G8BQ1oQC5nQS2NdAFAJSIC+S5SokMAwgIzQCmAHukwHYAmM8yqAewBuTAE6jsXJmkw48hEgnKVqUaLQBMzNpx5xEKGVlzRhYiVPzEyFIA).

Also, this PR enforces order between `override` and `static`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2018"
 license = "Apache-2.0/MIT"
 name = "swc"
 repository = "https://github.com/swc-project/swc.git"
-version = "0.17.0"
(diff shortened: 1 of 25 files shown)


kdy1 (approved):
Thanks! LGTM.

PR #1712, opened 2021-05-18
fix(swc): Fix bugs

kdy1:
swc_ecma_preset_env:
 - [x] Update compat data. (Close #1704)
 - [x] Support `export-namespace-from`.

swc_ecma_transforms_compat:
 - [x] Use proper argument for private field accesses in a private method. (Closes #1711)

Issue #1604, opened 2021-04-24
Direct usage of `core-js-compat`


zloirock:
As I could see, copy of `core-js-compat` data used in `swc` version of `preset-env` is critically obsolete - 1.5 year. Each new version of `core-js-compat` contains updates engines support data,
```js
// .swcrc:
{
  "env": {
    "targets": {
      "safari": "14"
    },
    "mode": "entry",
    "coreJs": "3.11"
  }
}
// input.js:
import 'core-js/stable';
// output.js:
import   "core-js/modules/es.typed-array.int32-array";
import   "core-js/modules/es.typed-array.from";
import   "core-js/modules/web.immediate";
import   "core-js/modules/web.url-search-params";
import   "core-js/modules/es.typed-array.uint32-array";
import   "core-js/modules/web.url";
import   "core-js/modules/es.typed-array.float64-array";
import   "core-js/modules/es.typed-array.uint8-array";
import   "core-js/modules/es.typed-array.int16-array";
import   "core-js/modules/es.typed-array.int8-array";
import   "core-js/modules/es.string.replace";
import   "core-js/modules/es.typed-array.uint8-clamped-array";
import   "core-js/modules/es.typed-array.float32-array";
import   "core-js/modules/web.url.to-json";
import   "core-js/modules/es.typed-array.of";
import   "core-js/modules/es.typed-array.uint16-array";
// expected:
import   "core-js/modules/web.immediate";
```
`core-js-compat` contains information about new modules and entry points. Without this information impossible to use it correctly - new modules will not be injected even if they are required.
```js
// .swcrc:
{
  "env": {
    "targets": {
      "safari": "13"
    },
    "mode": "entry",
    "coreJs": "3.11"
  }
}
// input.js:
import 'core-js/stable/promise';
// output.js:
import   "core-js/modules/es.promise.finally";
import   "core-js/modules/web.dom-collections.iterator";
// expected:
import   "core-js/modules/es.aggregate-error";
import   "core-js/modules/es.promise.any";
import   "core-js/modules/es.promise.finally";
import   "core-js/modules/web.dom-collections.iterator";
```
Compat data should be updated even without `swc` updates for making possible to use actual `core-js` versions properly, so the only way to do it - usage of `core-js-compat` as a dependency.

zloirock:
@kdy1 please, take a look at this since at this moment is used critically obsolete compact data from `core-js@3.6` - 2019.12.19 - 2 years old.

Austaras:
[...]

PR #1719, opened 2021-05-20
fix(swc): Fix bugs

kdy1:
swc_ecma_preset_env:
 - [x] Update `core-js-compat`. (https://github.com/swc-project/swc/issues/1712#issuecomment-844959483)


zloirock:
It's good as a temporal fix, direct usage of `core-js-compat` could prevent possible problems in the future (#1604).

PR #1758, opened 2021-05-27
Build FreeBSD binary

Brooooooklyn:


kdy1:
Thanks a lot!

kdy1:
cc @Brooooooklyn 
The build is failing with a rsync error.

https://github.com/swc-project/swc/runs/2703996836?check_suite_focus=true

Brooooooklyn:
Oh it's never happened before, please retry the pipeline and see if this was a random error

kdy1:
Oh, it was a random error.
I actually tried again before cc, but it failed. Anyway, thank you!

Issue #1759, opened 2021-05-27
TS1031 Syntax Error does not work


Shinyaigeek:
**Describe the bug**

`TS1031: declare modifier cannot appear on a class element` does not work on current master branch.

```typescript
class Hoge {
    declare hoge: string;
    declare aaa(): string;
    constructor() {
        this.hoge = "asdf";
    }
}
```

Syntax Error TS1031 should occur in compiling above code because class method declaration with declare keyword is not correct in typescript. But swc on the current master branch does not throw TS1031 Syntax Error.

Typescript Playground URL: https://www.typescriptlang.org/play?ssl=7&ssc=2&pln=1&pc=1#code/MYGwhgzhAEASD2BzAptA3gKGt6ATZoYATqgBZLIBc0EALkQJYB2iA3FjvoSdGHwBQBKanUYt2OaMHhNRAV2C14RIeg6TstUgwgA6cimgBeaACJIuAGamJOAL4YHQA 

**Input code**

```typescript
class Hoge {
    declare hoge: string;
    declare aaa(): string;
    constructor() {
        this.hoge = "asdf";
    }
}
```

output(compilation is successfully done, but compilation should be aborted with TS1031 SyntaxError )
```javascript
class Hoge {
    constructor(){
        this.hoge = "asdf";
    }
}
```

**Config**

```json
{
        "jsc": {
            "target": "es2020"
        }
}
```

**Expected behavior**

swc should throw TS1031 SyntaxError

**Version**
The version of @swc/core: master branch (2021/5/28) 

**Additional context**

Can I pick this issue?

updated: sorry I already picked 🙇

PR #1760, opened 2021-05-27
Fix: no response TS1031 SyntaxError

Shinyaigeek:
## Motivation

Fix: #1759 

I fixed the error that TS1031 SyntaxError does not occur in compiling invalid typescript code with declaration class method with declare statement.

if something wrong (incorrect approach or not enough test case...), please feel free to close this PR or point it out 🙏


kdy1 (approved):
Thanks!

PR #1763, opened 2021-05-28
chore(): update ts and remove unused deps

chyzwar:
Update typescript to 4.3.2. remove core-js and axios as these look unused?


kdy1:
core-js and axios is used from tests.
Those are used to test common js support of `spack`.

chyzwar:
ok, I see but why not core-js 3 ?

kdy1:
The dependency on corejs@2 does not have any serious meaning.

PR #1766, opened 2021-05-29

kdy1:
swc_ecma_minifier:
 - Implement more rules.

swc_ecma_transforms_base:
 - `resolver`: Fix syntax context of catch block.

swc_ecma_transforms_optimization:
 - `expr_simplifier`: Fix a bug related to `this` of call expressions.
 - `dead_branch_remover`: Don't reduce switch cases if test is now known.
 - `dead_branch_remover`: Don't break `&&`.


fabiosantoscode:
Hey @kdy1 I'm thinking of submitting a reduce_vars implementation. Let me know if I can send a PR to this repo's master or still use your fork

kdy1:
@fabiosantoscode I think the master would be better. Thanks a lot!

ecmascript/transforms/optimization/src/simplify/expr/mod.rs
```diff
@@ -1344,7 +1344,7 @@ impl Fold for SimplifyExpr {
                         ExprOrSuper::Expr(seq.exprs.into_iter().next().unwrap().fold_with(self))
                     } else {
                         match seq.exprs.get(0).map(|v| &**v) {
-                            Some(Expr::Lit(Lit::Num(..))) => {}
+                            Some(Expr::Lit(..) | Expr::Ident(..)) => {}
```
avindra, line 1347: Getting a build error on this line
```
    Compiling swc_ecma_transforms_proposal v0.24.0
 error[E0658]: or-patterns syntax is experimental
     --> /home/abuild/rpmbuild/BUILD/deno-1.11.5/vendor/swc_ecma_transforms_optimization/src/simplify/expr/mod.rs:1347:34
      |
 1347 | ...                   Some(Expr::Lit(..) | Expr::Ident(..)) => {}
      |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      |
      = note: see issue #54883 <https://github.com/rust-lang/rust/issues/54883> for more information
      = help: add `#![feature(or_patterns)]` to the crate attributes to enable
```

It was seen when trying to build Deno after upgrading `swc_bundler` from `0.40.0` to `0.44.0`  https://github.com/denoland/deno/pull/11197 


I also found that the Or pattern syntax being referred here is experimental per  https://github.com/rust-lang/rfcs/blob/master/text/2535-or-patterns.md

Should the crate be annotated as described, minimum Rust compiler bumped, or something else?

kdy1: `or_patterns` is already stabilized and that's why the crate does not have `#![feature(or_patterns)]`.
I didn't think about it carefully, but if required, I can modify it to allow using it with an older version of rustc.

avindra: Gotcha. No worries. I saw it was stabilized in [rustc 1.53.0](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1530-2021-06-17). 

openSUSE (what I'm using) is still waiting for [1.53.0](https://build.opensuse.org/request/show/902676). So I'll just try again later when rust is upgraded.

Thanks for your reply.

Issue #1764, opened 2021-05-28
When using minify in es2016 or higher target, the space before extends is deleted


haydnhkim:
**Describe the bug**
The problem arises when I use it with minify option by setting the target of jsc to es2016~es2020.
JavaScript cannot be executed because the class name and `extends` keyword are stuck.

**Input code**

```js
class Hello {
  constructor() {
    console.log('hello');
  }
}

class Hi extends Hello {
  constructor() {
    console.log('hi');
  }
}
```

**Output code**

```js
class Hello{constructor(){console.log('hello');}}class Hiextends Hello {constructor(){console.log('hi');}}
                                                       ^^^^^^^^^
```

**Config**

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript"
    },
    "target": "es2016"
  },
  "minify": true
}
```

**Version**
The version of @swc/core: 1.2.58

PR #1767, opened 2021-05-29
Fix: removed space before extends keyword on class declaration

Shinyaigeek:
## Motivation

Fix: #1764 

I fixed the error that with minify option and es2016 ~ 2020 target option `swc` outputs invalid javascript code due to removed space before `extends` keyword on class declaration.

if something wrong (incorrect approach or not enough test case...), please feel free to close this PR or point it out 🙏

### DIFF
--- a/ecmascript/codegen/src/lib.rs
+++ b/ecmascript/codegen/src/lib.rs
@@ -835,6 +835,7 @@ impl<'a> Emitter<'a> {
     #[emitter]
     fn emit_class_trailing(&mut self, node: &Class) -> Result {
         if node.super_class.is_some() {
+            space!();
             keyword!("extends");
             space!();
             emit!(node.super_class);
--- a/ecmascript/codegen/src/lib.rs
+++ b/ecmascript/codegen/src/lib.rs
@@ -823,10 +823,6 @@ impl<'a> Emitter<'a> {
             space!();
             emit!(i);
             emit!(node.class.type_params);
-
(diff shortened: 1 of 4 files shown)


kdy1 (approved):
Thanks!

Shinyaigeek:
Oops, some tests is failed :bow:
I am investigating why.

kdy1:
UPDATE=1 cargo test --all

will fix it I think.
It's bit slow, but it updates all specs of all tests.

ecmascript/codegen/src/lib.rs
```diff
@@ -835,6 +835,7 @@ impl<'a> Emitter<'a> {
     #[emitter]
     fn emit_class_trailing(&mut self, node: &Class) -> Result {
         if node.super_class.is_some() {
+            space!();
```
Shinyaigeek, line 838: I got my approach was wrong. This updated code will insert space between `class ident` and `extends keyword`, but when minify option is false, space already exists after `class ident`, so when minify option is false, two spaces are inserted between  `class ident` and `extends keyword`.  I will fix this.

Shinyaigeek:
> UPDATE=1 cargo test --all

@kdy1 OK, I will try it!! 👌  thank you!

Shinyaigeek:
@kdy1 I fix the approach(eventually I did not touch tests), please check it when you are free 🙇

PR #1770, opened 2021-05-31
Expose load_transformed() and TransformedModule.

tmpfs:
To allow external programs to use the bundler transform phase to
generate a module graph.

### DIFF
--- a/bundler/src/bundler/export.rs
+++ b/bundler/src/bundler/export.rs
@@ -45,7 +45,7 @@ pub(super) struct RawExports {
 }
 
 #[derive(Debug, Default)]
-pub(crate) struct Exports {
+pub struct Exports {
     pub items: Vec<Specifier>,
     pub reexports: Vec<(Source, Vec<Specifier>)>,
 }
--- a/bundler/src/bundler/helpers/mod.rs
+++ b/bundler/src/bundler/helpers/mod.rs
@@ -6,7 +6,7 @@ use swc_ecma_parser::{lexer::Lexer, Parser, StringInput};
 use swc_ecma_utils::{drop_span, prepend_stmts};
 
 #[derive(Debug, Default)]
-pub(crate) struct Helpers {
+pub struct Helpers {
     /// `__spack_require__`
     pub require: AtomicBool,
 }
--- a/bundler/src/bundler/load.rs
+++ b/bundler/src/bundler/load.rs
@@ -21,7 +21,7 @@ use swc_ecma_transforms::resolver_with_mark;
 use swc_ecma_visit::{noop_visit_type, FoldWith, Node, Visit, VisitWith};
 /// Module after applying transformations.
 #[derive(Debug, Clone)]
-pub(crate) struct TransformedModule {
+pub struct TransformedModule {
     pub id: ModuleId,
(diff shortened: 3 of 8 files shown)


tmpfs:
Also includes a small fix for compiling against the `spack` module here: https://github.com/swc-project/swc/pull/1770/files#diff-887764dffae8a5dff11fe5af51313928fda3627c6bfa1a3d7fb28c51017917f5.

kdy1:
I think this is more suitable for `swc_ecma_loader`. (at /ecmascript/loader/)
The bundler can depend on it, and other crates can depend on it in the future.
Are you ok with it?

tmpfs:
@kdy1, You mean move `TransformedModule` and dependent structs to `ecmascript/loader`? 

Sure happy to do that if my program can still compile and run :+1:

kdy1:
Yes. I meant it and thank you in advance!

tmpfs:
The problem with moving to `ecmascript/loader` is that there are issues resolving the `Helpers`.

These are the compiler errors I have after attempting to move the structs:

```
   Compiling swc_ecma_loader v0.6.0 (/home/muji/git/consensys/swc/ecmascript/loader)
error: cannot find derive macro `Is` in this scope
  --> ecmascript/loader/src/module.rs:48:24
   |
48 | #[derive(Debug, Clone, Is)]
   |                        ^^

error[E0433]: failed to resolve: use of undeclared crate or module `swc_ecma_transforms`
  --> ecmascript/loader/src/module.rs:23:26
   |
23 |     pub swc_helpers: Lrc<swc_ecma_transforms::helpers::Helpers>,
   |                          ^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `swc_ecma_transforms`

error[E0412]: cannot find type `Helpers` in this scope
  --> ecmascript/loader/src/module.rs:21:22
   |
10 | pub struct TransformedModule {
   |                             - help: you might be missing a type parameter: `<Helpers>`
...
21 |     pub helpers: Lrc<Helpers>,
   |                      ^^^^^^^ not found in this scope

error: aborting due to 3 previous errors
```

I cannot depend upon `swc_ecma_transforms` in `swc_ecma_loader` as it creates a cyclic package dependency and the other `Helpers` in `swc_bundler` seem like they should be in that package to me.

So in order to do this refactoring we would need to decouple both `Helpers` from `TransformedModule` - maybe with a wrapper type in `swc_bundler`?

But obviously this is quite a lot of work. What do you think?

kdy1:
The problem of `swc_ecma_transforms` can be workarounded by depending on `swc_ecam_transforms_base`, while the helper of bundler going to other crate seems wrong to me.

But I think bundler should not expose such apis. I will think about this bit more.

tmpfs:
It may be that that we should move the helpers to a separate module to sidestep the issue a bit.

I also noticed the `Hook` struct would be useful for me to access when I create a `Bundler` (at the moment I have copied the code over). Maybe we could put that in helpers too?

kdy1:
The behavior of the loader implementation of bundler is too opinionated for public apis.
It modifies span hygiene data in a way that can be handled by the bundler afterward.
It's an implementation detail and can be changed anytime if I found a better way.

PR #1809, opened 2021-06-09
build(node-swc): Support lower version glibc linux on non-x64 arch

Brooooooklyn:
Close https://github.com/Brooooooklyn/swc-node/issues/480


kdy1 (approved):
Thanks!

PR #1815, opened 2021-06-11
fix(es): Fix UB

### LINKED ISSUE #1782: The conversion of React props value, unicode character is unnecessary
Here is the swcrc config:
```
{
	"jsc": {
		"parser": {
			"syntax": "typescript",
			"tsx": true
		}
	}
}
```

This is the source file:
```js
import React from 'react';

export const HelloWorld = () => {
  return (<div title="您好SWC"></div>)
}

```

Result：

```js
import React from 'react';
export var HelloWorld = function() {
    return(/*#__PURE__*/ React.createElement("div", {
        title: "\\u60a8\\u597dSWC"
    }));
};

```

As we can see all the chinese character was converted to unicode. but that is not what we want.

kdy1:
swc_ecma_parser:
 - [x] Remove UB.

swc_ecma_codegen:
 - [x] Remove codes related to the UB of the parser. (Closes #1782)


Austaras:
Also if it seems now `swc` will transform
```ts
const a = '\n'
```
to
```
const a = '\\n'
```

kdy1:
> Also if it seems now `swc` will transform
> 
> ```ts
> const a = '\n'
> ```
> 
> to
> 
> ```
> const a = '\\n'
> ```

Really? I think `\n` will be handled with other codes.

Austaras:
Sorry, seems like I got it wrong

PR #1818, opened 2021-06-11
Fix the incorrect ExportDefaultDeclaration span.lo for `export default function` case

vovacodes:
Am I missing something or is it a bug? @kdy1

### DIFF
--- a/ecmascript/parser/src/parser/tests.rs
+++ b/ecmascript/parser/src/parser/tests.rs
@@ -1,4 +1,5 @@
 use crate::test_parser;
+use swc_common::BytePos;
 use swc_ecma_ast::*;
 
 fn program(src: &'static str) -> Program {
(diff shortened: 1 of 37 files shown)


ecmascript/parser/src/parser/tests.rs
```diff
@@ -68,3 +69,26 @@ fn parse_program_module_error_01() {
         ",
     );
 }
+
+#[test]
+fn parse_module_export_named_span() {
+    let m = module("export function foo() {}");
+    if let ModuleItem::ModuleDecl(ModuleDecl::ExportDecl(ExportDecl { span, .. })) = &m.body[0] {
+        assert_eq!(span.lo, BytePos(0));
```
vovacodes, line 77: This reports the correct `span.lo`

ecmascript/parser/src/parser/tests.rs
```diff
@@ -68,3 +69,26 @@ fn parse_program_module_error_01() {
...
+}
+
+#[test]
+fn parse_module_export_default_span() {
+    let m = module("export default function foo() {}");
+    if let ModuleItem::ModuleDecl(ModuleDecl::ExportDefaultDecl(ExportDefaultDecl {
+        span, ..
+    })) = &m.body[0]
+    {
+        assert_eq!(span.lo, BytePos(0));
```
vovacodes, line 90: This tells that `span.lo` is `15` instead of `0`

kdy1:
I'm not sure, I'll investigate further when I have some time.

vovacodes:
@kdy1 I added the fix fix, could you please take a look when you have a moment, this issue blocks us from shipping one of the features. 🙇‍♂️

kdy1:
Ah, thanks!
I missed this issue. I always use milestone page for looking tasks and didn't check PR tab.

kdy1 (approved):
Thanks!

vovacodes:
Hm @kdy1 those two tests that are failing, I don't think they are correct are they? It's `647e21f8f157c338.js` (parsing `\u2029`) and `8af69d8f15295ed2.js` (parsing `\u2029`)  that are supposed to fail according to the tests but I don't think that's correct.

kdy1:
If I remember correctly, `\u2029`was wrong for javascript, but spec changed.
Can you please add those to https://github.com/swc-project/swc/blob/33f2ab2d795ee8fbe935f535cf5568a252af1fcc/ecmascript/parser/tests/test262.rs#L101 ?

vovacodes:
Ah I see I accidentally committed some changes to the `tests/test262-parser` git submodule. Let me try and revert those

vovacodes:
@kdy1, I finally resolved that submodule issue, I think it's ready now

PR #1829, opened 2021-06-17
Fix #1786: use `_interopRequireWildcard` for import when required by an export

### LINKED ISSUE #1786: swc uses `_interopRequireDefault` when it should use `_interopRequireWildcard`
EDIT this description is wrong; see https://github.com/swc-project/swc/issues/1786#issuecomment-861420915

**Describe the bug**

swc does not inline `_interopRequireDefault` when it is needed and when `externalHelpers` is false.  Instead it inlines `_interopRequireWildcard`, although that function is not used.

**Input code**

I can reproduce the bug using SWC's JS API

```
swc.transformSync(`
  import Foo from "bar";
  export {Foo} from "bar";
`, {
  module: {type: 'commonjs'}
})
```

The output includes the `_interopRequireWildcard` helper, but omits `_interopRequireDefault`.

```
"use strict";
    value: true
});
var _bar = _interopRequireDefault(require("bar"));
    if (obj && obj.__esModule) {
        return obj;
    } else {
        var newObj = {
        };
        if (obj != null) {
            for(var key in obj){
                    };
                    if (desc.get || desc.set) {
                    } else {
                        newObj[key] = obj[key];
                    }
                }
            }
        }
        newObj.default = obj;
        return newObj;
    }
}
Object.defineProperty(exports, "Foo", {
    enumerable: true,
    get: function() {
        return _bar.Foo;
    }
});
```

**Config**

n/a

**Expected behavior**

swc inlines the `_interopRequireDefault` helper function when it is used.  

**Version**
The version of @swc/wasm: 1.2.58

**Additional context**

Possibly related to #138

### LINKED ISSUE #1790: Doesn't export the default value when `interopRequireDefault` is applied
**Describe the bug**

swc exports the incorrect value when the `interopRequireDefault` helper is applied to a separate import from the same library

**Input code**

Reproducible using the API

```js
const swc = require(`@swc/wasm`);
console.log(swc.transformSync(`
import semver from 'semver';

export {SemVer} from 'semver';

export function foo() {
  return new semver.SemVer('');
}
`, {
  module: {type: `commonjs`},
  jsc: {
    externalHelpers: true,
  },
}).code);
```

The output exports `_semver.SemVer` instead of `_semver.default.SemVer`

```js
"use strict";
    value: true
});
exports.foo = foo;
var swcHelpers = require("@swc/helpers");
var _semver = swcHelpers.interopRequireDefault(require("semver"));
Object.defineProperty(exports, "SemVer", {
    enumerable: true,
    get: function() {
        return _semver.SemVer;
    }
});
function foo() {
    return new _semver.default.SemVer('');
}
```

**Config**
n/a

**Expected behavior**

```patch
diff --git a/out.js b/out.js
index 3b5596341..8b5e63f09 100644
--- a/out.js
+++ b/out.js
@@ -8,7 +8,7 @@ var _semver = swcHelpers.interopRequireDefault(require("semver"));
 Object.defineProperty(exports, "SemVer", {
     enumerable: true,
     get: function() {
-        return _semver.SemVer;
+        return _semver.default.SemVer;
     }
 });
 function foo() {

```

**Version**
The version of @swc/wasm: 1.2.58

**Additional context**
https://github.com/yarnpkg/berry/pull/2228

cspotcode:
Based on advice in #1826 
Fixes #1786
Might also fix #1790 

The `ImportAnalyzer` and `CommonJs` folders need to share a `Scope`.  To facilitate this, I refactored each module folder into two `Fold` structs, for example, `CommonJs` and `CommonJsWorker`.  The worker is created in `fold_module`, given a `&mut Scope` to do its work, and then dropped afterward.  This let me avoid adding `borrow()` and `deref()` calls all over the place.


cspotcode:
I have fixed the merge and the test; everything is passing now.

src/builder.rs
```diff
@@ -177,18 +180,20 @@ impl<'a, 'b, P: swc_ecma_visit::Fold> PassBuilder<'a, 'b, P> {
             ))
         };
 
+        let scope_instance: Scope = Default::default();
```
kdy1, line 183: Can you changed this to `Scope::default()`?

ecmascript/transforms/module/tests/common_js.rs
```diff
@@ -4782,6 +4787,60 @@ test!(
     console.log('aFunc: ', (0, _testlibrary).aFunc(1, 2));
     "
 );
+test!(
+    syntax(),
+    |_| {
+        let scope_instance: Scope = Default::default();
```
kdy1, line 4793: Please change this to `Scope::default()`

ecmascript/transforms/module/src/util.rs
```diff
@@ -78,7 +78,7 @@ impl Default for Lazy {
 }
 
 #[derive(Clone, Default)]
-pub(super) struct Scope {
```
kdy1, line 81: If possible, please hide all fields with `pub(crate)` (on fields)

ecmascript/transforms/module/src/umd.rs
```diff
@@ -35,13 +37,40 @@ struct Umd {
...
+            config: &mut self.config,
+            scope,
+            exports: &mut self.exports,
+        };
+        umd.fold_module(module)
+    }
+}
+
+struct UmdWorker<'a> {
+    cm: &'a mut Lrc<SourceMap>,
```
kdy1, line 63: &mut is not necessary in `&mut Arc<T>` / `&mut Rc<T>`

ecmascript/transforms/module/src/amd.rs
```diff
@@ -22,18 +24,40 @@ pub fn amd(config: Config) -> impl Fold {
...
-
 struct Amd {
     config: Config,
     in_top_level: bool,
-    scope: Scope,
+    scope: RefCell<Scope>,
     exports: Exports,
 }
 
+impl Fold for Amd {
```
kdy1, line 38: Can you merge `Amd` and `AmdWorker`?
Creating a new visitor is  really bad for binary size.

ecmascript/transforms/module/src/common_js.rs
```diff
@@ -17,25 +20,56 @@ use swc_ecma_utils::IsDirective;
...
     config: Config,
-    scope: Scope,
+    scope: Rc<RefCell<Scope>>,
     in_top_level: bool,
 }
 
 impl Fold for CommonJs {
     noop_fold_type!();
 
+    fn fold_module(&mut self, module: Module) -> Module {
```
kdy1, line 50: Can you merge `CommonJs` and `CommonJsWorker`?
Creating a new visitor is  really bad for binary size.

ecmascript/transforms/module/src/import_analysis.rs
```diff
@@ -7,25 +10,36 @@ use swc_ecma_visit::noop_visit_type;
...
-    scope: Scope,
+pub struct ImportAnalyzerWorker<'a> {
+    scope: &'a mut Scope,
 }
 
-impl Fold for ImportAnalyzer {
+impl Fold for ImportAnalyzerWorker<'_> {
     noop_fold_type!();
 
     fn fold_module(&mut self, module: Module) -> Module {
```
kdy1, line 38: Can you merge visitors?

ecmascript/transforms/module/src/umd.rs
```diff
@@ -35,13 +37,40 @@ struct Umd {
...
     config: BuiltConfig,
-    scope: Scope,
+    scope: RefCell<Scope>,
     exports: Exports,
 }
 
 impl Fold for Umd {
     noop_fold_type!();
 
+    fn fold_module(&mut self, module: Module) -> Module {
```
kdy1, line 47: Can you merge visitors?

kdy1 (changes requested):
Thanks!

cspotcode:
Thanks, I should be able to push the requested changes soon.

Regarding merging the visitors, it seemed very inconvenient to constantly
borrow and dereference from the `RC<RefCell<>>`.  This is why I opted for
using two visitors.

Is there a good way to avoid this?  Can I change `chain()` to accept a
smaller visitor factory, and then teach chain to invoke the factory and use
the value it returns?  This would avoid the messy borrowing refactor and
also avoid creating new visitors.

kdy1:
> Is there a good way to avoid this?

No. As far as I know, runtime-borrows are cumbersome because they should be (in rust).

> Can I change `chain()` to accept a smaller visitor factory, and then teach chain to invoke the factory and use the value it returns?  This would avoid the messy borrowing refactor and also avoid creating new visitors.

I think you should create an alternative maco like `chain_factory!` instead of modifying it.

But this is a complex problem. You want to pass down `RefMut` or `&mut` instead of `Rc<RefCell>`, right?
If then, it would require creating a new visitor from the macro and it becomes even worse.
I already tried something similar.

This is because you can't pass down `RefMut` to the second one until the first one finishes. So you can't create a chained visitor with shared data.

kdy1:
I think this will conflict with https://github.com/swc-project/swc/pull/1834
I'll postpone #1834 if you are going to resolve issues soon.

cspotcode:
Yes I should be able to resolve these issues this weekend, today or
tomorrow.

kdy1:
Great, thanks!

cspotcode:
I was so obsessed with learning Rust, I attempted something slightly different than merging the visitors.  I implemented a `FoldFactory` which can give you a `dyn Fold` when you need to run the visitor.  For backwards-compatibility, every `Fold` also implements `FoldFactory` by returning a reference to itself.  Chains can be a mix of `FoldFactory` and `Fold`.  The chain logic has been update to expect `FoldFactory` and to create `Fold`s when the visitor runs, not sooner.

This allows `CommonJsFactory` to be a `FoldFactory`, not a full visitor.  When it runs, it creates a `CommonJs` visitor with a mutable borrow of `Scope`.  This borrow expires when the visitor finishes, but for the entire duration, `CommonJs` does not need to do extra work to dereference or borrow `Scope`.

This change adds `dyn Fold`, and I'm not sure if that will be bad for binary size.  Is `dyn Fold` adding another visitor?

I noticed that the sourcemap tests require node 12 or older.  Node 14 and 16 changed their stack trace output.  Should I make a PR adding this requirement to CONTRIBUTING.md?

cspotcode:
Sorry, I forgot to change the other `Fold` visitors, I'll push that code soon...

kdy1:
I'm not sure about such change.
Can you compare the size and performance of master and your branch?

kdy1:
I looked at the code of amd pass, and it's two visitor and it's not related to `dyn Fold` nor `FoldFactory`.
Did you patched amd pass?

kdy1:
I profiled the size on mac os.


master:
```
21550312  6 22 14:48 ./target/release/libnode.dylib
```
It's 21045 KB

pr:
```
58555296  6 22 15:00 ./target/release/libnode.dylib
```
It's 57182 KB


More than triple.


I also found that this pr makes compilation much slower (14m -> 27m)
Seems reasonable as there are much more code to generate.

cspotcode:
Yeah I accidentally forgot to change some of the visitors, I'm still working on that.

ecmascript/transforms/module/src/util.rs
```diff
@@ -443,15 +438,6 @@ impl Scope {
...
     ) -> Expr {
-        macro_rules! entry {
-            ($i:expr) => {
-                folder
-                    .scope_mut()
-                    .exported_vars
-                    .entry(($i.sym.clone(), $i.span.ctxt()))
-            };
-        }
-
```
cspotcode, line 454: This macro began raising an error that I did not understand, and I could not figure out how to fix it unless I stopped using the macro.  I think the error was caused by `folder.scope_mut()` returning a `RefMut<Scope>`.

ecmascript/transforms/module/src/util.rs
```diff
@@ -20,8 +21,8 @@ use swc_ecma_visit::{Fold, FoldWith, VisitWith};
 
 pub(super) trait ModulePass: Fold {
     fn config(&self) -> &Config;
-    fn scope(&self) -> &Scope;
-    fn scope_mut(&mut self) -> &mut Scope;
```
cspotcode, line 24: As far as I can tell, this change is necessary because we cannot return a `&Scope` without also passing ownership of the `Ref` that it comes from.

cspotcode:
I've removed all the `*Worker`s and `FoldFactory`.  `ImportAnalyzer` and `CommonJs` borrow and deref the scope as necessary.  Tests are passing locally, so I think this is ready for another review.

ecmascript/transforms/module/src/amd.rs
```diff
@@ -300,10 +305,13 @@ impl Fold for Amd {
...
                         // export { foo } from 'foo';
                         ModuleDecl::ExportNamed(export) => {
+                            let mut scope_ref_mut = self.scope.borrow_mut();
+                            let scope = &mut *scope_ref_mut;
                             let imported = export.src.clone().map(|src| {
-                                self.scope
-                                    .import_to_export(&src, !export.specifiers.is_empty())
+                                scope.import_to_export(&src, !export.specifiers.is_empty())
                             });
+                            drop(scope);
```
kdy1, line 313: Why is this required?

ecmascript/transforms/module/src/amd.rs
```diff
@@ -320,33 +328,32 @@ impl Fold for Amd {
...
                                     if is_import_default {
-                                        self.scope
+                                        scope
                                             .import_types
                                             .entry(src.value.clone())
                                             .or_insert(false);
                                     }
                                 }
-
+                                drop(scope);
```
kdy1, line 355: Ditto

ecmascript/transforms/module/src/common_js.rs
```diff
@@ -116,14 +124,18 @@ impl Fold for CommonJs {
                             ref specifiers,
                             ..
                         })) => {
-                            self.scope.import_to_export(&src, !specifiers.is_empty());
+                            scope.import_to_export(&src, !specifiers.is_empty());
                         }
                         _ => {}
                     }
-
+                    drop(scope);
```
kdy1, line 131: Same as above

ecmascript/transforms/module/src/common_js.rs
```diff
@@ -315,13 +329,15 @@ impl Fold for CommonJs {
...
+                            let mut scope = self.scope.borrow_mut();
                             let imported = export.src.clone().map(|src| {
-                                self.scope
-                                    .import_to_export(&src, !export.specifiers.is_empty())
+                                scope.import_to_export(&src, !export.specifiers.is_empty())
                             });
 
                             stmts.reserve(export.specifiers.len());
 
+                            drop(scope);
```
kdy1, line 339: Same as above

kdy1 (commented):
Mostly LGMT. Thanks a lot!

ecmascript/transforms/module/src/amd.rs
cspotcode: Without it I get this error:

```
error[E0502]: cannot borrow `*self` as mutable because it is also borrowed as immutable
   --> ecmascript/transforms/module/src/amd.rs:361:54
    |
308 |                             let mut scope_ref_mut = self.scope.borrow_mut();
    |                                                     ---------- immutable borrow occurs here
...
361 |                                     None => Box::new(Expr::Ident(orig.clone()).fold_with(self)),
    |                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ mutable borrow occurs here
...
409 |                         }
    |                         - immutable borrow might be used here, when `scope_ref_mut` is dropped and runs the destructor for type `RefMut<'_, Scope>`
```

As far as I understand, `scope` and `scope_ref_mut` constitute an immutable borrow of `self` which will not be dropped till near the bottom of this function.  This conflicts with mutable borrows made by `fold_with(self)`.  But if I drop `scope` and `scope_ref_mut` earlier, there is no conflict.

kdy1 (approved):
LGTM. Thanks!

PR #1831, opened 2021-06-18
fix(es/codegen): Fix newlines in tagged template literals

### LINKED ISSUE #1791: Functions breaks when given multilined tagged templates.
**Describe the bug**

Using a function that takes a tagged template string throws a runtime error. This is very likely due to SWC's minimization process where a regular string quote has a line break in it.

**Input code**

```typescript
const f = (strs : TemplateStringsArray, args : any []) : null =>
	null

f `
`
// Uncaught SyntaxError: "" string literal contains an unescaped line break
```

**Output code**

```javascript
// Is two-lined code. Line breaks occurs inside of "...".
function _taggedTemplateLiteral(strings,raw){if(!raw){raw=strings.slice(0);}return Object.freeze(Object.defineProperties(strings,{raw:{value:Object.freeze(raw)}}));}function _templateObject(){var data=_taggedTemplateLiteral(["
\n"]);_templateObject=function _templateObject(){return data;};return data;}var f=function(strs,args){return null;};f(_templateObject());
```

**Config**

```json
{
	"include"         : ["./ts/*.ts"],
	"compilerOptions" :
	{
		"lib"                          : ["ESNext", "DOM"],
		"target"                       : "ES6",
		"outDir"                       : "./js/",
		"alwaysStrict"                 : true,
		"noImplicitAny"                : true,
		"removeComments"               : true,
		"preserveConstEnums"           : true,
		"watch"                        : true,
		"strictNullChecks"             : true,
		"strictFunctionTypes"          : true,
		"strictPropertyInitialization" : true,
		"strictBindCallApply"          : true,
		"strict"                       : true,
		"pretty"                       : true,
		"noImplicitReturns"            : true,
		"noEmit"                       : true
	}
}
```

**Expected behavior**
Running `f` above should just return `null` with no errors attached.

**Version**
The version of @swc/core: 1.2.59

kdy1:
swc_ecma_codegen:
 - Fix codegen of tagged template literals. (Closes #1791)

### DIFF
--- a/ecmascript/codegen/Cargo.toml
+++ b/ecmascript/codegen/Cargo.toml
@@ -7,7 +7,7 @@ include = ["Cargo.toml", "src/**/*.rs"]
 license = "Apache-2.0/MIT"
 name = "swc_ecma_codegen"
 repository = "https://github.com/swc-project/swc.git"
-version = "0.58.0"
+version = "0.58.1"
 
 [dependencies]
 bitflags = "1"
--- a/ecmascript/codegen/src/lib.rs
+++ b/ecmascript/codegen/src/lib.rs
@@ -1190,20 +1190,7 @@ impl<'a> Emitter<'a> {
 
         self.emit_leading_comments_of_span(node.span(), false)?;
 
-        punct!("`");
-        let i = 0;
-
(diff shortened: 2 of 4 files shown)


kdy1:
Closing as this pr is simply wrong

PR #1835, opened 2021-06-19
fix(transform/typescript): incorrectly strip import

Austaras:
Currently when there's an import item share the same name with object destructuring in variable declaration like
```ts
import { A } from "./a";
import { B } from "./b";

const { A: AB } = B;

console.log(A, AB);
```
the first import will be striped because `swc` would consider it being shadowed by local variable, which is wrong


kdy1 (approved):
Thanks!
I missed it.

PR #1844, opened 2021-06-23
chore: upgrade arbitrary crate to v1

punkeel:
Upgrade the arbitrary (fuzzing) crate to v1. This is a major bump, but the API hasn't changed much. We only need to add lifetimes in some impl blocks.


kdy1:
Do you actually need this?

punkeel:
I tried using `swc` + `cargo-fuzz` for an experiment, and was blocked by the different `arbitrary` versions used. The experiment didn't go the way I wanted it to, so I may not "need" it ... but I figured I could still contribute this to swc :-)

Up to you, I won't mind too much if you reject this PR.

kdy1 (approved):
Thanks!
I asked it just because this is a technically breaking change :)

punkeel:
Thanks for merging it :-)
And, indeed, it technically is a breaking change. Good catch, I totally missed it!

PR #1857, opened 2021-06-25
fix(es/transforms/base): Wrap call exprs when in a new expr callee

motiz88:
Fixes a bug where the fixer transforms code like this:
```
new (getObj()).ctor()
// Pseudo AST: NewExpr {callee: MemberExpr { object: CallExpr { callee: "getObj" }, property: "ctor" } }
```

to this, which is not equivalent:

```
new getObj().ctor()
// Pseudo AST: CallExpr {callee: MemberExpr { object: NewExpr { callee: "getObj" }, property: "ctor" }}
```

This is my first PR to swc and one of my first times writing Rust, so I'm not confident in the actual fix. Consider this a bug report with a _proposed_ fix attached to it.

I noticed that there was an existing traversal over member expressions and an `is_new` context flag for callees, so my idea was to try to reuse that and wrap a Call expression in parens if:

1. it's the object of a Member expression that is the callee of a New expression, or
2. it's the object of a Member expression that satisfies (1).


ecmascript/transforms/base/src/fixer.rs
```diff
@@ -1219,4 +1226,19 @@ var store = global[SHARED] || (global[SHARED] = {});
...
+        "new (getObj().ctor)()",
+        "new (getObj()).ctor()"
+    );
+    test_fixer!(
+        new_member_call_3,
+        "new (x.getObj().ctor)()",
+        "new (x.getObj()).ctor()"
+    );
+    identical!(new_call, "new (getCtor())");
+    test_fixer!(new_member_1, "new obj.ctor()", "new obj.ctor()");
```
motiz88, line 1242: The fix doesn't change the output of the `new_member_*` tests but I figured I'd add them as negative cases. I don't exactly know why one has the trailing parens and one doesn't - it would probably be good to normalise both to the same form if we can.

kdy1: Arguments of `new` expressions are optional and the ast of swc stores if user specified parens.

kdy1 (approved):
LGTM. Thanks!

PR #1864, opened 2021-06-27
feat(node-swc): Babel plugin

kdy1:
node:
 - [ ] Add babel plugin runner.


jasikpark:
Does `swc` have the ability to run babel plugins? Or do plugins need to be written for `swc` specifically?

PR #1868, opened 2021-06-29
Enable or_patterns feature for swc_ecma_loader.

tmpfs:
So that I can compile against the `swc_ecma_loader` crate to use `NodeResolver`.

### DIFF
--- a/ecmascript/loader/src/lib.rs
+++ b/ecmascript/loader/src/lib.rs
@@ -1,2 +1,3 @@
+#![feature(or_patterns)]
 pub mod resolve;
 pub mod resolvers;


kdy1:
This is not acceptable

tmpfs:
My bad, I needed a nightly update as `or_patterns` are now stable.

AlexMoutonNoble:
Hi Folks!
Just a note from outside that this use of this advanced feature had repercussions.
The new parceljs bundler uses this as a dependency and it fails to build with what looks like current (1.52.0) mainline rustc toolchain.  Updating to nightly I'm still getting errors.  Lots of searching leads me here and I just wanted to help fix a bug in some javascript!

I'm using nixpkgs https://github.com/NixOS/nixpkgs/archive/7fc63651a2bcdf4257ec07d4bb572f59264a451b.tar.gz fwiw

Thanks
A

tmpfs:
@AlexMoutonNoble, according to the release changelog: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1530-2021-06-17 the `or_patterns` features was stabilized in `1.53.0` so if you are using `1.52.1` (as your edit suggests) then an upgrade should fix the problem for you.

You may want to use the specific pinned version of `nightly` to have more confidence, see: https://github.com/swc-project/swc/blob/master/rust-toolchain.

AlexMoutonNoble:
Thanks yes I found 1.53 in nixpkgs unstable and that's working for me
Just venting for any future generations who may hit the same thing

Thanks, and sorry =)
Alex

PR #1891, opened 2021-07-04
fix(es/ast): Fix Reserve Words

### LINKED ISSUE #1838: Invalid ES5 emitted if class has reserved word as method name
**Describe the bug**

SWC emits a `function let() {}` which causes an error in strict mode stating `'let' is a reserved word` (and thus cannot be used as a function name) if a class has a method called `let`. I would imagine this is also the case with other reserved words.

**Input code**

```javascript
class Foo {
    let() {}
}
```

outputs:

```javascript
function _classCallCheck(instance, Constructor) {
    if (!(instance instanceof Constructor)) {
        throw new TypeError("Cannot call a class as a function");
    }
}
function _defineProperties(target, props) {
    for(var i = 0; i < props.length; i++){
        var descriptor = props[i];
        descriptor.enumerable = descriptor.enumerable || false;
        descriptor.configurable = true;
        if ("value" in descriptor) descriptor.writable = true;
        Object.defineProperty(target, descriptor.key, descriptor);
    }
}
function _createClass(Constructor, protoProps, staticProps) {
    if (protoProps) _defineProperties(Constructor.prototype, protoProps);
    if (staticProps) _defineProperties(Constructor, staticProps);
    return Constructor;
}
var Foo = /*#__PURE__*/ function() {
    "use strict";
    function Foo() {
        _classCallCheck(this, Foo);
    }
    _createClass(Foo, [
        {
            key: "let",
            value: function let() {        // <--- INVALID JAVASCRIPT
            }
        }
    ]);
    return Foo;
}();
;
```

In firefox 89.0.1, this error is thrown:

```
Uncaught SyntaxError: let is a reserved identifier
```

**Config**

N/A - no config, just running straight swc:

```console
$ swc test.js -o test.es5.js
```

**Expected behavior**
Either `function let_` or something of the sort, or simply don't name those functions anything.

**Version**
The version of @swc/core: `1.2.61`

**Additional context**
None.

Austaras:
fixes #1838 
@kdy1 two problems left:
1. I don't know what's needed to be escaped in bundler, so I just escape everything
2. What does es3/props_lit do? It seems do the opposite of `@babel/plugin-transform-property-literals`. That babel plugin transform
```js
var foo = {
  // changed
  const: function() {},
  var: function() {},

  // not changed
  default: 1,
  [a]: 2,
  foo: 1,
};
```
to
```js
var foo = {
  // changed
  "const": function () {},
  "var": function () {},
  // not changed
  "default": 1,
  [a]: 2,
  foo: 1
};
```
while swc transform string literal in prop name back to ident

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2018"
 license = "Apache-2.0/MIT"
 name = "swc"
 repository = "https://github.com/swc-project/swc.git"
-version = "0.26.0"
+version = "0.27.0"
 
 [lib]
 name = "swc"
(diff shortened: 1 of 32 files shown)


kdy1 (approved):
LGTM. Thanks!

kdy1:
Sorry, I forgot the questions.

1. Escaping everything is fine.
2. If I remember correctly, property literals pass converts `foo.catch` to `foo['catch']` for old browsers.

kdy1:
Hmm... I'm not sure about property literals pass.
I don't know why it does opposite task of babel one.

PR #1893, opened 2021-07-05
feat(plugin): Groundwork for rust plugin system

kdy1:
This PR adds a plugin system for rust.

With the first version, only transform plugins are allowed. It means, each plugins can export **one** implementation of `Fold`, but wrapped as a `fn (ast_json: &str) -> String` or `fn (ast_json: &str) -> Result<String, Error>` (not determined yet).

---

The plugin system will be implemented along with custom plugins for

 - emotion
 - styled components

for dogfooding.

---

I'll implement plugins along with CLI pr.

See https://github.com/swc-project/plugins/pull/1


Brooooooklyn:
A simple idea (needs to be verified):

Make program and everything related `#[repr(C)]`, and expose the api as:

```rust
#[repr(C)]
pub enum Program ...

extern "C" fn swc_plugin(*mut c_void) -> *mut c_void;

// expose proc macro
#[swc_plugin]
fn emotion_plugin(input: &mut Program) -> Program {
}

```

kdy1:
@Brooooooklyn It will be broken if version of the ast crate is not compatible, because a change of ast crate version means different memory layout.

I considered such approach long time ago, but I gave up because of it.

devongovett:
Very excited about this! I agree with @Brooooooklyn though: ideally we'd avoid serializing and deserializing the AST to JSON for performance. But I can understand the challenges there.

I think you would still have compatibility issues with different AST versions even with this approach. For example, if a plugin is running on a newer version of the AST than the core, then deserialization could fail when trying to load an AST where the plugin is expecting a new field but it doesn't exist, even if that field is not used by the plugin. Obviously you'd always have an issue if the plugin actually needs to use such a field, but I think there could be plenty of cases where a new field is added that plugins don't touch.

I did some thinking about this a bit before. I think it could be possible to autogenerate a C ABI that could ensure forward compatibility. Basically you'd have to treat everything as an opaque pointer. Rather than exposing the structs themselves, you'd expose pointers to those structs along with generated getter functions for each field. This way you could safely add new fields to the struct without breaking ABI compatibility (plugins don't have access to the layout of the actual structs).

For example:

```c
// opaque struct
typedef struct Module Module;

// getters
Span *module_get_span(Module *module);
ModuleItem *module_get_body(Module *module, int *count);
char *module_get_shebang(Module *module, int *length);

// constructor (basically Module::default())
Module *module_create();

// setters
void module_set_span(Module *module, Span *span);
void module_set_body(Module *module, ModuleItem *body, int length);
void module_set_shebang(Module *module, char *shebang, int length);
```

A small rust wrapper around this C ABI could be built to provide a nicer API for plugins. But a cool thing is that this would also allow plugins to be written in any language that supports calling C functions (which is basically all of them) with no copying. This is essentially how napi works, for example.

Of course, this seems like a rather large project, but my hope would be that most of that could be completely autogenerated from the Rust source code.

kdy1:
Oh, it seems like a great idea, thanks!
I was actually worried about old plugin with the new version of core.
Unknown fields will be dropped because ast of an old plugin does not know about them.

Brooooooklyn:
> Unknown fields will be dropped because ast of an old plugin does not know about them.

I don't think ***cross version abi compatibility*** is a big issue at the moment, abi compatible with the same version already  powerful for doing amazing things like customize third party plugin.

My suggestions is mark **native plugin** feature as ***experimental*** and don't modify or delete field in `AST` if possible. Introduce independent version number for `Program` may be a good idea, like `Node-API` version for `Node.js`

kdy1:
> don't modify or delete field in AST if possible

This requires one of the below.

 - Use same version of ast for all plugins
 - Add `extra: HashMap<String, serde_json::Value>` to **all** ast nodes.


Otherwise, fields will be removed while passing ast to the plugin

kdy1:
I think about it more.
Marking it as `experimental` seems good and I'll do so.


TypeScript is evolving, and ast definitions are changed quite frequently.
So I think I should consider ast version compatibility issue while designing it. 


> don't modify or delete field in AST if possible 

This is required to use old plugins with a new core and I think the suggestion by @devongovett is close to the best way.
I may use https://docs.rs/abi_stable/0.9.3/abi_stable/ because I think providing convenient rust plugin api is more important than supporting multiple languages.

jdalton:
With access from the AST for plugins would it be possible for a plugin to expose the AST to JavaScript land (as with the now deprecated `parse` APIs)? 

I've been able to get several X performance gains by using `swc` over parsers like `acorn`. 
Wondering what a migration path in v2.0 could look like.

kdy1:
> With access to the AST for rust plugins would it be possible for a plugin to expose the AST to JavaScript land

No. I want to prevent it if possible. `parse` will be provided, but it will be parse-only.

PR #1895, opened 2021-07-05
chore: Publish v1.2.63

kdy1:


Qix-:
The actions associated with this PR seem to have completed but I'm still getting a 404 on version `@swc/core@1.2.63`. Is there a bug with the action(s)?

kdy1:
Yes. I guess using cross-env will fix it.

PR #1879, opened 2021-07-01
fix: comments in empty file should be in leading comments

dsherret:
Fixes #1878

PR #1896, opened 2021-07-05
fix: put comments in file with only shebang and comments in the correct map

dsherret:
Caused by #1879. I missed this scenario earlier.

Luckily, nobody is going to write files like this.


kdy1 (approved):
Thanks!

PR #1899, opened 2021-07-06
fix(opt_chain): fold optional chain when chain includes member call at the end

### LINKED ISSUE #1882: Chained optional chaining transpilation leaves partial chain
**Describe the bug**
When there is chain of optional chain, some of chain left as-is without transpilation.

**Input code**

```
const button = renderedView.find(LoadingButton);
button?.prop('onClick')?.({} as any);
```

output

```
var ref;
const button = renderedView.find(LoadingButton);
var _obj = button?.prop('onClick');
(ref = _obj) === null || ref === void 0 ? void 0 : ref.call(_obj, {
});
```

raises runtime exception

```
var _obj = button?.prop('onClick');
                              ^

    SyntaxError: Unexpected token '.'

```

**Config**

```json
{
  "jsc": {
    "target": "es2018",
    "parser": {
      "syntax": "typescript",
      "decorators": false,
      "dynamicImport": true
    }
  }
}
```

**Expected behavior**

**Version**
The version of @swc/core:  @swc/core@1.2.62

**Additional context**

kwonoj:
- closes #1882

This PR attempts to fix #1882 when optional chain ends with member call, does not fold optional chain and leaves it behind.

I tried to follow suggestion https://github.com/swc-project/swc/issues/1882#issuecomment-874552592 but I may place it incorrectly, while test _seems_ correctly resolves optional chain. Will update changes if this is not expected behavior.

### DIFF
--- a/ecmascript/transforms/compat/Cargo.toml
+++ b/ecmascript/transforms/compat/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2018"
 license = "Apache-2.0/MIT"
 name = "swc_ecma_transforms_compat"
 repository = "https://github.com/swc-project/swc.git"
-version = "0.24.0"
+version = "0.24.1"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


kdy1 (approved):
LGTM. Thanks!

PR #1908, opened 2021-07-08
refactor: remove SourceMap dependency from dep-graph

dsherret:
There isn't any need for `analyze_dependencies` to use a SourceMap. Instead it can stay working with span/byte pos units so it's the same as the AST (that way, we don't mix units).

Additionally, it doesn't need to be limited to `SingleThreadedComments` and can instead take anything that implements the `Comments` trait.

What I'm trying to do is hold onto swc ASTs in Deno's language server, but I need everything to be `Send` in order to do that. Source map doesn't allow that, so it's up on the chopping block right now. Next will be `SourceFile` and `SingleThreadedComments`.


ecmascript/dep-graph/src/lib.rs
```diff
@@ -39,46 +37,41 @@ pub struct DependencyDescriptor {
...
     fn get_leading_comments(&self, span: Span) -> Vec<Comment> {
-        self.comments
-            .with_leading(span.lo, |comments| comments.to_vec())
+        if let Some(comments) = self.comments.take_leading(span.lo) {
+            let comments_copy = comments.clone();
+            self.comments.add_leading_comments(span.lo, comments);
+            comments_copy
+        } else {
+            Vec::with_capacity(0)
+        }
```
dsherret, line 67: This is kind of strange. Perhaps it would be nice to create a `ReadonlyComments` trait that only has the ability to view what's in the comment maps.

dsherret: Actually, thinking about this more we definitely shouldn't do this. This could cause some race conditions in multi-threaded code.

kdy1: If you are not using a trait object, you can define some extension methods which hold the lock for the byte position while executing closure.

kdy1 (commented):
Btw, `Lrc<SourceMap>` implements `Send + Sync` if you enable `concurrent` feature of `swc_common`.

ecmascript/dep-graph/src/lib.rs
dsherret: Hmmm... I'm not sure what you mean. How would this code handle it?

kdy1: I mean an extension method in `Comments` with some object-safety guards like `Self: Sized`.
(I'm not sure if it's a breaking change, though. I'll test it)

It accepts a closure (`FnOnce(&[Comment]) -> Ret`)

And implementor of `Comments` can use it to lock during operation.

kdy1: I'm not sure about avoiding locks. `RefCell` is fast enough for almost all usecases.

I decided to use a global hashmap because I didn't want to pass down comments everywhere, but maybe it was a wrong decision.

dsherret:
> Btw, Lrc<SourceMap> implements Send + Sync if you enable concurrent feature of swc_common.

Interesting. I feel like I knew that, but forgot and that's good to know! I don't want to turn on that feature because it would be a global change, would most likely be slower in our case, and introduces the possibility of a new class of bugs. I think I will be able to manage by only holding onto things like `SourceFile` and `SourceMaps` structs when we need them for things like transpiling and bundling or perhaps creating those after the fact. The rest of it should ideally work via only holding onto the AST, source file text, and comments (I will take them out of the `SingleThreadedComments`).

kdy1:
I see. I'm okay with current changes, but I'm not sure if you want me to merge it.

dsherret:
@kdy1 yeah, not yet. I want to look into fixing this possible race condition. I'm just looking into your comments.

ecmascript/dep-graph/src/lib.rs
dsherret: I'm still not sure how to make that work. I have committed a change that adds some `get` methods that clone... it's not ideal though and is a breaking change... please feel free to change it. I also removed the default implementation of `add_pure_comment` as it had a race condition (also fixed the race condition in the Optimizer code).

I just need to fix some compile errors as the CI is failing... whoops...

kdy1:
Is this PR ready? If not, please ping me when you are done :)

dsherret:
Hey @kdy1, yeah, it is now. I’m just not sure about the changes to the comments trait and what would be best there.

PR #1914, opened 2021-07-13
fix(ecmascript/transforms/proposal/decorators/legacy): fix decorate class prop stmts order in constructor

### LINKED ISSUE #1913: Transform incorrect when Use legacy decorator and class property syntax
**Describe the bug**
call decorated class field decorator will cause `TypeError: this.doSomething is not a function`
**Input code**
``` ts
import { action } from 'mobx';

class Store {
    constructor() {
	this.doSomething();
    }

    @action
    doSomething = () => {
	    console.log('run');
    }
}

const store = new Store();
```
**Config**

```json

{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true,
      "decorators": true
    },
    "transform": {
      "legacyDecorator": true
    },
    "loose": false,
    "target": "es2020"
  },
  "module": {
    "type": "commonjs",

    "strict": false,
    "strictMode": true,
    "lazy": false,
    "noInterop": false
  }
}
```

**Expected behavior**
the example code works well.
**Version**
The version of @swc/core: 1.2.63

**Additional context**

IronLu233:
close #1913


ecmascript/transforms/proposal/src/decorators/legacy.rs
```diff
@@ -773,12 +773,30 @@ impl Legacy {
...
 
-            constructor
+            let decorate_stmts_insert_position = constructor
                 .body
                 .as_mut()
                 .unwrap()
                 .stmts
-                .extend(constructor_stmts)
+                .iter()
+                .position(move |stmt| {
```
kdy1, line 782: Is this `move` necessary?
I think closure does not use any variables from outer scopes.

ecmascript/transforms/proposal/src/decorators/legacy.rs
```diff
@@ -773,12 +773,30 @@ impl Legacy {
...
+                    if let Stmt::Expr(expr_stmt) = stmt {
+                        let expr_stmt = *expr_stmt.expr.clone();
```
kdy1, line 784: Is this clone necessary?

kdy1, line 784: `expr_stmt` is a bad name for storing `Expr`. Can you change this to `expr` instead?

ecmascript/transforms/proposal/src/decorators/legacy.rs
```diff
@@ -773,12 +773,30 @@ impl Legacy {
...
+                        }
+                    }
+
+                    return false;
+                })
+                .map_or(0, |p| p + 1);
+
+            constructor.body.as_mut().unwrap().stmts.splice(
+                decorate_stmts_insert_position..decorate_stmts_insert_position,
+                constructor_stmts.iter().cloned(),
```
kdy1, line 798: Is cloning necessary?

If `constructor_stmts ` is used multiple times, the code will be duplicated. If this behavior is not intended, please use `take(&mut constructor_stmts)` where `take` is `std::mem::take`.

kdy1 (changes requested):
I found some issues, but thank you anyway!

ecmascript/transforms/proposal/src/decorators/legacy.rs
IronLu233: it's unnecessary. I just pass `constructor_stmts` to `Vec::splice`, because `SmallVec` implement `Iterator` trait. Also we don't need `take`

PR #1923, opened 2021-07-15
fix(node-swc/types): fix `TsEnumDeclaration` property

plylrnsdy:
`TsEnumDeclaration` property `member` should be `members`


kdy1 (approved):
LGTM. Thanks!

PR #1931, opened 2021-07-17
fix(builder):  move import_analyzer after  regenerator-runtime transform

### LINKED ISSUE #1924: use generator in code does not inject _interopRequireDefault implement into output
**Describe the bug**
compile code using generator without any other import will  generate a broken code without `_interopRequireDefault`
**Input code**

export default function * foo() {}
**Config**

```json
{
  "module": {
    "type": "commonjs"
  },
  "jsc": {
  }
}
```

**Expected behavior**
generated code should include _interopRequireDefault implements.

**Version**
The version of @swc/core: 1.2.64

**Additional context**
generated code
``` js
"use strict";
    value: true
});
exports.default = foo;
var _regeneratorRuntime = _interopRequireDefault(require("regenerator-runtime"));
function foo() {
    return _regeneratorRuntime.default.wrap(function foo$(_ctx) {
            case 0:
            case "end":
                return _ctx.stop();
        }
    }, foo);
}

```

IronLu233:
close #1924

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -187,13 +187,13 @@ impl<'a, 'b, P: swc_ecma_visit::Fold> PassBuilder<'a, 'b, P> {
 
         let module_scope = Rc::new(RefCell::new(Scope::default()));
         chain!(
+            self.pass,
+            compat_pass,
             // module / helper
             Optional::new(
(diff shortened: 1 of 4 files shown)


kdy1 (approved):
LGTM. Thanks!

Issue #1937, opened 2021-07-19
dyld: lazy symbol binding failed: Symbol not found: _SecRandomCopyBytes


Janpot:
**Describe the bug**

Ran into this in the Next.js repo after [this PR](https://github.com/vercel/next.js/pull/27167) was merged. Was able to replicate the behavior of their build system in a dumbed down script. They seem to individually compile each file concurrently.

**Input code**

```js
// build.js
const { transform } = require("@swc/core");

const AMOUNT_OF_FILES = 20;
const AMOUNT_OF_LINES = 1000;

for (let i = 0; i < AMOUNT_OF_FILES; i++) {
  transform(`console.log('hello');\n`.repeat(AMOUNT_OF_LINES));
}
```

then run

```
node ./build.js
```

This sometimes just gets aborted with a

```
dyld: lazy symbol binding failed: Symbol not found: _SecRandomCopyBytes
  Referenced from: /Projects/test-swc/node_modules/@swc/core-darwin-arm64/swc.darwin-arm64.node
  Expected in: flat namespace

dyld: Symbol not found: _SecRandomCopyBytes
  Referenced from: /Projects/test-swc/node_modules/@swc/core-darwin-arm64/swc.darwin-arm64.node
  Expected in: flat namespace

zsh: abort      node build.js
```

It's not deterministic and increasing `AMOUNT_OF_FILES` and `AMOUNT_OF_LINES` increases the chance of this happening. The lower these values are, the higher the chance of successfully finishing the script. Sometimes the errors get printed mixed into each other. Using `transformSync` works without a problem.

**Expected behavior**
No errors

**Version**
The version of @swc/core: 1.2.65

**Additional context**
This happens on Apple M1, macOS Big Sur 11.4.
I've seen it happen on node.js 14, 15 and 16

kdy1:
I was aware of this issue, and seems like a bug related to cross-compilation from intel mac to m1 mac.

PR #1939, opened 2021-07-20
fix(node/swc): m1 macos build

Brooooooklyn:
Close https://github.com/swc-project/swc/issues/1937

Tested on m1 macOS-11.4


kdy1 (approved):
Thank you a lot!!

huozhi:
Thank you both for this fix! Can confirm the issue is gone on my m1 🙏

PR #1940, opened 2021-07-20
fix(import): allow namespace import with default import 

### LINKED ISSUE #1938: module: `commonjs` raises "import * as foo cannot be used with other type of import specifiers"
**Describe the bug**

When import is mixture of default (interoped) import with named import, `"import * as foo cannot be used with other type of import specifiers" ` occurs.

**Input code**

```
const { transformSync } = require('@swc/core');

const out = transformSync(`
import reducer, * as SocketStore from './redux/stores/socket-store';

console.log(reducer);
console.log(SocketStore);
`,
  {
    jsc: {
      target: 'es2017',
      parser: {
        syntax: 'typescript',
        decorators: false,
        tsx: true,
      },
    },
    module: {
      type: 'commonjs',
    },
  });
```

raises 

```
thread '<unnamed>' panicked at 'internal error: entered unreachable code: import * as foo cannot be used with other type of import specifiers', ecmascript/transforms/module/src/import_analysis.rs:104:55
stack backtrace:
   0:        0x106c95d09 - _rust_eh_personality
   1:        0x1069f051e - _napi_register_module_v1
```

**Config**

above code inclues it, 

```js
{
    jsc: {
      target: 'es2017',
      parser: {
        syntax: 'typescript',
        decorators: false,
        tsx: true,
      },
    },
    module: {
      type: 'commonjs',
    },
  }
```

Specifically, `module.type` should be specified as `commonjs`.

**Expected behavior**
Allow to import default export with other exports at once.

**Version**
The version of @swc/core: 1.2.65

**Additional context**

kwonoj:
- closes #1938 

This PR attempts to allow named wildcard imports when there are multiple import specifiers. Currently it is unreachable code paths, but it is possible to have imports like

```
import foo, * as actions from '...'
```

to have default import with named imports. PR treats namespace same as single import with namespace case. 

I feel this may not be feasible approach, or brings redundant code paths and would like to update as needed.

### DIFF
--- a/ecmascript/transforms/module/Cargo.toml
+++ b/ecmascript/transforms/module/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2018"
 license = "Apache-2.0/MIT"
 name = "swc_ecma_transforms_module"
 repository = "https://github.com/swc-project/swc.git"
-version = "0.26.1"
+version = "0.26.2"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
--- a/ecmascript/transforms/module/src/import_analysis.rs
+++ b/ecmascript/transforms/module/src/import_analysis.rs
@@ -101,9 +101,11 @@ impl Visit for ImportAnalyzer {
             let mut has_non_default = false;
             for s in &import.specifiers {
                 match *s {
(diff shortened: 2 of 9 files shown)


kdy1 (approved):
Thanks!

Issue #846, opened 2020-06-16
Decorators break code with es2019 target


ryuhhnn:
**Describe the bug**

If you have a class that extends a base class such as the following:

```js
class SomeClass {
  someMethod() {}
}

class OtherClass extends SomeClass {
  anotherMethod() {
    super.someMethod()
  }
}
```

You will get the following error: `super is only allowed in object methods and classes` because the compiler will output `OtherClass` as a returned object thus inhibiting the ability to use the `this` keyword. 

This is similar to an issue that Babel was experiencing a few years ago: https://github.com/babel/babel/issues/2775

**Config**
```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "decorators": true,
      "decoratorsBeforeExport": true,
      "classProperty": true,
      "nullishCoalescing": true,
      "optionalChaining": true
    },
    "target": "es2019"
  }
}
```

**Expected behavior**
A class should be able to call `super.*` on its parent class.

**Version**
```
@swc/cli@^0.1.24
@swc/core@^1.2.2
```

kdy1:
>     "target": "es2019"

This part is problematic. Because you told swc that you target es2019, swc does **not** transcompile classes, so it's just emitted as-is.
I have added some tests anyway.

ryuhhnn:
Not sure why this issue was closed, I provided a valid case where this breaks and it wasn't fixed. I understand that swc does not transpile classes when the target is es2019, but if that is the case, then the decorator transpilation needs to be modified to not return an object and instead return a class. Is this something I can help with? I really want to start using swc, but this is a critical bug preventing us from adopting.

kdy1:
I missed the decorator part as it was not in example. I'm Sorry for it.

Brooooooklyn:
Decorator should always be transpiled. Now it would only be transpiled under target <= `es2015`

kdy1:
@Brooooooklyn It's transpiled iff syntax.decorators is true, but it seems like transpiled code crashes for some reason

tvarsis:
There seems to be a lot of mentioned of this ticket from different milestones, but it does not seems resolved yet. Is there any update on this issue which makes our project fail to transpile?

kdy1:
@tvarsis 

I added this to the milestones because I wanted to fix it.
But the project is maintained by a one-man team and I couldn't fix it because I don't have time.
I have lots of tasks to do.

kdy1:
I did some investigation on this.
[...]

PR #1960, opened 2021-07-23
fix(ecmascript/transforms/classes/super_fields): Fix broken code when transpile super member access

### LINKED ISSUE #1959: Access super member in nested expr transform incorrect
## Describe the bug
transform `super.foo() + 1` will become `_get(_getPrototypeOf(Extended.prototype), 'getNext', this)() + 1` 
**Input code**
class Base {
	getNext() {
		return this.value++
	}
}

class Extended extends Base {
	getNext() {
		return super.getNext() + 1
	}
}

const ins = new Extended();
ins.getNext();
## Config

```json
{
  "module": {
    "type": "commonjs"
  }
}
```

## Expected behavior
should invoke super class member's function with call
``` js
 _get(_getPrototypeOf(Extended.prototype), "getNext", this).call(this) + 1;
```

**Version**
The version of @swc/core: 1.2.66

## Additional context
the code above will emit an error:
```
TypeError: Cannot read property 'value' of undefined
```

IronLu233:
close #1959


IronLu233:
@kdy1 Not sure this fix is the best way, I need your help for best fix approach.

kdy1 (approved):
I think clone can be removed, but I'll try it by myself. Thanks!

kdy1:
Postponing merge as tests fail.

kdy1:
Oh sorry I missed your comment.

`_get(_getPrototypeOf(Extended.prototype), "getNext", this).call(this) + 1;` is ideal, which is also emitted by babel.

IronLu233:
the CI test failed tells me this fix will cause a bug. I need more time to find a better way for fixing it. @kdy1

ecmascript/transforms/classes/src/super_field.rs
```diff
@@ -193,6 +193,36 @@ impl<'a> Fold for SuperCalleeFolder<'a> {
...
+                    Expr::Call(CallExpr {
+                        span: DUMMY_SP,
+                        callee: MemberExpr {
+                            span: DUMMY_SP,
+                            obj: ExprOrSuper::Expr(expr.clone()),
+                            prop: Box::new(Expr::Ident(quote_ident!("call"))),
+                            computed: false,
+                        }
+                        .as_callee(),
+                        args: iter::once(this_arg).chain(args.clone()).collect(),
```
kdy1, line 220: Is cloing required?

ecmascript/transforms/classes/src/super_field.rs
```diff
@@ -193,6 +193,36 @@ impl<'a> Fold for SuperCalleeFolder<'a> {
...
+                callee: ExprOrSuper::Expr(ref expr),
+                ref args,
+                ref type_args,
+                ..
+            }) => match &**expr {
+                Expr::Member(MemberExpr {
+                    obj: ExprOrSuper::Super(Super { span: super_token }),
+                    ..
+                }) => {
+                    let this_arg = ThisExpr {
```
kdy1, line 207: I looked at the diffs from test and this seems to be the cause. Maybe you need to use aliased variable instead.

IronLu233: Although I use 
``` rust
let this_arg = ThisExpr { span: DUMMY_SPAN };
```
the `this` still be transformed into `_this`

kdy1: I mean the reverse. Instead of using `ThisExpr`, you may try injecting `var _this = this`.

IronLu233:
@kdy1 I need some help. When `_this` will be replaced with `_this1` or other ident?

kdy1:
It depends on the context, but in this case `SuperCalleeFolder` ( https://github.com/swc-project/swc/blob/e916b35dd2ee3a191fee487ac61e64d9410b72ed/ecmascript/transforms/classes/src/super_field.rs )
This struct is used by all class transforms, including `decorators` which is used for tests for  #846

IronLu233:
> It depends on the context, but in this case `SuperCalleeFolder` ( https://github.com/swc-project/swc/blob/e916b35dd2ee3a191fee487ac61e64d9410b72ed/ecmascript/transforms/classes/src/super_field.rs )
> This struct is used by all class transforms, including `decorators` which is used for tests for #846

So you mean adding `var _this = this` is not the best way for this? And I need find a better way for fixing it?

kdy1:
I was confused by diff, sorry.
And this seems to require a different approach. Fixing `SuperCalleeFolder` seems good, but it's not a simple struct so it would take some time for debugging.

IronLu233:
I have a new idea for fixing it.
Now, when we fold code like `super() + 114514`, the fold order in  `SuperFieldAccessFolder#fold_expr` is:
1. fold self with SuperCalleeFolder
2. transform super field call if necessary
3. fold children with self(`SuperFiledAccessFolder`)
My idea is change the order to
1. fold children with self`SuperFieldAccessFolder`
2. transform super call to `super.foo.call(this)` or `super.foo.apply(this)`
3. fold transformed expr with `SuperCalleeFolder`;
Soon I will check whether it is available

IronLu233:
Hi, @kdy1 I meet another problem. When I ran the test `get_set_memoized_assign` using
``` bash
export RUST_BACKTRACE=full
export PATH="$PATH:$PWD/ecmascript/transforms/node_modules/.bin"
cargo test --package swc_ecma_transforms_compat --test es2015_classes --all-features -- get_set_memoized_assign --exact --nocapture
```

the unit test fails.
image
But when I run all unit test using
``` bash
    cargo test --all --all-features
```
it passes.
Why? seems unit test is not isolated.

And I want know how Mark works in span.
Please give me some help🙏

kdy1:
> And I want know how Mark works in span.

https://rustdoc.swc.rs/swc_ecma_transforms_base/resolver/fn.resolver_with_mark.html decribes how resolver works. Resolver attach`Mark`s to the identifiers, so it will help.

I'm not sure about the unit test. It's isolated. And diff says the decision to inject a variable is different between the expected one and the actual one. No test suite does such a thing.

ecmascript/transforms/classes/src/super_field.rs
```diff
@@ -101,7 +82,120 @@ impl<'a> Fold for SuperCalleeFolder<'a> {
...
+                    prop,
+                    computed,
+                    ..
+                }) => {
+                    let this = match self.constructor_this_mark {
+                        Some(mark) => quote_ident!(DUMMY_SP.apply_mark(mark), "_this").as_arg(),
+                        _ => ThisExpr { span: DUMMY_SP }.as_arg(),
+                    };
+
+                    let callee = self.super_to_get_call(*super_token, prop.clone(), *computed);
```
IronLu233, line 146: I don't know how to prevent `clone` in this match branch. I try to use *prop, but the compiler does not allow to do this.

ecmascript/transforms/proposal/src/decorators.rs
```diff
@@ -355,7 +355,7 @@ impl Decorators {
                     vars: &mut vars,
                     constructor_this_mark: None,
                     is_static: method.is_static,
-                    folding_constructor: true,
+                    folding_constructor: false,
```
IronLu233, line 358: Folding decorator seems not in constructor function, they are in the second parameter of `_decorate`, which is factory function for decorator.

ecmascript/transforms/compat/tests/es2020_class_properties.rs
```diff
@@ -1791,7 +1791,7 @@ function (A1) {
 
     _foo.set(_assertThisInitialized(_this), {
       writable: true,
-      value: _get(_getPrototypeOf(B.prototype), "foo", _assertThisInitialized(_this))()
+      value: _get(_getPrototypeOf(B.prototype), "foo", _assertThisInitialized(_this)).call(_this)
```
IronLu233, line 1794: Here, in my opinion, if without `call`, the `this` may not point to the instance.

ecmascript/transforms/classes/src/super_field.rs
kdy1: You can remove clone by changing `callee: ExprOrSuper::Expr(ref callee_expr), ` to `callee: ExprOrSuper::Expr(callee_expr)`, but I don't think it worth.
It requires reconstructing when `callee_expr` is not a member expression we wanted, and it means dirty code, and `memmove`.

I think clone is ok in this case, as `expr` in super[expr]` is very simple in most cases.

kdy1 (approved):
Thank you a lot!
Code becomes much more understandable. Thanks!

PR #1973, opened 2021-07-29
fix(es/codegen): Remove extra space of import decl

nuintun:
See: #1690


kdy1:
Why did you close this?

nuintun:
@kdy1 Because you updated, I'll re PR later~

nuintun:
@kdy1 #1975

PR #1977, opened 2021-07-29
fix: private method overloads should be removed when type stripping

dsherret:
Will fix https://github.com/denoland/deno/issues/11550


kdy1 (approved):
LGTM. Thanks!

Issue #1952, opened 2021-07-22
Redesign comment system


kdy1:
It seems like a misdesign.

### Convenience

API is tricky because of comments.

### Performance

The current design makes `fixer` slow because it requires moving comments in hashmap again and again.
Moving comments involves acquiring / releasing lock and rehashing.

It can be improved using buffer, but maybe there is a fundamental problem.

cc @dsherret as you may be interested in this.

dsherret:
Yeah, this would be good. I also still think there are race conditions in `with_leading` and `with_trailing` where two threads might remove one after the other and one thread would end up with different comments.

https://github.com/swc-project/swc/blob/f44e25c3afba2470e8013b757a6bacf30614b1f9/common/src/comments.rs#L291

PR #1996, opened 2021-08-03
fix: do not mutate comment maps while transforming

dsherret:
This change does three things:

1. Does not mutate the comment maps in transforms. This allows for another thread to be safely doing stuff with the comment maps without having another thread mutating it in a transform. Edit: Actually, I'm not sure if this makes sense... instead should a cloned version of the comment map be worked on?
2. It does not remove comments and forget about them when emitting.
3. Removes race condition in `with_leading` and `with_trailing` until #1952 is implemented.

### DIFF
--- a/tests/projects.rs
+++ b/tests/projects.rs
@@ -244,7 +244,7 @@ fn issue_406() {
     let s = file("tests/projects/issue-406/input.js").unwrap();
     println!("{}", s);
 
-    assert!(s.contains("return true"));
+    assert!(s.contains("return(true)"));
 }
 
 #[test]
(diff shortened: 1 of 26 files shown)


tests/projects.rs
```diff
@@ -244,7 +244,7 @@ fn issue_406() {
     let s = file("tests/projects/issue-406/input.js").unwrap();
     println!("{}", s);
 
-    assert!(s.contains("return true"));
+    assert!(s.contains("return(true)"));
```
dsherret, line 247: I'm not sure why this change was necessary? Output seems functionally the same.

dsherret:
On second thought, I'm not sure this should be done and this doesn't really solve what I was doing because `move_X` methods still occur and that makes sense to do.

PR #2042, opened 2021-08-09
fix(es/parser): Recover from `import.meta` in scripts

### LINKED ISSUE #2041: Do not return an error from `parse_program()` when the text contains `import.meta`
**Describe the feature**

When the text contains `import.meta` it would be nice to still be able to get an AST via `parse_program()`. Going along with the theme of #1178, I think it would be better for it to still return an AST, but put this in the errors diagnostic collection instead.

It would also be nice if the same thing happened for import and export declarations (IIRC it does this for them too).

kdy1:
swc_ecma_parser:
 - Recover from `import.meta` in scripts. (Closes #2041)
 - Allow `import.meta` when using `parse_program`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2440,7 +2440,7 @@ dependencies = [
 
 [[package]]
 name = "swc_ecma_parser"
-version = "0.66.1"
+version = "0.66.2"
 dependencies = [
  "either",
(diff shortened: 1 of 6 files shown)


bartlomieju:
Thanks for quick fix @kdy1!

Issue #1956, opened 2021-07-23
Support browser field of package.json


tmpfs:
**Describe the feature**

Support the `browser` field of `package.json` ([docs](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#browser)) to determine an entry point that overrides `main` when the target is a browser environment. This field is supported by Webpack and other bundlers and would be helpful for users migrating to spack/swc.

It would enable support for isomorphic packages such as [isomorphic-fetch](https://github.com/matthew-andrews/isomorphic-fetch/blob/master/package.json#L5).

**Additional context**

This should be enabled by default (the assumption is that the browser is the most common target) but an option needs to be provided to disable use of  the `browser` field for other targets such as server-side code. This could be achieved with a new configuration parameter in `.swcrc` which could be set on the command line when invoking `spack` to override the configuration.

Note I have assumed this is not yet supported as the `PackageJson` struct in `NodeResolver` does not declare the field.

**Reference**

* [Webpack aliasFields](https://webpack.js.org/configuration/resolve/#resolvealiasfields)
* [Browser field spec](https://github.com/defunctzombie/package-browser-field-spec)

kdy1:
Is this a bug about the bundler?

tmpfs:
I don't think it's a bug so I created a feature request as I figured you just haven't had time to implement this yet. Happy to help out to get this landed of course :+1: 

It is worth being aware that handling package entry points correctly from `package.json` configuration is a bit of a can of worms. So in addition to `browser` there is also the `module` field when the `type` is set to `module` and then there is the more recent `exports` ([docs](https://nodejs.org/dist/latest-v16.x/docs/api/packages.html#packages_package_entry_points)) which is specific to nodejs and supports different entry points for ESM (`import`) and CommonJS (`require`) - implementing this may involve passing a context to `NodeResolver` indicating whether to prefer ESM or CommonJS entry points.

Neither of these are part of the official `package.json` specification (unlike `browser`) but many bundlers support them and I think the goal for spack/swc should be compatibility so migrating from other bundlers is easy.

kdy1:
I mean is this feature request about the bundler of the resolver.

tmpfs:
It is related to the bundler as it is about resolving package entry points but I think the main part of the implementation would go in `NodeResolver` as that is where `package.json` is read and the main entry point is determined: https://github.com/swc-project/swc/blob/master/ecmascript/loader/src/resolvers/node.rs#L150.

[...]

PR #2045, opened 2021-08-10
Support target runtime environment.

### LINKED ISSUE #1955: Support shims for built in modules
**Describe the bug**

Compiling built in modules when the target is the browser and `paths` are configured does not use the packages from `paths` but compiles down to `import` statements. Possibly a duplicate of #1943.

**Input code**

An MVP reproducing and describing the issue is here:

https://github.com/tmpfs/swc-builtin-shim 

**Config**

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": true,
      "dynamicImport": false,
      "privateMethod": false,
      "functionBind": false,
      "exportDefaultFrom": false,
      "exportNamespaceFrom": false,
      "decorators": false,
      "topLevelAwait": false,
      "importMeta": false
    },
    "paths": {
      "assert": ["node_modules/assert"],
      "events": ["node_modules/events"]
    },
    "transform": null,
    "target": "es2016",
    "loose": false,
    "externalHelpers": false,
    "keepClassNames": false
  }
}
```

**Expected behavior**

We expect the packages configured in `paths` to be used instead of the default behavior for handling builtin modules.

**Version**
The version of @swc/core: 1.2.66

**Additional context**

N/A

tmpfs:
Update NodeResolver so it only special cases built in modules when
`TargetEnv::Node`.

This is the preliminary work to fix #1955. All we need to do now is thread the target environment (`TargetEnv`) from `spack` command line arguments into `NodeResolver::new` for proper support.

This is a breaking change as it now expects the browser to be the default target so builtin modules are no longer special-cased by default. I think it is reasonable to assume that the browser should be the default target for compiling/bundling.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -28,7 +28,7 @@ pub use self::{
 pub use ast_node::ast_serde;
 pub use ast_node::{ast_node, DeserializeEnum, Spanned};
 pub use from_variant::FromVariant;
-use serde::{Deserialize, Serialize};
+use serde::Serialize;
 use std::fmt::Debug;
 pub use swc_eq_ignore_macros::EqIgnoreSpan;
 pub use swc_eq_ignore_macros::TypeEq;
@@ -39,21 +39,6 @@ pub mod private;
 /// A trait for ast nodes.
 pub trait AstNode: Debug + PartialEq + Clone + Spanned + Serialize {
     const TYPE: &'static str;
-}
-
-/// Target runtime environment.
-#[derive(Debug, Serialize, Deserialize)]
-pub enum TargetEnv {
-    #[serde(rename = "browser")]
-    Browser,
-    #[serde(rename = "node")]
-    Node,
-}
-
-impl Default for TargetEnv {
-    fn default() -> Self {
-        TargetEnv::Browser
-    }
 }
 
 pub mod comments;
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -28,7 +28,7 @@ pub use self::{
 pub use ast_node::ast_serde;
 pub use ast_node::{ast_node, DeserializeEnum, Spanned};
 pub use from_variant::FromVariant;
-use serde::Serialize;
+use serde::{Serialize, Deserialize};
 use std::fmt::Debug;
 pub use swc_eq_ignore_macros::EqIgnoreSpan;
 pub use swc_eq_ignore_macros::TypeEq;
@@ -39,6 +39,21 @@ pub mod private;
 /// A trait for ast nodes.
 pub trait AstNode: Debug + PartialEq + Clone + Spanned + Serialize {
     const TYPE: &'static str;
+}
+
+/// Target runtime environment.
+#[derive(Debug, Serialize, Deserialize)]
+pub enum TargetEnv {
+    #[serde(rename = "browser")]
+    Browser,
+    #[serde(rename = "node")]
+    Node,
+}
+
+impl Default for TargetEnv {
+    fn default() -> Self {
+        TargetEnv::Browser
+    }
 }
 
 pub mod comments;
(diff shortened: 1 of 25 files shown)


tmpfs:
See: https://github.com/swc-project/cli/pull/50

tmpfs:
@kdy1, Would appreciate your feedback on this.

I suggest renaming `NodeResolver` in `swc_ecma_loaders` to `NodeModulesResolver` to be a bit clearer on it's purpose and to prevent a name collision with the `NodeResolver` type exported from `swc` to avoid this renaming: https://github.com/tmpfs/swc/blob/builtins-browser/spack/tests/fixture.rs#L22

kdy1:
> I suggest renaming NodeResolver in swc_ecma_loaders to NodeModulesResolver to be a bit clearer on it's purpose.

Seems like a good idea to me. As it's a breaking change anyway, I'm fine with such renaming.

common/src/lib.rs
```diff
@@ -41,6 +41,21 @@ pub trait AstNode: Debug + PartialEq + Clone + Spanned + Serialize {
     const TYPE: &'static str;
 }
 
+/// Target runtime environment.
+#[derive(Debug, Serialize, Deserialize)]
+pub enum TargetEnv {
```
kdy1, line 46: I'm a bit worried about adding this to `swc_common`. Ideally, only `node` and `browser` could be the only two target supported by ecmascript, but `swc_common` is for language-agnostic stuff.
Can you move this to `swc_ecma_ast`?

Reasoning is that, I may implement support for other targets but ecmascript does not support such target. I'm not sure if such case exist, but I want to stay safe.

kdy1 (commented):
Almost LGTM. Thanks!
I have one concern, though.

common/src/lib.rs
tmpfs: Thanks, I wasn't sure where to put this so that's helpful. It is moved to `swc_ecma_ast` now :+1:

tmpfs:
@kdy1, I also added an implementation for #1956 as we now know the target runtime environment it was simple to support.

I think that we should remove `swc_main` from `PackageJson`, is there a reason this exists rather than people just using `main`? I figured maybe it is legacy that can be removed as I have never seen it documented anywhere.

kdy1:
> I think that we should remove swc_main from PackageJson, is there a reason this exists rather than people just using main? I figured maybe it is legacy that can be removed as I have never seen it documented anywhere.

No, it's useless and I don't expect someone to using it as it's not documented at all. Can you please remove it?

tmpfs:
Cool, `swc_main` has been removed. I am tempted to remove `esnext` ([esnext info](https://github.com/stereobooster/package.json#esnext)) too as it is another legacy field that has been superceded by more modern alternatives like `exports`.

What do you think?

kdy1:
I think it's also fine. Thanks!

tmpfs:
Thanks @kdy1, that is done too! Hope we can get this merged now :+1: 

I will open a feature request for supporting the `exports` mapping in `package.json` but it is [quite convoluted](https://nodejs.org/dist/latest-v16.x/docs/api/packages.html#packages_package_entry_points) so best to handle that as a separate task.

kdy1 (approved):
LGTM. Thank you!

vjpr:
> I will open a feature request for supporting the `exports` mapping in `package.json` but it is [quite convoluted](https://nodejs.org/dist/latest-v16.x/docs/api/packages.html#packages_package_entry_points) so best to handle that as a separate task.

@tmpfs Where can I find this issue? I'm having some trouble with it over here: https://github.com/swc-project/swc/issues/5214 and https://github.com/swc-project/swc/issues/5144

tmpfs:
> > I will open a feature request for supporting the `exports` mapping in `package.json` but it is [quite convoluted](https://nodejs.org/dist/latest-v16.x/docs/api/packages.html#packages_package_entry_points) so best to handle that as a separate task.
> 
> @tmpfs Where can I find this issue? I'm having some trouble with it over here: #5214 and #5144

I can't find it either so I think I probably forgot to raise it; go ahead 👍

vjpr:
@tmpfs I'm actually not sure why swc needs to resolve imports at all. I wish I could just disable it.

PR #2053, opened 2021-08-11
feat(ecmascript/transforms/module): improve transform `misc import const throw`

### LINKED ISSUE #2049: A better transform for "misc import thorw"
**Describe the feature**
Now these code
``` js
import { bar } from 'foo';
bar = jest.fn()
```
will be transformed to
``` js
bar = (jest.fn(), (function() { throw new Error('"' + 'bar' + '" is read-only.') })())
```
If the assignee `bar` becomes `_foo.bar`(Babel's implement now), after jest plugin removes the assert, in the test, the test writer can override modules, for mock something.

IronLu233:
close #2049


kdy1 (approved):
LGTM. Thanks!

PR #2054, opened 2021-08-11
Support tsconfig paths in spack.

tmpfs:
So that setting paths in `options.jsc.paths` in `spack.config.js` is respected when resolving modules.


tmpfs:
@kdy1, I though this would solve my problem with aliasing builtin modules but it doesn't work so I think `spack` needs to introduce the concept of *aliases*.

For example, I have a dependency that requires the node builtin module `stream` but I am building for the browser so I want to use `stream-browserify` instead. Currently, paths do not allow me to do that as I get this error:

```
not processed by tsc resolver because base module is in node_modules
```

Can we add an `alias` configuration to `spack.config.js` that would just do a straight rewriting of a module `source` on a match:

```json
alias: {
  stream: "stream-browserify",
  fs: "localstorage-fs"
}
```

I think we could add this to `NodeModulesResolver`, would that work?

kdy1 (approved):
Seems like a good change. Thanks!

tmpfs:
@kdy1, I have added the `alias` functionality which is helping me a lot step through my build errors. One small detail I changed is that I figured it would be better to allow different aliases per target runtime environment for isomorphic builds, so the config now looks like:

```javascript
alias: {
  browser: {
    stream: "stream-browserify"
  }
}
```

The downside is that if somebody wants the same aliases for `node` and the `browser` some duplication is required but I think it is better to allow them to be configured individually.

I think this can be merged now if you approve :+1:

spack/src/config/mod.rs
```diff
@@ -48,6 +48,9 @@ pub struct Config {
 
     #[serde(default)]
     pub extenal_modules: Vec<JsWord>,
+
+    #[serde(default)]
+    pub alias: HashMap<TargetEnv, HashMap<String, String>>,
```
kdy1, line 53: Can you use `fxhash::HashMap` instead?

src/lib.rs
```diff
@@ -48,13 +48,34 @@ use swc_ecma_visit::FoldWith;
...
+        lru::CachingResolver, node::NodeModulesResolver, tsc::TsConfigResolver,
+    };
 
     pub type NodeResolver = CachingResolver<NodeModulesResolver>;
 
-    pub fn environment_resolver(target_env: TargetEnv) -> NodeResolver {
-        CachingResolver::new(40, NodeModulesResolver::new(target_env))
+    pub fn paths_resolver(
+        target_env: TargetEnv,
+        alias: HashMap<String, String>,
```
kdy1, line 62: Same here.

src/lib.rs
```diff
@@ -48,13 +48,34 @@ use swc_ecma_visit::FoldWith;
...
+            NodeModulesResolver::new(target_env, alias),
+            base_url.clone().into(),
+            paths.clone(),
+        );
+        CachingResolver::new(40, r)
+    }
+
+    pub fn environment_resolver(
+        target_env: TargetEnv,
+        alias: HashMap<String, String>,
```
kdy1, line 76: Same here.

kdy1 (changes requested):
Mostly looks good. Thanks!
I have some small changes to propose.

tmpfs:
Cool @kdy1, all done :+1:

kdy1 (approved):
LGTM. Thank you!

Issue #2046, opened 2021-08-10
Support exports mapping in package.json to determine entry points


tmpfs:
See the [node exports documentation](https://nodejs.org/dist/latest-v16.x/docs/api/packages.html#packages_package_entry_points).

The interesting thing about implementing this is that it *excludes access* to nested imports for files that are not explicitly mapped as exported by the package. Also conditional exports:

```json
{
  "main": "./main-require.cjs",
  "exports": {
    "import": "./main-module.js",
    "require": "./main-require.cjs"
  },
  "type": "module"
}
```

Would mean that `NodeModulesResolver::resolve` would now need to know whether the calling context is ESM or CommonJS in order to choose the appropriate file.

flickyiyo:
Does this has work already done?

PR #2060, opened 2021-08-12
Browser variants

tmpfs:
Improve the handling of the `browser` field in `package.json` to conform to [this specification](https://github.com/defunctzombie/package-browser-field-spec).


kdy1:
Is this actually used by libraries?

tmpfs:
@kdy1, this starts the process of updating the handling of the `browser` field to cover all variants.

Beforehand, I made a change that means that if an error occurs deserializing `package.json` it is no longer swallowed but now appears in the error context:

```
Caused by:
    0: failed to analyze module
    1: failed to resolve util from /home/muji/git/consensys/swc-paths-mvp/src/index.js
    2: failed to deserialize package.json
    3: invalid type: boolean `true`, expected a string at line 38 column 14
```

Which is important I think because this should only happen if a) the file is corrupted or b) we make an error handling the possible variants for deserialization.

Because my initial handling of `browser` was too simple and the deserialization error was being masked it took a while to track down the issue.

Then I added support for object type but we need to figure out how to handle the boolean `false` which should map to an empty module exports. Would appreciate your guidance on the best way to do that!

tmpfs:
> Is this actually used by libraries?

Yes, here is an example of the [object type](https://github.com/browserify/node-util/blob/master/package.json#L9-L11) and [readable-stream](https://github.com/nodejs/readable-stream/blob/master/package.json#L56-L63) uses a boolean `false` for the object value.

These packages are very widely used and I am sure there are many other packages using `browser` like this.

kdy1:
I think `FileName::Custom` with a special string can be used.
https://docs.rs/swc_common/0.11.4/swc_common/enum.FileName.html

tmpfs:
> I think `FileName::Custom` with a special string can be used.
> https://docs.rs/swc_common/0.11.4/swc_common/enum.FileName.html

Nice, let me look into that, thanks :+1:

kdy1:
Note: I renamed `spack` to `swc_node_bundler` and it now lives in `/node/bundler`

tmpfs:
Thanks for the heads up @kdy1, I will keep pulling master into this branch.

So I am getting closer with handling the `browser` field correctly but still an issue with the boolean `false` value. When I return `FileName::Custom` this gets transformed to ESM `import` statements which gives us this error in the browser:

```
Uncaught SyntaxError: Cannot use import statement outside a module
```

I think we may need another variant on `FileName` to handle this situation or we need to update the transformation for `FileName::Custom`.

Also I thought it would be a good idea to cache the parsed `package.json` data for when modules are referenced multiple times.

Be good if you can take a look at the changes so far and provide feedback, thanks :+1:

kdy1:
You should handle the special string within the Loader

kdy1:
For caching, CachingLoader will handle it.

tmpfs:
> For caching, CachingResolver will handle it.

Ok, I have removed the caching of `PackageJson` :+1: 

I think we need a new variant for `FileName` to complete this as the handling for builtin modules is not identical to when modules are disabled in the `browser` field.

For node builtins we still want the import statements (or `require` calls) but for modules that are disabled the locals symbols should be mapped to the empty object.

Can I add a variant like `FileName::Stub` and implement handling for disabled modules in `SwcLoader`? 

Alternatively we could add a flag to the `FileName::Custom` variant to distinguish this handling?

kdy1:
I'd prefer not breaking `swc_common`.
Can't you use `node:http` or something like that? Using `FileName::Macro` may also be viable, considering those imports are special and handled internally.

tmpfs:
Ok thanks @kdy1, i am going with the `node:` prefix for now.

I am still seeing a situation where imports of `events` and `assert` are being treated as external modules, so I think there are other places in the code I need to handle builtins only when `TargetEnv::Node`. Will investigate.

tmpfs:
Hi @kdy1, I think we can merge this now. It gets me past that `SyntaxError` and on to the next error :grin: 

I am not so happy about this code: https://github.com/tmpfs/swc/blob/browser-variants/node/bundler/src/loaders/swc.rs#L38-L80

As it also needs to handle the `export default {}` case, but we need to know whether the calling context is ESM or CJS to implement that (we would also need the calling context for #2046).

kdy1:
I agree that code is not looking great.
Maybe I can add some variants to `FileName` when I have to make a breaking change for `swc_common`.
I'm going to redesign comment system, and it means a breaking change of `swc_common`. 

I filed https://github.com/swc-project/swc/issues/2088

Thanks! I'll review this PR after handling some other tasks.

ecmascript/ast/src/lib.rs
```diff
@@ -142,3 +142,72 @@ impl Default for TargetEnv {
         TargetEnv::Browser
     }
 }
+
+/// List of built in packages for latest stable node with LTS (node@16).
+///
+/// Run `node -p "require('module').builtinModules"`
+pub const NODE_BUILTINS: [&'static str; 63] = [
```
kdy1, line 149: I think it is not a good place for this varianble.
Can you move this to `swc_ecma_loader/node`?

kdy1: Also, I think `&[&str]` is better because it reduces breaking changes.

kdy1 (changes requested):
I think ast crate should have minimal breaking changes because it will be used for plugins.

ecmascript/ast/src/lib.rs
tmpfs: Cool, I put it at `swc_ecma_loader::NODE_BUILTINS` because `swc_ecma_loader::resolvers::node` is behind a feature flag and this should be accessible without the flag IMO. I wonder whether putting `NodeModulesResolver` behind a feature flag is strictly necessary?

kdy1: > I wonder whether putting NodeModulesResolver behind a feature flag is strictly necessary?

Yes, but it's not necessary for `NODE_BUILTINS`.

I want to avoid regressing deno, but single const will be easily stripped out by the compiler, without regressing compile time by a margin.

kdy1 (approved):
LGTM. Thanks!

PR #2068, opened 2021-08-13
perf(node/swc): switch to mimalloc

Brooooooklyn:
### v1.2.76

```
@swc-node/core x 396 ops/sec ±1.83% (90 runs sampled)
esbuild x 465 ops/sec ±1.00% (88 runs sampled)
typescript x 26.83 ops/sec ±10.61% (52 runs sampled)
babel x 29.41 ops/sec ±9.19% (55 runs sampled)
Transform rxjs/AjaxObservable.ts benchmark bench suite: Fastest is esbuild
@swc-node/core x 1,023 ops/sec ±1.84% (72 runs sampled)
esbuild x 948 ops/sec ±1.17% (80 runs sampled)
Transform rxjs/AjaxObservable.ts parallel benchmark bench suite: Fastest is @swc-node/core
```

### This commit

```
@swc-node/core x 436 ops/sec ±0.57% (91 runs sampled)
esbuild x 468 ops/sec ±1.67% (87 runs sampled)
typescript x 27.01 ops/sec ±11.05% (51 runs sampled)
babel x 29.40 ops/sec ±8.73% (55 runs sampled)
Transform rxjs/AjaxObservable.ts benchmark bench suite: Fastest is esbuild
@swc-node/core x 1,089 ops/sec ±1.25% (75 runs sampled)
esbuild x 967 ops/sec ±1.14% (82 runs sampled)
Transform rxjs/AjaxObservable.ts parallel benchmark bench suite: Fastest is @swc-node/core
```


kdy1:
Wow, it's a big difference.
Is mimalloc always a win for any target?

Brooooooklyn:
It should be compatible for all x86  platforms and macOS-m1.

kdy1 (approved):
LGTM. Thanks!

kdy1:
Ignore the docs ci. I'll fix it with another PR after merging this.

Brooooooklyn:
`cargo +$(cat rust-toolchain) doc` should work

```diff
    uses: actions-rs/toolchain@v1
        with:
          profile: minimal
--        toolchain: nightly-2021-03-30
```

This line could be removed too, because `actions-rs/toolchain` supported the `rust-toolchain` file.

kdy1:
Yeah, you are correct. I'll remove the line.
Btw, is there any reason you used `mimalloc-rust ` instead of `mimalloc`?

Brooooooklyn:
> The `mimalloc` had to dynamically check if the memory type was align every time it was allocated, but now I've taken a closer look at Rust's ABI and found that all Rust types seem to be Align, so I'll just use Align.
> The second reason is that the previous library may cause memory errors because there are many data types that are not covered, especially when the extend function is used.

/cc @LemonHX

Issue #2087, opened 2021-08-16
Support environment variables that map to globals (spack)


tmpfs:
Support passing environment variables using a command line option that are transformed into global variables when building with **spack**.

This is typically used to switch backend API but has many uses beyond that. I suggest adding the `--env` option to spack which would take:

```
spack --env API_URL=http://localhost:3000
```

When compiling spack would define a `const` declaration for each environment variable at the top-level scope of the bundle:

```javascript
const API_URL = 'http://localhost:3000'
```

If a variable name contains a period it should be expanded to an object literal:

```
spack --env process.env.FOO=bar
```

Should yield:

```javascript
const process = {
  env: {FOO: 'bar'}
};
```

@kdy1, if you are happy with this feature request I can have a go at the implementation, thanks! :+1:

kdy1:
Yes, I think this is super useful, especially if someone can configure this from javascript.

tmpfs:
It is possible to do this using `jsc.transform.optimizer.globals.envs`, see #2093 and #2121.

PR #2093, opened 2021-08-17
Env option

tmpfs:
Initial preparation to support mapping environment variables to bundle globals.

See: https://github.com/swc-project/swc/issues/2087


tmpfs:
Hi @kdy1, I think this is ready to be merged if you approve :+1: 

We can now configure an `env` map in `spack.config.js` and extend/override them with the `--env` option (see https://github.com/swc-project/cli/pull/54).

These variables are converted into a tree structure in `EnvironmentGlobals` and then folded into the module in `finalize()`. When building multiple bundles they would be injected into each bundle.

kdy1:
I think this is already possible using `swc` option.
See: https://swc.rs/docs/configuring-swc/#jsctransformoptimizer

```json
{
  "jsc": {
    "transform": {
      "optimizer": {
        "globals": {
          "vars": {
            "__DEBUG__": "true"
          }
        }
      }
    }
  }
}
```

Although it's not documented, it acutally supports inlining environment variables.


```json
{
  "jsc": {
    "transform": {
      "optimizer": {
        "globals": {
          "vars": {
            "__DEBUG__": "true"
          },
          "env": ["NODE_ENV"]
        }
      }
    }
  }
}
```

should inline `process.env.NODE_ENV` if I remember it correctly.

tmpfs:
Thanks @kdy1, setting `envs` doesn't seem to be work in my test. Maybe I am doing something wrong?

Here is the entire config: 

```javascript
const path = require('path');
const { config } = require('@swc/core/spack')

module.exports = config({
  entry: {
    background: __dirname + '/app/scripts/background.js',
  },
  output: {
    path: path.join(__dirname, "dist", "chrome"),
  },
  module: {},
  alias: {
    browser: {
      fs: "localstorage-fs",
      constants: "constants-browserify",
      stream: "stream-browserify",
      path: "path-browserify",
      os: "os-browserify",
      crypto: "crypto-browserify",
      vm: "vm-browserify",
      http: "stream-http",
      https: "https-browserify",
      zlib: "browserify-zlib",
      tty: "tty-browserify",
    }
  },
  options: {
    jsc: {
      parser: {
        syntax: "ecmascript",
        jsx: true,
        dynamicImport: false,
        privateMethod: false,
        functionBind: false,
        exportDefaultFrom: false,
        exportNamespaceFrom: false,
        decorators: false,
        decoratorsBeforeExport: false,
        topLevelAwait: false,
        importMeta: false
      },
      transform: {
        optimizer: {
          globals: {
            envs: ['INFURA_PROJECT_ID']
          }
        }
      },
      target: "es2016",
      loose: false,
      externalHelpers: false,
      keepClassNames: false
    }
  }
});
```

Then I invoke `spack` with the environment variable:

```
INFURA_PROJECT_ID=1234 npx spack
```

And I am back to the `ReferenceError` I had before:

```
Uncaught ReferenceError: process is not defined
```

tmpfs:
Ok, I took a look at the code and it seems like when `env_map()` is called in `SwcLoader` to setup `InlineGlobals` it only processes `NODE_ENV`. Should I modify that to accept any name declared in `transform.optimizer.globals.envs` instead? I think that will fix it.

kdy1:
Yes, I think it's the right way to go.

tmpfs:
Closing in favor of #2121.

PR #2119, opened 2021-08-20
refactor: use dyn instead of impl in analyze_dependencies

dsherret:
It's easier to pass around these references as `dyn` rather than `impl` because then no type parameters are necessary. Do you think this is ok to do? I'm not sure about what the trade offs are.

### DIFF
--- a/ecmascript/dep-graph/src/lib.rs
+++ b/ecmascript/dep-graph/src/lib.rs
@@ -9,7 +9,7 @@ use swc_ecma_visit::{self, Node, Visit, VisitWith};
 
 pub fn analyze_dependencies(
     module: &ast::Module,
-    comments: &impl Comments,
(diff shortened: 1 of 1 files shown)


ecmascript/dep-graph/src/lib.rs
```diff
@@ -9,7 +9,7 @@ use swc_ecma_visit::{self, Node, Visit, VisitWith};
 
 pub fn analyze_dependencies(
     module: &ast::Module,
-    comments: &impl Comments,
+    comments: &dyn Comments,
```
dsherret, line 12: Are there performance implications for doing this? Maybe we should leave it as impl? I'm ok either way.

dsherret:
Actually, I think this is ok for us how it is and I think dyn might be less performant.

PR #2121, opened 2021-08-21
Support custom values in global envs map.

tmpfs:


tmpfs:
@kdy1, I am tempted to remove the special handling for `NODE_ENV`, what do you think?

kdy1:
> I am tempted to remove the special handling for NODE_ENV, what do you think?
I think it's correct way to go, but it should allow inlining only env vars from some allowlist.
Otherwise, it's a security issue.

That's why I only allowed `NODE_ENV`.

tmpfs:
But isn't the `transform.optimizer.globals.envs` the whitelist that the developer provides or am i misunderstanding?

kdy1:
> But isn't the transform.optimizer.globals.envs the whitelist that the developer provides or am i misunderstanding?

You are correct, `transform.optimizer.globals.envs` should act as an allowlist.

tmpfs:
Ok great, then I think this is good to be merged if you approve :+1: 

Because I removed the special handling for `NODE_ENV` anyone relying on the earlier behavior would now need to explicitly add `NODE_ENV` to the `envs` array. We should really be keeping a changelog of these breaking changes somewhere!

kdy1 (approved):
LGTM. Thanks!

PR #2172, opened 2021-08-28
Fix bug resolving node modules.

tmpfs:
Incorrectly passing the parent of the base directory would cause
some modules not to be found.


kdy1:
Thanks!

PR #2179, opened 2021-08-30
Fix for relative path resolution.

tmpfs:
Incorrectly using the parent directory instead of the base when
resolving relative specifiers.


tmpfs:
Sorry @kdy1, I should have caught this with the last fix!

tmpfs:
This was an error on my part actually, I was trying to resolve a relative file `./runtime.js` and using:

```rust
let base_dir = FileName::Real(std::env::current_dir()?);
```

When I needed to create a mock *filename* to resolve correctly:

```rust
let base_dir = FileName::Real(std::env::current_dir()?.join("index.js"));
```

PR #2182, opened 2021-08-30
Node base directory handling

tmpfs:
This allows passing `base_dir` as either a directory or a file and the behavior should be correct either way.

Typically when walking the module graph we have a *calling file* that performs the `import` (or `require`) so it makes sense to pass the file and automatically use the parent however there are times when it is convenient to resolve packages outside of the module graph and then it is more appropriate to use `std::env::current_dir()` for `base_dir`.


kdy1 (approved):
Thanks!

PR #2188, opened 2021-08-30
refactor: static blocks - use a block stmt instead of stmts vector and fix span

dsherret:
This change makes the `StaticBlock` node use a `BlockStmt` instead of `Vec<Stmt>` for the statements. This is more in line with the other class nodes (ex. `Constructor`) and aligns with what is done in the TypeScript compiler:

image

It will also make it easier for code formatting purposes to be able to get the span/range of the block.

Additionally, this fixes the span of `StaticBlock`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -939,7 +939,7 @@ dependencies = [
 
 [[package]]
 name = "jsdoc"
-version = "0.36.0"
+version = "0.37.0"
 dependencies = [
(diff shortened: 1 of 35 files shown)


kdy1 (approved):
Thanks! I'll bump version and merge it.

sosukesuzuki (commented):
Thank you!

Issue #2201, opened 2021-09-03
Support `FileName::Url(Url)` enum


kitsonk:
**Describe the feature**

In Deno, we have always used `FileName::Custom(String)` as the enum variant for our source maps, while in fact, everywhere else we use [Url](https://crates.io/crates/url/)s.  This isn't a "big" problem, but it does mean when we serialise a source map for a JSON representation of a source map (like to a base64 encoded string, or an external `.map` file) all of our sources get braced with `<>`.  This means that when they get loaded into a debugger, they get a bit strange, while in fact they are always fully qualified URLs, and it seems all debugger clients handle them just fine.

**Additional context**

I will raise a PR in hopes that there is no objection to it.

kdy1:
I'm fine with it, but please add `FileName::Internal(String)` too.
It's required but I didn't change it because it's a breaking change.

PR #2202, opened 2021-09-03
feat(common/syntax_pos): FileName supports URLs

kitsonk:
This allows URLs to be represented as file names without being modified.  For platforms, like Deno, that handle sources a "native" URLs this is better than converting the URL to a `FileName::Custom`.

Closes: #2201


kdy1 (approved):
Thanks!

PR #2227, opened 2021-09-10
fix(es/loader): Fix support for `jsc.paths`.

### LINKED ISSUE #2050: Path aliases don't seem to respect nested folder structure
**Describe the bug**
After https://github.com/swc-project/swc/issues/1934 and https://github.com/swc-project/swc/issues/1935 were resolved in `1.2.76`, I gave it a quick try and run into another issue.

It seems to me that `paths` simply rewrites the path prefix but doesn't respect the location of given file. I also tried using `baseUrl` but no luck. Am I missing some configuration?

**Input code**
Using `~` as an alias for `src` and having a file `src/subfolder/A.ts` with an `import { B } from '~/subfolder/B.ts';` (and considering the `src/subfolder/B.ts` exists and exports `B`), we'll run into `Error: Cannot find module './subfolder/B'`.

This is because the import was rewritten to `./subfolder/B` instead of just `./B` or `../subfolder/B` (that should be correct as these files are in the same subfolder indeed).

The command I'm running is: `yarn swc src -s -d ./dist`

**Config**
Note that `baseUrl` probably has no effect but I tried to add it to conform with `tsconfig.json`.

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
    },
    "target": "es2020",
    "baseUrl": ".",
    "paths": {
      "~/*": ["./*"]
    }
  },
  "module": {
    "type": "commonjs"
  }
}
```

Also note that in TypeScript the correct path in the config would be `"~/*": ["./src/*"]` in this case (including the `src` portion) since `baseUrl` is `.`. With SWC, however, I need to use just `./*`, probably because I'm passing `src` dir to the `swc` command. 

**Expected behavior**
Paths should be rewritten using current file location (not just using a static prefix).
In the given example the import from `'~/subfolder/B.ts` was rewritten to `./subfolder/B` instead of just `./B` (or `../subfolder/B`).

**Version**
The version of @swc/core: 1.2.76

**Additional context**
TypeScript also uses `baseUrl` to set the base path relative to which imports are rewritten.

kdy1:
swc_ecma_loader:
 - `TsConfigResolver`: Use `baseUrl`. (Closes #2050)

swc:
 - Change type of `JscConfig.base_url` to `PathBuf`.

testing:
 - Improve `NormalizedOutput::compare_to_file`.


stefee:
I am also still able to repro this issue (https://github.com/swc-project/swc/issues/2050#issuecomment-917684787)

PR #2230, opened 2021-09-10
feat: Dynamic swc.parser.tsx by filename in 4 node binding API

IronLu233:
From https://github.com/swc-project/swc/discussions/1916. but it seems not work in `transformFile`, `transformFileSync`, `transform`, `transformSync` APIs

Now I implement this feature.


kdy1 (approved):
LGTM. Thanks!

PR #2231, opened 2021-09-10
fix: missing new.target transform in es5

IronLu233:
https://github.com/swc-project/swc/pull/2129 implement it but not add it to transform es5


kdy1 (approved):
Thanks! I can't believe I forgot adding it. 👍

PR #2270, opened 2021-09-18
fix: Make crate features with global effects optional.

losfair:
The `swc` crate currently has two features that cause issues when used as a library:

- `log/release_max_level_info`. This feature silences `debug` and `trace` logs globally in *all* crates that depend on the `log` crate.
- `swc_node_base` which depends on the `mimalloc-rust` crate and sets a custom allocator. Users of the `swc` crate may want to provide their own allocator, and the global allocator feature requires a nightly compiler.

This pull request makes these two features optional, while leaving them enabled by default to preserve the current behavior.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ base64 = "0.13.0"
 dashmap = "4.0.2"
 either = "1"
 fxhash = "0.2.1"
-log = {version = "0.4", features = ["release_max_level_info"]}
+log = "0.4"
 lru = "0.6.1"
 once_cell = "1"
 pathdiff = "0.2.0"
@@ -74,13 +74,13 @@ swc_ecma_transforms_base = {version = "0.31.1", path = "./ecmascript/transforms/
 swc_ecma_utils = {version = "0.44.2", path = "./ecmascript/utils"}
 swc_ecma_visit = {version = "0.38.1", path = "./ecmascript/visit"}
 swc_ecmascript = {version = "0.63.1", path = "./ecmascript"}
-swc_node_base = {version = "0.3.0", path = "./node/base"}
 swc_visit = {version = "0.2.3", path = "./visit"}
 
 [dev-dependencies]
 rayon = "1"
 testing = {version = "0.13.0", path = "./testing"}
 walkdir = "2"
+swc_node_base = {version = "0.3.0", path = "./node/base"}
 
 [[example]]
 name = "usage"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -26,7 +26,7 @@ version = "0.54.0"
 name = "swc"
 
 [features]
-default = ["es3"]
+default = ["es3", "filter-logs", "custom-allocator"]
```
kdy1, line 29: Please remove these features completely and move them to `node/binding`.

Cargo.toml
```diff
@@ -36,6 +36,10 @@ es3 = []
 #
 # This feature will be removed on v2 of node bindings.
 wrong-target = []
+# Enable this feature to disable logs below the `info` level in release mode.
+filter-logs = ["log/release_max_level_info"]
+# Enable custom allocator. This may improve runtime performance but requires a nightly compiler.
+custom-allocator = ["swc_node_base"]
```
kdy1, line 42: This should be removed.

Cargo.toml
```diff
@@ -74,7 +78,7 @@ swc_ecma_transforms_base = {version = "0.31.1", path = "./ecmascript/transforms/
 swc_ecma_utils = {version = "0.44.2", path = "./ecmascript/utils"}
 swc_ecma_visit = {version = "0.38.1", path = "./ecmascript/visit"}
 swc_ecmascript = {version = "0.63.1", path = "./ecmascript"}
-swc_node_base = {version = "0.3.0", path = "./node/base"}
+swc_node_base = {version = "0.3.0", path = "./node/base", optional = true}
```
kdy1, line 81: Please move this to `dev-dependencies`. It's required to get correct profiling result.

kdy1: Also, please make it not optional.

kdy1 (changes requested):

kdy1:
I'm curious about the usecase. Can you share it?
Initially, I didn't expect any user to depend on `swc` crate.

losfair:
I'm using `swc` to compile TypeScript files to JavaScript as part of the build process in [rusty-workers](https://github.com/losfair/rusty-workers) v2 (the specific branch is WIP and isn't public yet). The usage looks like:

```rust
let compiler = Compiler::new(Arc::new(SourceMap::default()));
// ... prepare source, handlers and opts
let out = compiler.process_js_file(source, &handler, &opts).unwrap();
```

Is there any better ways to do that? It looks like the parser and codegen are there but needs some glueing to put it together so I just depended on `swc` directly.

kdy1 (approved):
LGTM. Thanks!

PR #2271, opened 2021-09-18
fix: emit static before readonly

dsherret:
Fixes:

image

I will add a test in a second. Edit: Actually, I couldn't get it to pick it up and gave up quickly 😅


kdy1 (approved):
Thank you!

PR #2272, opened 2021-09-19
fix(es/minifier): Don't drop used variables

### LINKED ISSUE #2262: minify remove necessary code
Example:

https://github.com/SoloJiang/minify-remove-necessary-code

You can start a http-server in the lib dir, and see `esm is not defined`, but you can see the source, esm is defined.

fabiosantoscode:
This reproduces #2262, though I'm not sure how to fix

---

Closes #2262

### DIFF
--- a/ecmascript/minifier/tests/compress/fixture/issues/2262/1/input.js
+++ b/ecmascript/minifier/tests/compress/fixture/issues/2262/1/input.js
@@ -10,11 +10,6 @@
 
     const esm = index;
 
-    var createNamedContext = function createNamedContext() {
-        esm();
-    };
-
-    createNamedContext();
-    createNamedContext();
+    esm();
 })()
 
--- a/ecmascript/minifier/tests/compress/fixture/issues/2262/1/input.js
+++ b/ecmascript/minifier/tests/compress/fixture/issues/2262/1/input.js
@@ -1,0 +1,17 @@
+(() => {
+    "use strict";
+
+    var commonjsGlobal = globalThis
+
+    var index = somethingGlobal;
+
+    const esm = index;
+
+    var createNamedContext = function createNamedContext() {
+        esm();
+    };
+
+    createNamedContext();
+    createNamedContext();
+})()
+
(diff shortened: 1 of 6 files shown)


kdy1:
Thank you a lot!
Reducing the error case is the hardest part.
May I send some PR to your fork?

ecmascript/minifier/tests/compress/fixture/issues/2262/1/input.js
```diff
@@ -0,0 +1,20 @@
+(() => {
+    "use strict";
+
+    var commonjsGlobal = globalThis
+
+    function createEventEmitter(value) {
+    }
```
fabiosantoscode, line 7: Removing this `createEventEmitter` or `commonjsGlobal` (both of which are unused) will cause the bug not to reproduce. No idea why, though

ecmascript/minifier/tests/compress/fixture/issues/2262/1/input.js
```diff
@@ -0,0 +1,20 @@
...
+
+    var index = somethingGlobal;
+
+    const esm = index;
```
fabiosantoscode, line 11: This indirection is important to the bug. `const esm = somethingGlobal `wouldn't reproduce it either.

fabiosantoscode:
> Reducing the error case is the hardest part.

I have lots of practice by now :D

fabiosantoscode:
Invited you to my fork @kdy1

kdy1:
Thanks!

fabiosantoscode:
Also @kdy1 is there a practical way to repeatedly run the SWC minifier against something? I managed to do that by repeatedly using the `./scripts/run.sh` with a written test, but I'm really used to the flow I had in Terser, where I'd have some file (usually bug.js) and then run Terser repeatedly on it using the CLI

fabiosantoscode:
Maybe I can contribute a `[bin]` target for `ecmascript/minifier` which could emulate the terser CLI options

kdy1:
Currently there's no way to do so, but a binary seems like a good idea to me.

PR #2273, opened 2021-09-19
CLI for minifier

fabiosantoscode:
Intended for development, and adapted from the minifier example (though not using the `testing` crate).


kdy1 (approved):
LGTM

PR #2280, opened 2021-09-21
fix(css/parser): url parsing

alexander-akait:
We can write `url(   "./image.png"   )` (whitespaces after `(` and before `)` )


css/parser/tests/fixture/function/url/span.rust-debug
```diff
@@ -0,0 +1,437 @@
...
+ --> $DIR/tests/fixture/function/url/input.css:7:27
+  |
+7 |     background-image: url('./image (1).jpg'   );
+  |                           ^^^^^^^^^^^^^^^^^^^^
+
+error: SpaceValues
+ --> $DIR/tests/fixture/function/url/input.css:7:27
+  |
+7 |     background-image: url('./image (1).jpg'   );
+  |                           ^^^^^^^^^^^^^^^^^^^^
```
kdy1, line 306: Is this intentional?
The property value should be [`UrlValue`](https://docs.rs/swc_css_ast/0.5.0/swc_css_ast/struct.UrlValue.html), not `SpaceValues`.

kdy1 (changes requested):
Thanks!
I found one problem, and that's why CI is failing.

alexander-akait:
@kdy1 Fixed, anyway here interesting case, we have:
```
a {
    color: var(   --color  );
}
```

our output:
```
{
  "type": "Stylesheet",
  "span": {
    "start": 0,
    "end": 35,
    "ctxt": 0
  },
  "rules": [
    {
      "type": "StyleRule",
      "span": {
        "start": 0,
        "end": 35,
        "ctxt": 0
      },
      "selectors": [
        {
          "type": "ComplexSelector",
          "span": {
            "start": 0,
            "end": 1,
            "ctxt": 0
          },
          "selectors": [
            {
              "type": "CompoundSelector",
              "span": {
                "start": 0,
                "end": 1,
                "ctxt": 0
              },
              "hasNestPrefix": false,
              "combinator": null,
              "typeSelector": {
                "type": "NamespacedName",
                "span": {
                  "start": 0,
                  "end": 1,
                  "ctxt": 0
                },
                "prefix": null,
                "name": {
                  "type": "Text",
                  "span": {
                    "start": 0,
                    "end": 1,
                    "ctxt": 0
                  },
                  "value": "a"
                }
              },
              "subclassSelectors": []
            }
          ]
        }
      ],
      "block": {
        "type": "DeclBlock",
        "span": {
          "start": 2,
          "end": 35,
          "ctxt": 0
        },
        "properties": [
          {
            "type": "Property",
            "span": {
              "start": 8,
              "end": 32,
              "ctxt": 0
            },
            "name": {
              "type": "Text",
              "span": {
                "start": 8,
                "end": 13,
                "ctxt": 0
              },
              "value": "color"
            },
            "values": [
              {
                "type": "FnValue",
                "span": {
                  "start": 15,
                  "end": 32,
                  "ctxt": 0
                },
                "name": {
                  "type": "Text",
                  "span": {
                    "start": 15,
                    "end": 18,
                    "ctxt": 0
                  },
                  "value": "var"
                },
                "args": [
                  {
                    "type": "SpaceValues",
                    "span": {
                      "start": 22,
                      "end": 31,
                      "ctxt": 0
                    },
                    "values": [
                      {
                        "type": "Text",
                        "span": {
                          "start": 22,
                          "end": 29,
                          "ctxt": 0
                        },
                        "value": "--color"
                      }
                    ]
                  }
                ]
              }
            ],
            "important": null
          }
        ]
      }
    }
  ]
}
```

As you can see we have `SpaceValues` in `args`, in fact, this is not a problem not only `url` functions, affected all functions, I think we should not put `SpaceValues` for all functions

alexander-akait:
@kdy1 Do we have something like `cargo test --test fixture --update`? Changing multiple files (snapshots) requires a lot of time

alexander-akait:
Commented broken cases, I will uncomment them in the next PR + more tests with functions like `var`/`env`/etc

kdy1 (approved):
LGTM. Thanks!

PR #2285, opened 2021-09-22
chore: ignore linguist detection for fixtures

g-plane:
swc is a Rust project, but fixtures are written in JS or TS, so GitHub Linguist detected them as source files (as the screenshot below). This PR let it be ignored.

image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,3 +8,6 @@
 *.js    text eol=lf merge=union 
 *.json  text eol=lf merge=union 
 *.debug text eol=lf merge=union 
+
+**/tests/**/*.js linguist-detectable=false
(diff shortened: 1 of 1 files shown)


alexander-akait:
hm, why is that bad?

kdy1 (approved):
I always wanted this feature but I didn't know it.
Thank you a lot!

kdy1:
@g-plane It seems like it's not working.

@alexander-akait swc contains really lots of typescript test files.

alexander-akait:
@kdy1 I think we need time until github recalculate files, don't think it happens immediately

g-plane:
I guess GitHub won't re-index again until actual source files changed. Not sure.

g-plane:
It works now:
图片

PR #2286, opened 2021-09-22
fix(css/parser): parsing `!important`

alexander-akait:


css/parser/src/parser/at_rule.rs
```diff
@@ -27,7 +27,7 @@ where
             }
         };
 
-        match &*name.to_lowercase() {
+        match &*name.to_ascii_lowercase() {
```
alexander-akait, line 30: My rust knowledge is still very young (if you known good books, I will be glad your recommendations), but `to_ascii_lowercase` is valid, because at-rules can be written using non ascii characters

alexander-akait:
Also `a { color: white !bad; }` in theory valid CSS, give me time to check it using spec

css/parser/tests/fixture/important/basic/output.json
```diff
@@ -0,0 +1,676 @@
...
+                  },
+                  "value": 1.0
+                },
+                "unit": {
+                  "span": {
+                    "start": 64,
+                    "end": 66,
+                    "ctxt": 0
+                  },
+                  "kind": null
```
alexander-akait, line 172: Why we don't write kind of unit in tests?

alexander-akait:
Using https://tomhodgins.com/demo/cssom/:
```css
a { color: red !bad; }
```

chrome doesn't parse that declaration, I don't have strong opinion, maybe we can revisit it late

alexander-akait:
Found many bugs while adding tests, postpone failed tests for future fixes, `url` parsing will be fixed in near future (as I promised)

css/parser/src/parser/at_rule.rs
kdy1: Yes, you are right.

css/parser/tests/fixture/important/basic/output.json
kdy1: I think it's a bug.

alexander-akait: :+1:

css/parser/tests/errors/important/basic/output.stderr
```diff
@@ -0,0 +1,6 @@
+error: Expected important
+ --> $DIR/tests/errors/important/basic/input.css:2:20
+  |
+2 |     color: white !!important;
```
kdy1, line 4: This seems not ideal. I think it span should include two bangs.

kdy1 (changes requested):
Thanks!
I found some small issues.

css/parser/src/parser/style_rule.rs
```diff
@@ -99,7 +103,18 @@ where
         let start = self.input.cur_span()?.lo;
 
         if !is!(self, EOF) && eat!(self, "!") {
-            expect!(self, "important");
+            self.input.skip_ws()?;
+
+            let is_important = match bump!(self) {
+                Token::Ident(value) => &*value.to_ascii_lowercase() == "important",
```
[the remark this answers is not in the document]
alexander-akait: hm, yep, because no `!important`, `important = true` should be only for `!important`, https://www.w3.org/TR/css-syntax-3/#!important-diagram, in theory we can change ast and output `important: true | false | "ident-token"`, i.e. for your example it will be `important: "foo"`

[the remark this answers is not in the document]
alexander-akait: If we want to allow to using parser for linting purposes it will be useful

css/parser/tests/errors/important/basic/output.stderr
alexander-akait: Do you mean?
```
color: white !!important;
             ^^^^^^^^^^^
```
  
First `!` is valid here, maybe better show only `!important`?

alexander-akait: Updates, I think we should output `Expected !important`, because `important` is just word

css/parser/src/parser/style_rule.rs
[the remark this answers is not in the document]
alexander-akait: @kdy1 friendly ping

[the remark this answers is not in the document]
kdy1: I think a warning for `red!foo` would be great, but can be fixed with follow-up PRs.

kdy1 (approved):
LGTM. Thanks!

css/parser/src/parser/style_rule.rs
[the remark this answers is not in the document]
alexander-akait: Yep, I think we can improve not valid parsing in future - after feedback, right now let's focus on parsing problems and ast

PR #2291, opened 2021-09-23
feat(node/swc): support for named export `Visitor`

xnuk:
In some environment, default exporting can be problematic.

### DIFF
--- a/node-swc/src/Visitor.ts
+++ b/node-swc/src/Visitor.ts
@@ -160,7 +160,7 @@ import {
   TsConstAssertion,
 } from "./types";
 
-export default class Visitor {
+export class Visitor {
   visitProgram(n: Program): Program {
     switch (n.type) {
       case "Module":
@@ -1723,3 +1723,5 @@ export default class Visitor {
     return n;
   }
 }
+
+export default Visitor;


kdy1 (changes requested):
Is this useful?
I'm not sure about this.

xnuk:
Currently `@swc/core/Visitor.js` is detected as **CommonJS module** by Node.js' ESM mode (`.mjs`, I know swc doesn't support ESM config files yet, but anyway), and it looks like this:

```javascript
"use strict";
class Visitor {
  // ...
}

exports.default = Visitor;
```

And following this:

> The ECMAScript Module Namespace representation of a CommonJS module is always a namespace with a `default` export key pointing to the CommonJS `module.exports` value.
>
> - [Node.js 16 documentation: "CommonJS Namespaces" section in "ECMAScript modules"](https://nodejs.org/dist/latest-v16.x/docs/api/esm.html#esm_commonjs_namespaces)

I need some workaround to import `Visitor` since the default export would be `{ default: Visitor }`:

```javascript
import SwcVisitor from '@swc/core';
const Visitor = SwcVisitor.default;
```

And named export would cause less problem I think?:

```javascript
import { Visitor } from '@swc/core';
```

kdy1 (commented):
I'm not sure if `Visitor` itself is useful, and I'm not sure if it's right to encourage usage of it.

kdy1 (approved):
I decided to merge this.
Thanks!

PR #2295, opened 2021-09-23
test(css/parser): quotes in value

alexander-akait:
Just tests for edge cases, all works fine here


alexander-akait:
We have bug :smile:

kdy1:
@alexander-akait Current css parser is quite naive, and there are really lots to do.
I hope I have more time.

alexander-akait:
@kdy1 don't worry, I will help you

kdy1:
Thank you a lot!

alexander-akait:
We really need good CSS parser, currently parser is not good, to be honestly same for HTML, but I think we will do it late

alexander-akait:
@kdy1 fixed and completed

Notes:
- in future we should covert `fn read_str(&mut self) -> LexResult<JsWord>` -> `fn read_str(&mut self) -> LexResult<Token>` and add `BadStringToken`, it allow do recover mode


- logic is some mixed, for example `read_escape` - https://github.com/swc-project/swc/blob/master/css/parser/src/lexer/mod.rs#L313

For spec:
> 4.3.7. Consume an escaped code point
This section describes how to consume an escaped code point. It assumes that the U+005C REVERSE SOLIDUS (\) has already been consumed and that the next input code point has already been verified to be part of a valid escape. It will return a code point.

But we use `self.input.eat_byte(b'\\')`, `\` should be already eaten, so we need do check/etc before run `read_escape`, it allow to have more pure logic

- we need more helpers like `is_newline`/`is_whitespace` and etc, to have logic in one place

I will improve this in other PRs, here just a fix, better to add more tests with edge cases and do refactor latest, it allow to prevent regressions

css/parser/src/lexer/value.rs
```diff
@@ -67,56 +67,57 @@ where
...
         loop {
-            if !was_escape {
-                let chunk = self.input.uncons_while(|peeked| match peeked {
-                    '\'' | '"' => peeked != quote,
-                    '\\' => false,
-                    _ => true,
-                });
-
-                buf.push_str(chunk);
-            }
```
alexander-akait, line 80: Faster than before

alexander-akait:
What do you think about `codegen`? How to better fix it? 

Codegen output:
```
content: 'ab';
```

Source:
```css
    content: "a\
b";
```

It happens due `\\n` logic, CSS ignore newline in this case, so newlines just stripped from output.

css/parser/src/lexer/value.rs
kdy1: Did you run the benchmark?
Seems like the new code is appending one char at a time, so I'm not sure how can it be faster than before.

kdy1 (commented):
For codegen, I prefer preserving input if it's possible without big effort.

css/parser/src/lexer/value.rs
alexander-akait: @kdy1 we don't look up above, but anyway for valid parsing escapes we should use this logic, maybe we can optimize it better in future

kdy1: I see, then it's fine.

alexander-akait:
@kdy1 if we want to keep string as is, we need introduce `String { value, rawValue }` and keep original string in `rawValue` and using this for codegen, do you agree with this?

kdy1:
Maintaining two `String` for values causes lots of problems in my experience.

So I think the answer depends on if `a\nb` is identical to `ab` in CSS. I'll search more about it.
If it's identical, I think it should be simply `ab` and if it's different, storing as `ab` is breaking input.

alexander-akait:
Original:
```
a {
  // These are same
  content: "a\
b";
  content: "ab";  
}
```

Spec - https://www.w3.org/TR/css-syntax-3/#consume-string-token

> U+005C REVERSE SOLIDUS (\)
> If the next input code point is EOF, do nothing.

Our case, just consume

> Otherwise, if the next input code point is a newline, consume it.

> Otherwise, (the stream starts with a valid escape) consume an escaped code point and append the returned code point to the <string-token>’s value.

kdy1:
Ah, I see. Then I think creating a type with two fields is better.
(I want to allow using the parser for various tools)

alexander-akait:
@kdy1 thanks for feedback, I will do it in near future

alexander-akait:
I am working on fixes, there are a lot of bugs with lexer/parser, hope I will fix it today

alexander-akait:
@kdy1 Found bugs/features: 
- `Token::Ident` should have raw value too, for example - `[attr=\2c] {}`, `\2c`
- `Token::Url` should be raw value too, for example - `url(./\2c.css)`
- Wrong ast for `@import (./style.css)`

kdy1:
> Token::Ident should have raw value too, for example - [attr=\2c] {}, \2c

I see. I think it's not avoidable, so I think ast change is the right way to go.

> Token::Url should be raw value too, for example - url(./\2c.css)

Same here.

> Wrong ast for @import (./style.css)

Seems like we need to change it as enum? I didn't know that it's allowed to use import like it.

alexander-akait:
> I see. I think it's not avoidable, so I think ast change is the right way to go.
> Same here.

Yes, I will do it in other PRs, when rebase it

> Seems like we need to change it as enum? I didn't know that it's allowed to use import like it.

Yes, possible values is `Str` - `@import './style.css'`, `Url` - `@import url(./style.css)`, `Fn` -> `@import url("./style.css")`

Maybe you can give advice how implement this, I am very new in rust and here a lot of magic for me, especial for attributes and macros

alexander-akait:
@kdy1 ready for merge

PR #2298, opened 2021-09-23
Fix attribute selector type

alexander-akait:
fix `type` for attribute selector, i.e. `AttrributeSelector` -> `AttributeSelector` + more tests


kdy1 (approved):
Thanks!

PR #2302, opened 2021-09-25
fix(es/parser): multiple import/export specifiers without comma

g-plane:
Before this PR, the code below should be invalid:

```js
import { a b c } from 'mod'

export { a b c }
```

but swc didn't report. This PR fixes it.

### DIFF
--- a/ecmascript/parser/src/parser/stmt.rs
+++ b/ecmascript/parser/src/parser/stmt.rs
@@ -1838,6 +1838,24 @@ export default function waitUntil(callback, options = {}) {
     }
 
     #[test]
+    #[should_panic(expected = "Expected ,, got b")]
+    fn import_specifiers_without_comma() {
+        test_parser(
+            "import { a b c } from 'mod'",
+            Syntax::Es(EsConfig::default()),
+            |p| p.parse_module(),
+        );
+    }
+
+    #[test]
+    #[should_panic(expected = "Expected ,, got b")]
+    fn export_specifiers_without_comma() {
+        test_parser("export { a b c }", Syntax::Es(EsConfig::default()), |p| {
+            p.parse_module()
+        });
+    }
+
+    #[test]
     #[should_panic(expected = "top level await is only allowed in module")]
     fn top_level_await_in_script() {
         let src = "await promise";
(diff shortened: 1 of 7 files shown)


ecmascript/parser/src/parser/stmt.rs
```diff
@@ -1837,6 +1837,24 @@ export default function waitUntil(callback, options = {}) {
         );
     }
 
+    #[test]
+    #[should_panic(expected = "Expected ,, got b")]
```
kdy1, line 1841: Can you add tests to https://github.com/swc-project/swc/tree/83d88ce388ee8553af6507032630ddbc7a100d14/ecmascript/parser/tests/typescript-errors instead?

kdy1 (changes requested):
Code changes LGTM, but I think fixture test would be better for error reporting test.
You can do `UPDATE=1 cargo test`.

g-plane:
How the error messages should be?

kdy1:
The current error message is fine. Just add a file named `input.ts` or `input.tsx` and run `UPDATE=1 cargo test`.
You may need to `touch` `tests/typescript.rs`

ecmascript/parser/src/parser/stmt.rs
g-plane: Should these tests be removed?

kdy1: I think it's better, as it's a duplicate.

kdy1 (commented):
The only issue I found it duplciated test :)
Thanks!

kdy1 (approved):
LGTM. Thanks!

ecmascript/parser/tests/typescript-errors/multiple-specifiers/export/input.ts.stderr
```diff
@@ -0,0 +1,6 @@
+error: Expected ,, got b
```
alexander-akait, line 1: @kdy1 Looks misleading, maybe better - `Expected ',', got b`, i.e. add quotes around expected characters

g-plane: See: https://github.com/swc-project/swc/pull/2304

PR #2309, opened 2021-09-27
feat(es/parser): support type-only import/export specifiers

g-plane:
Ref: https://github.com/microsoft/TypeScript/pull/45998


ecmascript/parser/src/parser/stmt/module_item.rs
```diff
@@ -177,18 +177,90 @@ impl<'a, I: Tokens> Parser<I> {
...
+                // Handle:
+                // `import { type xx } from 'mod'`
+                // `import { type xx as yy } from 'mod'`
+                // `import { type as } from 'mod'`
+                // `import { type as as } from 'mod'`
+                // `import { type as as as } from 'mod'`
+                if orig_name.sym == js_word!("type")
+                    && is!(self, IdentName)
+                    && self.syntax().typescript()
+                    // invalid: `import type { type something } from 'mod'`
```
dsherret, line 196: I think we should still parse this scenario even though it is invalid. It would be useful in dprint-plugin-typescript for example to parse this then convert `import type { type something} from 'mod'` to `import type { something } from "mod";`. Thoughts? (by the way, nice PR)

g-plane: I think dprint is a formatter, and it shoudn't accept code with invalid syntax.

g-plane: I prefer that this would be syntax errors, but I'm OK to treat it as valid syntax, because this isn't a huge change.

> The biggest annoying problem is that once I change the code, rustc will cost very very long time.

ecmascript/ast/src/module_decl.rs
```diff
@@ -199,6 +199,8 @@ pub struct ImportNamedSpecifier {
 
     #[serde(default)]
     pub imported: Option<Ident>,
+
+    pub is_type_only: bool,
```
kdy1, line 203: Please add `#[serde(default)]`

ecmascript/ast/src/module_decl.rs
```diff
@@ -244,4 +246,6 @@ pub struct ExportNamedSpecifier {
     /// `Some(bar)` in `export { foo as bar }`
     #[serde(default)]
     pub exported: Option<Ident>,
+    /// `type` in `export { type foo as bar }`
+    pub is_type_only: bool,
```
kdy1, line 250: Same here

ecmascript/codegen/src/lib.rs
```diff
@@ -184,7 +184,9 @@ impl<'a> Emitter<'a> {
         for specifier in &node.specifiers {
             match specifier {
                 ImportSpecifier::Named(ref s) => {
-                    specifiers.push(s);
+                    if !s.is_type_only {
```
kdy1, line 187: This seems wrong.
Type-only specifiers should be stripped out by `typescript::strip()`

ecmascript/codegen/src/lib.rs
```diff
@@ -317,6 +319,11 @@ impl<'a> Emitter<'a> {
                     }
                     result
                 }
+                ExportSpecifier::Named(ExportNamedSpecifier { is_type_only, .. })
+                    if *is_type_only =>
```
kdy1, line 323: Same here

ecmascript/parser/src/parser/stmt/module_item.rs
kdy1: It can be syntax error and can be valid ast in same time, and I think it's the right direction to go.
`emit_err` can be used for such purpose.

ecmascript/parser/src/parser/stmt/module_item.rs
```diff
@@ -177,18 +177,90 @@ impl<'a, I: Tokens> Parser<I> {
...
+                    && !is_type_only_import
+                {
+                    let possibly_orgi_name = self.parse_ident_name()?;
```
kdy1, line 199: I found a typo

ecmascript/parser/src/parser/stmt/module_item.rs
```diff
@@ -177,18 +177,90 @@ impl<'a, I: Tokens> Parser<I> {
...
+
+                // Handle:
+                // `import { type xx } from 'mod'`
+                // `import { type xx as yy } from 'mod'`
+                // `import { type as } from 'mod'`
+                // `import { type as as } from 'mod'`
+                // `import { type as as as } from 'mod'`
+                if orig_name.sym == js_word!("type")
+                    && is!(self, IdentName)
+                    && self.syntax().typescript()
```
kdy1, line 195: Also, please move `self.syntax().typescript()` to first condition.

ecmascript/parser/src/parser/stmt/module_item.rs
```diff
@@ -558,10 +631,73 @@ impl<'a, I: Tokens> Parser<I> {
...
+        let mut is_type_only = false;
+        let mut orig = self.parse_ident_name()?;
+
+        // Handle:
+        // `export { type xx }`
+        // `export { type xx as yy }`
+        // `export { type as }`
+        // `export { type as as }`
+        // `export { type as as as }`
+        if orig.sym == js_word!("type")
```
kdy1, line 646: Please move `self.syntax().typescript()` to first condition

kdy1 (changes requested):
Thanks!
Parser changes LGTM, but I found some small issues about the role of crates.

ecmascript/parser/src/parser/stmt/module_item.rs
g-plane: The reason why I don't put it first is that most users won't use "type" as specifier then this short circuit can bring better performance.

kdy1: Is u64 comparison faster than bool check?

Also, `self.syntax().typescript() ` is used to remove code for typescript parser.

ecmascript/transforms/module/src/amd.rs
```diff
@@ -350,6 +352,7 @@ where
                                          previous pass"
                                     ),
                                 })
+                                .filter(|e| !e.is_type_only)
```
g-plane, line 355: @kdy1 Should it be reverted?

kdy1: I think it's fine.

g-plane:
I've used `emit_err` in `ecmascript/parser/src/parser/stmt/module_item.rs`, but I can't hide the review conversation above.

kdy1 (approved):
LGTM. Thanks a lot!

Issue #1366, opened 2021-01-28
parseFileSync span bug


JiangWeixian:
**Describe the bug**

same file parse twice, different output

**Input code**

```
//test.jsx
const a = 1
```

import { parseFileSync } from '@swc/core';

const ast = parseFileSync('./test.jsx', { jsx: true, syntax: 'ecmascript', comments: true, dynamicImport: true, decorators: true, decoratorsBeforeExport: true, script: true })
const ast1 = parseFileSync('./test.jsx', { jsx: true, syntax: 'ecmascript', comments: true, dynamicImport: true, decorators: true, decoratorsBeforeExport: true, script: true })

console.log(JSON.stringify(ast))
console.log('====')
console.log(JSON.stringify(ast1))

```
{"type":"Module","span":{"start":0,"end":11,"ctxt":0},"body":[{"type":"VariableDeclaration","span":{"start":0,"end":11,"ctxt":0},"kind":"const","declare":false,"declarations":[{"type":"VariableDeclarator","span":{"start":6,"end":11,"ctxt":0},"id":{"type":"Identifier","span":{"start":6,"end":7,"ctxt":0},"value":"a","typeAnnotation":null,"optional":false},"init":{"type":"NumericLiteral","span":{"start":10,"end":11,"ctxt":0},"value":1},"definite":false}]}],"interpreter":null}
====
{"type":"Module","span":{"start":12,"end":23,"ctxt":0},"body":[{"type":"VariableDeclaration","span":{"start":12,"end":23,"ctxt":0},"kind":"const","declare":false,"declarations":[{"type":"VariableDeclarator","span":{"start":18,"end":23,"ctxt":0},"id":{"type":"Identifier","span":{"start":18,"end":19,"ctxt":0},"value":"a","typeAnnotation":null,"optional":false},"init":{"type":"NumericLiteral","span":{"start":22,"end":23,"ctxt":0},"value":1},"definite":false}]}],"interpreter":null}
```

**first ast span start at 0, second start at 12**

**Config**
```json
{
    // Please copy and paste your .swcrc file here
}
```

**Expected behavior**

**Version**
The version of @swc/core: 

"@swc/core": "^1.2.46",

**Additional context**

Brooooooklyn:
By design，just change the file name passed into the `parseFileSync` you'll get the same result.

JiangWeixian:
> By design，just change the file name passed into the `parseFileSync` you'll get the same result.

I think it's a bug

[...]

PR #2320, opened 2021-09-30
Support Wallaby AST ranges

vjpr:
Wallaby requires information about the original source AST - as do several other tools.

E.g. For the following statement `var a = b ? c() : d();`

It needs this 2D array of ranges:

`[start-line, start-column, end-line, end-column]`

```
 * ranges = [
 *   [ 1, 0, 1, 22 ],   // whole statement
 *   [ 1, 12, 1, 15 ],  // c() execution path
 *   [ 1, 18, 1, 21 ]   // d() execution path
 * ];
```

See: https://wallabyjs.com/docs/config/compilers.html#writing-a-custom-compiler

I have modified the `Output` type to also return `ranges`.

To generate it, I just traverse the AST using a visitor.

For now, I am just returning the byte offsets, and then in my JS code I map the offsets to line/col. I think this is what you do in `SourceMap#span_to_line`.

Here is the JS code where I am using SWC as a Wallaby "compiler".

```ts
function compile(filename, code, opts) {
  const defaultSourceMap = true
  //const defaultSourceMap = 'inline' // originally
  const finalOpts = {
    ...opts,
    sourceMaps:
      opts.sourceMaps === undefined ? defaultSourceMap : opts.sourceMaps,
  }

  const output: swc.Output = swc.transformSync(code, finalOpts)

  //console.log('sourcemap', output.map)
  //const ranges = await sourceMapToWallabyRanges(output.map)
  let ranges = JSON.parse(output.ranges)

  // Spans to lines.
  const lines = code.split('\n')
  ranges = ranges.map(range => {
    const [x, startByte, y, endByte] = range
    const [startLine, startCol] = findLineColForByte(lines, startByte)
    const [endLine, endCol] = findLineColForByte(lines, endByte)
    const out = [startLine, startCol, endLine, endCol]
    return out
  })

  console.log('ranges', ranges)

  return {
    map: output.map,
    code: output.code,
    ranges,
  }
}

////////////////////////////////////////////////////////////////////////////////

function findLineColForByte(lines, index) {
  let totalLength = 0
  let lineStartPos = 0
  for (let lineNo = 0; lineNo < lines.length; lineNo++) {
    totalLength += lines[lineNo].length + 1 // Because we removed the '\n' during split.
    if (index < totalLength) {
      const colNo = index - lineStartPos
      return [lineNo + 1, colNo]
    }
    lineStartPos = totalLength
  }
  console.warn('WARN:', {totalLength, index})
  return []
}
```

I am new to Rust, so probably haven't done things in the best way.

Looking for feedback on the best way to get this access to the AST.

## Related

- https://github.com/swc-project/swc/issues/1366
- https://github.com/wallabyjs/public/issues/2818#issuecomment-929757665

### DIFF
--- a/benches/typescript.rs
+++ b/benches/typescript.rs
@@ -133,6 +133,7 @@ fn bench_codegen(b: &mut Bencher, _target: JscTarget) {
                 None,
                 false,
                 None,
+                None,
             )
             .unwrap(),
         );
--- a/common/src/source_map.rs
+++ b/common/src/source_map.rs
@@ -193,6 +193,7 @@ impl SourceMap {
             .cloned()
     }
 
+    #[allow(unused)]
     fn next_start_pos(&self, len: usize) -> usize {
         // Add one so there is some space between files. This lets us distinguish
         // positions in the source_map, even in the presence of zero-length files.
@@ -222,6 +223,7 @@ impl SourceMap {
         let mut files = self.files.borrow_mut();
 
         let start_pos = self.next_start_pos(src.len());
+        //let start_pos = usize::MIN;
 
         let source_file = Lrc::new(SourceFile::new(
             filename,
--- a/node-swc/src/types.ts
+++ b/node-swc/src/types.ts
@@ -808,6 +808,10 @@ export interface Output {
    * Sourcemap (**not** base64 encoded)
    */
   map?: string;
+  /**
+   * Ranges for wallaby.js
+   */
+  ranges?: number[][];
 }
 
 export interface MatchPattern { }
--- a/node/binding/src/bundle.rs
+++ b/node/binding/src/bundle.rs
@@ -127,6 +127,7 @@ impl Task for BundleTask {
                             None,
                             minify,
                             None,
+                            None,
                         )?;
 
                         Ok((k, output))
--- a/node/binding/src/print.rs
+++ b/node/binding/src/print.rs
@@ -39,6 +39,7 @@ impl Task for PrintTask {
                 None,
                 self.options.config.clone().minify,
                 None,
+                None,
             )
             .convert_err()
     }
@@ -97,6 +98,7 @@ pub fn print_sync(cx: CallContext) -> napi::Result<JsObject> {
             &[],
             None,
             options.config.minify,
+            None,
             None,
         )
     }
--- a/node/bundler/tests/fixture.rs
+++ b/node/bundler/tests/fixture.rs
@@ -101,6 +101,7 @@ fn pass(input_dir: PathBuf) {
                         None,
                         false,
                         Some(true.into()),
+                        None,
                     )
                     .expect("failed to print?")
                     .code;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,6 +48,8 @@ use swc_ecma_transforms::{
     resolver_with_mark,
 };
 use swc_ecma_visit::{noop_visit_type, FoldWith, Visit, VisitWith};
+use crate::ecmascript::ast::Expr;
+use common::source_map::Pos;
 
 mod builder;
 pub mod config;
@@ -151,6 +153,8 @@ pub struct TransformOutput {
     pub code: String,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub map: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub ranges: Option<String>,
 }
 
 /// These are **low-level** apis.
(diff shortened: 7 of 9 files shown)


kdy1:
I think this is not an option for `@swc/core`.

vjpr:
How would you suggest getting this AST information?

kdy1:
I think a new node binding built on top of swc will be the best.

vjpr:
Thanks, I found https://github.com/vercel/next.js/tree/canary/packages/next/build/swc. This looks like a good template to follow.

vjpr:
@kdy1 

I am trying to wrap `swc` (using `next-swc` as a template) to extract AST span info (use case: https://github.com/swc-project/swc/pull/2320).

I have an `impl Visit` implementation that looks at every node and returns the span info. I'm looking for the correct location to run it without having to modify the `swc` crate.

I want it to be efficient so I only want `parse_js` to happen once.

Should I add it as a `before_pass` to `TransformTask#compute`?

```
    Input::Source(ref s) => {
        let before_pass = chain!(
        //     hook_optimizer(),
        //     Optional::new(next_ssg(), !self.options.disable_next_ssg),
        //     amp_attributes(),
        //     next_dynamic(s.name.clone(), self.options.pages_dir.clone()),
        //     styled_jsx(),
    		   range_collector()
        );
        self.c.process_js_with_custom_pass(
            s.clone(),
            &handler,
            &self.options.swc,
            // before_pass,
            before_pass(),
            noop(),
        )
    }
```

But `pass` only accepts a `Fold`. So I'm thinking `as_folder` is the way to go...

But I don't need a `VisitMut`, only a `Visit`...does this matter much?

```
pub fn range_collector() -> impl Fold + VisitMut {
    as_folder(RangeCollector)
}
```

vjpr:
So `process_js` takes a `Program` which I need to run my visitor on. So I just extracted some code from `process_js_with_custom_pass`, and then called `process_js`.

```rs
pub fn exec_transform<F>(cx: CallContext, op: F) -> napi::Result<JsObject>
where
    F: FnOnce(&Compiler, String, &TransformOptions) -> Result<Arc<SourceFile>, Error>,
{
    let c = get_compiler(&cx);

    let s = cx.get::<JsString>(0)?.into_utf8()?;
    let is_module = cx.get::<JsBoolean>(1)?;
    let options: TransformOptions = cx.get_deserialized(2)?;

    let output = try_with_handler(c.cm.clone(), |handler| {
        c.run(|| {
            if is_module.get_value()? {
                let program: Program =
                    serde_json::from_str(s.as_str()?).context("failed to deserialize Program")?;
                let res = c.process_js(&handler, program, &options.swc).unwrap();
                let ranges: Ranges = vec![vec![0, 0, 0, 0]];
                Ok(TransformOutputWithRanges { code: res.code, map: res.map, ranges: ranges })
            } else {
                let fm =
                    op(&c, s.as_str()?.to_string(), &options).context("failed to load file")?;

                // From `process_js_file`...

                let opts = &options.swc;

                let config = c.config_for_file(handler, opts, &fm.name)?;
                let config = match config {
                    Some(v) => v,
                    None => {
                        bail!("cannot process file because it's ignored by .swcrc")
                    }
                };

                let custom_before_pass = noop();
                let custom_after_pass = noop();

                let config = BuiltConfig {
                    pass: chain!(custom_before_pass, config.pass, custom_after_pass),
                    syntax: config.syntax,
                    target: config.target,
                    minify: config.minify,
                    external_helpers: config.external_helpers,
                    source_maps: config.source_maps,
                    input_source_map: config.input_source_map,
                    is_module: config.is_module,
                    output_path: config.output_path,
                    source_file_name: config.source_file_name,
                    preserve_comments: config.preserve_comments,
                    inline_sources_content: config.inline_sources_content,
                };

                let program = c.parse_js(
                    fm.clone(),
                    handler,
                    config.target,
                    config.syntax,
                    config.is_module,
                    true,
                )?;

                let ranges: Ranges = get_ranges(&program);

                let res = c.process_js(&handler, program, &options.swc).unwrap();

                // let res = c.process_js_file(fm, &handler, &options.swc).unwrap();
                Ok(TransformOutputWithRanges { code: res.code, map: res.map, ranges: ranges })
            }
        })
    })
    .convert_err()?;


    complete_output_with_ranges(cx.env, output)
}
```

kdy1:
I'm not sure, but I think you should fork swc crate.

vjpr:
> I'm not sure, but I think you should fork swc crate.

It's ok, I managed to get it done as a wrapper.

PR #2338, opened 2021-10-03
Fix tests failing due to CRLF on windows

vbfox:
A few tests were failing when the repository is checked out with native windows line endings (so the expected content actually has CRLF in it)

I could have changed my git setup but might as well fix it for everyone a the change is simple and seems harmless.


kdy1 (approved):
LGTM. Thanks!

PR #2358, opened 2021-10-06
test(css/parse): an-plus-b syntax

alexander-akait:
Many tests, some of them failed, we need fix them too


kdy1:
I think tests will fail, and requires some update of reference files.
Can you rebase this?

alexander-akait:
@kdy1 Ideally we need `An+B` syntax there to output correct AST, anyway tests should work (using tokens), I will rebase

alexander-akait:
Let's close, we already implemented it https://github.com/swc-project/swc/blob/main/crates/swc_css_parser/tests/fixture/selector/pseudo-class/an-plus-b/input.css

PR #2361, opened 2021-10-06
feat(css/ast): Add raw to `Text`

alexander-akait:
`raw` for nodes there it needed. We also need this for other node:
- `Str` https://github.com/swc-project/swc/pull/2295 (I will rebase after merge)
- `Url` (added todo, will be resolved in future)

Also I am working on more tests


alexander-akait:
In future we will remove `write_ident` from codegen, because we already have raw and so we don't need extra logic here

alexander-akait:
Most of changes is just `raw` value in snapshots

kdy1 (approved):
LGTM. Thanks!

PR #2363, opened 2021-10-06
fix(css/parse): at-rule import ast

alexander-akait:
import can be:
```css
/* Str */
@import './style.css';
/* Url (legacy) */
@import url(./style.css);
/* FnValue */
@import url('./style.css');
```

I think in future we will implement `parse` for `ast::Url` and move common logic there, because now we have logic duplicate for `url()` (legacy), but it require more work, will be done in future

Related: https://github.com/swc-project/swc/pull/2295

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2481,7 +2481,7 @@ dependencies = [
 
 [[package]]
 name = "swc_css"
-version = "0.10.0"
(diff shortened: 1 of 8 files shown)


alexander-akait:
Yes, in my roadmap on today

PR #2382, opened 2021-10-08
Test number

alexander-akait:
@kdy1 We have a lot of bug in lexer for `number`/`dimension`/`percent` tokens, also bugs in parser, here interesting case, I think we should not try to interested `+`/`-`/`*`/`/` for all values, because it is syntax only for `calc()`/`min()`/`max()`/`clamp()`.

I think we should stay out of the special syntax (for special syntax in values) for now, also we still do it invalid (and have invalid ast), in theory I can keep old behavior for math operations and covert them to `delim` token, I think we will improve AST for `calc`/etc in future because they are math operators and we should parser them as mathematical (left/right and operation regarding spec).

What do you think?

### DIFF
--- a/css/ast/src/token.rs
+++ b/css/ast/src/token.rs
@@ -36,32 +36,6 @@ pub enum Token {
     /// `#`
     Hash {
         is_id: bool,
-    /// `(`
-    LParen,
-
-    /// `)`
-    RParen,
-
-    /// `[`
-    LBracket,
-
-    /// `]`
-    RBracket,
-
-    Percent {
-        value: f64,
-    },
-
-    Dimension {
-        value: f64,
-        unit: JsWord
-    },
-
-    Number {
-        value: f64,
-    },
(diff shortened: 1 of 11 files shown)


kdy1:
> it is syntax only for calc()/min()/max()/clamp().

I didn't know it, thanks!

> I think we will improve AST for calc/etc in future because they are math operators and we should parser them as mathematical (left/right and operation regarding spec).

Accepting invalid tokens and reporting errors for all kinds of functions seems better to me. If parser rejects parsing, any tool depending on the parser can't work, so we should try to represent them in ast and report error instead.

kdy1:
Can you rebase this or close this?

alexander-akait:
@kdy1 Yep, only tests left here

alexander-akait:
Oh, we already have these tests and they work fine, let's close, we don't need it anymore

PR #2389, opened 2021-10-09
feat(css/parse): Add `raw` to `Url` node

alexander-akait:
Should be merged after https://github.com/swc-project/swc/pull/2295, we need improve whitespaces check from https://github.com/swc-project/swc/pull/2295

Also we still have problems with recovery mode:
1. We should return `BadUrlToken` and `UrlToken` in some places (added `TODO`)
2. Also we need implement https://www.w3.org/TR/css-syntax-3/#consume-the-remnants-of-a-bad-url (added `TODO`)

`1` and `2` I will do in another PR, after we merged https://github.com/swc-project/swc/pull/2295 and this

Also we have small problems with eating whitespaces (they should be part of `raw`) (added `TODO`) and I will solve it in another PR too, because we need refactor/rewrite (`FnToken` eating) other parts before doing it possible

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2481,7 +2481,7 @@ dependencies = [
 
 [[package]]
 name = "swc_css"
-version = "0.12.0"
+version = "0.13.0"
 dependencies = [
  "swc_css_ast",
  "swc_css_codegen",
@@ -2492,7 +2492,7 @@ dependencies = [
 
 [[package]]
 name = "swc_css_ast"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
  "is-macro",
  "serde",
@@ -2503,7 +2503,7 @@ dependencies = [
 
 [[package]]
 name = "swc_css_codegen"
-version = "0.10.0"
+version = "0.11.0"
 dependencies = [
  "auto_impl",
  "bitflags",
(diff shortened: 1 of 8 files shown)


kdy1:
Thanks!
What you said is correct, and for

> Also we have small problems with eating whitespaces (they should be part of raw) 

can we just reset to position and eat token?
It will not be performant, but we don't have to be performant in case of error.

alexander-akait:
> can we just reset to position and eat token?

I have ideas:
1. Allow to return `whitespaces` from `skip_whitespace`
2. Keep position before and after eating whitespaces and slice them from input

Solution depends on what we want CST or AST, for example `postcss` is keeping whitespaces in the `raws` property (https://astexplorer.net/), it allows linters and other tools write formatting rules, in other hard I have strong feeling what formatting rules is the worst thing, because ideal formatter should be pure and do not really on exiting formatting.

Anyway we can decide it in future, now I think better to use `2` point, because keeping whitespaces and rewrite AST to CST is just feature (can be even under flag to prevent big memory usage) and can be added late.

alexander-akait:
@kdy1 ready to merge, fixed problems with `whitespaces` in `raw`, also found bug in `read_ident_like` (added TODO), we need `FnToken` here, but it requires rewrite more things too, will be solved in future

I will just add `TODO` on problems to avoid big changes in a one PR and to make it easier to review

PR #2398, opened 2021-10-10
feat(css/ast): Add delim token

alexander-akait:
It is big refactor our lexer, here we introduce `delim` token according spec https://www.w3.org/TR/css-syntax-3/#typedef-delim-token (note - not all cases are covered due we need to refactor parser too, better do it step by step). Also we need this for https://github.com/swc-project/swc/pull/2382.

Why we need `delim` token? Because any characters in CSS is valid, now we have `Dollar`/`Caret`/`Tilde` and etc tokens, but it is wrong way, these characters are `Delim` tokens, it allows CSS be resistance to wrong input, for example:

```css
div {
  color: !?<>;
  color: red;
}
```

is should be parsable (you can try it in browser and see `red` color for text), now we have each separate token for special characters, but this is wrong way, because we cannot create a new token for an each character (allowed in CSS).

Also:
1. Change order in `read_token` according https://www.w3.org/TR/css-syntax-3/#consume-token
2. Remove wrong `Err`, because now it is `delim` tokens and we are more error recoverable
3. Parse non printable ascii
4. More helpers we will need them in future in other places

Ideally should be merged after https://github.com/swc-project/swc/pull/2389 

After this we will return to `Num` token and fix problems with number parsing/ast


kdy1 (commented):
Mostly LGTM, but I found some small issues.
These are very small issues, though.

alexander-akait:
@kdy1 ready for review, after this we can fix number parsing, then we will convert other weird tokens to delim tokens (with many fixes), also I think we will finish our lexer by the end of the month and I will focus on parser

kdy1 (approved):
LGTM.
Thanks for your all hard works!

alexander-akait:
@kdy1 Maybe I should do major release every time, becaue in most of my commits we change ast and tokens :smile:

PR #2409, opened 2021-10-12
fix(es/parser): Fix class getter/setter ASI problem

g-plane:
This affects:

```js
class C {
  get
  x() { return 0 }

  set
  x(value) {}
}
```

Previously, the example will be parsed as two class properties and two class methods. This PR let it be parsed as one getter and one setter.

Comparison:

- [Babel](https://astexplorer.net/#/gist/c0ea098c1ad6214773feb870c7be61cf/88cb83fedb6fd825e6350ab9605bd510010c9e1d)
- [TypeScript](https://www.typescriptlang.org/play?#code/MYGwhgzhAEDC0G8BQ1oHMCmAXF0AeAFAJSLQBO2ArmQHbQAM0AvkrhNroQG5giUYkELJkA)

### DIFF
--- a/ecmascript/parser/src/parser/class_and_fn.rs
+++ b/ecmascript/parser/src/parser/class_and_fn.rs
@@ -319,7 +319,7 @@ impl<'a, I: Tokens> Parser<I> {
 
         let declare_token = if declare {
             // Handle declare(){}
-            if self.is_class_method()? {
+            if self.is_class_method() {
                 let key = Either::Right(PropName::Ident(Ident::new(
                     js_word!("declare"),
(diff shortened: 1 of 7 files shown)


g-plane:
Will class property be inserted semicolon automatically when tokenizing?

kdy1:
No, macro for semicolons checks for it.
https://github.com/swc-project/swc/blob/d114e7d3643a0adc0f75e2664667f5990ea50317/ecmascript/parser/src/parser/macros.rs

g-plane:
I've rebased and it can be reviewed now.

PR #2407, opened 2021-10-11
fix(es/parser): Ensure comments in arrow func exprs with a type parameter are collected

dsherret:
Fixes #2339

PR #2410, opened 2021-10-12
chore(es/parser): cleanup comments buffer to not store length on a node

dsherret:
As per https://github.com/swc-project/swc/pull/2407#discussion_r726706547

There doesn't seem to be any benefit to storing this and it just uses more memory.


dsherret:
@kdy1 oops thanks! I fixed it. I'm so used to formatting being automatic on save and with my swc setup right now it's not... I should configure that.

PR #2411, opened 2021-10-12
perf(atoms): Update `string_cache`

kdy1:
swc_atoms:
 - Use `string_cache` with `parking_lot` patch applied.

wasm:
 - Pin `parking_lot_core` to `=0.8.0` to workaround https://github.com/Amanieu/parking_lot/issues/269.


dsherret:
I also just ran into this parking_lot issue 😅

PR #2414, opened 2021-10-12
fix(css/parser): Fix parsing of comments

alexander-akait:
Fix comment parsing, in css doesn't allow line comments, i.e.
```css
// Foo
```

is invalid comments.

Will be great to merge after https://github.com/swc-project/swc/pull/2398

### DIFF
--- a/.github/workflows/cargo.yml
+++ b/.github/workflows/cargo.yml
@@ -87,8 +87,6 @@ jobs:
         run: cargo check --color always --all --all-targets
   test:
     name: Test
-    needs:
-      - check
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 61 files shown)


kdy1 (changes requested):
I think it's right to bump minor version of `swc_css_parser` because public surface didn't change.

alexander-akait:
@kdy1 hm, before we generate ast for `// test`, now we don't do it, so if somebody rely on `// comment` parsing, it will be broken now, so I decided that it would be major change... Anyway I can change

alexander-akait:
@kdy1 Ready fro review, if we merge this and https://github.com/swc-project/swc/pull/2415, I will send a couple more fixes :star:

kdy1 (changes requested):
Code changes are fine, but I meant a minor version of the parser.

In detail, crates using `swc_stylis@ 0.11.x` couldn't select the correct version for `swc_css_parser`.

alexander-akait:
@kdy1 Done :star:

PR #2415, opened 2021-10-12
fix(css/ast): Fix delimiter token

alexander-akait:
Part of https://github.com/swc-project/swc/pull/2382 (we need it to solve #2382), delim tokens for `-`/`+`/`.` and fix logic for 

> If the input stream starts with a number, reconsume the current input code point, consume a numeric token, and return it.

https://www.w3.org/TR/css-syntax-3/#consume-token


kdy1 (approved):
Thanks!
you did right, because removing or adding enum variants can break compilation, if `#[non_exhuastive]` is not used.

PR #2416, opened 2021-10-12
feat(mdx): Add mdx parser

kdy1:


kdy1:
I'm doing this using some rest time.


> in theory we can reuse markdown parser here

We can use markdown library, but I don't think a markdown library will expose the whole internals to the level we need, because it's really not a comfortable thing to do.


AST is so simple, so is the parser.

schickling:
Is closing this PR good or bad news? 👀

alexander-akait:
@schickling mardown (and so mdx) will be implemented late, it is on my roadmap too, we just need to start with markdown parser (and late add support for mdx)

PR #2420, opened 2021-10-13
docs: improve react transform docs related to the resolver

dsherret:
I believe this should say to use `ts_resolver` when using TypeScript.


kdy1 (commented):
I think typescript nodes will be all removed by `typescript::strip`, which is expected to be applied before this pass.

What in order are you using?

PR #2425, opened 2021-10-13
refactor(css/ast): use `delim` token

alexander-akait:
Refactor all special tokens to `delim` tokens according spec + order of `token` from spec for better reading with spec.

Better to merge it after https://github.com/swc-project/swc/pull/2415.

Also after this and https://github.com/swc-project/swc/pull/2415, we will fix `Number` https://github.com/swc-project/swc/pull/2382.

Also I added `TODO` for missing tokens


kdy1 (approved):
LGTM. Thanks!

PR #2426, opened 2021-10-13
refactor(css/parser): Added `BadUrl` token + error recovery

alexander-akait:
Added `BadUrl` token + error recovery


alexander-akait:
Do you mean we should do breaking change here?

kdy1:
I think we should mark this as a breaking change of `swc_css_ast` because the changed interface is in it.

PR #2444, opened 2021-10-15
Fix parsing numbers

alexander-akait:
@kdy1 Fix parsing numbers, part of https://github.com/swc-project/swc/pull/2382, we need refactor value parser before introduce `Percent` and `Dimension` tokens, so let's avoid big changes and do it step by step.

Also some tests in `esbuild` test cases was wrong, it is invalid numbers.

Also added `raw` to `Num` node, it can be useful for minification and lints

### DIFF
--- a/.github/workflows/compilation.yml
+++ b/.github/workflows/compilation.yml
@@ -52,6 +52,10 @@ jobs:
         run: |
           (cd ecmascript/transforms && cargo hack check --feature-powerset --no-dev-deps)
 
+      - name: Check swc_ecma_transforms_compat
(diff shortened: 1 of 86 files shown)


alexander-akait:
For example: `& a { value: 10px }` is valid for CSS-in-JS and will be valid in CSS in near future, but it has different meaning and we will have to decide between hacks in `CSS-in-JS` or valid parsing using specs

alexander-akait:
`Cargo / Check license (pull_request)` is failed, but I think not related to my changes...

PR #2445, opened 2021-10-15
fix(css/ast): unit

alexander-akait:
Refactor `unit`, it will be `Dimension` node in future (from `<dimension-token>`). Also using `raw` give us more ability and faster codegen, because we don't need to convert them, also `kind: null` fixed


kdy1 (commented):
Code changes look fine to me.
I used `Unit` to optimize performance, but actually css parsing is not time-consuming task in most cases.

PR #2482, opened 2021-10-19
Feat added percent token

alexander-akait:
@kdy1 

Added `Percent` and `Dimension` tokens, but here some problems:
1. Our selector parser is fully wrong due invalid logic for `complex-selector`, so I move broken tests, to fix it we need big refactor, but now it is impossible, to solve this we need introduce `Fn` token, but without `Percent` and `Dimension` it is very hard to do in one PR and lead to big snapshots changes
2. We need simplify AST for `Percent` and `Dimension`, no need to store number, because `Percent` and `Dimension` are different nodes than `Number` (yes, they looks the same, but have different semantics)
3. Parsing `any` value is wrong (will be fixed in future)
4. In CSS context no `SpaceValues`, this node should be removed and we should eat whitespaces (will be fixed in future)

Anyway as I said before I think we should focus on our lexer and valid parsing and then focus on parsing, unfortunately to solve all the listed problems above hard in one PR or will lead to big changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2503,7 +2503,7 @@ dependencies = [
 
 [[package]]
 name = "swc_css"
-version = "0.22.0"
+version = "0.23.0"
 dependencies = [
  "swc_css_ast",
  "swc_css_codegen",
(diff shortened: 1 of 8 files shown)


alexander-akait:
@kdy1 Also ready for review

PR #2496, opened 2021-10-20
fix(es/transforms): Visit import equals rhs properly

nayeemrmn:
Addresses https://github.com/denoland/deno/issues/12395#issuecomment-940599288.


nayeemrmn:
Right now in this PR `import foo = bar.baz` counts as a concrete usage of `bar`, but this should depend on whether or not there is a concrete usage of `foo`. So we should figure that out before landing. I've added a test.

kdy1:
I updated test references to check if there's a regression.
I'll review this after eating lunch.

PR #2532, opened 2021-10-25
Refactor at rules

alexander-akait:
Many changes:
- fix bugs with `whitespaces`
- fix bug with `@keyframe "value" {}`
- fix bug with broken logic for `declaration`
- fix bug with parsing unknown at-rules (we need more fixes to fully support it)
- rename `Property` -> `Declaration` (like in specification)
- rename `DeclBlock` -> `Block` (like in spec), also we will have more block in future - `[`, `]`/`(`, `)`
- rename function to be align with specification
- remove `pub` from unnecessary prooerty
- more todos for future work


kdy1 (approved):
LGTM.
I only checked some of tests, but it's too much to review all of them 🤣

PR #2573, opened 2021-10-28
feat(css/ast): Add `SquareBracketBlock`

alexander-akait:
Added `SquareBracketBlock`, we have big problems with parsing `simple block`s https://www.w3.org/TR/css-syntax-3/#ref-for-simple-block%E2%91%A3

Shortly:
- We have three types of blocks - `(`/`)`, `[`/`]`, `{`/`}` in CSS (i.e simple block in spec)
- Each `simple block` can contain `component value`
- `Component value` can contain another `simple blocks`
- There are no commas/spaces or other values inside blocks, so we should remove empty (i.e. `SpacedValue`/ `CommaValue`)
- Also here fix when we can't parse grid multiple lines with whitespaces inside
- Also `simple blocks` can be part of `at-rules` value, so when we finish `simple blocks` we will simplify at-rules logic vey well 

Future fixes:
- I will content `(` and `)` to block too
- We should have two functions `parse_simple_block` and `parse_component_value` and they should call each other
- We still break AST for `+`/`-`/`/`/`*` but I will fix it after block improvements 

Also let use `children` instead `values`, because it is more general (postcss and other parser which I known uses `children` too)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2504,7 +2504,7 @@ dependencies = [
 
 [[package]]
 name = "swc_css"
-version = "0.25.0"
+version = "0.26.0"
 dependencies = [
  "swc_css_ast",
  "swc_css_codegen",
@@ -2515,7 +2515,7 @@ dependencies = [
 
 [[package]]
 name = "swc_css_ast"
-version = "0.23.0"
+version = "0.24.0"
 dependencies = [
  "is-macro",
  "serde",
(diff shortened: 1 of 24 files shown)


alexander-akait:
@kdy1 Our tests are broken:
```
running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

locally I have the same problem

PR #2583, opened 2021-10-29
fix(es/transforms/cjs): Allow mixing named exports and export stars.

### LINKED ISSUE #2548: export {default as name} can create duplicate symbols
**Describe the bug**

I frequently re-export the `default` export from another file as a named export.  When doing so with `babel`, it assigns the new named export inside the `_exportNames` array.  With `swc` this doesn't happen.  This causes apps that import my library to get an error:  `TypeError: Cannot redefine property:`

**Input code**

```
export {default as f} from "./f";
export * from "./f";

export {default as g} from "./g";
export * from "./g";
```

**Config**

```json
{
    "module": {
      "type": "commonjs"
    },
    "jsc": {
      "parser": {
        "syntax": "ecmascript",
        "jsx": true
      },
      "transform": {
          "react": {
              "runtime": "automatic"
          }
      },
      "loose": true
    }
  }
```

**Expected behavior**
With babel, I will find:

```
var _exportNames = {
  f: true,
  g: true
};
```

With `swc`, it produces:

```
var _exportNames = {
};
```

Where this is important is later on:

```
Object.keys(_f).forEach(function(key) {
    if (Object.prototype.hasOwnProperty.call(_exportNames, key)) return; // <- This doesn't return because _exportNames is wrong
    if (key in exports && exports[key] === _f[key]) return; 
        enumerable: true,
        get: function() {
            return _f[key];
        }
    });
});
```

**Version**
```
npx swc --version

@swc/cli: 0.1.51
@swc/core: 1.2.103
```

**Additional context**
I'd be happy to help track this down, if it's found to be a valid bug.

JeremyGrieshop:
This is my attempt at fixing #2548. 

---

 - Closes #2548

---

swc_ecma_transforms_module:
 - `common_js`: Fix _exportNames when exporting aliases and using `export *`s . (#2548)


ecmascript/transforms/module/src/common_js.rs
```diff
@@ -155,6 +155,19 @@ where
                             ref specifiers,
                             ..
                         })) => {
+                            // handle: export {sym as alias1, alias2, ...}
+                            for ExportNamedSpecifier { orig, exported, .. } in
+                                specifiers.into_iter().map(|e| match e {
```
kdy1, line 160: I think `filter_map` with `None` for other cases is better.
(Although I don't have concrete input which breaks this)

ecmascript/transforms/module/src/common_js.rs
```diff
@@ -155,6 +155,19 @@ where
...
+                                    ExportSpecifier::Named(e) => e,
+                                    _ => unreachable!(),
+                                })
+                            {
+                                if !exported.is_none() {
```
kdy1, line 165: You can use

```rust
if let Some(exported). = &exported {
    exports.push(exported.sym.clone());
}
```

kdy1 (changes requested):
Thanks!
Nice work, but I found some small issues.

ecmascript/transforms/module/src/common_js.rs
JeremyGrieshop: I've pushed a change to try to address this.  I hope it's in line with what you were thinking - I had previously used a pattern that I saw was elsewhere in the code.

kdy1 (approved):
LGTM. Thanks!

PR #2584, opened 2021-10-29
feat(css/parser): Added `OperatorValue` node

alexander-akait:
Added `OperatorValue` AST Node, it represents any `delim` tokens.

For example:
```css
a {
   font: italic small-caps bold 12px/30px Georgia, serif;
}
```

is valid CSS, but `/` should be not consider as division. `/` should be consider as `division` only for math functions like `calc()`/`min()`/`max()`/etc, we will improve them late.

The only thing I doubt is the name. 
Alternative - `DelimiterValue`.
Note - this node will be used for `,` in future.


alexander-akait:
@kdy1 Can you all and give your opinion?

kdy1:
Is this a groundwork for 

> / should be not consider as division. / should be consider as division only for math functions like calc()/min()/max()

?

I think I didn't get the point of this kind of node. Is it defined on the css specification?

alexander-akait:
@kdy1 

> I think I didn't get the point of this kind of node. Is it defined on the css specification?

In CSS spec no division, `/` can have many meanings, in

```css
a {
  font: italic small-caps bold 12px/30px Georgia, serif;
}
```

it is just separator in declaration value.

In:

```css
div {
  width: calc(100px / 2);
}
```

it is `/`` division.

Generally in CSS spec we have `<delim-token>`:

> <delim-token> has a value composed of a single code point.

So `*`, `+`, `-` and other not reserved characters (and even `,` (comma)) are delim characters.

Therefore, we need AST node in order to present it in AST. `OperatorValue` do it, but the name worries me a little, it can be misleading, so I suggest to use `DelimiterValue` and want to heart your opinion here.

alexander-akait:
@kdy1 Friendly ping

css/parser/tests/recovery/delim-token/ampersand/output.swc-stderr
```diff
@@ -1,6 +0,0 @@
-error: Expected Declaration value
```
kdy1, line 1: This test should have error message, as it's in `recovery` test.

css/parser/tests/recovery/delim-token/asterisk/output.json
```diff
@@ -64,59 +64,34 @@
         },
         "items": [
           {
-            "type": "Tokens",
```
kdy1, line 67: Is this change intentional?

css/parser/tests/recovery/delim-token/asterisk/output.swc-stderr
```diff
@@ -1,6 +0,0 @@
-error: Expected Declaration value
```
kdy1, line 1: Same here, it should report an error.

css/parser/tests/recovery/delim-token/bar/output.swc-stderr
```diff
@@ -1,6 +0,0 @@
-error: Expected Declaration value
```
kdy1, line 1: Same here.

kdy1 (changes requested):
I found some issues with updated test references, but I added review only on first few files as other cases are similar to them.


And it seems like tests are broken. Can you rebase it?

css/parser/tests/recovery/delim-token/bar/output.swc-stderr
alexander-akait: Yes, it is valid output now, we need to move them in other place

alexander-akait:
@kdy1 I would also like to know what you think about the `OperatorValue` name

kdy1:
Oh, I prefer `OpValue` over `OperatorValue`. Thanks! I missed it.

alexander-akait:
Close in favor https://github.com/swc-project/swc/pull/3402

PR #2585, opened 2021-10-30
fix(swc): Use standard base64 charset for inlined source maps

Kaciras:
Change inlined source map encoder to use the standard charset.

According to [Wikipedia](https://en.wikipedia.org/wiki/Data_URI_scheme):

>  Note that Base64-encoded data: URIs use the standard Base64 character set (with '+' and '/' as characters 62 and 63) rather than the so-called "URL-safe Base64" character set.

This fixes WebStorm cannot load source maps when using `@swc/jest`.


kdy1 (approved):
Thanks!

PR #2589, opened 2021-10-30
fix(es/lexer): Fix parsing of interpreter

antoniomuso:
Hello, I'm new to this codebase, During code reading, I found this strange code, and I think this is the correct implementation. Why is there an infinite loop?


kdy1 (approved):
LGTM. Thanks!

PR #2593, opened 2021-10-30
refactor(css/lexer): Refactor lexer to follow spec

alexander-akait:
@kdy1 Big lexer refactor:
1. Comments for easy maintenance
2. Fix order according spec
3. Simplify code
4. Reduce number of `self.input.cur()` 
5. Fix bug with `BadStr`
6. Fix bug with `BadURL`
7. Fix bug with invalid escape (partial)
8. Resolve couple TODO
9. More TODO fix invalid spec cases

I think that in the next 2-3 PRs I will finish lexer and it will be stable, it means other tools can use it, for example for bundling or simple minification, or any tooling.

The only thing I would like to ask - some places in spec require to return already consumed token, do we have api or better to implement this only for CSS lexer?


kdy1 (approved):
Thanks!
It was huge, but I think it's fine.

Issue #1733, opened 2021-05-23
Cannot resolve value from re-exported module


atsikov:
**Describe the bug**
swc returns `undefined` instead of exported const value when module is re-exported from upper-level index with circular dependencies

**Input code**

`src/components/A.ts`
``` js
export A = 'fooA'
```

`src/components/B.ts`
``` js
import { A } from '..' // <-- This import resolves A as undefined bacause of "B.ts -> index.ts -> B.ts" loop
import { A } from '../A' // <-- This import resolves A as expected

export B = `${A}'fooB`
```

`src/index.ts`
``` js
export * from './A'
export * from './B' // <- without this export, "import { A } from '..'" in B.ts works as expected
```

**Expected behavior**
swc correctly resolves modules even if there are circular references in index file with re-exports

**Version**
The version of @swc/core: 1.2.58

kdy1:
Are you using spack?

atsikov:
@kdy1 I was using `swc-loader`

chin2km:
Ran into this super unfortunate issue after migrating a project to `@vitejs/plugin-react-swc`.
Whats supper unfortunate is it does not throw any build errors but only runtime errors which costed much to the business.

PR #2598, opened 2021-10-31
fix(es/transforms/cjs): Preserve order of imports for reexports

### LINKED ISSUE #2594: Existing order of generated requires/defines can be problematic for inter-dependency relationships
**Describe the bug**

When importing from multiple modules, the current version of `swc` generates all of its `require` statements up front, followed by all of its `Object.defineProperty()` calls later.  You'll notice that `babel` does these one at a time, and this is important under certain conditions.  Consider the following input code:

**Input code**

```
export * from "./A";
export * from "./B";
```

The compiled commonjs results in:

```
"use strict";
    value: true
});
var _exportNames = {
};
var _a = _interopRequireWildcard(require("./A")); // <- all of the requires first
var _b = _interopRequireWildcard(require("./B")); // <- all of the requires first
    if (obj && obj.__esModule) {
        return obj;
    } else {
        var newObj = {
        };
        if (obj != null) {
            for(var key in obj){
                    };
                    if (desc.get || desc.set) {
                    } else {
                        newObj[key] = obj[key];
                    }
                }
            }
        }
        newObj.default = obj;
        return newObj;
    }
}
Object.keys(_a).forEach(function(key) {
    if (Object.prototype.hasOwnProperty.call(_exportNames, key)) return;
    if (key in exports && exports[key] === _a[key]) return;
    Object.defineProperty(exports, key, { // <- finally, our defineProperty calls
        enumerable: true,
        get: function() {
            return _a[key];
        }
    });
});
Object.keys(_b).forEach(function(key) {
    if (Object.prototype.hasOwnProperty.call(_exportNames, key)) return;
    if (key in exports && exports[key] === _b[key]) return;
    Object.defineProperty(exports, key, {. // <- finally, our defineProperty calls
        enumerable: true,
        get: function() {
            return _b[key];
        }
    });
});
```

`babel` produces a pattern that "interweaves" them (`require, defineProperty, require, defineProperty`):

```
"use strict";

  value: true
});

var _A = require("./A");

Object.keys(_A).forEach(function (key) {
  if (key in exports && exports[key] === _A[key]) return;
    enumerable: true,
    get: function get() {
      return _A[key];
    }
  });
});

var _B = require("./B");

Object.keys(_B).forEach(function (key) {
  if (key in exports && exports[key] === _B[key]) return;
    enumerable: true,
    get: function get() {
      return _B[key];
    }
  });
});
```

**Config**

```json
{
    "test": ".*.js$",
    "module": {
      "type": "commonjs"
    },
    "jsc": {
      "parser": {
        "syntax": "ecmascript",
        "jsx": true
      },
      "transform": {
          "react": {
              "runtime": "automatic"
          }
      },
      "loose": true
    }
  },
```

**Expected behavior**
Everything will compile, and quite often, bundlers such as `webpack` will get the resolutions correctly.  However, I've found `jest` for example, when importing the above example, may encounter a class from `B` that depends on `A`.  This breaks, because `A` hasn't been defined yet because the `defineProperty` hasn't been declared yet, resulting in a class construction with an `undefined` value for `A`.

By taking the `swc` output, and simply moving the requires around, everything runs just fine.

**Version**
The version of @swc/core:

```
@swc/cli: 0.1.51
@swc/core: 1.2.105
```

**Additional context**

kwonoj:
- closes #2594.

This PR attempts to order of the imports / reexports in transpiled runtime codebase. #2594 itself sufficiently explains the issue itself. It could be an implementation detail but so far based on observation tsc / babel works in the same way, which makes think it'd better to align as well. As issue described normally this won't appear but if there's some sort of circular dependencies, or reexports have dependencies across each other it may reveal different behavior.

For the change itself, PR changes the order of injecting handled export_all stmt by putting it after require stmt is created. In result, export_all stmt doesn't use extra_stmt (which appends to stmt at the end). Also changed `exports_alls` to map to stmt itself - when encountering ExportAll stmt it is gauranteed we'll need stmt anyway - hashmap contains generated stmt instead of only keeping name of export. This allowed easier lookup at the end.

Lastly, I noticed there is prior test cases from https://github.com/swc-project/swc/pull/133/ (https://github.com/swc-project/swc/pull/133/files#diff-6d08f604fc5444313bf8d80ca76a84f5abbc95129eb4e1311592637fbb17b089R1075) , looks like it is coming from original transform impl. Test cases are updated without adding new one.

### DIFF
--- a/ecmascript/transforms/module/src/common_js.rs
+++ b/ecmascript/transforms/module/src/common_js.rs
@@ -85,7 +85,7 @@ where
         let mut exports = vec![];
         let mut initialized = IndexSet::default();
 
-        let mut export_alls: AHashMap<JsWord, ModuleItem> = Default::default();
+        let mut export_alls: AHashMap<JsWord, Ident> = Default::default();
         // Used only if export * exists
         let mut exported_names: Option<Ident> = None;
 
@@ -233,11 +233,9 @@ where
                             }
 
                             export_alls.entry(export.src.value.clone()).or_insert(
-                                ModuleItem::Stmt(scope.handle_export_all(
-                                    quote_ident!("exports"),
-                                    exported_names.clone(),
-                                    export,
-                                )),
+                                scope
+                                    .import_to_export(&export.src, true)
+                                    .expect("Export should exists"),
                             );
 
                             drop(scope);
@@ -707,9 +705,13 @@ where
                 }
             }
 
-            let export_all = export_alls.remove(&src);
-            if let Some(export_all) = export_all {
-                stmts.push(export_all);
+            let exported = export_alls.remove(&src);
+            if let Some(export) = exported {
+                stmts.push(ModuleItem::Stmt(Scope::handle_export_all(
+                    quote_ident!("exports"),
+                    exported_names.clone(),
+                    export,
+                )));
             }
         }
 
--- a/ecmascript/transforms/module/src/common_js.rs
+++ b/ecmascript/transforms/module/src/common_js.rs
@@ -9,8 +9,8 @@ use std::{
     cell::{Ref, RefCell, RefMut},
     rc::Rc,
 };
-use swc_atoms::js_word;
-use swc_common::{FileName, Mark, Span, DUMMY_SP};
+use swc_atoms::{js_word, JsWord};
+use swc_common::{collections::AHashMap, FileName, Mark, Span, DUMMY_SP};
 use swc_ecma_ast::*;
 use swc_ecma_transforms_base::helper;
 use swc_ecma_utils::{
@@ -84,7 +84,10 @@ where
 
         let mut exports = vec![];
         let mut initialized = IndexSet::default();
-        let mut export_alls = vec![];
+
+        let mut export_alls: AHashMap<JsWord, ModuleItem> = Default::default();
+        // Used only if export * exists
+        let mut exported_names: Option<Ident> = None;
 
         for item in items {
             self.in_top_level = true;
@@ -174,6 +177,7 @@ where
                     }
                     drop(scope);
                     drop(scope_ref_mut);
+
                     match item {
                         ModuleItem::ModuleDecl(ModuleDecl::ExportAll(export)) => {
                             self.scope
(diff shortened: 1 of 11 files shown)


JeremyGrieshop:
Thank you for tackling this!  I suspect the test cases for my recent merge (`tests/fixture/issue-2548/`) also need to be updated in order for the workflow tests to pass.

kwonoj:
> Thank you for tackling this! I suspect the test cases for my recent merge (tests/fixture/issue-2548/) also need to be updated in order for the workflow tests to pass.

Thanks, I tried to update test cases as well.

tests/fixture/issue-2548/case2/output/index.js
```diff
@@ -8,10 +8,17 @@ Object.defineProperty(exports, "X", {
         return _z.default;
     }
 });
-var _exportNames = {
```
JeremyGrieshop, line 11: I'm afraid the current PR may actually regress issue #2548 in this state.

Here, _exportNames is removed by this PR in this case.  Which means, later on we will may an `Object.defineProperty()` for `X`, if `X` is exported by `Z` using `export * from "./Z"`.  This will result in multiple calls to `Object.defineProperty` for the same property.

tests/fixture/issue-2548/case2/output/index.js
```diff
@@ -8,10 +8,17 @@ Object.defineProperty(exports, "X", {
...
-    X: true
-};
 var _z = _interopRequireWildcard(require("./Z"));
+Object.keys(_z).forEach(function(key) {
+    if (key === "default" || key === "__esModule") return;
+    if (key in exports && exports[key] === _z[key]) return;
+    Object.defineProperty(exports, key, {
```
JeremyGrieshop, line 15: Here is where the call to `Object.defineProperty` can throw the error.  Previously, we injected:

`if (Object.prototype.hasOwnProperty.call(_exportNames, key)) return;`

to guard against this.

kwonoj: This is something I'm thinking around, there was previously condition to check if there are multiple exports which couldn't be used with changed.

ecmascript/transforms/module/src/common_js.rs
```diff
@@ -84,7 +84,10 @@ where
 
         let mut exports = vec![];
         let mut initialized = IndexSet::default();
-        let mut export_alls = vec![];
+
+        let mut export_alls: AHashMap<JsWord, ModuleItem> = Default::default();
```
kdy1, line 88: I think storing `ModuleItem` in a hash map will be very slow, due to `memmove`.
`ModuleItem` is a big struct, so I think this case be performance issue.

ecmascript/transforms/module/src/common_js.rs
```diff
@@ -174,14 +177,71 @@ where
...
+                                stmts.push(ModuleItem::Stmt(Stmt::Decl(Decl::Var(VarDecl {
+                                    span: DUMMY_SP,
+                                    kind: VarDeclKind::Var,
+                                    decls: vec![VarDeclarator {
+                                        span: DUMMY_SP,
+                                        name: Pat::Ident(exported_names_ident.clone().into()),
+                                        init: Some(Box::new(Expr::Object(ObjectLit {
+                                            span: DUMMY_SP,
+                                            props: exports
+                                                .clone()
```
kdy1, line 204: Is this clone necessary?

kdy1 (commented):
Thank you!

I have some concerns about changes.

ecmascript/transforms/module/src/common_js.rs
kwonoj: good point, was thinking that too but wasn't sure about actual implication so didn't try early optimizations. any suggestions?

ecmascript/transforms/module/src/common_js.rs
kwonoj: If I recall corretly yes, there was a move occurred somewhere.

ecmascript/transforms/module/src/common_js.rs
kwonoj: The reason I created moduleitem early is at the moment pushing stmt (https://github.com/swc-project/swc/pull/2598/files#diff-506530d02689c3c49a06cbec6fb8c9ffd7b1df4c551dcac29d8bd08d0248590fR712) there's mutable borrow to scope exists, so can't call `handle_export_all`. Curious if there's easy workaround.

JeremyGrieshop:
This may possibly resolve https://github.com/swc-project/swc/issues/1733, too.

ecmascript/transforms/module/src/common_js.rs
```diff
@@ -699,6 +704,15 @@ where
                     stmts.push(require.into_stmt().into());
                 }
             }
+
+            let exported = export_alls.remove(&src);
+            if let Some(export) = exported {
+                stmts.push(ModuleItem::Stmt(Scope::handle_export_all(
```
kwonoj, line 710: Curious if this approach makes sense? it was not possible to call `handle_export_all` due to move occurrence, I took out part to create Ident (which required &mut self) hold by hashmap, and actual stmt creation occurs in here.

kdy1: Yes, I think it's correct approach.

kdy1 (approved):
LGTM.
I found one issue, but it's not related to this PR.
I'll patch it after merging this. Thanks!

ecmascript/transforms/module/src/common_js.rs
kdy1: What you did is right.
Storing only minimum information to create `ModuleItem` and actually create big structs (`ModuleItem` in this case) works.

kwonoj:
Appreciate for the review for both, @kdy1 and @JeremyGrieshop :bow:

PR #2617, opened 2021-11-01
refactor(css/parser): Refactor parser for at rules

alexander-akait:
Refactor at-rules + fix edge cases bugs + tests + recovery mode for bad parsed at-rules + remove unnecessary methods + prepare for future improvements


kdy1 (changes requested):
Finally, we have some error recovery, thanks!
Code changes looks good to me.

But I'm not sure about the version. `cargo mono` depends on `crates.io`, so I think `cargo mono bump should be done again`.
Alternatively, you can just skip bumping version and I'll use a script to bump version.

(I don't checkout PRs anymore, so it will be separate commit)

alexander-akait:
> Alternatively, you can just skip bumping version and I'll use a script to bump version.

yes, it will be good

kdy1 (approved):
Thanks!
I force pushed to remove version changes from commit.

PR #2618, opened 2021-11-01
feat(css/parser): added `RoundBracketBlock`

alexander-akait:
Added `RoundBracketBlock`, we have 2 edge cases here:
1. custom property values (i.e. `:root { --foo: any-value }`), should be parser with `<any-value>` and use tokens (I will fix it in the next PR)
2. `{`/`}` block should be parsed as `simple_block` (will fix it after `1.` and merge this)


css/Cargo.toml
```diff
@@ -6,11 +6,11 @@ edition = "2018"
 license = "Apache-2.0/MIT"
 name = "swc_css"
 repository = "https://github.com/swc-project/swc.git"
-version = "0.26.0"
+version = "0.27.0"
 
 [dependencies]
 swc_css_ast = {version = "0.24.0", path = "./ast"}
```
kdy1, line 12: I think this is a breaking change of `swc_css_ast`.

css/parser/tests/fixture/rome/custom-properties/input.css
```diff
@@ -18,7 +18,7 @@
 	--at-keyword-known-block: @media {};
 	--cdo-at-top-level: <!--;
 	--cdc-at-top-level: -->;
-	--semicolon-not-top-level: (;);
+	/*--semicolon-not-top-level: (;);*/
```
kdy1, line 21: Why is this line commented out?

kdy1 (changes requested):

css/parser/tests/fixture/rome/custom-properties/input.css
alexander-akait: @kdy1 It is invalid, `;` is not allowed here, ideally we should move it in recovery mode, I will do it

alexander-akait:
@kdy1 Ready for review, we still lost recovery mode in some places, but it will not work before too, good recovery mode for values will be in future - after we fix delim token and `simple blocks` (i.e. `{`/`}`/` (will be after this),(`/`)` (here), `[`/`]` (solved))

kdy1 (approved):
Thanks!
All changes look good to me.

Issue #2606, opened 2021-10-31
does not allow reassign ref to exported fn for typescript


kwonoj:
### Describe the bug

This is followup edge case for https://github.com/swc-project/swc/issues/2549.

When assignment is a typescript expression like

```
export function warn() {
  throw new Error("this should not be called");
}

export const test = {};

Object.defineProperty(test, "warn", {
  get: () => warn,
  set: (v) => {
    (warn as any) = v;
  },
});
```

`(warn as any) = v;` won't reassign exports.

First glimpse this looks like issue around AST:

``` 
//(warn as any) = v;

TsAs(
        TsAsExpr {
            span: Span {
                lo: BytePos(
                    733,
                ),
                hi: BytePos(
                    744,
                ),
                ctxt: #0,
            },
            expr: Ident(
                Ident {
                    span: Span {
                        lo: BytePos(
                            733,
                        ),
                        hi: BytePos(
                            737,
                        ),
                        ctxt: #0,
                    },
                    sym: Atom('warn' type=inline),
                    optional: false,
                },
            ),
....
```

```
// warn = v;

Ident(
          BindingIdent {
              id: Ident {
                  span: Span {
                      lo: BytePos(
                          838,
                      ),
                      hi: BytePos(
                          842,
                      ),
                      ctxt: #0,
                  },
                  sym: Atom('warn' type=inline),
                  optional: false,
              },
              type_ann: None,
          }
```

in case tsc it doesn't create bindingident.

### Expected behavior

ts expression also should reassign exports

### Version

master

kwonoj:
Am I understanding correct this need to be updated other than transform, or it's still something to be fixed with transforms?

kwonoj:
I may wrong still though, peeking issue it looks like assignment is not evaluated when there's paren (https://github.com/swc-project/swc/blob/3280b4cd7a147cf72f517afa7a11cc5c39ccec06/ecmascript/parser/src/parser/pat.rs#L537) . Bit sure `how to fix` part, returning pat from inner expr breaks ast.

kwonoj:
@kdy1 would you mind share some thoughts, first if above seems correct triage & secondly some high level thoughts how to fix issue? :pray: I'll spin some time myself to dig further meanwhile.

kdy1:
[...]

PR #2631, opened 2021-11-02
fix(es/transforms/tsc): coerce bindingident with tsasexpr

kwonoj:
- closes https://github.com/swc-project/swc/issues/2606

This PR attempts to allow reassign fn when it is wrapped with `as` casting paren, like `(warn as any) = v`.

Change follows suggestion to coerece Pat::Ident if given Pat is wrapped paren expr - but instead of after visiting does check and coerece before to distinguish between plain assignment like `(x) = v`.

Peeking emitted output when we try to treat both as same it seems possible to treat both as same but can't 100% sure around minified outputs. For now, choose safest path to not regress existing fixtures.

### DIFF
--- a/ecmascript/transforms/typescript/src/strip.rs
+++ b/ecmascript/transforms/typescript/src/strip.rs
@@ -2217,6 +2217,34 @@ where
         d.visit_mut_children_with(self);
         d.definite = false;
     }
+
+    fn visit_mut_pat_or_expr(&mut self, node: &mut PatOrExpr) {
+        // Coerce bindingident to assign expr where parenthesis exists due to TsAsExpr
+        // like (warn as any) = v;
+        // We want to do this _before_ visiting children, otherwise handle_expr
+        // wipes out TsAsExpr and there's no way to distinguish between
+        // plain (x) = y vs. (x as any) = y;
+        match node {
+            PatOrExpr::Pat(pat) => match &**pat {
+                Pat::Expr(expr) => match &**expr {
+                    Expr::Paren(ParenExpr { expr, .. }) => match &**expr {
+                        Expr::TsAs(TsAsExpr { expr, .. }) => match &**expr {
+                            Expr::Ident(ident) => {
+                                *node = PatOrExpr::Pat(Box::new(Pat::Ident(ident.clone().into())));
+                            }
+                            _ => (),
+                        },
+                        _ => (),
+                    },
+                    _ => (),
+                },
+                _ => (),
+            },
+            _ => (),
+        }
+
+        node.visit_mut_children_with(self);
+    }
 }
 
 fn module_ref_to_expr(r: TsModuleRef) -> Expr {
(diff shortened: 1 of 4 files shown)


ecmascript/transforms/typescript/src/strip.rs
```diff
@@ -2217,6 +2217,34 @@ where
...
     }
+
+    fn visit_mut_pat_or_expr(&mut self, node: &mut PatOrExpr) {
+        // Coerce bindingident to assign expr where parenthesis exists due to TsAsExpr
+        // like (warn as any) = v;
+        // We want to do this _before_ visiting children, otherwise handle_expr
+        // wipes out TsAsExpr and there's no way to distinguish between
+        // plain (x) = y vs. (x as any) = y;
+        match node {
+            PatOrExpr::Pat(pat) => match &**pat {
```
kwonoj, line 2228: I'm feeling this nested match to deref Box is bit unidomatic, potentially due to lacks of my knowledge and would like to simplify if possible.

kdy1: It's possible with nightly rustc, but there are many swc users who use stable rustc, so it can't be simplified for now.

ecmascript/transforms/typescript/src/strip.rs
```diff
@@ -2217,6 +2217,34 @@ where
...
+                        _ => (),
+                    },
+                    _ => (),
+                },
+                _ => (),
+            },
+            _ => (),
+        }
+
+        node.visit_mut_children_with(self);
```
kdy1, line 2246: I think checking for `Expr::Ident` after `node.visit_mut_children_with(self);` would be more clear and hande more cases.

kdy1 (changes requested):
Thanks!

ecmascript/transforms/typescript/src/strip.rs
kwonoj: Yes, I used box signatuer for the internal plugin experiment. Sad it's not stable.

ecmascript/transforms/typescript/src/strip.rs
kwonoj: Do you mean before `match` to traverse node to find out ParenExpr? I tried it and hit by this https://github.com/swc-project/swc/pull/2631/files#diff-aee7604a9d35c517e7e1637b34d66b7ce8de17cc407ba6236d314471d00d4b5eR2224-R2226 - which invalidates few existing test cases like `(x) = v`. It's different to TsAsExpr which strips paren (which is expected) need to preserve as-is.

kwonoj: visit_mut_children triggers to visit expr, and strips out TsAsExpr with `handle_expr` so there wasn't way to distinguish if it's coming from as-casting vs. ecmascript runtime paren.

kdy1: Oh, I see, then it's fine.

kdy1 (approved):
Thanks!

PR #2638, opened 2021-11-03
fix(es/transforms/jsx): collapses multiple whitespace charactars

### LINKED ISSUE #2208: swc collapses multiple whitespace characters in JSX expressions.
**Describe the bug**
swc collapses two or more spaces into one space in JSX expressions. This causes unexpected formatting when using `white-space: pre`.

**Input code**
```tsx
const Component = () => {
  return <pre>|    |    |</pre>;
}
```

**Config**

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": true
    }
  }
}
```

**Expected behavior**

Whitespace is preserved. Compare output from swc:
```tsx
var Component = function() {
    return(/*#__PURE__*/ React.createElement("pre", null, "| | |"));
};
```

to output from babel:
```tsx
var Component = function Component() {
  return /*#__PURE__*/React.createElement("pre", null, "|    |    |");
};
```

**Version**
1.2.84

**Additional context**
This case for this is ASCII art rendered in React. The current workaround is wrapping every expression in a template string to preserve spaces.

RiESAEX:
attempt to fix #2208


ecmascript/transforms/react/src/jsx/mod.rs
```diff
@@ -1130,42 +1130,28 @@ fn to_prop_name(n: JSXAttrName) -> PropName {
...
-    for (last, s) in s
-        .split(|c: char| c != '\u{a0}' && c.is_ascii_whitespace())
-        .filter(|s| !s.is_empty())
+    static SPACE_START: Lazy<Regex> = Lazy::new(|| Regex::new("^[ ]+").unwrap());
+    static SPACE_END: Lazy<Regex> = Lazy::new(|| Regex::new("[ ]+$").unwrap());
+    t.lines()
+        .map(|s| s.replace('\t', " "))
+        .enumerate()
+        .filter(|(_, v)| v.len() != 0)
         .identify_last()
```
kdy1, line 1139: This seems like an inefficient code, as it allocates lots of `String`s.
I think using a buffer is preferrable.

kdy1 (changes requested):
Thanks!

ecmascript/transforms/react/src/jsx/mod.rs
```diff
@@ -1130,41 +1130,34 @@ fn to_prop_name(n: JSXAttrName) -> PropName {
...
-        .split(|c: char| c != '\u{a0}' && c.is_ascii_whitespace())
-        .filter(|s| !s.is_empty())
-        .identify_last()
-    {
-        buf.push_str(s);
-        if !last {
-            buf.push(' ');
+    static SPACE_START: Lazy<Regex> = Lazy::new(|| Regex::new("^[ ]+").unwrap());
+    static SPACE_END: Lazy<Regex> = Lazy::new(|| Regex::new("[ ]+$").unwrap());
+    let mut buf = String::from("");
```
kdy1, line 1135: Can you this change to `String::new()`?

kdy1 (changes requested):
Mostly LGTM.
Thnaks!

ecmascript/transforms/react/src/jsx/mod.rs
RiESAEX: of course.
btw, I looked into some failed test case, noticing their output is different with the output which I got with tsc playground, should i edit them? most of them added in #2456

kdy1: Yes, those tests exist to show difference of output of lots of inputs.
You can run `UPDATE=1 cargo test` from the repository root to update them.
Note that you may need to run it multiple time, until it exists normally.

kdy1 (approved):
LGTM. Thanks!

PR #2639, opened 2021-11-03
feat(css/ast): Added `SelectorList`

alexander-akait:
Improve AST for selectors

> https://www.w3.org/TR/selectors-4/#typedef-class-selector

> \<selector-list\> = \<complex-selector-list\>

> \<complex-selector-list\> = \<complex-selector\>#

> \<compound-selector-list\> = \<compound-selector\>#

Added `SelectorList` node, ideally we should avoid any `Vec<Something>`, because in most of time it means missing AST Node

Also to solve https://github.com/swc-project/swc/pull/2358, we need fix logic, this PR is part of these fixes

There are a lot of snapshots changes, but they are simple


kdy1 (approved):
LGTM. Thanks!

PR #2640, opened 2021-11-03
(Fix #2016) Add a es2015 transform for const support

### LINKED ISSUE #2016: `const` is being transformed to a `var` without it being read-only as well
Starting with the following script:

```js
const a = 2;
var ret = false;
try {
	a = 3;
} catch (e) {
	if (e instanceof TypeError) {
		ret = true;
	}
}

ret;
```

**Expected behavior**
Babel transpiles to:
```js
"use strict";

function _readOnlyError(name) { throw new TypeError("\"" + name + "\" is read-only"); }

var a = 2;
var ret = false;

try {
  3, _readOnlyError("a");
} catch (e) {
  if (e instanceof TypeError) {
    ret = true;
  }
}

ret;
```
**See the _readOnlyError function added

**Actual Behavior**
While swc transforms to:
```js
"use strict";
function _instanceof(left, right) {
    if (right != null && typeof Symbol !== "undefined" && right[Symbol.hasInstance]) {
        return right[Symbol.hasInstance](left);
    } else {
        return left instanceof right;
    }
}
var a = 2;
var ret = false;
try {
    a = 3;
} catch (e) {
    if (_instanceof(e, TypeError)) {
        ret = true;
    }
}
ret;
```

**Config**
The default config:
```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false,
      "dynamicImport": false,
      "privateMethod": false,
      "functionBind": false,
      "exportDefaultFrom": false,
      "exportNamespaceFrom": false,
      "decorators": false,
      "topLevelAwait": false,
      "importMeta": false
    },
    "transform": null,
    "target": "es5",
    "loose": false,
    "externalHelpers": false,
    "keepClassNames": false
  }
}
```

**Version**
The version of @swc/core: Current

**Additional context**

0xe:
Adds support for checking mutation on identifiers declared as const.

Transforms

```javascript
// vars aren't affected
var a;
a = 2;
a = 3;

try {
	var a = 4;
	a = 5;
} catch {}

// consts with initializers work as expected
const b = 2;
b = 3;

try {
	const b = 4;
	b = 5;
} catch {};

// FIXME: no-initializer consts should throw Syntax Error
const c;
c = 3;
c = 4;

try {
	const c;
	c = 5;
	c = 6;
} catch {};

// multi-declarations work as expected
const d = 42, e = 42;
d = 35;
e = 45;

// array/object setters work
const f = {a: 1, b: 2};
f["a"] = 3;

const g = [1, 2, 3, 4];
g[1] = 5;
```

into 

```javascript
// vars aren't affected
var a;
a = 2;
a = 3;
try {
    var a = 4;
    a = 5;
} catch (e) {
}
// consts with initializers work as expected
var b = 2;
3, _readOnlyError("b#1");
try {
    var b = 4;
    5, _readOnlyError("b#2");
} catch (e) {
}
;
// FIXME: no-initializer consts should throw Syntax Error
var c;
3, _readOnlyError("c#1");
4, _readOnlyError("c#1");
try {
    var c;
    5, _readOnlyError("c#3");
    6, _readOnlyError("c#3");
} catch (e) {
}
;
// multi-declarations work as expected
var d = 42, e = 42;
35, _readOnlyError("d#1");
45, _readOnlyError("e#1");
// array/object setters work
var f = {
    a: 1,
    b: 2
};
3, _readOnlyError("f#1");
var g = [
    1,
    2,
    3,
    4
];
5, _readOnlyError("g#1");
```

TODO:
- Raise a syntax error for [const declarations without initializers](https://tc39.es/ecma262/multipage/ecmascript-language-statements-and-declarations.html#sec-let-and-const-declarations-static-semantics-early-errors)
- Tests

### DIFF
--- a/ecmascript/transforms/compat/src/es2015/const_readonly.rs
+++ b/ecmascript/transforms/compat/src/es2015/const_readonly.rs
@@ -1,0 +1,141 @@
+use swc_common::DUMMY_SP;
+use swc_ecma_ast::*;
+use swc_ecma_transforms_base::helper;
+use swc_ecma_visit::{as_folder, noop_visit_mut_type, Fold, VisitMut, VisitMutWith};
+
+/// # Example
+///
+/// # In
+///
+/// ```js
+/// const a = 32;
+/// a = 4
+///
+/// const b = [1, 2, 3];
+/// b[0] = 5;
+/// ```
+///
+/// ## Out
+///
+/// ```js
+/// function _readOnlyError(name) { throw new TypeError("\"" + name + "\" is read-only"); }
+///
+/// var a = 32;
+/// 4, _readOnlyError("a#1");
+///
+/// var b = [1, 2, 3];
+/// 5, _readOnlyError("b#1");
+/// ```
+pub fn check_constants() -> impl Fold + VisitMut {
+    as_folder(Consts::default())
+}
+
+fn find_lhs_id(lhs: &PatOrExpr) -> Option<Ident> {
+    fn find_lhs_id_from_expr(lhs: &Expr) -> Option<Ident> {
+        match lhs {
+            Expr::Member(m) => {
+                let obj = match &m.obj {
+                    ExprOrSuper::Super(_) => return None,
+                    ExprOrSuper::Expr(obj) => &**obj,
+                };
+                match obj {
+                    Expr::Ident(i) => Some(i.clone()),
+                    Expr::Member(..) => find_lhs_id_from_expr(&obj),
+                    _ => None,
+                }
+            }
+            _ => None,
+        }
+    }
+
+    match lhs {
+        PatOrExpr::Expr(e) => find_lhs_id_from_expr(&e),
+        PatOrExpr::Pat(pat) => match &**pat {
+            Pat::Expr(e) => find_lhs_id_from_expr(&e),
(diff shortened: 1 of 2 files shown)


kdy1:
Is this useful?
I think editors will warn about them, and if it should be reported, I think it's better to report at compile time instead of runtime

kdy1:
I'm going to add an error reporting for unresolved variables and reassignment to constants to resolve https://github.com/vercel/next.js/issues/30800

I'm bit worried because this PR can be a duplicate work.

0xe:
> Is this useful? I think editors will warn about them, and if it should be reported, I think it's better to report at compile time instead of runtime

The [spec](https://tc39.es/ecma262/multipage/ecmascript-language-statements-and-declarations.html#sec-let-and-const-declarations-static-semantics-early-errors) seems to talk about only checking if const has a initializer during lexing. 

For the behavior in this PR, I referred to how babel does it in the [transform-block-scoping](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-block-scoping/src/index.js#L411) plugin.

I agree that it would be better to raise it at compile time.

kdy1:
@0xe Then I'll make it compile-time error instead.

kdy1:
I started working on https://github.com/swc-project/swc/pull/2647, which will make errors like this compile-time error.

PR #2641, opened 2021-11-03
Feat added nesting selector

alexander-akait:
Added `NestingSelector` as separate node, if we need to support:
```
.foo {
  color: blue;
  & > .bar { color: red; }
}
```

i.e. nested nesting selector, we should keep them in `SubclassSelector` (same for TypeSelector (NamespacedName)), and validate it is first or not (I will fix it in future)

Also we fix here span for `*` selector

### DIFF
--- a/css/codegen/src/lib.rs
+++ b/css/codegen/src/lib.rs
@@ -695,15 +695,13 @@ where
 
     #[emitter]
     fn emit_compound_selector(&mut self, n: &CompoundSelector) -> Result {
-        let ctx = Ctx { ..self.ctx };
-
-        emit!(&mut *self.with_ctx(ctx), n.nesting_selector);
+        emit!(&mut *self.with_ctx(self.ctx), n.nesting_selector);
 
         if let Some(combinator) = &n.combinator {
             self.wr.write_punct(None, combinator.as_str())?;
         }
 
-        emit!(&mut *self.with_ctx(ctx), n.type_selector);
+        emit!(&mut *self.with_ctx(self.ctx), n.type_selector);
 
         self.emit_list(&n.subclass_selectors, ListFormat::NotDelimited)?;
     }
--- a/css/codegen/src/lib.rs
+++ b/css/codegen/src/lib.rs
@@ -689,16 +689,20 @@ where
     }
 
     #[emitter]
+    fn emit_nesting_selector(&mut self, _: &NestingSelector) -> Result {
+        punct!(self, "&");
+    }
+
+    #[emitter]
     fn emit_compound_selector(&mut self, n: &CompoundSelector) -> Result {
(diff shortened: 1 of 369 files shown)


kdy1:
Is this in specificaiton?

alexander-akait:
@kdy1 Here https://www.w3.org/TR/css-nesting-1/, it is draft, but if we need support `&` we should follow to this spec (node `&` is supported by css-in-js) and it should be separated AST node

alexander-akait:
@kdy1 Can you review this too?

css/codegen/src/lib.rs
```diff
@@ -688,17 +688,21 @@ where
...
+    fn emit_nesting_selector(&mut self, _: &NestingSelector) -> Result {
+        punct!(self, "&");
+    }
+
     #[emitter]
     fn emit_compound_selector(&mut self, n: &CompoundSelector) -> Result {
-        if n.has_nest_prefix {
-            punct!(self, "&");
-        }
+        let ctx = Ctx { ..self.ctx };
```
kdy1, line 698: Seems like a redundant code.

kdy1 (changes requested):
Thanks!
I found one tiny issue.

css/codegen/src/lib.rs
alexander-akait: Hm, yep, fixed

kdy1 (approved):
Thanks!
I think rustc is smart enough to remove useless calls...

alexander-akait:
> Thanks!
> I think rustc is smart enough to remove useless calls...

Yep, I found some more places with this, remove them in other PRs

vpontis:
@alexander-akait / @kdy1, thank you for the great work!

Does SWC support nested CSS? I tried to use nested CSS w/ the newest version of Next.js and I got this:

CleanShot 2021-11-09 at 16 44 48@2x

Example repo: https://github.com/vpontis/swc-nested-css

---

For more context, we want to switch lu.ma's Next.js project from babel to swc but we use [`@styled-jsx/plugin-scss`](https://yarnpkg.com/package/@styled-jsx/plugin-sass) to give us support for nested CSS selectors. 

We would be able to get rid of SCSS if we can use native nesting in CSS. (Or we could use SWC if you have support for SCSS in styled-jsx, but I'm guessing that will be difficult...)

alexander-akait:
@vpontis Will support (some limitations right now), but not ready yet, anyway it should parse, can't you provide full CSS?

vpontis:
Cool thanks, @alexander-akait!

So it's going to be a pretty big migration for us to get rid of SCSS. And it'll be even bigger if we also need to remove nested classes. 

Also we really like using nested CSS for development, so we are trying to preserve that functionality for DX.

PR #2642, opened 2021-11-03
refactor(css/parser): rename `NamespacedName` to `TypeSelector`

alexander-akait:
Refactor - rename `NamespacedName` to `TypeSelector`, we don't have namespaced name in spec, mdn uses `type selector` too https://developer.mozilla.org/en-US/docs/Web/CSS/Type_selectors


alexander-akait:
@kdy1 Ready for review, from spec:
```
<type-selector> = <wq-name> | <ns-prefix>? '*'
```

alexander-akait:
@kdy1 Can you review?

kdy1 (approved):
Thanks!
I only checked for code changes, as github is too slow

PR #2643, opened 2021-11-03
Refactor attribute selector

alexander-akait:
Fixed:
- Avoid `NamespacedRule`/`TypeSelector` for `name`, because it is invalid
- Rename properties according spec
- Fix logic for invalid attribute matcher
- Fix logic for invalid attribute value
- Simplify
- Handle spaces between `modifier`
- matcher value can be `Ident` (`Text` AST Node) or `Str` (`Str` AST Node)


alexander-akait:
@kdy1 Ready for review

css/ast/src/selector.rs
```diff
@@ -108,6 +91,34 @@ pub enum AttrSelectorOp {
...
+    #[tag("Text")]
+    Text(Text),
+}
+
+#[ast_node("AttributeSelector")]
+pub struct AttrSelector {
+    pub span: Span,
+    pub prefix: Option<Text>,
+    pub name: Text,
+    pub matcher: Option<AttrSelectorMatcher>,
```
kdy1, line 109: Is `Matcher` a word used in the specification?

css/parser/tests/fixture.rs
```diff
@@ -317,6 +317,7 @@ impl Visit for SpanVisualizer<'_> {
     mtd!(AtRule, visit_at_rule);
     mtd!(AtSelector, visit_at_selector);
     mtd!(AtTextValue, visit_at_text_value);
+
```
kdy1, line 320: This seems like a mistake?

kdy1 (commented):
Code changes look fine to me.
I just have some questions.

css/ast/src/selector.rs
alexander-akait: Yes

> \<attr-matcher\> = [ '~' | '|' | '^' | '$' | '*' ]? '='

https://www.w3.org/TR/selectors-4/#grammar

css/parser/tests/fixture.rs
alexander-akait: Yep

css/ast/src/selector.rs
kdy1: Thanks for the clarification!

alexander-akait:
@kdy1 fixed and ready for review

kdy1 (approved):
LGTM.
Thank you!

PR #2659, opened 2021-11-05
fix(css/parser): Exclude whitespace from span

alexander-akait:
Small fix, we should not include spaces before


kdy1 (approved):
LGTM. Thanks!

PR #2660, opened 2021-11-05
refactor(css/ast): Refactor complex selector to represent all css

alexander-akait:
Improvements: 
- combinator is part of `complex` selector, not combinator

> \<complex-selector\> = \<compound-selector\> [ \<combinator\>? \<compound-selector\> ]*

- combinator is AST node, now
- keep `Combinator` in `ComplexSelector` because it is part of complex selector
- fix bug with spans
- improve logic
- refactor `last_pos` to use from last selector


alexander-akait:
@kdy1 Ready for review, in near future we finish selectors and they will be stable :star:

css/ast/src/selector.rs
```diff
@@ -12,7 +12,16 @@ pub struct SelectorList {
 #[ast_node("ComplexSelector")]
 pub struct ComplexSelector {
     pub span: Span,
-    pub selectors: Vec<CompoundSelector>,
+    pub children: Vec<ComplexSelectorChildren>,
+}
+
+#[ast_node]
+#[derive(Is)]
+pub enum ComplexSelectorChildren {
```
kdy1, line 20: What's the purpose of this enum?

kdy1: I think this change may make implementing transforms harder.

css/ast/src/selector.rs
```diff
@@ -21,13 +30,18 @@ pub struct CompoundSelector {
...
 
+#[ast_node("Combinator")]
+pub struct Combinator {
+    pub span: Span,
+    pub value: CombinatorValue,
+}
+
 #[derive(StringEnum, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Hash, EqIgnoreSpan)]
-pub enum SelectorCombinator {
+pub enum CombinatorValue {
```
kdy1, line 44: Can you add a type alias named `SelectorCombinator` with `#[deprecated = "Use CombinatorValue instead"]`?

kdy1 (commented):
Personally, I don't want to break existing css transforms.

css/ast/src/selector.rs
alexander-akait: Because combinator is not a part of compound selector, it is part of complex selector, compound selector can be written with/without combinator, in theory in future it will be possible to write:
```
.class {
  > .foo {
      color: red;
  }
}

.class > {
  .foo {
      color: red;
  }
}
```

note sass and less already support it, because they parse it as part of `complex` selector.

Look at grammar https://www.w3.org/TR/selectors-4/#grammar:
```
<complex-selector> = <compound-selector> [ <combinator>? <compound-selector> ]*
<compound-selector> = [ <type-selector>? <subclass-selector>*
                        [ <pseudo-element-selector> <pseudo-class-selector>* ]* ]!
```

as you can see complex selector has children:
- `compound-selector`
- `combinator` (optional)

css/ast/src/selector.rs
alexander-akait: I can do it, but if we will do major bump we can avoid to keep legacy stuff in code, I think it is normal to avoid keeping old stuff until we reached v1

alexander-akait:
Honestly, I think it will even improve transformation and not only, for example if we need to write linter and write a rule to disable some combinator we just visit all `Combinator` AST nodes and emit warning using `span`. For minification we remove spaces around combinator.

css/ast/src/selector.rs
kdy1: Then this is fine, thanks!

css/ast/src/selector.rs
kdy1: All crates exposing AST as a part of their API will never hit `v1`, so I think it's better to have one.

alexander-akait: @kdy1 I'm just worried that this way we may have a lot of garbage/legacy stuff in the future and as long as we are at an early stage, we can avoid this because our crates do not use many developers

kdy1 (approved):
Thanks!

PR #2667, opened 2021-11-07
fix(es/async_to_generator): Do not bind `this` to generator for prototype

kwonoj:
Attempt to close
- Closes #2402 
- Closes #2305 

This PR attempts to amend behavior around generator for the async fn to avoid bind to `this` context if given async fn is assignment to prototype. There was some other approach suggestion like https://github.com/swc-project/swc/issues/2305#issuecomment-927588761 , but doesn't seem to work as issue is prototype doesn't have proper context to attach from beginning.

As similar to existing flag `in_object_prop`, this PR introduces new flag `in_prototype_assignment` to indicate if given assign expr is for the prototype.

PR #2668, opened 2021-11-07
refactor: Remove trailing whitespace

kwonoj:
from discussion https://github.com/swc-project/swc/pull/2667#pullrequestreview-799506549

This PR removes trailing whitespace from codes in the repo. Note PR only touches src codes (*.rs). Some of fixture test will fail if fixture strips whitespace as it'll change spans.


kdy1 (approved):
Thanks!

PR #2672, opened 2021-11-07
fix(css/parse): pseudo element and pseudo class spaces

alexander-akait:
Fix, whitespace is not allowed between

> Between the ':'s, or between the ':' and \<ident-token\> or \<function-token\>, of a \<pseudo-element-selector\> or a \<pseudo-class-selector\>.

From https://www.w3.org/TR/selectors-4/#white-space


alexander-akait:
In future we will add recovery mode for selectors and all examples will be valid, but just output tokens, now we don't have it, but we will ad it late when we finish selector parser

kdy1 (approved):
LGTM, thanks!

PR #2673, opened 2021-11-07
refactor(css/ast): remove `UniversalSelector`

alexander-akait:
Remove `UniversalSelector`, we don't use it in code (maybe we forgot to remove it when we do refactor), also `UniversalSelector` is subtype of `TypeSelector`.


alexander-akait:
Or we can bring it back because we have https://www.w3.org/TR/selectors-4/#the-universal-selector and fix the logic of code, I don't have strong option

kdy1 (approved):
Thanks!

PR #2675, opened 2021-11-07
Refactor pseudo class and pseudo element

alexander-akait:
Added `PseudoClassSelector` amd `PseudoElementSelector`.

Based on grammar https://www.w3.org/TR/selectors-4/#grammar:

- pseudo element and pseudo classes is different types so let's us different AST nodes
- separating them in the future will allow for more efficient tranformations, for example 

> The four Level 2 pseudo-elements (::before, ::after, ::first-line, and ::first-letter) may, for legacy reasons, be represented using the <pseudo-class-selector> grammar, with only a single ":" character at their start.

For minimizing purpose we can covert `PseudoElemenetSelector` to `PseudoClassSelector`, we don't need visit all `PseudoSelector` nodes, because it will slow down.

Also `is:()` and etc selector can be converted to simple and versa vice:
```css
/* Selects any paragraph inside a header, main
   or footer element that is being hovered */
:is(header, main, footer) p:hover {
  color: red;
  cursor: pointer;
}

/* The above is equivalent to the following */
header p:hover,
main p:hover,
footer p:hover {
  color: red;
  cursor: pointer;
}
```

With some limitations, but we don't need to handle all `pseudo` selectors, because it will be slow too.


alexander-akait:
@kdy1 Can you review? Should I rebase?

kdy1:
I'll review code changes, but rebase is required to be merged I think?

alexander-akait:
Yep, In my TODO, want to fix problems with `an+b` syntax in our pseudo classes today too

kdy1 (approved):
LGTM, thanks!

kdy1:
(I looked at only code changes)

alexander-akait:
@kdy1 Ready for review, should be merged after https://github.com/swc-project/swc/pull/2759

alexander-akait:
@kdy1 Can you review? I have many good features :star: for selector parser

alexander-akait:
@kdy1 Can you review this, thanks

crates/swc_css_parser/tests/fixture.rs
```diff
@@ -13,7 +13,7 @@ use testing::NormalizedOutput;
 struct AssertValid;
 
 impl Visit for AssertValid {
-    fn visit_pseudo_selector(&mut self, s: &PseudoSelector, _: &dyn Node) {
+    fn visit_pseudo_class_selector(&mut self, s: &PseudoClassSelector, _: &dyn Node) {
```
kdy1, line 16: I think we should also implement `visit_pseudo_element_selector` too.

kdy1 (changes requested):
Mostyle fine, thanks!

kdy1 (approved):
Thanks!

PR #2676, opened 2021-11-07
refactor(es/transforms/bugfix): visit_mut for edge_default_param

kwonoj:
Context: https://discordapp.com/channels/889779439272075314/889785438938726441/906736454737219614

This PR attempts to replace existing fold in `edge_default_param` bugfix to visitor. The only reason I choose this module is this looks one of the smallest path I could try to migrate. 

Change's mostly straightforward, except couple of changes: 1. uses `as_folder` to plug migrated visitor to existing logics 2. instead of mapping `props` to reassign, directly replace elements in the vec.


kdy1 (approved):
LGTM. Thanks!

Issue #2677, opened 2021-11-07
default exported async fn does not transpiled into generator


kwonoj:
### Describe the bug

If async function is exported as default, it is not being transpiled to generator though configuration asks to transpile async / await.

### Input code

```typescript
async function region() {
}

export default async function someCall() {
  await region();
}
```

### Config

```json
{
  "jsc": {
    "target": "es2017"
  },
  "module": {
    "type": "commonjs",
    "noInterop": false
  },
  "env": {
    "targets": {
      "node": 14
    },
    "include": ["transform-async-to-generator"]
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.106&code=H4sIAAAAAAAAA0ssrsxLVkgrzUsuyczPUyhKTQdSGpoK1bxcvFy1ICK1oiC%2FqEQhJTUtsTSnRCERVUNxfm6qc2JODkSLgkJieWJmCdwYa6AZAHl8xMpiAAAA&config=H4sIAAAAAAAAA0WNSwrDMAxE76J1Fm2hXeQOPYRxleDgHxoFGozvXju4ZCfNPD0V2mBpLpSNgKVPOKKaL82kR2ZYcVlpIkWLFuPBtS1GVtaGMB63%2B7PVPiXwACYKLrrl6DKbQhYGrsrE1f%2FJ2lwhffYelPPf6XxRvRzjzuE9QJWd6w%2ByirtzuAAAAA%3D%3D

### Expected behavior

Transpile all of the async into generator

### Version

1.2.106

PR #2681, opened 2021-11-08
fix(es/transforms/compat): apply regenerator transform for default expr

kwonoj:
- attempt to close https://github.com/swc-project/swc/issues/2677

PR updates transform to check defaultexportdecl and fold if there's fn expression to transpile default exported async fn.

One thing bit unsure is unlike other fn's `FnDecl`, if it's `ExportDefaultDecl` fn is set to `FnExpr` instead. PR doesn't touches those node but only touches transform instead for now.


kwonoj:
Thinking this more, I'm curious if there's specific reason it has to be FnExpr for the default exported fn: it looks like babel treats it as declaration regardless of async or not for the default export: https://astexplorer.net/#/gist/4e88e71b8c3f7e7094dc5a2bc5085a36/666b57be5110383e94f552dbd2e0f976b003251e

kdy1:
>Thinking this more, I'm curious if there's specific reason it has to be FnExpr for the default exported fn: it looks like babel treats it as declaration regardless of async or not for the default export: >https://astexplorer.net/#/gist/4e88e71b8c3f7e7094dc5a2bc5085a36/666b57be5110383e94f552dbd2e0f976b003251e


Identifier of default function declaration is optional, while identifier of `FnDecl` is not optional.

kwonoj:
Got it. thoughts on current approach?

ecmascript/transforms/compat/src/es2017/async_to_generator.rs
```diff
@@ -362,6 +362,36 @@ impl Fold for Actual {
         }
     }
 
+    fn fold_module_item(&mut self, item: ModuleItem) -> ModuleItem {
+        let item = match item {
+            // if fn is ExportDefaultDecl, fn is not FnDecl but FnExpr
+            ModuleItem::ModuleDecl(ModuleDecl::ExportDefaultDecl(export_default)) => {
+                if let DefaultDecl::Fn(fn_expr) = export_default.decl {
+                    let expr = fn_expr.fold_children_with(self);
```
kdy1, line 370: Isn't this double-fold?

ecmascript/transforms/compat/src/es2017/async_to_generator.rs
```diff
@@ -362,6 +362,36 @@ impl Fold for Actual {
...
+                        ..export_default
+                    }))
+                } else {
+                    ModuleItem::ModuleDecl(ModuleDecl::ExportDefaultDecl(export_default))
+                }
+            }
+            _ => item,
+        };
+
+        item.fold_children_with(self)
```
kdy1, line 392: This will fold the same node multiple times.

kdy1 (changes requested):
Thanks!
I found some small performance issues.

ecmascript/transforms/compat/src/es2017/async_to_generator.rs
kwonoj: I may miss something, but this seems required? otherwise moduleitem skips folds ends up not transpiling bunch of async / await with test failures.

ecmascript/transforms/compat/src/es2017/async_to_generator.rs
kwonoj: Thanks for the catch, updated.

ecmascript/transforms/compat/src/es2017/async_to_generator.rs
kdy1: I mean you can add `return` to L380 to avoid duplicated folding.

kdy1 (changes requested):

ecmascript/transforms/compat/src/es2017/async_to_generator.rs
kwonoj: Oh :facepalm: yes, that seems do the trick.

ecmascript/transforms/compat/src/es2017/async_to_generator.rs
```diff
@@ -362,6 +362,37 @@ impl Fold for Actual {
...
+                    };
+
+                    return ModuleItem::ModuleDecl(ModuleDecl::ExportDefaultDecl(
+                        ExportDefaultDecl {
+                            decl: expr.into(),
+                            ..export_default
+                        },
+                    ));
+                } else {
+                    return ModuleItem::ModuleDecl(ModuleDecl::ExportDefaultDecl(export_default));
```
kdy1, line 387: I think `return` should be removed as `async`s in `export default class Foo {}` will not be folded otherwise.

kdy1 (changes requested):

kdy1 (approved):
Thank you!
LGTM.

ecmascript/transforms/compat/src/es2017/async_to_generator.rs
kwonoj: Uh, makes sense. Removing this makes match arm return type differences - updated code to fold for the remaining cases other than defaulyt expr.

PR #2686, opened 2021-11-08
test(css/parser): Add a test for `BOM`

alexander-akait:
Just `BOM` test to avoid future regressions

tomorrow I will try to finish our lexer and it will be stable :star:


kdy1 (approved):
Thanks!

PR #2693, opened 2021-11-09
Update README.md

tsriram:
TIL that swc stands for `speedy web compiler` ([reference](https://twitter.com/kdy1dev/status/1457552634244579330)). Thought it'd be good to add it to README :)


kdy1 (approved):
Thanks!

PR #2700, opened 2021-11-09
fix(css/parser): Fix parsing of escaped characters

alexander-akait:
@kdy1 Bug fix, it is very interesting bug, look at code:
```css
#\2d a {
  color: red;
}
```

html:
```
<div id="-a">Text</div>
```

color will be red, because `whitespace` between `\2d` and `a` is part of `\2d` due https://www.w3.org/TR/css-syntax-3/#consume-escaped-code-point

> If the next input code point is whitespace, consume it as well.

Now we interpreted it as two selectors and it is bug


css/parser/tests/fixture/esbuild/misc/-8o_H6sq86TDAHqF7YO0hg/span.rust-debug
```diff
@@ -20,31 +20,31 @@ error: SelectorList
  --> $DIR/tests/fixture/esbuild/misc/-8o_H6sq86TDAHqF7YO0hg/input.css:1:1
   |
 1 | \2d {}
-  | ^^^
+  | ^^^^
```
alexander-akait, line 23: Same for here, ` ` (whitespace) is part of escaped characters, it is not simple whitespace

alexander-akait:
@kdy1 Here is just fix, no breaking changes

kdy1 (approved):
Thank you!
CSS spec is stranger then ES spec, IMO.

PR #2702, opened 2021-11-09
Refactor lexer part 1

alexander-akait:
Part `1` of stable lexer. 

Now we have broken logic for `comments` and `whitespaces`, look at example
```css
a /* test */ b { }
```

Now our lexer output only one `whitespace`, because we skip `whitespaces` between `comments`, it is wrong, this PR refactor some places so we will fix it in future
Also previously we allow comments inside `functions`, i.e:
```css
a {
  background: url(/*test*/ "image.png");
}
```

It is wrong, now we output error on this (in future it will be covered by recovery mode).

Also we have a lot of bugs with comments, look at `span`, they are wrong, but I have idea how we fix it.


css/parser/src/lexer/mod.rs
```diff
@@ -138,13 +118,6 @@ where
                     }
                 }
 
-                if self.config.allow_wrong_line_comments {
-                    if self.input.is_byte(b'/') && self.input.peek() == Some('/') {
-                        self.skip_line_comment()?;
-                        self.start_pos = self.input.cur_pos();
-                    }
-                }
```
alexander-akait, line 146: We don't need to check on this in every place now, because comments eat in loop

kdy1:
I'll review css-related issues after upgrading css transforms in `styled-jsx`.
It's too confusing at the moment, sorrty.

alexander-akait:
@kdy1 Thas is bad :disappointed: I have a lot of improvements :disappointed: You can assign me to issues and ping me and I will fix them

kdy1:
Oh, thank you!

Can you take a look at https://github.com/kdy1/next.js/tree/c5abb419bc6f384dc308b9d79222465dc9d1fff4/packages/next/build/swc/src/styled_jsx ?
I tried to represent the same logic with `ComplexSelectorChildren`, but it resulted in some differences.

alexander-akait:
Yep, will look at this in near future (today)

alexander-akait:
@kdy1 Maybe you can provide instructions how to fast debug it?

alexander-akait:
@kdy1 Here is also no breaking changes, only fixes and refactor, part `2` contains breaking changes due new `EOF` token

css/parser/src/parser/style_rule.rs
```diff
@@ -16,15 +16,20 @@ where
...
             // TODO: remove `}`
-            if self.input.is_eof()? || is!(self, "}") {
+            if is!(self, "}") {
                 return Ok(rules);
             }
 
-            match cur!(self) {
-                tok!(" ") => {
-                    self.input.skip_ws()?;
+            let cur = self.input.cur()?;
```
kdy1, line 23: Code seems identical, is there a reason for this change?

css/parser/src/lexer/mod.rs
```diff
@@ -430,6 +408,83 @@ where
...
+    // This section describes how to consume comments from a stream of code points.
+    // It returns nothing.
+    fn read_comments(&mut self) -> LexResult<()> {
+        // If the next two input code point are U+002F SOLIDUS (/) followed by a U+002A
+        // ASTERISK (*), consume them and all following code points up to and including
+        // the first U+002A ASTERISK (*) followed by a U+002F SOLIDUS (/), or up to an
+        // EOF code point. Return to the start of this step.
+        // NOTE: We allow to parse line comments under the option.
+        let is_block_comment = self.input.cur() == Some('/') && self.input.peek() == Some('*');
+        let is_inline_comment = self.config.allow_wrong_line_comments
```
kdy1, line 421: Is this variables required?
I think `rustc` may optimize calls to `self.input.cur()` if `is_block_comment` is true, but I think checking these at each if statement will be better.

kdy1 (changes requested):
I have some tiny questions.

css/parser/src/parser/style_rule.rs
alexander-akait: To be honestly it is the same, we need refactor logic around spaces, because now we lose them between `comments`, but let's postpone it for other PRs, I will return original

css/parser/src/lexer/mod.rs
alexander-akait: Agree, fixed

alexander-akait:
@kdy1 fixed

css/parser/src/parser/style_rule.rs
kdy1: Did you push your changess?

alexander-akait: Yep

kdy1: Github diff view still shows changes.
Hmm...

alexander-akait: Oh, we need it due we should handle `None` in `match`, css parser and lexer in spec has two definitions:
1. consume (cur + bump)
2. reconsume (reset state)

But in most of places we do not respect it, because of this, we often need to pass arguments or store intermediate (i.e. span/cur token/etc) states somewhere else, it reduces performance and requires us have more code then we need. Ideally all in lexer and parser should use `loop + cur + bump + match` and `reset` if we need reconsume, here the first improvement, after lexer refactor it will be easy to convert it everywhere.

But I can avoid it here and do it late, I think it will be even better

kdy1 (approved):
Thanks!

(Btw, vscode github extension is so useful. I could look at all test diffs)

alexander-akait:
@kdy1 Can you provide steps how I can help you with https://github.com/kdy1/next.js/tree/c5abb419bc6f384dc308b9d79222465dc9d1fff4/packages/next/build/swc/src/styled_jsx, because we need continue our css lexer/parser

kdy1:
@alexander-akait I want to make https://github.com/kdy1/next.js/blob/c5abb419bc6f384dc308b9d79222465dc9d1fff4/packages/next/build/swc/src/styled_jsx/transform_css.rs#L156-L249 to work identically as it of canary.

Thank you and I will continue reviewing/merging css lexer/parser regardless of it.

alexander-akait:
@kdy1 I think you don't need handle `complex_selector`, you want to process only `.class` and `:pseudo` selectors (based on code), so I think better to handle only `CompoundSelector` (i.e. `visit_mut_compound_selector`)

PR #2704, opened 2021-11-10
Eof token

alexander-akait:
Refactor lexer part `2`

### DIFF
--- a/css/ast/src/token.rs
+++ b/css/ast/src/token.rs
@@ -18,7 +18,7 @@ pub struct TokenAndSpan {
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub enum Token {
     EOF,
-    
+
     Ident {
         value: JsWord,
         raw: JsWord,
--- a/css/ast/src/token.rs
+++ b/css/ast/src/token.rs
@@ -17,6 +17,8 @@ pub struct TokenAndSpan {
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub enum Token {
+    EOF,
+    
     Ident {
         value: JsWord,
         raw: JsWord,
--- a/css/parser/tests/fixture.rs
+++ b/css/parser/tests/fixture.rs
@@ -2,7 +2,6 @@ use std::path::PathBuf;
 use swc_common::{errors::Handler, input::SourceFileInput, Span, Spanned, DUMMY_SP};
 use swc_css_ast::*;
 use swc_css_parser::{
-    error::ErrorKind,
     lexer::Lexer,
     parse_tokens,
     parser::{input::ParserInput, Parser, ParserConfig},
@@ -26,9 +25,15 @@ impl Visit for AssertValid {
         }
 
         let mut errors = vec![];
+        let mut tokens = s.clone().args;
+
+        tokens.tokens.push(TokenAndSpan {
+            span: DUMMY_SP,
+            token: Token::EOF,
+        });
 
         let _selectors: SelectorList = parse_tokens(
-            &s.args,
+            &tokens,
             ParserConfig {
                 parse_values: true,
 
@@ -57,7 +62,13 @@ fn tokens_input(input: PathBuf) {
             let mut tokens = vec![];
 
             while let Ok(t) = lexer.next() {
+                let is_eof = t.token == Token::EOF;
+
                 tokens.push(t);
+
+                if is_eof {
+                    break;
+                }
             }
             Tokens {
                 span: Span::new(fm.start_pos, fm.end_pos, Default::default()),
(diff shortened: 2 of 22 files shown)


alexander-akait:
@kdy1 Can you look? Here is the breaking change, now tokens should be ended by `EOF` token, it is in spec and allow to improve recovery mode in some places

css/ast/src/token.rs
```diff
@@ -17,6 +17,8 @@ pub struct TokenAndSpan {
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub enum Token {
+    EOF,
```
kdy1, line 20: Can other token come after `EOF` token?
I'm not sure how can it help error recovery.

kdy1 (commented):
If it can only be the last, `None` is enough IMO.

css/ast/src/token.rs
alexander-akait: @kdy1 

> Can other token come after EOF token?

No, EOF is last

> I'm not sure how can it help error recovery.

There is an interesting situation here - I saw chrome CSS lexer and Servo CSS lexer (other implementations too), they all use it because it is clearly described in the specification (even `esbuild` uses it). 

One more problem - we use `ErrorKind::Eof` to indicate end of file, but actually there is no error here, we have to rely on token flow. This makes our code cleaner according to the spec.

In fact - yes, this does not fundamentally change nothing, but to be honest, I would like to be closer to the specification and not add anything new to it.

Here `esbuild` example of `EOF` token - https://github.com/evanw/esbuild/blob/master/internal/css_lexer/css_lexer.go#L257, if you need I can provide more links on sources

kdy1: Interesting. I don't like adding `Invalid` state to ast, (another token after `EOF` in this case), but I think there should be enough reason, so I'll merge it. Thanks!

kdy1 (dismissed):

kdy1:
Seems like tests are failing due to some match expressions. @alexander-akait Can you fix itt?

alexander-akait:
@kdy1 Done

css/parser/tests/fixture.rs
```diff
@@ -26,9 +25,15 @@ impl Visit for AssertValid {
         }
 
         let mut errors = vec![];
+        let mut tokens = s.clone().args;
+
+        tokens.tokens.push(TokenAndSpan {
+            span: DUMMY_SP,
+            token: Token::EOF,
+        });
```
kdy1, line 34: This seems like a severe regression of API.

kdy1 (commented):
I missed changes of `fixture.rs` in the previous review.

It's too bad regression IMHO.

css/parser/tests/fixture.rs
alexander-akait: Ideally we should never faced with it in real code, `parse_tokens` should always ending with `EOF` token, this change only because we can't parse `:pseudo-selector(content)` due broken selector parser, I think we will fix it in near future, now it is just workaround for our tests.

alexander-akait: Shorty - just put `EOF` token at the end if you need to parse using tokens, nothing more

alexander-akait: @kdy1 Friendly ping, any thoughts?

kdy1: I'm still against this PR, as it adds lots of `invalid state`s.

alexander-akait: @kdy1 But it helps to error reporting and recovery mode, for example if we have not closed string we will report about unexpected EOF token, but now we just report error (like expected value, but it is wrong, value exist, but EOF token unexpected here), without showing why (you can see updated snapshots here).

To be honestly when we handle tokens we have two situations:
1. broken tokens input (no EOF token at the end), if you handle tokens in stream (it very important) we can't rely on end of array (i.e. `None`), we can only understand end of file by presence EOF token
2. broken tokens input (wrong token order or unexpected token after other token), in this case we use error recovery and work as browser (i.e. skip invalid properties/values/declarations/rules/etc)

That is why almost everyone lexer has `EOF` token, many of them are written with the possibility with stream reading and just full file consumption. 

I think that in the future we may well have stream API, it can potentially increase performance in many places. Sometimes we do not need to read the entire file at once, but consumption occurs chunks.

I understand that this may look a little strange and look like `invalid state`s. But I can provide links/code of many lexers and you will find it everywhere, because as I said earlier, lexers are created no matter how the file is read in stream or buffer of the whole file or something other. 

Prof:
1. babel - https://github.com/babel/babel/blob/main/packages/babel-parser/src/tokenizer/state.js#L108
2. esbuild - https://github.com/evanw/esbuild/blob/master/internal/css_lexer/css_lexer.go#L257
3. css-tree - https://github.com/csstree/csstree/blob/a8e4f1b071a3bc2117b83eae8c3cc2aebc176405/lib/tokenizer/types.js#L3 

I can provide more if you need

alexander-akait:
Let's close, maybe we revisit it late, now `None` is `EOF`, it still can be useful if we will need stoppable parsers, but currently we don't need it, anyway CSS and HTML parsers design to be them

Issue #2549, opened 2021-10-27
Object.defineProperty does not allow reassign ref to exported fn


kwonoj:
**Describe the bug**

I created small snippet https://github.com/kwonoj/defineproperty-test to explain this bit better.

When there is a exported fn and an exported object contains ref to its fn, replacing ref in object via defineProperty does not replaces exports.

**Input code**

```
// component.js 

export function warn() {
  throw new Error('this should not be called');
}

export const test = {};

// set object test have ref to fn `warn`, also sets setter / getter
Object.defineProperty(test, 'warn', {
	configurable: true,
	get: () => warn,
	set: (v) => {
		warn = v;
	},
});
```

```
// mock_component.js

import { test } from './component';

function warnOverride() { console.log('override') }

// replace test.warn into `warnOverride`
try {
  test['warn'] = warnOverride;
} catch (e) {
}

if (test['warn'] !== warnOverride) {
  Object.defineProperty(test, 'warn', { value: warnOverride });
}
```

```
//test.js

import './mock_component';

import { warn } from './component';

warn(); // this throws 'this should not be called'

```

**Config**

```json
{
  "jsc": {
    "target": "es2017",
    "parser": {
      "syntax": "typescript"
    },
    "externalHelpers": false
  },
  "module": {
    "type": "commonjs",
    "noInterop": false
  },
  "env": {
    "targets": {
      "node": 14
    },
    "include": ["transform-async-to-generator"]
  }
}
```

**Expected behavior**
Hope it calls `warnOverride`

**Version**
The version of @swc/core: 1.2.103.

**Additional context**

Though I said `expected behavior`, I feel this is bit controversial since reassigning fn is not allowed anyway. The reason I raised issue is at least to discuss if this is expected behavior on the swc's side.

If we peek other transpiler's behavior, this is output from babel / tsc both:

// babel
```
exports.warn = warn;

function warn() {
  throw new Error('this should not be called');
}

var test = {};
exports.test = test;
Object.defineProperty(test, 'warn', {
  configurable: true,
  get: function get() {
    return warn;
  },
  set: function set(v) {
    exports.warn = warn = v;
  }
});
```

//tsc
```
function warn() {
    throw new Error('this should not be called');
}
exports.warn = warn;
exports.test = {};
Object.defineProperty(exports.test, 'warn', {
    configurable: true,
    get: () => warn,
    set: (v) => {
        exports.warn = warn = v;
    },
});
```

[...]

PR #2705, opened 2021-11-10
fix(es/transforms/cjs): bind reassign scope to default export fn

### LINKED ISSUE #2678: does not allow reassign ref to default exported fn
### Describe the bug

- Followup edge cases for https://github.com/swc-project/swc/issues/2549, https://github.com/swc-project/swc/issues/2606

When an function is async, default exported it does not reassign its ref to export as similar to original #2549 does.

### Input code

```typescript
export default function someCall() {
  region();
}

export const text = {};
Object.defineProperty(test, 'someCall', {
  set: (v) => {
    someCall = v;
  }
})
```

### Config

```json
{
  "jsc": {
    "target": "es2017"
  },
  "module": {
    "type": "commonjs",
    "noInterop": false
  },
  "env": {
    "targets": {
      "node": 14
    },
    "include": ["transform-async-to-generator"]
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.106&code=H4sIAAAAAAAAAz2NQQrCMBBF94Hc4e%2BaQPEAlrrxAHqFGqdSiUlJpqVScndHgm7f579H2xwT407jsHjGuATHUwzI8UXnwXtjsWsFJHoINrbTqmilFdWfiyEzmDZGj73Ierk9yfFBhFOga4ozJX4bpswtmp%2B1aas1Ex9hVov%2BVAH%2BYfGt3RdJrtgP6sRtQKcAAAA%3D&config=H4sIAAAAAAAAA0WNSwrDMAxE76J1Fm2hXeQOPYRxleDgHxoFGozvXju4ZCfNPD0V2mBpLpSNgKVPOKKaL82kR2ZYcVlpIkWLFuPBtS1GVtaGMB63%2B7PVPiXwACYKLrrl6DKbQhYGrsrE1f%2FJ2lwhffYelPPf6XxRvRzjzuE9QJWd6w%2ByirtzuAAAAA%3D%3D

### Expected behavior

Same as #2549, #2606, expect to get reassign like 

```
  set: (v) => {
	exports.someCall = someCall = v;
  }

```

### Version

1.2.106 (and potentially master TOT)

### Additional context

- Looks like potentially related to https://github.com/swc-project/swc/issues/2677 , but for now separating issue until we can triage if it shares same root cause.

kwonoj:
- closes #2678 

This PR attempt to fix #2678, one of the edge cases of #2549.

As same as fix for #2549, this PR sets scope to the exported. The only difference is it doesn't set scope to name of the declaration (or FnExpr's) as default export uses fixed name `default` for the export. Given current fold already creates default export member expr, reuses created one to assign into scope.

### DIFF
--- a/ecmascript/transforms/module/src/common_js.rs
+++ b/ecmascript/transforms/module/src/common_js.rs
@@ -372,6 +372,28 @@ where
                                     // init_export!("default");
                                     let ident = ident.unwrap_or_else(|| private_ident!("_default"));
 
+                                    let member_expr = member_expr!(DUMMY_SP, exports.default);
+                                    match &*member_expr {
+                                        Expr::Member(MemberExpr { prop, .. }) => match &**prop {
+                                            Expr::Ident(default_ident) => {
+                                                let mut scope = self.scope.borrow_mut();
+
+                                                scope
+                                                    .exported_bindings
+                                                    .entry((ident.sym.clone(), ident.span.ctxt()))
+                                                    .or_default()
+                                                    .push((
+                                                        default_ident.sym.clone(),
+                                                        default_ident.span.ctxt(),
+                                                    ));
+
+                                                drop(scope);
+                                            }
+                                            _ => (),
+                                        },
+                                        _ => (),
+                                    }
+
                                     extra_stmts.push(ModuleItem::Stmt(Stmt::Decl(Decl::Fn(
                                         FnDecl {
                                             ident: ident.clone(),
@@ -384,10 +406,7 @@ where
                                     stmts.push(
                                         AssignExpr {
                                             span: DUMMY_SP,
-                                            left: PatOrExpr::Expr(member_expr!(
-                                                DUMMY_SP,
-                                                exports.default
-                                            )),
+                                            left: PatOrExpr::Expr(member_expr),
                                             op: op!("="),
                                             right: Box::new(ident.into()),
                                         }
--- a/ecmascript/transforms/module/src/common_js.rs
+++ b/ecmascript/transforms/module/src/common_js.rs
@@ -372,26 +372,19 @@ where
                                     // init_export!("default");
                                     let ident = ident.unwrap_or_else(|| private_ident!("_default"));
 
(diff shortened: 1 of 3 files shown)


ecmascript/transforms/module/src/common_js.rs
```diff
@@ -372,6 +372,28 @@ where
                                     // init_export!("default");
                                     let ident = ident.unwrap_or_else(|| private_ident!("_default"));
 
+                                    let member_expr = member_expr!(DUMMY_SP, exports.default);
+                                    match &*member_expr {
```
kdy1, line 376: As value of `member_expr` is known, I think this can be improved for better readability.

ecmascript/transforms/module/src/common_js.rs
```diff
@@ -372,6 +372,28 @@ where
...
+                                                        default_ident.span.ctxt(),
+                                                    ));
+
+                                                drop(scope);
+                                            }
+                                            _ => (),
+                                        },
+                                        _ => (),
+                                    }
+
```
kdy1, line 396:
```rust
{
        let mut scope = self.scope.borrow_mut();

        scope
            .exported_bindings
            .entry((ident.sym.clone(), ident.span.ctxt()))
            .or_default()
            .push((js_word!("default"), Default::default()));
    }
}
```

will work.

Note that it's wrapped in `{` and `}` to make `scope` dropped.

kdy1 (changes requested):
Thanks!

ecmascript/transforms/module/src/common_js.rs
kwonoj: Good point, thanks. maybe this works? https://github.com/swc-project/swc/pull/2705/files#diff-506530d02689c3c49a06cbec6fb8c9ffd7b1df4c551dcac29d8bd08d0248590fR380

kdy1 (approved):
Thanks!

ecmascript/transforms/module/src/common_js.rs
kdy1: Yes, thanks!

PR #2696, opened 2021-11-09
refactor(es/transforms/compat): visitmut for async_to_generator part1

kwonoj:
This PR partially migrates `async_to_generator`'s fold into visit_mut. For easier migration there are some naive choices to copy few things. This may need to be revisited depends on the nature of the copy.

PR #2709, opened 2021-11-10
refactor(es/transforms/compat): visitmut for async_to_generator part2

kwonoj:
Continuation from https://github.com/swc-project/swc/pull/2696

This is second part of visit_mut refactoring for async_to_generator. Due to each fold depends on others, it was not able to do gradual migartion but had to replace all at once.

One notable changes in this PR is this change actually changed output bit, between order of generated fn. Since visit_mut doesn't return whole stmts vec now it appends extra_stmts each time it's being created which ends up emission order changes. Peeking babel's behavior, actually this aligns with babel's output and afaik this shouldn't cause runtime differences.


kdy1:
Can you rebase this?
I flattened the `ecmascript` directory because I think it's better to do so as the project got much bigger.

kwonoj:
Sure, rebased to latest master.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -44,61 +44,58 @@ struct Actual {
...
-            let stmt = stmt.fold_with(&mut actual);
 
-            buf.extend(actual.extra_stmts.into_iter().map(T::from_stmt));
-            buf.push(stmt);
-        }
+            let stmt = &mut (stmts[idx]);
+            stmt.visit_mut_with(&mut actual);
 
-        buf
+            stmts.extend(actual.extra_stmts.into_iter().map(T::from_stmt));
```
kdy1, line 80: I think this will append `extra_stmts` to the `stmts`, while previous behavior will prepend each injected statement prior to stmt.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -44,61 +44,58 @@ struct Actual {
...
+        Vec<T>: VisitMutWith<Self>,
     {
-        let stmts = stmts.fold_children_with(self);
+        stmts.visit_mut_children_with(self);
 
-        let mut buf = Vec::with_capacity(stmts.len());
+        let stmts_len = stmts.len();
 
-        for stmt in stmts {
+        for idx in 0..stmts_len {
```
kdy1, line 70: You can use `stmts.take()` instead.

As we have to re-create a vector anyway, `take()` is fine.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -172,17 +169,15 @@ impl Fold for Actual {
                 type_params: Default::default(),
                 return_type: Default::default(),
             },
-            ..m
+            key: m.key.take(),
```
kdy1, line 172: Is this required?
If `rustc` complains about it, it would be better to assign to `m.function` instead as this is useless `memmove`.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -172,17 +169,15 @@ impl Fold for Actual {
...
-    fn fold_expr(&mut self, expr: Expr) -> Expr {
+    fn visit_mut_expr(&mut self, expr: &mut Expr) {
         match expr {
-            Expr::Paren(ParenExpr { span, expr }) => {
-                return Expr::Paren(ParenExpr {
-                    span,
-                    expr: expr.fold_with(self),
-                });
+            Expr::Paren(ParenExpr { expr, .. }) => {
+                expr.visit_mut_with(self);
```
kdy1, line 180: Can you add `return` after this?

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -192,27 +187,32 @@ impl Fold for Actual {
                 args,
                 type_args,
             }) if callee.is_fn_expr() => {
-                return self.handle_iife(span, (*callee).fn_expr().unwrap(), args, type_args);
+                *expr = self.handle_iife(
```
kdy1, line 190: Same here, can you add `return` after this?

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -192,27 +187,32 @@ impl Fold for Actual {
...
+                    span.take(),
+                    &mut (*callee).take().fn_expr().unwrap(),
```
kdy1, line 192: `expect_fn_expr()` would be better than `fn_expr().unwrap()`.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -192,27 +187,32 @@ impl Fold for Actual {
...
                     Expr::Fn(..) => true,
                     _ => false,
                 },
                 _ => false,
             } =>
             {
-                return self.handle_iife(
-                    span,
-                    (*callee).paren().unwrap().expr.fn_expr().unwrap(),
+                *expr = self.handle_iife(
```
kdy1, line 211: Can you add `return` after this?

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -192,27 +187,32 @@ impl Fold for Actual {
...
+                    span.take(),
+                    &mut (*callee).take().paren().unwrap().expr.fn_expr().unwrap(),
```
kdy1, line 213: Same here. `.expect_paren().expr.expect_fn_expr()` would be more readable.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -192,27 +187,32 @@ impl Fold for Actual {
...
                     _ => false,
                 },
                 _ => false,
             } =>
             {
-                return self.handle_iife(
-                    span,
-                    (*callee).paren().unwrap().expr.fn_expr().unwrap(),
+                *expr = self.handle_iife(
+                    span.take(),
```
kdy1, line 212: `span` is `Copy`, so *span would be better.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -246,14 +246,14 @@ impl Fold for Actual {
                     _ => {}
                 }
 
-                let assign_expr = assign_expr.fold_children_with(self);
+                assign_expr.visit_mut_children_with(self);
                 self.in_prototype_assignment = false;
-                return Expr::Assign(assign_expr);
+                *expr = Expr::Assign(assign_expr.take());
```
kdy1, line 251: Can you add `return` after this?

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -266,24 +266,24 @@ impl Fold for Actual {
...
                 // Apply arrow
-                let used_this = contains_this_expr(&body);
+                let used_this = contains_this_expr(body);
 
                 let fn_expr = FnExpr {
                     ident: None,
                     function: Function {
                         decorators: vec![],
-                        span,
+                        span: span.take(),
```
kdy1, line 275: `*span` would be better

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -342,76 +345,76 @@ impl Fold for Actual {
...
+            return;
         }
 
-        let function = self.fold_fn(Some(f.ident.clone()), f.function, true);
-        FnDecl {
-            ident: f.ident,
-            function,
+        self.visit_fn(Some(f.ident.clone()), &mut f.function, true);
+        *f = FnDecl {
+            ident: f.ident.take(),
```
kdy1, line 363: Assigning to `f.function` would be better than using `ident.take()`.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -342,76 +345,76 @@ impl Fold for Actual {
...
-
-                    ModuleItem::ModuleDecl(ModuleDecl::ExportDefaultDecl(ExportDefaultDecl {
-                        decl: expr.into(),
-                        ..export_default
-                    }))
-                } else {
-                    let item =
-                        ModuleItem::ModuleDecl(ModuleDecl::ExportDefaultDecl(export_default));
-                    item.fold_children_with(self)
+                    item.visit_mut_children_with(self);
```
kdy1, line 389: Seems like a double visit for `DefaultDecl::Fn`.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -342,76 +345,76 @@ impl Fold for Actual {
...
 
         if !prop.function.is_async {
-            return prop;
+            return;
         }
-        let params = prop.function.params;
 
         let is_this_used = contains_this_expr(&prop.function.body);
 
+        let original_fn_params = prop.function.params.clone();
```
kdy1, line 405: `.take()` would be better than `.clone()`.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -427,9 +430,9 @@ impl Fold for Actual {
             })
         };
 
-        MethodProp {
+        *prop = MethodProp {
```
kdy1, line 433: Same here. Assignment to `prop.function` would be better

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -444,23 +447,18 @@ impl Fold for Actual {
...
+        n.visit_mut_children_with(self);
 
         self.in_object_prop = old;
-
-        n
-    }
-
-    fn fold_stmts(&mut self, n: Vec<Stmt>) -> Vec<Stmt> {
-        n
     }
```
kdy1, line 462: Seems like `fold_stmts` is removed?

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -579,15 +576,20 @@ impl Fold for MethodFolder {
...
                 Expr::Member(MemberExpr {
                     obj: ExprOrSuper::Super(..),
                     ..
                 }) => true,
                 _ => false,
             } =>
             {
-                return self.handle_assign_to_super_prop(span, left.member().unwrap(), op, right);
+                *expr = self.handle_assign_to_super_prop(
+                    span.take(),
```
kdy1, line 588: `*span` would be better

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -579,15 +576,20 @@ impl Fold for MethodFolder {
...
+                    left.take().member().unwrap(),
```
kdy1, line 589: `expect_member()` would be better.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -607,11 +609,11 @@ impl Fold for MethodFolder {
                 _ => false,
             } =>
             {
-                return self.handle_assign_to_super_prop(
-                    span,
-                    left.expr().unwrap().member().unwrap(),
-                    op,
-                    right,
+                *expr = self.handle_assign_to_super_prop(
```
kdy1, line 612: Can you add `return` after this?

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -607,11 +609,11 @@ impl Fold for MethodFolder {
...
+                    span.take(),
```
kdy1, line 613: `*span` would be better.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -607,11 +609,11 @@ impl Fold for MethodFolder {
...
+                    left.take().expr().unwrap().member().unwrap(),
```
kdy1, line 614: `expect_expr().expect_member()` would be better.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -677,12 +679,12 @@ impl Fold for MethodFolder {
                     definite: false,
                 });
 
-                Expr::Call(CallExpr {
-                    span,
+                *expr = Expr::Call(CallExpr {
+                    span: span.take(),
```
kdy1, line 683: Same here

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -703,9 +705,9 @@ impl Fold for MethodFolder {
                         body: BlockStmtOrExpr::Expr(Box::new(
                             MemberExpr {
                                 span: DUMMY_SP,
-                                obj: ExprOrSuper::Super(super_token),
-                                computed,
-                                prop,
+                                obj: ExprOrSuper::Super(super_token.take()),
```
kdy1, line 708: Same here. `.take()` is not required.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -715,14 +717,14 @@ impl Fold for MethodFolder {
                     definite: false,
                 });
 
-                Expr::Call(CallExpr {
-                    span,
+                *expr = Expr::Call(CallExpr {
+                    span: span.take(),
```
kdy1, line 721: Same here

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -731,17 +733,19 @@ impl Actual {
...
-        callee: FnExpr,
-        args: Vec<ExprOrSpread>,
+        callee: &mut FnExpr,
+        args: &mut Vec<ExprOrSpread>,
         type_args: Option<TsTypeParamInstantiation>,
     ) -> Expr {
         if !callee.function.is_async || callee.ident.is_some() {
-            let callee = Expr::Fn(callee).fold_with(self);
-            let args = args.fold_with(self);
+            callee.visit_mut_with(self);
```
kdy1, line 741: I think order should be reversed to be identical with previous code.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -750,31 +754,36 @@ impl Actual {
         if callee.ident.is_some()
             && contains_ident_ref(&callee.function.body, callee.ident.as_ref().unwrap())
         {
-            let callee = Expr::Fn(callee).fold_with(self);
-            let args = args.fold_with(self);
+            callee.visit_mut_with(self);
```
kdy1, line 757: Same here

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -804,7 +813,7 @@ impl Actual {
         let right = make_fn_ref(
             FnExpr {
                 ident: None,
-                function: f,
+                function: f.take(),
```
kdy1, line 816: Same here

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -868,8 +877,9 @@ impl Actual {
                 vec![quote_ident!("arguments").as_arg()],
             ))),
         });
-        Function {
-            span,
+
+        *f = Function {
+            span: span.take(),
```
kdy1, line 882: Same here

kdy1 (changes requested):

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
kwonoj: Yes it does. I wrote in PR for the changes

> One notable changes in this PR is this change actually changed output bit, between order of generated fn. Since visit_mut doesn't return whole stmts vec now it appends extra_stmts each time it's being created which ends up emission order changes. Peeking babel's behavior, actually this aligns with babel's output and afaik this shouldn't cause runtime differences.

and as said above, this actually makes output emission aligned to what babel does, thought it'd acceptable.

kdy1: Thank you for your explanation. Then it's fine.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
kwonoj: yes. doesn't seem to do anything? if it's visit_mut, should it explicitly noop?

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
kwonoj: Vec is actually not recreated: previously there's new buf which appends existing stmt with new extra_stmt. But in here stmts are used as-is, only appends new extra_stmts. `take()` makes existing stmts drain.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
kwonoj: Short curcuiting with return here actually makes some existing test fails, if fn is wrapped iife and it contains async.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
kwonoj: Actually I realized f.function itself is visited with mut, reassign it again was not needed.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
kwonoj: seems like I can't move Function in here?

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -44,61 +44,58 @@ struct Actual {
...
                 in_object_prop: false,
                 in_prototype_assignment: false,
                 extra_stmts: vec![],
             };
-            let stmt = stmt.fold_with(&mut actual);
 
-            buf.extend(actual.extra_stmts.into_iter().map(T::from_stmt));
-            buf.push(stmt);
-        }
+            let stmt = &mut (stmts[idx]);
```
kdy1, line 77: Is `()` required? I think it can be used like `stmts[idx].visit_mut_with(self)` instead?

kdy1: Also, this will visit injected statements instead of source statements.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -192,27 +184,33 @@ impl Fold for Actual {
                 args,
                 type_args,
             }) if callee.is_fn_expr() => {
-                return self.handle_iife(span, (*callee).fn_expr().unwrap(), args, type_args);
+                *expr = self.handle_iife(
+                    span.take(),
```
kdy1, line 188: Can you change this to `*span`?

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -192,27 +184,33 @@ impl Fold for Actual {
...
-                return self.handle_iife(
-                    span,
-                    (*callee).paren().unwrap().expr.fn_expr().unwrap(),
+                *expr = self.handle_iife(
+                    *span,
+                    &mut (*callee).take().expect_paren().expr.expect_fn_expr(),
                     args,
-                    type_args,
+                    type_args.take(),
                 );
```
kdy1, line 214: I think `return` is required here. too.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -246,14 +244,15 @@ impl Fold for Actual {
                     _ => {}
                 }
 
-                let assign_expr = assign_expr.fold_children_with(self);
+                assign_expr.visit_mut_children_with(self);
                 self.in_prototype_assignment = false;
-                return Expr::Assign(assign_expr);
+                *expr = Expr::Assign(assign_expr.take());
```
kdy1, line 249: This is redundant as `expr` is always `Expr::Assign(assign_expr)`.
You can just delete this line.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -342,76 +344,70 @@ impl Fold for Actual {
...
-                        let function = self.fold_fn(expr.ident.clone(), expr.function, true);
-                        FnExpr {
-                            ident: expr.ident,
-                            function,
-                        }
+                if let DefaultDecl::Fn(expr) = &mut export_default.decl {
+                    if expr.function.is_async {
+                        self.visit_fn(expr.ident.clone(), &mut expr.function, true);
+
+                        let fn_expr = FnExpr {
```
kdy1, line 371: Seems redundant as `expr` is already `FnExpr`.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -342,76 +344,70 @@ impl Fold for Actual {
...
+                            ident: expr.ident.take(),
+                            function: expr.function.take(),
+                        };
+
+                        *item = ModuleItem::ModuleDecl(ModuleDecl::ExportDefaultDecl(
+                            ExportDefaultDecl {
+                                decl: fn_expr.into(),
```
kdy1, line 378: You can use `expr.take()` here I guess?

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -427,40 +423,31 @@ impl Fold for Actual {
...
 
-        let n = n.fold_children_with(self);
+        n.visit_mut_children_with(self);
 
         self.in_object_prop = old;
-
-        n
-    }
-
-    fn fold_stmts(&mut self, n: Vec<Stmt>) -> Vec<Stmt> {
```
kdy1, line 462: Explicit `visit_mut_stmts` with noop body is required.

kdy1 (changes requested):
I found some issues, thanks to vscode github extension.

kwonoj:
Appreciate for the thorough review, I tried update PR as much. 

I expect I'll make some mistakes, but didn't expect this many. :sweat_smile: Hope I could use this to make next attempt better.

kwonoj:
Also some of failing check seems due to npm woes:

https://github.com/swc-project/swc/runs/4167846153?check_suite_focus=true

```
[3/5] Fetching packages...
error An unexpected error occurred: "registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.14.5.tgz: Request failed \"500 Internal Server Error\"".
info If you think this is a bug, please open a bug report with the information provided in "D:\\a\\swc\\swc\\yarn-error.log".
info Visit yarnpkg.com/en/docs/cli/install for documentation about this command.
Error: Process completed with exit code 1.
```

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -44,61 +44,56 @@ struct Actual {
...
             let mut actual = Actual {
                 in_object_prop: false,
                 in_prototype_assignment: false,
                 extra_stmts: vec![],
             };
-            let stmt = stmt.fold_with(&mut actual);
 
-            buf.extend(actual.extra_stmts.into_iter().map(T::from_stmt));
-            buf.push(stmt);
+            stmts[idx].visit_mut_with(&mut actual);
```
kdy1, line 77: I think this is still problematic, as `idx` is not modified when `stmts` is extended.

kdy1: If `stmts.len() == 3` (stmt 0, 1, 2) and `stmt 0` creates `extra_stmts`, this code will visit `stmts[1]` which is content of `extra_stmts` from `stmt 0`.

kdy1 (changes requested):
It's almost there! I found one issue.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
kwonoj: Yes, that's correct oversight I missed. I wanted to reuse existing vec without having separate one, but it seems bit tricky as extra_stmts are non deterministic variable.

kdy1: You can add the length of `extra_stmts` to `idx`.

kwonoj: :bulb:

kwonoj: actually, when loop starts ending index is already binded.. :thinking: 

I'll try to accept cost by having another vec: I think those micro optimization can come later once logic works correctly.

kdy1: Oh, yes you are right. You can add same value (the length) to end index, too.
But recreating vector is fine for now anyway.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -44,61 +44,60 @@ struct Actual {
...
 
-            buf.extend(actual.extra_stmts.into_iter().map(T::from_stmt));
-            buf.push(stmt);
+            stmt.visit_mut_with(&mut actual);
+            stmts_updated.push(stmt);
+            stmts_updated.extend(actual.extra_stmts.into_iter().map(T::from_stmt));
         }
 
-        buf
+        stmts_updated.shrink_to_fit();
```
kdy1, line 82: This will be noop as `stmts_updated.len()` == `stmts.len() + sum(extra_stmts.len())`.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -342,76 +346,64 @@ impl Fold for Actual {
...
-                        }
+                if let DefaultDecl::Fn(expr) = &mut export_default.decl {
+                    if expr.function.is_async {
+                        self.visit_fn(expr.ident.clone(), &mut expr.function, true);
+                        *item = ModuleItem::ModuleDecl(ModuleDecl::ExportDefaultDecl(
+                            ExportDefaultDecl {
+                                decl: DefaultDecl::Fn(expr.take()),
+                                ..*export_default
+                            },
+                        ));
```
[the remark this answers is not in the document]
kwonoj: Updated, fyi this is code path for `binding(this)` related, which likely goes away when I attempt to amend https://github.com/swc-project/swc/issues/2716 soon.

kdy1 (approved):
Thanks!

kwonoj:
@kdy1 Thanks for going through all of this 🙇

Issue #299, opened 2019-03-01
[Lexer] jsx backslash escape parse not correct


LuoZijun:
Spec: https://facebook.github.io/jsx/

backslash escape (`\`) parse is not correct in `JSXAttributeValue` .

Tests:
```
<Page num="\ "> ABC hehe en</Page>;
<Page num="\""> ABC hehe en</Page>;
<Page num="\\"> ABC hehe en</Page>;
```

PR #2723, opened 2021-11-11
fix(ecma/lexer/jsx): jsx back slash escape parse

### LINKED ISSUE #2162: Issue parsing JSX attribute with single back slash
**Input code**

```ts
function test() {
  return (
      <>
          <A b="\" />
      </>
  );
}
```

**Expected behavior**

Should parse because back slashes are valid like that in JSX.

Ex. https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgIilQ3wG4BYAKABMk0AbFIuNCAOwGd4BBALjhRsAnhUpVMAVzYZg7ODCRcAFAEo4Abypw4RGBKhs4SrdtMAeAHwnTNuGe5wARgF58AHXxwA9Fcq27PiYqogC+VEA (even though syntax highlighting breaks)

**Version**

swc_ecmascript 0.57.0

RiESAEX:
attempt to fix #2162 
maybe relate to #299
I think babel only escape `\` to `\\` [babel playground](https://babeljs.io/repl#?browsers=IE%208&build=&builtIns=false&corejs=false&spec=true&loose=true&code_lz=MYewdgzgLgBAhjAvDAPAPgFAoIIwEaIDkAOoTAPRoDcWuBJArgKwAsApgGxmU075HEAHgCYAjN2q1-JUhUl96xDDAm9KQA&debug=false&forceAllTransforms=true&shippedProposals=true&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=env%2Creact%2Ctypescript&prettier=true&targets=&version=7.16.3&externalPlugins=&assumptions=%7B%7D)

### DIFF
--- a/crates/swc_ecma_parser/src/lexer/jsx.rs
+++ b/crates/swc_ecma_parser/src/lexer/jsx.rs
@@ -157,9 +157,8 @@ impl<'a, I: Input> Lexer<'a, I> {
             if ch == '\\' {
                 has_escape = true;
                 out.push_str(self.input.slice(chunk_start, cur_pos));
-
-                let chars = self.read_jsx_escaped_char(&mut Raw(None))?;
-                out.extend(IntoIterator::into_iter(chars).flatten().flatten());
+                out.push_str("\\");
+                self.bump();
 
                 chunk_start = self.input.cur_pos();
                 continue;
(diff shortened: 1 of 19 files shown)


crates/swc_ecma_parser/src/lexer/jsx.rs
```diff
@@ -157,9 +157,8 @@ impl<'a, I: Input> Lexer<'a, I> {
             if ch == '\\' {
                 has_escape = true;
                 out.push_str(self.input.slice(chunk_start, cur_pos));
-
-                let chars = self.read_jsx_escaped_char(&mut Raw(None))?;
-                out.extend(IntoIterator::into_iter(chars).flatten().flatten());
+                out.push_str("\\");
+                self.bump();
```
kdy1, line 161: I think this will break all other valid jsx usages.

kdy1 (commented):
Let's see if tests work.

crates/swc_ecma_parser/src/lexer/jsx.rs
kdy1: Hmm...
Tests succeed, but I'm still confused. It makes all tokens handled as `raw`... So I thought it would break things, but seems like it does not?

kdy1 (approved):
I was stupid, thanks!

PR #2726, opened 2021-11-11
refactor(css/ast): `Identifier` AST node

alexander-akait:
Replace `Text` on `Ident` (`Identifier` inside AST, like we do for JS). In CSS no text data type, in spec we have clearly https://www.w3.org/TR/css-values-3/#textual-values:

> The textual data types include various keywords and identifiers as well as strings (\<string\>) and URLs (\<url\>).

> CSS identifiers, generically denoted by \<ident\>, consist of a sequence of characters conforming to the \<ident-token\> grammar. [CSS3SYN] Identifiers cannot be quoted; otherwise they would be interpreted as strings. 

Also we have the same logic in JS:
```rust
#[ast_node("Identifier")]
#[derive(Eq, Hash, EqIgnoreSpan)]
pub struct Ident {
    pub span: Span,
    #[serde(rename = "value")]
    pub sym: JsWord,

    /// TypeScript only. Used in case of an optional parameter.
    #[serde(default)]
    pub optional: bool,
}
```

Let's use `Ident` here too


kdy1 (dismissed):

kdy1:
@alexander-akait Can you run `cargo fmt --all`?

kdy1:
Also, I'll flatten `css` into `crates` after merging this PR.

alexander-akait:
@kdy1 Yep, WIP

alexander-akait:
@kdy1 Ready for merge

kdy1 (approved):
Thanks!

alexander-akait:
@kdy1 

> Also, I'll flatten css into crates after merging this PR.

Yep, let't do it, I will continue after you moved them

PR #2732, opened 2021-11-12
fix(ecma/parser): check ts class name

### LINKED ISSUE #1218: Swc permits ts class names with disallowed names
**Describe the bug**
Swc's `invalid_class_name` check should check the same identifiers as `typescript::parse_ts_interface` instead of only `any`. Here: 
https://github.com/swc-project/swc/blob/master/ecmascript/parser/src/parser/class_and_fn.rs#L1077-L1097

Edit: it appears that tsc doesnt enforce this for `class intrinsic {}`, so that should be omitted.

**Input code**
```ts
class number {}
```

**Config**
None

**Expected behavior**
`Class name cannot be 'number'`

**Version**
The version of @swc/core: `1.2.37`

**Additional context**
None, explained in the first section

RiESAEX:
attemp to fix #1218
js allow any identifier be a class name (I think), 
and ts disallow "any", "unknown", "never", "number", "bigint", "boolean", "string", "symbol", "void", "object".

### DIFF
--- a/crates/swc_ecma_parser/src/parser/class_and_fn.rs
+++ b/crates/swc_ecma_parser/src/parser/class_and_fn.rs
@@ -85,8 +85,10 @@ impl<'a, I: Tokens> Parser<I> {
             expect!(p, "class");
 
             let ident = p.parse_maybe_opt_binding_ident()?;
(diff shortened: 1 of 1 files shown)


kdy1 (approved):
Thanks!

Issue #2656, opened 2021-11-05
JSX automatic transform development incorrect


kitsonk:
### Describe the bug

When attempting to transform JSX, with the setting `development` to `true`, swc does not properly transpile like Babel or TypeScript.

It provides something like this:

```js
import { jsx as _jsx, Fragment as _Fragment } from "react/jsx-runtime";
function App() {
    return(/*#__PURE__*/ _jsx("div", {
        __source: {
            fileName: "https://deno.land/x/mod.tsx",
            lineNumber: 4
        },
        __self: this,
        children: /*#__PURE__*/ _jsx(_Fragment, {
        })
    }));
}
```

### Input code

```typescript
function App() {
  return (
    <div><></></div>
  );
}
```

Using a JSX pass with the options like this:

```rust
let options = swc_ecma_transforms_react::react::Options {
  runtime: Some(swc_ecma_transforms_react::react::Runtime::Automatic),
  development: true,
  import_source: "react".to_string(),
  ..Default::default()
};
```

### Expected behavior

This is what [Babel outputs with development plugin](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.6&spec=false&loose=false&code_lz=KYDwDg9gTgLgBAMwK4DsDGMCWEVwIJhgAUAlHAN4BQccUwMSUuR1NcAPACaYBuAfOwEB6Yd36sSAbkoBfIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=react%2Cstage-2%2Ctypescript&prettier=false&targets=&version=7.16.2&externalPlugins=%40babel%2Fplugin-transform-react-jsx-development%407.16.0&assumptions=%7B%7D):

```js
var _jsxFileName = "/repl.tsx";
import { Fragment as _Fragment } from "react/jsx-dev-runtime";
import { jsxDEV as _jsxDEV } from "react/jsx-dev-runtime";
export function App() {
  return /*#__PURE__*/_jsxDEV("div", {
    children: /*#__PURE__*/_jsxDEV(_Fragment, {}, void 0, false)
  }, void 0, false, {
    fileName: _jsxFileName,
    lineNumber: 3,
    columnNumber: 5
  }, this);
}
```

[...]

PR #2741, opened 2021-11-13
Fix: incorrect transform `jsx-runtime` in development

Shinyaigeek:
## Motivation 🔥 

Fix: #2656 

Currently, swc does not transform JSX correctly with new JSX Transform in development mode. It is importing `jsx` from `jsx-runtime` when it should be importing `jsxDEV` from `jsx-dev-runtime`. (The same applies to `jsxs`).

I fixed this.


kdy1 (changes requested):
This is not enough.
You have to handle special attributes like `__source`, too.

Shinyaigeek:
Oh! Sorry, I overlooked it :pray:

I updated code with passing `isStaticValidation`, `source`, `self` (this) to `jsxDEV`, and not set `__source` and `__self` as props with `"runtime": "automatic"` config.

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -596,13 +600,62 @@ where
 
                 self.top_level_node = top_level_node;
```
kdy1, line 602: I think this code will error if `/* @jsxRuntime automatic */` is used.

kdy1: Because of this, the babel plugin detect attributes named `__source` and `__self` and change those into arguments.

kdy1 (changes requested):
I think there's are small issues with jsx directives.

crates/swc_ecma_transforms_react/src/jsx/mod.rs
Shinyaigeek: I’m sorry for that I did not know about jsx directives.

> Because of this, the babel plugin detect attributes named __source and __self and change those into arguments.

Is this the process below?

https://github.com/babel/babel/blob/main/packages/babel-plugin-transform-react-jsx/src/create-plugin.ts#L502

I am worried if my understanding is correct, but should I fix the code to set `__source` and `__self` props in development mode as usual?

kdy1: Sorry for late response, I missed the email.


> Is this the process below?


Yes.

>  Should I fix the code to set __source and __self props as JSX attribute in development mode as usual?

I think so, as it's what babel does

Shinyaigeek: Thank you for your reply and help!

I fixed code like below

- set `__source` and `__self` to props with `jsx_src` and `jsx_self` (as usual)
- with `runtime: automatic` and `development: true` configuration, swc extract `__source` and `__self` from JSX’s attributes and set it as `_jsxDEV`’s arguments.

it looks like there are not test settings to test `react` module(= `jsx` with `jsx_src` and  `jsx_self`) with a specific configuration, so I added `integration` testing directory to test `react` module with a specific configuration.

If there is yet something I missed, I will be happy if you pointed it out.

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -489,6 +495,44 @@ where
                                         continue;
                                     }
 
+                                    if !use_create_element
+                                        && i.sym == String::from("__source")
```
kdy1, line 499: `String::from("__source")` is very inefficient.
You can compare `JsWord` to `str` like `*i.sym === "__source ".`

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -489,6 +495,44 @@ where
...
+                                            .map(|expr| ExprOrSpread { expr, spread: None });
+                                        assert_ne!(
+                                            __source, None,
+                                            "value of property '__source' should not be empty"
+                                        );
+                                        continue;
+                                    }
+
+                                    if !use_create_element
+                                        && i.sym == String::from("__self")
```
kdy1, line 518: Same here

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -468,6 +472,8 @@ where
                 };
 
                 let mut key = None;
+                let mut __source = None;
```
kdy1, line 475: I think variable name should not use `_` as a prefix as it has special meaning in rust.

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -596,13 +640,52 @@ where
...
+                    let key = match key {
+                        Some(key) => key,
+                        None => ExprOrSpread {
+                            spread: None,
+                            expr: Box::new(build_undefined_literal_value()),
+                        },
+                    };
+
+                    // set undefined literal to __source if __source is None
+                    let __source = match __source {
```
kdy1, line 655: Same here. `_` prefix is special in rust.

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -596,13 +640,52 @@ where
...
                 self.top_level_node = top_level_node;
 
+                let args = once(name.as_arg()).chain(once(props_obj.as_arg()));
+                let args = if self.development {
+                    // set undefined literal to key if key is None
+                    let key = match key {
+                        Some(key) => key,
+                        None => ExprOrSpread {
+                            spread: None,
+                            expr: Box::new(build_undefined_literal_value()),
```
kdy1, line 650: Please use `undefined(DUMMY_SP)` instead.

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -596,13 +640,52 @@ where
...
+                        },
+                    };
+
+                    // set undefined literal to __source if __source is None
+                    let __source = match __source {
+                        Some(__source) => __source,
+                        None => ExprOrSpread {
+                            spread: None,
+                            expr: Box::new(build_undefined_literal_value()),
```
kdy1, line 659: Same here, `undefined(DUMMY_SP)` would be better.

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -596,13 +640,52 @@ where
...
+                        },
+                    };
+
+                    // set undefined literal to __self if __self is None
+                    let __self = match __self {
+                        Some(__self) => __self,
+                        None => ExprOrSpread {
+                            spread: None,
+                            expr: Box::new(build_undefined_literal_value()),
```
kdy1, line 668: Same here, `undefined(DUMMY_SP)` would be better.

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -596,13 +640,52 @@ where
...
+                    };
+                    args.chain(Some(key))
+                        .chain(once(
+                            Lit::Bool(Bool {
+                                span: DUMMY_SP,
+                                value: use_jsxs,
+                            })
+                            .as_arg(),
+                        ))
+                        .chain(Some(__source))
```
kdy1, line 679: There's `std::iter::once` and it's better than using `Some` to create an iterator.

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -596,13 +640,52 @@ where
...
+                        .chain(Some(__self))
```
kdy1, line 680: Same here.

kdy1 (changes requested):
Thank you!

crates/swc_ecma_transforms_react/src/jsx/mod.rs
Shinyaigeek: 👍    1ed2c7b

crates/swc_ecma_transforms_react/src/jsx/mod.rs
Shinyaigeek: 👍  7fcafd2

I replaced `__source` with `source_props`.

crates/swc_ecma_transforms_react/src/jsx/mod.rs
Shinyaigeek: I did not notice `undefined` util module :bow:

:+1: 0ac5ded

crates/swc_ecma_transforms_react/src/jsx/mod.rs
Shinyaigeek: 👍   75a12b8

kdy1:
Code changes look fine to me, but can you add some more tests?
Actually this feature was once implemented by https://github.com/swc-project/swc/pull/2542, but I reverted it because it broke next.js

Shinyaigeek:
Thank you! 

I've added some tests on this, but I am wondering what more tests to add.

Can you share what happened or error logs when Next.js was broken? (I searched for the Next.js issue but couldn't find it)

kdy1:
@Shinyaigeek https://github.com/vercel/next.js/pull/30302#issuecomment-952042021
This was the issue, and thank you!

Shinyaigeek:
I’m sorry for the late response, I didn’t have time due to my graduation thesis.

Based on the report you shared, I investigated the contexts and test cases we missed and found the cause of the "next / dynamic'dev mode' should render dynamic server rendered values ​​on client mount" error. This is that `isStaticChildren` would always be undefined (cast false) in React.Fragment.

Another, "Invalid hrefs ›dev mode ›should not show error for function component with forwardRef" The cause of these errors is not well understood, but `next / link` should have only one child element. , I'm guessing that an error has occurred there. I will continue to investigate it a bit more

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -596,13 +665,52 @@ where
...
                 self.top_level_node = top_level_node;
 
+                let args = once(name.as_arg()).chain(once(props_obj.as_arg()));
+                let args = if self.development {
+                    // set undefined literal to key if key is None
+                    let key = match key {
+                        Some(key) => key,
+                        None => ExprOrSpread {
+                            spread: None,
+                            expr: Box::new(*undefined(DUMMY_SP)),
```
kdy1, line 675: You can do `undefined(DUMMY_SP)` instead of `Box::new(*undefined(DUMMY_SP))`.

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -596,13 +665,52 @@ where
...
+                        },
+                    };
+
+                    // set undefined literal to __self if __self is None
+                    let self_props = match self_props {
+                        Some(self_props) => self_props,
+                        None => ExprOrSpread {
+                            spread: None,
+                            expr: Box::new(*undefined(DUMMY_SP)),
```
kdy1, line 693: Same here.

crates/swc_ecma_transforms_react/tests/jsx/fixture/runtime/jsxdev-args-with-fragment/output.js
```diff
@@ -0,0 +1,13 @@
...
+        _jsxDEV("div", {
+            children: "hoge"
+        }, void 0, false, void 0, void 0),
+        _jsxDEV("div", {
+            children: "fuga"
+        }, void 0, false, void 0, void 0)
+    ]
+},
+void 0,
+true);
```
kdy1, line 13: I got different output from babel.

```js
var _jsxFileName = "/Users/kdy1/projects/lab/src/input.js";
import { jsxDEV as _jsxDEV } from "react/jsx-dev-runtime";
import { Fragment as _Fragment } from "react/jsx-dev-runtime";

var x = /*#__PURE__*/_jsxDEV(_Fragment, {
  children: [/*#__PURE__*/_jsxDEV("div", {
    children: "hoge"
  }, void 0, false, {
    fileName: _jsxFileName,
    lineNumber: 3,
    columnNumber: 9
  }, this), /*#__PURE__*/_jsxDEV("div", {
    children: "fuga"
  }, void 0, false, {
    fileName: _jsxFileName,
    lineNumber: 4,
    columnNumber: 9
  }, this)]
}, void 0, true);
```

Seems like children of fragments are not visited?

kdy1: Can you add `e.visit_mut_children_with(self)` to https://github.com/swc-project/swc/blob/79f6e2e2a406c6d7c98ea91884dd61e756b7925d/crates/swc_ecma_transforms_react/src/jsx_src/mod.rs#L38 and https://github.com/swc-project/swc/blob/79f6e2e2a406c6d7c98ea91884dd61e756b7925d/crates/swc_ecma_transforms_react/src/jsx_self/mod.rs#L39

kdy1 (changes requested):
Thanks!
I found some tiny issues.

crates/swc_ecma_transforms_react/src/jsx/mod.rs
Shinyaigeek: Thanks ! 👍   8a8fc1e

crates/swc_ecma_transforms_react/tests/jsx/fixture/runtime/jsxdev-args-with-fragment/output.js
Shinyaigeek: I think this works. `jsxdev-args-with-gragment` test is placed in `fixture`, not in `integration`. This means `jsx-src` and `jsx-self` does not work while this test run. This output will be the same with babel’s output if we move this test into `integration`. I think this is a little confusing, so I moved this test case from `fixture` into `integration` in fdaaaa2

Shinyaigeek:
I’m sorry I did not understand why #2542 broke next.js (https://github.com/vercel/next.js/pull/30302#issuecomment-952042021). I think there are enough test cases for this change, but I am also worried that this change will break next.js as https://github.com/vercel/next.js/pull/30302#issuecomment-952042021 . Is there any documentation for running Next.js tests with the local version of swc? (I rewrote Cargo.toml of `next-swc` with {path = "..."}, I get a compile error ...)

kdy1:
@Shinyaigeek Same here. I don't understand the reason.

I think in this case it will be fine, as you added tests for all input. But if you want to try it, you can patch all swc dependencies in https://github.com/vercel/next.js/tree/c004322d138e1949437efb0b80e0973f3a3775cf/packages/next-swc/crates to use git or path dependency targeting your own branch and create a PR.

kdy1:
But I really think it will be fine in this case, as we tested all input.

kdy1:
But I really think it will be fine in this case, as we tested all input.

Shinyaigeek:
@kdy1 I see. It seems to be a tough task to test with `next-swc`, so is it okay to get a review as it is, and send a follow up issue & PR if there is any problem?

kdy1:
Yeah, I think it is better. Thanks for the pr again and I'll review it soon

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -396,10 +401,33 @@ where
                     }
                 }
 
+                let args = vec![fragment.as_arg(), props_obj.as_arg()];
+
+                let args = if self.development {
+                    args.into_iter()
+                        .chain(
```
kdy1, line 408: I think `.chain(once(undefined(DUMMY_SP).as_arg())).chain(once(Lit::Bool(Bool {
                                    span: DUMMY_SP,
                                    value: use_jsxs,
                                })
                                .as_arg()))` would be better.

kdy1 (changes requested):
Thanks! I found one perf/code style issue.

crates/swc_ecma_transforms_react/src/jsx/mod.rs
Shinyaigeek: OK 👍  Thank you for helping! I fixed in 5be5cf0

kdy1 (approved):
LGTM. Thanks!

Issue #2370, opened 2021-10-07
Performance bugs with known fixes


kdy1:
### general

 - [ ] Dynamic linking of iconv

 - [ ] Dynamic linking of libresolv

 - [x] Use `ahash` #2384()

### `swc_atoms`

 - [x] Switch to parking_lot (https://github.com/swc-project/swc/pull/2411)

Related: https://github.com/servo/string-cache/issues/242

 - [ ] Thread-local interning, in cost of more memory use while `LocalInterner` is alive.

### `swc_common`

 - [ ] Global / Local pool for `Mark` allocation.

This will result in a huge difference as `Mark` is used for various purposes.

### `swc_ecma_ast`

 - [ ] Methods of `IdentLike` can be optimized.

Comparison of `js_word` is much faster than string comparison.

### `swc_ecma_parser`

 - [ ] Skip typescript types, without storing them.

 - [x] Optimize `parse_block` (#2406)

We should create a vector with good initial capacity.

 - [ ] Optimize parsing of expressions. (#2437)

We perform too much backtracking.

### `swc_ecma_codegen`

 - [ ] Don't allocate while print a number.

 - [ ] Use faster library like `ryu` for printing numbers.

### `swc_ecma_transforms_base`

 - [x] `resolver`: Use better way to fetch parent `Mark`. (#2392)

It uses `Mark::parent` but it's very slow operation.
This also shows up in the profile result, as `Mark::parent()` involves a global lock.

 - [ ] `hygiene`: Reduce allocation of `JsWord`

Creating a `JsWord` requires locking global interner.

### `swc_ecma_utils`

 - [x] Parallel `collect_decls`. (#2394)

It's easy to parallelize it.

### `swc_ecma_transforms_base`

 - [ ] Use single `Mutex` instead of lots of `AtomicBool`s for helper.

### `swc_ecma_transforms_macros`

 - [x]  `#[fast_path]` has exponential time complexity. (#2439)

### `swc_ecma_transforms_compat`

 - [ ] `destructuring` visits nodes for exponential times.

 - [ ] `for_of` visits nodes for exponential times.

 - [ ] `object_rest` has exponential time complexity.

 - [x] `async_to_generator` visits nodes for exponential times. (#2442)

These are simple, but I postponed because of #2441.

---

`Fold` results in a lot of `memmove`. So transforms in `@swc/core` should use `VisitMut` instead of `Fold`.

The order of tasks is ordered by real-world data.
The sooner it comes, the higher effect it has.

[...]

PR #2751, opened 2021-11-15
refactor(es/transforms/compat): apply visitmut for export_namespace_from

### LINKED ISSUE #2716: Async to regenerator explicitly binds this context when created
### Describe the bug

I'm hitting some series of https://github.com/swc-project/swc/issues/2305, with various edge cases. We've fixed by not binding to context for some known cases. While, comparing emitted output between swc / babel, both do the similar like

```
function tt() {
  return _tt.apply(this, arguments);
}

function _tt() {
  _tt = _asyncToGenerator(function* () {
    ...
  }); // if swc, do the explicit .bind(this) here
  return _tt.apply(this, arguments);
}
```

which returns wrapped fn calls .apply(this) to bind this context when fn is being called

### Input code

_No response_

### Config

_No response_

### Playground link

_No response_

### Expected behavior

this context would be preserved for each fn per call time.

### Version

master

### Additional context

discord discussion thread: https://discord.com/channels/889779439272075314/889785438938726441/908256585313312819

kwonoj:
relates to https://github.com/swc-project/swc/issues/2370.

I found it's bit tricky to resolve https://github.com/swc-project/swc/issues/2716 and trying to do some refresher, continuing #2370 by replace export_namespace_from into visitMut.

There seems some changes to optimize things further, but did choose straightforward migration for now and do the another round of optimization if it's applicable.

### DIFF
--- a/crates/swc_ecma_transforms_compat/src/es2020/export_namespace_from.rs
+++ b/crates/swc_ecma_transforms_compat/src/es2020/export_namespace_from.rs
@@ -1,27 +1,24 @@
 use swc_common::DUMMY_SP;
 use swc_ecma_ast::*;
 use swc_ecma_utils::{IdentExt, IsDirective};
-use swc_ecma_visit::{noop_fold_type, Fold};
+use swc_ecma_visit::{as_folder, noop_fold_type, noop_visit_mut_type, Fold, VisitMut};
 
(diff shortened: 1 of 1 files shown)


kdy1 (approved):
LGTM. Thanks!

PR #2756, opened 2021-11-15
refactor(css/ast): added `QualifiedRule` AST node

alexander-akait:
@kdy1 There is not `StyleRule` in CSS spec, `StyleSheet` can contain:
1. \<CDO-token\>
2. \<CDC-token\>
3. \<whitespace-token\>
4. Qualified rule
5. At-rule

So I rename `StyleRule` on `QualifiedRule` according to spec.

Also there is no `selectors` property on these (`At-rule` and `Qualified rule`) AST nodes, spec has more generic name -  `prelude`:

> at-rule
> An at-rule has a name, a prelude consisting of a list of component values, and an optional block consisting of a simple {} block.
Note: This specification places no limits on what an at-rule’s block may contain. Individual at-rules must define whether they accept a block, and if so, how to parse it (preferably using one of the parser algorithms or entry points defined in this specification).

> qualified rule
> A qualified rule has a prelude consisting of a list of component values, and a block consisting of a simple {} block.
Note: Most qualified rules will be style rules, where the prelude is a selector [SELECT] and the block a list of declarations.

Why better to use `prelude`? Because `Most qualified rules will be style rules`, but not all of them, so better to use more generic names like in spec.


alexander-akait:
@kdy1 Also please review this, in near future we will finish big AST changes and focused on bugs and features

kdy1 (approved):
Thanks!
I didn't look at tests as there's no change in test input and ci passing means changes are all about renamed stuffs.

alexander-akait:
@kdy1 Looks like random fail

PR #2759, opened 2021-11-15
feat(css/ast): Add types for `An+b` syntax

alexander-akait:
`An+B` syntax (https://www.w3.org/TR/css-syntax-3/#anb-microsyntax):
- new syntax supported
- new AST nodes (you can see them in the PR)
- recovery mode

It was not easy, the specification is complex and not clear in some places, but I have finished.

Also here we have new experimental syntax (not implemented in this PR):

```
:nth-child(2n+1 of li,.test) {}
```

But we need fix small problems in our selector parser before fix it, I will do it after we merge it and https://github.com/swc-project/swc/pull/2675, better to merge this PR firstly


alexander-akait:
@kdy1 Ready for review, should be merged before https://github.com/swc-project/swc/pull/2675 (to avoid big rebase)

crates/swc_css_ast/src/selector.rs
```diff
@@ -130,12 +131,48 @@ pub struct AttrSelector {
...
+    Ident(Ident),
+}
+
 #[ast_node("PseudoSelector")]
 pub struct PseudoSelector {
     pub span: Span,
     pub is_element: bool,
     pub name: Ident,
-    pub args: Tokens,
+    pub children: Option<PseudoSelectorChildren>,
```
kdy1, line 175: Is empty `Tokens` not enough?

kdy1:
I have a question about the change.

crates/swc_css_ast/src/selector.rs
alexander-akait: @kdy1 we will have more different types here in future, for example `:is` will contains selector list, other normal values, also some of them do not contains `args`, for example `:focus {}`, so `children` (previously `args`) is optional

alexander-akait: We do the same with any optional fields for our AST Nodes

kdy1: Thanks for description!

kdy1 (approved):

alexander-akait:
@kdy1 Can we merge?

kdy1:
I'll update base branch once https://github.com/swc-project/swc/pull/2764  is merged.

PR #2762, opened 2021-11-16
fix(es/transforms/compat): preserve span for the method

### LINKED ISSUE #2758: async object methods annotated by comments are incorrectly compiled into methods returning immediately after the `return` keyword
### Describe the bug

With swc, code like the following:

```js
const obj = {
    // A comment
    async foo() {
        console.log("This code will never execute!");
    }
}
```

will be transpiled into the following, when targeting ES5 (but the same problems exist when targeting ES2015 and ES2016 too):

```js
// Using external helpers for making output more easily readable
var obj = {
    foo: function() {
        return // A comment
        swcHelpers.asyncToGenerator(regeneratorRuntime.mark(function _callee() {
                    case 0:
                        console.log("This code will never execute!");
                    case 1:
                    case "end":
                        return _ctx.stop();
                }
            }, _callee);
        }))();
    }
};
```

This code will never execute anything after the `return` keyword, as the [Automatic Semicolon Insertion (ASI)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#automatic_semicolon_insertion) rules dictate that syntax such as this:

```js
return
a + b
```

Is transformed by ASI into this:

```js
return;
a + b;
```

So, this is why the engine interprets the newline after the `return` keyword followed by a comment as a statement.

In the code example in the top of the issue, swc will produce incorrect code when a method inside an object literal is annotated with a comment. It can be a single-line comment, multi-line comment or a JSDoc comment - it doesn't matter.

Here's how it looks to the engine:

```js
return; // A comment
swcHelpers.asyncToGenerator( /* ... /*)
```

I found this bug after investigating why async object methods always came out completely empty after tree-shaking only to realize all async methods within object literals with comments returned immediately.

### Config

This happens with any swc config targeting an EcmaScript version below ES2017

### Playground link

https://play.swc.rs/?version=1.2.106&code=H4sIAAAAAAAAA0vOzysuUchPylKwVajm5VIAAn19BUeF5Pzc3NS8EohIYnFlXrJCWn6%2BhiZMEQgkA%2FXm56Tq5eSnayiFZGQWA0VSUhXKM3NyFPJSy1KLFFIrUpNLS1IVlTStIdpqeblqAfEGEn1zAAAA&config=H4sIAAAAAAAAAz2NMQ7DIBAE%2F7I1bVLwgjR5xImcIyIM6O4sgSz%2FPWBZLnc0u7vjpwF%2BRypFGX6hpOzAzVgypRenyqLwJtvAlURZpq49GzV4cFhJg8RqcGOrXROHg5F82aaiD4y8ls%2BWeJatVz758%2BQxx6Xf11Hflzg%2Fjz8EB6o5oQAAAA%3D%3D

### Expected behavior

You could either always wrap the entire thing in a parenthesized expression such as:

```js
var obj = {
    foo: function() {
        return (// A comment
            swcHelpers.asyncToGenerator(/* ... /*);
        );
    }
};
```

Or, you could move the comment up and ensure that the method body continues on the same line as the `return` keyword:

```js
var obj = {
    // A comment
    foo: function() {
        return swcHelpers.asyncToGenerator(/* ... /*);
    }
};
```

### Version

1.2.108

### Additional context

_No response_

kwonoj:
- closes #2758 

It looks like span gets swapped between inner fn body to fn itself when updating fn expr for the object props. PR preserves original span for the fn if available, otherwise fallback to dummy span instead.


kdy1 (approved):
Thank you!

PR #2764, opened 2021-11-16
chore: Fix typo in docs

KnorpelSenf:
Fixes a typo on the main page of the docs, visible at https://rustdoc.swc.rs/swc/

Also removes the incorrect word (“cutomizing” instead of “customizing”) from the local cspell vocabulary. It was added there in #2472. Reformats the vocabulary with Prettier according to the config in this git tree.


kdy1 (approved):
Thanks!

Issue #2701, opened 2021-11-09
Comments above an object method declaration cause an early return to be added to the function body


hexmodr:
### Describe the bug

The output generated by SWC adds in an early return to async function when defined using the short-hand method syntax. This leads to no code in the method body being executed

### Input code

```typescript
const x = {
  // i am some comment
  async hello(arg) {
    console.log("Hello")
  }
}
```

### Config

```json
{
    "jsc": {
        "externalHelpers": false,
        "parser": {
            "syntax": "typescript",
            "tsx": true,
            "decorators": true
        }
    },
    "module": {
        "type": "commonjs"
    }
}
```

### Playground link

https://play.swc.rs/?version=1.2.106&code=H4sIAAAAAAAAA0vOzysuUahQsFWo5lJQ0NdXyFRIzFUozs9NVUjOz81NzSsBCicWV%2BYlK2Sk5uTkayQWpWuC1SoAFeQV5%2Bek6uXkp2soeYBklTSBErVctVwAp5%2F48lcAAAA%3D&config=H4sIAAAAAAAAAy2NQQrDMAwE%2F7Jnv8Av6KWPEI4KCbZlJAUSgv9eq%2FQ6O8M%2BOKwgP%2BDLWTvVF9fBasgfqsYJg9RYw7C7O13I8HuwFd2HI8FtIddzqRsXUXKJOsicCU22s3LkUa24SGvSD8Mad3v%2F59%2FZ%2FALw%2FWoujQAAAA%3D%3D

### Expected behavior

It to not generate an invalid early return in the function

### Version

1.2.106

UMASHIBA1:
Can I pick this?

kdy1:
@UMASHIBA1 Of course, thank you!

PR #2766, opened 2021-11-16
fix(es/transform/compat) fix early return caused by comment above async func in object

UMASHIBA1:
Fix https://github.com/swc-project/swc/issues/2701
I fixed early return caused by comment above async method in object.
if something wrong, please point it out🙇‍♂️


UMASHIBA1:
If I'm wrong I'm sorry. Does this PR resolved same problem...?
https://github.com/swc-project/swc/pull/2762

kwonoj:
:oh-no: yes, it looks like aiming same issues while I tried to reuse exising span. Sorry I didn't aware if there was overlapping work.

kdy1 (approved):
The new code can panic if someone trolls, but rust code is required to such trolling, so it's fine IMO.

kwonoj:
@kdy1 do you think this is still required after #2762? My impression was preserving original span would be ok.

kwonoj:
Looks like #2758 was dupe of #2701 and I didn't realize it.

kdy1:
@kwonoj Oh, you are right. I'll test if it works with this PR reverted.

kdy1:
@kwonoj I'll revert this PR as it's not required.

UMASHIBA1:
> :oh-no: yes, it looks like aiming same issues while I tried to reuse exising span. Sorry I didn't aware if there was overlapping work.

OK👍 Thank you for your great work. I'm glad the issue was solved early.

PR #2770, opened 2021-11-16
fix(css): better recovery mode

alexander-akait:
Better recovery mode, now we can parse
```css
{}
a { color: red }
```

Why we need `selectors` as optional?
- Recovery mode
- Ideally recovery mode should be implement inside `parser/selector/mod.rs`, because we have two bad states - bad selector and bad simple block, now we use the same logic for both states, but due this we lose ability to recovery a lot of case (after merge this PR I will send a new with more cases)
- Also our selector parser is greedy, but it is invalid (and wrong by spec), because we can't reuse parser for parsing selectors inside other things, for example
```css
/* 'li, .test' is selector */
:nth-child(2n+1 of li,.test) {}
```
Also we will need it for `:is()`/`:has()`/etc pseudo-classes

Due greedy behavior we always faced `Err(InvalidSelector)` if the next token is not accepted by selector parser, we should refactor it on `lazy`, it means some function like `parse_selectors` and `parse_complex_selector` should return `PResult<Option<SelectorList>>`/`PResult<Option<ComplexSelector>>`, so it give ability to - reuse parser inside other things + good recovery mode

Yep, it can be slightly dirty right now, but it for future improvements :star:


kdy1 (changes requested):
What's the point of recovering from `{}`?
I think such error recovery should be the internal logic of parser, without changing ast type definitions.

alexander-akait:
Yes, it is invalid, I found better solution for this, I will send a PR

PR #2776, opened 2021-11-17
fix(es/transforms/compat): capture this context scope for the class method

kwonoj:
Part 1 for #2716.

This PR is an attempt to remove explicit binding to `this`. As a first step trying to remove scoped this in class method.
However, I *do not think* this is ok to check in as-is due to some of the notable issues to be resolved. I left some comments for the known part, and there may be possibly more.

As Part 1 implies this won't resolve all of the cases. But I found trying to solve all at once would take some amount of effort,  trying to make gradual progress with known possible paths.

### DIFF
--- a/crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
+++ b/crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
@@ -63,8 +63,6 @@ impl AsyncToGenerator {
         T: StmtLike + VisitMutWith<Actual>,
         Vec<T>: VisitMutWith<Self>,
     {
-        stmts.visit_mut_children_with(self);
-
         let mut stmts_updated = Vec::with_capacity(stmts.len());
 
         for mut stmt in stmts.drain(..) {
@@ -80,6 +78,7 @@ impl AsyncToGenerator {
         }
 
         *stmts = stmts_updated;
+        stmts.visit_mut_children_with(self);
     }
 }
 
@@ -121,17 +120,38 @@ impl VisitMut for Actual {
         }
         let params = m.function.params.clone();
 
-        let mut folder = MethodFolder { vars: vec![] };
+        let mut folder = MethodFolder {
+            vars: vec![],
+            scoped_this_ident: None,
+        };
         m.function.params.clear();
 
         m.function.visit_mut_children_with(&mut folder);
+
+        let this_decl = if let Some(this_ident) = folder.scoped_this_ident {
+            let decl = VarDecl {
+                span: DUMMY_SP,
+                kind: VarDeclKind::Var,
+                declare: false,
+                decls: vec![VarDeclarator {
+                    span: DUMMY_SP,
+                    name: Pat::Ident(this_ident.into()),
+                    init: Some(Box::new(Expr::This(ThisExpr { span: DUMMY_SP }))),
+                    definite: false,
+                }],
+            };
+
+            Some(Stmt::Decl(Decl::Var(decl)))
+        } else {
+            None
+        };
 
         let expr = make_fn_ref(
             FnExpr {
                 ident: None,
                 function: m.function.take(),
             },
-            self.in_object_prop || self.in_prototype_assignment,
+            false,
         );
 
         let hoisted_super = if folder.vars.is_empty() {
@@ -154,6 +174,7 @@ impl VisitMut for Actual {
                 span: DUMMY_SP,
                 stmts: hoisted_super
                     .into_iter()
+                    .chain(this_decl.into_iter())
                     .chain(iter::once(Stmt::Return(ReturnStmt {
                         span: DUMMY_SP,
                         arg: Some(Box::new(Expr::Call(CallExpr {
@@ -459,6 +480,49 @@ impl VisitMut for Actual {
     fn visit_mut_stmts(&mut self, _n: &mut Vec<Stmt>) {}
 }
 
+struct ThisReplacer<'a> {
+    to: &'a Ident,
+}
+
+impl VisitMut for ThisReplacer<'_> {
+    noop_visit_mut_type!();
+
+    fn visit_mut_arrow_expr(&mut self, _: &mut ArrowExpr) {}
+
+    fn visit_mut_constructor(&mut self, _: &mut Constructor) {}
+
+    fn visit_mut_expr(&mut self, e: &mut Expr) {
+        e.visit_mut_children_with(self);
+
+        match e {
+            Expr::This(..) => {
+                *e = Expr::Ident(self.to.clone());
+            }
+
+            _ => {}
+        }
+    }
+
+    fn visit_mut_function(&mut self, _: &mut Function) {}
+
+    /// Don't recurse into fn
+    fn visit_mut_getter_prop(&mut self, n: &mut GetterProp) {
+        n.key.visit_mut_with(self);
+    }
+
+    /// Don't recurse into fn
+    fn visit_mut_method_prop(&mut self, n: &mut MethodProp) {
+        n.key.visit_mut_with(self);
+        n.function.visit_mut_with(self);
+    }
+
+    /// Don't recurse into fn
+    fn visit_mut_setter_prop(&mut self, n: &mut SetterProp) {
+        n.key.visit_mut_with(self);
+        n.param.visit_mut_with(self);
+    }
+}
+
 /// Hoists super access
 ///
 /// ## In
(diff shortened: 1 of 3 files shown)


crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -80,6 +78,7 @@ impl AsyncToGenerator {
         }
 
         *stmts = stmts_updated;
+        stmts.visit_mut_children_with(self);
```
kwonoj, line 81: This is the biggest part of the headache. I know this is not feasible, as this'll require redundant visit against `extra_stmt` generated. But the order of visit actually matters, if this occurs before visiting individual stmts, inner scope of fn gets visited first before class_method reaches and there's no easy way to detect if fn's `this` scope is inside of class or not.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -459,6 +480,49 @@ impl VisitMut for Actual {
     fn visit_mut_stmts(&mut self, _n: &mut Vec<Stmt>) {}
 }
 
+struct ThisReplacer<'a> {
```
kwonoj, line 483: this probably needs some :eye: , since it's copy-paste from arrow and maybe need to be reused instead of duplication.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -564,6 +629,39 @@ impl VisitMut for MethodFolder {
     noop_visit_mut_type!();
 
     fn visit_mut_expr(&mut self, expr: &mut Expr) {
+        match expr {
```
kwonoj, line 632: To visit inner scoped fn body extended MethodFolder. maybe ok? or should it be separate visitor?

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
Austaras: You would need `arguments` replaced too

kwonoj: Would you mind elaborate? not sure if I correctly understand what need to be updated.

Austaras: You still need `ArgumentsReplacer` from `arrow`

kwonoj: Yes, I aware of those - I was curious if it's related with current `this` replacement or you're thinking it's better to mimic what arrow does regardless since both are going through nearly same path. as discussed in the issue itself and wrote in here (`maybe need to be reused instead of duplication`) probabaly, eventually these two need to be consoliated though.

Austaras: IMO we could have a combined visitor for `this` and `arguments`, and use it both here and in arrow

kwonoj: sounds like a considerable approach, but I'd probabaly suggest to tackle it as separate PR & issue since it'll contain bit out of the scope work for the current PR.

for those, I'm curious how to handle differences between 2 visitor, for example argreplacer does for `visit_mut_member_expr`, thisreplacer for `visit_mut_arrow_expr`. Anyway, I think it's worth to file an issue and track.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
Austaras: Wait, why do we need to know whether it's inside class or not?

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
Austaras: `MemberExpr` doesn't need any special handle?

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
kwonoj: to determine captured `this` to be inserted in the fn. (`var _this = this`), not all of this context needs this.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
kwonoj: I don't know, reason raised q. For those two visitor, one explicitly skips visit and declared as noop and one are not.

Austaras: Also I should correct myself for what I say about `ThisReplacer`, it is also needed in `ClassMethod` and `MethodProperty`

kwonoj: Either cases I think we can take this discussion out of this pr.

kwonoj: > it is also needed in ClassMethod

This pr is specifically for the classmethod.

kwonoj: https://github.com/swc-project/swc/issues/2786

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
Austaras: Oh, you mean whether to insert `var _this = this` to parent scope or within function? If I'm understanding `babel/helper-wrap-function` correctly, there's a definite list for that we don't need to access parent scope: `class_method` `class_static_property` and `object_property`, we could label them manually

kwonoj: that's interesting. Curious if we have equivalent already? @kdy1 otherwise I guess we may need it as well.

kwonoj: It may makes bit tricky to pass renamed ident across maybe, but have to see it if that's the direction we'll go.

Austaras: Yes, 'this_id: a &'a mut Ident' seems to be a bit weird

kwonoj:
I'd like to proceed this PR in these steps, if this sounds acceptable approach:

1. Resolve immediate questions, notably https://github.com/swc-project/swc/pull/2776#discussion_r750908883. I think it'd be ok to accept immeidate workaround
2. Land #2786, update code to utilize refactored
3. See if we can explore https://github.com/swc-project/swc/pull/2776#discussion_r751519526 , probably need separate issue to track
4. New PR completes remaning work not covered in this PR.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
kdy1: I agree that we need one, [contains_this_expr](https://docs.rs/swc_ecma_utils/0.52.0/swc_ecma_utils/fn.contains_this_expr.html) is a similar one, but it's a visitor.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
kdy1: I think `arguments` replacer should handle member expressions, otherwise `foo.arguments` will be replaced.
It sounds like a bug.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
kdy1: After thinking a lot, I concluded that it's fine because `visit_mut_stmts` of `Actual` will prevent it from being too deep.

kdy1 (changes requested):
It's fine, but I found some code style issues.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -564,6 +629,39 @@ impl VisitMut for MethodFolder {
...
+            Expr::Arrow(ArrowExpr { body, .. }) if contains_this_expr(body) => {
+                if self.scoped_this_ident.is_none() {
```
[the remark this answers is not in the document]
kwonoj: Thx. I choose `get_or_insert` instead of `get_or_insert_with`, as closure doesn't seem required.

kwonoj:
Cool, once this goes in I'll wait for #2786 and apply further progresses on top of it.

kdy1 (approved):
LGTM.

PR #2782, opened 2021-11-17
fix(ecma/parser): allow static/declare to be optional class property 

### LINKED ISSUE #1219: optional class properties with a name of `static` or `declare` are incorrectly disallowed
**Describe the bug**
`static? = 5` is allowed as a ts class property, however, swc does not allow it:
```
error: Unexpected token `?`. Expected identifier, string literal, numeric literal or [ for the computed key
 --> tests/main.ts:2:9
  |
2 |   static?() {}
  |         ^
```
```
error: `declare` modifier cannot appear on a class element
 --> tests/main.ts:2:3
  |
2 |   declare?() {}
  |   ^^^^^^^

error: Unexpected token `?`. Expected identifier, string literal, numeric literal or [ for the computed key
 --> tests/main.ts:2:10
  |
2 |   declare?() {}
  |          ^
```

**Input code**
```ts
class Foo {
  static? = 5;
  declare? = 5;
}
```

**Config**
None

**Expected behavior**
No parser errors

**Version**
The version of @swc/core: `1.2.37`

**Additional context**
It is caused by the [is_class_property](https://github.com/swc-project/swc/blob/master/ecmascript/parser/src/parser/class_and_fn.rs#L820) check, which wont work correctly if the current token is `?` when checking for `declare` and `static` tokens in class_member.

RiESAEX:
fix #1219


kdy1 (approved):
Thanks!

PR #2783, opened 2021-11-17
Add support for helpers in commonjs scripts

### LINKED ISSUE #1814: With isModule: false, emits references to non-emitted helpers
**Describe the bug**

When transforming a module, swc emits definitions for the helpers it references (such as `_typeof`); however, when transforming a script, it references helpers without emitting their definitions. This leads to a `ReferenceError` when the result is run.

**Input code**

```js
> swc.transformSync("typeof Math")  // works
{
  code: 'var _typeof = function(obj) {\n' +
    '    return obj && typeof Symbol !== "undefined" && obj.constructor === Symbol ? "symbol" : typeof obj;\n' +
    '};\n' +
    'typeof Math === "undefined" ? "undefined" : _typeof(Math);\n'
}
> swc.transformSync("typeof Math", {isModule: false})  // fails
{
  code: 'typeof Math === "undefined" ? "undefined" : _typeof(Math);\n'
}
```

**Config**
none

**Expected behavior**
If I’m understanding correctly, swc should emit the same `var _typeof = …` in script mode that it emits in module mode. (Perhaps an IIFE is needed to scope the extra definitions in a script, although swc doesn’t do that for `let` at a script’s top level so I’m not sure.)

**Version**
The version of @swc/core: 1.2.60

sventschui:
- [x] Support for `externalHelpers: false`
- [x] Clean up code
- [ ] Test in real-life project

Fixes #1814

### DIFF
--- a/crates/swc/src/builder.rs
+++ b/crates/swc/src/builder.rs
@@ -270,7 +270,20 @@ impl<'a, 'b, P: swc_ecma_visit::Fold> PassBuilder<'a, 'b, P> {
             ),
             compat::reserved_words::reserved_words(),
             Optional::new(export_namespace_from(), need_interop_analysis),
-            Optional::new(helpers::inject_helpers(), self.inject_helpers),
+            Optional::new(
+                helpers::inject_helpers(match module {
+                    Option::None => swc_ecma_transforms_base::helpers::ModuleType::Es6, /* TODO: is this assumption correct? */
+                    Option::Some(ModuleConfig::CommonJs(_)) =>
+                        swc_ecma_transforms_base::helpers::ModuleType::CommonJs,
+                    Option::Some(ModuleConfig::Umd(_)) =>
+                        swc_ecma_transforms_base::helpers::ModuleType::Umd,
+                    Option::Some(ModuleConfig::Amd(_)) =>
+                        swc_ecma_transforms_base::helpers::ModuleType::Amd,
+                    Option::Some(ModuleConfig::Es6) =>
+                        swc_ecma_transforms_base::helpers::ModuleType::Es6,
+                }),
+                self.inject_helpers
+            ),
             ModuleConfig::build(
                 self.cm.clone(),
                 base_url,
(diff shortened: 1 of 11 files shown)


crates/swc_ecma_transforms_base/src/helpers/mod.rs
```diff
@@ -393,8 +561,32 @@ let _throw1 = null;
...
+                    module_type: ModuleType::CommonJs,
+                })
+            },
+            "let _throw = null",
+            "function _throw(e) {
+    throw e;
+}
+let _throw1 = null;
+",
+            true, // TODO: we have an extra Stmt in `acutal` AST, can this be fixed?
```
sventschui, line 566: @kdy1 is it required to fix this test with `false` passed here? The generated AST has an extra `Stmt` that the expected AST doesn't have.

crates/swc/src/builder.rs
```diff
@@ -270,7 +270,20 @@ impl<'a, 'b, P: swc_ecma_visit::Fold> PassBuilder<'a, 'b, P> {
             ),
             compat::reserved_words::reserved_words(),
             Optional::new(export_namespace_from(), need_interop_analysis),
-            Optional::new(helpers::inject_helpers(), self.inject_helpers),
+            Optional::new(
+                helpers::inject_helpers(match module {
+                    Option::None => swc_ecma_transforms_base::helpers::ModuleType::Es6, /* TODO: is this assumption correct? */
```
sventschui, line 275: @kdy1 is it correct to assume ESM when we do not have a module config?

kdy1: Yes. It's default behavior.

crates/swc_ecma_transforms_base/src/helpers/mod.rs
kdy1: I think it's better to fix it, as it will ensure that the `hygiene` pass does not break injected helpers.

kdy1 (commented):
Mostly LGTM. Thanks!

kdy1:
Closing as it's implemented by https://github.com/swc-project/swc/pull/4025

PR #2787, opened 2021-11-17
Refactor parser and lexer

alexander-akait:
@kdy1 Very important fix, because we handle whitespaces incorrectly, look at example:
```css
a { 
  color: /* comment */ /*comment*/ red;
        ^             ^           ^
          /* whitespace tokens */
}
```

Before we eat whitespaces tokens between comments, so tokens in recovery mode was wrong, after this PR now we have valid count of whitespace tokens.

I remove `skip_whitespaces`, we should not rely on whitespaces inside file, lexer generates whitespace tokens, tokens input - already have whitespace tokens inside, also this logic is invalid, because whitespaces can be between comments.

Also we have broken `span` in some cases because they did not take into account whitespaces and comments at the beginning and at the end of the file, so `span` for `Stylesheet` has incorrect values, but `Stylesheet` should represent the entire file, not where the code start.

Also skipping whitespaces for peeked value is wrong, because there are situations when whitespaces token followed by whitespace token (between comments) and rely on whitespace characters in code is wrong

Also we don't need `is_eof`, because we have macro.

Also added `TODO` for future `span` fixes (because we get span before whitespaces in some places, but it is wrong).

Example:
```
div /* test */ > a { color: red; }
```

In this example comment should be part of complex selector.

### DIFF
--- a/crates/swc_css_parser/src/lexer/mod.rs
+++ b/crates/swc_css_parser/src/lexer/mod.rs
@@ -58,16 +58,6 @@ where
             .map_err(|kind| Error::new(span, kind))
     }
 
-    fn skip_whitespaces(&mut self) -> PResult<()> {
-        let start = self.input.cur_pos();
-
-        let token = self.skip_ws();
-        let end = self.input.cur_pos();
-        let span = Span::new(start, end, Default::default());
-
-        token.map_err(|kind| Error::new(span, kind))
-    }
-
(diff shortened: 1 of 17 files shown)


kdy1 (dismissed):
LGTM. Thanks!

alexander-akait:
@kdy1 Rebased, ready for review again

kdy1 (approved):
Thanks!

PR #2791, opened 2021-11-18
fix(es/transforms/compat): Optioanl call function call result with param

Austaras:


Austaras:
The code of optional chaining is a bit confusing, maybe it could be refactored into more like babel's implementation, but that would require 2 or 3 loop instead of one

kdy1 (approved):
Yeah, I agree. I think 2/3 loops would be fine.
But thanks! This changes LGTM.

PR #2792, opened 2021-11-18
fix(es/utils): Fix `collect_decls`

### LINKED ISSUE #2785: `swc_ecma_utils::BindingCollector` does not handle array, rest, and object binding patterns
### Describe the bug

For example, if you have a variable declaration like the following:

```ts
const { test } = { test: 5 };
```

The result of `swc_ecma_utils::collect_decls` will be empty. I believe it should contain this declaration?

See here:

https://github.com/swc-project/swc/blob/e5604c3e56e8ebdc96d4a15e1610f5dbcae09876/crates/swc_ecma_utils/src/lib.rs#L2244-L2249

### Expected behavior

I believe it should return results for these types of declarations

### Version

latest main

kdy1:
swc_ecma_utils:
 - `collect_decls`: Handle assignment pattern property. (Closes #2785)

### DIFF
--- a/crates/swc_ecma_utils/src/lib.rs
+++ b/crates/swc_ecma_utils/src/lib.rs
@@ -2165,6 +2165,14 @@ where
 {
     noop_visit_type!();
 
+    fn visit_assign_pat_prop(&mut self, node: &AssignPatProp, _: &dyn Node) {
+        node.value.visit_with(node, self);
+
+        if self.is_pat_decl {
+            self.add(&node.key);
+        }
+    }
+
     fn visit_class_decl(&mut self, node: &ClassDecl, _: &dyn Node) {
         node.visit_children_with(self);
 
@@ -2246,14 +2254,6 @@ where
                 Pat::Ident(i) => self.add(&i.id),
                 _ => {}
             }
-        }
-    }
-
-    fn visit_assign_pat_prop(&mut self, node: &AssignPatProp, _: &dyn Node) {
-        node.value.visit_with(node, self);
-
-        if self.is_pat_decl {
-            self.add(&node.key);
         }
     }
 
--- a/crates/swc_ecma_utils/src/lib.rs
+++ b/crates/swc_ecma_utils/src/lib.rs
@@ -2249,7 +2249,7 @@ where
         }
     }
 
-    fn visit_object_pat_prop(&mut self, node: &ObjectPatProp, _: &dyn Node) {
+    fn visit_assign_pat_prop(&mut self, node: &AssignPatProp, _: &dyn Node) {
         node.value.visit_with(node, self);
 
         if self.is_pat_decl {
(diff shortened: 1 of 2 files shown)


crates/swc_ecma_utils/src/lib.rs
```diff
@@ -2165,6 +2165,14 @@ where
 {
     noop_visit_type!();
 
+    fn visit_assign_pat_prop(&mut self, node: &AssignPatProp, _: &dyn Node) {
```
dsherret, line 2168: This does not fix the issue completely. I will open a PR.

kdy1: Oh, thanks!
I thought it was enough.

dsherret: I think we need to handle `const [ test ] = asdf;`. I'm just writing a test to confirm it.

kdy1: [playground link](https://play.swc.rs/?version=1.2.106&code=H4sIAAAAAAAAA%2BNKzs8rLlGIVkjLz1eIVbBVqK4FALEg6qoTAAAA&config=H4sIAAAAAAAAA0WNSwrDMAxE76J1ts0id%2BghjKsEF%2F%2FQKBBjfPfawSU7aebpqdIXlrZK2QhYxoQS1Vy0kZbMsOKy0kKKHu3Gg1tfjBysHWG8eudTAs92oeCi28sw2RSyMPBUJh7%2BT7YuCulzjqDez27hSu1xzDuH9wRVTm4%2FOkwQUrUAAAA%3D)


You can change to AST mode to see how array patterns are consturcted.


```json
{
  "type": "Module",
  "span": {
    "start": 2633,
    "end": 2651,
    "ctxt": 0
  },
  "body": [
    {
      "type": "VariableDeclaration",
      "span": {
        "start": 2633,
        "end": 2651,
        "ctxt": 0
      },
      "kind": "const",
      "declare": false,
      "declarations": [
        {
          "type": "VariableDeclarator",
          "span": {
            "start": 2639,
            "end": 2651,
            "ctxt": 0
          },
          "id": {
            "type": "ArrayPattern",
            "span": {
              "start": 2639,
              "end": 2646,
              "ctxt": 0
            },
            "elements": [
              {
                "type": "Identifier",
                "span": {
                  "start": 2641,
                  "end": 2644,
                  "ctxt": 0
                },
                "value": "foo",
                "optional": false,
                "typeAnnotation": null
              }
            ],
            "optional": false,
            "typeAnnotation": null
          },
          "init": {
            "type": "ObjectExpression",
            "span": {
              "start": 2649,
              "end": 2651,
              "ctxt": 0
            },
            "properties": []
          },
          "definite": false
        }
      ]
    }
  ],
  "interpreter": null
}
```

dsherret: Thanks @kdy1! It is actually working. My bad! I misread the code—ArrayPatterns have elements that are `Pat` and so it will visit those identifiers.

I'm going to submit my tests in a PR though 🙂

kdy1: Thanks!

PR #2798, opened 2021-11-18
fix: extract_var_ids should handle object patterns

dsherret:
Adds a few tests for collect_decls and then fixes an issue with extract_var_ids

### DIFF
--- a/crates/swc_ecma_utils/Cargo.toml
+++ b/crates/swc_ecma_utils/Cargo.toml
@@ -26,4 +26,4 @@ unicode-xid = "0.2"
 
 [dev-dependencies]
 testing = {version = "0.15.0", path = "../testing"}
-swc_ecma_parser = {path = "../swc_ecma_parser"}
+swc_ecma_parser = {version = "0.78.4", path = "../swc_ecma_parser"}
--- a/crates/swc_ecma_utils/Cargo.toml
+++ b/crates/swc_ecma_utils/Cargo.toml
@@ -26,3 +26,4 @@ unicode-xid = "0.2"
 
 [dev-dependencies]
 testing = {version = "0.15.0", path = "../testing"}
+swc_ecma_parser = {path = "../swc_ecma_parser"}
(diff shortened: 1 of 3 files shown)


crates/swc_ecma_utils/src/lib.rs
```diff
@@ -411,6 +411,12 @@ pub struct Hoister {
 impl Visit for Hoister {
     noop_visit_type!();
 
+    fn visit_assign_pat_prop(&mut self, node: &AssignPatProp, _: &dyn Node) {
+        node.value.visit_with(node, self);
+
+        self.vars.push(node.key.clone());
```
dsherret, line 417: Does this `Hoister` need to do the `is_pat_decl` stuff that `BindingCollector` is doing?

dsherret:
Oops... I somehow updated a submodule

crates/swc_ecma_utils/Cargo.toml
```diff
@@ -26,3 +26,4 @@ unicode-xid = "0.2"
 
 [dev-dependencies]
 testing = {version = "0.15.0", path = "../testing"}
+swc_ecma_parser = {path = "../swc_ecma_parser"}
```
kdy1, line 29: Can you specify the version?
Otherwise `cargo`will refuse to publish.

crates/swc_ecma_utils/src/lib.rs
kdy1: I don' think so because `visit_pat` does not check if it's decl.

kdy1 (changes requested):
Thank you!
I found an issue that can block publishing.

crates/swc_ecma_utils/Cargo.toml
dsherret: Oh, I thought it would be ok because it's a dev dependency. I will update it.

dsherret: Fixed in a724e83efc275e72f47a024973d1ee948169af32.

kdy1 (approved):
Thanks!

PR #2806, opened 2021-11-19
fix(es/transforms/compat): do not explicitly bind this for fn

kwonoj:
Part 2 for #2716,

This PR progresses #2716, removes explicit binding to this for plain fn. There is a new test case (https://github.com/swc-project/swc/commit/3be8642765caa1fca22351d0d1318ae95e80b8d3#) to verify this behavior change correctly binds this context for the functions.


kdy1 (approved):
Thanks!

PR #2812, opened 2021-11-19
refactor(es/transforms/compat): Lift this replacer out of arrow

### LINKED ISSUE #2786: [Refactor] (es/transforms/compat): consolidate replacer visitor for arrow / fns
### Describe the feature

Context: https://github.com/swc-project/swc/pull/2776#discussion_r751427342

With effort for the #2716, there seems some overlapping between few visitors to handle replacing context specific (`this`, etcs). It maybe possible we could reuse instead of repeating same steps.

Probably reusing `ThisReplacer`, and moreover potentially make a single visitor instead fo `This` + `Arguments`. But it requires proper audit to do so.

_No response_

### Additional context

_No response_

Austaras:
Here is my proposed API, it might seem to be a bit weird, and it definitely need a better name.

should close #2786


kwonoj:
async_to_generator currently have thisreplacer as well which is copy paste from this. probably it need to be updated as well?

Austaras:
> async_to_generator currently have thisreplacer as well which is copy paste from this. probably it need to be updated as well?

Updated. The first commit is to showcase the api

crates/swc_ecma_utils/src/function.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+impl<'a> VisitMut for FunctionWrapper<'a> {
+    noop_visit_mut_type!();
+
+    fn visit_mut_expr(&mut self, e: &mut Expr) {
+        match e {
+            Expr::Ident(id) if id.sym == js_word!("arguments") => {
+                let arguments = self.state.args.get_or_insert(private_ident!("_arguments"));
```
kdy1, line 79: This should be `get_or_insert_with(|| private_ident!("_arguments"))`. Otherwise, global lock will be acquired every time.

crates/swc_ecma_utils/src/function.rs
```diff
@@ -0,0 +1,123 @@
...
+                *e = Expr::Ident(arguments.clone());
+            }
+            Expr::This(..) => {
+                let this = self.state.this.get_or_insert(private_ident!("_this"));
```
kdy1, line 83: Same here

crates/swc_ecma_transforms_compat/src/es2015/arrow.rs
```diff
@@ -83,6 +83,7 @@ impl VisitMut for Arrow {
                 ..
             }) => {
                 params.visit_mut_with(self);
+                params.visit_mut_with(&mut FunctionWrapper::new(&mut self.state));
```
kdy1, line 86: Isn't this exponential?

kdy1: Nevermind. `n` will be super small.

crates/swc_ecma_utils/src/function.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /// Don't recurse into constructor
+    fn visit_mut_class(&mut self, _: &mut Class) {}
+
+    /// Don't recurse into fn
+    fn visit_mut_function(&mut self, _: &mut Function) {}
+
+    /// Don't recurse into getter/setter/method except computed key
+    fn visit_mut_getter_prop(&mut self, p: &mut GetterProp) {
```
kdy1, line 106: I think you can implement `visit_mut_prop_name` which only visits children if it's `PropName::Computed`.

crates/swc_ecma_transforms_compat/src/es2015/arrow.rs
```diff
@@ -94,56 +95,9 @@ impl VisitMut for Arrow {
...
-                        span: DUMMY_SP,
-                        name: Pat::Ident(id.into()),
-                        init: Some(Box::new(Expr::Ident(Ident::new(
-                            js_word!("arguments"),
-                            DUMMY_SP,
-                        )))),
-                        definite: false,
-                    });
-                }
+                body.visit_mut_with(&mut FunctionWrapper::new(&mut self.state));
```
kdy1, line 100: This has exponential time complexity, but it's fine for most case because `n` is super small. I think it's better to have a comment mentioning it.

crates/swc_ecma_transforms_compat/src/es2015/arrow.rs
```diff
@@ -178,127 +132,34 @@ impl VisitMut for Arrow {
     }
 
     fn visit_mut_function(&mut self, f: &mut Function) {
```
kdy1, line 134: This can be removed, as `f.visit_mut_children_with(self);` is default behavior.

kdy1 (changes requested):
I found some small issues.

crates/swc_ecma_utils/src/function.rs
Austaras: This is new to me

crates/swc_ecma_utils/src/function.rs
Austaras: No I can't, since I want args and function body in getter/setter/method be ignored

crates/swc_ecma_transforms_compat/src/es2015/arrow.rs
Austaras: Could you explain why this is exponential?

kdy1: I was confused, so I made a benchmark that only profiles nested visitor.
You can see the content of benchmark at https://github.com/swc-project/swc/pull/2820/files and


```
    Finished bench [optimized + debuginfo] target(s) in 1m 20s
     Running unittests (/Users/kdy1/projects/s/issues/target/release/deps/time_complexity-81379e7c22c61e80)

running 6 tests
test time_10 ... bench:         238 ns/iter (+/- 33)
test time_15 ... bench:         480 ns/iter (+/- 27)
test time_20 ... bench:         874 ns/iter (+/- 29)
test time_40 ... bench:       5,086 ns/iter (+/- 212)
test time_5  ... bench:          71 ns/iter (+/- 3)
test time_60 ... bench:      13,085 ns/iter (+/- 480)

test result: ok. 0 passed; 0 failed; 0 ignored; 6 measured; 0 filtered out; finished in 2.80s
```


This is what I get.

Austaras: I see, maybe I could figure out a way to avoid this

Austaras: Wait, wouldn't swc translate the innermost arrow expression first, and every time a `FunctionWrapper` is started, it will be stopped to visit inner arrow by `FnExpr`?

Austaras: See the new test in the new commit, only 6 expressions is printed, I guess it's fine

kdy1: The number it printed does not matter.
At the benchmark above I posted, `visit count` is also tested.
It's exponential because it's deeply nested, not because a node is visited exponential times.


> Wait, wouldn't swc translate the innermost arrow expression first, and every time a FunctionWrapper is started, it will be stopped to visit inner arrow by FnExpr?

About this, I think it will be fine for this line, but not for `body` below
https://github.com/swc-project/swc/blob/9221b0744bc921957bd9a491f1499f55fa5c9443/crates/swc_ecma_transforms_compat/src/es2015/arrow.rs#L100

kdy1:
Changes look fine to me.
Can you ping me or request review after fixing the failing test case?

Austaras:
Sure, I have some IRL issues to handle, it may takes a day or two

crates/swc_ecma_transforms_compat/src/es2015/arrow.rs
Austaras: Ok, let's move this lengthy discussion to github discussion

kdy1 (approved):
LGTM. Thanks!

Issue #2809, opened 2021-11-19
TypeScript emits wrong code on string enums when one member references another


mernen:
### Describe the bug

TypeScript consider [string enums](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) to be enums where every value is either a string literal *or* a reference to another string enum member, like the code below.

### Input code

```typescript
enum Color {
    Aqua = '#00ffff',
    Cyan = Aqua,
}
```

### Playground link

https://play.swc.rs/?version=1.2.111&code=H4sIAAAAAAAAA0vNK81VcM7PyS9SqOZSAALHwtJEBVsFdWUDgzQgUNcBizpXJuYBRUGSOly1XAD2dZgDNgAAAA%3D%3D&config=H4sIAAAAAAAAA0WNSwrDMAxE76J1ts0id%2BghjKsEF%2F%2FQKBBjfPfawSU7aebpqdIXlrZK2QhYxoQS1Vy0kZbMsOKy0kKKHu3Gg1tfjBysHWG8eudTAs92oeCi28sw2RSyMPBUJh7%2BT7YuCulzjqDez27hSu1xzDuH9wRVTm4%2FOkwQUrUAAAA%3D

### Expected behavior

The value of `Cyan` should be evaluated at compile time to the same of member `Aqua`; instead, it's being treated as a variable reference.

Expected to be the same as `tsc` output:

```javascript
var Color;
(function (Color) {
    Color["Aqua"] = "#00ffff";
    Color["Cyan"] = "#00ffff";
})(Color || (Color = {}));
```

### Version

1.2.111

### Additional context

Current `swc` output:

```javascript
var Color1;
(function(Color) {
    Color["Aqua"] = '#00ffff';
    Color["Cyan"] = Aqua;
})(Color1 || (Color1 = {
}));
```

PR #2816, opened 2021-11-20
fix(es/transforms/typescript): allow aliases on string enums (#2809)

mernen:
A fix for #2809.

This is my first time with this codebase! Please report if I got something wrong.

---

Clsoes #2809


crates/swc_ecma_transforms_typescript/src/strip.rs
```diff
@@ -1042,9 +1042,11 @@ where
                                         right: if rhs_should_be_name {
                                             Box::new(Expr::Lit(Lit::Str(value.clone())))
                                         } else {
-                                            m.init.unwrap_or_else(|| {
+                                            if m.init.is_some() {
+                                                Box::new(val)
+                                            } else {
                                                 Box::new(Expr::Lit(Lit::Str(value.clone())))
-                                            })
+                                            }
```
mernen, line 1049: I am *pretty* sure this test is unnecessary for *valid* inputs.

Only situation I can think of that would reach the `else` case is code like:

```typescript
enum Test {
    Foo = '?',
    Bar,
}
```

Which violates the rule about string enums being all initialized. `tsc` produces a warning "Enum member must have initializer" and the following output:

```javascript
var Test;
(function (Test) {
    Test["Foo"] = "?";
    Test[Test["Bar"] = void 0] = "Bar";
})(Test || (Test = {}));
```

While `swc` (with this patch or not) produces:

```javascript
var Test;
(function (Test) {
    Test["Foo"] = "?";
    Test[Test["Bar"] = 0] = "Bar";
})(Test || (Test = {}));
```

If we remove this condition and always return `Box::new(val)`, the result will be:

```javascript
var Test;
(function (Test) {
    Test["Foo"] = "?";
    Test[Test["Bar"] = 0] = 0;
})(Test || (Test = {}));
```

Which I think one could be argue is no worse than any of the others.

kdy1: > tsc produces a warning "Enum member must have initializer" and the following output:

If it's warning, `tsc` will generate js code. This branching is cheap enough to be justified IMO.

kdy1 (approved):
Thank you!

crates/swc_ecma_transforms_typescript/src/strip.rs
kdy1: > tsc produces a warning "Enum member must have initializer" and the following output:

If it's warning, `tsc` will generate js code. This branching is cheap enough to be justified IMO.

PR #2817, opened 2021-11-20
refactor(es/transforms/compat): visit_mut for block_scoping

kwonoj:
relates to #2370

This PR refactors block_scoping to use visit_mut instead of fold. Same as other work, this is straightforward conversion.


kdy1 (approved):
Thanks!

PR #2839, opened 2021-11-22
Refactor lexer

alexander-akait:
@kdy1 Refactor lexer part 4 (penultimate)

Done:
- move logic from `cur` on `next` (as the spec is written)
- resolve todos
- bug fix with spaces around `URL` token

In spec we have https://www.w3.org/TR/css-syntax-3/#tokenizer-definitions
```
next input code point
The first code point in the input stream that has not yet been consumed.

current input code point
The last code point to have been consumed.

reconsume the current input code point
Push the current input code point back onto the front of the input stream, so that the next time you are instructed to consume the next input code point, it will instead reconsume the current input code point.
```

also spec uses `Consume the next input code point.`

So we have:
- `self.cur()`
- `self.next()` (and `self.next_next()`, and `self.next_next_next()`)
- `self.consume()`
- `self.reconsume()`

Also in spec we have https://www.w3.org/TR/css-syntax-3/#starts-with-a-valid-escape, https://www.w3.org/TR/css-syntax-3/#would-start-an-identifier, https://www.w3.org/TR/css-syntax-3/#starts-with-a-number
```
The algorithm described here can be called explicitly with two code points, or can be called with the input stream itself. In the latter case, the two code points in question are the current input code point and the next input code point, in that order.
```

So now we use `self.cur()` and `self.next()` without lookup ahead


crates/swc_css_parser/src/lexer/mod.rs
```diff
@@ -267,31 +288,25 @@ where
...
                 return Ok(tok!(";"));
             }
             // U+003C LESS-THAN SIGN (<)
             Some(c) if c == '<' => {
-                self.input.bump();
-
                 // If the next 3 input code points are U+0021 EXCLAMATION MARK U+002D
                 // HYPHEN-MINUS U+002D HYPHEN-MINUS (!--), consume them and return a
                 // <CDO-token>.
-                if self.input.is_byte(b'!')
```
kdy1, line 288: Actually `is_byte` is an optimization and it's much faster than comparing `char`.

crates/swc_css_parser/src/lexer/mod.rs
```diff
@@ -1007,19 +996,19 @@ where
...
         // Look at the first code point:
-        let first = maybe_first.or(self.input.cur());
+        let first = maybe_first.or(self.cur());
 
         match first {
             // U+002B PLUS SIGN (+)
             // U+002D HYPHEN-MINUS (-)
             Some('+') | Some('-') => {
-                match maybe_second.or(self.input.peek()) {
+                match maybe_second.or(self.next()) {
```
kdy1, line 1005: I think this will call `self.next()` unconditionally.

crates/swc_css_parser/src/lexer/mod.rs
```diff
@@ -1007,19 +996,19 @@ where
         maybe_third: Option<char>,
     ) -> LexResult<bool> {
         // Look at the first code point:
-        let first = maybe_first.or(self.input.cur());
+        let first = maybe_first.or(self.cur());
```
kdy1, line 999: I think this will call `self.next()` unconditionally.

crates/swc_css_parser/src/lexer/mod.rs
```diff
@@ -119,10 +154,8 @@ where
             }
             // U+0023 NUMBER SIGN (#)
             Some(c) if c == '#' => {
```
kdy1, line 156: I found another issue.
It should be `Some('#')`, not `Some(c) if c == '#'`.

But this is not related to this PR, so let's skip it.

crates/swc_css_parser/src/lexer/mod.rs
```diff
@@ -948,12 +937,12 @@ where
...
     ) -> LexResult<bool> {
         // Look at the first code point:
-        let first = maybe_first.or(self.input.cur());
+        let first = maybe_first.or(self.cur());
 
         match first {
             // U+002D HYPHEN-MINUS
             Some('-') => {
-                let second = maybe_second.or(self.input.peek());
+                let second = maybe_second.or(self.next());
```
kdy1, line 945: Seems like an inefficeint utf8 decoding.

crates/swc_css_parser/src/lexer/mod.rs
```diff
@@ -965,7 +954,7 @@ where
                     // or the second and third code points are a valid escape
                     // return true.
                     Some(_) => {
-                        let third = maybe_third.or(self.input.peek_ahead());
+                        let third = maybe_third.or(self.next_next());
```
kdy1, line 957: Seems like an inefficeint utf8 decoding.

crates/swc_css_parser/src/lexer/mod.rs
```diff
@@ -984,7 +973,7 @@ where
             // If the first and second code points are a valid escape, return true. Otherwise,
             // return false.
             Some('\\') => {
-                let second = self.input.peek();
+                let second = maybe_second.or(self.next());
```
kdy1, line 976: Seems like an inefficeint utf8 decoding.

crates/swc_css_parser/src/lexer/mod.rs
```diff
@@ -1007,19 +996,19 @@ where
...
             Some('+') | Some('-') => {
-                match maybe_second.or(self.input.peek()) {
+                match maybe_second.or(self.next()) {
                     // If the second code point is a digit, return true.
                     Some(second) if second.is_digit(10) => return Ok(true),
                     // Otherwise, if the second code point is a U+002E FULL STOP (.) and the
                     // third code point is a digit, return true.
                     Some('.') => {
-                        if let Some(third) = maybe_third.or(self.input.peek_ahead()) {
+                        if let Some(third) = maybe_third.or(self.next_next()) {
```
kdy1, line 1011: Seems like an inefficeint utf8 decoding.

kdy1 (changes requested):
I have some concerns.

crates/swc_css_parser/src/lexer/mod.rs
alexander-akait: @kdy1 Can you clarify about `inefficeint utf8 decoding`?

crates/swc_css_parser/src/lexer/mod.rs
alexander-akait: Good catch I will fix it

crates/swc_css_parser/src/lexer/mod.rs
alexander-akait: @kdy1 Why? We should look here on currrent and the next if no options were provided, from spec:

> This section describes how to check if three code points would start an identifier. The algorithm described here can be called explicitly with three code points, or can be called with the input stream itself. In the latter case, the three code points in question **are the current input code point and the next two input code points**, in that order.

crates/swc_css_parser/src/lexer/mod.rs
alexander-akait: We rewrite logic from looking ahead on current looking as in spec, i.e. we use
- `self.cur()` 
- `self.next()`
- `self.next_next()`

https://www.w3.org/TR/css-syntax-3/#starts-with-a-number

> This section describes how to check if three code points would start a number. The algorithm described here can be called explicitly with three code points, or can be called with the input stream itself. In the latter case, the three code points in question are the current input code point and the next two input code points, in that order.

Same for other places

crates/swc_css_parser/src/lexer/mod.rs
kdy1: I meant  `.or` should be changed to `.or_else` to avoid useless utf8 decoding for cases where `first` is `Some`.

But nevermind, I found that original source code was the same.  It can be handled with another PR.

crates/swc_css_parser/src/lexer/mod.rs
alexander-akait: It was one place where we used `is_byte`, but if it is faster we can rewrite `self.cur()` and `self.next()`/etc on bytes, I tried it right now, I don't see any perf, maybe I missed something?

crates/swc_css_parser/src/lexer/mod.rs
alexander-akait: Do you mean use something like:
```rust
match self.cur() {
  Some(ch) => ch == c as char,
  _ => false,
}
```

If yes we can rewrite on it, but before we don't use, maybe we can postpone it to the next PR, it is easy, but I wan't to mix it

crates/swc_css_parser/src/lexer/mod.rs
alexander-akait: Ah, I see, let's fix it, I can fix it here or we can postpone it to the next PR

alexander-akait:
Fixed all `inefficeint utf8 decoding`

crates/swc_css_parser/src/lexer/mod.rs
alexander-akait: Got it and fixed

kdy1 (approved):
LGTM.
I'll merge this after publishing a new version of `@swc/core`.

Issue #2275, opened 2021-09-20
failed to resolve fs when using electron or node targets


schontz:
**Describe the bug**

When attempting to run `npx spack` on an electron project, the following error is encountered:

```
(node:84253) UnhandledPromiseRejectionWarning: Error: load_transformed failed

Caused by:
    0: failed to analyze module
    1: failed to resolve fs from /path/to/node_modules/electron/index.js
    2: index not found
```

**Config**

**.swcrc**

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "decorators": false,
      "dynamicImport": false
    },
    "transform": null,
    "target": "es2019",
    "loose": false,
    "externalHelpers": false,
    "keepClassNames": true
  },
  "env": {
    "targets": {
      "electron": 14
    }
  },
  "module": {
    "type": "commonjs",
    "strict": false,
    "strictMode": true,
    "lazy": false,
    "noInterop": false
  }
}
```

**spack.config.js**

```js
const { config } = require('@swc/core/spack');
const packageJson = require('./package.json');

// Electron copies over all dependencies during build, so exclude them from bundling.
const PACKAGE_DEPENDENCIES = Object.keys(packageJson.dependencies).map((pkg) => pkg);

module.exports = config({
  externalModules: PACKAGE_DEPENDENCIES,
  entry: {
    main: __dirname + '/src/main.ts',
  },
  output: {
    path: __dirname + '/swc',
  },
});
```

I have also tried `"targets": { "node": 14 }`, but it produces the same error.

**Expected behavior**

Expect swc to ignore node native modules.

**Version**
The version of @swc/core: 1.2.88

PR #2840, opened 2021-11-22
[swc_bundler] Ignore external require

ErDmKo:
May be this fix will help with https://github.com/swc-project/swc/issues/2275. But it is for shure helps me

### DIFF
--- a/crates/swc_bundler/src/bundler/mod.rs
+++ b/crates/swc_bundler/src/bundler/mod.rs
@@ -131,7 +131,7 @@ where
             }
         })
     }
-    pub fn is_external(&self, src: &JsWord) -> bool {
+    pub(crate) fn is_external(&self, src: &JsWord) -> bool {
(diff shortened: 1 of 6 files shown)


crates/swc_bundler/src/bundler/mod.rs
```diff
@@ -131,7 +131,9 @@ where
             }
         })
     }
-
+    pub fn is_external(&self, src: &JsWord) -> bool {
```
kdy1, line 134: Can you reduce the visibility of this method to `pub(crate)`?

kdy1 (changes requested):
Almost LGTM. Thanks!

crates/swc_bundler/src/bundler/mod.rs
ErDmKo: done

kdy1 (approved):
Thank you!

PR #2847, opened 2021-11-23
Refactor parser

alexander-akait:
Just small refactor to avoid logic duplication

Ideally we need migrate everywhere on `self.parse()?` to avoid code duplication, but it requires more refactor, put it on my roadmap


kdy1 (approved):
LGTM. Thanks!

PR #2848, opened 2021-11-23
fix(css/parser): keyframes ast

alexander-akait:
Small fix, `Str` should be `Str`, not `Ident` to be align with token, because

> \<keyframes-name\> = \<custom-ident\> | \<string\>

https://drafts.csswg.org/css-animations/#keyframes

Also use `name` instead `id`, because in spec we have `keyframes-name`


alexander-akait:
@kdy1 friendly ping

kdy1 (approved):
Thanks!

PR #2849, opened 2021-11-23
feat(css/parser): Better recovery mode

alexander-akait:
Better recovery mode, we don't need `is_in_delimited_value`.

To be honestly we don't need `Ctx` (but it requires more refactor too), every time when you want to use `Ctx`, it means you missed something from spec, because all css can be parsed using https://www.w3.org/TR/css-syntax-3/#parsing

`Ctx` is only valid to parse non standard js, but right now we support `// comment` only, also we want to support Nesting selectors (we already parse some simple cases, but we need more tests and fixes)

### DIFF
--- a/crates/swc_css_parser/src/parser/mod.rs
+++ b/crates/swc_css_parser/src/parser/mod.rs
@@ -38,8 +38,6 @@ struct Ctx {
 
     disallow_comma_in_media_query: bool,
 
-    is_in_delimited_value: bool,
-
     allow_at_selector: bool,
 
     recover_from_property_value: bool,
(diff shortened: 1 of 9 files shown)


alexander-akait:
@kdy1 Rebased

kdy1:
Which one should be reviewed/merged first?

alexander-akait:
@kdy1 Better to merge it https://github.com/swc-project/swc/pull/2897 firstly

alexander-akait:
@kdy1 Ready for review :star:

kdy1 (approved):
LGTM, thanks!

PR #2857, opened 2021-11-24
refactor(es/transforms/compat): visit_mut for computed_props

kwonoj:
relates to #2370

This PR refactors computed_props to use visit_mut instead of fold. Mostly straightfoward conversion, except assiging expr when props_cnt ==1 since move occurs inside of temporal move due to drain - moved out assignment out of the loop.


kdy1 (approved):
LGTM. Thanks!

PR #2868, opened 2021-11-24
fix(es/transforms/module): fix helper injection detection across import order

### LINKED ISSUE #2856: `_interopRequireWildcard` is injected without definition
### Describe the bug

When `import React from 'react'` is used, and also a `{...props}` with a `key={}` in JSX, `_interopRequireWildcard` is injected at the top of the TSX file while not being initialized. 

This create the error `ReferenceError: _interopRequireWildcard is not defined`

### Input code

```typescript
import React from 'react';

export const element = (
  <div {...innerTbodyProps} key={true ? 1 : 2} />
);
```

### Config

For TS:

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "transform": {
      "react": {
        "runtime": "automatic"
      }
    }
  },
  "module": {
    "type": "commonjs"
  }
}
```

For JS:

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": true
    },
    "transform": {
      "react": {
        "runtime": "automatic"
      }
    }
  },
  "module": {
    "type": "commonjs"
  }
}
```

Both have the same issue

### Output code

```tsx
"use strict";
    value: true
});
exports.element = void 0;
var _react = _interopRequireWildcard(require("react"));
    if (key in obj) {
            value: value,
            enumerable: true,
            configurable: true,
            writable: true
        });
    } else {
        obj[key] = value;
    }
    return obj;
}
function _interopRequireDefault(obj) {
    return obj && obj.__esModule ? obj : {
        default: obj
    };
}
function _objectSpread(target) {
    for(var i = 1; i < arguments.length; i++){
        var source = arguments[i] != null ? arguments[i] : {
        };
        var ownKeys = Object.keys(source);
        if (typeof Object.getOwnPropertySymbols === "function") {
            ownKeys = ownKeys.concat(Object.getOwnPropertySymbols(source).filter(function(sym) {
                return Object.getOwnPropertyDescriptor(source, sym).enumerable;
            }));
        }
        ownKeys.forEach(function(key) {
            _defineProperty(target, key, source[key]);
        });
    }
    return target;
}
const element = /*#__PURE__*/ (0, _react).createElement("div", _objectSpread({
}, innerTbodyProps, {
    key: true ? 1 : 2
}));
exports.element = element;
```

### Playground link

_No response_

### Expected behavior

it should haved inject `_interopRequireWildcard`, just like if `import * as React from 'react'` was used, so something like that:

```tsx
"use strict";
    value: true
});
exports.element = void 0;
var React = _interopRequireWildcard(require("react"));
    if (key in obj) {
            value: value,
            enumerable: true,
            configurable: true,
            writable: true
        });
    } else {
        obj[key] = value;
    }
    return obj;
}
    if (obj && obj.__esModule) {
        return obj;
    } else {
        var newObj = {
        };
        if (obj != null) {
            for(var key in obj){
                    };
                    if (desc.get || desc.set) {
                    } else {
                        newObj[key] = obj[key];
                    }
                }
            }
        }
        newObj.default = obj;
        return newObj;
    }
}
function _objectSpread(target) {
    for(var i = 1; i < arguments.length; i++){
        var source = arguments[i] != null ? arguments[i] : {
        };
        var ownKeys = Object.keys(source);
        if (typeof Object.getOwnPropertySymbols === "function") {
            ownKeys = ownKeys.concat(Object.getOwnPropertySymbols(source).filter(function(sym) {
                return Object.getOwnPropertyDescriptor(source, sym).enumerable;
            }));
        }
        ownKeys.forEach(function(key) {
            _defineProperty(target, key, source[key]);
        });
    }
    return target;
}
const element = /*#__PURE__*/ (0, React).createElement("div", _objectSpread({
}, innerTbodyProps, {
    key: true ? 1 : 2
}));
exports.element = element;
```

### Version

1.2.112

### Additional context

I tried to use the swc playground to share a link, but it wasn't working. So here is a screenshot proving what I'm saying:

image

kwonoj:
- closes #2856

As commented in issue, crux is depends on import order import_analysis does not flag wildcard interop need to be used correctly. PR includes a scoped property to contain import sets to determine which interop is requred regardless of import order.

### DIFF
--- a/crates/swc_ecma_transforms_module/src/import_analysis.rs
+++ b/crates/swc_ecma_transforms_module/src/import_analysis.rs
@@ -1,6 +1,6 @@
 use super::util::Scope;
-use std::{cell::RefCell, rc::Rc};
-use swc_atoms::js_word;
+use std::{cell::RefCell, collections::HashSet, rc::Rc};
+use swc_atoms::{js_word, JsWord};
 use swc_common::DUMMY_SP;
 use swc_ecma_ast::*;
 use swc_ecma_transforms_base::enable_helper;
@@ -9,11 +9,15 @@ use swc_ecma_visit::{
 };
 
 pub fn import_analyzer(scope: Rc<RefCell<Scope>>) -> impl Fold + VisitMut {
-    as_folder(ImportAnalyzer { scope })
+    as_folder(ImportAnalyzer {
+        scope,
+        import_srcs: HashSet::new(),
+    })
 }
 
 pub struct ImportAnalyzer {
     scope: Rc<RefCell<Scope>>,
+    import_srcs: HashSet<JsWord>,
 }
 
 /// Inject required helpers methods **for** module transform passes.
@@ -97,6 +101,7 @@ impl Visit for ImportAnalyzer {
             }
         } else {
             let mut has_non_default = false;
+
             for s in &import.specifiers {
                 match *s {
                     ImportSpecifier::Namespace(ref _ns) => {
@@ -105,10 +110,10 @@ impl Visit for ImportAnalyzer {
                         }
                     }
                     ImportSpecifier::Default(_) => {
-                        scope
-                            .import_types
-                            .entry(import.src.value.clone())
-                            .or_insert(false);
+                        let src = import.src.value.clone();
+                        let src_already_exist = self.import_srcs.contains(&src);
+
+                        scope.import_types.entry(src).or_insert(src_already_exist);
                     }
                     ImportSpecifier::Named(ref i) => {
                         let ImportNamedSpecifier {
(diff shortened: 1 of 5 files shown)


kdy1 (changes requested):
Sorry, I reviewed this, but forgot submitting it.

kwonoj:
> Sorry, I reviewed this, but forgot submitting it.

No worries, it's holiday around here and I'm slowing down as well.

crates/swc_ecma_transforms_module/src/import_analysis.rs
```diff
@@ -1,19 +1,23 @@
...
-    as_folder(ImportAnalyzer { scope })
+    as_folder(ImportAnalyzer {
+        scope,
+        import_srcs: Default::default(),
+    })
 }
 
 pub struct ImportAnalyzer {
     scope: Rc<RefCell<Scope>>,
+    import_srcs: AHashSet<JsWord>,
```
kwonoj, line 20: Guess this is it?

kdy1: Yes, correct.

kdy1 (approved):
Thanks!

PR #2869, opened 2021-11-24
fix: custom properties parsing

alexander-akait:
@kdy1 Fix parsing custom properties and any values. 

Our mistakes:
- We apply grammar on custom properties, but spec clearly says no grammar on custom properties, values from custom variables can be used in JS, for example you can get put JSON in custom property, get it using JS and parse, even more you can put `function` there and run it using JS, so we can't apply grammar on it, we should use `parse_declaration_value` https://www.w3.org/TR/css-variables-1/#defining-variables and https://www.w3.org/TR/css-syntax-3/#typedef-declaration-value
- Also we have bugs in our balanced implementation, so I rewrite it for `parse_any_value` too.
- Fixed spaces in custom variables, because `--var: ;` is valid and has real code usage https://github.com/w3c/csswg-drafts/issues/774#issuecomment-434571254
- Rewriting fix bug with wrong `span`
- Also fix a lot of edge cases with improve recovery mode

And small refactor - just cleanup code because we don't need to have duplicated code


crates/swc_css_parser/src/parser/style_rule.rs
```diff
@@ -157,25 +157,35 @@ where
...
+                _ => {
+                    let ctx = Ctx {
+                        allow_operation_in_value: false,
+                        recover_from_property_value: true,
+                        ..self.ctx
+                    };
+                    let (mut parsed_value, parsed_last_pos) =
+                        self.with_ctx(ctx).parse_property_values()?;
+
+                    value.append(&mut parsed_value);
```
kdy1, line 185: I think `value.extend(parsed_value)` is better here.

kdy1 (changes requested):
It's mostly fine, thank you!
I find one small perf/code style issue. (I mean specialization by perf)

alexander-akait:
@kdy1 Done :+1:

crates/swc_css_parser/tests/fixture.rs
```diff
@@ -20,7 +20,14 @@ impl Visit for AssertValid {
             }
 
             match &tokens.tokens[0].token {
```
kdy1, line 22: This seems like a regression?

kdy1 (commented):
Why are tests for pseudo selectors reduced?
It seems like we are skipping more compared to main branch.

crates/swc_css_parser/tests/fixture.rs
alexander-akait: Yep, because we should not try to test it in this way, it is wrong, tokens from custom properties and pseudo selectors can't be reparsed in the same, because they are in other context, i.e. we have
```
:pseudo(foo { color: red }) { }
```

When we try to parse it using tokens input we send tokens to parser:
```
foo { color: red }
```

and our parser thinks it is `QualifiedRule`, but it is wrong, because it is just tokens of pseudo selectors, in `any-value` or declaration-value` according spec we can have almost any value, but they can't be reparsing in right way.

I thought to raise this issue and fix our tests a little later.

alexander-akait: Ideally we should remove this check and output tokens as is in our tests, I can fix it here, it is easy

alexander-akait: You can compare real characters in code and our tokens, and they are same

alexander-akait: Removed and nothing changes :smile:

kdy1: Removing visitor is even more regression, because the only purpose of the visitor is to `panic` if input is invalid. It's not normalizer.

alexander-akait: But we can't use `parse_tokens` in this case, because `parse_tokens` expecting we will have `Stylesheet` at the top, but here no `Stylesheet`, it is part of custom property or pseudo selector.

I think we need to use other approach to tests tokens, we should parse original code (in fixtures) to AST, then use `parse_tokens` to parse original code,and then compare AST, so we can check AST is the same as for tokens input and as for string input

alexander-akait: Also we can add just `parse_tokens` tests and snapshot them, so we can see what is changed in our tokens

alexander-akait: You can try to remove filtering tokens in this branch and you will see why we can't test tokens in this way...

kdy1 (approved):
I see, thank you for description!

alexander-akait:
@kdy1 I put in my todo - adding tokens tests, it should be not hard

Issue #1178, opened 2020-10-22
Better support for getting AST in spite of syntax errors


dsherret:
If swc is used in language services then it will need to be able to understand code that is half typed. This would help a lot in deno_lint as linting errors could be provided as the user works in a file.

## Understand code that is missing certain tokens

For example, the following should parse as an if statement and give back a diagnostic error about a missing close brace:

```ts
if (true) {
```

https://ts-ast-viewer.com/#code/JYMwBAFALgTgrgUwJRgN5A

Other examples that the TS compiler handles (not necessarily handles well, but it still provides an AST):

* https://ts-ast-viewer.com/#code/MYGwhgzhAEDeBQ0nQC4FMIoFzUwJwEsA7Acx32JIG54BfIA
* https://ts-ast-viewer.com/#code/C4JwngBA3gUBEGMD2A7AzsCAzJSIF4IAiHJIgbhgF9EBDYBACwgAoBLLV2gSmkVTRIANgFMAdEKQBzFkVpoAJliLcqvWPGTph4yTKIJaAVymNgKyjSxsUtIUMgb+20ROmzrt+2ArdqQA
* https://ts-ast-viewer.com/#code/MYGwhgzhAECC0G8BQ1oDMD2GAUBKRKq0wGAdhAC7RjQC80GARgFYB0hAvkh0A
* https://ts-ast-viewer.com/#code/MYGwhgzhAEAqCmEAu0DeAod1vSYpAXNMgE4CWAdgObpA
* https://ts-ast-viewer.com/#code/G4QwTgBALgpgzlA-ALggsBLAdgcwgXggFYBuIA

## Fail less, but still provide a diagnostic for invalid tokens in certain positions

Right now in certain scenarios the TS and Roslyn (C#) parsers expand the possibilities in the tree to include invalid possibilities (this is done less than I remembered actually). This is not necessary in most cases though and probably 99.9% of the time a diagnostic could just be created while parsing. For a node missing any child nodes, it could just put in a child node in with a zero width (ts compiler does that in many scenarios).

* https://ts-ast-viewer.com/#code/MYewdgzgLgBADgJxHGBeGBWA3AKFJWKAU2jRgG8cZ4k4BCAGhwF8g
* I could get more actual examples...

Anyway, something to start thinking about.

PR #2874, opened 2021-11-25
feat(es/parser): non-fatal unterminated string parsing

dsherret:
This is part of #1178 for improving SWC support in editors. It allows getting an AST in spite of a unterminated string literal similar to the TypeScript compiler: https://ts-ast-viewer.com/#code/OQFwpgziCWB2DmAoR4oG5kCIRA

In Deno, this will help us do import completions when the user hasn't yet added a quote on a module specifier:

```ts
import * as something from "./
```


kdy1 (approved):
LGTM. Thanks!

PR #2876, opened 2021-11-26
refactor(es/transforms/compat): remove explicit bindings for objectprop

kwonoj:
Continuing #2716. 

This PR is part of #2716, gets rid of explcit state context flag `in_object_prop`. It looks like previous refactor made this sort of noop, test cases seems passing without additional changes.


kdy1 (approved):
Thanks!

PR #2881, opened 2021-11-26
fix(node/swc): correctly deserialize parsefile options

### LINKED ISSUE #2878: Passing plugins to `swc.transformFile` throws an error
### Describe the bug

Passing plugins to `swc.transformFile` throws an error:

```
/path/node_modules/@swc/core/index.js:78
            const res = yield bindings.parseFile(path, toBuffer(options));
                                       ^

Error
    at Compiler.<anonymous> (/path/node_modules/@swc/core/index.js:78:40)
    at Generator.next (<anonymous>)
    at /path/node_modules/@swc/core/index.js:18:71
    at new Promise (<anonymous>)
    at __awaiter (/path/node_modules/@swc/core/index.js:14:12)
    at Compiler.parseFile (/path/node_modules/@swc/core/index.js:75:16)
    at Compiler.<anonymous> (/path/node_modules/@swc/core/index.js:147:38)
    at Generator.next (<anonymous>)
    at /path/node_modules/@swc/core/index.js:18:71
    at new Promise (<anonymous>) {
  code: 'StringExpected'
}
```

### Input code

```typescript
// test.js
const swc = require('@swc/core');
const Visitor = require('@swc/core/Visitor').default;

swc.transformFile('./foo.js', {
  plugin: m => new class extends Visitor {
    visitIdentifier(node) {
      console.log(node);
      return node;
    }
  }().visitProgram(m),
});

// foo.js
console.log('Hello, world!');
```

### Config

_No response_

### Playground link

_No response_

### Expected behavior

```bash
$ node test.js
> {
  type: 'Identifier',
  span: { start: 0, end: 7, ctxt: 0 },
  value: 'console',
  optional: false
}
{
  type: 'Identifier',
  span: { start: 8, end: 11, ctxt: 0 },
  value: 'log',
  optional: false
}
```

### Version

@swc/cli: 0.1.51@swc/core: 1.2.112

### Additional context

`swc.transformFileSync` works as expected

kwonoj:
- closes #2878 

minor fix to parseFile, per its interface https://github.com/swc-project/swc/blob/2462b9941f94bc475cf9ff9c67e3b7c1f98739cc/node-swc/src/index.ts#L237-L248 option is always an object but parse_file trying to read it as string and deserialize later.

### DIFF
--- a/crates/node/src/parse.rs
+++ b/crates/node/src/parse.rs
@@ -182,7 +182,7 @@ pub fn parse_file_sync(cx: CallContext) -> napi::Result<JsString> {
 pub fn parse_file(cx: CallContext) -> napi::Result<JsObject> {
     let c = get_compiler(&cx);
     let path = PathBuf::from(cx.get::<JsString>(0)?.into_utf8()?.as_str()?);
-    let options = cx.get::<JsString>(1)?.into_utf8()?.as_str()?.to_string();
+    let options = cx.get_buffer_as_string(1)?;
 
     cx.env
         .spawn(ParseFileTask { c, path, options })


crates/node/src/parse.rs
```diff
@@ -182,7 +180,7 @@ pub fn parse_file_sync(cx: CallContext) -> napi::Result<JsString> {
 pub fn parse_file(cx: CallContext) -> napi::Result<JsObject> {
     let c = get_compiler(&cx);
     let path = PathBuf::from(cx.get::<JsString>(0)?.into_utf8()?.as_str()?);
-    let options = cx.get::<JsString>(1)?.into_utf8()?.as_str()?.to_string();
+    let options: ParseOptions = cx.get_deserialized(1)?;
```
kdy1, line 183: This is executed in js thread.
That's why I used `String`

kdy1 (changes requested):

crates/node/src/parse.rs
kwonoj: does execution context related to what type of value we'll try to grab? since parse_file passes object, I thought it requires serialization other than coerce into string and current error is due to current path attempts to deserialize object into string.

kdy1: If so, we should pass it as a JSON form and store it to the task as bytes/string and parse from the background thread.

crates/node/src/parse.rs
```diff
@@ -182,7 +182,7 @@ pub fn parse_file_sync(cx: CallContext) -> napi::Result<JsString> {
 pub fn parse_file(cx: CallContext) -> napi::Result<JsObject> {
     let c = get_compiler(&cx);
     let path = PathBuf::from(cx.get::<JsString>(0)?.into_utf8()?.as_str()?);
-    let options = cx.get::<JsString>(1)?.into_utf8()?.as_str()?.to_string();
+    let options = cx.get_buffer_as_string(1)?;
```
kwonoj, line 185: per suggested in https://github.com/swc-project/swc/pull/2881#discussion_r757747507, I left deserialization inside of task as-is.

kdy1 (approved):
Thanks!

PR #2883, opened 2021-11-26
fix(es/transforms/module/commonjs): fix named exports overrides

### LINKED ISSUE #2101: Runtime issue with named export overrides
**Describe the bug**
Based on https://github.com/swc-project/swc/issues/1714

Valid JS is generated, but at runtime if the override is not the same (identity-wise) as the export from `*` the code won't load.

**Input code**
```
import { customRender } from './customRender';

// re-export everything
export * from '@testing-library/react';

// override render method
export { customRender as render };
```

**Config**

```json
{
  "sourceMaps": true,
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true,
      "dynamicImport": true
    },
    "transform": {
      "react": {
        "runtime": "automatic",
        "refresh": true
      },
      "hidden": {
        "jest": true
      }
    }
  },
  "module": {
    "type": "commonjs"
  },
  "env": {
    "coreJs": 3,
    "mode": "usage"
  }
}
```

**Expected behavior**
The transpilation should respect the behaviour that the named exports can override the wildcard exports.
e.g. [Babel REPL](https://babeljs.io/repl#?browsers=node%2014&build=&builtIns=false&corejs=3.6&spec=false&loose=false&code_lz=JYWwDg9gTgLgBAbzgYwK4GcYRAJQKYB2AJnlHAL5wBmU2cA5AHQD0am2-xp9A3AFB9mzOFDwBaPAA9IsOHgBupAJ4wAFsAIBzPlJnwAVNVogGAARh5MGzWIA2wAEZQAhlCXNRz5DF4ChcCEUoKGASEUISMhA8NQgiHWloeCQ2LFwI0jhndHCuMnJ-IA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env,react&prettier=false&targets=Node-14&version=7.14.1&externalPlugins=) generates the `_exportNames` object for overrides and then excludes those when defining the wildcard export.
```
var _exportNames = {
  render: true
};

Object.keys(_react).forEach(function (key) {
  if (Object.prototype.hasOwnProperty.call(_exportNames, key)) return; // <-- This check is missing in the swc output
  if (key in exports && exports[key] === _react[key]) return; // <-- previous fix added this check, but only works when the override is the same as the * export
    enumerable: true,
    get: function () {
      return _react[key];
    }
  });
});
```

**Version**
The version of @swc/core: 1.2.77

**Additional context**
This code should have this behaviour too:
```
export { customRender as render } from './customRender';

// re-export everything, but keep named 'render' exported above
export * from '@testing-library/react';
```

JeremyGrieshop:
This PR closes #2101 and the details can be found in the issue description.

This is somewhat of a followup to PR #2583, as the PR was a incomplete solution.  It only fixed named exports declared as import/export statements.  Furthermore, the order of the named exports could appear before or after the wildcard exports, causing cases to be missed.

This fix requires a preliminary pass through the items to find all named exports first in order to address the ordering.  Furthermore, I've moved the logic from PR #2583 into this preliminary loop instead to catch both cases where export/imports are declared as well as simple named exports.

I've used Babel REPL to compare my output to multiple use cases, which also required changing a couple of internal test outputs.  

This re-ordering can cause the `_exportNames` array to be declared sooner than previously found.  I've found that Babel seems to always declare it right away.

It should also be noted that the testing has changed much since my first PR, and I've tried to follow the same patterns and locations to place and run tests.


crates/swc_ecma_transforms_module/src/common_js.rs
```diff
@@ -89,6 +89,31 @@ where
         // Used only if export * exists
         let mut exported_names: Option<Ident> = None;
 
+        // make a preliminary pass through to collect exported names ahead of time
+        for item in items.clone() {
```
kdy1, line 93: I think we don't need `clone()`.

kdy1 (changes requested):
I found one perf issue, thanks!

crates/swc_ecma_transforms_module/src/common_js.rs
JeremyGrieshop: I'm fairly new to Rust.  Is there a better way to iterate a vector twice?  Removing clone would cause the second iteration to fail.

kdy1: You can use `for item in &items` instead, as `&Vec<T>` implements `IntoIterator<Item = &T>`.

JeremyGrieshop: Ah, thank you!  One of these days, I will learn Rust properly.

kdy1 (approved):
Thanks!

PR #2888, opened 2021-11-27
refactor(es/transforms/compat): visit_mut for spread

kwonoj:
relates to #2370

This PR refactors spread to use visit_mut instead of fold. Mostly straightforward conversion except one place match returns Some() only it needs to create a new expr, and other uses .take() for the existing callee value.

concat_args has not been touched as well since there are few places actually need to rely on returned value. It may need revisit later.


kdy1 (approved):
Thanks!

PR #2889, opened 2021-11-27
chore(es/preset-env): Upgrade `browserslist-rs`

g-plane:
The new version has fixed some bugs and added support of loading configuration.


Cargo.lock
```diff
@@ -271,6 +273,7 @@ dependencies = [
  "serde-wasm-bindgen",
  "serde_json",
  "thiserror",
+ "ureq",
```
kdy1, line 276: Seems like a wrong dependency?
Maybe `[dev-dependencies]` can be better.

Cargo.lock
```diff
@@ -2074,6 +2083,21 @@ version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "448296241d034b96c11173591deaa1302f2c17b56092106c1f92c1bc0183a8c9"
 
+[[package]]
+name = "ring"
```
kdy1, line 2087: CI failed because of this.

kdy1 (changes requested):

Cargo.lock
g-plane: It's used in `build.rs`, so it's under `[build-dependencies]`.

Cargo.lock
g-plane: Should I switch to another HTTP client library?

kdy1: I think it's preferrable. I checked `cargo-deny`, but there was no option to exclude build-time dependencies.

kdy1 (approved):
LGTM. Thanks!

PR #2891, opened 2021-11-27
refactor(es/transform/compat): visit_mut for opt_chaining

kwonoj:
relates to #2370

This PR refactors opt_chaining to use visit_mut instead of fold. There are some explicit clone to bend existing handle_ impl, I feel this may possible to optimize further by rewriting those fns, or I may miss some straightforward workarounds.


kdy1:
If we use `.clone()`, there's no point of using `VisitMut`.
I'm trying to use `VisitMut` to avoid useless `memmove` operations, but cloning is much more costly.

kwonoj:
Yes, agreed. It's not blindly cloning all the things though. Some may legit as previous few cases did, but I feel some are not and raised PR to discuss & move forward.

kwonoj:
I'm trying to improve PR to remove those, meanwhile it'd be appreciated to point out few obvious one I missed while working on it, reason I created PR to share. @kdy1, Does it sound ok to you?

kwonoj:
@kdy1 , OK, I've updated PR to remove clone. Still, I feel there are some non-idomatic paths that need to be updated and would be happy to update further if you may be able to give some peek for the review. 🙏

kdy1:
Did you find the fix? I found that contents are changed.

kwonoj:
@kdy1 I'm not sure if it's `right` fix, but yes, PR has updated to remove `clone` across codes. 

It changed some paths for `handle_` calls - to assign changed values into given expr directly, now it accepts expr directly and then internally matches necessary paths. As wrote, there maybe non idiomatic changes and will fix if there are any.

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
```diff
@@ -30,64 +32,56 @@ pub struct Config {
...
 impl OptChaining {
     /// Returned statements are variable declarations without initializer
-    fn fold_one_stmt_to<T>(&mut self, stmt: T, new: &mut Vec<T>)
+    fn visit_mut_one_stmt_to<T>(&mut self, mut stmt: T, new: &mut Vec<T>)
     where
-        T: StmtLike + FoldWith<Self>,
+        T: StmtLike + VisitMutWith<Self>,
     {
-        let stmt = stmt.fold_with(self);
+        (&mut stmt).visit_mut_with(self);
```
kdy1, line 83: Is this `(&mut )` required?

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
```diff
@@ -166,490 +162,536 @@ impl OptChaining {
...
-
-        Expr::Unary(e)
-    }
-
-    /// Only called from [Fold].
-    fn handle_call(&mut self, e: CallExpr) -> Result<CondExpr, Expr> {
-        match e.callee {
-            ExprOrSuper::Expr(callee) if callee.is_opt_chain() => {
-                let callee = callee.opt_chain().unwrap();
-                let expr = self.unwrap(callee);
```
kdy1, line 233: Can't we just return `CondExpr` from `unwrap`, by taking all components of it?
I think it wil make code much better.

kdy1 (changes requested):
I tried to find all issues, but I'm not sure.

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
kwonoj: no, looks like this was leftover.

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
kwonoj: That was I tried initially and somehow couldn't make it work at the moment, maybe now all the refs are replaced it may possible. I'll give another shot and either update PR or leave comment here if it doesn't work.

kwonoj: @kdy1 , ok, looks like it was able to make handle_ to return as suggested. Thanks for the tip, PR is updated as suggested.

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
```diff
@@ -166,18 +171,17 @@ impl OptChaining {
...
 impl OptChaining {
-    /// Only called from [Fold].
-    fn handle_unary(&mut self, e: UnaryExpr) -> Expr {
+    /// Only called from [VisitMut].
+    fn handle_unary(&mut self, e: &mut UnaryExpr) -> Expr {
         let span = e.span;
 
         if let op!("delete") = e.op {
-            match *e.arg {
+            match &mut *e.arg.take() {
```
kdy1, line 184: This is wrong if `e.arg` si not optional chaining expr nor member expr.

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
```diff
@@ -222,58 +228,58 @@ impl OptChaining {
...
-    /// Only called from [Fold].
-    fn handle_call(&mut self, e: CallExpr) -> Result<CondExpr, Expr> {
-        match e.callee {
+    /// Only called from [VisitMut].
+    fn handle_call(&mut self, e: &mut CallExpr) -> Result<CondExpr, Expr> {
+        match &mut e.callee {
             ExprOrSuper::Expr(callee) if callee.is_opt_chain() => {
-                let callee = callee.opt_chain().unwrap();
-                let expr = self.unwrap(callee);
+                let mut callee = (&mut *callee).take().opt_chain().unwrap();
```
kdy1, line 238: `(&mut *callee).take()` semes identical to `callee.take()`.
Is there a reason you used it?

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
```diff
@@ -222,58 +228,58 @@ impl OptChaining {
...
 
                 return Ok(CondExpr {
                     span: DUMMY_SP,
                     alt: Box::new(Expr::Call(CallExpr {
                         callee: ExprOrSuper::Expr(expr.alt),
-                        ..e
+                        ..e.take()
                     })),
-                    ..expr
+                    test: expr.test,
```
kdy1, line 247: Seems like unnecessary change?

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
```diff
@@ -222,58 +228,58 @@ impl OptChaining {
...
+                    cons: expr.cons,
                 });
             }
             ExprOrSuper::Expr(callee) if callee.is_member() => {
-                let callee = callee.member().unwrap();
-                let callee = self.handle_member(callee);
+                let mut callee = (&mut *callee).take().member().unwrap();
```
kdy1, line 252: Same here. `(&mut *callee).take()` seems redundant.

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
```diff
@@ -328,50 +336,50 @@ impl OptChaining {
...
-        match *e.expr {
+        match &mut *e.expr {
             Expr::Member(MemberExpr {
                 obj: ExprOrSuper::Expr(obj),
                 prop,
                 computed,
                 span: m_span,
             }) if obj.is_opt_chain() => {
-                let obj = obj.opt_chain().unwrap();
+                let mut obj = (&mut *obj).take().opt_chain().unwrap();
```
kdy1, line 372: Same here.

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
```diff
@@ -381,26 +389,25 @@ impl OptChaining {
...
             }
-
             Expr::Call(CallExpr {
                 span,
                 callee: ExprOrSuper::Expr(callee),
                 args,
                 type_args,
             }) if callee.is_opt_chain() => {
-                let callee = callee.opt_chain().unwrap();
+                let mut callee = (&mut *callee).take().opt_chain().unwrap();
```
kdy1, line 398: Same here.

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
```diff
@@ -495,23 +502,25 @@ impl OptChaining {
...
                         ..
                     }) => true,
                     _ => false,
                 };
 
-                let (left, right, alt) = match *obj {
-                    Expr::Ident(..) => (obj.clone(), obj, e.expr),
-                    _ if is_simple_expr(&obj) && self.c.pure_getter => (obj.clone(), obj, e.expr),
+                let (left, right, alt) = match **obj {
+                    Expr::Ident(..) => (obj.clone(), obj.clone(), e.expr.take()),
```
kdy1, line 514: Isn't it safe to use `.take()` for second object?

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
```diff
@@ -495,23 +502,25 @@ impl OptChaining {
...
+                    _ if is_simple_expr(&obj) && self.c.pure_getter => {
+                        (obj.clone(), obj.clone(), e.expr.take())
```
kdy1, line 516: Same here.

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
kwonoj: nope, probabaly leftover while I did this https://twitter.com/bascule/status/1051247460859797504

kdy1 (changes requested):
Thanks!

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
kwonoj: it is safe in terms of compile, and it makes obj to `Invalid` as it's taken over and made existing test fail.

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
kwonoj: yes, same as https://github.com/swc-project/swc/pull/2891#discussion_r758869920

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
kwonoj: Yes, thanks for pointing out. lot of these are leftover while I made changes on top of each, changing one and forgot to bring it back.

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
kwonoj: Thanks, take seems not needed even. 🤦

kwonoj:
Cool, addressed comments as much as I could.

kdy1 (approved):
LGTM. Thanks!

PR #2897, opened 2021-11-27
Refactor code declarations

alexander-akait:
Refactor code:
- Avoid duplicate code
- Fix parsing nested at-rules in list of declarations, it will allow to refactor our at-rules parser very well - we miss `Block` for many at-rules and we don't need duplicate logic parsing for nested at-rules, we will just reuse this logic, I will do refactor at-rules in the next PR
- More tests for unknown at-rules and nested declarations

And no tests changed :smile: 

We are still do some weird things in some places, but we will fix it in near future


kdy1 (approved):
LGTM. Thanks!
I was writing a question about the nested at rules, but I found that this can be useful for some cases.

PR #2905, opened 2021-11-28
feat(css/parser): number type flag

alexander-akait:
Added `type` flag for some tokens. Why we need it?
1. Spec says to do it, https://www.w3.org/TR/css-syntax-3/#consume-number
2. We need it to implement `Integer` AST node, due https://www.w3.org/TR/css-values-3/#integers, CSS grammar distinguish `Number` and `Integer` types, so to output valid grammar AST, we need to introduce `Integer` node (I will do it late)
3. Also in some places we need validate `Num` token to correctly output errors on using, for example https://www.w3.org/TR/css-syntax-3/#the-anb-type, only `type: "integer"` is valid, but right now we allow to use any numbers, but `20.3n + 2` is invalid grammar and we should output on this error (I will improve it after we merged it)

So to solve `2.` and `3.` we need this


alexander-akait:
@kdy1 Can you look at this too? Thanks

kdy1 (approved):
Thanks!

PR #2911, opened 2021-11-29
feat(es/transforms/compat): Add loose mode to parameters

### LINKED ISSUE #2800: Problem with functions with default parameters for ES5 target
### Describe the bug

Some libraries (e.g. ramda) relies on the Function.prototype.length property for currying. Unfortunately, SWC creates wrong output that makes code unusable.

### Input code

```typescript
//this code returns 1 in browser (checked it in Chrome, FF)
const getFunctionLength = fn => fn.length;
function x(a, b='defaultValue') {
}
console.log(getFunctionLength(x))
//but the transpiled code by SWC returns 2:
"use strict";
var getFunctionLength = function(fn) {
    return fn.length;
};
function x(a, param) {
    var b = param === void 0 ? 'someValue' : param;
}
console.log(getFunctionLength(x));
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.106&code=H4sIAAAAAAAAA0vOzysuUUhPLXErzUsuyczP80nNSy%2FJULBVSMtTsLUDkno5YBFrrjSoCoUKjUSdJFv14vzc1LDEnNJUdU2Faq5armSgUfk5qXo5%2BekaGAZqVGhqAgBjKQ75bAAAAA%3D%3D&config=H4sIAAAAAAAAA0WMTQrEIAxG75K1286id5hDBCctFv9IUhgR714tlu7C915ehUMsrBUyshCPS0pU%2FMMKZAOKZZcVTNf6tKEXagYUeScdiiyd%2BZSEJjUQXHRbGSWbQmYSeRHG3T9m66GQfucYKmjJdAc%2F0N7G%2FHPynaLySe0Cj9ke9LUAAAA%3D

### Expected behavior

Transpiling should work correctly and swc should produce correct code for ES5 target

### Version

1.2.110

### Additional context

_No response_

### LINKED ISSUE #2825: Function Length Error beacuse of Simply Adding Parameters
### Describe the bug

When printing the length of a function, the number of formal parameters of the function should be printed. The number of formal parameters does not include the number of remaining parameters, but only "**the number of parameters before the first one with default value**"
However, after SWC conversion, the length of the function is redundant, and the number of default value parameters is calculated.

### Input code

```typescript
function sum(x, y = 0) {
}

console.log(sum.length); // 1
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false,
      "dynamicImport": false,
      "privateMethod": false,
      "functionBind": false,
      "exportDefaultFrom": false,
      "exportNamespaceFrom": false,
      "decorators": false,
      "topLevelAwait": false,
      "importMeta": false
    },
    "transform": null,
    "target": "es5",
    "loose": false,
    "externalHelpers": false
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.111&code=H4sIAAAAAAAAA0srzUsuyczPUyguzdWo0FGoVLBVMNBUqOaq5eJKzs8rzs9J1cvJT9cASuvlpOall2RoWivo6ysYAgCtZO0SOQAAAA%3D%3D&config=H4sIAAAAAAAAA0VOOw6DMAy9i2fWLtygA4ewgkFBSRzZRmqKuHsTNcBkv4%2Ff8wGbOhgPyChK0jYtyfADI5CLqE58NhiqrVILBqVzAENZyZpFX1ULzEpdHSD65JfSkhzHLKT6SJjWcDnPGhR53htxgJVcZzuJnDatqWrind23fzjx%2FDQF%2FJYbJH4nI%2BH8vHl90h1ep15nstP5A1SlDpf7AAAA

### Expected behavior

SWC should not simply add parameters, here is a solution
```js
function sum(x) {
  var y = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : 0;
}
console.log(sum.length); // 1
```

### Version

1.2.111

### Additional context

I run swc use cmd as follows
```bash
```

Austaras:
closes #2825, closes #2800


Austaras:
I would need some help @kdy1 
The test case of issue_2811 is failing because it inject `_this` to wrong scope, so the result will be something like
```js
var f = function () { ... } 
if (true) {
} else {
   var _this = this
}
```
Why is this happening? I guess it's because of `Parallel`, how does that work and how should I fix it?

kdy1:
`#[parallel]` only works if the number of statements (at the same level) is more than 64, so I don't think it's the case.

I think `visit_mut_stmts` should store some 'shared' variables like `self.vars` to somewhere, because otherwise variables from different statements will be injected by `visit_mut_stmts`.

kdy1:
```ts
var f = function () { ... } 
if (true) {
} else {
   var _this = this
}

```

for this case, `visit_mut_stmts` for the top level generates variable, and `visit_mut_stmts` for 
```js
{
   var _this = this
}
```

takes the variables, just as those variables are generated by the nested block.

Austaras:
But `parallel` doesn't allow a custom `visit_mut_stmts`, what should I do?

kdy1:
You can remove `#[parallel]` for now. I'll modify `#[parallel]` in future.

crates/swc_ecma_utils/src/function.rs
```diff
@@ -13,6 +13,14 @@ impl WrapperState {
     pub fn new(this: Option<Ident>, args: Option<Ident>) -> Self {
         Self { this, args }
     }
+    pub fn merge(&mut self, other: Self) {
+        if let Some(this) = other.this {
+            self.this = Some(this)
```
kdy1, line 18: This modifies even if `self.this` is `Some`. Is this intended?
If so, `merge` seems like a confusing name,

kdy1 (changes requested):
Thanks! Almost LGTM.

crates/swc_ecma_utils/src/function.rs
Austaras: Oh, this is made for `parallel`, let me remove it

kdy1 (approved):
Thank you!
I'm bit worried because circular dependency (in `dev-dependencies`) may block publising, but let's try.

PR #2918, opened 2021-11-30
feat(plugin_runner/resolve): supports resolve to path

kwonoj:
This PR is super naive approach to support to resolve local path when load plugins. 

It is based on assumption

- user either passes relative / absolute path can be resolved to file
- user takes responsibility to interop platform specific file path or file name.

It makes api for the plugin config bit polymorphic, like

```
 plugins: Array<[Path | pkgname, Config]> // Path resolves first, if fails try to resolve it as pkg name
```

In result, plugin_name will be explicit path if resolve to local path as well.

This may not be suffecient / or feasible interfaces for the users, like we may want to internal interop to resolve specific file per each platform based on the path, etcs. But rather introduce somewhat complex logic first, guess we can have simplest approach first then update it based on feedback.


kdy1 (commented):
So this allows the user to specify the full path to dylib, right?
I think then it's fine for now.

kwonoj:
> So this allows the user to specify the full path to dylib, right?

Yes, that's the purpose. User specifies full path including file name to load specific plugin binary.

kdy1 (approved):

PR #2919, opened 2021-11-30
fix(css/parser): namespace at-rule

alexander-akait:
fix `@namespace` at-rule:
- resolve todo
- remove wrong `Ident` with zero span
- rename properties according spec


kdy1 (approved):
LGTM, thanks!

PR #2922, opened 2021-11-30
fix(css/parser): unknown at-rule parsing

alexander-akait:
Fix unknown at-rule parsing:
1. Added `SimpleBlock`, it will be union with `Block` in future, to fix it we need to refactor our value and rule parser, it requires some more PRs, partially done here
2. ~~Added `Function`, it will be union with `FnValue`, to fix it we need to refactor our value logic more~~ Done in this PR
3. Also I think we can avoid using `Tokens` struct, because each `TokenAndSpan` already have `span`, so our output can be simplified, will be done in future, after `1.` and `2.`

After this PR we can parse almost all at-rules, even invalid and unknown at-rules, without errors.

Anyway parsing css can be done in two modes:
- grammar parsing - parse on tokens -> consume component value -> apply grammar, in spec - https://www.w3.org/TR/css-syntax-3/#parse-grammar (we do it by default)
- non grammar parsing - parse on tokens -> consume component value, in spec it is by default, for example for component values - https://www.w3.org/TR/css-syntax-3/#parse-component-value (will fix it firstly)

For bundling and dev tools purposes non grammar parsing is enough and it should be faster, for minification better to parse it as grammar because we can apply more optimizations.

Ideally we should switch on non grammar parsing by default in future, but it will be better to provide options for this, anyway to fix it we need more refactor and more PRs, that's what I'm working on, the current PR allows us to start to implement this.

Shorty, our logic should be:

String -> Tokens -> non grammar parsing using [`component_value`] (https://www.w3.org/TR/css-syntax-3/#parse-component-value) -> apply grammar on `component_value` tokens (if grammar is invalid, we well keep non grammar AST)


crates/swc_css_parser/src/parser/at_rule.rs
```diff
@@ -182,71 +182,87 @@ where
...
+                                match val {
+                                    Value::Lazy(token) => {
+                                        at_rule.prelude.tokens.extend(token.tokens);
+                                    }
+                                    _ => {
+                                        unreachable!();
+                                    }
+                                }
+                            }
+                        }
```
alexander-akait, line 256: It is dirty workaround, because we need more refactor, I show it in `1.`, it will be simplified and removed in future

alexander-akait: Resolved in this PR

PR #2938, opened 2021-12-01
fix(es/transforms): reduce renames of constructor and method parameters

dsherret:
I'm breaking up my #2887 into smaller PRs.

This change causes some additional top level declarations (non-exported) to be renamed which could cause issues in script files. However, this is an existing issue with swc that already occurs with having declarations with the same name in nested function declarations. I think the hygiene pass needs to be improved and it's something I will look into a bit more today.


kdy1 (approved):
Seems innocent, thanks!

PR #2939, opened 2021-12-01
fix(ecma/parser): transform tagged template with invalid escape sequence

### LINKED ISSUE #2674: Throw error on valid tagged template literal
### Describe the bug

throws
error: Expected 4 hex characters
 --> input.js:6:5
  |
6 | tag`\unicode and \u{55}`;
  |     ^^

### Input code

```typescript
tag`\unicode and \u{55}`;
```

### Config

_No response_

### Playground link

_No response_

### Expected behavior

should compile

### Version

latest

### Additional context

_No response_

RiESAEX:
fix #2674 

although the syntax restriction of escape sequences in tagged templates is removed in ES2018
we should transform it to any target.

I changed the `swc_ecma_parser::token::Token::Template.cooked` 
to make it store the error from lexer, and let parser decide should it be throw out


crates/swc_ecma_parser/src/lexer/mod.rs
```diff
@@ -972,7 +972,7 @@ impl<'a, I: Input> Lexer<'a, I> {
 
         // TODO: Optimize
         let mut has_escape = false;
-        let mut cooked = Some(String::new());
+        let mut cooked = Ok(String::new());
```
kdy1, line 975: I think `Option<String>` is enough in the code.

kdy1: We don't use value stroed `Err` variant.

kdy1 (approved):
Thanks!
I was bit confused, but I think tests are enough.

PR #2940, opened 2021-12-01
fix(swc/transforms): do not rename top level declarations in hygiene transform

### LINKED ISSUE #2687: Shadowed top level variables are renamed
### Describe the bug

Top level variables that are shadowed are renamed. This is problematic in `<script src="...">` tags, because top level variables become globals on the `window` object. 

### Input code

```typescript
var myNum = 32;
function testShadowing() {
    let myNum = 999;
    console.log("Shadowed myNum is " + myNum); // 999
    console.log("Global myNum is " + window.myNum); // Should be 32, parcel outputs "undefined"
}
testShadowing();
```

### Config

_No response_

### Playground link

https://play.swc.rs/?version=1.2.107&code=H4sIAAAAAAAAA2XOzQ6CMAwH8PueotkJooFET8vi2RsXnmCwAkvKSmDTGOO7O8UYP3pr8%2F%2B1PZkZxksVRzjAfqdFF30bHHsIuIR6MJbPzvdZDlcBqQjDO66U0s9hy35hwoK4z%2BRq0L5ibgEJm7XJNZTlg%2F2rI3Fj6Nukw2lT8UHrgSNZaDC9uoXJzC0ScAxTDIlEb7FzHq0UN%2FHzvr4DSKZB5ugAAAA%3D&config=H4sIAAAAAAAAA0WMSwrDMAxE76K1F00WXeQOPYRwlODiH5ICNcZ3r11cshtm3rwKb7GwVcjIQjySlKj4gQ3IBhTLLiuYjvXqQC%2FUDCjySToQWR%2Fr0mefktAEDAQX3VGGzKaQmUTuCePp%2F2TrrpD2axQVtGT6OZ%2FQbsf8OXlNUPmi9gUibsF1uAAAAA%3D%3D

### Expected behavior

Ideally, no renaming (but I'm sure there is a reason this is done in the first place)

But at least in this case, it would be fine to ensure that top-level variables are never renamed and instead rename the variable in the function.	

### Version

1.2.107

### Additional context

https://github.com/parcel-bundler/parcel/issues/7270

### LINKED ISSUE #2808: swc renames global namespaces and enums in script file
### Describe the bug

SWC renames a TypeScript namespace in a script file to have a `1` on the end, but a name in a script file could be used by consumers in other files that aren't compiled by swc.

This only started happening recently in swc.

### Input code

Also happens with enums, but here's an example with namespaces:

```typescript
// test.ts
namespace Test {
  export const test = 1;
}
```

```
npx swc test.ts
```

### Playground link

~~The playground has different behaviour because it's not doing a hygiene pass.~~

Edit: actually, it is doing this when mangle is turned off, but maybe that is ok because module is set to "common js"? The source type is set to script though.

https://play.swc.rs/?version=1.2.111&code=H4sIAAAAAAAAA8tLzE0tLkhMTlUISS0uUajm5VJQSK0oyC8qUUjOzwOKhCXmlKYq2CoYWvNy1fJy8XKBhPNzUvVy8tM1QHr0wCo0rQFv9PpSSgAAAA%3D%3D&config=H4sIAAAAAAAAA0WOQQ6DMAwE%2F%2BIzh5Yjf%2BgjomBQEIkjr5GKEH%2BvobTcrN2dkTeaEKnbqAYF63FhLRbe1JGtlRE1VaOGDB4NYQY31HMUDSYK6kwX3r0OOrI5xGgf7dOBWQT8R3IqaVgPfZRclYG7CmWcf8vdXVn65Qi28wN3OpKlTKD9Fl1wwutaf%2FEP2XIXPNAAAAA%3D

### Expected behavior

```js
var Test;
(function(Test) {
    Test.test = 1;
})(Test || (Test = {
}));
```

### Actual behaviour

```js
var Test1;
(function(Test) {
    Test.test = 1;
})(Test1 || (Test1 = {
}));
```

### Version

1.2.111

### Additional context

_No response_

dsherret:
~~This is branched from #2938 so that one requires merging first.~~

Closes #2808
Closes #2687


crates/swc_ecma_transforms_base/src/hygiene/usage_analyzer.rs
```diff
@@ -31,6 +31,7 @@ pub(super) struct ScopeData {
 pub(super) struct CurScope<'a> {
     pub parent: Option<&'a CurScope<'a>>,
     pub data: ScopeData,
+    pub depth: u16,
```
dsherret, line 34: What would probably be better is to instead provide `top_level_mark` to `hygiene(top_level_mark)` then that could be checked instead.

dsherret:
I tried to run `cargo test --all-features -all`, but then I get:

```
Doc-tests swc_css
error[E0658]: `#[doc(cfg)]` is experimental
 --> /home/david/swc/crates/swc_css/src/lib.rs:4:20
  |
4 | #[cfg_attr(docsrs, doc(cfg(feature = "minifier")))]
  |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: see issue #43781 <https://github.com/rust-lang/rust/issues/43781> for more information
  = help: add `#![feature(doc_cfg)]` to the crate attributes to enable

error: aborting due to previous error
```

I then tried to run `cargo test --all`, but it looks like it still doesn't run these tests that are failing on the CI.

kdy1:
@dsherret Are you using custom cargo install? `rust-toolchain` is handled by `rustup`, but it seems like stable version of `cargo`  is used.

crates/swc_ecma_transforms_base/src/hygiene/usage_analyzer.rs
kdy1: Yeah, it's a big change and depth check is enough for now.

kdy1: I think `resolver` should distinguish top level bindings and unresolved references, so I filed https://github.com/swc-project/swc/issues/2956
Maybe in future `hygiene` pass can be updated to aware of it, but it's not a scope of this PR.

kdy1 (approved):
Code changes LGTM.
Thank you!
(`hygiene` was headache to me.)


Can you rebase the PR?

dsherret:
@kdy1 Updated.

Also, sorry something keeps going really weird with my merge commits as I'm using linux to develop here and I normally use windows. I will eventually figure it out 😅

> @dsherret Are you using custom cargo install? `rust-toolchain` is handled by `rustup`, but it seems like stable version of `cargo` is used.

It seems to use nightly:

```
/swc$ cargo -V
cargo 1.58.0-nightly (2e2a16e98 2021-11-08)
```

crates/swc_ecma_transforms_base/src/hygiene/usage_analyzer.rs
dsherret: Yes, that sounds better! Then you wouldn't need to pass around a top_level_mark, which can be disruptive.

kdy1 (approved):
LGTM, thansk!

PR #2944, opened 2021-12-02
fix(ci): Fix freebsd builds

kdy1:


csi-lk:
Thanks for jumping on this just hit this issue locally, if you need any help debugging etc. let me know

PR #2947, opened 2021-12-02
refactor(es/transforms/compat): visit_mut for destructuring

kwonoj:
One more check off for https://github.com/swc-project/swc/issues/2370, refactor destructuring into visit_mut.

I tried to make straightforward refactor as same as before.


crates/swc_ecma_transforms_compat/src/macros.rs
```diff
@@ -168,3 +168,159 @@ macro_rules! impl_fold_fn {
         }
     };
 }
+
+macro_rules! impl_visit_mut_fn {
```
kwonoj, line 172: There are few places still using `impl_fold_fn` macro, so PR do not delete it yet.

crates/swc_ecma_transforms_compat/src/es2015/destructuring.rs
```diff
@@ -518,22 +535,27 @@ impl Destructuring {
...
-            .chain(body.stmts)
-            .collect()
+            let mut stmt = Stmt::Decl(Decl::Var(VarDecl {
+                span: DUMMY_SP,
+                kind: VarDeclKind::Let,
+                decls,
+                declare: false,
+            }));
+
+            stmt.visit_mut_children_with(self);
```
kdy1, line 547: I think this is bit different from original, as original one is `.fold_with()`.
So it should be `visit_mut_with(self)` to be identical.

kdy1: But well, as there's no `visit_mut_stmt` so it does not matter.

kdy1 (approved):
Thanks!

PR #2950, opened 2021-12-02
refactor(es/parser/typescript): remove typescript import_assertions config

dsherret:
This doesn't need to be config for Typescript as it is always enabled with TypeScript.


crates/swc_ecma_parser/src/lib.rs
```diff
@@ -148,10 +148,8 @@ impl Syntax {
         match self {
             Syntax::Es(EsConfig {
                 import_assertions, ..
-            })
-            | Syntax::Typescript(TsConfig {
-                import_assertions, ..
             }) => import_assertions,
+            Syntax::Typescript(_) => true,
```
kdy1, line 152: This is the only required part.

kdy1 (changes requested):
Removing this flag can break some code.

kdy1 (approved):
Nevermind. I decided to remove such options to prevent user from being confused.

PR #2970, opened 2021-12-04
fix(ecma/transform): update helpers about array

### LINKED ISSUE #2689: Non-iterable Instance Spreadable Operator is Allowed Incorrectly
### Describe the bug

When the Symbol.iteraor of a element set to undefined, it should not  **spreadable**, 
there is a **unsupportedIterableToArray** check for spreadable operator
For non-iterable instance, the following reminder is required
```bash
Invalid attempt to spread non-iterable instance.
In order to be iterable, non-array objects must have a [Symbol.iterator]() method.
```
but in swc, a variable can be spreaded even if it's Symbol.iterator is set to undefined

### Input code

```typescript
let s = Object('abc');
s[Symbol.iterator] = undefined;
try {
  console.log([...s].length); // TypeError: s is not iterable
}
catch(e) {
  console.log(e.message);
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false,
      "dynamicImport": false,
      "privateMethod": false,
      "functionBind": false,
      "exportDefaultFrom": false,
      "exportNamespaceFrom": false,
      "decorators": false,
      "topLevelAwait": false,
      "importMeta": false
    },
    "transform": null,
    "target": "es5",
    "loose": false,
    "externalHelpers": false
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.107&code=H4sIAAAAAAAAA2WOsQ6CMBBA9yb9h9uA5dgljs4OuhGGtpyAKT3SOwdi%2BHerq%2FvLey%2BSgsAZrv5JQevK%2BVA1nTXS3%2FbVc8RFKTvlPBTolUZ6LInGAmje4W0NQOAkHAkjT3WPiDJgpDTp3HTQtnDfN7rkzPlUOotAYoWf00ey5rAmOA1zTc2%2FjXAlETfRd%2Bj4AEpXwuqqAAAA&config=H4sIAAAAAAAAA0WMTQrEIAxG75K1286id5hDBCctFv9IUhgR714tlu7C915ehUMsrBUyshCPS0pU%2FMMKZAOKZZcVTNf6tKEXagYUeScdiiyd%2BZSEJjUQXHRbGSWbQmYSeRHG3T9m66GQfucYKmjJdAc%2F0N7G%2FHPynaLySe0Cj9ke9LUAAAA%3D

### Expected behavior

Here is the expected conversion results
```js
function _toConsumableArray(arr) {
    return _arrayWithoutHoles(arr) || _iterableToArray(arr) || _unsupportedIterableToArray(arr) || _nonIterableSpread();
}

function _nonIterableSpread() {
    throw new TypeError("Invalid attempt to spread non-iterable instance.\nIn order to be iterable, non-array objects must have a [Symbol.iterator]() method.");
}

function _unsupportedIterableToArray(o, minLen) {
    if (!o) return;
    if (typeof o === "string") return _arrayLikeToArray(o, minLen);
    var n = Object.prototype.toString.call(o).slice(8, -1);
    if (n === "Object" && o.constructor) n = o.constructor.name;
    if (n === "Map" || n === "Set") return Array.from(o);
    if (n === "Arguments" || /^(?:Ui|I)nt(?:8|16|32)(?:Clamped)?Array$/.test(n)) return _arrayLikeToArray(o, minLen);
}

function _iterableToArray(iter) {
    if (typeof Symbol !== "undefined" && iter[Symbol.iterator] != null || iter["@@iterator"] != null) return Array.from(iter);
}

function _arrayWithoutHoles(arr) {
    if (Array.isArray(arr)) return _arrayLikeToArray(arr);
}

function _arrayLikeToArray(arr, len) {
    if (len == null || len > arr.length) len = arr.length;
    for (var i = 0, arr2 = new Array(len); i < len; i++) {
        arr2[i] = arr[i];
    }
    return arr2;
}

var s = Object('abc');
s[Symbol.iterator] = undefined;

try {
    console.log(_toConsumableArray(s).length); // TypeError: s is not iterable
} catch (e) {
    console.log(e.message);
}

```
I think the swc transformation lacks a **_unsupportedIterableToArray** check.

### Version

1.2.107

### Additional context

I run swc use cmd as follows
```bash
```

RiESAEX:
fix #2689


kdy1:
I'll merge this after moving external helpers into this repository.

kdy1 (changes requested):
Can you reabse and modify `packages/swc-helpers`, too?

packages/swc-helpers/src/_unsupported_iterable_to_array.js
```diff
@@ -0,0 +1,9 @@
+export default function _unsupportedIterableToArray(o, minLen) {
+  if (!o) return;
+  if (typeof o === "string") return _arrayLikeToArray(o, minLen);
```
kdy1, line 3: I think `import` for this is required.

crates/swc_ecma_transforms_base/src/helpers/mod.rs
```diff
@@ -164,6 +164,7 @@ macro_rules! define_helpers {
 
 define_helpers!(Helpers {
     apply_decorated_descriptor: (),
+    array_like_to_array: (),
     array_with_holes: (),
```
kdy1, line 168: Can you add a dependency on `_arrayLikeToArray`?

kdy1 (changes requested):
Thank you!
I found two dependency issues.

kdy1 (approved):
LGTM/. Thanks!

PR #2958, opened 2021-12-03
feat(node): upgrade to napi v2

Brooooooklyn:
Merge after napi-rs 2 released

PR #2983, opened 2021-12-07
fix(node/swc): allow binding location to be customized with SWC_BINDING environment variable

alexeagle:
When specified, it should point to the location on disk of the .node binding file, either absolute or relative to the working directory.
Allows tools like Bazel to manage the binding file without the platform-specific npm packages being installed, for example to build the binding from Rust sources.


kwonoj:
The only thing I'm thinking around is if we need more explicit name than `SWC_BINDING`, if it's a new user may not clear what this env does. But also not sure what'll be good name instead. :thinking:

alexeagle:
I also didn't see anywhere in the repo where I could add documentation explaining this, do you know where it goes?

kwonoj:
If I recall correctly, this one? https://github.com/swc-project/website

node-swc/src/index.ts
```diff
@@ -12,7 +13,10 @@ export * from "./types";
 import { BundleInput, compileBundleOptions } from "./spack";
 import { loadBinding } from "@node-rs/helper";
 
-const bindings = loadBinding(__dirname, "swc", "@swc/core")
+// Allow overrides to the location of the .node binding file
+const bindingsOverride = process.env["SWC_BINDING"];
```
kwonoj, line 17: @alexeagle actually thinking this bit more - how much deatil we should expose?

currently it looks like it only cares about the path to the pkg, rest relies on what `loadBinding` assumes around pkg structure (https://github.com/napi-rs/node-rs/blob/main/packages/helper/src/loader.ts#L10)

But if we want full control, say bazel project have a pkg build to swc binary itself and want to point to path to those specific binary 1. will it follow same pkg structure `loadBinding` assumes? 2. will it follow same file name structures to the binary?

maybe something like `SWC_BINARY_PATH` which explicitly requires to absolute / relative path `includes binary file name`, then skips `loadBinding` for those cases would be better?

kdy1 (changes requested):
I think https://github.com/swc-project/swc/pull/2983/files#r763630802 should be resolved.

Brooooooklyn:
`@node-rs/helper` will be removed after napi-rs upgrade to `v2`: https://github.com/swc-project/swc/pull/2958

node-swc/src/index.ts
alexeagle: Sure, makes sense, the napi-rs helper isn't helping for this use case. Changed as you suggested.

alexeagle:
@kdy1 PTAL

kwonoj (approved):
looks ok to me, probabaly need double confirmation.

kdy1 (approved):
Thanks!

alexeagle:
CI failure looks like it needs a retry
```
error An unexpected error occurred: "https://registry.yarnpkg.com/rxjs/-/rxjs-7.3.0.tgz: ESOCKETTIMEDOUT".
info If you think this is a bug, please open a bug report with the information provided in "D:\\a\\swc\\swc\\yarn-error.log".
```

PR #2993, opened 2021-12-08
refactor(es/transforms/compat): visit_mut for class properties

kwonoj:
Followup https://github.com/swc-project/swc/pull/2966 for https://github.com/swc-project/swc/issues/2370. 

This mostly wraps up class properties work. There is a remaining place with fold_* still, it's due to related struct is not yet migrated into visit_mut requires revisit later.


kdy1 (approved):
LGTM. Thanks!

PR #2997, opened 2021-12-09
refactor(es/transforms/compat): visit_mut for object_rest_spread

kwonoj:
#2370 

This is final piece of visit_mut for ecma_transfoms (:yay:) along with current opened PR + remaining piece for class.


kdy1 (approved):
LGTM. Thank you!

PR #3000, opened 2021-12-09
fix: fix regression in enums that reference themselves

dsherret:
I accidentally caused this regression because the resolver pass happens before type stripping now.


crates/swc_ecma_transforms_typescript/src/strip.rs
```diff
@@ -18,7 +18,7 @@ use swc_ecma_utils::{
 use swc_ecma_visit::{as_folder, Fold, Visit, VisitMut, VisitMutWith, VisitWith};
 
 /// Value does not contain TsLit::Bool
-type EnumValues = AHashMap<Id, TsLit>;
+type EnumValues = AHashMap<JsWord, TsLit>;
```
dsherret, line 21: I think it's ok to use `JsWord` here. It seems typescript itself doesn't really handle enums nested within enums that gracefully (because who writes code like that?)

dsherret:
By the way, if this could be including in patch versions of the current crates that would save me a ton of time because I wouldn't have to bump all the crates we have. No big deal if not though.

crates/swc_ecma_transforms_typescript/src/strip.rs
kdy1: You are right.

kdy1 (approved):
LGTM, thanks!
I don't think this is a breaking change, so `cargo update -p swc_ecma_transforms_typescript` will be enough.

PR #3005, opened 2021-12-10
fix(dep_graph): parsing of import assertions

bartlomieju:
Fixes parsing of import assertions for identifiers and adds
parsing of import assertion for dynamic imports.


kdy1 (approved):
Thanks!

PR #3007, opened 2021-12-10
refactor(es/transforms/compat): visit_mut for regenerator (part 1)

kwonoj:
#2370

This is not listed, but while trying to dig one bug realized this wasn't migrated. I'm trying to migrate first, then dig into bugs.


kdy1 (approved):
LGTM, thanks!

PR #3008, opened 2021-12-11
build(package): remove redundant dependencies

kwonoj:
## Description

I may wrong, but after we flattened crates it looks like nested dependenices for ecma_transforms are not being used anymore: actually, except 1 pkg root pkg seems including all necessary dependencies.

PR simply deletes it and CI seems to passing. Still there maybe some manual path requires this dependencies - if that's case I'd like to update those to use root pkg instead.


kdy1 (approved):
Nice catch, thank you!

PR #3009, opened 2021-12-11
build(cargo): use mocha for exec_ test in cargo

kwonoj:
## Description

This PR replaces test runner from `jest` to `mocha` for the tests executed via rust test cases, notably using `exec_with_jest`.

### Motivation

jest is great test framework but it is quite heavyweight. It have several stages of test environment setup for its features like test isolation in vm, transpilation support etcs. In result, each bootstrap to jest takes amount of time always. In case of cargo test per each test cases generated as temporary js test suite we spin up new node.js processes with jest per each file. This effectively makes jest's test isolation not required but still have to spend process warmup time per each file. 

Mocha in opposite it is lightweight test runner does not provide those functionaility, instead performs quite faster when it comes to bootstrap. Since Mocha doesn't provide any assertions by default we can use jest's assertion library without replacing lot of test cases. (Actually one 1 test case accessing global jest. need to be updated).

Here's brief execution time comparision between 2 when running `swc_ecma_transforms_compat`. It's from local machine and I think number may vary on CI but also shows obvious gain.

(`hyperfine -w 3 "cargo test --package swc_ecma_transforms_compat"`)
```
//mocha

Benchmark 1: cargo test --package swc_ecma_transforms_compat
  Time (mean ± σ):     17.259 s ±  1.893 s    [User: 129.996 s, System: 17.067 s]
  Range (min … max):   14.240 s … 19.987 s    10 runs
```

```
//jest

Benchmark 1: cargo test --package swc_ecma_transforms_compat
  Time (mean ± σ):     116.293 s ±  2.252 s    [User: 846.903 s, System: 281.885 s]
  Range (min … max):   114.217 s … 121.382 s    10 runs
```

**PR doesn't replace jest for node-swc tests**: I cannot 100% sure if there's specific reason to use jest for those legit node.js side test cases. For now PR doesn't touch it, depends on usecases we _may_ replace it. But having dedicated test runners for each different purpose won't harm in my opinion.


kdy1 (dismissed):
Thank you!

kwonoj:
There's CI failure for path lookup different to local path - I'll fix it up quickly.

.mocharc.js
```diff
@@ -0,0 +1,12 @@
...
+ * https://github.com/swc-project/swc/pull/3009
+ *
+ * There are 2 test runners in this repo:
+ * - jest is being used for unit testing node-swc packages with javascript test cases
+ * - mocha is being used for cargo's test requires js runtime to validate its transform.
+ *
+ * This config is for the mocha test runner invoked by cargo to resolve its global setup file.
+ */
+module.exports = {
+  require: require.resolve('./.mocha.setup.js')
```
kwonoj, line 11: Instead of using CARGO_WORKSPACE_DIR workaround to let cargo resolve path to root pkg, using config file mocha can resolve automatically.

kdy1 (approved):

PR #3014, opened 2021-12-11
feat(es/transforms/react): refresh support typescript namespace

Austaras:
This is an abandoned feature and little to nobody actually use it, but official babel implementation supported it so anyway

### DIFF
--- a/crates/swc_ecma_transforms_react/src/refresh/mod.rs
+++ b/crates/swc_ecma_transforms_react/src/refresh/mod.rs
@@ -11,7 +11,7 @@ use std::{
     collections::{HashMap, HashSet},
     mem,
 };
-use swc_atoms::JsWord;
+use swc_atoms::{js_word, JsWord};
 use swc_common::{
     collections::{AHashMap, AHashSet},
     comments::Comments,
@@ -107,6 +107,7 @@ pub fn refresh<C: Comments>(
         used_in_jsx: HashSet::default(),
         curr_hook_fn: Vec::new(),
         scope_binding: IndexSet::new(),
+        ts_namespace: Vec::new(),
     }
 }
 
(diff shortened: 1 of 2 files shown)


Austaras:
@kdy1 why is `private_ident` not working in ts namespace?

Austaras:
Oh, and to make this actually work, `swc` needs to put jsx transform before ts transform -- I don't know if it worths it.

kdy1:
@Austaras Maybe because `hygiene` does not handle typescript?
But typescript types should be converted to a variable before invoking `hygiene`, so I'm not sure at the moment.

And I think `typescript::strip_with_jsx` should be applied before react jsx pass, because of binary size and performance.

If we apply jsx before strip, we can make code of `strip` much cleaner, but it's impossible because of `noop_fold_type!` in custom passes. (Code written by people who depends on the crate `swc`)

Austaras:
Fine, let's just ignore this.

Austaras:
Oh, maybe we could have a seperate react refresh pass and run it before ts, would that have any significant impact on performance? If not, I will figure out if hygiene could be fixed.

kdy1:
Yeah, it will be also possible if required. It can be injected after resolver.
https://github.com/swc-project/swc/blob/aab3326b3eec376d2fd754c42f3b2505af163c40/crates/swc/src/config/mod.rs#L372-L373

But is there a reason which makes it impossible to transform after applying `typescript::strip`?

Austaras:
Yes, after ts strip, there won't be any module/namspace left to analysis, so unless we do some control flow analysis, we need it to be placed before ts transform

Austaras:
react refresh only transforms top level react component, and by top level it means top level function/variable and top level ts module/namespace, after transform it will be in an iife so it's hard to do transform.

PR #3020, opened 2021-12-12
refactor(es/transforms/compat): visit_mut for es2015::class

RiESAEX:
#2370 
This is final piece of visit_mut for es2015::class


kdy1 (approved):
Thank you!

PR #3038, opened 2021-12-14
refactor(es/ast): change ClassProp definition

### LINKED ISSUE #2516: class with same name static property is wrongly renamed
**Describe the bug**

**Input code**
```js
class A {
  static A = class {}
}
```

and output is
```js
class A1 {
  static A = class {}
}
```

**Config**

```json
{
  "env": {
     "target": "chrome 80"
  }
}
```

**Expected behavior**
Output should be identical to input

**Version**

### LINKED ISSUE #3026: Change definition of ClassProp
### Discussed in https://github.com/swc-project/swc/discussions/3025

<div type='discussions-op-text'>

<sup>Originally posted by **Austaras** December 12, 2021</sup>
Change `key` field to `PropName` and remove `computed`. There's two reasons:
1. ClassProp key and ClassMethod key are both `ClassElementName` in spec, it's better to unify them
图片
图片
2. Current definition may generate invalid state. When `computed` is false, `key` could only be `Number` `String` `BigInt`

Do we need to wait for a new major release to fix this?</div>

Austaras:
**Description:**
make ClassProp key PropName

**BREAKING CHANGE:**
It will change the definition of ast tree

It will close #3026, and as an intended side effect, close #2516 and some unreported decorator issue


crates/swc_ecma_codegen/src/lib.rs
```diff
@@ -1209,7 +1209,7 @@ where
             space!()
         }
 
-        if n.computed {
+        if n.key.is_computed() {
```
kdy1, line 1212: You can replace this with `emit!(n.key)`.
The emit method for `PropName` is aware of `Computed`.

kdy1 (changes requested):
Mostly LGTM. Thank you!
Can you rebase this?

Austaras:
Done

kdy1 (approved):
Thanks!

PR #3050, opened 2021-12-16
fix(swcpack): Fully-Specified ESM Import Specifier

### LINKED ISSUE #3043: swcpack: Fully-Specified ESM Import Specifier Fails
### Describe the bug

Similar to [this `webpack` bug](https://github.com/webpack/webpack/issues/13252), `swcpack` aka `spack` doesn't match the behavior and semantics of TypeScript when importing using a fully-specified path with a `.js` extension.

Eg. the following fails:

```ts
// index.ts
import { two } from './two.js';
```

```ts
// two.ts
export function two() {}
```

But does not fail when building with TSC.

Full repo with failing `build` script: https://github.com/karlhorky/swcpack-esm-fully-specified-demo

### Input code

_No response_

### Config

_No response_

### Playground link

_No response_

### Expected behavior

`swcpack` should match the semantics of TSC and TypeScript, and allow for fully-specified import paths with `.js` extensions for ESM.

See @RyanCavanaugh's comment here: https://github.com/microsoft/TypeScript/issues/41887#issuecomment-741968855

> TypeScript doesn't modify JavaScript code you write, the import path you write should be the one you want to appear in the output .js file

### Version

@swc/cli@0.1.55, @swc/core@1.2.120

### Additional context

_No response_

magic-akari:

fix #3043

### DIFF
--- a/crates/swc_bundler/tests/common/mod.rs
+++ b/crates/swc_bundler/tests/common/mod.rs
@@ -167,10 +167,39 @@ impl NodeResolver {
             return Ok(path.to_path_buf());
         }
 
-        for ext in EXTENSIONS {
-            let ext_path = path.with_extension(ext);
-            if ext_path.is_file() {
-                return Ok(ext_path);
+        if let Some(name) = path.file_name() {
+            let mut ext_path = path.to_path_buf();
+            let name = name.to_string_lossy();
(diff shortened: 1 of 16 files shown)


kdy1 (approved):
LGTM. Thank you!

Issue #2735, opened 2021-11-12
Optional call `eval` shall be treated as indirect eval call


Austaras:
### Describe the bug

title

### Input code

```typescript
eval?.()
```

### Config

```json
{}
```

### Expected behavior

should output
```js
eval === null || eval === void 0 ? void 0 : (0, eval)();
```

### Version

latest

PR #3068, opened 2021-12-18
fix(es/transforms/compat): transform optional eval call to indirect eval call

magic-akari:

fix https://github.com/swc-project/swc/issues/2735

### DIFF
--- a/crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
+++ b/crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
@@ -4,7 +4,9 @@ use swc_common::{util::take::Take, Spanned, DUMMY_SP};
 use swc_ecma_ast::*;
 use swc_ecma_transforms_base::perf::Check;
 use swc_ecma_transforms_macros::fast_path;
-use swc_ecma_utils::{alias_if_required, prepend, private_ident, undefined, ExprFactory, StmtLike};
+use swc_ecma_utils::{
+    alias_if_required, prepend, private_ident, quote_ident, undefined, ExprFactory, StmtLike,
+};
 use swc_ecma_visit::{
     as_folder, noop_visit_mut_type, noop_visit_type, Fold, Visit, VisitMut, VisitMutWith,
 };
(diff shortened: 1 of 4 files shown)


crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
```diff
@@ -501,16 +503,24 @@ impl OptChaining {
                 ..
             }) => {
                 let obj_span = obj.span();
-                let is_super_access = match **obj {
+                let is_super_access = match obj.as_ref() {
```
kdy1, line 506: I think this change is redundant.

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
```diff
@@ -501,16 +503,24 @@ impl OptChaining {
...
                     Expr::Member(MemberExpr {
                         obj: ExprOrSuper::Super(..),
                         ..
                     }) => true,
                     _ => false,
                 };
 
-                let (left, right, alt) = match **obj {
-                    Expr::Ident(..) => (obj.clone(), obj.clone(), e.expr.take()),
+                let (left, right, alt) = match obj.as_ref() {
```
kdy1, line 514: Same here.

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
```diff
@@ -501,16 +503,24 @@ impl OptChaining {
...
+                    Expr::Ident(ident) => (
+                        obj.clone(),
+                        obj.clone(),
+                        if *ident.sym == *"eval" {
```
kdy1, line 518: This comparison can be replaced with `ident.sym == js_word!("eval")` which is much faster.

kdy1 (changes requested):
Thank you!
I found some tiny issues.

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
magic-akari: I changed it to `&**obj`

kdy1 (approved):
LGTM. Thank you!

PR #3069, opened 2021-12-19
feat(es/preset-env): Upgrade `browserslist-rs`

g-plane:
Upgraded `browserslist-rs`.

## Features

- Added support of `throwOnMissing` option which is introduced since Browserslist v4.19.
- Added support of querying percentage by region, such as `> 1% in CN`.

## Fixes

- Query which starts with `not` can be detected and reports error.
- Renamed WebAssembly entry to avoid collision.


kdy1 (changes requested):
Seems like it has a compilation bug related to wasm?

kdy1 (approved):
LGTM. Thanks!

Issue #1589, opened 2021-04-19
Reimplement CLI in rust


kdy1:
Node binding should export `runCLI`.

This would help lots of cases.

tmpfs:
Hey @kdy1, I keep looking at the Typescript CLI entry points and wonder why they exist and it's not just straight Rust binaries using `clap` (or `structopt`).

Is there functionality that is enabled by having the command line programs exposed as node executables? Is it part of a plan for a Javascript-based plugin system maybe?

I just wanted to look into the reason why `--ignore` isn't working for example and it seems like it was never implemented from what I can tell. Implementing this would be much more straightforward if it was all just Rust :pray:

kdy1:
> Is there functionality that is enabled by having the command line programs exposed as node executables? Is it part of a plan for a Javascript-based plugin system maybe?

No. It's because to make managing version more convenient for the end-user

tmpfs:
Thanks, that helps! I will try to find some time to tackle this one as I think it would simplify things quite a bit - we should be able to remove out quite a bit of typescript :+1: 

I assume you would prefer to use `clap` over `structopt`?

kwonoj:
It's quite late to realize but I forget js-side plugin support in @swc/cli. This may be some headache if we want 100% feature parity.

For now I'll skip those until new cli becomes stable enough. This is also related with experimental plugin stablization and may possible to think about whole plan accordingly based on those.

kwonoj:
`swc_cli` is a ground to implement this. Roughly, these are the tasks need to be implemented and if anyone want to dive in, please feel free to send PR

1. `compile` command
- port rest of cli flag features in subcommand flag struct
- `watch` mode implementation
2. `minify` command
3. `lint` command
4. `bundle` command - accept JSON based configs 

Currently, implementations are sort of exploration, does not have to be 1:1 match to existing `@swc/cli`'s behavior. Specifically, there will be rough edges to deal with dynamic values which requires js runtime. We want to give suffecient time to iterate, finalize behaviors based on feedbacks & some constraints we have.

younes-io:
I would really like to work on one of the commands, but I need some guidance
Anyone could help me to start ?

StemCll:
Hi @kwonoj 
Is this PR still valid? https://github.com/swc-project/swc/pull/5003

stephanrotolante:
Hey @kwonoj,

[...]

PR #3070, opened 2021-12-19
feat(swc/cli): setup packages for binary

kwonoj:
**Description:**

This PR attempts to set up initial binary packages for `@swc/cli`. PR **intentionally** does not contain actual implementations yet. Actual implementation will come later once after we get agreement on the migration paths.

1. `@swc/core` v2 

- `swc_cli` will be built as native binary, supports a single subcommand `compile` which current `@swc/cli` supports.
- `@swc/core`'s platform specific optional deps will includes binary to swc_cli for each platform
- `@swc/cli` provides interop to subcommand-less invocation
```
// pseudo 
// `@swc/cli/index.ts

if (args[0] != 'compile') {
  console.warn(`deprecated: v3 will require to invoke cli with subcommand only');
  exec('swc_cli_bin', 'compile', ...args);
}
else {
 exec('swc_cli_bin', ...args);
}
```

2. `@swc/core` v3
- retiring, let `@swc/cli` mark as deprecated 
- `@swc/core` specifics `bin` property in package.json to point native binary directly or have wrapper shell script to do those. Either cases, node.js should not be required to invoke binaries
- all of cli development consolidated into main repo

**Pros**
- full rust binary
- aligned development in single repo
- 0 v8 warmup time: invoking cli becomes very effective, similar to esbuild's cli performs.

**Cons**
- binary size: this effectively increases pkg size as we need to ship additional binary
- compilation time: may increase time to create binary & node bindings

- relates with #1589


kdy1 (commented):
As an alternative native binary, we can add `runCli` to `@swc/core` and use it.
This would result in slower startup, though.

Another alternative is using a shared dynamic library between node bindings and CLI.

kwonoj:
> we can add runCli to @swc/core and use it

That's first thing I was considered, but eventually came to approach native binary instead. If we were to accept node.js being entrypoint to cost warmup time in my opinion it doesn't have a lot of gain we're trying to migrate logics into rust.

> Another alternative is using a shared dynamic library between node bindings and CLI.

This sounds intriguing. What do we need to do exactly?

kdy1:
We should allow compiling the crate `swc` as dylib.

https://github.com/swc-project/swc/blob/25a17e8b115452474f1010e373abac9ed32caa18/crates/swc_common/Cargo.toml#L15-L16

then we can use 
```toml
[build]
rustflags = ["-C", "prefer-dynamic"]
```

But I'm not sure about the way to force `dylib` only for the crate `swc`.

I expect replacing

```toml
[lib]
crate-type = ["lib", "dylib"]
```

with

```toml
[lib]
crate-type = ["dylib"]
```

on ci will work

kwonoj:
Thanks, I'll try few and get back if I'm stuck.

kwonoj:
Marking as draft for now while trying dylib.

kwonoj:
OK, I tried this bit and _looks_ like dylib based binary can be generated at least. 

But while trying, I noticed few things

- As you already noted, there isn't way to have fine-grained control over specific dependencies link type. Once we enable `prefer-dynamic`, All deps gets dynamically linked including any external dependencies.
- This could cause some trouble, first external deps are out of our control so any abi breakage might occur any time. Also, for the distribution we need to pkging several list of files instead of single statically linked binary.

I think it's worth topic to pursue but also this could be considered as `future optimization`. Thoughts on if we do this? @kdy1 

- Accept current approach (separate bin / node bindings with static linked)
- Trying to implement few features actually using swc_common in swc_cli
- See how badly dev workflow affected (CI times, etcs) and think about further
- @swc/cli won't include to use this binary for those time being, so we can safely introduce any breaking changes depends on outcome.

kdy1:
I think it's fine, but I'm not sure about

> Trying to implement few features actually using swc_common in swc_cli

What feature did you mean?

kwonoj:
> What feature did you mean?

Port over what `@swc/cli` does into this binary. Currently this PR is empty and does nothing as you can see already. What I meant was make progress to port over cli logics and see how does it affect to build time & others.

kdy1:
I see. Thank you for the clarification!

kwonoj:
Changed status to ready for review again - once this goes in I'll try to port over `compile` subcommand into this binary crate.

kdy1 (approved):
Thanks!

Issue #3079, opened 2021-12-20
Unable to setup development environment


laqda:
### Describe the bug

I am trying to set up my development environment as described in [CONTRIBUTING.md](https://github.com/swc-project/swc/blob/main/CONTRIBUTING.md).
The last step instruct to launch tests with the `--all-features` flag.

I get this error :

```
error[E0428]: the name `_generated_bundle_generated_` is defined multiple times
   --> crates/node/src/bundle.rs:235:12
    |
178 | pub(crate) fn bundle(cx: CallContext) -> napi::Result<JsObject> {
    |            ---------------------------------------------------- previous definition of the value `_generated_bundle_generated_` here
...
235 | pub(crate) fn bundle(cx: CallContext) -> napi::Result<JsObject> {
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `_generated_bundle_generated_` redefined here
    |
    = note: `_generated_bundle_generated_` must be defined only once in the value namespace of this module

error[E0428]: the name `bundle` is defined multiple times
   --> crates/node/src/bundle.rs:234:1
    |
177 | #[js_function(1)]
    | ----------------- previous definition of the value `bundle` here
...
234 | #[js_function(1)]
    | ^^^^^^^^^^^^^^^^^ `bundle` redefined here
    |
    = note: `bundle` must be defined only once in the value namespace of this module
    = note: this error originates in the attribute macro `js_function` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0119]: conflicting implementations of trait `napi::Task` for type `bundle::BundleTask`
   --> crates/node/src/bundle.rs:163:1
    |
45  | impl Task for BundleTask {
    | ------------------------ first implementation here
...
163 | impl Task for BundleTask {
    | ^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `bundle::BundleTask`

Some errors have detailed explanations: E0119, E0428.
For more information about an error, try `rustc --explain E0119`.
error: could not compile `node` due to 3 previous errors
warning: build failed, waiting for other jobs to finish...
error: build failed
```

The duplication comes from the features `swc_v1` and `swc_v2` being used at the same time.
The two features `swc_v1` and `swc_v2` are clearly incompatible. I will open a PR to throw a `compile_error` if both features are enabled.

### Expected behavior

[...]

PR #3080, opened 2021-12-20
feat(node): update features check

laqda:
**Description:**

This PR contains two proposals.

First, move feature checks to build script. The `compile_error` error message won't be mixed with other errors that are consequences. What do you think of that ?

Second, add a check that insures that features `swc_v1` and `swc_v2` are not being used at the same time.

https://github.com/swc-project/swc/issues/3079


kdy1:
I don't think we need a build script here

laqda:
The build script was already there. I wasn't sure if it was a good place to have such tests. 
Should I just move the new check next to the existing one ?

kdy1:
Yeah I think so

kdy1 (approved):
LGTM, thank you!

PR #3088, opened 2021-12-21
ci(actions): include swc_cli into optional deps package

kwonoj:
**Description:**
This PR continues CLI implementation, update publish script to each platform specific optional dep to include cli binary as well. Main entrypoint (@swc/cli) and optional dep doesn't point to those binary yet, so this is safe from any breaking changes. 

To conform with existing artifact packaging I did small bend over to copy release binary into specific platform path side by side to node bindings.

### DIFF
--- a/.github/workflows/publish-node.yml
+++ b/.github/workflows/publish-node.yml
@@ -22,15 +22,22 @@ jobs:
           - host: macos-latest
             target: x86_64-apple-darwin
             build: |
+              cargo build -p swc_cli --release
+              cp ./target/release/swc .
               yarn build
               strip -x *.node
           - host: windows-latest
-            build: yarn build
+            build: |
+              yarn build
+              cargo build -p swc_cli --release
+              cp target/release/swc .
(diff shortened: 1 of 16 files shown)


kdy1 (approved):
LGTM, thanks!

alexeagle (commented):
Hey OJ - should this PR have caused the new CLI binary to appear on https://github.com/swc-project/swc/releases ?  I still only see node bindings there. We'd like to benchmark the pure-rust CLI under Bazel :)

kwonoj:
@alexeagle We'll need it but afaik it's not yet included via this PR. For now, you can directly access included binary in `@swc/core`'s native pkg or download it via cdns like unpkg.com. 

I'll try to peek this.

kwonoj:
I wouldn't say it is wrong - please refer announcements that it is currently an experiemental approach to test cli & its deployment mechanism. It is indeed a possible burden who won't need cli binary, reason we'd like to try to couple of different possible approaches to find a suitable way.

PR #3095, opened 2021-12-22
feat: Add DependencyDescriptor::dynamic_import_assertions

### LINKED ISSUE #3052: swc_ecma_dep_graph: Dynamic import assertions empty vs dynamic
Currently there is no way of knowing if any `DependencyDescriptor::import_assertions` are dynamic/unknown or just empty. This information will let the caller do accurate early analysis of dynamic import assertions. I propose to change its type from `HashMap<String, String>` to `Option<HashMap<String, Option<String>>` according to the following specification:
```ts
import("./foo.js");                                                  // import_assertions = Some(HashMap[])
import("./foo.js", {});                                              // import_assertions = Some(HashMap[])
import("./foo.js", dynamic);                                         // import_assertions = None
import("./foo.js", { assert: {} });                                  // import_assertions = Some(HashMap[])
import("./foo.js", { assert: dynamic });                             // import_assertions = None
import("./foo.js", { assert: { type: "json" } });                    // import_assertions = Some(HashMap[("type", Some("json"))])
import("./foo.js", { assert: { type: dynamic } });                   // import_assertions = Some(HashMap[("type", None)])
import("./foo.js", { assert: { type: "json", ...dynamic } });        // import_assertions = None
import("./foo.js", { assert: { type: "json", [dynamic]: "json" } }); // import_assertions = None
```
If the set of assertion keys can not be fully determined (because a non-object-literal was passed or there are any computed keys), set `import_assertions = None`. Otherwise set `import_assertions = Some(HashMap[])` and for each key, if the value is not a string literal set `import_assertions[key] = None` otherwise set `import_assertions[key] = Some(value)`.

Plus there is a bug currently in `parse_import_assertions()` from not handling these cases. It can report assertions that are overridden by computed keys e.g. `import("./foo.js", { type: "json", ["type" + ""]: "blah" })`.

nayeemrmn:
Closes #3052.

### DIFF
--- a/crates/swc_ecma_dep_graph/src/lib.rs
+++ b/crates/swc_ecma_dep_graph/src/lib.rs
@@ -69,7 +69,8 @@ pub struct DependencyDescriptor {
     /// Import assertions for this dependency.
     /// NOTE: it's filled only for static imports and exports.
     pub import_assertions: HashMap<String, String>,
-    /// Import assertions for dynamic imports. Present if `is_dynamic` is true.
+    /// Import assertions for dynamic imports. Present if `is_dynamic` is `true`
+    /// and `kind` is not `DependencyKind::Require`.
     pub dynamic_import_assertions: Option<DynamicImportAssertions>,
 }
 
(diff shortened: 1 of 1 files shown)


kdy1 (approved):
LGTM. Thanks!

PR #3098, opened 2021-12-22
fix(ecma/transforms): correct export ref in object descructuring

### LINKED ISSUE #2858: Improper export with object descructuring
### Describe the bug

When `es5` or lower is picked, `swc` isn't able to properly transpile the following code:

```js
export const {
    a: a2,
    b: b2
} = {a: 1, b: 2}
```

### Input code

```typescript
const {
    a: a1,
    b: b1
} = {a: 1, b: 2}

export const {
    a: a2,
    b: b2
} = {a: 1, b: 2}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript"
    },
    "target": "es5"
  }
}
```

### Playground link

_No response_

### Expected behavior

```js
var ref = {
    a: 1,
    b: 2
}, a1 = ref.a, b1 = ref.b;
var ref1 = {
    a: 1,
    b: 2
}, a2 = ref1.a, b2 = ref1.b;
export { a2, b2 };
```

### Version

1.2.111

### Additional context

In output `ES Modules`: we can see that an `ref` object is exported, but in `Common JS`, all 3 `ref`,  `a2`, and `b2` are exported (when only `a2` an `b2` should be exported):

ES Modules | CommonJS
-|-
image  | image

magic-akari:
**Description:**

fix #2858

### DIFF
--- a/crates/swc/tests/fixture/issue-2858/1/input/.swcrc
+++ b/crates/swc/tests/fixture/issue-2858/1/input/.swcrc
@@ -1,0 +1,11 @@
+{
+    "jsc": {
+        "parser": {
+            "syntax": "ecmascript"
+        },
+        "target": "es5"
+    },
+    "module": {
+        "type": "es6"
+    }
+}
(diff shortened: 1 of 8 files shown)


kdy1 (approved):
LGTM, thanks!

PR #3104, opened 2021-12-23
fix(es/transforms): fix `destructuring binding` in `exprs assign`

### LINKED ISSUE #2850: Consecutive Assignment Error at Object Pattern in Destructure
### Describe the bug

In the poc.js, the destructuring target is the Object pattern. After transforming, there is a consecutive assignment error.
The bug report is as follows when run by node after transforming.
```bash
}, ref1 = ref.foo, foo = ref1 === void 0 ? 123 : ref1, x = 123 = ref.bar, ref);
                                                           ^^^
SyntaxError: Invalid left-hand side in assignment
```

### Input code

```typescript
const obj = {
  foo = 123,
  bar: x = 123,
} = { foo: 24, bar: 45 };
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false,
      "dynamicImport": false,
      "privateMethod": false,
      "functionBind": false,
      "exportDefaultFrom": false,
      "exportNamespaceFrom": false,
      "decorators": false,
      "topLevelAwait": false,
      "importMeta": false
    },
    "transform": null,
    "target": "es5",
    "loose": false,
    "externalHelpers": false
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.111&code=H4sIAAAAAAAAA0vOzysuUchPylKwVajmUlBIy88HsgyNjHWAnKTEIiuFChi%2FFqQEpMBKwchEByJpYqpQaw0AIU%2BnXkQAAAA%3D&config=H4sIAAAAAAAAA0WMTQrEIAxG75K1286id5hDBCctFv9IUhgR714tlu7C915ehUMsrBUyshCPS0pU%2FMMKZAOKZZcVTNf6tKEXagYUeScdiiyd%2BZSEJjUQXHRbGSWbQmYSeRHG3T9m66GQfucYKmjJdAc%2F0N7G%2FHPynaLySe0Cj9ke9LUAAAA%3D

### Expected behavior

No SyntaxError after transforming.

### Version

1.2.111

### Additional context

I run swc use cmd as follows
```bash
```

magic-akari:
**Description:**

**BREAKING CHANGE:**

fix #2850


crates/swc_ecma_transforms_compat/src/es2015/destructuring.rs
```diff
@@ -931,23 +946,35 @@ impl VisitMut for AssignFolder {
                         });
                     }
                     Pat::Assign(pat) => {
-                        let ref_ident = make_ref_ident(self.c, &mut self.vars, None);
+                        let ref_ident = match *pat.right {
+                            Expr::Ident(ref i) if i.span.ctxt() != SyntaxContext::empty() => {
```
kdy1, line 950: What is this check?
All passes are invoked after applying `resolver`, so I think all identifier reference expressions will have non-empty `SyntaxContext`.

crates/swc_ecma_transforms_compat/tests/es2015_destructuring.rs
```diff
@@ -180,12 +180,105 @@ test!(
 ref = [], ref1 = ref[0], code = ref1 === void 0 ? 1 : ref1, ref;"
 );
 
+test!(
+    syntax(),
+    |_| tr(),
```
kdy1, line 185: Can you change the definition of `tr` to 

```rust
fn tr() -> impl Fold {
    chain!(resolver(), destructuring(Config { loose: true }))
}
```

?

This is to match actual behavior of `@swc/core`.

kdy1 (changes requested):
Thank you!

crates/swc_ecma_transforms_compat/tests/es2015_destructuring.rs
magic-akari: done.

magic-akari: You're right. It changes a test result to be more in line with the actual behavior.

kdy1 (approved):
LGTM. Thanks!
I couldn't verify logic mentally so I checked tests :)

PR #3107, opened 2021-12-24
fix(es/compat): Fix `classes` pass

### LINKED ISSUE #3106: Methods with non-identifier characters transpile with invalid output
### Describe the bug

When methods have non-identifier characters (and use quotes in the input), the output contains a function with an invalid name.

### Input code

```typescript
class ValidationStrategy {
    static "object?"(value) {
        return 1;
    }

    static "string!"(value) {
        return 2;
    }
}
console.log(ValidationStrategy);
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "env": {
    "coreJs": 3,
    "mode": "usage"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.115&code=H4sIAAAAAAAAA0vOSSwuVghLzMlMSSzJzM8LLilKLElNr1So5lIAguISoGiyglJ%2BUlZqcom9kkZZYk5pqiZUFgSKUktKi%2FIUDK3BIrVcKNqKS4oy89IVcWszgmmr5UrOzyvOz0nVy8lP18B0j6Y1FwD2vn2wqwAAAA%3D%3D&config=H4sIAAAAAAAAA0WNTQoDIQxG75K1u9Iu5giFHiLYjFj8I9FSEe9enTp0J997eTZ4iYatQUIW4vmSGjJ%2BYAPSHkWzTRnU0Ma0oxPqCjKyoTwVuQ7mYhRaVIG3we51lnT0iUnkjzAYd5p9hHx8ljk0yDXREbzB2Cm8fwGm%2Bzi%2FHObkRdDQNM5fVtnKY6UyF%2BpfKYX%2FzdcAAAA%3D

### Expected behavior

The output should be valid JS

### Actual behavior

```js
    _createClass(ValidationStrategy, null, [
        {
            key: "object?",
            value: function object?(value) { // <----------
                return 1;
            }
        },
```

### Version

1.2.115

### Additional context

https://github.com/parcel-bundler/parcel/issues/7468

kdy1:
swc_ecma_ast:
 - Add `Ident::verify_symbol`.

swc_ecma_transforms_compat:
 - Don't generate invalid identifiers. (Closes #3106)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2817,6 +2817,7 @@ dependencies = [
  "string_enum",
  "swc_atoms 0.2.9",
  "swc_common",
+ "unicode-xid",
 ]
 
 [[package]]
--- a/crates/swc_ecma_ast/Cargo.toml
+++ b/crates/swc_ecma_ast/Cargo.toml
@@ -24,6 +24,7 @@ serde = {version = "1.0.88", features = ["derive"]}
 string_enum = {version = "0.3.1", path = "../string_enum"}
 swc_atoms = {version = "0.2", path = "../swc_atoms"}
 swc_common = {version = "0.15.0", path = "../swc_common"}
+unicode-xid = "0.2"
 
 [dev-dependencies]
 serde_json = "1"
--- a/crates/swc_ecma_ast/src/ident.rs
+++ b/crates/swc_ecma_ast/src/ident.rs
@@ -114,6 +114,13 @@ impl Ident {
     ///
     /// Returns [Cow::Owned] if it's modified.
     pub fn symbol_for_str(s: &str) -> Cow<str> {
+        if s.is_reserved() || s.is_reserved_in_strict_mode(true) || s.is_reserved_in_strict_bind() {
+            let mut buf = String::with_capacity(s.len() + 1);
+            buf.push('_');
+            buf.push_str(s);
+            return Cow::Owned(buf);
+        }
+
         {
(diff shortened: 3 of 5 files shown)


magic-akari:
https://github.com/babel/babel/blob/69ae3b5a250d60618702751aa2aa4f56ccd43988/packages/babel-types/src/converters/toIdentifier.ts#L4-L26

Babel uses algorithms to normalize Identifier.
But can we use anonymous functions without names here?

kdy1:
@magic-akari What do you mean by

> But can we use anonymous functions without names here?

magic-akari:
> @magic-akari What do you mean by
> 
> > But can we use anonymous functions without names here?

```diff
{
    key: "object?",
-   value: function object?(value) {
+   value: function(value) {
        return 1;
    }
},
```

kdy1:
Yeah, it's possible but it will result in a bad stack trace. (I expect js engines to assign the name `value` to those functions)

magic-akari:
> Yeah, it's possible but it will result in a bad stack trace. (I expect js engines to assign the name `value` to those functions)

It makes sense. I forgot the stack trace.
Developers experience is important.

PR #3108, opened 2021-12-24
fix(es/transforms): fix classes method name

magic-akari:
**Description:**

Some functions of class become reserved words after translation.

input:
```js
class Foo {
  "eval?"(value) {
    return value;
  }
}
```

output:
```js
// ...
var Foo = function() {
    "use strict";
    function Foo() {
        _classCallCheck(this, Foo);
    }
    _createClass(Foo, [
        {
            key: "eval?",
            value: function eval(value) {
                return value;
            }
        }
    ]);
    return Foo;
}();
```

This PR will fix this case by rechecking the reserved words.

**BREAKING CHANGE:**

 - #3106
 - #3107


kdy1 (approved):
LGTM. Thanks!

PR #3113, opened 2021-12-24
fix(es/codegen): Implement codegen for static import assertions

### LINKED ISSUE #2802: Keep import assertions intact in targets that support it
### Describe the bug

SWC removes import assertions from esm imports which isn't the expected result, since I am trying to run this file in the browser and it requires import assertions for json imports. I think the import assertions should be kept intact if `jsc.parser.importAssertions`  is set to `true` and the target supports it

### Input code

```typescript
import json from './file.json' assert { type: "json" }
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false,
      "importAssertions": true
    },
    "target": "es2021",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.107&code=H4sIAAAAAAAAA8vMLcgvKlHIKs7PU0grys9VUFDX00%2FLzEnVAwmpKyQWF6cC5asVSioLUq0UlECiSgq1AMdd9yk3AAAA&config=H4sIAAAAAAAAA0WOsQ7DIAxE%2F8UzQ5uhQ7Z%2BQD8CUSeiAoxsRypC%2FHshSprNOt%2B7uwofcTBXyJYFeVxSktovzIAuWnHss4Lpti4tNgga8DET61M6oJ6SwKy8YTOgllfUgcp0m%2B4dC0SCfzD65JcyShzFzChyvWxaw%2BlsPSvSextCBS0Z98wHtCvj3CKvw7hv%2BAFPJhDn0AAAAA%3D%3D

### Expected behavior

The import assertions should stay intact if the target environment is set to esnext, module system is esm and `importAssertions` in the config is set to `true`.

The expected output would be:

```js
import json from  './file.json' assert { type: "json" }
```

### Version

1.2.107

### Additional context

_No response_

### LINKED ISSUE #3110: Import assertions are skipped when transpiled from TS to JS
### Describe the bug

Import assertions affect whether a module is valid in the module graph, depending on the JS runtime, and so should not be stripped when transpiling.

### Input code

```typescript
import data from "./test.json" assert { type: "json" };

console.log(data);
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2022",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.122&code=H4sIAAAAAAAAA8vMLcgvKlFISSxJVEgrys9VUNLTL0ktLtHLKs7PU1JILC5OBUpXK5RUFqRaKShBRGutubiS8%2FOK83NS9XLy0zVAujWtAVSAl05LAAAA&config=H4sIAAAAAAAAA0WNSwrDMAxE76J1FsWLLnKHHsK4SnDxD40CDcZ3rx1cspNmnp4qfeBorVSsgGVMOJPaL62kZ2E48UVpIUWPNhvArS9WdtaOMMzDmF6HnMETWCj65LdzyFyORRi4K5v28Cdbd8X8PkZQr3%2BX80ntdsw7j9cEVQ5uPx%2Bqcl%2B4AAAA

### Expected behavior

```js
import data from "./test.json" assert { type: "json" };
console.log(data);
```

### Actual behavior

```js
import data from "./test.json";
console.log(data);
```

### Version

1.2.122

### Additional context

_No response_

andreubotella:
swc_ecma_codegen:

* Implement codegen for static import assertions.

swc_ecma_transforms_proposal:

* Modify the existing `import_assertions` transform to support assertions in exports.
* Add tests.

swc:

* Add a `keep_import_assertions` option to `JscExperimental` that will keep import assertions in the compiled output.

node-swc:

* Add a `keepImportAssertions` option to the typings for `JscConfig.experimental`.
* Also fix a typo in the typings where `JscConfig.experimental` was spelled as "experimetal".

**Description:**

Currently import assertions are correctly parsed if the input is TypeScript, or EcmaScript with `import_assertions: true`, but the codegen step strips them no matter what. This PR adds a `keep_import_assertions` option to `JscConfig` to keep them.

**BREAKING CHANGE:**

None

Fixes #2802.


RiESAEX:
looks like this will fix #2802 too

crates/swc_ecma_parser/src/parser/stmt/module_item.rs
```diff
@@ -153,8 +153,16 @@ impl<'a, I: Tokens> Parser<I> {
         let _ = cur!(self, false);
         let asserts = if self.input.syntax().import_assertions()
             && !self.input.had_line_break_before_cur()
-            && eat!(self, "assert")
+            && is!(self, "assert")
         {
+            let valid = self.target() >= EsVersion::Es2022;
```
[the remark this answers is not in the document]
kdy1: Also, checking for `es2022` is also wrong. Import assertions is not `es2022`. It's stage 3, not 4.

kdy1 (changes requested):

crates/swc_ecma_parser/src/parser/stmt/module_item.rs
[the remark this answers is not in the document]
andreubotella: I'm not sure they should be stripped out, since host environments can fail an import without assertions that would succeed with them, as happens with JSON modules in HTML's module loader algorithm. But I guess this doesn't have to be the case generally, so stripping them might be fine, since it's not like the codegen will result in code that fails at runtime.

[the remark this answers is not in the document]
andreubotella: > Also, checking for `es2022` is also wrong. Import assertions is not `es2022`. It's stage 3, not 4.

There's no variant for stage 3 or "ESNext" in EsVersion. I went with ES2022 because with it shipping in browsers now, it's extremely likely that it'll be bumped to stage 4 in the next few months. But what should I use instead?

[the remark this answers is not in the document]
kdy1: Maybe an experimental option I think?

crates/swc/src/builder.rs
```diff
@@ -177,6 +177,10 @@ impl<'a, 'b, P: swc_ecma_visit::Fold> PassBuilder<'a, 'b, P> {
             ))
         } else {
             Either::Right(chain!(
+                Optional::new(
+                    compat::experimental::experimental(),
```
kdy1, line 181: This should not be variant of `EsVersion`

kdy1: I meant `Experiemental` type declared in `swc`.

kdy1 (changes requested):

crates/swc/src/builder.rs
andreubotella: You mean adding an option to `JscExperimental` to _keep_ import assertions? I think I was in the mindset that the transformation to strip them had to be just like any other compat transformation, and I missed the fact that this was an option.

andreubotella: Also, I just noticed that there is already a transformation to strip import assertions in `swc_ecma_transforms_proposal`, although it doesn't handle exports.

kdy1: > You mean adding an option to JscExperimental to keep import assertions? 

Yes, it's still stage 3, so keeping it should be opt-in.
Also, it's experimental because it's stage 3 and will be moved to `EsVersion` in the future.

crates/swc/src/config/mod.rs
```diff
@@ -932,6 +935,9 @@ pub struct JscExperimental {
     /// This requires cargo feature `plugin`.
     #[serde(default)]
     pub plugins: Option<Vec<PluginConfig>>,
+    /// If true, keeps import assertions in the output.
+    #[serde(default)]
+    pub import_assertions: bool,
```
kdy1, line 940: Can you change this to `keep_import_assertions`?

kdy1 (changes requested):
Other changes except this LGTM. Thank you!

I think this can make users think it's an option to `enable` instead of to `keep`.

crates/swc/src/config/mod.rs
andreubotella: Done, and updated the OP.

kdy1 (approved):
Thank you!

andreubotella:
> Thank you!

Thank you for the review!

guybedford:
This configuration doesn't seem to apply through the `.swcrc` file correctly currently in the Node.js CLI run, instead only when set manually via `-C jsc.experimental.keepImportAssertions=true`. Perhaps something was overlooked in this PR?

kdy1:
@guybedford Ah, my bad. I think it's because there's no code to handle the field in https://github.com/swc-project/swc/blob/a4e93eb2dff0614e7a39c4301c84654412cf68dc/crates/swc/src/config/mod.rs#L985-L991

PR #3118, opened 2021-12-25
fix(es/transforms): transform bigint literial to string

magic-akari:
**Description:**
bigint literal method name in class should be transformd to string.

input:

```javascript
class Foo {
  1n() {}
  get 2n() {}
  set 3n(x) {}
}
```

SWC output
```javascript
// ...
var Foo = /*#__PURE__*/ function() {
    "use strict";
    function Foo() {
        _classCallCheck(this, Foo);
    }
    _createClass(Foo, [
        {
            key: 1n,
            value: function value() {
            }
        },
        {
            key: 2n,
            get: function get() {
            }
        },
        {
            key: 3n,
            set: function set(x) {
            }
        }
    ]);
    return Foo;
}();
```

expected output
```javascript
// ...
var Foo = /*#__PURE__*/ function() {
    "use strict";
    function Foo() {
        _classCallCheck(this, Foo);
    }
    _createClass(Foo, [
        {
            key: "1",
            value: function value() {
            }
        },
        {
            key: "2",
            get: function get() {
            }
        },
        {
            key: "3",
            set: function set(x) {
            }
        }
    ]);
    return Foo;
}();
```

- [SWC Playground](https://play.swc.rs/?version=1.2.115&code=H4sIAAAAAAAAA0vOSSwuVnDLz1eo5lJQMMzT0FSorgWy0lNLFIzgvGIgzzhPowLMrQUACGYRpDQAAAA%3D&config=H4sIAAAAAAAAAy2LQQ6AIAwE%2F7Jnr3rwDz6CYDEYVNKWREL4u2C47U5mCk6xWAuiYSHuS%2FKt5sUKzZHEso%2BKCSoNOROEajuGD9KmkMxo%2F3r2FKjHvfn58nN%2Fe5dHN8HLNkTlRPUDfb%2F2O3wAAAA%3D)

- [Babel Playground](https://babeljs.io/repl#?browsers=chrome%2045&build=&builtIns=false&corejs=3.6&spec=false&loose=false&code_lz=MYGwhgzhAEBiD29oG8BQ1oEYB2AKAlCgL7rQDmApgC7QBMehyJGE10AzHgB6MlFA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.16.6&externalPlugins=&assumptions=%7B%7D)

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_transforms_compat/src/es2015/classes/mod.rs
+++ b/crates/swc_ecma_transforms_compat/src/es2015/classes/mod.rs
@@ -727,7 +727,15 @@ where
                     }
                     PropName::Str(ref s) => Box::new(Expr::Lit(Lit::Str(s.clone()))),
                     PropName::Num(n) => Box::new(Expr::Lit(Lit::Num(n))),
-                    PropName::BigInt(ref b) => Box::new(Expr::Lit(Lit::BigInt(b.clone()))),
+                    PropName::BigInt(ref b) => Box::new(Expr::Lit(
+                        Str {
+                            span: b.span.clone(),
+                            value: format!("{}", b.value.clone()).into(),
+                            has_escape: false,
+                            kind: Default::default(),
+                        }
+                        .into(),
+                    )),
                     PropName::Computed(ref c) => c.expr.clone(),
                 },
             })
--- a/crates/swc_ecma_transforms_compat/src/es2015/classes/mod.rs
+++ b/crates/swc_ecma_transforms_compat/src/es2015/classes/mod.rs
@@ -729,8 +729,8 @@ where
                     PropName::Num(n) => Box::new(Expr::Lit(Lit::Num(n))),
                     PropName::BigInt(ref b) => Box::new(Expr::Lit(
                         Str {
-                            span: b.span.clone(),
-                            value: format!("{}", b.value.clone()).into(),
+                            span: b.span,
+                            value: format!("{}", b.value).into(),
                             has_escape: false,
                             kind: Default::default(),
                         }
--- a/crates/swc_ecma_transforms_compat/tests/es2015_classes.rs
+++ b/crates/swc_ecma_transforms_compat/tests/es2015_classes.rs
@@ -33,6 +33,28 @@ fn spec_tr(tester: &Tester) -> impl Fold {
 test!(
     syntax(),
     |t| tr(t),
+    bigint_literial_methods,
+    "class Foo{ 1n(){} }",
+    r#"
+let Foo = function () {
+  "use strict";
+  function Foo() {
+    _classCallCheck(this, Foo);
+  }
+  _createClass(Foo, [
+    {
+      key: "1",
+      value: function () {},
+    },
+  ]);
+  return Foo;
+}();
+"#
+);
+
+test!(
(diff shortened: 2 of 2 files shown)


crates/swc_ecma_transforms_compat/src/es2015/classes/mod.rs
```diff
@@ -727,7 +727,15 @@ where
                     }
                     PropName::Str(ref s) => Box::new(Expr::Lit(Lit::Str(s.clone()))),
                     PropName::Num(n) => Box::new(Expr::Lit(Lit::Num(n))),
-                    PropName::BigInt(ref b) => Box::new(Expr::Lit(Lit::BigInt(b.clone()))),
+                    PropName::BigInt(ref b) => Box::new(Expr::Lit(
+                        Str {
+                            span: b.span.clone(),
+                            value: format!("{}", b.value.clone()).into(),
```
kdy1, line 733: I think `.clone()` can be removed.

crates/swc_ecma_transforms_compat/src/es2015/classes/mod.rs
```diff
@@ -727,7 +727,15 @@ where
                     }
                     PropName::Str(ref s) => Box::new(Expr::Lit(Lit::Str(s.clone()))),
                     PropName::Num(n) => Box::new(Expr::Lit(Lit::Num(n))),
-                    PropName::BigInt(ref b) => Box::new(Expr::Lit(Lit::BigInt(b.clone()))),
+                    PropName::BigInt(ref b) => Box::new(Expr::Lit(
+                        Str {
+                            span: b.span.clone(),
```
kdy1, line 732: `.clone()` is redundant as `Span` implements `Copy`.

kdy1 (changes requested):
Thank you!

magic-akari:
Literal numbers should probably be converted to strings.
The current behavior will have no impact on runtime.
But it won't merge functions which have same name.

example code:
```JavaScript
class Foo {
  1() {}
  1n() {}

  2() {}
  "2"() {}
}
```

Babel output:
```JavaScript 
// ...
var Foo = /*#__PURE__*/function () {
  function Foo() {
    _classCallCheck(this, Foo);
  }

  _createClass(Foo, [{
    key: "1",
    value: function _() {}
  }, {
    key: "2",
    value: function _() {}
  }]);

  return Foo;
}();
```

SWC output:
```
// ...
var Foo = /*#__PURE__*/ function() {
    "use strict";
    function Foo() {
        _classCallCheck(this, Foo);
    }
    _createClass(Foo, [
        {
            key: 1,
            value: function value() {
            }
        },
        {
            key: "1",
            value: function value() {
            }
        },
        {
            key: 2,
            value: function value() {
            }
        },
        {
            key: "2",
            value: function 2() {
            }
        }
    ]);
    return Foo;
}();
```

kdy1:
I'm not sure about the merging of such functions.
This may matter if it's overridden.

kdy1 (approved):
Thank you!

PR #3119, opened 2021-12-25
fix(es/transforms): hoist functions in regenerator

### LINKED ISSUE #2556: Function hoisting ignored in async function
For deeply nested async function, hoisting is ignored, and function declared after the return statement are droped.

With this code : 
```js
function requester() {
	return pureRequester
  
    async function pureRequester() {
      	await refreshThenRequest()
        return true;
        
        async function refreshThenRequest() {
        
        }
    }
}
```

and this config : 
```
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true,
      "dynamicImport": true
    },
    "transform": {
      "react": {
        "runtime": "classic",
        "pragma": "React.createElement",
        "pragmaFrag": "React.Fragment"
       }
    }
  }
}
```
We get this : 
```js
function asyncGeneratorStep(gen, resolve, reject, _next, _throw, key, arg) {
    try {
        var info = gen[key](arg);
        var value = info.value;
    } catch (error) {
        reject(error);
        return;
    }
    if (info.done) {
        resolve(value);
    } else {
        Promise.resolve(value).then(_next, _throw);
    }
}
function _asyncToGenerator(fn) {
    return function() {
        var self = this, args = arguments;
        return new Promise(function(resolve, reject) {
            var gen = fn.apply(self, args);
            function _next(value) {
                asyncGeneratorStep(gen, resolve, reject, _next, _throw, "next", value);
            }
            function _throw(err) {
                asyncGeneratorStep(gen, resolve, reject, _next, _throw, "throw", err);
            }
            _next(undefined);
        });
    };
}
function requester() {
    var pureRequester = function pureRequester() {
        return _pureRequester.apply(this, arguments);
    };
    return pureRequester;
    function _pureRequester() {
        _pureRequester = _asyncToGenerator(regeneratorRuntime.mark(function _callee() {
                    case 0:
                        _ctx.next = 2;
                        return refreshThenRequest();
                    case 2:
                        return _ctx.abrupt("return", true);
                    case 5:
                    case "end":
                        return _ctx.stop();
                }
            }, _callee);
        }));
        return _pureRequester.apply(this, arguments);
    }
}
```
[REPL link](https://swc.rs/repl#%7B%22rawCode%22%3A%22%20%20%5Cn%20%20%20%20async%20function%20pureRequester()%20%7B%5Cn%20%20%20%20%20%20%5Ctawait%20refreshThenRequest()%5Cn%20%20%20%20%20%20%20%20return%20'bar'%3B%5Cn%20%20%20%20%20%20%20%20%5Cn%20%20%20%20%20%20%20%20async%20function%20refreshThenRequest()%20%7B%5Cn%20%20%20%20%20%20%20%20%5Ctreturn%20'foo'%5Cn%20%20%20%20%20%20%20%20%7D%5Cn%20%20%20%20%7D%5Cn%22%2C%22rawConfig%22%3A%22%7B%5Cn%20%20%5C%22jsc%5C%22%3A%20%7B%5Cn%20%20%20%20%5C%22parser%5C%22%3A%20%7B%5Cn%20%20%20%20%20%20%5C%22syntax%5C%22%3A%20%5C%22typescript%5C%22%2C%5Cn%20%20%20%20%20%20%5C%22tsx%5C%22%3A%20true%2C%5Cn%20%20%20%20%20%20%5C%22dynamicImport%5C%22%3A%20true%5Cn%20%20%20%20%7D%2C%5Cn%20%20%20%20%5C%22transform%5C%22%3A%20%7B%5Cn%20%20%20%20%20%20%5C%22react%5C%22%3A%20%7B%5Cn%20%20%20%20%20%20%20%20%5C%22runtime%5C%22%3A%20%5C%22classic%5C%22%2C%5Cn%20%20%20%20%20%20%20%20%5C%22pragma%5C%22%3A%20%5C%22React.createElement%5C%22%2C%5Cn%20%20%20%20%20%20%20%20%5C%22pragmaFrag%5C%22%3A%20%5C%22React.Fragment%5C%22%5Cn%20%20%20%20%20%20%20%7D%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%7D)

You can see that the function `refreshThenRequest` is called in `_pureRequester`, but the definition of this function has been droped.

This seems to only happen in the context of async function. If we change to sync functions, the declaration of `refreshThenRequest` is preserved.

### LINKED ISSUE #2620: Transpiling async functions as class properties causes error
### Describe the bug

```
dist/index.js:694
                        console.log(!!_this);
                                      ^
ReferenceError: _this is not defined
```

### Input code

```typescript
class Weird1 {
	weird = async () => {
		console.log(!!this);
	};

	decoy1 = async () => {};
	decoy2 = () => {};
}

class Weird2 {
	weird = async () => {
		console.log(!!this);
	};

	//    decoy1 = async () => { };
	decoy2 = () => {};
}

class Weird3 {
	weird = async () => {
		console.log(!!this);
	};

	decoy1 = async () => {};
	//    decoy2 = () => { };
}

class Weird4 {
	decoy1 = async () => {};
	decoy2 = () => {};

	weird = async () => {
		console.log(!!this);
	};
}

// This will succeed
new Weird4().weird();

// This will succeed
new Weird3().weird();

// This will succeed
new Weird2().weird();

// This will fail with "ReferenceError: _this is not defined"
new Weird1().weird();
```

### Config

```json
{
	"env": {},
	"module": {
		"type": "commonjs"
	}
}
```

package.json (the browserslist config is important):
```json
{
	"browserslist": "> 0.2%",
	"dependencies": {
		"@swc/cli": "^0.1.51",
		"@swc/core": "^1.2.106",
		"browserslist": "^4.17.5",
		"regenerator-runtime": "^0.13.9"
	}
}
```

### Playground link

_No response_

### Expected behavior

Prints "true" four times.

### Version

1.2.106

### Additional context

https://github.com/parcel-bundler/parcel/issues/7147

magic-akari:
**Description:**

**BREAKING CHANGE:**

- fix #2556 
- fix #2620

### DIFF
--- a/crates/swc_ecma_transforms_compat/src/es2015/regenerator/hoist.rs
+++ b/crates/swc_ecma_transforms_compat/src/es2015/regenerator/hoist.rs
@@ -71,7 +71,7 @@ impl Hoister {
             left: PatOrExpr::Pat(Box::new(ident.clone().into())),
             op: op!("="),
             right: Box::new(Expr::Fn(FnExpr {
-                ident: private_ident!(ident.span, format!("_{}", ident.sym)).into(),
+                ident: Some(Ident::new(format!("_{}", ident.sym).into(), ident.span)),
                 function,
             })),
         }
--- a/crates/swc_ecma_transforms_compat/src/es2015/regenerator/hoist.rs
+++ b/crates/swc_ecma_transforms_compat/src/es2015/regenerator/hoist.rs
@@ -14,6 +14,7 @@ where
     let mut v = Hoister {
         vars: Default::default(),
         arguments: None,
+        functions: Default::default(),
     };
     node.visit_mut_with(&mut v);
 
@@ -24,6 +25,7 @@ where
 pub(super) struct Hoister {
     pub vars: Vars,
     pub arguments: Option<Ident>,
+    pub functions: Vec<Stmt>,
 }
 
 impl Hoister {
@@ -53,6 +55,27 @@ impl Hoister {
             span: DUMMY_SP,
             exprs,
         })
+    }
+
+    fn function_to_expr(&mut self, mut func: FnDecl) -> Expr {
+        func.visit_mut_children_with(self);
+
+        let FnDecl {
+            ident, function, ..
+        } = func;
+
+        self.vars.push(ident.clone());
+
+        AssignExpr {
+            span: function.span,
+            left: PatOrExpr::Pat(Box::new(ident.clone().into())),
+            op: op!("="),
+            right: Box::new(Expr::Fn(FnExpr {
+                ident: private_ident!(ident.span, format!("_{}", ident.sym)).into(),
+                function,
+            })),
+        }
+        .into()
     }
 }
 
@@ -103,13 +126,29 @@ impl VisitMut for Hoister {
     }
 
     fn visit_mut_stmt(&mut self, s: &mut Stmt) {
-        if let Stmt::Decl(Decl::Var(var)) = s {
-            let span = var.span;
-            let expr = Box::new(self.var_decl_to_expr(var.take()));
+        match s {
+            Stmt::Decl(Decl::Var(var)) => {
+                let span = var.span;
+                let expr = Box::new(self.var_decl_to_expr(var.take()));
 
-            *s = Stmt::Expr(ExprStmt { span, expr });
-            return;
+                *s = Stmt::Expr(ExprStmt { span, expr });
+                return;
+            }
+            Stmt::Decl(Decl::Fn(func)) => {
+                let span = func.function.span;
+                let expr = Box::new(self.function_to_expr(func.take()));
+
+                let stmt = ExprStmt { span, expr }.into();
+
+                self.functions.push(stmt);
+
+                // cleanup
+                *s = Stmt::dummy();
+                return;
+            }
+            _ => {}
         }
+
         s.visit_mut_children_with(self);
     }
 
--- a/crates/swc_ecma_transforms_compat/tests/es2015_regenerator.rs
+++ b/crates/swc_ecma_transforms_compat/tests/es2015_regenerator.rs
@@ -1317,6 +1317,62 @@ function requester() {
 "#
 );
 
+test_exec!(
+    Syntax::default(),
+    |_| chain!(async_to_generator(), tr(())),
+    issue_2620,
+    r#"
+async function main() {
+    class Weird1 {
+        weird = async () => {
+            return !!this;
+        };
+
+        decoy1 = async () => {};
+        decoy2 = () => {};
+    }
+
+    class Weird2 {
+        weird = async () => {
+            return !!this;
+        };
+
+        //    decoy1 = async () => { };
+        decoy2 = () => {};
+    }
+
+    class Weird3 {
+        weird = async () => {
+            return !!this;
+        };
+
+        decoy1 = async () => {};
+        //    decoy2 = () => { };
+    }
+
+    class Weird4 {
+        decoy1 = async () => {};
+        decoy2 = () => {};
+
+        weird = async () => {
+            return !!this;
+        };
+    }
+
+    return Promise.all([
+        new Weird4().weird(),
+        new Weird3().weird(),
+        new Weird2().weird(),
+        new Weird1().weird(),
+    ]);
+}
+
+return main().then((results) => {
+    expect(results).toEqual([true, true, true, true]);
+});
+"#
+);
+
 test!(
     Syntax::default(),
     |_| tr(()),
--- a/crates/swc_ecma_transforms_compat/tests/es2015_regenerator.rs
+++ b/crates/swc_ecma_transforms_compat/tests/es2015_regenerator.rs
@@ -1172,6 +1172,154 @@ test_exec!(
 test!(
     Syntax::default(),
     |_| tr(()),
+    hoist_function_in_generator_issue_2556_1,
+    r#"
+function* foo() {
+    return bar;
+    function bar() { }
+}
+"#,
+    r#"
+var regeneratorRuntime = require("regenerator-runtime");
+var _marked = regeneratorRuntime.mark(foo);
+function foo() {
+  var bar;
+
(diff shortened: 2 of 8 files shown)


crates/swc_ecma_transforms_compat/src/es2015/regenerator/hoist.rs
```diff
@@ -54,6 +56,27 @@ impl Hoister {
...
+        } = func;
+
+        self.vars.push(ident.clone());
+
+        AssignExpr {
+            span: function.span,
+            left: PatOrExpr::Pat(Box::new(ident.clone().into())),
+            op: op!("="),
+            right: Box::new(Expr::Fn(FnExpr {
+                ident: private_ident!(ident.span, format!("_{}", ident.sym)).into(),
```
kdy1, line 74: I think this will break a function that calls itself if it reaches here.
Can you add a test for it? (Or some comments describing why the function declarations cannot reach here)

You can fix it by using same syntax context as original one.
`Ident::new(&format!("_{}", ident.sym), ident.span)` will fix the issue.

kdy1 (changes requested):
Thanks!

crates/swc_ecma_transforms_compat/tests/fixture/async-to-generator/issue-2556/1/exec.js
```diff
@@ -0,0 +1,17 @@
+async function init() {
+    return fib;
+    async function fib(n) {
+        if (n <= 1) {
+            return n;
+        }
+        const x = await fib(n - 1);
```
magic-akari, line 7: Hi, @kdy1 ,
I added two tests with recursive calls.
Is that what you mean by `a function that calls itself` ?

kdy1: Yes, but I missed one.
Can you add a variable named `fib`?
Renaming bug will be triggered by `hygiene` and `hygiene` requires a conflicting variable.

magic-akari: > Yes, but I missed one.
> Can you add a variable named `fib`?
> Renaming bug will be triggered by `hygiene` and `hygiene` requires a conflicting variable.

Where should I add it?
Is it below the `return fib` statement?

crates/swc_ecma_transforms_compat/tests/fixture/async-to-generator/issue-2556/1/exec.js
```diff
@@ -0,0 +1,21 @@
+async function init() {
+    return fib;
+
+    const fib = 55;
```
magic-akari, line 4: @kdy1 I add the fib verbal here which is unreachable code.
It should be removed in the transform process.
But the actual behavior does not match.

crates/swc_ecma_transforms_compat/tests/es2015_regenerator.rs
```diff
@@ -1169,6 +1169,311 @@ test_exec!(
...
+
+test!(
+    Syntax::default(),
+    |_| chain!(async_to_generator(), tr(())),
+    hoist_function_in_async_issue_2556_6,
+    r#"
+async function foo(a) {
+    return bar1;
+
+    const bar1 = "bar1";
```
magic-akari, line 1384: @kdy1 Same test code here.
Dead codes are correctly removed.

I cannot to figure out the difference between test files and inline test code.

magic-akari (commented):
@kdy1 I need some help, please see the review section.

crates/swc_ecma_transforms_compat/tests/fixture/async-to-generator/issue-2556/1/exec.js
kdy1: I meant a global variable.

crates/swc_ecma_transforms_compat/tests/fixture/async-to-generator/issue-2556/1/exec.js
kdy1: Can you post the transformed code?
I think `resolver` has a bug related to scoping of `const` variables.

crates/swc_ecma_transforms_compat/tests/es2015_regenerator.rs
kdy1: I think the difference is `resolver`.

kdy1 (commented):
Can you prepend `resolver_with_mark` to https://github.com/swc-project/swc/blob/68db26dba7b24063b84b01e312fbba3e7ea9ba65/crates/swc_ecma_transforms_compat/tests/es2017_async_to_generator.rs#L3033-L3039?
I think it's required to test the case I mentioned.

crates/swc_ecma_transforms_compat/tests/es2015_regenerator.rs
```diff
@@ -1169,6 +1169,410 @@ test_exec!(
...
+var fib = function fib() {
+    return 42;
+};
+function init() {
+    return _init.apply(this, arguments);
+}
+function _init() {
+    _init = _asyncToGenerator(
+        regeneratorRuntime.mark(function _callee1() {
+            var fib1, fib1, _fib;
```
magic-akari, line 1512: Hi, @kdy1 ,
The Transformed code is here.
You're right, the variable names are a bit buggy.

magic-akari:
My bad.

```JavaScript
async function foo() {
  return bar;
  const bar = 1;
  async function bar() { }
}
```

This code is not valid before translation.

I will update the test code.

crates/swc_ecma_transforms_compat/tests/fixture/async-to-generator/issue-2556/1/exec.js
magic-akari: It seems that the translated code works well. No naming conflicts.

kdy1 (approved):
LGTM. Thank you!

Issue #2680, opened 2021-11-08
Object.defineProperty Called on Non-object


clhiker:
### Describe the bug

When swc converts object definition properties, there is non-object defineProperty. 
Here is result converted by swc, pay attention to the define **var _obj, _mutatorMap = {}, _obj1, _mutatorMap1 = {};**
```js
function _defineEnumerableProperties(obj, descs) {
    for (var key in descs) {
        var desc = descs[key];
        desc.configurable = desc.enumerable = true;
        if ("value" in desc) desc.writable = true;
        Object.defineProperty(obj, key, desc);
    }
    if (Object.getOwnPropertySymbols) {
        var objectSymbols = Object.getOwnPropertySymbols(descs);
        for (var i = 0; i < objectSymbols.length; i++) {
            var sym = objectSymbols[i];
            var desc = descs[sym];
            desc.configurable = desc.enumerable = true;
            if ("value" in desc) desc.writable = true;
            Object.defineProperty(obj, sym, desc);
        }
    }
    return obj;
}

var a1 = "outside";
var _obj, _mutatorMap = {}, _obj1, _mutatorMap1 = {};
var obj1 = ({}, _mutatorMap1[({}, _mutatorMap[a1] = _mutatorMap[a1] || {}, _mutatorMap[a1].get = function () {
    var a = "inside";
    return a;
}, _defineEnumerableProperties(_obj, _mutatorMap), _obj).outside] = _mutatorMap1[({}, _mutatorMap[a1] = _mutatorMap[a1] || {}, _mutatorMap[a1].get = function () {
    var a = "inside";
    return a;
}, _defineEnumerableProperties(_obj, _mutatorMap), _obj).outside] || {}, _mutatorMap1[({}, _mutatorMap[a1] = _mutatorMap[a1] || {}, _mutatorMap[a1].get = function () {
    var a = "inside";
    return a;
}, _defineEnumerableProperties(_obj, _mutatorMap), _obj).outside].get = function () {
    var a = "middle";
    return a;
}, _defineEnumerableProperties(_obj1, _mutatorMap1), _obj1);
```
Swc defines a undefine variable **var _obj**, but what is needed is an object variable.
When running the code using V8, it gets the following error report,
```bash
swc-trans/poc.js:44: TypeError: Object.defineProperty called on non-object
        Object.defineProperty(obj, key, desc);
               ^
TypeError: Object.defineProperty called on non-object
    at Function.defineProperty (<anonymous>)
[...]

PR #3120, opened 2021-12-26
fix(es/transforms): init obj when using computed props

magic-akari:
**Description:**

**BREAKING CHANGE:**

- fix: #2680

### DIFF
--- a/crates/swc_ecma_transforms_compat/src/es2015/computed_props.rs
+++ b/crates/swc_ecma_transforms_compat/src/es2015/computed_props.rs
@@ -56,6 +56,7 @@ struct ComputedProps {
 #[derive(Default)]
 struct ObjectLitFolder {
     vars: Vec<VarDeclarator>,
+    used_define_enum_props: bool,
     c: Config,
 }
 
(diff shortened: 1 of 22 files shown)


crates/swc_ecma_transforms_compat/src/es2015/computed_props.rs
```diff
@@ -308,6 +309,28 @@ impl VisitMut for ObjectLitFolder {
...
+    node.visit_children_with(&mut visitor);
+    visitor.used_define_enum_props
+}
+
+#[derive(Default)]
+struct DefineEnumVisitor {
+    used_define_enum_props: bool,
+}
+
+impl Visit for DefineEnumVisitor {
```
kdy1, line 323: I think this visitor should be noop for nested functions.

```ts
const obj = {
    foo() {
        const obj2 = {
            get [1]() {}
        }
    }
};
```

The inputs like this can be broken.

kdy1 (changes requested):

crates/swc_ecma_transforms_compat/src/es2015/computed_props.rs
magic-akari: Yes, this is a point worth considering.
This visitor should not visit the nest functions.

There is another problem.
ComputedProps will recursively access the inner AST, and collect all inner VarDeclarator.
So VarDeclarator is misplaced to the outermost layer.

```diff
-   var _obj, _mutatorMap = {};
    var obj1 = {
        foo: function() {
+   var _obj, _mutatorMap = {};
            var obj2 = (
                _obj = {}, 
                _mutatorMap[1] = _mutatorMap[1] || {}, 
                _mutatorMap[1].get = function() {}, 
                _defineEnumerableProperties(_obj, _mutatorMap), 
                _obj
            );
        }
    };
```

I merged the ObjectLitFolder into ComputedProps to ensure that the VarDeclarator is also collected and updated each time the internal stmts are accessed.

kdy1 (approved):
Thank you!

PR #3122, opened 2021-12-26
feat(es/preset-env): Upgrade `browserslist-rs`

g-plane:
## Features

- Added support of querying `cover 99% in XX`

## Fixes

- This release contains a notable fix about WebAssembly support, because previous versions broke the playground.


kdy1 (approved):
Thank you!

PR #3125, opened 2021-12-26
fix(binding): recover on non existing ldd

idan-at:
**Description:**

There are linux environments where LDD does not exist (for example, think of a `node` binary that is linked statically and it built `FROM scratch`)
In order to support those environments, in case `LDD` doesn't exist, the code will use the `gnu` binary.

### DIFF
--- a/node-swc/src/binding.js
+++ b/node-swc/src/binding.js
@@ -5,8 +5,18 @@ const { platform, arch } = process
 
 let nativeBinding = null
 let localFileExisted = false
-let isMusl = false
 let loadError = null
+
+const isMusl = () => {
+  try {
+    const isMusl = readFileSync('/usr/bin/ldd', 'utf8').includes('musl')
+    
+    return isMusl;
+  } catch(e) {
+    return false;
+  }
+}
+
 
 switch (platform) {
   case 'android':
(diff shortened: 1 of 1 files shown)


kdy1 (changes requested):
This seems too expensive. I'm not sure about this, but I think support for musl libc must not make gnu case slower.

idan-at:
> This seems too expensive. I'm not sure about this, but I think support for musl libc must not make gnu case slower.

WDYM by slower? from my understanding, the file was being read anyway. All I added was the fallback to gnu when `ldd` is missing

node-swc/src/binding.js
```diff
@@ -5,9 +5,19 @@ const { platform, arch } = process
 
 let nativeBinding = null
 let localFileExisted = false
-let isMusl = false
 let loadError = null
 
+const isMusl = () => {
+  try {
+    const isMusl = readFileSync('/usr/bin/ldd', 'utf8').includes('musl')
```
kdy1, line 12: I mean this file is needlessly loaded as a node js string.

idan-at: I see. tho it's worth mentioning this PR doesn't change that

I don't mind changing to something else, what did you have in mind?

kdy1: Oh, I see.
Sorry and then it's fine.

kdy1 (approved):
Thank you!

PR #3127, opened 2021-12-26
feat(ecma/compat): object super

### LINKED ISSUE #2684: SWC incorrectly Converted Super Class
### Describe the bug

There is a supper called in attribute assignment method in object, which is not allowed in es5.
when using node run transformed code, it get an error as follows:
```bash
console.log(super.x);
                    ^^^^^
SyntaxError: 'super' keyword unexpected here

```

### Input code

```typescript
obj = {
  foo() {
    console.log(super.x);
  }
}
obj.foo(); // undefined
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false,
      "dynamicImport": false,
      "privateMethod": false,
      "functionBind": false,
      "exportDefaultFrom": false,
      "exportNamespaceFrom": false,
      "decorators": false,
      "topLevelAwait": false,
      "importMeta": false
    },
    "transform": null,
    "target": "es5",
    "loose": false,
    "externalHelpers": false
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.106&code=H4sIAAAAAAAAA8tPylKwVajm5VJQSMvP19CEMBUUkvPzivNzUvVy8tM1iksLUov0KjStQVK1vFxAlJ%2BUpQdWb62gr69QmpeSmpaZl5oCABVRCgFOAAAA&config=H4sIAAAAAAAAA0WMTQrEIAxG75K1286id5hDBCctFv9IUhgR714tlu7C915ehUMsrBUyshCPS0pU%2FMMKZAOKZZcVTNf6tKEXagYUeScdiiyd%2BZSEJjUQXHRbGSWbQmYSeRHG3T9m66GQfucYKmjJdAc%2F0N7G%2FHPynaLySe0Cj9ke9LUAAAA%3D

### Expected behavior

It should be a correct conversion like this
```js
var _obj;

function _get(target, property, receiver) {
    if (typeof Reflect !== "undefined" && Reflect.get) {
        _get = Reflect.get;
    } else {
        _get = function _get(target, property, receiver) {
            var base = _superPropBase(target, property);
            if (!base) return;
            var desc = Object.getOwnPropertyDescriptor(base, property);
            if (desc.get) {
                return desc.get.call(receiver);
            }
            return desc.value;
        };
    }
    return _get(target, property, receiver || target);
}

function _superPropBase(object, property) {
    while (!Object.prototype.hasOwnProperty.call(object, property)) {
        object = _getPrototypeOf(object);
        if (object === null) break;
    }
    return object;
}

function _getPrototypeOf(o) {
    _getPrototypeOf = Object.setPrototypeOf ? Object.getPrototypeOf : function _getPrototypeOf(o) {
        return o.__proto__ || Object.getPrototypeOf(o);
    };
    return _getPrototypeOf(o);
}

obj = _obj = {
    foo: function foo() {
        console.log(_get(_getPrototypeOf(_obj), "x", this));
    }
};
obj.foo(); // undefined
```

### Version

1.2.107

### Additional context

I run swc use cmd as follows
```bash
```

RiESAEX:
**Description:**
Compile ES2015 object super to ES5
[plugin-transform-object-super](https://babeljs.io/docs/en/babel-plugin-transform-object-super) in babel

fix #2684

### DIFF
--- a/crates/swc_ecma_transforms_compat/src/es2015/object_super.rs
+++ b/crates/swc_ecma_transforms_compat/src/es2015/object_super.rs
@@ -1,0 +1,264 @@
+use swc_atoms::js_word;
+use swc_common::{util::take::Take, DUMMY_SP};
+use swc_ecma_ast::*;
+use swc_ecma_transforms_base::helper;
+use swc_ecma_utils::{private_ident, quote_ident, ExprFactory, UsageFinder};
+use swc_ecma_visit::{as_folder, noop_visit_mut_type, Fold, VisitMut, VisitMutWith};
+
+struct ObjectSuper;
+pub fn object_super() -> impl Fold + VisitMut {
+    as_folder(ObjectSuper)
+}
+
+impl VisitMut for ObjectSuper {
+    noop_visit_mut_type!();
+
+    fn visit_mut_expr(&mut self, expr: &mut Expr) {
+        expr.visit_mut_children_with(self);
+        if let Expr::Object(ObjectLit { span, props }) = expr {
+            let obj_ref = private_ident!("_obj");
+            let mut replacer = SuperReplacer {
+                obj: &obj_ref,
+                processed: false,
+            };
+            for prop_or_spread in props.iter_mut() {
+                if let PropOrSpread::Prop(ref mut prop) = prop_or_spread {
+                    match &mut **prop {
+                        Prop::Method(_) => {}
+                        _ => {
+                            prop.visit_mut_with(&mut replacer);
+                        }
+                    }
+                }
+            }
+            if replacer.processed {
+                *expr = Expr::Assign(AssignExpr {
+                    span: DUMMY_SP,
+                    op: AssignOp::Assign,
+                    left: PatOrExpr::Expr(Box::new(Expr::Ident(obj_ref))),
+                    right: Box::new(expr.take()),
+                })
+            }
+        }
+    }
+}
+
+struct SuperReplacer<'a> {
+    obj: &'a Ident,
+    processed: bool,
+}
+impl SuperReplacer<'_> {
+    fn get_proto(&self) -> ExprOrSpread {
+        Expr::Call(CallExpr {
+            span: DUMMY_SP,
+            callee: helper!(get_prototype_of, "getPrototypeOf"),
+            args: vec![self.obj.clone().as_arg()],
+            type_args: Default::default(),
+        })
+        .as_arg()
+    }
+}
+impl VisitMut for SuperReplacer<'_> {
+    fn visit_mut_expr(&mut self, expr: &mut Expr) {
+        match expr {
+            Expr::Member(MemberExpr {
+                obj: ExprOrSuper::Super(..),
+                prop,
+                computed,
+                ..
+            }) => {
+                *expr = Expr::Call(CallExpr {
+                    span: DUMMY_SP,
+                    callee: helper!(set, "get"),
+                    args: vec![
+                        self.get_proto(),
+                        if *computed {
+                            prop.take().as_arg()
+                        } else {
+                            match *prop.take() {
+                                Expr::Ident(Ident { sym, span, .. }) => Expr::Lit(Lit::Str(Str {
+                                    span,
+                                    value: sym,
+                                    has_escape: false,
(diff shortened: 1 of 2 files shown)


kdy1 (changes requested):
Thank you!

crates/swc_ecma_transforms_compat/tests/es2015_object_super.rs
```diff
@@ -0,0 +1,202 @@
...
+    r#"var _super, _obj;
+    var Base = {
+        test: "1"
+    };
+    var obj = _obj = {
+        bar: function bar() {
+            return _set(_getPrototypeOf(_obj), "test", (_super = +_get(_getPrototypeOf(_obj), "test", this)) + 1, this, true), _super;
+        }
+    };
+    Object.setPrototypeOf(obj, Base);"#
```
RiESAEX, line 149: babel will inject `_super` in function `bar` but I guess it's ok?

RiESAEX:
babel will generate code like `var obj = (_obj = { xxx });`,
but looks like swc removes ParenExpr in some pass.

kdy1:
@RiESAEX `fixer` (https://github.com/swc-project/swc/blob/70c2f3b3a57ec9afc1f28ec1c6377dbbf4e920b6/crates/swc_ecma_transforms_base/src/fixer.rs) removes/adds parenthesis.

crates/swc_ecma_transforms_compat/src/es2015/object_super.rs
```diff
@@ -0,0 +1,623 @@
...
+                            init: None,
+                            definite: false,
+                        })
+                        .collect(),
+                }))),
+            );
+        }
+    }
+    fn visit_mut_block_stmt(&mut self, stmts: &mut BlockStmt) {
+        stmts.visit_mut_children_with(self);
```
kdy1, line 47: I found a problem.
This will take all variables even it should not.


```js

const foo = {
    foo() {
        super.bar();
        const nested = {
            nesteedMethod() {
                super.foo();
            }
        };
    }
};


```


The variable for `super.bar();` will be injected into `nesteedMethod`.

To prevent this, you can preserve original variables like

```rust
fn visit_mut_block_stmt(&mut self, s: &mut BlockStmt) {
    let orig_vars = self.extra_var.take();

    s.visit_mut_children_with(self);

    // Prepend

    self.extra_var = orig_vars;
}

```

kdy1 (changes requested):
Thank you!

crates/swc_ecma_transforms_compat/src/es2015/object_super.rs
RiESAEX: I think `self.extra_var` is empty before `visit_mut_children_with(self)`.
since `self .extra_vars.drain(..)`

RiESAEX:
for code like 
```js
let outer = {
  foo() {
    let inter = {
      bar: function bar() {
        super.d; // 'super' is only allowed in object methods and classes.
      },
    };
  },
};
```
should this transformer throw an error or just ignore it (and implement a lint rule)?

kdy1:
Yeah, I think we should create an error from linter, as it will move a runtime error to the compile time.
Will you implement it? It's fine if you decide to file an issue instead to postpone it.

RiESAEX:
I took a look at babel and found that the error was thrown by the parser. It also has a parser option `allowSuperOutsideMethod`.
This is a little out of control for me, so I will file an issue for it.

crates/swc_ecma_transforms_compat/src/es2015/object_super.rs
```diff
@@ -0,0 +1,623 @@
...
+            }
+        }
+    }
+}
+
+struct SuperReplacer {
+    obj: Option<Ident>,
+    vars: Vec<Ident>,
+}
+impl VisitMut for SuperReplacer {
```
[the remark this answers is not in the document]
RiESAEX: it's used in the test case nested_object_super_property_in_key,
While can't have a no-op, it can be optimized to only visit method keys

kdy1 (approved):
LGTM. Thank you!

PR #3129, opened 2021-12-27
refact(ecma/transform/react): use VisitMut for refresh

### LINKED ISSUE #2261: Missing refresh signature for hook calls outside of assignments
**Describe the bug**

If a React hook call isn't assigned but instead used in an expression, then the refresh transformer doesn't seem to recognize it and create a hash.

(This is a valid hook call, and the React Hooks eslint rule doesn't complain)

**Input code**

```js
import { useState } from "react";

export function App() {
	console.log(useState())

	return null;
}
```

**Config**

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": true
    },
    "transform": {
      "react": {
        "refresh": true,
        "development": true
      }
    }
  }
}

```

**Current behaviour** 

```js
import { useState } from "react";
export function App() {
    console.log(useState());
    return null;
}
_c = App;
var _c;
$RefreshReg$(_c, "App");
```

**Expected behavior**
With the offical `react-refresh/babel` plugin:

```js
var _s = $RefreshSig$();

import { useState } from "react";
export function App() {
  _s();

  console.log(useState());
  return null;
}

_s(App, "VhlnxxH3xh7Qs17KADUJ8wvRSZs=");

_c = App;

var _c;

$RefreshReg$(_c, "App");
```

**Version**
20ce326909670d0e40c9f1bbd804387c2958eae6

Austaras:
**Description:**

migrate react refresh transform to `VisitMut`

**BREAKING CHANGE:**

closes #2261 too

### DIFF
--- a/crates/swc_ecma_transforms_react/src/refresh/hook.rs
+++ b/crates/swc_ecma_transforms_react/src/refresh/hook.rs
@@ -1,0 +1,523 @@
+use std::mem;
+
+use indexmap::IndexSet;
+use once_cell::sync::Lazy;
+use regex::Regex;
+use sha1::{Digest, Sha1};
+use swc_atoms::JsWord;
+use swc_common::{util::take::Take, SourceMap, Spanned, DUMMY_SP};
+use swc_ecma_ast::*;
+use swc_ecma_utils::{private_ident, quote_ident, ExprFactory};
+use swc_ecma_visit::{
+    noop_visit_mut_type, noop_visit_type, Visit, VisitMut, VisitMutWith, VisitWith,
+};
+
+use crate::RefreshOptions;
+
+use super::util::{is_builtin_hook, make_call_expr, make_call_stmt, CollectIdent};
+
+// function that use hooks
+struct HookSig {
+    handle: Ident,
+    // need to add an extra register, or alreay inlined
+    hooks: Vec<Hook>,
+}
+
+impl HookSig {
+    fn new(hooks: Vec<Hook>) -> Self {
+        HookSig {
+            handle: private_ident!("_s"),
+            hooks,
+        }
+    }
+}
+
+struct Hook {
+    callee: HookCall,
+    key: String,
+}
+
+// we only consider two kinds of callee as hook call
+enum HookCall {
+    Ident(Ident),
+    Member(Expr, Ident), // for obj and prop
+}
+pub struct HookRegister<'a> {
+    pub options: &'a RefreshOptions,
+    pub ident: Vec<Ident>,
+    pub extra_stmt: Vec<Stmt>,
+    pub scope_binding: IndexSet<JsWord>,
+    pub cm: &'a SourceMap,
+    pub should_reset: bool,
+}
+
+impl<'a> HookRegister<'a> {
+    pub fn gen_hook_handle(&mut self) -> Stmt {
+        Stmt::Decl(Decl::Var(VarDecl {
+            span: DUMMY_SP,
+            kind: VarDeclKind::Var,
+            decls: self
+                .ident
+                .take()
+                .into_iter()
+                .map(|id| VarDeclarator {
+                    span: DUMMY_SP,
+                    name: Pat::Ident(BindingIdent::from(id)),
+                    init: Some(Box::new(make_call_expr(quote_ident!(self
+                        .options
+                        .refresh_sig
+                        .clone())))),
+                    definite: false,
+                })
+                .collect(),
+            declare: false,
+        }))
+    }
+
+    // The second call is around the function itself. This is used to associate a
+    // type with a signature.
+    // Unlike with $RefreshReg$, this needs to work for nested declarations too.
+    fn wrap_with_register(&self, handle: Ident, func: Expr, hooks: Vec<Hook>) -> Expr {
+        let mut args = vec![func.as_arg()];
+        let mut sign = Vec::new();
+        let mut custom_hook = Vec::new();
+
+        for hook in hooks {
+            let name = match &hook.callee {
+                HookCall::Ident(i) => i,
+                HookCall::Member(_, i) => i,
+            };
+            sign.push(format!("{}{{{}}}", name.sym, hook.key));
+            match &hook.callee {
+                HookCall::Ident(ident) if !is_builtin_hook(ident) => {
+                    custom_hook.push(hook.callee);
+                }
+                HookCall::Member(obj, prop) if !is_builtin_hook(prop) => {
+                    if let Expr::Ident(ident) = obj {
+                        if ident.sym.as_ref() != "React" {
+                            custom_hook.push(hook.callee);
+                        }
+                    }
+                }
+                _ => (),
+            };
+        }
+
+        // this is just for pass test
+        let has_escape = sign.len() > 1;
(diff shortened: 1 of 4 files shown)


crates/swc_ecma_transforms_react/src/refresh/hook.rs
```diff
@@ -0,0 +1,523 @@
...
+                .collect(),
+            declare: false,
+        }))
+    }
+
+    // The second call is around the function itself. This is used to associate a
+    // type with a signature.
+    // Unlike with $RefreshReg$, this needs to work for nested declarations too.
+    fn wrap_with_register(&self, handle: Ident, func: Expr, hooks: Vec<Hook>) -> Expr {
+        let mut args = vec![func];
```
kdy1, line 81: I think changing the type of `args` to `Vec<ExprOrSpread>` will be better for performance as it reduces memory operations.

kdy1: You can use `let mut args = vec![func.as_arg()];` instead.

crates/swc_ecma_transforms_react/src/refresh/hook.rs
```diff
@@ -0,0 +1,523 @@
...
+        args.push(Expr::Lit(Lit::Str(Str {
+            span: DUMMY_SP,
+            value: sign.into(),
+            has_escape,
+            kind: StrKind::Synthesized,
+        })));
+
+        let mut should_reset = self.should_reset;
+
+        let mut custom_hook_in_scope = Vec::new();
```
kdy1, line 126: Same here. `Vec<ExprOrSpread>` will avoid memory operations.

crates/swc_ecma_transforms_react/src/refresh/hook.rs
```diff
@@ -0,0 +1,523 @@
...
+
+    fn visit_mut_block_stmt(&mut self, b: &mut BlockStmt) {
+        let mut current_scope = IndexSet::new();
+
+        // TODO: merge with collect_decls
+        for stmt in &b.stmts {
+            stmt.collect_ident(&mut current_scope);
+        }
+        let orig_binding = self.scope_binding.len();
+        self.scope_binding.extend(current_scope.into_iter());
```
kdy1, line 225: Is `.into_iter()` required?
`.extend` accepts `impl IntoIterator` so this seems redundant.

crates/swc_ecma_transforms_react/src/refresh/hook.rs
```diff
@@ -0,0 +1,523 @@
...
+        }?;
+        let name = if IS_HOOK_LIKE.is_match(&ident.sym) {
+            Some(ident)
+        } else {
+            None
+        }?;
+        let mut key = if let Some(name) = lhs {
+            self.cm
+                .span_to_snippet(name.span())
+                .unwrap_or("".to_string())
```
kdy1, line 438: I think we should use `unwrap_or_else` or `.unwrap_or(String::new())`.

Maybe specialization used by std may optimize this, but just to be safe.

crates/swc_ecma_transforms_react/src/refresh/hook.rs
```diff
@@ -0,0 +1,523 @@
...
+        } else {
+            "".to_string()
```
kdy1, line 440: Same here

crates/swc_ecma_transforms_react/src/refresh/hook.rs
```diff
@@ -0,0 +1,523 @@
...
+        };
+        // Some built-in Hooks reset on edits to arguments.
+        if &name.sym == "useState" && expr.args.len() > 0 {
+            // useState first argument is initial state.
+            key += &format!(
+                "({})",
+                self.cm
+                    .span_to_snippet(expr.args[0].span())
+                    .unwrap_or("".to_string())
```
kdy1, line 449: Same here

crates/swc_ecma_transforms_react/src/refresh/hook.rs
```diff
@@ -0,0 +1,523 @@
...
+            );
+        } else if &name.sym == "useReducer" && expr.args.len() > 1 {
+            // useReducer second argument is initial state.
+            key += &format!(
+                "({})",
+                self.cm
+                    .span_to_snippet(expr.args[1].span())
+                    .unwrap_or("".to_string())
```
kdy1, line 457: Same here

crates/swc_ecma_transforms_react/src/refresh/mod.rs
```diff
@@ -497,24 +174,24 @@ impl<C: Comments> Refresh<C> {
...
                 reg.push((reg_ident.clone(), reg_str));
                 if let Persist::Hoc(hoc) =
-                    self.get_persistent_id_from_possible_hoc(expr, reg, ignore)
+                    self.get_persistent_id_from_possible_hoc(expr, reg, hook_reg)
                 {
-                    *first_arg = Box::new(make_assign_stmt(reg_ident.clone(), first_arg.take()));
-                    if let Some(hook) = &hoc.hook {
-                        let span = call_expr.span;
-                        let first = ExprOrSpread {
+                    let mut first = first_arg.take();
```
kdy1, line 179: This will make the output invalid if `hoc.hook` is `None`.

I'm not sure if this is possible, though.

crates/swc_ecma_transforms_react/src/refresh/mod.rs
```diff
@@ -497,24 +174,24 @@ impl<C: Comments> Refresh<C> {
...
+                        let mut args = vec![ExprOrSpread {
                             spread: None,
-                            expr: Box::new(Expr::Call(call_expr.take())),
-                        };
-                        let mut args = vec![first];
-                        args.extend(hook.args.clone());
-                        *call_expr = CallExpr {
+                            expr: first,
+                        }];
+                        args.append(&mut rest_arg.clone());
```
kdy1, line 186: `args.extend(hook.args.clone());` seems better because of specializations.

crates/swc_ecma_transforms_react/src/refresh/mod.rs
```diff
@@ -869,12 +383,20 @@ impl<C: Comments> Fold for Refresh<C> {
...
+            if let Persist::Hoc(_) = persistent_id {
+                // we need to make hook transform happens after component for
+                // HOC
+                items.push(item);
+            } else {
+                item.visit_mut_children_with(&mut hook_visitor);
+                self.scope_binding.truncate(current_binding);
+
+                items.push(item);
+                let stmts = hook_visitor.extra_stmt.take();
```
kdy1, line 395: `items.extend(hook_visitor.extra_stmt.take().into_iter().map(ModuleItem::Stmt))` would be better for performance because of `TrustedLen`.

kdy1 (changes requested):
Thank you!

crates/swc_ecma_transforms_react/src/refresh/mod.rs
Austaras: It won't, see L194

kdy1 (approved):
LGTM. Thanks!

Issue #2924, opened 2021-12-01
[mangle] exporting from along with importing the same variable name doesn't work


neuronetio:
### Describe the bug

just run the code

### Input code

```typescript
import { y } from 'z';
export { y } from 'z';
```
### Output

```javascript
import { y as a } from 'z';
export { a as y } from 'z'; // <- a is not defined in z
```

### Playground link

[Playground](https://play.swc.rs/?version=1.2.111&code=H4sIAAAAAAAAA8vMLcgvKlGoVqhUqFVIK8rPVVCvUrfmSq3AJgwAW0YioS0AAAA%3D&config=H4sIAAAAAAAAA22PSwrDMAxE76J1FkkWpeQOPUMRrhLc%2BiMsJTQE371OyYdCVxqNHmJmgacY6BZgTEJpVTIHxTd0oDOTmGRZoQKVYvXohHJZMA2kBSFp67YpZxej0AZU4G2w%2Fbw%2B8xgGR6vSyI4mcgf0IuK7cSgS0JP8%2Bv0fj5OdUKnMyOfJ0vXQgj0m29RnUhM9J5Idz8Xy8TFukUrBb4kL5DP0%2FlduG6hppPwBXE3XBSkBAAA%3D)

### Expected behavior

exported value should keep proper name

### Version

1.2.114

dsherret:
This also sometimes occurs when transforming and not mangling. It's a problem in the resolver code.

PR #3130, opened 2021-12-28
fix(es/transform/resolver): avoid ident collapse for the reexport

### LINKED ISSUE #3114: `import type {foo} from 'foo'` should not remove `foo` from `export {foo} from 'foo'`
### Describe the bug

Hi! First of all thank u for the swc.

Looks like it has a bug with `type` keyword.

### Input code

_No response_

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "target": "es2021",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.115&code=H4sIAAAAAAAAA8vMLcgvKlEoqSxIVahOy8%2BvVUgrys9VUAcy1bm4UisQsm75%2BQq2YGZ%2BmgJQ2houja4PAGYvU81UAAAA&config=H4sIAAAAAAAAA0WOSwrDMAxE76J1FsWLLnyHHsK4SnDxD0mBGuO7VzYp2Q0zo6fp8GEPtkN1xEhTccvivmBBWkX2FKrABsJqCZ04VDs6ULSBbB7GaBpLYQS7u8i4QQo57G2yfEmVkPmOXD7ivzmUlcr7nEZf7xbzCeNmXHeBX1dxbfgBaADhoLcAAAA%3D

### Expected behavior

Like in [TS playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBDAnmApnA3gMwhAvnTKCEOAcmwlICgqUAPSWBZNAMRzgF5nUJMCcAbhr1G8LDnyFiZCqSA)

```js
export { foo } from 'foo';
```

### Actual behavior

```js
export { };
```

### Version

1.2.122

### Additional context

_No response_

kwonoj:
**Description:**

Attempts to close #3114. 

This PR applies new mark to named export if

- named export comes from direct reexport (`export { some } from 'xxx'`)
- and export's ident collapses to existing ident

In case of #3114, existing logic treat reexported ident is same as imported one via `import type {...}`, ends up stripping its existence. However, if named export comes from direct reexport it is ensured it won't come from any ident in existing scope. PR attempts to assign new context for those cases.

Disclaimer this may not be the correct approach and I may did wrong.

- closes #3114

### DIFF
--- a/crates/swc_ecma_transforms_base/src/resolver/mod.rs
+++ b/crates/swc_ecma_transforms_base/src/resolver/mod.rs
@@ -3,7 +3,7 @@ use std::cell::RefCell;
 use swc_atoms::JsWord;
 use swc_common::{collections::AHashSet, Mark, SyntaxContext};
 use swc_ecma_ast::*;
-use swc_ecma_utils::{find_ids, Id};
+use swc_ecma_utils::{find_ids, ident::IdentLike, Id};
 use swc_ecma_visit::{as_folder, noop_visit_mut_type, Fold, VisitMut, VisitMutWith};
 
 #[cfg(test)]
@@ -106,6 +106,9 @@ struct Scope<'a> {
 
     /// All types declared in the scope
     declared_types: AHashSet<JsWord>,
+
+    /// Id of import declarations reexports ident need to check overlap
+    import_declarations: AHashSet<Id>,
 }
 
 impl<'a> Scope<'a> {
@@ -117,6 +120,7 @@ impl<'a> Scope<'a> {
             declared_symbols: Default::default(),
             hoisted_symbols: Default::default(),
             declared_types: Default::default(),
+            import_declarations: Default::default(),
         }
     }
 }
@@ -418,7 +422,21 @@ impl<'a> VisitMut for Resolver<'a> {
         // We need to analyze these identifiers for type stripping purposes.
         self.ident_type = IdentType::Binding;
         self.in_type = n.type_only;
+
         n.visit_mut_children_with(self);
+
+        for specifier in &n.specifiers {
+            match specifier {
+                ImportSpecifier::Default(ImportDefaultSpecifier { ref local, .. })
+                | ImportSpecifier::Named(ImportNamedSpecifier { ref local, .. })
+                | ImportSpecifier::Namespace(ImportStarAsSpecifier { ref local, .. })
+                    if n.type_only =>
+                {
+                    self.current.import_declarations.insert(local.to_id());
+                }
+                _ => {}
+            }
+        }
     }
 
     fn visit_mut_arrow_expr(&mut self, e: &mut ArrowExpr) {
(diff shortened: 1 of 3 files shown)


kdy1 (commented):
I think the parser is not a good place for such operation, because using `Mark` requires `swc_common::GLOBLAS` to be configured.

(And being able to use the parser without it is one of the explicit requests I received)

Is this not possible from `typescript::strip`?

kwonoj:
> Is this not possible from typescript::strip?

It is original place I started, I think it'll work if we try to ignore strip for the specific ident collapses. The reason I choose to update parser was to have a proper context for the ident _maybe_ useful to other transforms as it'll be guaranteed to be treated as different ident.

I'll give it another go and update PR if it works.

kdy1:
Oh, I see.
We apply resolver right after parsing, so it can be the place IMO.

kwonoj:
Cool, I'll try to peek resolver as well.

crates/swc_ecma_transforms_base/src/resolver/mod.rs
```diff
@@ -133,6 +136,7 @@ struct Resolver<'a> {
     handle_types: bool,
     in_type: bool,
     in_ts_module: bool,
+    import_declarations: AHashMap<Id, bool>,
```
kdy1, line 139: Can you add a comment describing what information is stored as `bool`?

crates/swc_ecma_transforms_base/src/resolver/mod.rs
```diff
@@ -418,7 +423,20 @@ impl<'a> VisitMut for Resolver<'a> {
         // We need to analyze these identifiers for type stripping purposes.
         self.ident_type = IdentType::Binding;
         self.in_type = n.type_only;
+
         n.visit_mut_children_with(self);
+
+        for specifier in &n.specifiers {
+            match specifier {
+                ImportSpecifier::Default(ImportDefaultSpecifier { ref local, .. })
+                | ImportSpecifier::Named(ImportNamedSpecifier { ref local, .. })
```
kdy1, line 432: I think we also have to check for `type` of `ImportNamedSpecifier`.

```ts
import { type foo, bar } from 'baz';

```

kdy1: Also, it would be nice to have a test case for the import namespace specifier with `type` keywortd.

crates/swc_ecma_transforms_base/src/resolver/mod.rs
```diff
@@ -846,7 +864,34 @@ impl<'a> VisitMut for Resolver<'a> {
         }
 
         // Phase 2.
-        stmts.visit_mut_children_with(self)
+        stmts.visit_mut_children_with(self);
+
+        for item in stmts.iter_mut() {
+            match item {
+                ModuleItem::ModuleDecl(ModuleDecl::ExportNamed(export)) if export.src.is_some() => {
+                    let mark = swc_common::Mark::fresh(swc_common::Mark::root());
```
kdy1, line 872: `swc_common::Mark` can be `Mark` as it's already imported.

crates/swc_ecma_transforms_base/src/resolver/mod.rs
```diff
@@ -846,7 +864,34 @@ impl<'a> VisitMut for Resolver<'a> {
...
+
+                    // `look back` exported specifiers if given export is reexport,
+                    // if there's collapsing ident in the same scope apply new context instead.
+                    // Refer https://github.com/swc-project/swc/issues/3114 for an example.
+                    for specifier in export.specifiers.iter_mut() {
+                        if let ExportSpecifier::Named(named_specifier) = specifier {
+                            if let Some(type_only) =
```
kdy1, line 879: We can use `if let Some(true) =` instead.

kdy1 (changes requested):
Thanks!

crates/swc_ecma_transforms_base/src/resolver/mod.rs
kwonoj: actually I could just use hashset instead.

crates/swc_ecma_transforms_base/src/resolver/mod.rs
kwonoj: I may misunderstood but `import { type ..}` seems just works? tried to add test cases here

https://github.com/swc-project/swc/pull/3130/files#diff-92b6d0b97f5a9184ecd099b4e07cf97b45aef1222a7da1d40f891ec385d8fc83R2

dsherret:
Does this also fix #2924?

kwonoj:
> Does this also fix #2924?

doesn't look like so. It seems like there's some other offender in mangle side, or my changes are not suffecient to deal with all cases. If this goes in, I'll try to peek mangle side if there's an easy fix.

kwonoj:
Closing in favor of https://github.com/swc-project/swc/pull/3139

PR #3136, opened 2021-12-28
fix(es/transforms): Fix async function parameter length

### LINKED ISSUE #3135: Incorrect async function arity (function.length)
### Describe the bug

It looks like swc async wrappers are messing with function.length results ([https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length)).

### Input code

```typescript
const myFunc = async (one, two, three) => {
    return 'irrelevant'
};

console.log('arity', myFunc.length)
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.115&code=H4sIAAAAAAAAAy2MQQqAIBAA775ibxaIHwg79g%2BRRQXbhW0rJPp7HprDzG0S06Gw9%2B2kBAHi0UcnJnSgNw8VQZwhrPAYGAjqKQS2imDDK5Ja8y7GpPHhhr5xnmyUqt26f%2BsbUtYyfzMfQzVrAAAA&config=H4sIAAAAAAAAA0WMSwrDMAxE76K1t93kDj2EcJXg4B8aBWqM714bUrKSmPdmOp3wtHWqrBBdH1o2%2FtJG4hPDa6hGbmoz2jlChiNjPcSWgtdksRTITR2lkMPe1pIvqaoAD%2BJ8xL855lAqn2sFnazVeVcllXyCxjN0lwPet216yfgBCKaI9boAAAA%3D

### Expected behavior

When running the transpiled code (I'm using node to run it), the console.log should print **arity 3**

### Actual behavior

When running the transpiled code (I'm using node to run it), the console.log prints **arity 0**

### Version

1.2.115

### Additional context

_No response_

magic-akari:
**Description:**
This pull request fixes the async/generator function parameter length and function name.
It seems that there are still some issues here that I can't fix right now.

The code uses `bind` to handle `this` issue.
But there is still the arguments/super/new.target issue to deal with. 

- hoisting variable/function may be better than bind. ( #2716)

**BREAKING CHANGE:**

- fix #3135


crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -737,6 +593,314 @@ impl VisitMut for MethodFolder {
...
+                        return_type: return_type.take(),
+                    },
+                };
+
+                let ref_ident = quote_ident!("ref");
+                let real_fn_ident = private_ident!(ref_ident.span, format!("_{}", ref_ident.sym));
+
+                let right = if let Some(this) = this {
+                    Expr::Call(CallExpr {
+                        span: span.take(),
```
kdy1, line 748: `span` is `Copy`, so `*span` will be better

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -905,25 +1078,27 @@ impl Actual {
...
                                 function: f,
                             })),
                             Stmt::Return(ReturnStmt {
                                 span: DUMMY_SP,
-                                arg: Some(Box::new(Expr::Ident(raw_ident.clone().unwrap()))),
+                                arg: Some(Box::new(Expr::Ident(function_ident.clone().unwrap()))),
                             }),
                         ]
                     } else {
+                        let ident = if function_ident.is_some() {
```
kdy1, line 1094: `function_ident.unwrap_or(binding_ident)` will be more readable.

kdy1 (changes requested):
Mostly LGTM. Thank you!
I find some tiny code style issues.

crates/swc/tests/tsc-references/unionAndIntersectionInference3_es5.1.normal.js
```diff
@@ -34,17 +34,22 @@ concatMaybe([
     3
 ], 4);
 // Repros from #32247
-var g = _asyncToGenerator(regeneratorRuntime.mark(function _callee(com) {
```
kdy1, line 37: This seems like a regression of optimization.

crates/swc/tests/fixture/issue-2248/es5/output/index.js
```diff
@@ -28,18 +28,23 @@ function _asyncToGenerator(fn) {
         });
     };
 }
-export var foo = _asyncToGenerator(regeneratorRuntime.mark(function _callee() {
```
kdy1, line 31: Seems like a regression.
I'm looking for the cause...

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -737,6 +593,314 @@ impl VisitMut for MethodFolder {
...
+                                stmts: vec![Stmt::Return(ReturnStmt {
+                                    span: expr.span(),
+                                    arg: Some(expr),
+                                })],
+                            },
+                        }),
+                        type_params: type_params.take(),
+                        return_type: return_type.take(),
+                    },
+                };
```
kdy1, line 741: I think we need optimizaiton for IIFE here.

kdy1 (changes requested):

crates/swc/tests/tsc-references/unionAndIntersectionInference3_es5.1.normal.js
magic-akari: Do you mean to add a pure comment?

```
/* #__PURE__ */
```

kdy1: I mean direct call of `_asyncToGenerator` without a wrapper function.

magic-akari: This is a trade-off.
If we don't return a new function, the function name will be lost.
[See the Babel playground.](https://babeljs.io/repl#?browsers=chrome%2050&build=&builtIns=false&corejs=3.6&spec=false&loose=false&code_lz=MYewdgzgLgBA5gLhgHgKoBoYCVMGUB8AFKALZKECUMAvPjAJJQCmATgIZQgvK6YbZ0APjACCEAJ5hgjVhy48-mLPiq0YABRYgSASwhM0dajDYSpKfjhgFi28qroz2nbrxiWhos9ObP5bjwokTW09A1QjOgBvACgYGCgACy0AdxhSAG4YgF8MoA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=true&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Ctypescript&prettier=false&targets=&version=7.16.6&externalPlugins=&assumptions=%7B%7D)
Maybe we should add a pure comment.

magic-akari: And, we need to ensure that the length of this g function is 1 before and after the conversion.

kdy1: Oh, I see. Thank you for the clarification!

kdy1 (approved):

PR #3145, opened 2021-12-29
fix(ecma/transform/compat): transform new.target in arrow

### LINKED ISSUE #2885: `new.target` in arrow function wrongly transformed
### Describe the bug

This is similar to #2811, the problem lies in chrome 46 support `new.target` but not `arrow`

### Input code

```typescript
function foo() {
  const a = (a) => new.target
}
```

### Config

```js
{
  jsc: {
    parser: {
      syntax: "ecmascript"
    }
  },
  env: {
    targets: {
      chrome: "46"
    },
    mode: "entry",
    coreJs: 3
  }
}
```

### Playground link

_No response_

### Expected behavior

should output
```js
function foo() {
  var _newtarget = new.target;

  var a = function a(_a) {
    return _newtarget;
  };
}
```

### Version

1.2.112

### Additional context

There are two solutions.

1. Remove arrow related transform in `new.target` and move it to `arrow`
   pros: it's clean and can be reused in other transforms
   cons: it will need `arrow` transform be placed before `new.target`
2. Make `new.target` aware arrow function don't have its own `new.target` and reuse it in other transforms
   cons: it's ugly

Austaras:

closes #2885 

Also why `new_target` transform not in `preset_env`?

### DIFF
--- a/crates/swc_ecma_utils/src/function.rs
+++ b/crates/swc_ecma_utils/src/function.rs
@@ -7,16 +7,25 @@ use swc_ecma_visit::{noop_visit_mut_type, VisitMut, VisitMutWith};
 pub struct WrapperState {
     this: Option<Ident>,
     args: Option<Ident>,
+    new_target: Option<Ident>,
 }
 
 impl WrapperState {
-    pub fn new(this: Option<Ident>, args: Option<Ident>) -> Self {
-        Self { this, args }
+    pub fn new(this: Option<Ident>, args: Option<Ident>, new_target: Option<Ident>) -> Self {
+        Self {
+            this,
+            args,
+            new_target,
+        }
     }
     pub fn to_decl(self) -> Vec<VarDeclarator> {
-        let Self { this, args } = self;
+        let Self {
+            this,
+            args,
+            new_target,
+        } = self;
 
-        let mut decls = Vec::with_capacity(2);
+        let mut decls = Vec::with_capacity(3);
         if let Some(this_id) = this {
             decls.push(VarDeclarator {
                 span: DUMMY_SP,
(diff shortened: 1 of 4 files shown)


kdy1:
Can you add it?
There's no good way to ensure that we used all compat transforms in preset-env

Austaras:
Added. I'm just making sure there isn't a specific reason not to

crates/swc_ecma_utils/src/function.rs
```diff
@@ -7,14 +7,15 @@ use swc_ecma_visit::{noop_visit_mut_type, VisitMut, VisitMutWith};
 pub struct WrapperState {
     this: Option<Ident>,
     args: Option<Ident>,
+    target: Option<Ident>,
```
kdy1, line 10: Can you add a comment that this `target` is `new.target`?

crates/swc_ecma_utils/src/function.rs
```diff
@@ -7,14 +7,15 @@ use swc_ecma_visit::{noop_visit_mut_type, VisitMut, VisitMutWith};
...
-    pub fn new(this: Option<Ident>, args: Option<Ident>) -> Self {
-        Self { this, args }
+    pub fn new(this: Option<Ident>, args: Option<Ident>, target: Option<Ident>) -> Self {
+        Self { this, args, target }
     }
     pub fn to_decl(self) -> Vec<VarDeclarator> {
-        let Self { this, args } = self;
+        let Self { this, args, target } = self;
 
         let mut decls = Vec::with_capacity(2);
```
kdy1, line 20: This can be 3 I guess?

crates/swc_ecma_utils/src/function.rs
```diff
@@ -54,7 +66,7 @@ impl WrapperState {
     }
 
     pub fn to_stmt_in_subclass(self) -> (Option<Stmt>, Option<Ident>) {
-        let Self { this, args } = self;
+        let Self { this, args, target } = self;
 
         let mut decls = Vec::with_capacity(2);
```
kdy1, line 71: This can be 3 I guess?

kdy1 (changes requested):
Almost LGTM. Thank you!

Austaras:
Done

kdy1 (approved):
LGTM. Thanks!

Issue #2853, opened 2021-11-23
swc can't handle unescaped `\0`


SukkaW:
### Describe the bug

I am developing a [Rollup](https://rollupjs.org/) plugin based on `swc` (using swc as a transformer plugin).

Rollup [uses `\0` as its `virtual module` identifier](https://rollupjs.org/guide/en/#conventions), e.g. rollup's commonjs module [uses `\0commonjsHelpers.js` as the name of its virtual module](https://github.com/rollup/plugins/blob/02fb349d315f0ffc55970fba5de20e23f8ead881/packages/commonjs/src/helpers.js#L15):

```js
import * as commonjsHelpers from "\0commonjsHelpers.js";
```

However, swc would throw `Syntax Error` on `\0`:

```
     Error: error: Unterminated string constant

  |
1 | import * as commonjsHelpers from "
  |                                  ^



Caused by:
    0: failed to process js file
    1: Syntax Error
```

### Input code

```typescript
// Minimum reproduction
// Check on repl.it: https://replit.com/@SukkaW/SweetSmallModulus

const { transformSync: swcTransform } = require('@swc/core');
const { transformSync: esbuildTransform } = require('esbuild');

const code = `
import * as commonjsHelpers from "\0commonjsHelpers.js";
`;

try {
  console.log(esbuildTransform(code, {}));
} catch (e) {
  console.trace(e);
}

try {
  console.log(swcTransform(code), {});
} catch (e) {
  console.trace(e);
}

try {
  console.log(swcTransform(`const a = "\0a"`), {});
} catch (e) {
  console.trace(e);
}
```

### Config

```json
{}
```

### Expected behavior

Shouldn't throw Syntax Error on `\0`, either escape it (no matter the given target) silently or leave it as it is.

### Version

1.2.111

PR #3147, opened 2021-12-29
test(es/parser): Ensures that null escapes are respected

jimberlage:
**Description:**

I took a quick look at https://github.com/swc-project/swc/issues/2853 to see if I could make a dent in it.

This adds tests to ensure that the lexer/parser does not result in `SyntaxError::UnterminatedStrLit` when encountering a string with `\0` in it.

I think the parser/lexer is behaving correctly.  My best guess is that some code that feeds data into the parser is stopping when it encounters a literal NULL character.  It seems that the parser does OK with `\0` characters, but it's getting the string `const a = "` (at least, that's my guess.)  I'm not familiar enough with the code to know where I should look for a fix for that.  Here's a more detailed repl.it that shows the parser behaving correctly vs. when it blows up:

https://replit.com/join/djxrxgfolk-jimberlage

You can see that

```javascript
transformSync(`const a = "\0a"`);
```

blows up, while

```javascript
transformSync(`const a = "\\0a"`);
```

results in `{ code: 'var a = "\\x00a";\n' }`.

Another thing that I held off on changing, but you can tackle if you want - looks like esbuild's transformSync results in `{ warnings: [], code: 'const a = "\\0a";\n', map: '' }` for the same code.

**Related issue (if exists):** https://github.com/swc-project/swc/issues/2853


kdy1 (approved):
Thanks!

PR #3162, opened 2021-12-31
fix(es/typescript): Declare the exported variables in namespace

### LINKED ISSUE #3158: Undefined reference inside namespaces.
### Describe the bug

If you were to execute the sample code, you'll notice it will crash because `foo` is undefined in function `xyz`.

Using the [TypeScript playground](https://www.typescriptlang.org/play?ts=4.5.4#code/HYQwtgpgzgDiDGEAEAVaAXJAoA3lpSEAHjAPYBOmAZgK7DzoCWpwSIARvABQCUSOSchHQ1yrAIwAGANxIAvlnyESFTPBZRqpUkgC8SWvSYte-QcNGsATDPmKCxMpQN0GzVkQCeAL15K8BARCImJsnKYAVAbavNJKCgpYaJoAdF6+PNJAA) it yields the correct result compiling the call to `foo()` as `Test.foo()`.

The expected behavior would be to have an output similar as `tsc`.
 

### Input code

```typescript
namespace Test 
{
  export function abc() { return 10; }

  export const foo = function() { return 20; }

  export function xyz()
  {
    return abc() * foo();
  }
}

Test.xyz();
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2021",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": false
}
```

### Playground link

https://play.swc.rs/?version=1.2.124&code=H4sIAAAAAAAAA8tLzE0tLkhMTlUISS0uUeDlqublUlBIrSjILypRSCvNSy7JzM9TSExK1tBUqFYoSi0pLcpTMDSwVqjl5UJSmZyfB9Sdlp%2BvYAvXhazDCEMH3OyKyioNTZAE2GYFmA6IlVogMzU0rUFSQP1gI0AO1QPrsgYAnXBSZb4AAAA%3D&config=H4sIAAAAAAAAA0WNQQ7DIAwE%2F%2BJzDm2O%2BUMfgagTEQFGLJWKEH%2Bvaalys72z40YnLG2NksngPCbUWMybNio1MWx2qdBCBXrajQd3XUw%2BuCjCWG%2FrXWMvAp7AQsFFt9chsxJSZuCKTDz8n%2BzqCvJ8jUP7%2FlOnVoLEE9Qv0Sw7PCb9q38AcUsrT74AAAA%3D

### Expected behavior

```js
// From TypeScript playground.
"use strict";
var Test;
(function (Test) {
    function abc() { return 10; }
    Test.abc = abc;
    Test.foo = function () { return 20; };
    function xyz() {
        return abc() * Test.foo(); // <-- Here foo is correctly handled by tsc.
    }
    Test.xyz = xyz;
})(Test || (Test = {}));
Test.xyz();
```

### Actual behavior

```js
// From SWC playground.
var Test;
(function(Test1) {
    function abc() {
        return 10;
    }
    Test1.abc = abc;
    Test1.foo = function() {
        return 20;
    };
    function xyz() {
        return abc() * foo(); // <-- Here foo will be undefined.
    }
    Test1.xyz = xyz;
})(Test || (Test = {
}));
Test.xyz();
```

### Version

1.2.124

### Additional context

_No response_

magic-akari:
**Description:**
This Pull Request follows the behavior of Babel, which is slightly different from the behavior of TypeScript Compiler.

```javascript
var Test;
(function(Test1) {
    function abc() {
        return 10;
    }
    Test1.abc = abc;
    var foo = Test1.foo = function() {
    //  ^^^~~~~~ Babel declares the variable here
        return 20;
    };
    function xyz() {
        return abc() * foo();
    }
    Test1.xyz = xyz;
})(Test || (Test = {
}));
Test.xyz();
```

TypeScript output:
```javascript
var Test;
(function (Test) {
    function abc() {
        return 10;
    }
    Test.abc = abc;
    Test.foo = function () {
        return 20;
    };
    function xyz() {
        return abc() * Test.foo();
        //              ^~~~~~~ tsc add the `Test` here
    }
    Test.xyz = xyz;
})(Test || (Test = {}));
Test.xyz();
```

**BREAKING CHANGE:**

- fix #3158


crates/swc_ecma_transforms_typescript/tests/fixture/issue-3158/2/1.md
```diff
@@ -0,0 +1,56 @@
+<!-- Thanks for submitting a pull request! Please provide enough information so that others can review your pull request. -->
+
```
kdy1, line 2: Can you remove this file?

kdy1:
I think it's better to follow behavior of `tsc` in general, but in this case behavior of babel is preferrable because it's much likely to be optimized.

crates/swc/tests/tsc-references/invalidInstantiatedModule_es2015.2.minified.js
```diff
@@ -1,5 +1,9 @@
-var M, M1, M2;
-(M = M1 || (M1 = {
-})).Point = class {
-}, M.Point = 1, (M2 || (M2 = {
+var M, M2;
+!function(M1) {
+    class Point {
+    }
+    M1.Point = Point;
+    var Point = M1.Point = 1;
```
kdy1, line 6: This is regression. There's a class named `Point` so `var Point` is wrong.

crates/swc/tests/tsc-references/invalidInstantiatedModule_es5.2.minified.js
```diff
@@ -6,7 +6,8 @@ var M, M2;
             if (!(instance instanceof Constructor)) throw new TypeError("Cannot call a class as a function");
         }(this, Point);
     };
-    M1.Point = Point, M1.Point = 1;
+    M1.Point = Point;
+    var Point = M1.Point = 1;
```
kdy1, line 10: Same here

kdy1 (changes requested):
Thank you!
I found some issues while checking tests.

crates/swc/tests/tsc-references/invalidInstantiatedModule_es2015.2.minified.js
magic-akari: https://github.com/swc-project/swc/blob/1aa55e616a7bc2ffa791600e908ddb10a4009e7b/crates/swc_ecma_parser/tests/typescript/tsc/invalidInstantiatedModule/input.ts#L1-L4

HI  @kdy1 ,
This seems to be expected behavior.
There is no way to export both `class Point` and `var Point`.
The comments also indicate that this is an error which should be reported.

magic-akari: Maybe we should report this error in lint

kdy1: Oh, I see. Thank you!

kdy1 (approved):
Thank you!

PR #3165, opened 2022-01-01
Test that string input sends null characters

jimberlage:
**Description:**

This ensures that the code responsible for reading JS strings handles NULL character literals by escaping them.

I'm pretty sure, given these tests, that the problem lies somewhere in wasm-bindgen, or was fixed somehow by recent changes.

**Related issue (if exists):** https://github.com/swc-project/swc/issues/2853

### DIFF
--- a/crates/swc_common/src/lib.rs
+++ b/crates/swc_common/src/lib.rs
@@ -70,6 +70,8 @@ pub mod serializer;
 pub mod source_map;
 pub mod sync;
 mod syntax_pos;
+#[cfg(test)]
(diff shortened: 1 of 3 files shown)


kdy1 (changes requested):
I don't think these tests are required.
null character is simply selected because I didn't think anyone to use it. (It's intentional)

kdy1:
See https://docs.rs/swc_ecma_parser/0.82.6/swc_ecma_parser/index.html#null-character-after-

PR #3171, opened 2022-01-02
chore(es/codegen): pretty output for empty lit object and empty statement 

magic-akari:
**Description:**

Here are some examples

before
```JavaScript 
function pick(obj, keys) {
    return keys.reduce((a, k)=>{
        a[k] = obj[k];
        return a;
    }, {
    });
}
```

after
```JavaScript 
function pick(obj, keys) {
    return keys.reduce((a, k)=>{
        a[k] = obj[k];
        return a;
    }, {});
}
```

---

before
```JavaScript
var Bar;
(function(Bar1) {
    function foo() {
        console.log("foo");
    }
    Bar1.foo = foo;
})(Bar || (Bar = {
}));
export { Bar as Bar };
```

after
```JavaScript 
var Bar;
(function(Bar1) {
    function foo() {
        console.log("foo");
    }
    Bar1.foo = foo;
})(Bar || (Bar = {}));
export { Bar as Bar };
```

---

**BREAKING CHANGE:**


kdy1 (changes requested):
Thanksø﻿

kdy1 (approved):

PR #3172, opened 2022-01-02
fix(es/minifier): obj, num and str comparison

cuyl:
Description:

optimize_lit_cmp does not compare `number` and `string` correctly.

input
```js
const a = 3;
const b = 4;
const c = "3";
const d = "4";
const e = {};
const f = {};
const g = true;
const h = false;
const j = null;
console.log(a === b, c === d, e === f, g === h, h === j);
```
previous out
```js
const a = 3;
const b = 4;
const c = "3";
const d = "4";
const e = {};
const f = {};
const g = true;
const h = false;
const j = null;
console.log(true, true, true, false, false);
```
current out
```js
const a = 3;
const b = 4;
const c = "3";
const d = "4";
const e;
const f;
const g = true;
const h = false;
const j = null;
console.log(false, false, false, ({}) === {}, false);
```

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/optimize/ops.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/ops.rs
@@ -78,28 +78,42 @@ where
 
     ///
     /// - `1 == 1` => `true`
+    /// - `1 == 2` => `false`
     pub(super) fn optimize_lit_cmp(&mut self, n: &mut BinExpr) -> Option<Expr> {
-        match n.op {
-            op!("==") | op!("!=") => {
-                // Abort if types differ, or one of them is unknown.
-                if n.left.get_type().opt()? != n.right.get_type().opt()? {
-                    return None;
-                }
-
+        if n.op != op!("==") && n.op != op!("!=") {
+            return None;
+        }
+        let flag = n.op == op!("!=");
+        let mut make_lit_bool = |value: bool| {
+            self.changed = true;
+            Some(Expr::Lit(Lit::Bool(Bool {
+                span: n.span,
+                value: flag ^ value,
+            })))
+        };
+        match (n.left.get_type().opt()?, n.right.get_type().opt()?) {
+            // Abort if types differ, or one of them is unknown.
+            (lt, rt) if lt != rt => {}
+            (Type::Obj, Type::Obj) => {}
+            (Type::Num, Type::Num) => {
+                let l = n.left.as_number().opt()?;
+                let r = n.right.as_number().opt()?;
+                tracing::debug!("Optimizing: literal comparison => num");
+                return make_lit_bool(l == r);
+            }
+            (Type::Str, Type::Str) => {
+                let l = &n.left.as_string().opt()?;
+                let r = &n.right.as_string().opt()?;
+                tracing::debug!("Optimizing: literal comparison => str");
+                return make_lit_bool(l == r);
+            }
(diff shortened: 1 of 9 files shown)


kdy1:
Can you run `UPDATE=1 cargo test` from `/crates/swc`?

cuyl:
@kdy1 updated

crates/swc_ecma_minifier/tests/terser/compress/comparing/lit_comparisons/config.json
```diff
@@ -0,0 +1,5 @@
+{
```
kdy1, line 1: Please move this to `tests/compress/fixture/issues/**`
`terser` directory is directory for tests ported from terser.

kdy1 (changes requested):

crates/swc_ecma_minifier/tests/terser/compress/comparing/lit_comparisons/config.json
cuyl: Ok, moved.

crates/swc_ecma_minifier/src/compress/optimize/ops.rs
```diff
@@ -78,28 +78,42 @@ where
...
+            })))
+        };
+        match (n.left.get_type().opt()?, n.right.get_type().opt()?) {
+            // Abort if types differ, or one of them is unknown.
+            (lt, rt) if lt != rt => {}
+            (Type::Obj, Type::Obj) => {}
+            (Type::Num, Type::Num) => {
+                let l = n.left.as_number().opt()?;
+                let r = n.right.as_number().opt()?;
+                tracing::error!("Optimizing: literal comparison => num");
```
kdy1, line 101: Can you change this to `debug!`?

crates/swc_ecma_minifier/src/compress/optimize/ops.rs
```diff
@@ -78,28 +78,42 @@ where
...
+                return make_lit_bool(l == r);
+            }
+            (Type::Str, Type::Str) => {
+                let l = &n.left.as_string().opt()?;
+                let r = &n.right.as_string().opt()?;
+                tracing::error!("Optimizing: literal comparison => str");
```
kdy1, line 107: Same here.

crates/swc_ecma_minifier/src/compress/optimize/ops.rs
```diff
@@ -78,28 +78,42 @@ where
...
-
-                let value = if n.op == op!("==") { l == r } else { l != r };
-
-                tracing::debug!("Optimizing: literal comparison => bool");
-                self.changed = true;
-                return Some(Expr::Lit(Lit::Bool(Bool {
-                    span: n.span,
-                    value,
-                })));
+                tracing::error!("Optimizing: literal comparison => bool");
```
kdy1, line 113: Same here.

kdy1 (changes requested):
Mostly LGTM. Thanks!

crates/swc_ecma_minifier/src/compress/optimize/ops.rs
cuyl: Opps, I just use error to debug. moved back to `debug!` now.

kdy1 (approved):
LGTM. Thank you!

PR #3178, opened 2022-01-03
mod(es): change definition of ast

Austaras:
Some cases (optional_chaining__optional_super_property and optional_chaining__optional_super_property_class) for parser will be invalid after this change, what should I do? Just delete it?


kdy1:
Yeah, I think so.

dsherret:
Is there an outline of what changes are taking place and why the changes are taking place somewhere? My browser keeps crashing looking at the diff 😅

Austaras:
> Is there an outline of what changes are taking place and why the changes are taking place? My browser keeps crashing looking at the diff 😅

See [here](https://github.com/swc-project/swc/discussions/3153) or first commit

crates/swc_ecma_transforms_base/src/hygiene/usage_analyzer.rs
```diff
@@ -580,8 +580,14 @@ impl Visit for UsageAnalyzer<'_> {
     fn visit_member_expr(&mut self, e: &MemberExpr) {
         e.obj.visit_with(self);
 
-        if e.computed {
-            e.obj.visit_with(self);
```
Austaras, line 584: I suppose this should be `e.prop.visit_with` @kdy1 ?

Austaras:
Also, is there a difference between `optimize_property_of_member_expr` and `handle_known_computed_member_expr`? They seems to have identical functionality

crates/swc_ecma_transforms_base/src/hygiene/usage_analyzer.rs
kdy1: Yes you are right

Austaras: Oh, let's spin a separate pr real quick to avoid massive snapshot overwrite...

kdy1:
Hmm... Maybe it's a duplicate logic.
Can you remove one of them?

Austaras:
图片

crates/swc_ecma_ast/src/expr.rs
```diff
@@ -63,6 +64,10 @@ pub enum Expr {
     #[tag("MemberExpression")]
     Member(MemberExpr),
 
+    #[tag("SuperPropExpression")]
+    #[is(name = "super_")]
+    Super(SuperPropExpr),
```
kdy1, line 69: I think `SuperProp` would be better, as `Expr::Super` looks like `super` instead of `super.foo`.

kdy1 (commented):
I didn't fully reviewed yet, but I found one concern which will invalidate lots of diffs.

crates/swc_ecma_ast/src/expr.rs
Austaras: If name could remain the same I could push an commit to rename this, and rebase it before update parser change

Austaras: Well the change is not that big, I could handle it mannually

kdy1: I disabled auto-merge of parser PR. Thanks!

Austaras:
> Hmm... Maybe it's a duplicate logic. Can you remove one of them?

Seems like remove either one of them would break some tests, I will return to this later

node-swc/src/binding.js
```diff
@@ -5,19 +5,9 @@ const { platform, arch } = process
 
 let nativeBinding = null
 let localFileExisted = false
+let isMusl = false
```
kdy1: This seems like a wrong change.

crates/swc_ecma_transforms_compat/src/es2015/regenerator/mod.rs
```diff
@@ -338,8 +338,8 @@ impl Regenerator {
         let ctx = private_ident!("_ctx");
         let mut handler = CaseHandler::new(&ctx);
 
-        f.body
-            .visit_mut_with(&mut FnSentVisitor { ctx: ctx.clone() });
+        // f.body
```
kdy1: Why did you commented this out?

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
```diff
@@ -554,25 +529,21 @@ impl OptChaining {
...
-                                        left: PatOrExpr::Pat(Box::new(Pat::Ident(
-                                            this_obj.clone().into(),
-                                        ))),
-                                        right: obj.obj.take().expect_expr(),
-                                    })
-                                    .as_obj(),
-                                    prop: obj.prop.take(),
-                                    computed: obj.computed,
-                                })),
+                                Expr::Member(obj @ MemberExpr { .. }) => {
```
kdy1: ` @ MemberExpr { .. }` seems redundant

kdy1 (changes requested):
Whoa, thank you so much for the hard work.

I used few hours to review this :)

crates/swc_ecma_transforms_compat/src/es2020/opt_chaining.rs
Austaras: It is

Austaras:
Thank you for take your time to review this

kdy1 (approved):

Issue #3167, opened 2022-01-02
[Meta] Wasm-based experimental plugin support


kwonoj:
### Describe the feature

Originating from https://github.com/swc-project/swc/issues/2635#issue-1043151813 . 

It looks like we are able to push forward wasm based plugin, this issue is filed for tracking its subtasks to officially release features. Still it'll be exposed under `experimental` flag to iterate implementaion with possibly breaking changes.

- [x] runtime replacement (https://github.com/swc-project/swc/pull/3196 / https://github.com/wasmerio/wasmer/issues/2324) : hope to use wasmer.
- [x] deprecate `abi_stable` (https://github.com/swc-project/swc/pull/3179)
- [x] deprecate `plugin_ast`
- [x] `swc_common/plugin` to utilize https://wasmerio.github.io/wasmer/crates/doc/wasmer/macro.imports.html
- [x] support loading / bytecode caching wasm from `swc_plugin_runner` (https://github.com/swc-project/swc/pull/3179)
- [x] configuration support for bytecode cache location (https://github.com/swc-project/swc/pull/3179)
- [ ] compiled bytecode cache support
- [ ] swc_cli supports for plugin project scafolding (we'll consolidate https://github.com/swc-project/swc-dev/blob/main/src/main.rs into `swc`'s subcommand)
- [x] writing short adr (https://github.com/swc-project/swc/pull/3249)
- [ ] Cache expiration strategy (Probably will not be implemented in experimental phase, need bit more thought. May spin off new issue later).
- [x] Optimize wasm loading strategy: current impl (https://github.com/swc-project/swc/blob/fc4c6708f24cda39640fbbfe56123f2f6eeb2474/crates/swc/src/plugin.rs#L19) is incorrect as it is being called per each file transform operation. We should read plugin binary only once per whole transform. (https://github.com/swc-project/swc/pull/3186)
- [x] println / dbg! or tracing support (https://github.com/swc-project/swc/pull/3271)
- [x] error handling (https://github.com/swc-project/swc/pull/3300)

kwonoj:
Subtasks:

- https://github.com/swc-project/swc/issues/3311
- https://github.com/swc-project/swc/issues/3309
- https://github.com/swc-project/swc/issues/3308
- https://github.com/swc-project/swc/issues/3307

Now plugin implementation reached `at least working` stage, further tasks will be filed separately instead of using meta issue. This issue will be closed soon once a few remaining tasks here make some progress.

[...]

PR #3179, opened 2022-01-03
refactor(swc/plugins): load wasm plugin with cache (#3167 part 1)

kwonoj:
**Description:**

This PR is part 1 of #3167. PR contains a minimal set of changes, including

- deprecates abi_stable usage (in swc/plugins)
- revise plugin_runner to load wasm binary with cache support.

PR doesn't run actual transforms from loaded binary yet, only `loading` and compiling part are included.

**BREAKING CHANGE:**

Techinically this is breaking changes to abi_stable based plugin, but under experimental flag we expected that'll occur.

#3167

### DIFF
--- a/crates/swc/src/config/mod.rs
+++ b/crates/swc/src/config/mod.rs
@@ -421,7 +421,7 @@ impl Options {
                 syntax.typescript()
             ),
             lint_to_fold(swc_ecma_lints::rules::all()),
-            crate::plugin::plugins(experimental.plugins),
+            crate::plugin::plugins(experimental),
             custom_before_pass(&program),
             // handle jsx
             Optional::new(
@@ -971,6 +971,13 @@ pub struct JscExperimental {
     /// If true, keeps import assertions in the output.
     #[serde(default)]
     pub keep_import_assertions: bool,
+    /// Location where swc may stores its intermediate cache.
+    /// Currently this is only being used for wasm plugin's bytecache.
+    /// Path should be absolute directory, which will be created if not exist.
+    /// This configuration behavior can change anytime under experimental flag
+    /// and will not be considered as breaking changes.
+    #[serde(default)]
+    pub cache_root: Option<String>,
 }
 
 impl Merge for JscExperimental {
--- a/crates/swc/src/plugin.rs
+++ b/crates/swc/src/plugin.rs
@@ -14,10 +14,16 @@ use swc_ecma_visit::{noop_fold_type, Fold};
 #[serde(deny_unknown_fields, rename_all = "camelCase")]
 pub struct PluginConfig(String, serde_json::Value);
 
-pub fn plugins(names: Option<Vec<PluginConfig>>) -> impl Fold {
+pub fn plugins(config: crate::config::JscExperimental) -> impl Fold {
     #[cfg(feature = "plugin")]
     {
-        RustPlugins { plugins: names }
+        let cache_root =
+            swc_plugin_runner::resolve::resolve_plugin_cache_root(config.cache_root).ok();
+
+        RustPlugins {
+            plugins: config.plugins,
+            plugin_cache: cache_root,
+        }
     }
 
     #[cfg(not(feature = "plugin"))]
@@ -28,11 +34,15 @@ pub fn plugins(names: Option<Vec<PluginConfig>>) -> impl Fold {
 
 struct RustPlugins {
     plugins: Option<Vec<PluginConfig>>,
+    /// TODO: it is unclear how we'll support plugin itself in wasm target of
+    /// swc, as well as cache.
+    #[cfg(feature = "plugin")]
+    plugin_cache: Option<swc_plugin_runner::resolve::PluginCache>,
 }
 
 impl RustPlugins {
     #[cfg(feature = "plugin")]
-    fn apply(&self, mut n: Program) -> Result<Program, anyhow::Error> {
+    fn apply(&mut self, mut n: Program) -> Result<Program, anyhow::Error> {
         use anyhow::Context;
 
         if let Some(plugins) = &self.plugins {
@@ -42,7 +52,13 @@ impl RustPlugins {
 
                 let path = swc_plugin_runner::resolve::resolve(&p.0)?;
 
-                n = swc_plugin_runner::apply_js_plugin(&p.0, &path, &config_json, n)?;
+                n = swc_plugin_runner::apply_js_plugin(
+                    &p.0,
+                    &path,
+                    &mut self.plugin_cache,
+                    &config_json,
+                    n,
+                )?;
             }
         }
 
--- a/crates/swc/src/plugin.rs
+++ b/crates/swc/src/plugin.rs
@@ -14,14 +14,59 @@ use swc_ecma_visit::{noop_fold_type, Fold};
 #[serde(deny_unknown_fields, rename_all = "camelCase")]
 pub struct PluginConfig(String, serde_json::Value);
 
+struct PluginBinary(String, Vec<u8>, String);
+
 pub fn plugins(config: crate::config::JscExperimental) -> impl Fold {
     #[cfg(feature = "plugin")]
     {
+        use anyhow::Context;
+
+        let mut plugin_binaries = vec![];
+        let mut plugin_errors = vec![];
+
+        if let Some(plugins) = &config.plugins {
+            for p in plugins {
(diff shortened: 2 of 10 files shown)


deny.toml
```diff
@@ -74,6 +74,7 @@ unlicensed = "deny"
 allow = [
   "MIT",
   "Apache-2.0",
+  "Apache-2.0 WITH LLVM-exception",
```
kwonoj, line 77: This is due to transitive dependencies around cranelift-* licenses. It is bit cryptic to read, but peeking https://bugzilla.mozilla.org/show_bug.cgi?id=1476798 I think it's reasonably safe to allow.

crates/swc_plugin_runner/tests/integration.rs
```diff
@@ -31,6 +31,8 @@ fn build_plugin(dir: &Path) -> Result<PathBuf, Error> {
     Err(anyhow!("Could not find built plugin"))
 }
 
+//TODO: https://github.com/swc-project/swc/issues/3167
```
kwonoj, line 34: Test is skipped as test fixture doesn't produces wasm plugin for testing yet.

crates/swc_macros_common/Cargo.toml
```diff
@@ -15,4 +15,4 @@ quote = "1"
 
 [dependencies.syn]
 features = ["derive", "visit", "parsing", "full", "printing", "extra-traits"]
-version = "=1.0.65"
+version = "=1.0.84"
```
kwonoj, line 18: wasmer-derive have conflict semver requirements to this which blocks compilation.

crates/swc/src/config/mod.rs
```diff
@@ -971,6 +971,13 @@ pub struct JscExperimental {
     /// If true, keeps import assertions in the output.
     #[serde(default)]
     pub keep_import_assertions: bool,
+    /// Location where swc may stores its intermediate cache.
+    /// Currently this is only being used for wasm plugin's bytecache.
+    /// Path should be absolute directory, which will be created if not exist.
+    /// This configuration behavior can change anytime under experimental flag
+    /// and will not be considered as breaking changes.
+    #[serde(default)]
+    pub cache_root: Option<String>,
```
kwonoj, line 980: I don't have clear pictures around how we'll expose configuration around cache yet, so put it under experimental flag. Easiest way is as current does accepting a string to the path - but maybe we want to some level of configuration (i.e what should we do if we want to opt in mem cache? or control over each cache targets instead of root?) instead.

crates/swc_plugin_runner/src/lib.rs
```diff
@@ -1,41 +1,87 @@
...
-    config_json: &str,
-    program: Program,
-) -> Result<Program, Error> {
-    (|| -> Result<_, Error> {
-        let plugin_rt = swc_common::plugin::get_runtime_for_plugin(plugin_name.to_string());
+/// Load plugin from specified path.
+/// If cache is provided, it'll try to load from cache first to avoid
+/// compilation.
+fn load_plugin(plugin_path: &Path, cache: &mut Option<PluginCache>) -> Result<Instance, Error> {
+    let wasmer_store = Store::default();
```
kdy1, line 14: I think it's better to cache file content.

`SwcJsPluginRef` did this internally, but now we are reading the `.wasm` file every time `apply_js_plugin` is called.

crates/swc/src/config/mod.rs
kdy1: Yeah, I'm also not sure about this at the moment,

kdy1: I think it should be `plugin_cache_name`. (Or something includes `plugin`)

kdy1 (changes requested):

crates/swc/src/config/mod.rs
kwonoj: I was thinking more broad to not to include `plugin` explicitly:

```
{
..
cache_root: //some_path
}
```

then there can be

```
- cache_root
   - plugins
   - some other for future, i.e bundle, etcs..
```

even if we have single cache forever for plugin only, I think it's better to have root level config instead.

crates/swc_plugin_runner/src/lib.rs
kwonoj: How does cache invalidation works if we store full file contents and do not read again? Having hash from full file is in my opinion most correct way to ensure cache can be invalidated when file changes, like having new versions. Other way to calculate hash (name + path + mtime...) is quite error prone afaik.

There is def. overhead around reading file over and over again, but I expect it's somewhat acceptable for the first implementation and can optimize later if we heavily offended by those.

kwonoj: (fyi, wasmer itself does same for the hashing : https://github.com/wasmerio/wasmer/blob/d5a444d61f74babc438cf669470ca7e62d891a46/lib/cli/src/commands/run.rs#L225-L246)

kdy1: > we store full file contents and do not read again

This was true for previous version, and I'm not sure if it's useful to support updating plugins within the same process.
Hmm... Might be useful if we have a server mode.

kwonoj: No, current cache targets filesystem so this is permanant. So it becomes matter to get right hash to load up compiled binary for the subsequent process invocation.

For example, if we incorrectly calculate hash like `name + path` which'll be same over different version, user'll see unexpected behavior by we serve old versions of cache. mtime is being used for those purpose for some cases, but that's also error prone on the CI systems.

kdy1: What do you mean by `current`?
This PR or previous one?

kwonoj: I meant this PR.

Pseudo workflow like

1. swc runs plugin A for the first time, cache bytecode (and exit)
2. next time swc runs, swc finds hash for A
2.1 cache hits - returns compiled bytecode from filesystem cache
2.2 cache miss - consider this is same as 1, go back to 1

The part in discussion is incorrect 2.1 
- plugin A.1 (new version of A) hash calculated to plugin A, swc returns compiled bytecode of A instead.

kwonoj: Also, we need to cache `dual` if we want to store file contents as well - first we need to store raw file contents, secondly need to store compiled bytecode which is actual cached contents we need to serve to avoid compilation.

kdy1: Ah, I was talking about `std::fs::read(plugin_path)` being called multiple times, within a single invocation.

kwonoj: ugh I misunderstood your comment completely then 😅 

But also - how frequent cases will there be if the same plugin will be specified more than once in a config? Will we allow it even? Unless I misunderstand again, reading file should occur for any unique plugin binaries.

kdy1: The problem is not a single plugin being specified multiple times. If we compile 3000 files, this will read the wasm file at least 3000 times.

kwonoj: oh, right 🤦 I somehow thought this is being called per each loading phase, not for the each transform.

crates/swc/src/plugin.rs
```diff
@@ -14,10 +14,61 @@ use swc_ecma_visit::{noop_fold_type, Fold};
...
+pub fn plugins(config: crate::config::JscExperimental) -> impl Fold {
     #[cfg(feature = "plugin")]
     {
-        RustPlugins { plugins: names }
+        use anyhow::Context;
+
+        let mut plugin_binaries = vec![];
+        let mut plugin_errors = vec![];
+
+        if let Some(plugins) = &config.plugins {
```
kwonoj, line 27: Somewhat ugly, but thoughts? Peeking paths I realized there are 2 things occurring per each process for 1. Config for plugin 2. Reading file itself.

Instead of trying to cache both individually inside of runner, I put those upfront and pass into each plugin invocation. This'll gets rid of internal cache mangement and makes cache lookup to plugin contents (with config) to zero, but as you could see in code error handling gets little dirty due to lacks of my knowledge.

@kdy1

kdy1: Seems much better than the previous one to me.

kdy1 (approved):

PR #3183, opened 2022-01-03
feat: improve DependencyDescriptor to combine import_assertions with dynamic_import_assertions

### LINKED ISSUE #3137: Improve `DependencyDescriptor` to combine `import_assertions` with `dynamic_import_assertions`
### Describe the feature

The type of `DependencyDescriptor` is error prone because it does not force you to check the correct value:

```rs
struct DependencyDescriptor {
    // ...etc...

    /// Import assertions for this dependency.
    /// NOTE: it's filled only for static imports and exports.
    pub import_assertions: HashMap<String, String>,
    /// Import assertions for dynamic imports. Present if `is_dynamic` is `true`
    /// and `kind` is not `DependencyKind::Require`.
    pub dynamic_import_assertions: Option<DynamicImportAssertions>,
}

#[derive(Clone, Debug, Eq, PartialEq)]
pub enum DynamicImportAssertions {
    /// The set of assertion keys could not be statically analyzed.
    Unknown,
    /// The set of assertion keys is statically analyzed, though each respective
    /// value may or may not not be.
    Known(HashMap<String, DynamicImportAssertion>),
}
```

Instead it would be better to do something along the lines of:

```rs
struct DependencyDescriptor {
    // ...etc...

    /// Import assertions for this dependency when not `DependencyKind::Require`.
    pub import_assertions: Option<ImportAssertions>,
}

#[derive(Clone, Debug, Eq, PartialEq)]
pub enum ImportAssertions {
    /// The set of assertion keys could not be statically analyzed.
    Unknown,
    /// The set of assertion keys is statically analyzed, though each respective
    /// value may or may not not be for dynamic imports.
    Known(HashMap<String, ImportAssertion>),
}

#[derive(Clone, Debug, Eq, PartialEq)]
pub enum ImportAssertion {
    /// The value of this assertion could not be statically analyzed.
    Unknown,
    /// The value of this assertion is a statically analyzed string.
    Known(String),
}
```

_No response_

### Additional context

_No response_

dsherret:
**Description:**

This makes the `DependencyDescriptor` interface a little less error prone by combining `import_assertions` with `dynamic_import_assertions`.

**BREAKING CHANGE:** Combines `dynamic_import_assertiosn` with `import_assertions` in `DependencyDescriptor`

**Related issue (if exists):** https://github.com/swc-project/swc/issues/3137

Closes #3137

### DIFF
--- a/crates/swc_ecma_dep_graph/src/lib.rs
+++ b/crates/swc_ecma_dep_graph/src/lib.rs
@@ -30,7 +30,9 @@ pub enum DependencyKind {
 }
 
 #[derive(Clone, Debug, Eq, PartialEq)]
-pub enum DynamicImportAssertion {
+pub enum ImportAssertion {
+    /// There was not an import assertion.
+    None,
     /// The value of this assertion could not be statically analyzed.
     Unknown,
     /// The value of this assertion is a statically analyzed string.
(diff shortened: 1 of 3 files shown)


crates/swc_ecma_dep_graph/src/lib.rs
```diff
@@ -225,107 +227,115 @@ impl<'a> Visit for DependencyCollector<'a> {
...
+    ImportAssertions::Known(import_assertions)
 }
 
 /// Parses import assertions from the second arg of a dynamic import.
-fn parse_dynamic_import_assertions(arg: Option<&ast::ExprOrSpread>) -> DynamicImportAssertions {
+fn parse_dynamic_import_assertions(arg: Option<&ast::ExprOrSpread>) -> ImportAssertions {
+    let arg = match arg {
+        Some(arg) => arg,
+        None => return ImportAssertions::None,
+    };
```
dsherret, line 262: The nesting was out of control in this function so I reduced it.

kdy1 (approved):
LGTM. Thank you!

PR #3186, opened 2022-01-04
refactor(swc/plugin): avoid redundant filesystem read

kwonoj:
**Description:**

Previous implementation https://github.com/swc-project/swc/blob/fc4c6708f24cda39640fbbfe56123f2f6eeb2474/crates/swc/src/plugin.rs#L19-L44 attmpts to avoid reading filesystem was not correct: `plugins` will be called per transform, so bubbling up won't make any difference.

This PR falls back to naive hashmap based cache based on the path of plugin instead. As noted in comments, this does not consider few known edge cases we may need to deal later.

#3167


crates/swc_plugin_runner/src/lib.rs
```diff
@@ -9,7 +17,40 @@ pub mod resolve;
...
+fn load_plugin(plugin_path: &Path, cache: &mut Option<PluginCache>) -> Result<Instance, Error> {
+    static BYTE_CACHE: Lazy<Mutex<AHashMap<PathBuf, Arc<Vec<u8>>>>> =
+        Lazy::new(|| Default::default());
+
+    // TODO: This caching streategy does not consider few edge cases.
+    // 1. If process is long-running (devServer) binary change in the middle of
+    // process won't be reflected.
+    // 2. If reading binary fails somehow it won't bail out but keep retry.
+    let module_bytes_key = plugin_path.to_path_buf();
+    let module_bytes = if let Some(cached_bytes) = BYTE_CACHE.lock().get(&module_bytes_key).cloned()
```
kdy1, line 39: I think type of this can be `Arc<Vec<u8>>>` instead of `Result` as the cons and alt produce identical type.

kdy1 (changes requested):

crates/swc_plugin_runner/src/lib.rs
kwonoj: yes, it's my oversight.

kdy1 (approved):
LGTM. Thanks!

PR #3188, opened 2022-01-04
refactor(swc): deprecate abi_stable, plugin_ast (#3167 part 2)

kwonoj:
**Description:**

This PR purely focused to delete some of existing code, mostly related with `abi_stable` and `plugin_ast`. Due to dependencies across packages it was not trivial to take out small portion of codes only, in result PR removes somewhat lot. 

There maybe some accidental deletion but if that's the case we can easily recover.

**BREAKING CHANGE:**
Same as part 1 this'll keep breaks old abi_stable based plugins. new plugin impl is not complete yet.

#3167


kdy1 (approved):
Thanks!

PR #3192, opened 2022-01-04
fix(visit): noop_visit_type should still visit typescript code that executes at runtime

dsherret:
**Description:** These are not technically types as they're emitted at runtime.

**BREAKING CHANGE:** noop_visit_type and noop_visit_mut_type macros all visit TypeScript runtime declarations now.


crates/swc_ecma_transforms_base/tests/ts-resolver/internalModules/codeGeneration/importStatements/output.ts
```diff
@@ -1,28 +1,28 @@
 module A {
-    export class Point__2 {
```
dsherret, line 2: Actually this might be wrong? I will look into it.

kdy1:
They should be translated to js before emitting, so it's better to skip them to reduce binary size.

kdy1 (changes requested):
What's the usecase of this?

dsherret:
@kdy1 yeah this doesn't seem exactly right.

When analyzing TS code using `Visit` the `noop_visit_type` will cause it to stop analyzing code with stuff like `const a = <SomeType> (otherCodeHere)` because of the TS assertion. Also, it doesn't analyze code in namespace declarations. Maybe we just shouldn't be using `noop_visit_type` and I will create a separate macro in deno_lint that still visits these nodes that contain runtime code.

dsherret:
@kdy1 I changed this to not do it for `fold`, but I think it makes sense for `noop_visit_type`. People writing visitors might still want runtime code that's in a TS node to be analyzed when using `noop_visit_type`.

Maybe I'm mistaken about the purpose of `Visit` and `noop_visit_type`. It would be trivial for us to add our own macro that does most of this for deno_lint so just let me know what would be best.

Also, fixed the files that were being changed to not be.

kdy1 (commented):
I think this is fine, but I'll wait for binary size CI action to ensure.
Thanks!

kdy1 (approved):
Previous: 47943160 = (45.7221603 MB)
After: 48138608 = (45.9085541 MB)


I think it's fine, thanks!

PR #3196, opened 2022-01-05
refactor(swc/plugin): replace wasm runtime

kwonoj:
**Description:**
We started plugin implememtation with wasmer as runtime, but looks like it doesn't support all of the platforms swc currently targets.

PR attempts to replace runtime to `wasmtime` to see if it can be used for those platforms. This PR attempts to verify target support only, so replacement to wasmtime actually doesn't contain correct code to utilize plugins. It should followup after we confirm target support.

**BREAKING CHANGE:**

https://github.com/wasmerio/wasmer/issues/2324


rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2022-01-02
\ No newline at end of file
+nightly-2021-12-15
```
kwonoj, line 1: due to https://github.com/bytecodealliance/rustix/issues/142 ; once wasmtime bumps up its dependencies we can revert this.

kdy1 (approved):
Thank you!

PR #3199, opened 2022-01-06
feat(swc/plugin_runner): pass serialized ast (#3167 part3)

kwonoj:
**Description:**

This PR makes plugin_runner actually passes serialized ast into wasm plugins. It doesn't do any transform yet, but re-enables integration testing so we can at least verify instantiation of plugin won't fail. I wanted to naively confirm if ast gets correctly passed, but realized I can't use `println` or `dbg!()` inside of guest (wasm plugin) side. This is probably a future homework.

PR also includes some bikeshedding - `swc_plugin` becomes main crate for the authoring plugin. It reexports few for the conveniences. `swc_plugin_js_api` is not actively being used for those reason at the moment.

#3167


tests/rust-plugins/swc_internal_plugin/Cargo.toml
```diff
@@ -5,16 +5,18 @@ edition = "2021"
 name = "swc_internal_plugin"
 publish = false
 version = "0.1.0"
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [lib]
-crate-type = ["rlib", "cdylib"]
+crate-type = ["cdylib"]
+
+# swc_atoms -> string_cache -> parking_lot -> https://github.com/Amanieu/parking_lot/issues/269
```
kwonoj, line 12: this'll be some headache if this won't be resolved in our dep tree in the future when we need to stablize scaffolding template.

crates/swc_plugin_runner/src/lib.rs
```diff
@@ -96,14 +97,42 @@ fn load_plugin(plugin_path: &Path, cache: &mut Option<PluginCache>) -> Result<In
...
+            };
 
             Instance::new(&module, &import_object).context("Failed to create plugin instance")
         }
         Err(err) => Err(err.into()),
     };
 }
 
+fn copy_memory_to_instance(instance: &Instance, program: &Program) -> Result<(i32, u32), Error> {
+    let alloc = instance.exports.get_native_function::<u32, i32>("alloc")?;
```
kwonoj, line 114: this may not be required, instead may need to use https://docs.rs/wasmer/2.0.0/wasmer/struct.Memory.html#method.grow instead. I'll double check later.

crates/swc_plugin/Cargo.toml
```diff
@@ -11,13 +11,10 @@ version = "0.21.0"
...
-abi_stable = "0.10.3"
 anyhow = "1.0.41"
 serde = {version = "1.0.126", features = ["derive"]}
 serde_json = "1.0.64"
-swc_atoms = {version = "0.2.7", path = "../swc_atoms"}
 swc_common = {version = "0.16.0", path = "../swc_common", features = ["plugin-mode"]}
-swc_ecma_ast = {version = "0.62.0", path = "../swc_ecma_ast"}
-swc_ecma_visit = {version = "0.48.0", path = "../swc_ecma_visit"}
-swc_node_base = {version = "0.5.1", path = "../swc_node_base"}
-swc_plugin_js_api = {version = "0.8.0", path = "../swc_plugin_js_api"}
```
kdy1, line 23: Can you remove `swc_plugin_js_api`?

crates/swc_plugin_runner/Cargo.toml
```diff
@@ -17,12 +17,13 @@ parking_lot = "0.11"
 serde = {version = "1.0.126", features = ["derive"]}
 serde_json = "1.0.64"
 swc_atoms = {version = "0.2.7", path = '../swc_atoms'}
-swc_common = {version = "0.16.0", path = "../swc_common", features = ["plugin-rt"]}
-swc_ecma_ast = {version = "0.62.0", path = "../swc_ecma_ast"}
+swc_common = {version = "0.16.0", path = "../swc_common", features = ["plugin-mode"]}
```
kdy1, line 20: This should be `plugin-rt`.

crates/swc_plugin/src/lib.rs
```diff
@@ -1,39 +1,31 @@
...
+pub use swc_ecma_ast::*;
+pub use swc_ecma_visit::*;
 
-use abi_stable::std_types::{RResult, RStr, RString};
-use anyhow::Context;
-use serde::de::DeserializeOwned;
-/// Reexported for convenience.
-pub use swc_plugin_js_api::*;
+#[cfg(target_arch = "wasm32")]
+pub mod memory {
```
kdy1, line 7: I think it's better to have `#[doc(hidden)]`.

crates/swc_plugin_runner/src/lib.rs
```diff
@@ -96,14 +97,42 @@ fn load_plugin(plugin_path: &Path, cache: &mut Option<PluginCache>) -> Result<In
 
     return match module {
         Ok(module) => {
-            let import_object = imports! {};
+            let memory = Memory::new(&wasmer_store, MemoryType::new(1, None, false))?;
```
kdy1, line 100: What does this code do?

kdy1 (changes requested):

crates/swc_plugin/src/lib.rs
kwonoj: I'm somehow seeing weird invalid memory reference err when running `test -p lib`, so guarded with target arch instead. I'll dig around memoryview's page instead for the future PRs.

kdy1: I mean additional `#[doc(hidden)]` on the module.

crates/swc_plugin_runner/src/lib.rs
kwonoj: This creates memory `instance` for the given wasm instance we runs. then expose it into guest (wasm) instance via https://docs.rs/wasmer/2.0.0/wasmer/#externs.

crates/swc_plugin/src/lib.rs
kwonoj: Got it, thanks for the clarification.

crates/swc_plugin/Cargo.toml
kwonoj: Sure thing!

crates/swc_plugin_runner/Cargo.toml
kwonoj: Updated - currently afaik feature definition for rt === mode. maybe we can just consolidate later depends on usecases?

kdy1: No, it should not be consolidated. It was intentionally separate, because users otherwise may try using features only for the runtime (`@swc/core`)

crates/swc_common/src/plugin.rs
```diff
@@ -8,8 +8,9 @@ use crate::{syntax_pos::Mark, SyntaxContext};
 use anyhow::Error;
 use std::any::type_name;
 
+/// Serialize ast to pass into plugins. This is being called from host side.
 /// TODO: https://github.com/swc-project/swc/issues/3167
-#[cfg(feature = "plugin-mode")]
+#[cfg(feature = "plugin-rt")]
 pub fn serialize_for_plugin<T>(t: &T) -> Result<rkyv::AlignedVec, Error>
```
kdy1, line 14: It think it should be `plugin-base`, as both of runtime and plugin require this.

kdy1: Previously it was used by both. (Before removing codes in `swc_common/plugin`.

crates/swc_plugin_runner/src/lib.rs
```diff
@@ -112,8 +141,15 @@ pub fn apply_js_plugin(
...
-        let _instance = load_plugin(path, cache)?;
-        // TODO: actually apply transform from plugin
+        let instance = load_plugin(path, cache)?;
+
+        let plugin_process = instance
+            .exports
+            .get_native_function::<(i32, u32), i32>("process")?;
+
+        let (alloc_ptr, len) = copy_memory_to_instance(&instance, &program)?;
+        plugin_process.call(alloc_ptr, len)?;
```
kdy1, line 151: I think we need to call `free` after this?

kdy1 (changes requested):

crates/swc_common/src/plugin.rs
kwonoj: yes, you're correct. I only realized right after changing this but went sleep. 😅

crates/swc_plugin_runner/src/lib.rs
kwonoj: yes eventually. I left it intentionally for now - first would like to verify passing / transform with serialization first and also want to check bit more if alloc is needed in this form.

crates/swc_plugin_runner/src/lib.rs
```diff
@@ -96,14 +97,42 @@ fn load_plugin(plugin_path: &Path, cache: &mut Option<PluginCache>) -> Result<In
...
+    let serialized_len = serialized.len();
+
+    let alloc_ptr = alloc.call(serialized_len.try_into()?)?;
+    let memory = instance.exports.get_memory("memory")?;
+    let view = memory.view::<u8>();
+
+    // loop over the Wasm memory view's bytes, assign bytes value of alignedvec from
+    // serialized
+    let ptr_start: usize = alloc_ptr.try_into()?;
+    for (cell, byte) in view[ptr_start..ptr_start + serialized_len + 1]
```
kdy1, line 126: I'm not sure about this.
`view` is already a section. Does it require range operation to access?

kdy1: Also, I think this should be something faster like https://doc.rust-lang.org/beta/std/ptr/fn.copy.html

kwonoj: I also think there might be some more effecient way, this is mostly copying over from some known examples. Primary focus is `make it working` for now. :sweat_smile: 

I bet most of these code will be refactored anyway while doing further polishings. But having foundational ground something works + having test ensures I can change codes later.

kdy1 (approved):
I think we can revisit the allocator issue later.
Section of memory (`MemoryView`) + global memory offset seems strange to me, but it works so maybe it's correct.

PR #3202, opened 2022-01-06
refactor(es/transforms): `async_to_generator`

### LINKED ISSUE #3182: Incorrect async function.length when function is passed as argument
### Describe the bug

Hi 👋 

Some days ago I reported an issue - #3135 - that was fixed very quickly (thank you very much @magic-akari and @kdy1).

Even thought it is fixed, I noticed that when the function is passed as an argument to another function, the value seems to still be 0.

Thank you!

### Input code

```typescript
const myWrapper = (myFunc) => {
  console.log("arity", myFunc.length);
};

myWrapper(async (one, two, three) => {
  return "irrelevant";
});
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.126&code=H4sIAAAAAAAAAz3MMQqEMBBG4T6n%2BEkVQbyAuOVeYesQBhXijIzjShDvbkCwedXjS8KbYSk%2FjetKigFhKd%2BdU4Phg9MBqR6SqcsyBh91tuJbPE%2BXiUebmt5dvXMvEuJWOCEIUws7pGZSoldUsl0ZflalTP%2FI5qtQlRs7lss7jQAAAA%3D%3D&config=H4sIAAAAAAAAA0WMSwrDMAxE76K1t93kDj2EcJXg4B8aBWqM714bUrKSmPdmOp3wtHWqrBBdH1o2%2FtJG4hPDa6hGbmoz2jlChiNjPcSWgtdksRTITR2lkMPe1pIvqaoAD%2BJ8xL855lAqn2sFnazVeVcllXyCxjN0lwPet216yfgBCKaI9boAAAA%3D

### Expected behavior

Prints "arity 3"

### Actual behavior

Prints "arity 0"

### Version

1.2.126

### Additional context

_No response_

magic-akari:
**Description:**

This pull request implements the Function Wrapper, which corresponds to [three templates in Babel](https://github.com/babel/babel/blob/ddd93b9c61a197711c2e3eae8a8b175783ea6976/packages/babel-helper-wrap-function/src/index.ts#L14-L39).

The original "FunctionWrapper" is renamed to ~~"FunctionEnvironmentUnwraper"~~ `FnEnvHoister`.

cc @Austaras 

**BREAKING CHANGE:**

- try to fix  #3182


Austaras:
I would still prefer a shorter name

kdy1:
Same here, I prefer a shorter name with documentation telling that it's the counterpart of `FunctionEnvironmentUnwraper` of babel.

magic-akari:
```javascript
function () {
    var REF = FUNCTION;
    function NAME(PARAMS) {
        return REF.apply(this, arguments);
    }
    return NAME;
})()
```

Since we use the named expression template to handle functions which have function name, we got rid of the handle iife issue.
Internal hoist function takes care potential recursive function references very well.

---

As a result, the issue with `in_prototype_assignment` has gone. ( I'm not sure, but I ran the test. )
cc @kwonoj

kwonoj:
Those prototype assignment check was introduced to avoid `this` binding issue (https://github.com/swc-project/swc/pull/2667) - as long as it won't regress I think it's fine.

crates/swc/tests/tsc-references/objectRest2_es5.1.normal.js
```diff
@@ -58,24 +58,29 @@ function _objectSpread(target) {
...
-                    case 2:
-                        objects = _ctx.sent.objects;
-                        return _ctx.abrupt("return", _objectSpread({}, connectionFromArray(objects, args)));
-                    case 4:
-                    case "end":
-                        return _ctx.stop();
-                }
-            }, _callee);
-        }))
+        resolve: (function() {
```
kdy1, line 61: This seems like a regression of IIFE optimization.

crates/swc/tests/tsc-references/objectRest2_es5.2.minified.js
```diff
@@ -29,35 +29,40 @@ function _objectSpread(target) {
...
 !function(name) {
     return {
-        resolve: (function(fn) {
-            return function() {
-                var self = this, args = arguments;
-                return new Promise(function(resolve, reject) {
-                    var gen = fn.apply(self, args);
-                    function _next(value) {
-                        asyncGeneratorStep(gen, resolve, reject, _next, _throw, "next", value);
+        resolve: (function() {
```
kdy1, line 32: Same here

kdy1 (changes requested):
I didn't review it fully.

crates/swc/tests/tsc-references/objectRest2_es5.1.normal.js
magic-akari: Hi @kdy1 

https://github.com/swc-project/swc/blob/56d1f03556f6145ef2e454fac3a4450b06034354/crates/swc_ecma_parser/tests/typescript/tsc/objectRest2/input.ts#L5-L15


We should make sure:
```diff
+ expect(rootConnection('test').resolve.length).toBe(2);
```

crates/swc_ecma_utils/src/function/function_wrapper.rs
```diff
@@ -0,0 +1,394 @@
...
+        // A function may be called internally recursively if it has a name
+        if let Some(name_ident) = self.function_ident.as_ref().cloned() {
+            self.build_named_expression_wrapper(name_ident)
+        }
+        // An internal recursive call has to refer to the external binding_name,
+        // It is safe to using expression wrapper.
+        else if self.params.len() > 0 || self.binding_ident.is_some() {
+            self.build_anonymous_expression_wrapper()
+        } else {
+            // Optimization: no name, no params
```
magic-akari, line 380: @kdy1  We can only optimize it when there is no function name and no parameters.

Austaras: maybe functions like
```js
function (a = 10, ...rest) {}
```
could also be optimized?

magic-akari:
```javascript
(function (a = 10, ...rest) {}).length === 0
```

I will create a test to check it.

magic-akari: Optional parameters are not counted in the function parameter length.
So when all are optional parameters, or rest parameters, the function parameter length is 0, and we can safely optimize it.

crates/swc/tests/tsc-references/objectRest2_es5.1.normal.js
kdy1: Oh, I see. Thank you!
Then this is a field of the minifier, not transform.

crates/swc_ecma_utils/src/function/function_wrapper.rs
magic-akari: FYI, The `self.params` here are cloned and preprocessed by `Self::get_params`.

crates/swc_ecma_utils/src/function/function_wrapper.rs
```diff
@@ -0,0 +1,394 @@
...
+            self.function
+        }
+    }
+}
+
+impl Into<(FnDecl, FnDecl)> for FunctionWrapper {
```
kdy1, line 386: This `impl` seems too opinionated compared to the public API.
What `FnDecl` it creates? And, why it returns two function declarations?
(I know because I'm looking at source code, but uses will only see `.into()`)

I think it should use a dedicated method which has a describing name.

crates/swc_ecma_utils/src/function/function_wrapper.rs
```diff
@@ -0,0 +1,394 @@
...
+    fn into(mut self) -> (FnDecl, FnDecl) {
+        let name_ident = self.function_ident.as_ref().or(self.binding_ident.as_ref());
+        assert!(name_ident.is_some());
+        let name_ident = name_ident.unwrap().clone();
```
kdy1, line 390: `name_ident.expect("describtion")` seems better here.

crates/swc_ecma_utils/src/function/function_wrapper.rs
```diff
@@ -0,0 +1,394 @@
...
+            || private_ident!("_ref"),
+            |ident| private_ident!(ident.span, format!("_{}", ident.sym)),
+        );
+
+        let FnExpr { function, .. } =
+            self.build_function_forward(ref_ident.clone(), Some(name_ident.clone()));
+
+        let assign_stmt = AssignExpr {
+            span: DUMMY_SP,
+            op: AssignOp::Assign,
```
kdy1, line 296: Please. use `op!("=")`.

crates/swc_ecma_utils/src/function/function_wrapper.rs
```diff
@@ -0,0 +1,394 @@
...
+            params: Self::get_params(fn_expr.function.params.iter()),
+            function: fn_expr.into(),
+        }
+    }
+}
+
+impl TryFrom<VarDeclarator> for FunctionWrapper {
+    type Error = ();
+
+    fn try_from(var: VarDeclarator) -> Result<Self, Self::Error> {
```
kdy1, line 92: I think this method should return `VarDeclarator` on error. Otherwise the user have to clone it.

kdy1: This pattern is quite common in rust.

crates/swc_ecma_utils/src/function/function_wrapper.rs
```diff
@@ -0,0 +1,394 @@
...
+                _ => Err(()),
+            }
+        } else {
+            Err(())
+        }
+    }
+}
+
+impl FunctionWrapper {
+    fn get_params<'a, T, P>(params_iter: T) -> Vec<Param>
```
kdy1, line 125: Can you add a comment that this method is used to preserve `length` of a function?

kdy1 (changes requested):
Thanks! I'll review test changes later

crates/swc_ecma_utils/src/function/function_wrapper.rs
```diff
@@ -0,0 +1,422 @@
...
+            binding_ident: None,
+            function_ident: None,
+            params: Self::get_params(fn_expr.function.params.iter()),
+            function: fn_expr.into(),
+            _type: Default::default(),
+        }
+    }
+}
+
+impl Into<Expr> for FunctionWrapper<Expr> {
```
kdy1, line 330: Oh, I didn't know this impl is possible.

I'm not sure why does it work. It should be one of `specialization` or `#[fundameltal]`.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -594,101 +609,12 @@ impl VisitMut for MethodFolder {
...
-            Expr::Arrow(ArrowExpr {
-                is_async: true,
-                span,
-                params,
-                body,
-                is_generator,
-                type_params,
-                return_type,
-            }) => {
+            Expr::Arrow(arrow_expr) if arrow_expr.is_async => {
```
kdy1, line 615: `Expr::Arrow(arrow_expr @ ArrowExpr { is_async: true, .. })` would be better.

crates/swc_ecma_transforms_compat/src/es2017/async_to_generator.rs
```diff
@@ -756,357 +654,23 @@ impl Actual {
...
-                    ..
-                },
-            ) => {
-                self.visit_fn(
-                    fn_expr.ident.clone(),
-                    binding_ident,
-                    &mut fn_expr.function,
-                    false,
-                );
+            Expr::Fn(fn_expr) if fn_expr.function.is_async => {
```
kdy1, line 660: Same here, `@ FnExpr{ function: Function { is_async: true, .. }, .. }` would be better.

kdy1 (changes requested):
Mostly LGTM. Thanks!
I found some code style issues, where rust has a dedicated syntax.
I think dedicated syntax (`@ Struct {}`) should be preferred over generic if guard.

crates/swc/tests/fixture/issue-1718/case2/output/index.ts
```diff
@@ -32,13 +32,18 @@ function scanUser(groups) {
...
     _scanUser = _asyncToGenerator(function*(groups) {
-        yield Promise.all(groups.map(_asyncToGenerator(function*({ users  }) {
-            for (const user of users){
-                console.log('user', user);
-                yield new Promise((resolve)=>setTimeout(resolve, 30)
-                );
-            }
-        })));
+        yield Promise.all(groups.map(function() {
+            var _ref = _asyncToGenerator(function*({ users  }) {
```
kdy1, line 36: I used some time to this about this, but I failed to find the way we can we optimize back from the minifier.

kdy1: Maybe we need escape analysis in transform, not minifier.

kdy1 (commented):
I'm bit worried about the regression of output size. (`length` of function does not matter in most cases)

How do you think about this?

crates/swc/tests/fixture/issue-1718/case2/output/index.ts
magic-akari: > I'm bit worried about the regression of output size. (`length` of function does not matter in most cases)
> 
> How do you think about this?

There's a option in Babel's assumptions: [ignoreFunctionLength](https://babeljs.io/docs/en/assumptions#ignorefunctionlength).

If we introduce the assumptions, we can optimize further.

```rust
!assumptions.ignore_function_length && self.params.len() > 0
```

kdy1: Oh, seems like a clever solution. I'll quickly add a type for it. Thanks!

kdy1: Did it, can you rebase this PR?

magic-akari: Should I add the assumptions support for `async_to_generator` right now?

kdy1: Hmm, no I'll merge it and let's gradually implement loose mode.

kdy1 (approved):

PR #3211, opened 2022-01-07
Fix #3081 and lazy imports

### LINKED ISSUE #3081: Issues with lazy requires codegen when using nested closures
### Describe the bug

SWC generates faulty code when using `module.lazy = true` and `module.type = "commonjs"` along with nested closures.

It will generate a mix of lazy and non-lazy code:

```js
import { spawn } from "child_process";

function log() {
  console.log(spawn);
}

const other = () => {
  const nestedClosure = () => {
    spawn("ls");
  };
};
```

Generates 

```js
"use strict";
var _childProcess = require("child_process");
function log() {
    console.log(_childProcess().spawn);
}
const other = ()=>{
    const nestedClosure = ()=>{
        (0, _childProcess).spawn("ls");
    };
};
```

Notice that the `require` statement and the call in `nestedClosure` are non-lazy, but the call in `function log()` is `lazy`.

### Inverting the functions

Inverting the order of the `log` and `other` functions will not produce the same issue. Instead, `swc` will generate completely non-lazy, but functional code.

```js
import { spawn } from "child_process";

const other = () => {
  const nestedClosure = () => {
    spawn("ls");
  };
};

function log() {
  console.log(spawn);
}
```

Generates

```js
"use strict";
var _childProcess = require("child_process");
const other = ()=>{
    const nestedClosure = ()=>{
        (0, _childProcess).spawn("ls");
    };
};
function log() {
    console.log(_childProcess.spawn);
}
```

### Avoid using nested closures

Replacing either `other` or `nestedClosure` with a named function will also produce valid, lazy code:

```js
import { spawn } from "child_process";

function log() {
  console.log(spawn);
}

function other() {
  const nestedClosure = () => {
    spawn("ls");
  };
};
```

Generates

```js
"use strict";
function _childProcess() {
    const data = require("child_process");
    _childProcess = function() {
        return data;
    };
    return data;
}
function log() {
    console.log(_childProcess().spawn);
}
function other() {
    const nestedClosure = ()=>{
        _childProcess().spawn("ls");
    };
}
```

### Input code

```typescript
import { spawn } from "child_process";

function log() {
  console.log(spawn);
}

const other = () => {
  const nestedClosure = () => {
    spawn("ls");
  };
};
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript"
    },
    "target": "es2020"
  },
  "module": {
    "type": "commonjs",
    "lazy": true
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.115&code=H4sIAAAAAAAAA1WOMQrDMBAE%2B3vFospu8gHhNHlICPI5Fsg6oTuTwujvkRMIpN2dHTZuRarhgJbHK6NhqbLBhTWm%2BV6qBFZ1nmjZc7AoGUmew4iDgCBZJfHlTD7r0VMjOmOD2MoVEzo7XX%2B4IbMaz7ckulf%2B6%2FG9MLikrpuA1nWe3kTS5YKhAAAA&config=H4sIAAAAAAAAA0WNSQ6DMAxF7%2BI1iypL7tBDWKlBQWSQbaSmEXevg4LY2X94v8EmHuYGBVmI%2ByU1KX5hBq2FxHMoChOomLTgLnTag7ySWoTEvZwze89ZaAQmiCGFpXaYz7EwiTwWpnW%2Fk6exYv4cXWjXnjGtEnPapFPxZxTlo4%2Fe0AEK8h7Ny%2F8DWryisckAAAA%3D

### Expected behavior

I expect the output to produce a lazy `require` statement, along with lazy `_childProcess()` calls.

```js
"use strict";
function _childProcess() {
    const data = require("child_process");
    _childProcess = function() {
        return data;
    };
    return data;
}
function log() {
    console.log(_childProcess().spawn);
}
const other = ()=>{
    const nestedClosure = () => {
        _childProcess().spawn("ls");
    }
};
```

### Version

@swc/cli: 0.1.55, @swc/core: 1.2.121

### Additional context

_No response_

gmalette:
## Description

Fixes #3081 

Lazy imports generate broken code if an usage of the import precedes the blacklisting of the identifier. 

The ultimate cause was that the blacklisting of identifiers was done as the generated statements get emitted, causing statements prior to be emitted as lazy, even if an identifier was subsequently blacklisted.

A proximal cause of the first identified issue (dealing with nested closures) was an unidentified issue in tracking the top-level, and nested-closures believing they were top-level.

## Background

An imported identifier are (currently) blacklisted for two reasons:
1. It is being re-exported
2. It is used at the top level

These two reasons can be generalized as only being used at the top-level, since that's also what exports are.

## Implementation

To solve the problem, we must perform two additional prepasses on the AST:
- First, populate a `Scope` with imports, to be able to map identifiers to
  their sources
- Second, navigate all top-level identifiers and mark them as non-lazy

The second pass is performed using a new `Visitor` which only visits top-level nodes.

**Related issue:** #3081


crates/swc_ecma_transforms_module/src/common_js.rs
```diff
@@ -55,6 +58,57 @@ where
...
+impl LazyIdentifierVisitor {
+    fn new(scope: Rc<RefCell<Scope>>) -> Self {
+        LazyIdentifierVisitor {
+            scope,
+            top_level_idents: Default::default(),
+        }
+    }
+}
+
+impl Visit for LazyIdentifierVisitor {
```
kdy1, line 74: I think this Visitor should have `visit_labelled_stmt` which ignores the identifier in label.

kdy1: Additionally, a break statement or a continue statement can have label. Those methods can be empty, just like visitor methods of import/exports.

crates/swc_ecma_transforms_module/src/common_js.rs
```diff
@@ -55,6 +58,57 @@ where
...
+    fn visit_function(&mut self, _: &Function) {}
+    fn visit_constructor(&mut self, _: &Constructor) {}
+    fn visit_setter_prop(&mut self, _: &SetterProp) {}
+    fn visit_getter_prop(&mut self, _: &GetterProp) {}
+    fn visit_class_prop(&mut self, _: &ClassProp) {}
+
+    fn visit_decl(&mut self, decl: &Decl) {
+        match decl {
+            Decl::Class(ref c) => {
+                if let Some(super_class) = &c.class.super_class {
```
kdy1, line 96: I think this `if let` is redundant.

`c.class.super_class.visit_with(self)` will work.

crates/swc_ecma_transforms_module/src/common_js.rs
```diff
@@ -55,6 +58,57 @@ where
...
+    fn visit_export_decl(&mut self, _: &ExportDecl) {}
+    fn visit_named_export(&mut self, _: &NamedExport) {}
+    fn visit_export_default_decl(&mut self, _: &ExportDefaultDecl) {}
+    fn visit_export_default_expr(&mut self, _: &ExportDefaultExpr) {}
+
+    fn visit_export_all(&mut self, export: &ExportAll) {
+        self.top_level_idents.insert(export.src.value.clone());
+    }
+
+    fn visit_function(&mut self, _: &Function) {}
```
kdy1, line 87: I think `visit_arrow_expr` is missing?

crates/swc_ecma_transforms_module/src/common_js.rs
```diff
@@ -55,6 +58,57 @@ where
...
+    fn visit_constructor(&mut self, _: &Constructor) {}
+    fn visit_setter_prop(&mut self, _: &SetterProp) {}
+    fn visit_getter_prop(&mut self, _: &GetterProp) {}
+    fn visit_class_prop(&mut self, _: &ClassProp) {}
+
```
kdy1, line 92: I think having a visitor method for `PropName` is preferrable.

kdy1: It visits children iff it's `PropName::Computed`.

crates/swc_ecma_transforms_module/src/common_js.rs
```diff
@@ -125,6 +179,24 @@ where
...
+            match item {
+                ModuleItem::ModuleDecl(ModuleDecl::Import(import)) => {
+                    self.scope.borrow_mut().insert_import(import.clone())
+                }
+                _ => {}
+            }
+        }
+
+        let mut visitor = LazyIdentifierVisitor::new(self.scope.clone());
+        for item in &items {
```
kdy1, line 193: `items.visit_with(&mut visitor)` would be better.

crates/swc_ecma_transforms_module/src/common_js.rs
```diff
@@ -125,6 +179,24 @@ where
...
+            item.visit_with(&mut visitor);
+        }
+        for ident in &visitor.top_level_idents {
+            self.scope.borrow_mut().lazy_blacklist.insert(ident.clone());
```
kdy1, line 197: This `clone()` seems redundant.

kdy1 (changes requested):
Thank you!

crates/swc_ecma_transforms_module/src/common_js.rs
gmalette: I don't understand what you mean here, can you elaborate please?

crates/swc_ecma_transforms_module/src/common_js.rs
gmalette: Entirely possible! To be honest I don't really know JavaScript all that well (or, as you may have guessed, Rust for that matter). I tried to emulate the `mark_as_nested!` macro.

Can you give me an example of what would make an `ArrowExpr`? I'd like to add a test for it

crates/swc_ecma_transforms_module/src/common_js.rs
gmalette: I also cannot make a test fail for this case, even without `visit_labeled_stmt` and `visit_continue_stmt`, here's what I tried:

```javascript
import { x } from "lib";

x:
console.log(1);

continue x;

function fn() {
    x()
}
```

This produces correctly lazy code, do you have an idea of what I'm doing wrong?

crates/swc_ecma_transforms_module/src/common_js.rs
gmalette: I think you're saying that we need to visit the child of an expression if it's like `obj[prop]`, yes?

kdy1: I meant it would be good to have a method `visit_prop_name`, while it does

```rust

match n {
    PropName::Computed(n) => {
        n.visit_with(self);
    }
    _ => {}
}

```

This is to prevent accidental breakage due to AST changes.

gmalette:
Thanks @kdy1 for your feedback, I really appreciate it!

I think I included all your suggestions, but I couldn't produce failing tests for `visit_arrow_expr`, `visit_labeled_stmt`, and `visit_continue_stmt`.

For `visit_labeled_stmt` and `visit_continue_stmt` it's expected; their context differs from other top-level identifiers.

I don't worry all that much about it because now the failure mode should produce correct but unoptimized code (lazy but immediately required), whereas before it produced incorrect code.

crates/swc_ecma_transforms_module/src/common_js.rs
kdy1: Yeah, it's expected as currently swc does not attach context to labels, which is used by label/continue/break statements.
I was talking about being safe from future changes.

crates/swc_ecma_transforms_module/src/common_js.rs
kdy1: I didn't think a lot about it. I found several visitor methods to prevent visiting nested nodes, but I failed to find arrow expression from the method list.

kdy1:
Yeah, it's fine, thank you!
I'll review this soon.

kdy1 (approved):
LGTM. Thanks!

magic-akari:
I'm implementing lazy require for the new module transformation #4758 .

I checked at Babel's transformations.

An imported identifier cannot be lazy for two reasons:
- raw import
- reexport

```
import "foo"; // raw import without any local binding
export * from "bar"
```

It has nothing to do with top level usage.
I will follow Babel's implementation.

PR #3216, opened 2022-01-09
feat(swc/plugin): allow multivalue for plugin fn signature

kwonoj:
**Description:**

`tests/rust-plugins/swc_internal_plugin/README.md` includes crux of changes in this PR.

Our plugin's transform doesn't guarantee deterministic size of returned value, so `process` should able to pass `ptr` + `size` to reconstruct `AlignedVec` to deserialize its result. Thanksfully there is accepted spec (https://github.com/WebAssembly/multi-value) allows to return multiple values from wasm exported fn. 

Problem is generating those bytes - while compiler supports `target-feature=+multivalue` it is not working as expected at least for the target `wasm32-unknown-unknown`. We can bring whole `wasm-bindgen` and its cli to utilize wasm-bindgen's bindings, but it's yet unclear we'll require `wasm-bindgen` for all plugin authoring.

This PR instead picks up necessary polyfill only for the multivalue against the generated wasm binary. This can be safely removed once we know compiler flag issues gets fixed. If we release plugin features before compiler fix, we may integrate this steps into our cli as well as scaffolding plugin template to avoid manual postbuild steps. For now, test-plugin relies on naive scripts to chain commands.

### DIFF
--- a/crates/swc_plugin_runner/src/lib.rs
+++ b/crates/swc_plugin_runner/src/lib.rs
@@ -145,10 +145,10 @@ pub fn apply_js_plugin(
 
         let plugin_process = instance
             .exports
-            .get_native_function::<(i32, u32), i32>("process")?;
+            .get_native_function::<(i32, u32), (i32, i32)>("process")?;
 
         let (alloc_ptr, len) = copy_memory_to_instance(&instance, &program)?;
-        plugin_process.call(alloc_ptr, len)?;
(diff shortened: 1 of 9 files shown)


kdy1 (approved):
LGTM, nice work! 


This is the point I was blocked when I was investigating the wasm-based plugin before the `abi_stable`-based one, and TIL it's a bug of a compiler.

PR #3219, opened 2022-01-09
fix(es/hygiene): block scope

### LINKED ISSUE #2627: Wrong Conversion of Let and Const Variables Defined in Block
### Describe the bug

Const and let only exist in the blocks they are defined in, but after swc conversion, they all become var variables and can print value out of the block.

### Input code

```typescript
{
  let a = 1;
  const b = 1;
  var c = 1;
}
console.log(a); 
console.log(b); 
console.log(c);
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false,
      "dynamicImport": false,
      "privateMethod": false,
      "functionBind": false,
      "exportDefaultFrom": false,
      "exportNamespaceFrom": false,
      "decorators": false,
      "topLevelAwait": false,
      "importMeta": false
    },
    "transform": null,
    "target": "es5",
    "loose": false,
    "externalHelpers": false
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.106&code=H4sIAAAAAAAAA6vm5VJQyEktUUhUsFUwtAbxkvPziksUkuD8ssQihWQor5aXCySdn5Oql5OfrpGoaa2AKpKEIZIMFAEAhYlu3mYAAAA%3D&config=H4sIAAAAAAAAA0WMTQrEIAxG75K1286id5hDBCctFv9IUhgR714tlu7C915ehUMsrBUyshCPS0pU%2FMMKZAOKZZcVTNf6tKEXagYUeScdiiyd%2BZSEJjUQXHRbGSWbQmYSeRHG3T9m66GQfucYKmjJdAc%2F0N7G%2FHPynaLySe0Cj9ke9LUAAAA%3D

### Expected behavior

const and let only exist in the blocks they are defined in. After transforming, they should like this
```javascript
// const and let only exist in the blocks they are defined in.
{
  var _a = 1;
  var _b = 1;
  var c = 1;
}
console.log(a); // ReferenceError
console.log(b); // ReferenceError
console.log(c); // Prints 1
```
or like this
// const and let only exist in the blocks they are defined in.
```javascript
{
  var a$__0 = 1;
  var b$__1 = 1;
  var c = 1;
}
console.log(a); // ReferenceError
console.log(b); // ReferenceError
console.log(c); // Prints 1
```

### Version

@swc/cli: 0.1.51 @swc/core: 1.2.106

### Additional context

I run swc use cmd as follows
```bash
```

### LINKED ISSUE #3213: Variable being renamed unnecessarily is breaking code
### Describe the bug

Running the function `f()` from the input code, causes the console to log `"original string"`, whereas running the one outputted by the compiler results in `"new string"` being logged.

### Input code

```typescript
function f() {
  var value$2 = 'original string';
  if (true) {
    var value$21 = 'new string'
    console.log(value$2)
  }
  [].map(function x(_) {
    var value$2; // this causes the previous value$2 to become value$21
    return value$2
  })
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": true,
    "minify": {
      "compress": false,
      "mangle": false
    },
    "keepClassNames": true
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.124&code=H4sIAAAAAAAAA21OzQqCQBC%2B%2BxRzCFwvSl2lJ4mIbZm1gXVG9keD8N1bLYWgwwfDfL82sYkkDFZV8CoARu0zXMLDCc5QiqeOWDsI0RN3ZZslZEFFn%2FBj%2BLEcFw%2FjtMlX3ggHcVg76dRXV2Vizrhc614Pym4rnur2J7WFpoH4oABGp4Ah3wiDx5EkhX1tFLijkR73MWuOx5g8b7%2Bltyrm4g21Prmx%2BAAAAA%3D%3D&config=H4sIAAAAAAAAA0WNOw7DIBBE77I1bVLQpk7usCJrC4efGCwFId894CC5W828edtogyHdKHGG5HGhhsJf0iTGM0y2qZDqWI8WdpBDUeG8ShkIbr1zMUJIl7yLIm%2BDXeoQmehTFmDuesVhdXJpPiLp4Rh4sRf8BT328b0PrFGpSc4vdzou87RZPCd47n5%2FjK7bygAAAA%3D%3D

### Expected behavior

I think that there are two problems:

1. The compiler shouldn't have renamed the first `value$2` to `value$21` in the first place. It does that because of the second declaration inside `function x(_)`, but it's not necessary since variables are scoped to its closures, so the shadowing is valid.
2. The compiler should notice that the `var value$21` declaration inside the `if` block clashes with the compiler's renaming of `value$2` to `value$21`, so it should probably rename it to something else (or not at all, as mentioned in the previous point.). Notice that the bug is gone if the code is taken out off the `if` block.

### Actual behavior

_No response_

### Version

1.2.124

### Additional context

This might seem like an unrealistic example, but it actually broke our application written in [Rescript](https://rescript-lang.org/), whose compiler outputs a lot of variables suffixed with numbers.

Update: Looks like this only started happening on `v1.2.102`

### LINKED ISSUE #3235: Block-scoping bug?
I have the following script:

let res = [];

let a = 2;
res.push(a === 2);

{
    let b = 1;
}
res.push(typeof b === "undefined");

if(true) {
	let b = 0;
}
res.push(typeof b === "undefined");

for(let b = 0; b < 10; b++) {
}
res.push(typeof b === "undefined");

function test() {
	let b = 7;
}
res.push(typeof b === "undefined");

console.log(res);

I expect this to evaluate to [true, true, true, true, true], but am getting [true, false, false, false, false] because the first block-scoped variable b is not renamed or prepended with an underscore. Here's the [link](https://play.swc.rs/?version=1.2.124&code=H4sIAAAAAAAAA8tJLVEoSi1WsFWIjrXm4soBchOBHCNrLqCoXkFpcYYGkG8LFNEESldzKQABSFESUJGhNVctQllJZUFqfhpIAqhaqTQvJTUtMy81RQmkLzNNo6SoNFVToZqLE6bbgGjdaflFGghdQMpGwRBEa2uDDCTWkNK85JLM%2FDyFktTiEg0Ul5gT7RKgImsAjBOGyzIBAAA%3D&config=H4sIAAAAAAAAA0WMSwrDMAxE76K1t93kDj2EcJXg4B8aBWqM714bUrKSmPdmOp3wtHWqrBBdH1o2%2FtJG4hPDa6hGbmoz2jlChiNjPcSWgtdksRTITR2lkMPe1pIvqaoAD%2BJ8xL855lAqn2sFnazVeVcllXyCxjN0lwPet216yfgBCKaI9boAAAA%3D) to the swc-playground transpilation. In babel the transpilation looks like the following with all block-scoped let assignments prepended with underscores and incremented:

var res = [];
var a = 2;
res.push(a === 2);
{
  var _b = 1;
}
res.push(typeof b === "undefined");

if (true) {
  var _b2 = 0;
}

res.push(typeof b === "undefined");

for (var _b3 = 0; _b3 < 10; _b3++) {}

res.push(typeof b === "undefined");

function test() {
  var b = 7;
}

res.push(typeof b === "undefined");
res;

magic-akari:
**Description:**
This pull request try to fix block scope issues.

Previous behavior:
Rename every ident if conflict.

Current behavior:
Only the idnet marked as `block_scoped` will be renamed.

`block_scoped` ident is an identifier which is `var`, but should behave like `let`/`const`.
There is a similar [definition at Babel](https://github.com/babel/babel/blob/35ca17aab6c9daa631a2e80e09afc09da8ce11ec/packages/babel-types/src/constants/index.ts#L66).

- When `let`/`const` transforms to `var`, we mark it as `block_scoped` ident.
- We assume all `private_ident` is `block_scoped` ident.

TODO:
- [ ] ~~There are too many cases where `privae_ident` should be used but is not, which will result in name conflicts.~~
- [ ] ~~`Ident::new` or direct construction should be disabled. It is better only use `private_ident` and `quote_ident` to construct `Ident`.~~

**BREAKING CHANGE:**

- fix #2627 
- fix #3213 
- fix #3235


magic-akari:
Hi @kdy1 ,
There are some issues with `swcHelpers` need to be resolved.

- With config `externalHelpers: true`, we should mark `swcHelpers` as `block_scoped` by using `private_ident`.
And pass ident into transformed code as ref ident to ensure they are connected when renaming.

- With config `externalHelpers: false`, we should mark all exposed ident as `private_ident` and ref them during transformation. By doing so, we will solve the [issue](https://github.com/swc-project/swc/issues/3213#issuecomment-1012819505) mentioned by @Austaras .


The pull request is not completed. But it is better to review it now.

crates/swc/tests/fixture/issue-2372/output/index.js
```diff
@@ -14,7 +14,7 @@ describe('example test that should fail due to compilation', ()=>{
             2,
             4
         ], 
-    ])('for entry %s', (_1, a, b, expected)=>{
+    ])('for entry %s', (_, a, b, expected)=>{
         const result = (0, _).sum(a, b);
```
magic-akari, line 18: We should rename the import ident by using `private_ident`.

crates/swc/tests/fixture/issue-2854/1/output/index.js
```diff
@@ -4,7 +4,7 @@ export function App() {
 export function Form(param) {
     var _onChange = param.onChange, onChange = _onChange === void 0 ? function() {} : _onChange;
     return React.createElement("input", {
-        onChange: function onChange1() {
+        onChange: function onChange() {
```
magic-akari, line 7: should use `private_ident`.

kdy1:
> With config externalHelpers: true, we should mark swcHelpers as block_scoped by using private_ident.
> And pass ident into transformed code as ref ident to ensure they are connected when renaming.

Why should we make `swcHelpers` block scoped? Single `Mark` for `swcHelpers` is enough I think?


> With config externalHelpers: false, we should mark all exposed ident as private_ident and ref them during transformation. > By doing so, we will solve the issue mentioned by @Austaras .

Why? We can use a helper mark, shared by helpers.

magic-akari:
> > With config externalHelpers: true, we should mark swcHelpers as block_scoped by using private_ident.
> > And pass ident into transformed code as ref ident to ensure they are connected when renaming.
> 
> Why should we make `swcHelpers` block scoped? Single `Mark` for `swcHelpers` is enough I think?
> 
> > With config externalHelpers: false, we should mark all exposed ident as private_ident and ref them during transformation. > By doing so, we will solve the issue mentioned by @Austaras .
> 
> Why? We can use a helper mark, shared by helpers.

If there are some codes as following:

```javascript
{
    var foo = 1;
}

console.log(foo);
```

Should we rename it? It depends.
If the code is user-entered, keep the name. The `foo` should be hoisted from the block.

But if they are transformed form following:

[REPL](https://play.swc.rs/?version=1.2.124&code=H4sIAAAAAAAAA6vmUgCCnNQShbT8fAVbBUNrrlouruT8vOL8nFS9nPx0DaC4pjUARembNCcAAAA%3D&config=H4sIAAAAAAAAA0WOSw6DMAwFr1J5zaKtShecoJsewkoNTZWfbFMRodydBCGxs57njb3CTwwMKyRkIW6T5KC4wABkPIphmxS6itVoRCdUOlDkibQhcr%2Fe%2Brp2MQodQAfeBjvmJjPRJyaRc4VhcnSqaFHigO5FLhFXTnluuY%2BfuXEraE60n3pCOdWHzsr7AFuv6sJ%2F7%2BwPVhuYL0dPl0cPpWyRCl%2B%2F7QAAAA%3D%3D)

```javascript
{
    let foo = 1;
}

console.log(foo);
```

Should we rename it? No, the `foo` is block scoped and the `foo` in `console.log` refers to unresolved binding which may be `window.foo`.

How do we recognize them?
`Mark` is not enough. We need more information.
When a block-scoped `Ident` (`let` or `const`) is transformed into `var`, mark it as `block_scoped`.
[Babel did it in a similar way.](https://github.com/babel/babel/blob/35ca17aab6c9daa631a2e80e09afc09da8ce11ec/packages/babel-plugin-transform-block-scoping/src/index.ts#L157-L158)

We treat all `Ident` constructed during code transformed as `block_scoped` as well.
In the process of code transformation, We use `private_ident` to avoid naming conflicts.
In most cases, we do not want `private_ident` to be hoisted. It makes sense to mark `private_ident` as `block_scoped`.

In result we should never rename `Ident` that untouched by code transformation.
We only resolve the case of naming conflicts caused by our code transform.

It there is naming conflicts in user input, it's better to report it from lint and leave it untouched.

kdy1:
I mean, the helper mark is almost the same as `private_ident`, but just shared by various helper functions.

All helpers are injected on top-level, and we can detect conflict from `hygiene`, without any more information. If user references an unresolved identifier, it's decl-ref conflict as helpers use their own `SyntaxContext`.

kdy1:
Private identifier is simply a symbol + unique `SyntaxContext`, while helper identifier is a symbol + unique `SyntaxContext` (per file)

crates/swc_ecma_transforms_base/src/hygiene/usage_analyzer.rs
```diff
@@ -313,36 +344,73 @@ impl UsageAnalyzer<'_> {
...
-                    ctxts_of_decls.len() > 1
-                } else {
+                let conflicts: Vec<DeclInfo> = ctxts_of_decls
+                    .iter()
+                    .filter(
+                        |DeclInfo {
+                             ctxt,
+                             block_scoped: bs,
+                             ..
+                         }| ctxt == &id.1 && (block_scoped || *bs),
```
magic-akari, line 359: @kdy1 
Do you mean that if we detect some specific SyntaxContext, we should treat it as `private_ident` as well?

kdy1: Yes, actually `SyntaxContext` is the trick behind `private_ident!`. 

`private_ident!` works by creating a unique `SyntaxContext` (by `SyntaxContext::empty(Mark::fresh(Mark::root()))`), and creating `Ident` witth provided symbol and the `SyntaxContext` we made.

Actual renaming is done by the `hygiene` pass.

crates/swc/tests/fixture/issue-3235/7/output/index.js
```diff
@@ -0,0 +1,18 @@
...
+            while(1)switch(_ctx.prev = _ctx.next){
+                case 0:
+                case "end":
+                    return _ctx.stop();
+            }
+        }, _callee);
+    }));
+    return _foo.apply(this, arguments);
+}
+var swcHelpers = 1;
```
magic-akari, line 18: Hi, @kdy1 ,
I add a test case.
It seems to be a regression in my changes.

magic-akari (commented):
Hi, @kdy1 ,
In the original code, The first Ident will be renamed.
https://github.com/swc-project/swc/blob/848e9c8b8a14da971358a2afe33c284d1bbcb014/crates/swc_ecma_transforms_base/src/hygiene/usage_analyzer.rs#L321

For imported `swcHelpers`, the first `Ident` is what we import and should be renamed.

But it cannot handle these two cases at the same time
[REPL](https://play.swc.rs/?version=1.2.124&code=H4sIAAAAAAAAAytLLFJIVLBVMLTm4qrmUgCCnNQSmEgtAMhK4o8eAAAA&config=H4sIAAAAAAAAA0WOSw6DMAwFr1J5zaKtShecoJsewkoNTZWfbFMRodydBCGxs57njb3CTwwMKyRkIW6T5KC4wABkPIphmxS6itVoRCdUOlDkibQhcr%2Fe%2Brp2MQodQAfeBjvmJjPRJyaRc4VhcnSqaFHigO5FLhFXTnluuY%2BfuXEraE60n3pCOdWHzsr7AFuv6sJ%2F7%2BwPVhuYL0dPl0cPpWyRCl%2B%2F7QAAAA%3D%3D)

```javascript
var a = 1;

{
    let a = 1;
}
```

[REPL](https://play.swc.rs/?version=1.2.124&code=H4sIAAAAAAAAA6vmUgCCnNQShUQFWwVDa65aLq6yxCIoDwAly%2FyBHgAAAA%3D%3D&config=H4sIAAAAAAAAA0WOSw6DMAwFr1J5zaKtShecoJsewkoNTZWfbFMRodydBCGxs57njb3CTwwMKyRkIW6T5KC4wABkPIphmxS6itVoRCdUOlDkibQhcr%2Fe%2Brp2MQodQAfeBjvmJjPRJyaRc4VhcnSqaFHigO5FLhFXTnluuY%2BfuXEraE60n3pCOdWHzsr7AFuv6sJ%2F7%2BwPVhuYL0dPl0cPpWyRCl%2B%2F7QAAAA%3D%3D)

```javascript
{
    let a = 1;
}

var a = 1;
```

Notice: only the `let a` should be renamed.

I handle these two cases by detect the `block_scoped`.

It means I cannot determine which one should be renamed if two `Ident` are in the same context without extra information.

crates/swc/tests/fixture/issue-3235/7/output/index.js
kdy1: If two identifier has same symbol and same syntax context, those are equal to each other.
If different variable get a same syntax context, then it's bug of `resolver` pass.

magic-akari: in this [REPL](https://play.swc.rs/?version=1.2.124&code=H4sIAAAAAAAAA6vmUgCCnNQShUQFWwVDa65aLq6yxCIoDwAly%2FyBHgAAAA%3D%3D&config=H4sIAAAAAAAAA0WOSw6DMAwFr1J5zaKtShecoJsewkoNTZWfbFMRodydBCGxs57njb3CTwwMKyRkIW6T5KC4wABkPIphmxS6itVoRCdUOlDkibQhcr%2Fe%2Brp2MQodQAfeBjvmJjPRJyaRc4VhcnSqaFHigO5FLhFXTnluuY%2BfuXEraE60n3pCOdWHzsr7AFuv6sJ%2F7%2BwPVhuYL0dPl0cPpWyRCl%2B%2F7QAAAA%3D%3D), we get the wrong result because we can't distinguish which one needs to be renamed.

kdy1: Hmm... Then we may remove the logic for retaining identifiers of top-level items?

magic-akari: it is not related to top level. We should not touch the `var` Ident.
See this [swc playground](https://play.swc.rs/?version=1.2.133&code=H4sIAAAAAAAAA6vmUgCCajAJAjmpJQqJCrYKhtYK%2BvoKxRn5pTkpCkWpeYm5qWA1tVxgqiyxCFNZdmpqgQJYJVBVWmlecklmfp5CWn6%2BhibUBqrbAwACzhNBvwAAAA%3D%3D&config=H4sIAAAAAAAAA0WOSw6DMAwFr1J5zaKtShecoJsewkoNTZWfbFMRodydBCGxs57njb3CTwwMKyRkIW6T5KC4wABkPIphmxS6itVoRCdUOlDkibQhcr%2Fe%2Brp2MQodQAfeBjvmJjPRJyaRc4VhcnSqaFHigO5FLhFXTnluuY%2BfuXEraE60n3pCOdWHzsr7AFuv6sJ%2F7%2BwPVhuYL0dPl0cPpWyRCl%2B%2F7QAAAA%3D%3D) and this [Babel REPL](https://babeljs.io/repl#?browsers=chrome%2045&build=&builtIns=false&corejs=false&spec=false&loose=false&code_lz=N4KABBapNgNgUwC5gIZgLxgIwG4wD0BYAzgBYD2ArnACZgBOCAdqgLYLiQC-IXEAN1QM0mHPiKlKNegGsECAA5hWHELxAAzKswDGSAJYVmYTRQoAKAJRR-t2JEQp0WPIWLlqdRi3acYGjBCIi7i7lJecgrKqpzcQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=true&sourceType=module&lineWrap=true&presets=env&prettier=true&targets=&version=7.16.9&externalPlugins=&assumptions=%7B%7D).

kdy1: Oh... You are right, although I'm not sure how can we fix this.

Scoping in rust is super-hard because of `no two &mut to same data`.

magic-akari: You could check the first commit in this pull request that I added some test cases about it.

kdy1:
Honestly I don't think adding a field to `Ident` is a good idea.

magic-akari:
Without adding a field to the `Ident`, is there any way to mark a `Ident` as `block_scoped` in code  transformation and consume it in hygiene pass?

FYI, Babel store them in NodePath.

magic-akari:
> If we can collect block-scoped `Id`s from somewhere, it would be possible. If we only make block-scoped identifiers some of passes, we can store it in some variable (e.g. shared by resolver, block_scoping, hygiene)

https://github.com/swc-project/swc/blob/af53b9487f74ff28438928903fb1f2db93fe4fa8/crates/swc/src/builder.rs#L267-L299

Maybe I can pass a hashset to collect `Ident` from here.

crates/swc_ecma_transforms_base/src/hygiene/tests.rs
```diff
@@ -136,6 +136,7 @@ fn simple() {
 }
 
 #[test]
+#[ignore]
```
kdy1: Why is this ignored?

kdy1 (commented):
Thank you so much for the hard work!

crates/swc_ecma_transforms_base/src/hygiene/tests.rs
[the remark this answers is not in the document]
magic-akari: Oh, I have no idea about how to handle multiple files.
What do you think about it?

crates/swc_ecma_transforms_compat/tests/es2015_block_scoping.rs
```diff
@@ -11,7 +11,7 @@ use swc_ecma_transforms_testing::{compare_stdout, test, test_exec, Tester};
 
 test!(
     ::swc_ecma_parser::Syntax::default(),
-    |_| block_scoping(),
+    |_| block_scoping(Default::default()),
```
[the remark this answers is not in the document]
magic-akari: You are right.
I just updated the test config to use `resolver` and `hygiene`.

crates/swc_ecma_transforms_compat/src/es2015/block_scoping.rs
```diff
@@ -535,7 +540,18 @@ impl VisitMut for BlockScoping {
 
         self.var_decl_kind = old;
 
-        var.kind = VarDeclKind::Var;
+        if var.kind != VarDeclKind::Var {
+            var.kind = VarDeclKind::Var;
+
+            var.decls.iter().for_each(|decl| {
```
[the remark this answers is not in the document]
magic-akari: We only care about the ident which becomes `var` from `const`/`let` here.

crates/swc_ecma_transforms_base/src/ident_scope.rs
```diff
@@ -0,0 +1,76 @@
...
+    ident_scope_record: IdentScopeRecord,
+}
+
+impl VisitMut for UnblockIdentCollector {
+    noop_visit_mut_type!();
+
+    fn visit_mut_var_decl(&mut self, var: &mut VarDecl) {
+        var.visit_mut_children_with(self);
+
+        for decl in var.decls.iter() {
```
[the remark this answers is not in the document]
magic-akari: We record the var declaration here.
It means that user is declaring  a `var` ident here which we should not touch it while renaming. 

oh, thanks.
I forget to handle other Pat.

crates/swc_ecma_transforms_base/src/hygiene/tests.rs
```diff
@@ -313,8 +317,8 @@ fn for_loop() {
         },
         "
             for(var a=1;;) {}
-            for(var a1 of foo) {}
-            for(var a2 = 3;;) {}
+            for(var a of foo) {}
```
magic-akari, line 320: One of the main idea of this PR is to leave the user input `vars` untouched.
But when updating this test file, I found many cases where the `var` was intentionally changed.
Maybe I'm thinking in the wrong way.

kdy1: Sometimes it's inevitable, as `var` and `let`/`const` have different scoping.

crates/swc_ecma_transforms_base/src/hygiene/tests.rs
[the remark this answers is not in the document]
kdy1: Hmm... I also don't have one at the moment.

crates/swc_ecma_transforms_base/src/ident_scope.rs
```diff
@@ -0,0 +1,110 @@
...
+    fn visit_ident(&mut self, ident: &Ident) {
+        self.ident_scope_record
+            .borrow_mut()
+            .entry(IdentScope {
+                sym: ident.sym.clone(),
+                ctxt: ident.span.ctxt(),
+            })
+            .or_insert_with(|| self.ident_scope_kind);
+    }
+
```
kdy1: We need `visit_object_pat_prop` because it creates a binding.

kdy1: But seems like this does the exact thing `collect_bindings` do.

crates/swc_ecma_transforms_compat/src/es2015/block_scoping.rs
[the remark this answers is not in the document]
kdy1: Do you mean

```js
var { foo }
{
    let { foo }
}
```

this is fine?

kdy1 (changes requested):

crates/swc_ecma_transforms_compat/src/es2015/block_scoping.rs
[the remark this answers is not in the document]
magic-akari: There seems to be only one place where `const`/`let` becomes `var` since the `destructuring` is applied before `block_scoping`.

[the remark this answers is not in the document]
kdy1: Ah, I see. Thank you!

crates/swc_ecma_transforms_base/src/hygiene/tests.rs
magic-akari: If there are some cases where `var` ident of different scopes are together and conflict with each other, this pull request is meaningless and I will close it.

crates/swc_ecma_transforms_base/src/ident_scope.rs
magic-akari: https://github.com/swc-project/swc/blob/2ea81db19fb5a36ecd949c06fa242525d62c8997/crates/swc_ecma_ast/src/pat.rs#L116-L125

According to the definition of `ObjectPatProp`, I think it is only necessary to handle with `KeyValuePatProp` and `AssignPatProp`.

See:

https://github.com/swc-project/swc/blob/3afcaba887539e2cce40f5cdd4fbce0511c77ff0/crates/swc_ecma_transforms_base/src/ident_scope.rs#L101-L109

kdy1: Oh, I missed it. My bad, sorry.

magic-akari: Yes, we collect Bindings.

```javascript
let a = 1,
  [b, ...c] = [2, 3],
  { d: e } = {},
  { f = g } = {},
  { h: i = j } = {},
  { ...k } = {};
```

In this case, we got `a`, `b`, `c`, `e`, `f`, `i` and `k`.

kdy1: Is there a reason you used another visitor instead of using `collect_bindings`?
Each visitor uses more space in the final binary than expected

crates/swc_ecma_transforms_base/src/hygiene/tests.rs
kdy1: Those updates are mostly not intentional.
The renamed `var`s in tests are renamed just because it works and it was required to pass the test.

kdy1: Oh, almost all renamings in his files are intentional.
They have different syntax contexts, so those are different variables.

crates/swc_ecma_transforms_base/src/ident_scope.rs
magic-akari: Oh, do you mean `BindingCollector`, I just find it.

crates/swc_ecma_transforms_base/src/hygiene/tests.rs
kdy1: Tests in this file are extracted from other tests, and those are tests for injected identifiers.

crates/swc_ecma_transforms_base/src/ident_scope.rs
magic-akari: I have tried the `collect_decls` but got the wrong result.
Maybe I should use `find_ids`, but I need to ignore `key` in `KeyValuePatProp` and `value` in `AssignPatProp`.

Just like what I do as mentioned:
https://github.com/swc-project/swc/blob/3afcaba887539e2cce40f5cdd4fbce0511c77ff0/crates/swc_ecma_transforms_base/src/ident_scope.rs#L101-L109

magic-akari: Ok, `find_ids ` works well.

magic-akari:
another case

input:
```javascript
if (true) {
    const a = 1;
    const b = 2;
    function foo(a, b) {
        use(a, b);
    }
} else {
    const a = 1;
    const b = 2;
    function bar(a, b) {
        use(a, b);
    }
}
```

expect output:
```javascript 
if (true) {
    var foo = function foo(a, b) {
        use(a, b);
    };
    var a = 1;
    var b = 2;
} else {
    var bar = function bar(a, b) {
        use(a, b);
    };
    var a1 = 1;
    var b1 = 2;
}
```

main branch output([REPL](https://play.swc.rs/?version=1.2.136&code=H4sIAAAAAAAAA8tMU9AoKSpN1VSo5lIAguT8vOIShUQFWwVDaySBJKCAEUQgrTQvuSQzP08hLT9fI1FHIQmmFQRKi1MhYhC1tVy1Cqk5xamkG56UWESE4VwAPHNdpb8AAAA%3D&config=H4sIAAAAAAAAA0WNSwrDMAxE76J1ts0id%2BghjKsEF%2F%2FQKBBjfPfawSU7aebpqdIXlrZK2QhYxoQS1Vy0kZbMsOKy0kKKHu3Gg1tfjBysHWG8eudTAs92oeCi28sw2RSyMPBUJh7%2BT7YuCulzjqDez27hSu1xzDuH9wRVTm4%2FOkwQUrUAAAA%3D)):
```javascript
if (true) {
    var foo = function foo(a3, b3) {
        use(a3, b3);
    };
    var a2 = 1;
    var b2 = 2;
} else {
    var bar = function bar(a, b) {
        use(a, b);
    };
    var a1 = 1;
    var b1 = 2;
}
```

this PR output:
```javascript
if (true) {
    var foo = function foo(a3, b3) {
        use(a3, b3);
    };
    var a2 = 1;
    var b2 = 2;
} else {
    var bar = function bar(a, b) {
        use(a, b);
    };
    var a1 = 1;
    var b1 = 2;
}
```

I will take a look at why I got this result.

magic-akari:
Closed due to outdated

PR #3220, opened 2022-01-10
feat(swc/plugin_runner): perform actual plugin transform

kwonoj:
**Description:**

This PR implements full end-to-end plugin workflow and it works 🎉 
While there are still lots of places to polish, at least functionality wise swc can load plugin and perform transformation.

PR includes deserialization from plugin's transformed results, also improve integration test actually verifies output from plugin transform. With this test further refactoring can gaurantee transform workflow won't break.

### DIFF
--- a/crates/swc_plugin/src/lib.rs
+++ b/crates/swc_plugin/src/lib.rs
@@ -1,7 +1,10 @@
 // Reexports
-pub use swc_common::chain;
-pub use swc_ecma_ast::*;
-pub use swc_ecma_visit::*;
+pub use swc_common::{chain, DUMMY_SP};
+pub mod ast {
+    pub use swc_atoms::*;
+    pub use swc_ecma_ast::*;
+    pub use swc_ecma_visit::*;
+}
 
 #[doc(hidden)]
 #[cfg(target_arch = "wasm32")]
--- a/crates/swc_plugin_runner/src/lib.rs
+++ b/crates/swc_plugin_runner/src/lib.rs
@@ -7,9 +7,10 @@ use anyhow::{Context, Error};
 use once_cell::sync::Lazy;
 use parking_lot::Mutex;
 use resolve::PluginCache;
+use rkyv::{AlignedVec, Deserialize};
 use swc_common::{collections::AHashMap, plugin::serialize_for_plugin};
 use swc_ecma_ast::Program;
-use wasmer::{imports, Instance, Memory, MemoryType, Module, Store};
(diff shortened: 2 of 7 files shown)


crates/swc_plugin/src/lib.rs
```diff
@@ -1,7 +1,10 @@
 // Reexports
-pub use swc_common::chain;
-pub use swc_ecma_ast::*;
-pub use swc_ecma_visit::*;
+pub use swc_common::{chain, DUMMY_SP};
+pub mod ast {
```
kwonoj, line 3: this is small bikeshedding for the reexports namespace. Bit unsure this would be better or just having everything under top-level reexport would be better, would like to try couple of interatiions.

One gain for having namespace is author can to wildcard for the ast only like

```
swc_plugin::{ast:*, ...} 
```

instead of wildcard import every top level exports.

crates/swc_plugin_runner/src/lib.rs
```diff
@@ -143,14 +148,39 @@ pub fn apply_js_plugin(
...
+        // Reconstruct AlignedVec from ptr returned by plugin
+        let memory = instance.exports.get_memory("memory")?;
+        let ptr: WasmPtr<u8, Array> = WasmPtr::new(returned_ptr as _);
+
+        let mut transformed_serialized = AlignedVec::with_capacity(returned_len);
+
+        // Deref & read through plguin's wasm memory space via returned ptr
+        let derefed_ptr = ptr
+            .deref(memory, 0, returned_len.try_into().unwrap())
+            .unwrap();
```
kwonoj, line 168: these are just good enough verifies end-to-end workflow, further polishing should introduce better handlings (refer //todo comment).

crates/swc_plugin_runner/tests/integration.rs
```diff
@@ -67,11 +91,19 @@ fn internal() -> Result<(), Error> {
...
-                .unwrap();
+                .expect("Plugin should apply transform");
+
+        let mut visitor = TestVisitor {
+            plugin_transform_found: false,
+        };
+        program.visit_with(&mut visitor);
 
-        Ok(())
+        visitor
```
kwonoj, line 103: now integration test actually verifies plugin transformation.

crates/swc_plugin/src/lib.rs
kdy1: Seems like a good idea. I love this :)

kdy1 (approved):
LGTM. Nice work!

PR #3227, opened 2022-01-11
feat(swc/plugin_runner): free guest memory after transform

kwonoj:
**Description:**

PR mostly cleans up previous workaround such as not using swc_common::plugin, and notably introduces few more changes

- now we correctly frees up allocated memory
- between each plugin invocation, no longer does serialization / deserialization on the host side: we know bytes are serialized already so we can pass through plugin transformation then construct Program at the last stage. 

We have few features should be implementated such as error handling, but overall getting quite close to move over some ergonomic support (i.e macro support in plugin fn signatures)


crates/swc/src/plugin.rs
```diff
@@ -42,27 +42,38 @@ struct RustPlugins {
...
+        // transform.
         if let Some(plugins) = &self.plugins {
             for p in plugins {
                 let config_json = serde_json::to_string(&p.1)
                     .context("failed to serialize plugin config as json")?;
 
                 let path = swc_plugin_runner::resolve::resolve(&p.0)?;
 
-                n = swc_plugin_runner::apply_js_plugin(
+                serialized = swc_plugin_runner::apply_js_plugin(
```
kwonoj, line 64: This is the part host skips serialize / deserialize: We do not need to pass `Program` into each plugin since each plugin invocation need to serialize & copy over into guest anyway: instead we supply previous (or initial) raw bytes which can be copied into guest directly.

crates/swc_common/src/plugin.rs
```diff
@@ -8,38 +8,58 @@ use crate::{syntax_pos::Mark, SyntaxContext};
...
-/// Serialize ast to pass into plugins. This is being called from host side.
-/// TODO: https://github.com/swc-project/swc/issues/3167
-#[cfg(feature = "plugin-base")]
-pub fn serialize_for_plugin<T>(t: &T) -> Result<rkyv::AlignedVec, Error>
+/// A struct wraps internal representation of serialized Program to be passed
+/// across plugin's host to guest. Plugin author should not rely on specific
+/// details serialized byte format struct contains: it is strictly
+/// implementation detail which can change anytime.
+#[cfg(any(feature = "plugin-mode", feature = "plugin-rt"))]
+pub struct SerializedProgram(pub rkyv::AlignedVec);
```
kwonoj, line 16: This is mostly to hide implementation detail from plugin author. We may change internal implementation any time, but plugin author should never care about specific data type we use (`AlignedVec`)

crates/swc_common/src/lib.rs
```diff
@@ -62,7 +62,7 @@ pub mod input;
 pub mod iter;
 pub mod macros;
 pub mod pass;
-#[cfg(feature = "plugin-base")]
+#[cfg(any(feature = "plugin-rt", feature = "plugin-mode"))]
```
kdy1, line 65: Is this required?
`plugin-base` is depended by `plugin-rt` and `plugin-mode`.

crates/swc_common/src/plugin.rs
```diff
@@ -8,38 +8,58 @@ use crate::{syntax_pos::Mark, SyntaxContext};
...
 
-/// Serialize ast to pass into plugins. This is being called from host side.
-/// TODO: https://github.com/swc-project/swc/issues/3167
-#[cfg(feature = "plugin-base")]
-pub fn serialize_for_plugin<T>(t: &T) -> Result<rkyv::AlignedVec, Error>
+/// A struct wraps internal representation of serialized Program to be passed
+/// across plugin's host to guest. Plugin author should not rely on specific
+/// details serialized byte format struct contains: it is strictly
+/// implementation detail which can change anytime.
+#[cfg(any(feature = "plugin-mode", feature = "plugin-rt"))]
```
kdy1, line 15: `plugin-base` seems like a better feature to use.

crates/swc_common/src/plugin.rs
kdy1, line 16: I think storing this in `swc_common` is bad for maintenance, because of breaking changes.

kdy1: I made `swc_plugin_js_api` because of this.

crates/swc_common/src/plugin.rs
```diff
@@ -8,38 +8,58 @@ use crate::{syntax_pos::Mark, SyntaxContext};
...
+pub struct SerializedProgram(pub rkyv::AlignedVec);
+
+impl SerializedProgram {
+    #[cfg(any(feature = "plugin-mode", feature = "plugin-rt"))]
+    pub fn new(bytes: Vec<u8>, len: i32) -> SerializedProgram {
+        let mut vec = rkyv::AlignedVec::with_capacity(
+            len.try_into()
```
kdy1, line 22: What should happen if `bytes.len() != len`?
Also, is there a reason to have `len` when we can extract it from `bytes.len()`?

crates/swc_common/src/plugin.rs
```diff
@@ -8,38 +8,58 @@ use crate::{syntax_pos::Mark, SyntaxContext};
...
-pub fn serialize_for_plugin<T>(t: &T) -> Result<rkyv::AlignedVec, Error>
+/// A struct wraps internal representation of serialized Program to be passed
+/// across plugin's host to guest. Plugin author should not rely on specific
+/// details serialized byte format struct contains: it is strictly
+/// implementation detail which can change anytime.
+#[cfg(any(feature = "plugin-mode", feature = "plugin-rt"))]
+pub struct SerializedProgram(pub rkyv::AlignedVec);
+
+impl SerializedProgram {
+    #[cfg(any(feature = "plugin-mode", feature = "plugin-rt"))]
```
kdy1, line 19: `plugin-base` seems better.

crates/swc/Cargo.toml
```diff
@@ -53,7 +53,7 @@ serde = {version = "1", features = ["derive"]}
 serde_json = "1"
 sourcemap = "6"
 swc_atoms = {version = "0.2", path = "../swc_atoms"}
-swc_common = {version = "0.16.0", path = "../swc_common", features = ["sourcemap", "concurrent"]}
+swc_common = {version = "0.16.0", path = "../swc_common", features = ["sourcemap", "concurrent", "plugin-rt"]}
```
kdy1, line 56: `plugin-rt` should be enalbed via a cargo feature in the `node` crate.

`node/plugin` -> `swc/plugin` -> `swc_common/plugin-rt`.

crates/swc_common/src/plugin.rs
```diff
@@ -8,38 +8,58 @@ use crate::{syntax_pos::Mark, SyntaxContext};
...
+                Error::msg("SharedSerializeMapError")
+            }
+        })
 }
 
-/// TODO: https://github.com/swc-project/swc/issues/3167
-#[cfg(feature = "plugin-mode")]
-pub fn deserialize_for_plugin<T>(bytes: &[u8]) -> Result<T, Error>
+/// Deserialize given raw bytes into Program.
+#[cfg(any(feature = "plugin-rt", feature = "plugin-mode"))]
```
kdy1, line 51: `plugin-base` seems better.

crates/swc_common/src/plugin.rs
```diff
@@ -8,38 +8,58 @@ use crate::{syntax_pos::Mark, SyntaxContext};
...
+            }
+            rkyv::ser::serializers::CompositeSerializerError::SharedError(e) => {
+                Error::msg("SharedSerializeMapError")
+            }
+        })
 }
 
-/// TODO: https://github.com/swc-project/swc/issues/3167
-#[cfg(feature = "plugin-mode")]
-pub fn deserialize_for_plugin<T>(bytes: &[u8]) -> Result<T, Error>
```
kdy1, line 31: `deserialize_for_plugin` exists because we need to serialize/deserialize diagnostics generated by plugins.
This should not be renamed, and I don't think we have to make a breaking change here.

tests/rust-plugins/swc_internal_plugin/src/lib.rs
```diff
@@ -29,28 +30,27 @@ impl VisitMut for ConsoleOutputReplacer {
...
-    // allocated memory, copy bytes and pass ptr to it.
+    // allocate memory, copy bytes and pass ptr to plugin.
     let raw_serialized_bytes =
         unsafe { std::slice::from_raw_parts(ast_ptr, len.try_into().unwrap()) };
 
-    // Reconstruct AlignedVec from raw bytes
-    let mut serialized_program = AlignedVec::with_capacity(len.try_into().unwrap());
-    serialized_program.extend_from_slice(raw_serialized_bytes);
+    // Reconstruct SerializedProgram from raw bytes
+    let serialized_program = SerializedProgram::new(raw_serialized_bytes.to_vec(), len);
```
kdy1, line 40: `to_vec` clones memory. Is this cloning required?

kdy1: Original code also copy memory, though.

kdy1 (changes requested):
Thank you!

crates/swc_common/src/plugin.rs
kdy1: I think `swc_common` is not a good place to store this, because breaking change of `swc_common` means breaking change of `swc_ecma_ast`. (Because of the way cargo works)

crates/swc_common/src/lib.rs
kwonoj: Yes, but looks like i.e if `swc_plugin` sets feature to `plugin-mode` plguin-base cfg won't kick off for those target? also this is more explicitly readable this is supported in both modes.

crates/swc_common/src/plugin.rs
kwonoj: as commented above, plugin-runner uses `plugin-rt` and swc_plugin uses `plugin-mode` and `plugin-base` cfg  doesn't seem to work for those two.

crates/swc_common/src/plugin.rs
kwonoj: do you mean swc_common itself's breaking changes, or this struct itself? mind elaborate bit more?

crates/swc_common/src/plugin.rs
kwonoj: vec's len is actually constructed from `len` which is returned from raw bytes ptr + size, so I expect length difference would not occur. I decided to accept size explicitly though, as it's easier to reason about we are constructing a struct from raw ptr instead of vec itself. Not a specific technical reason we have to but I felt it gives better implication.

crates/swc_common/src/plugin.rs
kdy1: I mean a breaking change of this struct is a breaking change of `swc_common` because it's exported.

crates/swc_common/src/plugin.rs
kdy1: Oh, I see. Thanks!

crates/swc/Cargo.toml
kwonoj: I think this is leftover while I move things around.

crates/swc_common/src/plugin.rs
kwonoj: I.. think that's kinda expected? this struct will be only used between host / and the plugin, and by the time we make breaking changes I expect both should have breaking version bump as well.

crates/swc_common/src/lib.rs
kdy1: https://github.com/swc-project/swc/blob/d51904b93e4f252e0036bdd1b6ab8ed44836bce4/crates/swc_common/Cargo.toml#L24-L25

`plugin-rt` and `plugin-mode` both enables `plugin-base`.

crates/swc_common/src/plugin.rs
kdy1: If so, naming seems problematic. We will also serialize other things, not only `swc_ecma_ast::Program`.
How do you think about making the name `Serialized` and renaming `deserialize_for_plugin` as `Serialized.deserialize`, and `serialize_for_plugin` -> `Serialized::new_from`?

kdy1: We need to serialize many things, including diagnostics from a plugin.
Also, in future, we may expand plugin API to bundler/css plugins.

crates/swc_common/src/plugin.rs
kwonoj: It was bikeshedding to have somewhat better clarity, and should be fine to revert.

crates/swc_common/src/plugin.rs
kwonoj: Sounds legit to me. Probabaly we need something like `Serialized<T>` then?

kdy1: Oh, seems like a great idea. Yeah, it will prevent deserializing as a different type.

tests/rust-plugins/swc_internal_plugin/src/lib.rs
kwonoj: probably..not? Since this is plugin code for the testing, I honestly choose some kind of shortcut here.

kdy1: Oh, you are correct. It does not matter I think.
We need to think about once we provide a macro for plugins.

crates/swc_common/src/plugin.rs
kwonoj: actually T seems bit tricky. I'll play with things bit and try to update PR, either having T or not having it.

kdy1: You are right. I took a look at this and it can be hard because it's passed between `extern "C"` boundaries.

crates/swc_common/src/plugin.rs
```diff
@@ -6,40 +6,66 @@
...
-    let archived = unsafe { rkyv::archived_root::<T>(&bytes[..]) };
-
-    archived
-        .deserialize(&mut rkyv::Infallible)
-        .map(|r: rkyv::with::With<T, T>| r.into_inner())
-        .with_context(|| format!("failed to deserialize `{}`", type_name::<T>()))
+pub type PluginSerializedByte = Serialized<rkyv::AlignedVec>;
+
+#[cfg(feature = "plugin-base")]
+impl Serialized<rkyv::AlignedVec> {
```
kwonoj, line 21: @kdy1 thoughts on this? I don't feel this is elegant enough, but this seems could be a way to make `Serialized` struct to be generic across different usecases later.

kdy1: I think this is fine.

crates/swc_common/src/plugin.rs
```diff
@@ -8,38 +8,64 @@ use crate::{syntax_pos::Mark, SyntaxContext};
...
-        }
-        rkyv::ser::serializers::CompositeSerializerError::SharedError(e) => {
-            Error::msg("SharedSerializeMapError")
-        }
-    })
+/// Wraps internal representation of serialized data. Consumers should not
+/// rely on specific details of byte format struct contains: it is
+/// strictly implementation detail which can change anytime.
+pub struct Serialized<T> {
+    field: T,
```
kdy1, line 15: I think `Serialized` should have `AlignedVec` with `PhantomData<T>`.
This seems a bit weird as `T` is not serialized.

crates/swc_common/src/plugin.rs
kdy1: Oh, sorry. I misread the code.

`Serialized<rkyv::AlignedVec>` seems weird to me, as it's not a serialized form of `rkyv::AlignedVec`.

crates/swc_common/src/plugin.rs
kwonoj: if we're going to use rkyv for any kind of serialization, that may works better. I thought we may have some other cases insetad.

crates/swc_common/src/plugin.rs
kwonoj: Yes, that's the part I'm feeling bit.. awkward.

kwonoj: in here T represents internal field type, not the type of original type to be deserialized into. I presume there should be some better way?

crates/swc_common/src/plugin.rs
kdy1: I'm not sure about it at the moment, but if we want to store alternative `Serialized` in `swc_common`, we can make a breaking change.

But I think we can use `rkyv` for almost everything as `rkyv` is zero copy serialization framework. It's faster than `memmove` which is required for `abi_stable`

kwonoj: yes, makes sense. Probably then we don't need to have generic for the wrapping struct, as long as we can hide alignedvec from consumer.

crates/swc_common/src/plugin.rs
kwonoj: ok, removed generic to simplify things, based on discussion https://github.com/swc-project/swc/pull/3227#discussion_r781868216 that we'll rely on rkyv in most of cases for the serialization.

But also, field is hidden and there's no public direct access to rkyv from the consumers, we may not need worry too much for the breaking changes.

kdy1 (approved):
LGTM. Thank you!

PR #3234, opened 2022-01-11
fix(parser): fix `ComputedPropName` span

dsherret:
Fixes the `ComputedPropName` node's span.

### DIFF
--- a/crates/swc_ecma_parser/src/parser/expr.rs
+++ b/crates/swc_ecma_parser/src/parser/expr.rs
@@ -1127,11 +1127,15 @@ impl<'a, I: Tokens> Parser<I> {
             && eat!(self, '['))
             || eat!(self, '[')
         {
+            let bracket_lo = self.input.prev_span().lo;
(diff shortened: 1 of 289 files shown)


crates/swc_ecma_parser/src/parser/expr.rs
```diff
@@ -1127,11 +1127,15 @@ impl<'a, I: Tokens> Parser<I> {
             && eat!(self, '['))
             || eat!(self, '[')
         {
+            let bracket_lo = self.input.prev_span().lo;
             let prop = self.include_in_expr(true).parse_expr()?;
             expect!(self, ']');
             let span = Span::new(obj.span().lo(), self.input.last_pos(), Default::default());
             debug_assert_eq!(obj.span().lo(), span.lo());
-            let prop = ComputedPropName { span, expr: prop };
```
dsherret, line 1134: This line was added and it used `span`, which is the parent span. The span of the `ComputedPropName` should be `[expr]`.

kdy1 (approved):
LGTM. Thank you!

PR #3240, opened 2022-01-12
feat(swc/plugin_macro): plugin macro helper support

kwonoj:
**Description:**

This PR amends ergonomics to write plugin by introducing `plugin_module` macro. There may be better name.

Macro wraps most of low level pointer operation from plugin author. In result, to write plugin required fn signature is like below.

```
#[plugin_module]
fn some_transform(program: Program, config: String) -> Result<Program, Error>
```

macro doesn't care about fn's name, it could be named in any way. But for the convention we'll probably propose recommended name. (`process`? `transform`?)

There are number of undefined open questions in above signatures - consider it is not fully finalized. Especially error bubbling requires some exploration to pass between host to guest.


tests/rust-plugins/swc_internal_plugin/src/lib.rs
```diff
@@ -23,35 +24,18 @@ impl VisitMut for ConsoleOutputReplacer {
...
+/// returning ptr back to host.
+///
+/// It is possible to opt out from macro by writing transform fn manually via
+/// `__plugin_process_impl(*const u8, i32, *const u8, i32) -> (i32,i32)`
+/// if plugin need to handle low-level ptr directly.
+///
+/// TODO:
+/// - better developer experience: println / dbg!() doesn't emit anything
+#[plugin_module]
+pub fn process(program: Program, _plugin_config: String) -> Result<Program, std::fmt::Error> {
```
kwonoj, line 38: Now we're back to original signature passes `Program` and `config` for both and macro hides all pointer interop to reconstruct `Program`.

- config is `string`: I was thinking to have some strict typed config, then realized this is config for the each plugin: plugin should able to change interface to config without changing swc's core. In result, this is a string and plugin author need to decompose as needed.

Few open questions

- should there be some deserialization helper for config into json (or other data types) or we should leave it to userland? (TBD for now)
- will we need any other metadata to be passed? or entire swc config? 
- Return value type: especially error handling is not defined yet, so `std:fmt::Error` in here is stub. Need to dig bit more to serialize error to pass into host.

kdy1: Actually, if we want to support babel-like configuration for plugins, the config can be an arbitrary type.

Type used in `extern "C"` should be fixed, but we can sielently serialize/deserialize for plugin authors. (Maybe by sing serde_json)

I'm not sure if this is good enough.

kdy1: Instead of passing information via `extern "C"`, we can allow plugin authors request for the required information using `imports! {}` macro I guess?
Of course it will require some work, but actually there are really lots of metadata and some of them are quite hard to pass.

kwonoj: Since we get json-string, we can convert / deserialize as we want. But I'm not sure about ergonomics at this moment. Rather to try get early conclusion let's start from string first? in worst cases we could interop into something similar to serde_json:value and let user dynamically look up values.

kwonoj: > information using imports! {} macro I guess

imports! macro is to expose host fn into guest, isn't it?

kdy1: I think starting with `String` is a good idea.

kwonoj: not sure if tls need to be involved here since this lives in wasm's memory space, but as you said it may worth consider acquiring necessary type information to be reconstructed into plugin fn. 

I'd like to suggest this as open q for now, with getting some feedback once we release experimental release for the plugin.

kdy1: I agree. The only thing I worry about is `swc_common::errorrs::Handler` but it will be quite easy to implement support for it.

crates/swc_plugin/Cargo.toml
```diff
@@ -1,19 +1,14 @@
...
-edition = "2021"
+edition = "2018"
 license = "Apache-2.0"
 name = "swc_plugin"
 repository = "https://github.com/swc-project/swc.git"
 version = "0.24.0"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
 [dependencies]
```
kdy1, line 11: We don't re-export plugin macro from `swc_plugin`?

kdy1 (commented):
Mostly LGTM. I have a question.

crates/swc_plugin/Cargo.toml
kwonoj: Good point, I think I just forgot it. Updated.

kdy1 (approved):
LGTM. Nice work!

PR #3242, opened 2022-01-12
Ignore source maps in args in generated fns for classes

0xe:
swc generates mappings in through identifiers passed as arguments in functions generated for providing class support such as _classCallCheck, _inherits. This PR implements changes to ignore these mappings.

Before:
before

After:

after2


kdy1 (changes requested):
This can be done by giving `DUMMY_SP` to corresponding nodes instead.
We should not patch codegen to avoid generating sourcemap of some nodes.

0xe:
> This can be done by giving `DUMMY_SP` to corresponding nodes instead. We should not patch codegen to avoid generating sourcemap of some nodes.

If I give DUMMY_SP for those nodes before clone it treats it as a new identifier. It generates this for example:

gen

kdy1:
You should preserve the syntax context (`.span.ctxt`) of `Ident`. 
https://docs.rs/swc_ecma_ast/0.65.0/swc_ecma_ast/struct.Ident.html

crates/swc_ecma_transforms_compat/src/es2015/classes/mod.rs
```diff
@@ -443,7 +454,7 @@ where
                     init: Some(Box::new(Expr::Call(CallExpr {
                         span: DUMMY_SP,
                         callee: helper!(create_super, "createSuper"),
-                        args: vec![class_name.clone().as_arg()],
+                        args: vec![Expr::Ident(class_name_sym).as_arg()],
```
kdy1, line 457: Please us `class_name_sym.as_arg()` instead of `Expr::Ident(class_name_sym).as_arg()`.

kdy1 (changes requested):
Thank you!

kdy1 (approved):
LGTM. Thank you!

PR #3244, opened 2022-01-12
fix(transforms/typescript): remove rogue println

dsherret:


kwonoj (approved):
I think this is something I can approve confidently 😅 (still it won't allow me to check in though)

kdy1 (approved):
Oh... I missed it. Thanks!

Issue #2635, opened 2021-11-03
[Meta] Plugin system for 2.0


kdy1:
I will modify this as required.

# Javascript plugins

If it's possible to make it fast, I'll support js plugins, too.

# Rust-based plugins

### Task: Proxy `swc_common` (#2646)

 - `swc_common`: Create a trait for thread-local APIs. (cargo feature `plugin-base`)
 - `swc_common`: Implement the trait. (cargo feature `plugin-rt`)
 - `swc_common`: Replace APIs in plugin mode. (cargo feature `plugin-mode`)
 - Pass the trait to plugins and use it to replace `swc_common` APIs.

### Task: `rplugin`: System to pass AST directly (#2671)

This is important for performance.

## Design

I experimented lots of ways for plugin system, and decided to go with [abi_stable](). But there's a problem.

### Problem: Identical API

I want to provide identical API as core transforms. And it requires sharing data stored in thread-local variables.

#### Triage: Dynamic library

I initially thought I need to make `swc_common` a dynamic library so plugins can invoke it. But dynamic linking does not fit for swc plugins. As `swc_common` depends on `std`, the version of `rustc` should be all same among plugins, and chaning version of `rustc` becomes a huge breaking change. I don't want to break all plugins by upgrading `rustc`. So wee need another way than a dynamic library.

---

I posted some questions on the rust community while trying this option, in case you are curious about what I tried.

 - https://www.reddit.com/r/rust/comments/ql1c5m/questions_about_sharing_thread_locals_between/
 - https://users.rust-lang.org/t/sharing-a-crate-with-crate-type-dylib/66768

I also experimented with `extern "C"`. See https://github.com/kdy1/rust-dylib-test

#### Solution: Pass function to plugin with stable abi

Instead of making `swc_common` dynamic library, we can pass a big trait that contains functions related to thread-local variables in `swc_common`.

The plugin runner will pass this trait while invoking plugins. [abi_stable][] can be used to do this.

For plugins, operations of `swc_common` will be replaced by the trait. In other words, functions in `swc_common` will work by calling a method in the passed trait object.

---

[abi_stable]: https://docs.rs/abi_stable/

devongovett:
Excited to see progress here! I think there's still one potential issue with this approach. Adding/updating/removing fields from AST nodes is potentially a breaking change when types are nested without boxes. For example:

```rs
struct Foo {
  foo: u32,
  bar: Bar,
  baz: u32
}

struct Bar {
  test: u32
}
```

[...]

PR #3249, opened 2022-01-13
docs(adr): setting up adr, start to write 00001-plugin

kwonoj:
**Description:**
One of the biggest challenge I had to deal with large codebase was finding out historical context for some behavior, in short `why we do this from beginning`? And sometimes, I made changes without knowing those context then find out changed attempt was already tried and not chosen for various reasons.

For those reason I came to believe maintaining some decision records, history even if it's some short form of text can be meaningful. Among those I am in favor of https://adr.github.io/ and would like to propose we'd do the similar thing.

As a first step, this PR attempts to create initial template - mostly blind copy-paste from https://github.com/mozilla/uniffi-rs/tree/main/docs/adr and trying to fill some gaps for the plugin features we are changing now. I don't have full context for all the stories of plugin as well, so I may ask @kdy1 to fill some gaps especially prior decisions we made & why we don't do it.

- https://github.com/swc-project/swc/issues/2635

### DIFF
--- a/docs/adr/00000-whats-the-big-idea.md
+++ b/docs/adr/00000-whats-the-big-idea.md
@@ -1,0 +1,1 @@
+### This adr attempts to describe 10,000ft view for the swc project itself. To be filled gradually.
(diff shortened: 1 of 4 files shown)


kwonoj:
@kdy1 please feel free to add commits or modify for the blank part of plugin docs.

martindisch:
Thank you! I have been wondering what the deciding factors were between `wasmtime` and `wasmer`, so this ADR was perfect.

PR #3258, opened 2022-01-13
feat(css/parser): `@layer` at-rule

alexander-akait:
**Description:**

`@layer` parsing https://www.w3.org/TR/css-cascade-5/#at-layer

There are couple problems were found:
- `@import` is not supported `supports()`:
```css
@import url("narrow.css") supports(display: flex) handheld and (max-width: 400px);
```
- `@supports` is not supported `declaration` parsing and `selector` parsing

These things will be fixed in future

**BREAKING CHANGE:**

Do we have it for CSS?

No


crates/swc_css_codegen/src/list.rs
```diff
@@ -29,7 +29,8 @@ add_bitflags!(
         SpaceDelimited: 1 << 2,
         SemiDelimited: 1 << 3,
         CommaDelimited: 1 << 4,
-        DelimitersMask: SpaceDelimited | SemiDelimited | CommaDelimited,
+        DotDelimited: 1 << 5,
+        DelimitersMask: SpaceDelimited | SemiDelimited | CommaDelimited | DotDelimited,
```
kdy1, line 33: I think we should change value of`AllowTrailingComma` to `1 << 6`>

kdy1 (changes requested):
Mostly LGTM, thanks!

I found an issue related to bitflags.

alexander-akait:
@kdy1 Fixed :+1:

kdy1 (approved):
LGTM. Thank you!

PR #3260, opened 2022-01-13
fix: resolve todo

alexander-akait:
**Description:**

resolve todo, first todo was due problems with spaces around legacy `url` parsing, now we avoid `reset_to`, and just use `last_pos` (it was design for this goals) and it should be faster

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thank you!

PR #3269, opened 2022-01-14
feat(lint): "no-console" implementation

ArturAralin:
**Description:**
Add simple "no-console" rule implementation. 

p.s. I trying to figure out how to write lint rules with swc


crates/swc_ecma_lints/src/rules/mod.rs
```diff
@@ -5,12 +5,14 @@ use swc_ecma_visit::{noop_fold_type, Fold};
...
 mod duplicate_bindings;
 mod duplicate_exports;
+mod no_console;
 
 pub fn all() -> Vec<Box<dyn Rule>> {
     vec![
         const_assign::const_assign(),
         duplicate_bindings::duplicate_bindings(),
         duplicate_exports::duplicate_exports(),
+        no_console::no_console(),
```
kdy1, line 15: `no_console` should not be a hard error, unless there's config for it.

crates/swc_ecma_lints/src/rules/no_console.rs
```diff
@@ -0,0 +1,31 @@
+use crate::rule::{visitor_rule, Rule};
+use swc_common::errors::HANDLER;
+use swc_ecma_ast::*;
+use swc_ecma_visit::{noop_visit_type, Visit};
+
+const MESSAGE: &'static str = "Unexpected console statement";
+
+pub fn no_console() -> Box<dyn Rule> {
```
kdy1, line 8: You can accept `top_level_ctxt: SyntaxContext` instead.

crates/swc_ecma_lints/src/rules/no_console.rs
```diff
@@ -0,0 +1,31 @@
...
+    visitor_rule(NoConsole::default())
+}
+
+#[derive(Debug, Default)]
+struct NoConsole;
+
+impl NoConsole {
+    fn check(&mut self, id: &Ident) {
+        if id.sym.eq("console") {
+            if id.span.ctxt.outer().as_u32() == 1 {
```
kdy1, line 18: This is wrong.

crates/swc_ecma_lints/src/rules/no_console.rs
```diff
@@ -0,0 +1,31 @@
...
+            }
+        }
+    }
+}
+
+impl Visit for NoConsole {
+    noop_visit_type!();
+
+    fn visit_ident(&mut self, id: &Ident) {
+        self.check(id);
```
kdy1, line 29: This will also error for `foo.console`

kdy1 (changes requested):

crates/swc_ecma_lints/src/rules/mod.rs
ArturAralin: should I use `struct_span_warn` instead of `struct_span_err`?

crates/swc_ecma_lints/src/rules/no_console.rs
ArturAralin: I was try solve this with `if id.span.ctxt.outer().as_u32() == 1 {`, but you told that is wrong. I need advice

crates/swc_ecma_lints/src/rules/no_console.rs
kdy1: You have to check if it's `top_level_ctxt`, which should be provided on creation.

ArturAralin: I can't understand where I should put top_level_ctxt into `fn no_console()`. It means that I should call `no_console(some_top_level_ctx)` ?

kdy1: Yes

ArturAralin: then I should compare top file context and identifier span context?

smth like?

```
if id.span == self.top_file_context {
  // write warn
}
```

kdy1: Not span itself,`.span.ctxt`

ArturAralin: Ok. And one more question. How to get access to file top level context? Should I use visit_module/visit script for it?

kdy1: > You have to check if it's top_level_ctxt, which should be provided on creation

You should accept it from `no_console`.

ArturAralin: It's correct? 

```
let m: Module = ....;
// do stuff

let rule = no_console(m.span.ctxt.clone());
rule.lint_module(&m);
```

kdy1: No. I mean `pub fn no_console(top_level_ctxt: SyntaxContext)`

ArturAralin: Above I assume rule call code.

Of course I'll add argument to no_console fn, place SyntaxContext to struct and other things

kdy1: No, the span of the module does not have any context.
You should use the `Mark` (or `SytnaxContext`) passed to `resolver_with_mark`.

ArturAralin: Can you get my any example? A bit hard to understand

kdy1: Many passes accept `top_level_mark` or `top_level_ctxt`. See https://github.com/swc-project/swc/blob/a4e93eb2dff0614e7a39c4301c84654412cf68dc/crates/swc/src/config/mod.rs#L367-L456

Top level context here is `SyntaxContext::empty().apply_mark(top_level_mark)`.

kdy1 (changes requested):
Can you add some tests about this?

kdy1:
@kwonoj Thoughts?
I think this should be an off-by-default config option, at least.

kwonoj:
Whem / where do we invoke lint rules? Transform / or separate paths to execute rules?

Either cases all lints should be off by default in my opinion: user need to configure as they want.

P.s: impl looks ok to me.

ArturAralin:
@kdy1 @kwonoj any decision made? I'd want to implement several rules like this on this weekend

kdy1:
Can you make a type `Config` in `swc_ecma_lints` which has `#[non_exhaustive]`, `#[serde(rename_all = "camelCase")]`, #[derive(Debug, Clone, Copy, Deserialize)]` with a field `no_console: bool`, and then expose it via `JscConfig.lints` with `#[serde(default)]`?

I think `no-console` should be opt-in.

ArturAralin:
@kdy1 sure, I can. 

Should I consider lint config in tests or just keep check all rules?

ArturAralin:
Also I think we're should use enum with following variants OFF, WARNING, ERROR instead of boolean

kdy1:
I agree with the enum idea :)
Sorry for late response

crates/swc_ecma_lints/src/rules/mod.rs
```diff
@@ -1,17 +1,25 @@
...
-    vec![
+pub fn all(lint_config: &LintConfig, top_level_ctxt: SyntaxContext) -> Vec<Box<dyn Rule>> {
+    let mut rules = vec![
         const_assign::const_assign(),
         duplicate_bindings::duplicate_bindings(),
         duplicate_exports::duplicate_exports(),
-    ]
+    ];
+
+    if let Some(rule) = no_console::no_console(&lint_config.no_console, top_level_ctxt.clone()) {
```
kdy1, line 18: `.clone()` is not required as it's `Copy`.

kdy1, line 18: Also, this can be converted into `rules.extend`

crates/swc_ecma_lints/src/rules/no_console.rs
```diff
@@ -0,0 +1,58 @@
...
+impl NoConsole {
+    fn new(expected_reaction: LintRuleReaction, top_level_ctxt: SyntaxContext) -> Self {
+        Self {
+            expected_reaction,
+            top_level_ctxt,
+        }
+    }
+
+    fn check(&mut self, id: &Ident) {
+        if id.sym.eq("console") {
```
kdy1, line 36: You can use `&*id.sym == "console"` instead

crates/swc_ecma_lints/src/rules/no_console.rs
kdy1: This is wrong because `foo.console` is not a target.of this rule

crates/swc_ecma_lints/tests/fixture.rs
```diff
@@ -37,9 +40,16 @@ fn pass(input: PathBuf) {
         let mut parser = Parser::new_from(lexer);
 
         let mut m = parser.parse_module().unwrap();
-        m.visit_mut_with(&mut resolver());
+        let top_level_mark = Mark::fresh(Mark::root());
 
-        let rules = all();
+        m.visit_mut_with(&mut resolver_with_mark(top_level_mark.clone()));
```
kdy1, line 45: `clone()` seems redundant

crates/swc_ecma_lints/tests/pass/no_console/1/input.js
```diff
@@ -0,0 +1,8 @@
+object.console.log("message");
+
+(console) => {
+    console.log("message");
+};
+
```
kdy1, line 6: Seems like test file is missing output.

kdy1 (changes requested):

crates/swc_ecma_lints/tests/pass/no_console/1/input.js
ArturAralin: @kdy1 quite don't understand what you're mean

kdy1: I mean it does not have any reference file to compare with.
`UPDATE=1 cargo test` should generate it, if it's implemented correctly.

ArturAralin: @kdy1 now results compare is not implemented for lints crate. If I understand correctly I should read output from handler and write it to file with `NormalizedOutput` to compare outputs in future. Now I can't understand how to get output from Handler

kdy1: I think you can use integration test suite of the `swc` crate in this case.

You can add tests in https://github.com/swc-project/swc/tree/74b0624c0c572d462ca7462a3325c6b0f89ae3b5/crates/swc/tests/errors/lints with `.swcrc` which is json file like

```json
{
    "jsc": {
        "lints": {
            "noConsole": "error"
        }
    }
}
```


Note that file name should be `input.js`, and .`swcrc` is searched upward starting from the directory of `input.js`.

After doing so, you can run `UPDATE=1 cargo test` on `swc` crate directory to create test reference files.

crates/swc_ecma_lints/src/config.rs
```diff
@@ -0,0 +1,22 @@
...
+    fn default() -> Self {
+        Self::Off
+    }
+}
+
+#[derive(Debug, Clone, Default, Copy, Serialize, Deserialize)]
+#[non_exhaustive]
+#[serde(rename_all = "camelCase")]
+pub struct LintConfig {
+    pub no_console: LintRuleReaction,
```
kdy1, line 21: Can you add `#[serde(default)]` here?

crates/swc/src/config/mod.rs
```diff
@@ -368,6 +369,8 @@ impl Options {
             .global_mark
             .unwrap_or_else(|| Mark::fresh(Mark::root()));
 
+        let top_level_ctxt = SyntaxContext::empty().apply_mark(top_level_mark.clone());
```
kdy1, line 372: `.clone()` seems redundant.

kdy1 (changes requested):
Thank you!

crates/swc/src/config/mod.rs
ArturAralin: oops. missed here

crates/swc_ecma_lints/src/config.rs
ArturAralin: @kdy1 ok.

Now I think about many rules that have settings e.g. no-console has exceptions for methods seems like

```
"no-console": ["error", { "allow": ["warn"] }]
```

maybe we're should use similar structure? 

```
struct RuleConfig<T: Serialize + Deserialize + rest...>(LintRuleReaction, T);

struct NoConsoleConfig {
  allow: Optional<HashSet<String>>,
}

struct LintConfig {
  no_console: LintRuleReaction<NoConsoleConfig>
}
```

kdy1: Yeah, I think it's the correct approach, and we need to accept various values like `"error"` or `{ "allow": "warn" }`

But I think we can postpone it at the moment.

ArturAralin: Ok. I'll fix now

kdy1: Oh, cool!
Thanks

crates/swc_ecma_lints/src/rules/no_console.rs
```diff
@@ -0,0 +1,69 @@
...
+use swc_common::{errors::HANDLER, SyntaxContext};
+use swc_ecma_ast::*;
+use swc_ecma_visit::{noop_visit_type, Visit};
+
+const MESSAGE: &str = "Unexpected console statement";
+
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct NoConsoleConfig {
+    // not used for now
+    allow: Option<HashSet<String>>,
```
kdy1, line 16: Can you use `AHasetSet` instead?
One in `swc_common`

crates/swc_ecma_lints/src/config.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct RuleConfig<T: Debug + Clone + Serialize + Default>(
+    #[serde(default)] LintRuleReaction,
+    #[serde(default)] T,
+);
+
+impl<T: Debug + Clone + Serialize + Default> RuleConfig<T> {
+    pub fn get_rule_reaction(&self) -> &LintRuleReaction {
```
kdy1, line 26: Can you change this to `pub(crate)`?

crates/swc_ecma_lints/src/config.rs
```diff
@@ -0,0 +1,41 @@
...
+        &self.0
+    }
+
+    pub fn get_rule_config(&self) -> &T {
```
kdy1, line 30: Same here, `pub(crate)` seems better.

kdy1 (changes requested):
Thank you! 
It's quite close 👍

kdy1 (approved):
Thank you!

crates/swc_ecma_lints/src/config.rs
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    pub(crate) fn get_rule_config(&self) -> &T {
+        &self.1
+    }
+}
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[non_exhaustive]
+#[serde(rename_all = "camelCase")]
```
magic-akari, line 37: For better compatibility with the existing lint config, it is better to use the `kebab-case`.
@ArturAralin @kdy1

kdy1: Maybe we can add lots of aliases instead

ArturAralin: Aliases bring more boilerplate code

PR #3270, opened 2022-01-14
feat(swc/plugin_runner): free allocated memory when errors

kwonoj:
**Description:**

This is small bikeshedding for the host side runner. There are multiple paths can short-curcuit if error occurs, and by having simple `Drop` host try to free memory as much as it can. A struct was introduced to keep reference to allocated ptr to track.

### DIFF
--- a/crates/swc_plugin_macro/src/lib.rs
+++ b/crates/swc_plugin_macro/src/lib.rs
@@ -25,6 +25,8 @@ fn handle_func(func: ItemFn) -> TokenStream {
     let ret = quote! {
         #func
 
+        use swc_plugin::Serialized;
+
         // TODO:
         // error handling & bubbling
         #[no_mangle]
--- a/crates/swc_plugin_runner/src/lib.rs
+++ b/crates/swc_plugin_runner/src/lib.rs
@@ -108,70 +108,129 @@ fn load_plugin(plugin_path: &Path, cache: &mut Option<PluginCache>) -> Result<In
     };
 }
 
-/// Copy host's serialized bytes into guest (plugin)'s allocated memory.
-/// Once transformation completes, host should free allocated memory.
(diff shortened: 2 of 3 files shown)


crates/swc_plugin_runner/src/lib.rs
```diff
@@ -108,70 +108,129 @@ fn load_plugin(plugin_path: &Path, cache: &mut Option<PluginCache>) -> Result<In
...
+    }
+}
 
-    Ok(Serialized::new_for_plugin(&transformed_raw_bytes[..], len))
+impl Drop for PluginTransformTracker {
+    fn drop(&mut self) {
+        for ptr in self.allocated_ptr_vec.iter() {
+            self.exported_plugin_free
+                .call(ptr.0, ptr.1)
+                .expect("Failed to free memory allocated in the plugin");
```
kwonoj, line 228: `free` also can fail, but if that occurs it's probably happening in runtime (wasmer) side which we can't recover to proceed anyway, unwrapping via `expect`.

kwonoj:
cargo-crev installation fails: https://github.com/RustCrypto/utils/issues/714

crates/swc_plugin_macro/src/lib.rs
```diff
@@ -25,6 +25,8 @@ fn handle_func(func: ItemFn) -> TokenStream {
     let ret = quote! {
         #func
 
+        use swc_plugin::Serialized;
```
kwonoj, line 28: this is minor fix to import module in macro correctly to avoid actual plugin have to manually import this.

kdy1 (approved):
LGTM. I checked some code related to `Serialized` because I thought it may cause UB, but we were actually copying the slice, so it's fine :)
Nice work!

PR #3271, opened 2022-01-15
feat(swc/plugin_runner): support plugin targets wasm32-wasi

kwonoj:
**Description:**

This PR changes plugin target to be built against `wasm32-wasi`, instead of previous target `wasm32-unknown-unknown`. Plugin runner adjusted to setup proper env to run wasi plugin binaries.

`unknown-unknown` works well for our general purpose, but due to it assumes host target is `unknown`, lot of std calls gets into stub:

> wasm32-unknown-unknown - this target, like the WASI one, is focused on producing single *.wasm binaries. The standard library, however, is largely stubbed out since the "unknown" part of the target means libstd can't assume anything. This means that while binaries will likely work in wasmtime, common conveniences like println! or panic! won't work.

(https://docs.wasmtime.dev/wasm-rust.html)

As description says, plugin author won't have easy way to use `println!` or any other std related calls while making a plugin. wasi target allows to use those seamlessly.

Comparing size between those 2 using our `internal-plugin` test cases, bytesize increase is negligible.

```
./bulid.sh --release

 993335 Jan 15 00:08 swc_internal_plugin.wasm //wasm32-unknown-unknown
1021056 Jan 15 00:11 swc_internal_plugin.wasm //wasm32-wasi
```

**This is opt-in change: plugin_runner supports `wasm32-unknown-unknown` and `wasm32-wasi` binary both. Plugin author can choose depends on requirement to std in plugin.**

** Possible problem
- it is unclear how this'll affect runtimes like browser, However, we don't have any clear path to support plugin for `@swc/wasm` at all yet. so it's not a hard blocker for now.

** Possible alternative
- we may able to expose limited capabilites via wrapper fn exposed into guest from host, like `print_via_host` which proxies call to host. It'll involve serialization / deserialization and possibly limits output format as only known one.

**BREAKING CHANGE:**

~I doubt it, but if any plugin is written in `wasm32-unknown-unknown` plugin need to be re-built against wasm32-wasi.~

PR updated to support unknown-unknown and wasi both.

### DIFF
--- a/crates/swc_plugin_runner/src/lib.rs
+++ b/crates/swc_plugin_runner/src/lib.rs
@@ -8,8 +8,9 @@ use once_cell::sync::Lazy;
 use parking_lot::Mutex;
 use resolve::PluginCache;
 use swc_common::{collections::AHashMap, plugin::Serialized};
-use wasmer::{imports, Array, Instance, Memory, MemoryType, Module, Store, WasmPtr};
+use wasmer::{Array, Exports, Instance, Memory, MemoryType, Module, Store, WasmPtr};
 use wasmer_cache::{Cache, Hash};
+use wasmer_wasi::WasiState;
 
 pub mod resolve;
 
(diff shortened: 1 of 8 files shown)


crates/swc_plugin_runner/src/lib.rs
```diff
@@ -96,9 +97,37 @@ fn load_plugin(plugin_path: &Path, cache: &mut Option<PluginCache>) -> Result<In
     return match module {
         Ok(module) => {
             let memory = Memory::new(&wasmer_store, MemoryType::new(1, None, false))?;
-            let import_object = imports! {
-                "env" => {
-                    "memory" => memory
+            let wasi_version = get_wasi_version(&module, true);
```
kwonoj, line 100: If plugin binary have correct & strict wasi version, we expect it is compiled as `wasm32-wasi`, setting up runtime import object for wasi. Otherwise we'll fall back to default import_object.

kdy1: I think this is a good change, but if I failed to use `wasi` with wasm compiled to `unknown-unknown`.
Seems like you find a way :)

kdy1 (approved):
Nice work!
I was blocked once because of it.

PR #3279, opened 2022-01-15
refact(ecma/utils): merge super field visitor with FnEnvHoister

### LINKED ISSUE #2716: Async to regenerator explicitly binds this context when created

### LINKED ISSUE #2895: Reference to `arguments` is invalid inside an async function and a class
### Describe the bug

If an async method of a class is using the `arguments` object, the reference to the object is lost.

I have a reproduction here https://github.com/egilsster/swc-jest-issue/blob/main/src/args.test.ts

This starts to work with `target` set to `es2018`

Related to https://github.com/swc-project/swc/issues/2786

### Input code

```typescript
export class Quirk {
  async doStuff() {
    const args = arguments;
    console.log(args);
    return { foo: null, ...args[0] };
  }
}
```

```typescript
const args = { one: 1, two: 2 };
const quirk = new Quirk();
// @ts-ignore using old js syntax
const res = await quirk.doStuff(args);
```

### Config

```json
{
  "jsc": {
    "target": "es2015",
    "parser": {
      "syntax": "typescript"
    }
  },
  "sourceMaps": true,
  "inlineSourcesContent": true
}
```

### Playground link

_No response_

### Expected behavior

I expect the `arguments` object to be `{ '0': { one: 1, two: 2 } }` in my case.

### Version

1.2.112

### Additional context

_No response_

Austaras:
**Description:**
`FnEnvHoister` should transform super field access too
**BREAKING CHANGE:**
rename `FnEnvState` into `FnEnvHoister`. The origin API design doesn't make any sense
closes #2895 closes #2716

This is still not enough: for browsers like chrome 46 that supports class but not arrow, super field access cannot be transfomed into arrow calls, but I don't believe anyone is actually using them.


Austaras:
Some snapshot tests are failing, why that happens?

Austaras:
> Some snapshot tests are failing, why that happens?

Seems like default hashmap has a random iter order

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
```diff
@@ -126,57 +145,285 @@ impl FnEnvState {
...
                 kind: MetaPropKind::NewTarget,
                 ..
             }) => {
                 let target = self
-                    .state
                     .new_target
                     .get_or_insert_with(|| private_ident!("_newtarget"));
                 *e = Expr::Ident(target.clone());
             }
+            // super.foo = 123 => super_get_foo = (value) => super.foo = value
```
kdy1, line 203: Doesn't we need special handling for `super.foo++`?

crates/swc_ecma_utils/src/factory.rs
```diff
@@ -60,6 +60,18 @@ pub trait ExprFactory: Into<Expr> {
         })
     }
 
+    #[inline]
+    fn call_fn(self, args: Vec<ExprOrSpread>, span: Span) -> Expr {
```
kdy1, line 64: Can you change the order of arguments?
I think span should come first for consistency.

kdy1 (changes requested):
Thanks! Mostly LGTM!

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
Austaras: Yes, I somehow forget that

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
```diff
@@ -126,57 +145,378 @@ impl FnEnvState {
...
+
+                            let op = op.to_update();
+
+                            let args = if let Some(op) = op {
+                                let tmp = private_ident!("tmp");
+                                self.extra_ident.push(tmp.clone());
+                                vec![
+                                    Expr::Assign(AssignExpr {
+                                        span: DUMMY_SP,
+                                        left: PatOrExpr::Pat(Box::new(Pat::Ident(
```
kdy1, line 252: Can you change this to `PatOrExpr::Pat(tmp.clone().into())`?

Recently I did some work to ease creation of AST and core transforms are the things which plugin authors will look if they are blocked.

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
```diff
@@ -126,57 +145,378 @@ impl FnEnvState {
...
+                                        ))),
+                                        op: op!("="),
+                                        right: c.expr.take(),
+                                    })
+                                    .as_arg(),
+                                    Expr::Bin(BinExpr {
+                                        span: DUMMY_SP,
+                                        left: Box::new(Expr::Call(CallExpr {
+                                            span: DUMMY_SP,
+                                            callee: Callee::Expr(Box::new(
```
kdy1, line 263: I think `self.super_get_computed(DUMMY_SP).as_callee()` would be better.

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
```diff
@@ -126,57 +145,378 @@ impl FnEnvState {
...
+                                    })
+                                    .as_arg(),
+                                ]
+                            } else {
+                                vec![c.expr.take().as_arg(), right.take().as_arg()]
+                            };
+                            *e = Expr::Call(CallExpr {
+                                span: *span,
+                                args,
+                                callee: Expr::Ident(callee).as_callee(),
```
kdy1, line 280: Please use `callee.as_callee()` instead.

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
```diff
@@ -126,57 +145,378 @@ impl FnEnvState {
...
+                ..
+            }) => {
+                if let Expr::SuperProp(super_prop) = &mut **expr {
+                    match &mut super_prop.prop {
+                        SuperProp::Computed(c) => {
+                            let callee = self.super_get_computed(super_prop.span);
+                            let call = Expr::Call(CallExpr {
+                                span: *span,
+                                args: vec![c.expr.take().as_arg()],
+                                callee: Expr::Ident(callee).as_callee(),
```
kdy1, line 324: `callee.as_callee()` would be better.

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
```diff
@@ -126,57 +145,378 @@ impl FnEnvState {
...
+
+                    let op = match op {
+                        op!("++") => op!(bin, "+"),
+                        op!("--") => op!(bin, "-"),
+                    };
+
+                    let update_expr = Expr::Bin(BinExpr {
+                        span: DUMMY_SP,
+                        left: Box::new(Expr::Ident(tmp.clone())),
+                        right: Box::new(Expr::Lit(Lit::Num(Number {
```
kdy1, line 375: Can you use `right: 1.0.into()` instead?

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
```diff
@@ -126,57 +145,378 @@ impl FnEnvState {
...
+                            value: 1_f64,
+                        }))),
+                        op,
+                    })
+                    .as_arg();
+
+                    let mut exprs = match prop {
+                        SuperProp::Ident(id) => vec![
+                            Box::new(Expr::Assign(AssignExpr {
+                                left: PatOrExpr::Pat(Box::new(tmp.clone().into())),
```
kdy1, line 386: Same here. `Box::new` can be removed I think?

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
```diff
@@ -126,57 +145,378 @@ impl FnEnvState {
...
+                                    .as_call(*arg_span, vec![update_expr]),
+                            ),
+                        ],
+                        SuperProp::Computed(c) => {
+                            let prop = private_ident!("prop");
+                            self.extra_ident.push(prop.clone());
+                            vec![
+                                Box::new(Expr::Assign(AssignExpr {
+                                    span: *arg_span,
+                                    left: PatOrExpr::Pat(Box::new(tmp.clone().into())),
```
kdy1, line 405: I think `Box::new` can be removed.

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
```diff
@@ -126,57 +145,378 @@ impl FnEnvState {
...
+                                    op: op!("="),
+                                    right: Box::new(self.super_get_computed(*arg_span).as_call(
+                                        *arg_span,
+                                        vec![Expr::Assign(AssignExpr {
+                                            span: c.span,
+                                            left: PatOrExpr::Pat(Box::new(Pat::Ident(
```
kdy1, line 411: Please use `prop.clone().into()` instead of `Box::new(Pat::Ident(prop.clone().into()))`.

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
```diff
@@ -126,57 +145,378 @@ impl FnEnvState {
...
+                    e.visit_mut_children_with(self)
+                }
+            }
+            Expr::SuperProp(SuperPropExpr { prop, span, .. }) => match prop {
+                SuperProp::Computed(c) => {
+                    c.expr.visit_mut_children_with(self);
+                    *e = Expr::Call(CallExpr {
+                        span: *span,
+                        args: vec![c.expr.take().as_arg()],
+                        callee: Expr::Ident(self.super_get_computed(*span)).as_callee(),
```
kdy1, line 441: `Expr::Ident()` can be omitted.

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
```diff
@@ -126,57 +145,378 @@ impl FnEnvState {
...
+                        type_args: None,
+                    })
+                }
+                SuperProp::Ident(id) => {
+                    *e = Expr::Call(CallExpr {
+                        span: *span,
+                        args: Vec::new(),
+                        callee: Expr::Ident(self.super_get(id, *span)).as_callee(),
```
kdy1, line 449: `Expr::Ident()` can be omitted.

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
```diff
@@ -269,3 +609,104 @@ impl<'a> VisitMut for InitThis<'a> {
...
+            }))),
+            definite: false,
+        }
+    }));
+    if let Some(id) = set.computed {
+        let prop = private_ident!("_prop");
+        let value = private_ident!("_value");
+        decls.push(VarDeclarator {
+            span: DUMMY_SP,
+            name: Pat::Ident(id.into()),
```
kdy1, line 687: `Pat::Ident()` can be omitted.

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
```diff
@@ -269,3 +609,104 @@ impl<'a> VisitMut for InitThis<'a> {
...
+            return_type: None,
+            type_params: None,
+        }))),
+        definite: false,
+    }));
+    if let Some(id) = get.computed {
+        let param = private_ident!("_prop");
+        decls.push(VarDeclarator {
+            span: DUMMY_SP,
+            name: Pat::Ident(id.into()),
```
kdy1, line 636: `Pat::Ident()` can be omitted.

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
```diff
@@ -269,3 +609,104 @@ impl<'a> VisitMut for InitThis<'a> {
...
+                type_params: None,
+            }))),
+            definite: false,
+        });
+    }
+    decls.extend(set.ident.into_iter().map(|(key, ident)| {
+        let value = private_ident!("_value");
+        VarDeclarator {
+            span: DUMMY_SP,
+            name: Pat::Ident(ident.into()),
```
kdy1, line 660: This `Pat::Ident` can be omitted.

crates/swc_ecma_utils/src/function/fn_env_hoister.rs
```diff
@@ -269,3 +609,104 @@ impl<'a> VisitMut for InitThis<'a> {
         }
     }
 }
+
+fn extend_super(decls: &mut Vec<VarDeclarator>, get: SuperField, set: SuperField) {
+    decls.extend(get.ident.into_iter().map(|(key, ident)| VarDeclarator {
+        span: DUMMY_SP,
+        name: Pat::Ident(ident.into()),
```
kdy1, line 616: `Pat::Ident` can be omitted.

kdy1 (changes requested):
Thank you for the hard work!
I found some code style issues.

kdy1 (approved):
LGTM. Thank you!

Issue #3237, opened 2022-01-12
panicked when trying to transform a array with Template literal


jerebtw:
### Describe the bug

panicked when trying to transform a array with Template literals

```bash
thread '<unnamed>' panicked at 'internal error: entered unreachable code: jsonify: Expr Tpl(Tpl { span: Span { lo: BytePos(12), hi: BytePos(17), ctxt: #0 }, exprs: [], quasis: [TplElement { span: Span { lo: BytePos(13), hi: BytePos(16), ctxt: #0 }, tail: true, cooked: Some(Str { span: Span { lo: BytePos(13), hi: BytePos(16), ctxt: #0 }, value: Atom('lel' type=inline), has_escape: false, kind: Normal { contains_quote: false } }), raw: Str { span: Span { lo: BytePos(13), hi: BytePos(16), ctxt: #0 }, value: Atom('lel' type=inline), has_escape: false, kind: Normal { contains_quote: false } } }] }) cannot be converted to json', crates\swc_ecma_transforms_optimization\src\json_parse.rs:132:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
failed to handle: internal error: entered unreachable code: jsonify: Expr Tpl(Tpl { span: Span { lo: BytePos(12), hi: BytePos(17), ctxt: #0 }, exprs: [], quasis: [TplElement { span: Span { lo: BytePos(13), hi: BytePos(16), ctxt: #0 }, tail: true, cooked: Some(Str { span: Span { lo: BytePos(13), hi: BytePos(16), ctxt: #0 }, value: Atom('lel' type=inline), has_escape: false, kind: Normal { contains_quote: false } }), raw: Str { span: Span { lo: BytePos(13), hi: BytePos(16), ctxt: #0 }, value: Atom('lel' type=inline), has_escape: false, kind: Normal { contains_quote: false } } }] }) cannot be converted to json
Failed to compile 1 file with swc.
Error: Failed to compile
    at initialCompilation (C:\Users\Jere\Desktop\Software\GitHub\TypescriptSWC\node_modules\@swc\cli\lib\swc\dir.js:162:19)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async dir (C:\Users\Jere\Desktop\Software\GitHub\TypescriptSWC\node_modules\@swc\cli\lib\swc\dir.js:222:5)
```

### Input code

```typescript
let test = [`lel`];

console.log(test);
```

### Config

```json
{
[...]

PR #3282, opened 2022-01-16
fix(ecma/transforms/optimizer/jsonify): handle tpl when jsonify

RiESAEX:
**Description:**
make it can handle template 
if the template is empty str or don't have any expr.

fix: #3237


crates/swc_ecma_transforms_optimization/src/json_parse.rs
```diff
@@ -129,6 +129,13 @@ fn jsonify(e: Expr) -> Value {
         Expr::Lit(Lit::Num(Number { value, .. })) => Value::Number((value as i64).into()),
         Expr::Lit(Lit::Null(..)) => Value::Null,
         Expr::Lit(Lit::Bool(v)) => Value::Bool(v.value),
+        Expr::Tpl(Tpl { quasis, .. }) => Value::String(match quasis.get(0) {
```
kdy1, line 132: This will error for template literal with interpolation, because there's no check for length of `quasis`.

kdy1 (changes requested):
Can you add some tests involving template literals with interpolations?

crates/swc_ecma_transforms_optimization/src/json_parse.rs
RiESAEX: we checked length of tpl.expr at here https://github.com/swc-project/swc/blob/7ec6e94cb05929c36e28a59670b23f54472079f9/crates/swc_ecma_utils/src/lib.rs#L1466. 
so I think at here the length of quasis will always be 0 or 1.

kdy1: Oh, you are right. Thanks!

kdy1 (approved):
Thank you!

Issue #3283, opened 2022-01-16
expr_simplifier removes `new String("...")` constructor call


mischnic:
### Describe the bug

`expr_simplifier` optimizes away a `new String("...")` String constructor call. But this isn't always safe. 

The input code will log "hello", while the incorrectly optimized output code prints undefined.

### Input code

```typescript
var s = new String("*");
s.test = "hello";
console.log(s.test);
```

### Config

```rust
use swc_common::{
    chain, comments::SingleThreadedComments, sync::Lrc, FileName, Globals, Mark, SourceMap,
    SyntaxContext, DUMMY_SP,
};
use swc_ecma_ast::{Decl, Ident, ImportSpecifier, Invalid, Module};
use swc_ecma_codegen::{text_writer::JsWriter, Config, Emitter};
use swc_ecma_parser::{lexer::Lexer, EsConfig, PResult, Parser, StringInput, Syntax, TsConfig};
use swc_ecma_preset_env::{preset_env, BrowserData, Mode, Query, Targets, Version};
use swc_ecma_transforms::{
    compat::reserved_words::reserved_words,
    fixer, helpers, hygiene,
    optimization::simplify::{dead_branch_remover, expr_simplifier},
    react,
    resolver::resolver_with_mark,
};
use swc_ecma_transforms_compat::es2015;
use swc_ecma_visit::FoldWith;

fn main() {
    let cm = Lrc::<SourceMap>::default();
    let src = r#"
var s = new String("*");
s.test = "hello";
console.log(s.test);
"#;
    let (module, comments) = parse(src, "test.js", &cm).unwrap();

    swc_common::GLOBALS.set(&Globals::new(), || {
        helpers::HELPERS.set(&helpers::Helpers::default(), || {
            let global_mark = Mark::fresh(Mark::root());
            let module = module.fold_with(&mut resolver_with_mark(global_mark));

            let transform = &mut expr_simplifier(Default::default());
            let module = module.fold_with(transform);

            let module = module.fold_with(&mut chain!(
                reserved_words(),
                hygiene(),
                fixer(Some(&comments))
            ));

            let code = emit(&module, &comments, cm);
            println!("{}", code);
        });
    });
}

fn parse(
    code: &str,
    filename: &str,
    cm: &Lrc<SourceMap>,
) -> PResult<(Module, SingleThreadedComments)> {
    let source_file = cm.new_source_file(FileName::Real(filename.into()), code.into());
    let comments = SingleThreadedComments::default();

[...]

PR #3284, opened 2022-01-16
fix(es/transforms): Remove unsafe `new String("...")` optimization

mischnic:
**Description:**
After removing the `preserve_string_call` setting, `ExprSimplifierConfig` is actually empty now. Not sure if that should be removed as well or left as is

**BREAKING CHANGE:**
The API changed did change..

Closes https://github.com/swc-project/swc/issues/3283


kdy1 (dismissed):
LGTM. Thanks!

kdy1:
Can you remove the failing test case?

kdy1 (approved):
Thank you!

PR #3289, opened 2022-01-17
fix(es/resolver): Ignore names of jsx attributes

### LINKED ISSUE #2854: Infinite loop code generated with JSX + ES5 onChange attribute via anonymous function
### Describe the bug

Hi there,

I discovered this bug while working on my big Next.js 12 application using SWC and then reduced it to a very simple example on a fresh new create-next-app with Next.js 12. And I was able to reproduce it on the SWC playground too.

PS: The playground share functionality is not working at the moment.

### Input code

```tsx
function App() {
  return <Form />
}

function Form({ onChange = function () {} }) {
  return (
    <input
      onChange={function () {
        onChange();
      }}
    />
  );
}
```

### Output code

```js
function App() {
    return React.createElement(Form, null);
}
function Form(param) {
    var _onChange = param.onChange, onChange = _onChange === void 0 ? function() {
    } : _onChange;
    return React.createElement("input", {
        onChange: function onChange() {
            onChange();
        }
    });
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": true
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

_SWC playground share functionality not working at the moment_

### Expected behavior

The generated code should not trigger an infinite loop

### Version

1.2.11

### Additional context

Do we need to rename anonymous functions when compiling to ES5?

Here's a live Next.js demo:
- https://swc-maximum-call-stack-error.vercel.app/
- https://github.com/vvo/swc-maximum-call-stack-error

kdy1:
**Description:**

swc_ecma_transforms_base:
 - Ignore names of jsx attributes. (Closes #2854)

**BREAKING CHANGE:**

 - Closes #2854.


kwonoj (approved):

kdy1 (commented):
@kwonoj The `resolver` pass was attaching the same syntax context as the parameter, so `onChange` in 

```js
       <input
            onChange={function () {
            }}
        />
```

got same syntax context as `onChange` in the paramter. And as `es2015::function_name` pass clones `onChange` of `onChange: fuction () {}` (generated by jsx pass), it resulted in a bug.

kwonoj (approved):

PR #3292, opened 2022-01-17
fix(ecma): apply static block first

Austaras:
**Description:**
static block will be transformed into private static property, which many browsers doesn't support


kdy1 (approved):
LGTM. Thanks!

PR #3298, opened 2022-01-17
feat: update contributing docs

dpchamps:
**Description:**

Starting from a fresh pull, it can be difficult to get up and running. The changes I've proposed here represent findings from digging around the discord channel and a few closed issues.

I think it'd also be nice to call out a high-level explanation of what the difference between `svc_v1` and `svc_v2` is, but I don't know enough about the features. Happy to add it here if someone would give some guidence.

**BREAKING CHANGE:**

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,41 +65,59 @@ Steps to get started:
 
 ## Getting your development environment set up
 
-After cloning the project there are a few steps required to get the project running.
+After cloning the project there are a few steps required to get the project running. 
+For running all tests, take the following steps:
 
-1.  Fetch submodules to pull ECMAScript test suites.
+1. Fetch submodules to pull ECMAScript test suites.
 
     ```bash
     git submodule update --init --recursive
     ```
 
-2.  Install js dependencies.
+2. Install js dependencies.
 
     ```bash
     yarn
     ( cd crates/swc_ecma_transforms; yarn )
     ```
 
-3.  Setup some environment variables which is required for tests.
+3. Setup some environment variables which is required for tests.
 
     ```bash
-    export RUST_BACKTRACE=full
+    export RUST_BACKTRACE=0
     export PATH="$PATH:$PWD/crates/swc_ecma_transforms/node_modules/.bin"
     export RUST_MIN_STACK=16777216
     ```
 
-4.  Install deno, if you are going to work on the bundler.
+4. Install deno
 
     See [official install guide of deno](https://deno.land/manual/getting_started/installation) to install it.
 
-5.  Ensure you're using Node.JS >= 16
+5. Add wasm32-wasi target
+
+   `rustup target add wasm32-wasi` 
+
+6. Ensure you're using Node.JS >= 16
 
     Since tests make use of `atob` which was only introduced in node 16.
 
-6.  Run tests
+7. Run tests
 
     ```bash
-    cargo test --all --all-features
+    cargo test --all --no-default-features --features swc_v1
+    ```
+    Or
+    ```bash
+   cargo test --all --no-default-features --features swc_v2 
+   ```
+
+## Running tests per package
+
+While working on specific packages, individual tests can be run by specifying a package to the
+cargo test runner, e.g.
+
+    ```bash
+    cargo test -p swc_ecma_transforms --all-features
     ```
 
 ## Pull requests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,17 +75,17 @@ For running all tests, take the following steps:
     ```
 
 2. Install js dependencies.
+   Ensure [Yarn Package Manager is installed](https://yarnpkg.com/getting-started/install)
 
-    ```bash
-    yarn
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -65,41 +65,59 @@ Steps to get started:
...
 
     ```bash
-    export RUST_BACKTRACE=full
+    export RUST_BACKTRACE=0
     export PATH="$PATH:$PWD/crates/swc_ecma_transforms/node_modules/.bin"
     export RUST_MIN_STACK=16777216
     ```
 
-4.  Install deno, if you are going to work on the bundler.
+4. Install deno
```
dpchamps, line 92: This, and `wasm32-wasi` is required for running all tests successfully

dpchamps, line 92: This, and `wasm32-wasi` is required for running all tests successfully

CONTRIBUTING.md
```diff
@@ -65,41 +65,59 @@ Steps to get started:
...
     yarn
     ( cd crates/swc_ecma_transforms; yarn )
     ```
 
-3.  Setup some environment variables which is required for tests.
+3. Setup some environment variables which is required for tests.
 
     ```bash
-    export RUST_BACKTRACE=full
+    export RUST_BACKTRACE=0
```
kdy1, line 87: This seems wrong?

CONTRIBUTING.md
kdy1: But if you are not going to work on the bundler, you don't need to run bundler tests.

CONTRIBUTING.md
```diff
@@ -65,41 +65,59 @@ Steps to get started:
...
     ```bash
     git submodule update --init --recursive
     ```
 
-2.  Install js dependencies.
+2. Install js dependencies.
 
     ```bash
     yarn
     ( cd crates/swc_ecma_transforms; yarn )
```
kwonoj, line 81: I think this is no longer required.

kdy1: Oh, you are right.

RiESAEX: I remember we need run yarn at root folder to install mocha,
and setup a path to node_modules/.bin

CONTRIBUTING.md
dpchamps: see my description, perhaps theres something else that needs to be configured?

CONTRIBUTING.md
dpchamps: As someone pulling the code for the first time, I usually want to run all tests to make sure everything went smoothly.

I tried to cover this by calling out "to run all tests..." in this subsection header.

Maybe a better way is to have a different subsection that outlines required dependencies per package, steps for running all tests, and a callout for how to run individual package tests?

CONTRIBUTING.md
dpchamps: what should we change this to? no longer needed, or different command. 

Should also call out yarn as a dependency probably.

kdy1: No longer needed, but `yarn` in root directory is required.

CONTRIBUTING.md
```diff
@@ -65,41 +65,59 @@ Steps to get started:
...
     ```
 
-3.  Setup some environment variables which is required for tests.
+3. Setup some environment variables which is required for tests.
 
     ```bash
-    export RUST_BACKTRACE=full
+    export RUST_BACKTRACE=0
     export PATH="$PATH:$PWD/crates/swc_ecma_transforms/node_modules/.bin"
```
RiESAEX, line 88: I think it should be `export PATH="$PATH:$PWD/node_modules/.bin"`

CONTRIBUTING.md
kdy1: We need `RUST_LIB_BACKTRACE=0`

kdy1: We can configure it in the error reporting test, and I think we should.

CONTRIBUTING.md
dpchamps: @kdy1 thoughts?

kdy1: Oh, yeah you are right.

CONTRIBUTING.md
dpchamps: @kdy1 Great catch on this. `RUST_BACKTRACE=0` Was a suggestion I dug up on the discord server.

The error output is deceiving: it looked like it was emitting a backtrace... but actually, the captured error was being passed into `NormalizedOutput` with `Debug` formatting. 

With backtrace enabled, the lint errors were being captured with stack traces enabled. This is an easy fix: https://github.com/swc-project/swc/pull/3298/files#diff-d807373603a279f7b50a670e39c4b65357c79d13622bca50583de75692c3b6baL76

dpchamps:
@RiESAEX , @kdy1 , @kwonoj  thanks for the feedback. I've updated based on your suggestions.

kdy1 (approved):
LGTM, thanks!

PR #3299, opened 2022-01-17
Use updated `siphasher` to avoid duplicate code and undefined behavior

asquared31415:
This PR removes the implementation of siphash from the `swc_common` crate in favor of the [`siphasher`](https://crates.io/crates/siphasher) crate which is already used transitively by dependencies.

This is done to avoid some undefined behavior in the previous implementation where a pointer that only had read access to one byte would be used to read 8 bytes.  It also lets the project use another crate with a more extensive implementation that can get bugfixes on its own.

This PR also indirectly updates the existing transitive dependency which fixes some other undefined behavior.


kwonoj (approved):
Do we have any example code of UB by previous dependencies?

asquared31415:
Every use of the previous siphash implementation that wrote more than a single byte caused UB.  The UB was in the `load_int_le` macro which was used in every code path that wrote to the hasher.  This code was in the previous implementation and the `get_unchecked` produces a pointer that is only valid for the byte at the passed index.
```rs
        ptr::copy_nonoverlapping(
            $buf.get_unchecked($i),
            &mut data as *mut _ as *mut u8,
            mem::size_of::<$int_ty>(),
        );
```

The `siphasher` crate received an update earlier today to remove the `get_unchecked` in favor of `buf.as_ptr().offset($i)` which produces a pointer valid for the whole slice and offsets it.

PR #3300, opened 2022-01-18
feat(swc/common): implement PluginError

kwonoj:
**Description:**

This PR introduces priliminary error handling story from plugin to the host (SWC). Given plugin runs under wasm runtime, we don't have transparent way to just pass through error to the host. PR introduces simple enum to wraps up possible errors which can be serialized / deserialized into, then host catches it up and bubble up as needed.

The list of possible `PluginError` is incomplete. It should be expanded over actual usecase & feedback.


kdy1 (approved):
LGTM. Nice work!

PR #3304, opened 2022-01-18
reface(ecma): add `visit_obj_and_computed` macro

Austaras:
**Description:**

There're lots of code like this in swc
```rust
visit_member_prop() {
   obj.visit_with(self);
   if let MemeberProp::Computed(c) = &e.prop {
      c.visit_with(self)
   }
}
```
this pr add a macro to simplify it

### DIFF
--- a/crates/swc_ecma_transforms_base/src/helpers/mod.rs
+++ b/crates/swc_ecma_transforms_base/src/helpers/mod.rs
@@ -5,9 +5,7 @@ use swc_common::{FileName, FilePathMapping, Mark, SourceMap, DUMMY_SP};
 use swc_ecma_ast::*;
 use swc_ecma_parser::{lexer::Lexer, Parser, StringInput};
 use swc_ecma_utils::{prepend_stmts, quote_ident, quote_str, DropSpan};
-use swc_ecma_visit::{
-    as_folder, noop_visit_mut_type, visit_mut_obj_and_computed, Fold, VisitMut, VisitMutWith,
-};
+use swc_ecma_visit::{as_folder, noop_visit_mut_type, Fold, VisitMut, VisitMutWith};
 
 #[macro_export]
 macro_rules! enable_helper {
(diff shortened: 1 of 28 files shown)


crates/swc_ecma_transforms_base/src/helpers/mod.rs
```diff
@@ -321,17 +323,7 @@ impl VisitMut for Marker {
         i.span = i.span.apply_mark(self.0);
     }
 
-    fn visit_mut_member_prop(&mut self, p: &mut MemberProp) {
```
kdy1, line 324: Previously we don't visit object here.

kdy1 (commented):
Alterantively, we can remove `visit_mut_member_expr` and `visit_mut_super_prop_expr` from visitors without `visit_mut_ident`.

Previously `MemberExpr.prop` will go into `visit_mut_expr`, but for now, it will only have effect if the visitor overrides `visit_mut_ident`.

crates/swc_ecma_transforms_base/src/helpers/mod.rs
Austaras: By that logic we could also remove some `visit_mut_prop_name`

kdy1: Yes, I think you are right.

kdy1: @kwonoj How do you think about this?

crates/swc_ecma_transforms_base/src/helpers/mod.rs
```diff
@@ -321,17 +323,7 @@ impl VisitMut for Marker {
...
-            p.visit_mut_with(self);
-        }
-    }
-
-    fn visit_mut_super_prop(&mut self, p: &mut SuperProp) {
-        if let SuperProp::Computed(p) = p {
-            p.visit_mut_with(self);
-        }
-    }
+    visit_mut_obj_and_computed!();
```
kdy1, line 326: We don't visit `obj` here.

https://github.com/swc-project/swc/blob/b4f268fa23f90becfea42783e1dd3505fc115d0b/crates/swc_ecma_transforms_base/src/helpers/mod.rs#L324-L328

crates/swc_ecma_transforms_optimization/src/simplify/const_propgation.rs
```diff
@@ -97,13 +99,7 @@ impl VisitMut for ConstPropagation<'_> {
...
     }
 
-    fn visit_mut_member_expr(&mut self, e: &mut MemberExpr) {
-        e.obj.visit_mut_with(self);
-
-        if let MemberProp::Computed(c) = &mut e.prop {
-            c.visit_mut_with(self);
-        }
-    }
+    visit_mut_obj_and_computed!();
```
kdy1, line 102: Is this required?

kdy1 (approved):
Thank you!

PR #3313, opened 2022-01-19
fix(es/minifier): Fix logic for checking `arguments`

kdy1:
**Description:**

swc_ecma_minifier:
 - Don't mark `arguments` in nested scope as a inline blocker.


kdy1:
I verified that it's harmless, by looking at each files.

PR #3318, opened 2022-01-19
Fix media query parsing

alexander-akait:
**Description:**

Fixes/improvements:
1. Parse according https://www.w3.org/TR/mediaqueries-5/#typedef-mf-lt
2. Support `range` syntax
3. Support `not` syntax
4. Support nested parens parsing, i.e. `((((min-width: 100px))))`
5. Fix parsing in many edge cases and simplify logic
6. Better error reporting
7. Fix span
8. Refactor many parts and simplify it
9. Fixed `media` in `@import`

After this I will focus on `@support`, so we will supports the latest `@import` spec for future bundling CSS

Used https://github.com/chromium/chromium/blob/main/third_party/blink/renderer/core/css/parser/media_query_parser.cc#L462, so our behavior is almost the same with chrome, but with new features and abilities

**BREAKING CHANGE:**

Yes, new AST, I think because CSS is still experimental we can avoid migration 

No

There are couple todos in code, so still WIP

Notes for future fixes:
- `// TODO uppercase acceptable` I want to implement macros `is_case_insensitive_ident`, so we can use it for case insensitive CSS syntax (will be next PR) in whole CSS code base
- Added `TODO` for better error recovery


crates/swc_css_parser/src/parser/at_rule.rs
```diff
@@ -678,123 +678,448 @@ where
...
+
+        self.input.skip_ws()?;
+
+        let left = self.parse()?;
+
+        self.input.skip_ws()?;
+
+        match cur!(self) {
+            tok!(")") => {
+                eat!(self, ")");
```
kdy1, line 957: `bump!(self)` would be better, as we know current token.

crates/swc_css_parser/src/parser/at_rule.rs
```diff
@@ -678,123 +678,448 @@ where
...
+                    }
+                };
+
+                Ok(MediaFeature::Boolean(MediaFeatureBoolean {
+                    span: span!(self, span.lo),
+                    name,
+                }))
+            }
+            tok!(":") => {
+                eat!(self, ":");
```
kdy1, line 972: Same here.

kdy1 (changes requested):
Mostly LGTM, nice work!
I found some small code style issues.

kdy1 (approved):

PR #3324, opened 2022-01-20
build: fix Windows arm64 build OOM

Brooooooklyn:
OOM disappeared after `LTO` disabled and `codegen-units' increased` https://github.com/Brooooooklyn/swc/actions/runs/1722159297


kdy1 (approved):
Thank you!

PR #3328, opened 2022-01-20
Fix at rules

alexander-akait:
**Description:**

fixes:
- todo
- error recovering for unknown at-rules

**BREAKING CHANGE:**

No

No

### DIFF
--- a/crates/swc_css_parser/src/parser/at_rule.rs
+++ b/crates/swc_css_parser/src/parser/at_rule.rs
@@ -1092,7 +1092,10 @@ where
                         }
                     }
                     _ => {
-                        return Err(Error::new(span, ErrorKind::Expected("'>' or '<' operators")));
+                        return Err(Error::new(
+                            span,
+                            ErrorKind::Expected("'>' or '<' operators"),
+                        ));
                     }
                 };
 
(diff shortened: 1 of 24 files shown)


crates/swc_css_parser/src/parser/at_rule.rs
```diff
@@ -1113,7 +1118,30 @@ where
                     }
                 };
 
-                // TODO validate comparison
+                let is_valid_operator = match left_comparison {
```
kdy1, line 1121: I think `clippy` will be angry at this.

crates/swc_css_parser/tests/recovery/at-rule/layer/block/output.json
```diff
@@ -23,7 +23,223 @@
...
+              },
+              "token": {
+                "WhiteSpace": {
+                  "value": "   "
+                }
+              }
+            }
+          ]
+        }
+      ],
```
alexander-akait, line 242: We lost many tokens, because not reset state

kdy1 (approved):
Code changes look good, but not sure if it will pass CI.

crates/swc_css_parser/src/parser/at_rule.rs
alexander-akait: hm, let's check it, what is command run clippy?

kdy1: You can run `cargo clippy`

alexander-akait: clippy is fine :smile:

PR #3331, opened 2022-01-20
Fix keyframes parsing

alexander-akait:
**Description:**

bugfixe `@keyframes`:
- Introduce new type - `<custom-ident>`, spec https://drafts.csswg.org/css-values-4/#identifier-value
- fix span
- empty `@keyframes` is not allowed

**BREAKING CHANGE:**

Yes, `CustomIdent` should be used not `Ident`

No


kdy1 (approved):
LGTM. Thanks!

PR #3334, opened 2022-01-21
fix: remove jsx file extension

padorang684:
Adds `.jsx` to the list of file extensions that get removed by `NodeImportResolver`


kdy1 (approved):
Thank you!

PR #3336, opened 2022-01-21
Refactor block

alexander-akait:
**Description:**

Refactor from https://github.com/swc-project/swc/pull/2922/, start to union all blocks in `SimpleBlock` https://www.w3.org/TR/css-syntax-3/#consume-simple-block, i.e. all things in `{}`/`()`/`[]` is `SimpleBlock`.

Next things:
- Refactor `@page` at-rule, because we use `Block` here and it is invalid, there are bugs in parsing
- Union `Block` and `SimpleBlock`

**BREAKING CHANGE:**

Yes, we union all in `SimpleBlock`

No

### DIFF
--- a/crates/swc_css_ast/src/style_rule.rs
+++ b/crates/swc_css_ast/src/style_rule.rs
@@ -32,3 +32,10 @@ pub struct Declaration {
     /// The span includes `!`
     pub important: Option<Span>,
 }
+
+#[ast_node("SimpleBlock")]
+pub struct SimpleBlock {
+    pub span: Span,
+    pub name: char,
+    pub value: Vec<Value>,
+}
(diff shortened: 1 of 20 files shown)


crates/swc_css_ast/src/style_rule.rs
```diff
@@ -1,6 +1,15 @@
 use crate::{AtRule, Ident, SelectorList, Tokens, Value};
 use swc_common::{ast_node, Span};
 
+#[ast_node("SimpleBlock")]
+pub struct SimpleBlock {
+    pub span: Span,
+    // TODO Create a simple block with its associated token set to the current input token and with
+    // its value initially set to an empty list.
+    pub name: char,
```
alexander-akait, line 9: TODO will be resolved after union with `Block`

kdy1 (approved):
Thank you!
I think this change will also make implementing transforms easier.

PR #3340, opened 2022-01-22
fix(bundler): support pnpm

### LINKED ISSUE #2945: spack: error when resolving packages with pnpm
### Describe the bug

Running `spack` with [pnpm](https://pnpm.io/) errors because spack is not able to resolve the packages from `node_modules`. It works fine with `npm`.

Error:
```
node:internal/process/promises:246
triggerUncaughtException(err, true /_ fromPromise _/);
^

[Error: load_transformed failed

Caused by:
0: failed to analyze module
1: failed to resolve is-odd from /home/sachinraja/test/spack-pnpm-bug/node_modules/is-even/index.js
2: index not found] {
code: 'GenericFailure'
}
```

### Input code

```typescript
import isEven from "is-even";

isEven(2);
```

### Config

```js
// spack.config.js
const { config } = require("@swc/core/spack");

module.exports = config({
  entry: {
    web: __dirname + "/index.ts",
  },
  output: {
    name: "index.js",
    path: __dirname + "/dist",
  },
});
```

### Reproduction
https://github.com/sachinraja/spack-pnpm-bug

### Expected behavior

spack resolves and inlines the `is-even` package from `node_modules`

### Version

1.2.116

### Additional context

[pnpm uses symlinks](https://pnpm.io/symlinked-node-modules-structure)

RiESAEX:
**Description:**
after swc_ecma_loader find the module path, call canonicalize to make it resolve the symbolic links.

- Closes #2945


kdy1 (commented):
This will slow down greatly, even for users not using `pnpm`.

crates/swc_bundler/src/bundler/mod.rs
```diff
@@ -44,6 +44,9 @@ pub struct Config {
 
     /// Type of emitted module
     pub module: ModuleType,
+
+    /// If it's false, try to find the real path.
+    pub preserve_symlinks: bool,
```
kdy1, line 49: I think this should be `resolve_symlinks`, and should default to false.

kdy1: `pnpm` is special case and most users will use default config.

kdy1 (changes requested):
Thank youi!

crates/swc_bundler/src/bundler/mod.rs
RiESAEX: I agree with it will slow down the speed.
but I'm not sure if we can preserve symlinks by default. 
I think resolving symlinks is the default behavior of [node](https://nodejs.org/api/cli.html#--preserve-symlinks)/[webpack](https://webpack.js.org/configuration/resolve/#resolvesymlinks)/[esbuild](https://esbuild.github.io/api/#preserve-symlinks).

kwonoj: We don't offer this yet, so even other tools are doing by default we have chance to reconsider default behavior.

Also even though this is not intended, if we make this as default this'll be breaking change _if anyone relies on current resolving_. I doubt there are many but technically.

I'd like to follow suggestion to make this not-enabled by default, then consider separately to make it as default.

kdy1: Does this really fix pnpm issue?

I investigated a bit, and I found that

- pnpm uses hard links (https://pnpm.io/faq)
- hard links are real files, and all hard links are treated as file, not link
- canonicalize uses `realpath` on unix (https://doc.rust-lang.org/std/fs/fn.canonicalize.html)

RiESAEX: pnpm use symlink and hard links. hard links it for pnpm's global cache, and the sym link is to reduce the depth of node_modules
https://pnpm.io/blog/2020/05/27/flat-node-modules-is-not-the-only-way
for example in #2945
pnpm link
```
node_modules/is-even -> node_modules/.pnpm/is-even@1.0.0/node_modules/is-even
node_modules/.pnpm/is-even@1.0.0/node_modules/is-odd -> node_modules/.pnpm/is-odd@0.1.2/node_modules/is-odd
node_modules/.pnpm/is-odd@0.1.2/node_modules/is-number -> node_modules/.pnpm/is-number@3.0.0/node_modules/is-number
node_modules/.pnpm/is-number@3.0.0/node_modules/kind-of -> node_modules/.pnpm/kind-of@3.2.2/node_modules/kind-of
```
```
  DEBUG Resolve is-even from Real(
    "/spack-pnpm-bug/index.ts",
) for Node

  DEBUG Resolve is-odd from Real(
    "/spack-pnpm-bug/node_modules/.pnpm/is-even@1.0.0/node_modules/is-even/index.js",
) for Node

  DEBUG Resolve is-number from Real(
    "/spack-pnpm-bug/node_modules/.pnpm/is-odd@0.1.2/node_modules/is-odd/index.js",
) for Node

  DEBUG Resolve kind-of from Real(
    "/spack-pnpm-bug/node_modules/.pnpm/is-number@3.0.0/node_modules/is-number/index.js",
) for Node

  DEBUG Resolve is-buffer from Real(
    "/spack-pnpm-bug/node_modules/.pnpm/kind-of@3.2.2/node_modules/kind-of/index.js",
) for Node

```

kdy1: If then, canonicalize is fine, but shouldn't it be called by resolver?

crates/swc_ecma_loader/src/resolvers/node.rs
```diff
@@ -83,14 +83,23 @@ enum StringOrBool {
...
+        resolve_symlinks: bool,
+    ) -> Self {
+        Self {
+            target_env,
+            alias,
+            resolve_symlinks,
+        }
     }
 
     fn wrap(&self, path: Option<PathBuf>) -> Result<FileName, Error> {
```
kdy1, line 105: Isn't it better to canonicalize here?

kdy1: (Assuming the entry is not an symlink)

kdy1: If we canonicalize at here, `base` on next resolving would use the canonicalized path, and it will reduce duplicated call to `base.canonicalize`.

kdy1 (changes requested):
Thank you!
I think I found a way to reduce duplicated syscalls

kdy1 (approved):
LGTM. Thank you!

Issue #3308, opened 2022-01-19
Support `swc_common::errors::HANDLER` in plugin


kwonoj:
### Describe the bug

Context: https://github.com/swc-project/swc/blob/f5af22ef716db76a39ea22cc88f3f92745fb4aad/crates/swc_common/src/plugin.rs#L65-L90

Previous `abi_stable` based runtime had those support, while new runtime remove those logics. Host should expose those features to plugins as same as before.

### Expected behavior

-

### Version

master TOT

kdy1:
This can be tested using code in https://github.com/swc-project/swc/pull/2616/files and https://github.com/swc-project/plugins/blob/6b6bedf5e33249a3f365ea9435dc6351600c2818/packages/internal-test/src/lib.rs

PR #3346, opened 2022-01-24
refactor(swc/plugin): remove multivalue polyfill

kwonoj:
**Description:**

Previously our plugin does not need more than 1 returned `value` from plugin. Value means its transformed result or either error. However, https://github.com/swc-project/swc/issues/3308 and few other reveals necessary of setting additional value which is not straightforward to use single return value. The way I solved those is using imported fn (host injects fns into guest, with an external state guest can modify). 

Then I came to realize it could be used for all kind of return values including transform result as well removed need of multivalue return.

PR introduces a imported fn `set_transform_result`, revises `plugin_macro` to call it to set its transformed result. Plugin's transform fn still have return value, but it's single `i32` only indicates if transform was successful or not. This is strictly internal changes - plugin author supposed to use `plugin_macro` and do not deal with mutlivalue return by themselves. However, this removes manual polyfill step to create plugin binary so we can make scaffolding script more easily.


crates/swc_plugin_runner/src/lib.rs
```diff
@@ -11,12 +11,49 @@ use swc_common::{
...
+use wasmer::{
+    imports, Array, Exports, Function, Instance, LazyInit, Memory, Module, Store, WasmPtr,
+};
 use wasmer_cache::{Cache, Hash};
-use wasmer_wasi::{get_wasi_version, WasiState};
+use wasmer_wasi::{is_wasi_module, WasiState};
 
 pub mod resolve;
 
+/// Set plugin's transformed result into host's enviroment.
```
kwonoj, line 22: I'll probably write some more texts how does flow actually works.

crates/swc_plugin_runner/src/lib.rs
```diff
@@ -116,10 +163,10 @@ fn load_plugin(plugin_path: &Path, cache: &mut Option<PluginCache>) -> Result<In
                 // Generate an `ImportObject` from wasi_env
                 let mut import_object = wasi_env.import_object(&module)?;
 
-                // Inject `memory` into generated import_object, this is required to allocate
-                // guest (plugin)'s memory space from the host (SWC).
+                // This'll inject few interfaces into plugin's namespace to let plugin
+                // communicate with host
                 let mut env = Exports::new();
-                env.insert("memory", memory);
+                env.insert("__set_transform_result", set_transform_result_fn_decl);
```
kwonoj, line 169: instead of expose memory and host deals with creating instance, now we pass environment to declared fn and let wasmer runtime sets memory accordingly. `set_transform_result` can access guest memory instance when guest calls it.

crates/swc_plugin_runner/tests/integration.rs
```diff
@@ -12,18 +12,11 @@ use swc_ecma_visit::{Visit, VisitWith};
...
-        } else {
-            let mut c = Command::new("sh");
-            c.args(&["-c", "./build.sh"]);
-            c
-        };
-
+        let mut cmd = Command::new("cargo");
         cmd.current_dir(dir);
-        cmd.stderr(Stdio::inherit()).output()?;
+        cmd.args(["build", "--target=wasm32-wasi"])
```
kwonoj, line 17: simplifed!

crates/swc_plugin_macro/src/lib.rs
```diff
@@ -116,11 +129,8 @@ fn handle_func(func: ItemFn) -> TokenStream {
                 return construct_error_ptr(err);
             }
 
-            (
-                0,
-                serialized_result.as_ptr() as _,
-                serialized_result_len.expect("Should be an i32"),
-            )
+            set_transform_result_volatile(serialized_result.as_ptr() as _, serialized_result_len.expect("Should be an i32"));
```
kwonoj, line 132: this is main change in macro side. Instead of return value, it calls host imported fn first to set value. When it's being called, host's fn access env & allocate host side values.

crates/swc_plugin_runner/src/lib.rs
```diff
@@ -153,18 +202,17 @@ struct PluginTransformTracker {
     instance: Instance,
     // Reference to the pointers succesfully allocated which'll be freed by Drop.
     allocated_ptr_vec: Vec<(i32, i32)>,
+    transform_result: Arc<Mutex<Vec<u8>>>,
```
kdy1, line 205: Is there a reason we can't use `Rc<RefCell<Vec<u8>>>`?
I checked the code and seems like it does not pass thread boundary (https://github.com/swc-project/swc/blob/1bb0dcb18f1ad33879f87304d187d5a449b6bc8c/crates/swc_plugin_runner/src/lib.rs#L328-L330)

kwonoj: I also don't believe it doesn't pass thread boundary currently, but choose it for more of future precaution: wasm thread will be available, and we _may_ need to support it in somehow later - if that occurs we may need to revisit impl.

kwonoj: so this was for guest (plugin) side thread boundary access instead of host side.

kdy1: I see, then I think it's fine.

kdy1 (approved):
`Arc<Mutex<T>>>` was the only issue I found. (Resolved)

Thank you!

PR #3351, opened 2022-01-24
Refactor namespace at rule

alexander-akait:
**Description:**

Implement new `Url` type (spec https://www.w3.org/TR/css-values-4/#urls): 
1. Fix bugs with parsing around `url()`/`src()`
2. Supports modifiers
3. Keep original name `Url` token
4. Refactor `at-rules`, now they accept `Url` type according spec
5. Fix parsing `@namespace`
6. Fix some clippy warnings
7. Added `TODO` for future fixes

**BREAKING CHANGE:**

Yes, new type

No

### DIFF
--- a/crates/swc_css_parser/src/lexer/mod.rs
+++ b/crates/swc_css_parser/src/lexer/mod.rs
@@ -549,16 +549,16 @@ where
                                 raw_name: ref mut raw_url_name,
                                 ..
                             } => {
-                                *url_name = name.0.into();
-                                *raw_url_name = name.1.into();
+                                *url_name = name.0;
+                                *raw_url_name = name.1;
                             }
                             Token::BadUrl {
                                 name: ref mut url_name,
                                 raw_name: ref mut raw_url_name,
                                 ..
                             } => {
-                                *url_name = name.0.into();
-                                *raw_url_name = name.1.into();
+                                *url_name = name.0;
+                                *raw_url_name = name.1;
                             }
                             _ => {}
                         }
@@ -1243,7 +1243,7 @@ where
             }
         }
 
-        return Ok((value, raw));
+        Ok((value, raw))
     }
 }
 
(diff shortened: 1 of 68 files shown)


crates/swc_css_parser/src/lexer/mod.rs
```diff
@@ -542,7 +542,29 @@ where
                 }
                 // Otherwise, consume a url token, and return it.
                 _ => {
-                    return self.read_url();
+                    return self.read_url().map(|mut token| {
```
kdy1, line 545: Why does `read_url` return an incomplete token?
I think accepting `name` as a parameter and returning something in a valid state is better

kdy1: No partial-init is one of philosophies of rust.
Is there a drawback of this approach?

kdy1 (commented):
Thank you! I have some questions about the changes.

crates/swc_css_parser/src/lexer/mod.rs
alexander-akait: Yep, initially I do it, but name in one function, but token creation in other, so I tried to keep them independent, but it is not problem and I can refactor it

kdy1: Can you do the refactor? I think it's a better way to go.

alexander-akait: Yes, I will do it in near future :+1:

kdy1: Thank you! I think we should follow the philosophy of the language if we can.

alexander-akait: @kdy1 Done :+1: Found couple TODO with `url`, I will fix them in the next PR

kdy1 (approved):
Thank you!

PR #3356, opened 2022-01-24
Fix page at rule parsing

alexander-akait:
**Description:**

Fix `@page` parsing:
1. `@page` at-rule has selector list (i.e. comma separated syntax) in prelude
2. fix multiple pseudo parsing
3. keep parsing nested `@page` at-rules
4. fix span
5. added `todo` for codegen, we should implement `ListFormat::SemiAtEnd` for minify printing (i.e. the latest declaration can omit `;` if nothing more after `declaration`), currently we have weird logic and don't print `;` in many cases

Notes:
1. We still missing `SimpleBlock` for all `at-rules` (i.e. `@viewport { prop: value }` and we don't have `SimpleBlock` for `{`/`}`), it is bug and I will fix it after we finish union `Block` and `SimpleBlock`, because we need more refactor
2.  We have `TODO: change on `parse_simple_block`` for `KeyframesRule` rule, and we have wrong logic, I will fix it after `1.` above

**BREAKING CHANGE:**

Yes

No


crates/swc_stylis/tests/prefixer.rs
```diff
@@ -458,15 +458,15 @@ fn error_recovery_1() {
...
+            animation: slide 3s ease infinite",
     );
 
     t(
         "animation: slide 3s ease infinite;
             __styled-jsx-placeholder__1
         ",
         "-webkit-animation:slide 3s ease infinite;animation:slide 3s ease \
-         infinite;__styled-jsx-placeholder__1\n        ;",
+         infinite;__styled-jsx-placeholder__1\n        ",
```
alexander-akait, line 469: Want to verify it will not break nothing, don't know context of `// This behavior is wrong, but it's what `stylis@3` does.`, should we print `;` in this case or not, in CSS only latest declaration can omit `;`

Currently we print `;` in all valid cases where it is valid, but we will improve it in future - for `minify: true` we will omit `;` at the end and decrease size of minified CSS

crates/swc_css_parser/src/parser/at_rule.rs
```diff
@@ -136,6 +136,49 @@ where
...
+            | "bottom-center"
+            | "bottom-right"
+            | "bottom-right-corner"
+            | "left-top"
+            | "left-middle"
+            | "left-bottom"
+            | "right-top"
+            | "right-middle"
+            | "right-bottom"
+                if self.ctx.in_page_at_rule == true =>
```
kdy1, line 155: Please remove the comparison with `true`

crates/swc_css_parser/src/parser/at_rule.rs
```diff
@@ -1324,27 +1367,22 @@ where
...
-                    break;
-                }
-            }
-            items
-        };
+        self.ctx.in_page_at_rule = true;
 
         let block = self.parse()?;
 
+        self.ctx.in_page_at_rule = false;
```
kdy1, line 1382: I think we can use `.with_ctx` instead?
`with_ctx` returns a RAII guard which restores the original context on drop.

kdy1:
```rust
let ctx = Ctx {
    in_page_at_rule: true,
    ..self.ctx
};
let block = self.with_ctx(ctx).parse()?;
```

will do the trick.

crates/swc_css_parser/src/parser/at_rule.rs
```diff
@@ -1353,105 +1391,161 @@ where
...
-                eat!(self, ";");
+    fn parse(&mut self) -> PResult<PageSelectorPseudo> {
+        let span = self.input.cur_span()?;
+
+        expect!(self, ":");
 
-                Ok(p)
+        let value = match cur!(self) {
+            Token::Ident { value, .. }
+                if &*value.to_ascii_lowercase() == "left"
```
kdy1, line 1503: This seems ineffcient.
We can do `matches!(&*value.to_ascii_lowercase(), "left" | "right" | "first" | "blank")` instead.

crates/swc_css_parser/src/parser/at_rule.rs
```diff
@@ -1353,105 +1391,161 @@ where
...
-            allow_at_selector: true,
-            ..self.ctx
-        };
-        let prelude = self.with_ctx(ctx).parse_selectors()?;
+    fn parse(&mut self) -> PResult<PageMarginRule> {
+        let span = self.input.cur_span()?;
+
+        expect!(self, "{");
+
+        self.ctx.in_page_at_rule = false;
```
kdy1, line 1534: Same here. We can use RAII guard instead.

crates/swc_stylis/tests/prefixer.rs
kdy1: Yeah, we should emit `;` if it's not last because it will be replaced by style property like `color: red`.

To be exact, `__styled-jsx-placeholder__1` becomes `color: red`

kdy1 (changes requested):
Thank you!

crates/swc_css_parser/src/parser/at_rule.rs
alexander-akait: Yep, I will do it firstly when rebase this PR we already need this in many places

alexander-akait:
@kdy1 For stylish look like we need extra option, ideally stylish should generate `__styled-jsx-placeholder__1;` (i.e. always put `;` at the end, to protect interpolated styles, but I think we can't fix it in `stylish`), I will thing how we can better to fix it

kdy1:
Can you rebase this?

alexander-akait:
Yes, today

alexander-akait:
@kdy1 Ready for review

kdy1 (approved):
LGTM. Thank you!

PR #3359, opened 2022-01-25
feat(swc_plugin): pass hostcontext to plugin for diagnostics emission

kwonoj:
**Description:**

This PR attempts to solve #3308 by exposing simple proxy to `handler.emit` into plugin side.

Plugin's transform fn signature now accepts 3rd param named `HostContext`, which contains a wrapper to serialize `Diagnostics` to call host fn. Host fn calls `handler.emit` once it receives a diagnostics. I've been debating implementation approaches for the plugin side proxy (previously named pluginEmitter) and ended up to have a single interface to pass `Diagnostic` as is, and let plugin choose to construct Diagnostic as needed. This simplifies overall communication between host to guest as we only need to serialize / deserialize `Diagnostic`.

Test case was added to resemble previous example and _looks_ like it's working, but bit unsure if this is proper way to test. Needs some eye probably.

**BREAKING CHANGE:**

- Plugin's transform macro no longer requires to return `Result` from plugin fn. As commented errors should go through diagnostic emission
- Plugin's transform macro fn now passes 3rd param for the host environment context.

In result, plugin fn signature looks like below:

```
#[plugin_module]
fn transform(program: Program, config_str: String, ctx: &HostContext);
```

Note ctx is reference: unlike other values (program / config_str) it is constructed inside of plugin_macro which lives until transform fn lifecycle completes.

- https://github.com/swc-project/swc/issues/3308


tests/rust-plugins/swc_internal_plugin/src/lib.rs
```diff
@@ -36,10 +36,11 @@ impl VisitMut for ConsoleOutputReplacer {
...
 /// if plugin need to handle low-level ptr directly. However, there are important steps
-/// manually need to be performed like transformed results back to host. Refer swc_plugin_macro
+/// manually need to be performed like sending transformed results back to host. Refer swc_plugin_macro
 /// how does it work internally.
 #[plugin_module]
-pub fn process(program: Program, _plugin_config: String) -> Result<Program, PluginError> {
-    let transformed_program = program.fold_with(&mut as_folder(ConsoleOutputReplacer));
-    Ok(transformed_program)
+pub fn process(program: Program, _plugin_config: String, ctx: &HostContext) -> Program {
+    let test_diag = Diagnostic::new(Level::Error, "test_diag_error_from_plugin");
```
kdy1, line 43: What happens if a user use

```rust

HANDLER.with(|handler| {
     handler.struct_span_err(sp, "test").emit();
})
```

?

I think we should call user-defined transform function within a closure with `HANDLER` configured.

kdy1: I think it will be easy since the only required tasks are creating a `Handler` implementation which calls `HostContext.emit` on `Handler::emit`, and configuring `HANDLER` using `HANDLER.set`

kwonoj: I'm not sure if I understood correctly, but `HANDLER` supposed to be unique in host's context, isn't it? Plugin does not have way to access those unique context inside of wasm.

kdy1: No, it's a scoped thread-local variable.
`HANDLER` is configured only in closures provided to `HANDLER.set`.

kdy1: The macro for plugin should configure one, and call the transform function in the closure provided to `HANDLER.set`

kwonoj: Ok, I think I'm quite confused. As you said, HANDLER is scoped_thread_local and since plugin is wasm, it is not possible to compile plugin with wasm-* target if we import HANDLER, since wasm currently doesn't have thread support as well as scoped-tls is not able to target wasm-*.

Maybe I'm missing some obvious? 🤔

kwonoj: hmm, I recall first time I tried to import `HANDLER` there was a build failure but somehow attempting once again it seems at least build. Not sure what's changed.

Back to the original question, would you mind elaborating thoughts around importing whole handler in the plugin side? I think I'm not getting full pictures correctly in here.

kdy1: I experimented, but seems like thread-local requires `target-feature=+atomic`, `target-feature=+mutable-globals`, `target-feature=+bulk-memory` and something more.
So... Using `scoped-tls` with wasm will be quite tricky at the moment.

I stopped the investigation, but I think we need a way to report errors/create private identifiers from the plugin, while allowing testing them.
User plugin will depend on `swc_common/plugin-mode`, but while testing we may need to use normal `swc_common`, not one for `plugin-mode`.


For this PR, I think this is the best we can do at the moment. Thank you!

kdy1 (approved):

PR #3362, opened 2022-01-25
Fix todo in url

alexander-akait:
**Description:**

Resolve `TODO`:
1. Error recovery for invalid escapes
2. Keep `BadUrl` values as is and do not lose whitespaces
3. Tests

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

PR #3363, opened 2022-01-25
refactor: rename `PercentValue` to `Percent`

alexander-akait:
**Description:**

Rename `PercentValue` to `Percent`:
1. CSS spec show explicity `Percent` https://www.w3.org/TR/css-values-4/#percentages
2. I think we don't need `Value` suffix, because only one type for percent values

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
LGTM. Thanks!

PR #3364, opened 2022-01-25
Feat dashed ident

alexander-akait:
**Description:**

Implemented:
1. `<deshaed-ident>` - https://www.w3.org/TR/css-values-4/#dashed-idents
2. Fix span for `Ident` and `DashedIdent` in `at-rules` names
3. Implemented `@color-profile` https://www.w3.org/TR/css-color-5/#at-profile

**BREAKING CHANGE:**

Yes

No


kdy1 (changes requested):
Why do we need a separate dashed identifier?
Can't we just store dash in `Ident`?

alexander-akait:
@kdy1 They are different in semantic and in grammar parsing, `dashed-ident` used for custom properties/custom-media/etc and they don't have grammar, they can have any values, `custom-ident` should be used in property values, i.e. `foo: my-ident` (`my-ident` is `custom-ident`, currently we use `ident` and should fix it), `ident` used only for built-in CSS names except values, we will use different codegen/minification/visitor logic here for different ident.

Also it will allow use to avoid potential unsafe transformations for custom properties, because we will not do it where `dashed-ident` used

kdy1 (approved):
Thank you!

PR #3369, opened 2022-01-26
Feat improve minify output

alexander-akait:
**Description:**

Improve basic and minify output

Notes:
1. We should improve block outputting for basic and minify output, we need to finish union (`Block` and `SimpleBlock`), I will do in the next PR
2. We should introduce `SemiDelimitedExcludeLast` to printng declaration list, it will print `;` at the end of declaration list in basic output and avoid it in minify output, also it will allow to remove `semi_after_property`, we don't need it
3. `increase_indent` and `decrease_indent` are broken, we should fix it too, but better to do it after union `Block` and `SimpleBlock`
4. We should remove `SpacedValues` and keep only `CommaSeparated`, i.e. by default print all using space ` ` and print comma when it is inside `CommaSeparated`, CSS always print all using space, unless otherwise showed
5. There are a lot of improvements to minify output, I will continue it

**BREAKING CHANGE:**

No

No


crates/swc_css_codegen/tests/fixture/at-rules/import/1/output.min.css
```diff
@@ -0,0 +1 @@
+@import'custom.css';@import url("chrome://communicator/skin/");@import url("fineprint.css") print;@import url("bluish.css") speech;@import"common.css" screen;@import url('landscape.css') screen and (orientation:landscape);
```
kdy1, line 1: I tested using w3c css validator, and semes like the space after `)` is not required.

It's fine if it's intentional, I'm asking just to ensure it's not a mistake.

alexander-akait: Do you mean `@import url('landscape.css') screen`? Yes, also space is not required after `@import"common.css"screen;`

I will improve

alexander-akait: @kdy1 Done :+1:

kdy1 (dismissed):
Thank you!

alexander-akait:
I will fix clippy :+1:

PR #3386, opened 2022-01-27
Fix selector parser

alexander-akait:
**Description:**

Fix selector parser + added `bootstrap` to codegen test (almost all things are good), we still have one problem with `/`, to solve it we need refactor `calc` and `/`/`*`/`+`/`-` parsing, I will fix it late, also there is improvement for `@media` output, I will fix it late too

**BREAKING CHANGE:**

No

No


kdy1 (commented):
Thank you!

PR #3390, opened 2022-01-28
fix(ecma): dynamic import with name from another import

### LINKED ISSUE #3246: Wrong variable used in dynamic import and import
### Describe the bug

with a code like `import(variable)`, SWC will output a code that will crash with `ReferenceError: variable is not defined` when `variable` is imported from another module

### Input code

```typescript
import { foo } from 'bar';

import(foo);
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    }
  },
  "isModule": true,
  "module": {
    "type": "commonjs"
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.128&code=H4sIAAAAAAAAA8vMLcgvKlGoVkjLz1eoVUgrys9VUE9KLFK35uLKBMtpAGU0rbkA8UzezSkAAAA%3D&config=H4sIAAAAAAAAAy2KQQrAIAwE%2F7JnX%2BAf%2BgixKShaJYlQkfy9FnramWEXskT4hR5YiD%2BSeWt44EGxBomcusLt205XKEJmDkmOdo5C8MqDHOpvCzr7XsRWa7uzwOwFk8nycWMAAAA%3D

### Expected behavior

When we look at the output of this code:

image

We can see that `foo` is transpiled to `_bar.foo`. So I'd expect it to behave the same way here.

### Actual behavior

But when we look at our code sample:

image

We can see that `require(foo)` is used, when `foo` was never injected in the global scope (`_bar.foo` is)

### Version

1.2.128

(but it was also buggy in all versions I checked, and I went back up to the version 1.2.83, but I couldn't go further in the REPL)

### Additional context

It doesn't happen with `require` instead of `import()`:

image

---

Nor with `require` instead of `import from`:

image

---

It crashes too with default import:

image

---

But it works with `import *`

image

RiESAEX:
**Description:**
```js
import { foo }from 'bar';
import(foo);
```
to 
```js
var _bar = require("bar");
Promise.resolve().then(function() {
    return _interopRequireWildcard(require(_bar.foo)); // not return _interopRequireWildcard(require(foo));
});
```
babel generates like 
```js
Promise.resolve("" + _bar.foo).then(function (s) {
  return _interopRequireWildcard(require(s));
});
```
But I think they work the same.
fix #3246

### DIFF
--- a/crates/swc_ecma_transforms_module/src/util.rs
+++ b/crates/swc_ecma_transforms_module/src/util.rs
@@ -506,13 +506,34 @@ impl Scope {
             Expr::Call(CallExpr {
                 span,
                 callee: Callee::Import(_),
-                args,
+                mut args,
                 ..
             }) if !folder.config().ignore_dynamic
                 // TODO: import assertion
                 && args.len() == 1 =>
             {
-                folder.make_dynamic_import(span, args)
+                let expr = match *(args.pop().unwrap().expr) {
+                    Expr::Ident(ident) => match Self::fold_ident(folder, ident) {
+                        Ok(mut expr) => {
+                            if let Expr::Member(member) = &mut expr {
+                                if let Expr::Ident(ident) = member.obj.as_mut() {
+                                    member.obj = Box::new(Expr::Paren(ParenExpr {
+                                        expr: Box::new(Expr::Seq(SeqExpr {
+                                            span,
+                                            exprs: vec![Box::new(ident.take().into())],
+                                        })),
+                                        span,
+                                    }))
+                                }
+                            };
+                            expr
+                        }
+                        Err(ident) => Expr::Ident(ident),
+                    },
+                    expr => expr,
+                };
+
+                folder.make_dynamic_import(span, vec![expr.as_arg()])
             }
 
             Expr::Call(CallExpr {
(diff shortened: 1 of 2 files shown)


crates/swc_ecma_transforms_module/src/util.rs
```diff
@@ -506,13 +506,34 @@ impl Scope {
...
-                folder.make_dynamic_import(span, args)
+                let expr = match *(args.pop().unwrap().expr) {
+                    Expr::Ident(ident) => match Self::fold_ident(folder, ident) {
+                        Ok(mut expr) => {
+                            if let Expr::Member(member) = &mut expr {
+                                if let Expr::Ident(ident) = member.obj.as_mut() {
+                                    member.obj = Box::new(Expr::Paren(ParenExpr {
+                                        expr: Box::new(Expr::Seq(SeqExpr {
+                                            span,
+                                            exprs: vec![Box::new(ident.take().into())],
```
kdy1, line 523: Is `SeqExpr` required?
Seems like it has only one expression, and due to how codegen works, the output will be exact the same as the inner expression.

kdy1 (approved):
Thank you!

PR #3394, opened 2022-01-28
feat(es/lint): no-alert implementation

ArturAralin:
**Description:**

no-alert rule implementations.

Note: current implementation is not supports calls on "this" like `this.alert('msg')`

**BREAKING CHANGE:**
No


crates/swc/tests/errors/lints/no-alert/alert-ident-hoisting/output.swc-stderr
```diff
@@ -0,0 +1,6 @@
+error: Unexpected alert
+ 
+  |
+1 | alert();
```
kdy1, line 4: I think this should not be an error, as user is calling user-defined `alert`.
If we want to report an error, it should be usage before declarationn.

kdy1 (changes requested):

crates/swc/tests/errors/lints/no-alert/alert-ident-hoisting/output.swc-stderr
ArturAralin: you're mean is should be caught by another rule?

kdy1: I meant this should not be treated as a call to the global `alert`.

ArturAralin: Ok. I'll fix it

ArturAralin: @kdy1 while I investigating how to solve your comment I did found different syntax context resolving behaviour in tests in swc and swc_ecma_lint crates. I managed to solve different behaviour by adding to `crates/swc/src/config/mod.rs` following command 

```
program.visit_mut_with(&mut resolver_with_mark(top_level_mark));
```

It is correct solution or it should be fixed with another way?

kdy1: It's wrong and I think we need a feature gate for non-critical errors.

ArturAralin: So I need help with understanding how to fix problem with different syntax contexts resolving 😃

kdy1: Ah, are you using collecting bindings on creating?

kdy1: I think you can do it in `lint_module` instead?
If it does not work, it needs some thinking about desgin...

kdy1: Oh, I think your fix is correct way to go (applying resolver), although we need to remove the next call to the resolver.

ArturAralin: Okay. Happy to hear it

ArturAralin: hm... My fix solves my problem, but it fail some tests 

```
    tests_tests__fixture__issue_1160__input
    tests_tests__fixture__issue_1345__input
    tests_tests__fixture__issue_1362__case1__input
    tests_tests__fixture__issue_1421__case1__input
    tests_tests__fixture__issue_1869__decorator__input
    tests_tests__fixture__issue_846__case1__input
 ```

Comment (ArturAralin) [reply crates/swc/tests/errors/lints/no-alert/alert-ident-hoisting/output.swc-stderr:4]: I can commit my fixes or provide terminal output

Comment (kdy1) [reply crates/swc/tests/errors/lints/no-alert/alert-ident-hoisting/output.swc-stderr:4]: I'll create another PR to fix the wrong design and you can rebase after it's merged.

Comment (kdy1) [reply crates/swc/tests/errors/lints/no-alert/alert-ident-hoisting/output.swc-stderr:4]: I made https://github.com/swc-project/swc/pull/3400

Comment (ArturAralin) [reply crates/swc/tests/errors/lints/no-alert/alert-ident-hoisting/output.swc-stderr:4]: Thank you

**Comment by ArturAralin** [comment]
@kdy1 what about feature gate, I'd make it later

**Comment by kdy1** [comment]
It's to prevent binary size regression of `nexg-swc` and other users of `swc` crate.
I think `no-alert` and `no-console` should be removed by the feature gate by default, and enabled only for `@swc/core` or `@swc/wasm*`

**Comment by ArturAralin** [comment]
Ok. So if I'd add feature gate in https://github.com/swc-project/swc/pull/3399 it's okay? Just don't want update two other opened PRs

**Comment by kdy1** [comment]
Okay, I think it would be fine.

**Comment by ArturAralin** [comment]
@kdy1 good. Then I'll be wait for you review this PR

File: crates/swc_ecma_lints/src/rules/no_alert.rs [hunk 2]
```diff
@@ -0,0 +1,153 @@
...
+        top_level_declared_vars: AHashSet<Id>,
+        top_level_ctxt: SyntaxContext,
+        es_version: EsVersion,
+    ) -> Self {
+        Self {
+            expected_reaction,
+            top_level_ctxt,
+            top_level_declared_vars,
+            pass_call_on_global_this: es_version < EsVersion::Es2020,
+            fn_names: FN_NAMES.iter().map(|n| *n).collect(),
```
kdy1, line 57: Why is this required?

crates/swc_ecma_lints/src/rules/no_alert.rs
```diff
@@ -0,0 +1,153 @@
...
+            obj_names: OBJ_NAMES.iter().map(|n| *n).collect(),
```
kdy1, line 58: Same here, we can just use `static` item from rule.

crates/swc_ecma_lints/src/rules/no_alert.rs
```diff
@@ -0,0 +1,153 @@
...
+    noop_visit_type!();
+
+    fn visit_call_expr(&mut self, call_expr: &CallExpr) {
+        self.call_expr_span = Some(call_expr.span);
+
+        call_expr.visit_children_with(self);
+    }
+
+    fn visit_opt_chain_expr(&mut self, opt_chain: &OptChainExpr) {
+        opt_chain.visit_children_with(self);
```
kdy1, line 133: This impl seems redundant.
This is default implementation of `visit_opt_chain_expr`.

crates/swc_ecma_lints/src/rules/no_alert.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        // TODO: handle call alert on "this"
+    }
+}
+
+impl Visit for NoAlert {
+    noop_visit_type!();
+
+    fn visit_call_expr(&mut self, call_expr: &CallExpr) {
+        self.call_expr_span = Some(call_expr.span);
```
kdy1, line 127: I think this will be broken for nested calls.

kdy1: e.g.

```js
alert(() => alert("foo"))
```

crates/swc_ecma_lints/src/rules/no_alert.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        call_expr.visit_children_with(self);
+    }
+
+    fn visit_opt_chain_expr(&mut self, opt_chain: &OptChainExpr) {
+        opt_chain.visit_children_with(self);
+    }
+
+    fn visit_ident(&mut self, id: &Ident) {
+        if id.span.ctxt != self.top_level_ctxt {
```
kdy1, line 137: I think handling `Expr::Ident` from `visit_expr` is generally better, as it's much more efficient and much likely to work properly.

kdy1:
```js
(() => {
    const obj = {
        alert,
    };
})()
```

It's in a call, so `call_expr_span` is `Some` and `visit_ident` is called also for shorthand properties.

kdy1: Also,

```js
(() => {
    alert = () => {};
})()
```

it will have the top-level context and `visit_ident` is also called for patterns.

kdy1: One more,

```js
(() => {
    console.log(alert);
})()
```

is this a target of the lint rule?

kdy1 (changes requested):

crates/swc_ecma_lints/src/rules/no_alert.rs
ArturAralin: You're mean that I can use static hashset?

crates/swc_ecma_lints/src/rules/no_alert.rs
ArturAralin: Yes, you're right

crates/swc_ecma_lints/src/rules/no_alert.rs
kdy1: No, you don't need hastset for this.

crates/swc_ecma_lints/src/rules/no_alert.rs
```diff
@@ -0,0 +1,158 @@
...
+                    return;
+                }
+
+                if self.call_expr_span.is_some() {
+                    self.check_names(None, &*id.sym);
+                }
+            }
+            Expr::Call(call_expr) => {
+                if self.call_expr_span.is_none() {
+                    self.call_expr_span = Some(call_expr.span);
```
kdy1, line 139: I think this will error for

```js
foo(() => alert("foo"));
```

as `call_expr_span` points `foo` call while vaildating `alert`.

crates/swc_ecma_lints/src/rules/mod.rs
```diff
@@ -6,10 +6,25 @@ use swc_ecma_visit::{noop_fold_type, Fold};
 mod const_assign;
 mod duplicate_bindings;
 mod duplicate_exports;
+pub mod no_alert;
```
kdy1, line 9: Can you remove `pub`?

crates/swc_ecma_lints/src/rules/mod.rs
```diff
@@ -6,10 +6,25 @@ use swc_ecma_visit::{noop_fold_type, Fold};
...
 pub mod no_console;
```
kdy1, line 10: Also, I missed this one on previous review. Can you remove `pub` here, too?

crates/swc_ecma_lints/src/rules/no_alert.rs
```diff
@@ -0,0 +1,180 @@
...
+            self.emit_report(call_span, fn_name);
+        }
+    }
+}
+
+impl Visit for NoAlert {
+    noop_visit_type!();
+
+    fn visit_call_expr(&mut self, call_expr: &CallExpr) {
+        let mut call_collector = CallCollector {
```
kdy1, line 97: Using a nested visitor without a stopper method is bad for performance.

kdy1: Also, we don't need two visitor here, and visitor is quite big in final binary.

kdy1: (That's why methods for collecting bindings uses single visitor)

kdy1 (changes requested):

crates/swc_ecma_lints/src/rules/no_alert.rs
ArturAralin: Hmm.. Seems like I need advice how visit indents/members only on callee

ArturAralin: I rewrote implementation without nested visitor, but it looks a bit awkward by my mind

crates/swc_ecma_lints/src/rules/mod.rs
ArturAralin: changed in on pub(crate). Can't remove `pub` because config structs placed into modules

ArturAralin:
@kdy1 also I did feature gate

crates/swc_ecma_lints/src/config.rs
```diff
@@ -1,44 +1,61 @@
...
-#[derive(Debug, Clone, Default, Serialize, Deserialize)]
-#[non_exhaustive]
-#[serde(rename_all = "camelCase")]
-pub struct LintConfig {
-    #[serde(default)]
-    pub no_console: RuleConfig<NoConsoleConfig>,
-
-    #[serde(default)]
-    pub no_debugger: RuleConfig<()>,
+#[cfg(not(feature = "non_critical_lints"))]
```
kdy1, line 53: This will make these lints removed when the featrure is enabled.
Can you revert commits related to feature gates?

ArturAralin: A bit didn't understand you. Why lints will be removed if features is enabled? Here is `cfg(not(feature...`

kdy1: Because `#[cfg(not(feature = "non_critical_lints"))]` means this code (under `cfg`) should be built only if `not(feature = "non_critical_lints")`.
It means if we enable `non_critical_lints `, it's `not(true)`, so it's removed.

ArturAralin: then if we're not enable non_critical_lints then it's not(false) -> true.

if it is then it's expected.

here is I include all non critical lints if feature flag provided https://github.com/swc-project/swc/pull/3394/files/aade1992b77528596ce47ab63fa45ecf2e4374e3#diff-804fa350bc6183d7f8eb97b783ba6090005850a1f7a3c176f62b572871ff558dR1

and here is not https://github.com/swc-project/swc/pull/3394/files/aade1992b77528596ce47ab63fa45ecf2e4374e3#diff-804fa350bc6183d7f8eb97b783ba6090005850a1f7a3c176f62b572871ff558dR53

ArturAralin: Here I provide empty LintConfig struct if `non_critical_lints` feature flag isn't provided

ArturAralin: Possibly here is should be used another name

https://github.com/swc-project/swc/pull/3394/files/aade1992b77528596ce47ab63fa45ecf2e4374e3#diff-804fa350bc6183d7f8eb97b783ba6090005850a1f7a3c176f62b572871ff558dR54

crates/swc/Cargo.toml
```diff
@@ -50,7 +50,7 @@ swc_common = {version = "0.17.0", path = "../swc_common", features = ["sourcemap
 swc_ecma_ast = {version = "0.65.3", path = "../swc_ecma_ast"}
 swc_ecma_codegen = {version = "0.89.1", path = "../swc_ecma_codegen"}
 swc_ecma_ext_transforms = {version = "0.50.0", path = "../swc_ecma_ext_transforms"}
-swc_ecma_lints = {version = "0.11.1", path = "../swc_ecma_lints"}
+swc_ecma_lints = {version = "0.11.1", path = "../swc_ecma_lints", features = ["non_critical_lints"]}
```
kdy1, line 53: This should not be enabled by the swc crate.
`node` and `wasm` should depend on `swc_ecma_lints` and enable this feature.

crates/swc_ecma_lints/src/config.rs
```diff
@@ -1,44 +1,61 @@
...
-
-#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
-#[serde(rename_all = "lowercase")]
-pub enum LintRuleReaction {
-    Off,
-    Warning,
-    Error,
-}
+#[cfg(feature = "non_critical_lints")]
+#[path = ""]
```
kdy1, line 2: I think this can be removed?

crates/swc_ecma_lints/src/config.rs
```diff
@@ -1,44 +1,61 @@
...
+    use crate::rules::non_critical_lints::no_console::NoConsoleConfig;
+    use serde::{Deserialize, Serialize};
+    use std::fmt::Debug;
 
-impl Default for LintRuleReaction {
-    fn default() -> Self {
-        Self::Off
+    #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+    #[serde(rename_all = "lowercase")]
+    pub enum LintRuleReaction {
```
kdy1, line 10: I think this type should exist regardless of features.
The only things we need to feature-gate is impl itself (module) and the function `all`.

kdy1 (changes requested):

crates/swc/Cargo.toml
ArturAralin: Then tests in `swc` crate will fail

kdy1: Then we can add `swc_ecma_lints ` with features enabled to `[devDependencies]`, as we use 2021 edition.

ArturAralin: Sounds good

crates/swc/Cargo.toml
```diff
@@ -91,6 +91,7 @@ rayon = "1"
 swc_node_base = {version = "0.5.0", path = "../swc_node_base"}
 testing = {version = "0.18.0", path = "../testing"}
 walkdir = "2"
+swc_ecma_lints = {path = "../swc_ecma_lints", features = ["non_critical_lints"]}
```
kdy1, line 94: Please add version here, otherwise `cargo` will deny publishing.

kdy1 (changes requested):
Thank you! Code changes look good to me.

crates/swc/Cargo.toml
ArturAralin: oops. missed it

kdy1 (approved):
Thank you!

PR #3397, opened 2022-01-28
Refactor space and comma delimiter

alexander-akait:
**Description:**

Implement `Delimiter`, because CSS can have any values in `declaration` we implement special `Delimiter` struct, it allow to fix bug in `bootstrap` (will be next after merging) parsing with `/`.

Also we don't need anymore `CommaValues` and `SpaceValues`, and now output valid AST for `Function`.

Note - there are some improvements for minify mode (PR will be late)

**BREAKING CHANGE:**

Yes

No


crates/swc_css_codegen/tests/fixture/function/1/input.css
```diff
@@ -0,0 +1,6 @@
+div {
+    color: rgb(255, 255, 255);
+    color: rgb(255, 255, 255,.5);
+    color: rgb(255 255 255);
+    color: rgb(255 255 255 / .5);
```
alexander-akait, line 5: Here example `function` without `,`

crates/swc_css_codegen/tests/fixture/function/1/output.min.css
```diff
@@ -0,0 +1 @@
+div{color:rgb(255,255,255);color:rgb(255,255,255,.5);color:rgb(255 255 255);color:rgb(255 255 255 / .5)}
```
alexander-akait, line 1: Potential optimization, will be fixed in the next PR (after we merge it)

kdy1 (approved):
Thank you!

PR #3399, opened 2022-01-28
feat(es/lint): implement prefer-regex-literals

ArturAralin:
**Description:**

feat(es/lint): implement prefer-regex-literals

**BREAKING CHANGE:**
No


crates/swc_ecma_lints/src/rules/prefer_regex_literals.rs
```diff
@@ -0,0 +1,269 @@
...
+    "Regular expression literal is unnecessarily wrapped within a 'RegExp' constructor";
+const UNEXPECTED_REDUNDANT_REG_EXP_WITH_FLAGS_MESSAGE: &str =
+    "Use regular expression literal with flags instead of the 'RegExp' constructor";
+
+const MAX_VALID_ARGS_COUNT: usize = 2;
+
+const GLOBAL_THIS: &str = "globalThis";
+
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct PreferRegexLiteralsConfig {
```
kdy1, line 23: I think `#[serde(rename_all = "camelCase"]` is better for consistency.

crates/swc_ecma_lints/src/rules/prefer_regex_literals.rs
```diff
@@ -0,0 +1,269 @@
...
+        self.call_span = None;
+        self.first_arg = None;
+        self.second_arg = None;
+    }
+
+    fn set_state(&mut self, call_span: Span, args: &[ExprOrSpread]) {
+        self.call_span = Some(call_span);
+
+        if let Some(ExprOrSpread { expr, .. }) = args.get(0) {
+            let value = match expr.as_ref() {
```
kdy1, line 103: I think this logic can be extracted as a function?
Seems like same logic is used twice.

kdy1 (changes requested):
Thanks!

kdy1 (approved):
Thanks!

PR #3401, opened 2022-01-29
refactor: dimension

alexander-akait:
**Description:**

Refactor `UnitValue`:
1. Use `Dimension` (spec https://www.w3.org/TR/css-values-4/#dimensions)
2. Use `Ident` in `unit`, we don't need extra AST type, because (from spec):

> When written literally, a dimension is a number immediately followed by a unit identifier, which is an identifier. 

Yes

No


kdy1 (approved):
Thanks!

kdy1 (changes requested):
Can you rebase?

alexander-akait:
WIP

alexander-akait:
@kdy1 Done :+1:

kdy1 (approved):
Thanks!

PR #3402, opened 2022-01-29
Fix solidus delimiter

alexander-akait:
**Description:**

Bug fix, parse `/` as delimiter, now we can parse bootstrap without any problems

**BREAKING CHANGE:**

No

No


crates/swc_css_parser/src/parser/value/mod.rs
```diff
@@ -33,13 +33,27 @@ where
...
+            println!("{:?}", self.input.cur());
+
             hi = v.span().hi;
             values.push(v);
             state = self.input.state();
 
             if !eat!(self, " ")
                 && !is_one_of!(
-                    self, ",", "function", "ident", "percent", "str", "#", "url", "[", "{", "("
+                    self,
```
alexander-akait, line 42: This will be refactored in near future, we should move recovery mode inside `Declaration` parsing, not here, so it will be simplified

kdy1 (approved):
Thank you!
Seems like I can merge this without rebasing.

PR #3403, opened 2022-01-29
bugfix and tests

alexander-akait:
**Description:**

Uncomment some tests + fix error location for url modifiers

**BREAKING CHANGE:**

No

No


crates/swc_css_parser/tests/fixture/function/url/input.css
```diff
@@ -45,4 +45,5 @@ a {
     background: url(image.png\D799);
     background: url(image.png\E000);
     background: url(image.png\10FFFF);
+    background: url(18);
```
alexander-akait, line 48: This is valid `url`

kdy1 (approved):
Thanks!

PR #3405, opened 2022-01-29
test: more tests for codegen

alexander-akait:
**Description:**

Just test, found many places for optimizations (want to track them in future to avoid regressions)

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

PR #3410, opened 2022-01-30
Refactor code

alexander-akait:
**Description:**

Refactor code:
1. Rename `property` to `name` due spec https://www.w3.org/TR/css-syntax-3/#consume-declaration

> Consume the next input token. Create a new declaration with its name set to the value of the current input token and its value initially set to an empty list.

In spec declarations use `name` to store property name

2. Remove `AtTextValue` from `Value` because it is invalid, in this case we use recovery mode

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

PR #3411, opened 2022-01-30
refactor: color

alexander-akait:
**Description:**

Refactor colors:
1. Rename to `HexColor` due https://www.w3.org/TR/css-color-4/#typedef-color

There are more subtypes of color type, they will be implemented late

Also it allows to minify/transform colors faster and easy

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

PR #3415, opened 2022-01-31
Fix math functions

alexander-akait:
**Description:**

Fix:
1. Apply math logic only inside math function, in other places they are invalid
2. Add tests with `calc` (there are a lot of bugs, we should rewrite logic according spec, I will do it late, currently regressions tests)

**BREAKING CHANGE:**

No

No

### DIFF
--- a/crates/swc_css_parser/src/parser/value/mod.rs
+++ b/crates/swc_css_parser/src/parser/value/mod.rs
@@ -255,8 +255,6 @@ where
 
             tok!("str") => return Ok(Value::Str(self.parse()?)),
 
-            tok!("num") => return self.parse_numeric_value(),
-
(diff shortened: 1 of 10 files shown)


crates/swc_css_parser/src/parser/value/mod.rs
```diff
@@ -885,21 +877,43 @@ where
...
+                | "atan"
+                | "atan2"
+                | "pow"
+                | "sqrt"
+                | "hypot"
+                | "log"
+                | "exp"
+                | "abs"
+                | "sign"
+        );
```
alexander-akait, line 909: https://www.w3.org/TR/css-values-4/#math

kdy1 (approved):
Thanks!

PR #3423, opened 2022-02-01
Feat improve number minification

alexander-akait:
**Description:**

Improve `Number` and `String` minify output

**BREAKING CHANGE:**

No

No


crates/swc_css_codegen/src/lib.rs
```diff
@@ -938,15 +942,64 @@ where
...
+        }
     }
 
     #[emitter]
     fn emit_ration(&mut self, n: &Ratio) -> Result {
         emit!(self, n.left);
-        punct!(self, "/");
 
         if let Some(right) = &n.right {
+            punct!(self, "/");
```
alexander-akait, line 1002: Bug fix

crates/swc_css_codegen/src/lib.rs
```diff
@@ -764,7 +764,11 @@ where
 
     #[emitter]
     fn emit_str(&mut self, n: &Str) -> Result {
-        self.wr.write_raw(Some(n.span), &n.raw)?;
+        if self.config.minify {
+            self.wr.write_str(Some(n.span), &n.value)?;
+        } else {
+            self.wr.write_raw(Some(n.span), &n.raw)?;
+        }
```
alexander-akait, line 771: That is why we need `raw` and `value`, also `raw` can be used by plugins and in other tools without re parsing original source, because parsing some things in CSS is not easy and require to implement lexer

kdy1: Oh, I see. Great to see actual usecases of these fields.

kdy1 (approved):
Thank you!

PR #3425, opened 2022-02-02
feat: better number compression

alexander-akait:
**Description:**

Bugfix and feature - compress numbers better, also fix bug with trailing `0` and compression, example to reproduce `1000020000`

WIP on tests, also want to try to improve some numbers with `0x`

**BREAKING CHANGE:**

No

No


crates/swc_ecma_codegen/src/lib.rs
```diff
@@ -530,19 +530,65 @@ where
                 self.wr.write_str_lit(num.span, "-")?;
             }
             self.wr.write_str_lit(num.span, "Infinity")?;
-        } else if num.value.is_sign_negative() && num.value == 0.0 {
```
alexander-akait, line 533: We can't have negative number here, because `-100` is `UnaryExpression + NumericLiteral`

kdy1 (approved):
Thanks!

PR #3427, opened 2022-02-02
feat(ecma/parser): add allowSuperOutsideMethod option

### LINKED ISSUE #3230: use super where is not a method
### Describe the bug

normally should receive a error `'super' is only allowed in object methods and classes.` or `Uncaught SyntaxError: 'super' keyword unexpected here`
babel has a parser option for this `allowSuperOutsideMethod `

### Input code

```typescript
super.x
class a{
	a = function a(){
      super.x
    }
}
let obj = {
	a: function a(){
    	super.x
    }
}
```

### Config

_No response_

### Playground link

_No response_

### Expected behavior

should throw error

### Actual behavior

_No response_

### Version

1.2.128

### Additional context

_No response_

RiESAEX:
**Description:**
throw an error when use `super` outside of class and object methods

**BREAKING CHANGE:**
esConfig:
add `allowSuperOutsideMethod`: By default, super use is not allowed outside of class and object methods. Set this to true to accept such code.

for TS, I found that tsc will output the result and throw an error at the same time, so I make this always true
close #3230


crates/swc_ecma_parser/tests/span/js/super/obj1.js.spans
```diff
@@ -0,0 +1,6 @@
+error: Invalid access to super
```
kdy1, line 1: Oh... I never thought that span test can be used for error reporting 🤣

kdy1 (approved):

PR #3430, opened 2022-02-02
test: verify AST is not broken after minify

alexander-akait:
**Description:**

Test we don't break AST after minify

**BREAKING CHANGE:**

No

No


crates/swc_css_codegen/src/lib.rs
```diff
@@ -1022,7 +1022,7 @@ where
     fn emit_bin_value(&mut self, n: &BinValue) -> Result {
         emit!(self, n.left);
 
-        let need_space = matches!(n.op, BinOp::Add | BinOp::Mul);
+        let need_space = matches!(n.op, BinOp::Add | BinOp::Sub);
```
alexander-akait, line 1025: Found bug

alexander-akait: Fixed

kdy1 (approved):
Thanks!

PR #3431, opened 2022-02-02
fix(ecma/transform/compat): handle private field in nested class

### LINKED ISSUE #3368: Wrongly transformed private field in nested class
### Describe the bug

JS has a lexcial scope of private field, so any function and class inside a class could use the private field outside of it

### Input code

_No response_

### Config

_No response_

### Playground link

https://play.swc.rs/?version=1.2.133&code=H4sIAAAAAAAAAz2NQQrDMAwE73rFQg6xL%2FpBD%2BlPVGO3BWEF26WU4r83MWl0EGh20AaVWrHgS8AkuGBOKc3bkcycHxgosb1KRhju9YDATcqp7BMsV9PIanfXHs%2FKk%2Fgj7PTfnTpRjm8szvMo4bVYs%2FZZI28fOYiqG4L%2FAQfD0aWdAAAA&config=H4sIAAAAAAAAA0WMWwoCMQxFtyL57p%2BoMHtwEaFmtNIXSUYsQ%2FduWgb8u49z7w5v8bDsUJGFeChpWfELC5BPKJ5DVXCGWbRiFOoOFPlJOhC5WBdLEfq3qTy2SONKW6VJXWHkIYe1HZyDIPcDVN7MU%2F7MzfwWm%2FkXl0Sn2xl6%2FwFhK12uqAAAAA%3D%3D

### Expected behavior

Code could run

### Actual behavior

throw `a is not found`

### Version

1.2.133

### Additional context

_No response_

Austaras:
closes #3368


Austaras:
This is actually another kind of variable shadowing, hopefully this could work as a base for a new ident rewrite system

crates/swc/tests/tsc.rs
```diff
@@ -20,6 +20,10 @@ use testing::{NormalizedOutput, Tester};
         "propertyAccessChain\\.3.ts",
         "objectRestNegative.ts",
         "objectRestPropertyMustBeLast.ts",
+        // shall panic because not finding private field
```
kdy1, line 23: I think we should not panic.
Instead of panic, we can point the usage of `private`s using `swc_common::error::HANDLER`.

kdy1 (changes requested):
Mostly LGTM.
I'm just curious if it's intentional or a mistake.

crates/swc_ecma_transforms_compat/src/es2022/class_properties/private_field.rs
```diff
@@ -1,21 +1,68 @@
...
 use swc_ecma_utils::{alias_ident_for, alias_if_required, prepend, quote_ident, ExprFactory};
 use swc_ecma_visit::{noop_visit_mut_type, VisitMut, VisitMutWith};
 
-pub(super) struct BrandCheckHandler<'a> {
-    /// Mark for the private `WeakSet` variable.
+#[derive(Clone)]
+pub(super) struct Private {
     pub mark: Mark,
+    pub class_name: Ident,
+    pub ident: AHashMap<JsWord, PrivateKind>,
```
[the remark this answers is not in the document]
Austaras: Somehow `ClassProperties` get a `clone`, which seems like not required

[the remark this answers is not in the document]
kdy1: Oh, I see.
I also think it's not required.

crates/swc/tests/tsc.rs
kdy1: I think panic is fine for now.
Which do you prefer? Merge it as is, or improve the message?

Austaras: I'm not quite sure how to use HANDLER, could you take a look at my new commit?

kdy1: Using it is a bit hard as you should not use `panic`.
You have to return some dummy value like `None` and stop by using e.g. `return`

Austaras: Ok, let's do it later then

kdy1 (approved):
Thank you!

PR #3441, opened 2022-02-03
Refactor keyframes

alexander-akait:
**Description:**

Refactor `@keyframes name {}` logic:
1. Keyframes at-rules can contain only `KeyframeBlock` and can't contain any `AtRule`
2. Validate ident, because only `from`/`to`/`<percantage>` allowed
3. Rename `KeyframeBlock` fields according spec
4. This PR is required to complete union `SimpleBlock` and `Block` in one Node, so I separate PRs, hope we will finish it on this week
 
**BREAKING CHANGE:**

Yes

No


crates/swc_css_parser/src/parser/util.rs
```diff
@@ -14,29 +12,6 @@ impl<I> Parser<I>
...
-            let res = self.parse()?;
-            items.push(res);
-
-            if !ParseDelmited::eat_delimiter(self)? {
-                break;
-            }
-        }
-
-        Ok(items)
-    }
```
alexander-akait, line 38: In future (when we refactor parser using grammars in right way - we are working on it) we will have the same util, but it will be refactor to using `ComponentValue`, currently we don't need it

kdy1 (approved):
Thanks!

PR #3444, opened 2022-02-03
Fix lexer

alexander-akait:
**Description:**

Bugfix for URL and UrlRawValue:
1. Fix lexer and do not lose whitespaces in token
2. Print valid output for `Url` and `BadUrl` tokens
3. Fix span
4. Improve AST for `UrlValueRaw`, so no whitespaces will be in `value`, it will allow to get real URL without hacks (for example for bundling or linting purposes)
5. Improve minified output, currently we output whitespaces at the end of `UrlRawValue`, but it is unnecessary

**BREAKING CHANGE:**

yes

No


crates/swc_css_parser/src/lexer/mod.rs
```diff
@@ -1034,11 +1046,11 @@ where
...
-    // Consume a name
-    // This section describes how to consume a name from a stream of code points. It
-    // returns a string containing the largest name that can be formed from adjacent
-    // code points in the stream, starting from the first.
-    fn read_name(&mut self) -> LexResult<(JsWord, JsWord)> {
+    // Consume an ident sequence
+    // This section describes how to consume an ident sequence from a stream of code
+    // points. It returns a string containing the largest name that can be formed
+    // from adjacent code points in the stream, starting from the first.
+    fn read_ident_sequence(&mut self) -> LexResult<(JsWord, JsWord)> {
```
alexander-akait, line 1053: Just rename according spec, because it was renamed recently

kdy1 (commented):
LGTM. Thanks!

kdy1 (approved):

kdy1 (approved):

PR #3445, opened 2022-02-03
feat: error recovery for function

alexander-akait:
**Description:**

Feat - error recovery for functions, we will apply the same logic for defalcation too after merged, it will allow for us match grammar (in future) and simplify logic

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

PR #3448, opened 2022-02-04
test(es): Update tsc test suite

kdy1:


tido64:
@kdy1: It looks like 1.2.137 is now pulling in `ts-node`. Is that intentional? From the looks of this change, it's only needed for tests? Can we move it to `devDependencies`?

tido64:
> Seems like I made a mistake. I'll remove it and publish a new version tomorrow

No problem. Thanks for the quick response ❤️

PR #3450, opened 2022-02-04
fix(ecma/modules): dynamic import with expr from another import

### LINKED ISSUE #3438: ReferenceError in imported variables in dynamic imports
### Describe the bug

When an imported variable is used in a dynamic import, SWC doesn't use the proper variable, resulting in a `ReferenceError: <variable> is not defined`

### Input code

```typescript
import { foo } from 'bar';

const baz = 'baz'

import(`world/${foo(baz)}.js`);
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    }
  },
  "isModule": true,
  "module": {
    "type": "commonjs"
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.136&code=H4sIAAAAAAAAA8vMLcgvKlGoVkjLz1eoVUgrys9VUE9KLFK35uJKzs8rLlFISqxSsAWJValzcWWClWsklOcX5aToq1QDdWkAZTRr9bKKEzStuQAuSOPOTwAAAA%3D%3D&config=H4sIAAAAAAAAAy2KQQrAIAwE%2F7JnX%2BAf%2BgixKShaJYlQkfy9FnramWEXskT4hR5YiD%2BSeWt44EGxBomcusLt205XKEJmDkmOdo5C8MqDHOpvCzr7XsRWa7uzwOwFk8nycWMAAAA%3D

### Expected behavior

If I don't want it in an `import()`, here what SWC outputs:

image

we can see that it doesn't use `foo` but `_bar.foo` (or more specifically `(0, _bar).foo(baz)`). So I'd assume that something similar would happen with dynamic imports

### Actual behavior

Here is the code outputted by SWC:

image

We can see the
```js
var _bar = require("bar");
// ...
foo(baz)
```

instead of `_bar.foo(baz)` (or what swc would do: `(0, _bar).foo(baz)`)

### Version

1.2.136

### Additional context

_No response_

RiESAEX:
**Description:**
fold expr first then generate dynamic import
fix #3438


kdy1 (approved):
LGTM. Thanks!

PR #3455, opened 2022-02-04
Test more selectors

alexander-akait:
**Description:**

test selectors from https://github.com/postcss/postcss-selector-parser (union some tests, because they test the same)

Found two bugs:
1. span on type selector for ident (will fix in the next PR)
2. need improve AST for attribute matcher (will fix in the next PR)

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

PR #3456, opened 2022-02-04
feat(es/lint): implement no-use-before-define rule

ArturAralin:
**Description:**

no-use-before-define rule implementation

**BREAKING CHANGE:**
No


crates/swc/tests/errors/lints/no-use-before-define/check-let-and-const/.swcrc
```diff
@@ -0,0 +1,7 @@
+{
+    "jsc": {
+        "lints": {
+            "noUseBeforeDefine": ["off"]
```
kdy1, line 4: Why does this lint work when it's `off`?

ArturAralin: let and const checks regardless of enabled or disabled rule. I did it on purpose. Of course I can keep empty lints config, but I decide to disable rule visibly

crates/swc/tests/errors/lints/no-alert/alert-ident-hoisting/.swcrc
```diff
@@ -1,7 +0,0 @@
-{
```
kdy1, line 1: Why is this removed?

ArturAralin: Oh... It's should be reversed back. I was remove this test because it covers by no-use-before-define

crates/swc_ecma_lints/src/rules/no_use_before_define.rs
```diff
@@ -0,0 +1,233 @@
...
+
+#[derive(Debug, Default)]
+struct NoUseBeforeDefine {
+    expected_reaction: LintRuleReaction,
+    // means variables defined with "var" keyword
+    check_vars: bool,
+    check_functions: bool,
+    check_classes: bool,
+
+    scoped_indents: AHashMap<Span, AHashSet<Id>>,
```
kdy1, line 44: Wow... I love the idea.

crates/swc_ecma_lints/src/rules/no_use_before_define.rs
```diff
@@ -0,0 +1,233 @@
...
+                    }
+                });
+            }
+            Pat::Object(ObjectPat { props, .. }) => {
+                props.iter().for_each(|prop| match prop {
+                    ObjectPatProp::Assign(AssignPatProp { key, .. }) => {
+                        self.check_ident(es6_var_check, key);
+                    }
+                    ObjectPatProp::KeyValue(KeyValuePatProp { value, .. }) => {
+                        self.check_pat(es6_var_check, value.as_ref());
```
kdy1, line 168: I think we need `self.check_ident(es6_var_check, key);` here, as `key` of `KeyValuePatProp` creates a binding, just like `Pat::Ident`

crates/swc_ecma_lints/src/rules/no_use_before_define.rs
```diff
@@ -0,0 +1,233 @@
...
+
+        self.scoped_indents.remove(&block.span);
+        self.scope.pop();
+    }
+
+    fn visit_var_decl(&mut self, var_decl: &VarDecl) {
+        let es6_var_check = !matches!(&var_decl.kind, VarDeclKind::Var);
+
+        var_decl.decls.iter().for_each(|declarator| {
+            if let Some(init) = &declarator.init {
```
kdy1, line 208: I think`declarator.init.visit_with(self)` is better, without `if let Some()`.

kdy1: Description: Behavior is identical, as

 - `visit_opt_expr` is not defined. `if let Some` part is done by default impl of `visit_opt_expr`.
 - `visit_expr` is not defined. `visit_children_with(self)` is done by default impl of `visit_expr`.

crates/swc_ecma_lints/src/rules/no_use_before_define.rs
```diff
@@ -0,0 +1,233 @@
...
+                if !self.check_vars {
+                    return;
+                }
+            }
+
+            self.check_pat(es6_var_check, &declarator.name);
+        });
+    }
+
+    fn visit_fn_decl(&mut self, function: &FnDecl) {
```
kdy1, line 222: I think we should recures into the parameters and the body of functions

crates/swc_ecma_lints/src/rules/no_use_before_define.rs
```diff
@@ -0,0 +1,233 @@
...
+        if self.check_functions {
+            self.check_ident(false, &function.ident);
+        }
+    }
+
+    fn visit_class_decl(&mut self, class: &ClassDecl) {
```
kdy1, line 228: I think we should recures into the contents of the classes

kdy1 (changes requested):
Thank you!

ArturAralin:
I'll resolve your issues after I back home

crates/swc_ecma_lints/src/rules/no_use_before_define.rs
ArturAralin:
```
function f() {
  const {
    x: { y }
  } = { x: { y: 10, z: 20 } }

  console.log(x); // ReferenceError: x is not defined
}

f();
```

kdy1 (approved):
LGTM. Thanks!

Issue #2117, opened 2021-08-20
Class transformer not working with swc


DanielRamosAcosta:
**Describe the bug**

When applying the `plainToClass` function to a class defined with `class-validator`, it returns the class with no attributes

**Input code**

[Here is a repo with a test reproducing the error](https://github.com/DanielRamosAcosta/swc-class-transformer-bug)

```ts
const usersOptionsInput = plainToClass(GetUsersOptionsInput, {
  query: "Query",
  pagination: { skip: 0, limit: 20 },
});

expect(usersOptionsInput.query).toEqual("Query"); // error
```

**Config**

You can also find the config at the [repo reproducing the error](https://github.com/DanielRamosAcosta/swc-class-transformer-bug/blob/main/.swcrc) 

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "decorators": true
    },
    "transform": {
      "legacyDecorator": true,
      "decoratorMetadata": true
    },
    "target": "es2018"
  },
  "module": {
    "type": "commonjs",
    "noInterop": true
  }
}
```

**Expected behavior**
I would expect the instance of the class to have all the defined attributes.

**Version**
The version of @swc/core: 1.2.80

**Additional context**

Just one more time, [I created this repo in order to reproduce the error](https://github.com/DanielRamosAcosta/swc-class-transformer-bug)

This error is related to #1362

unlight:
Issue on `class-transformer` side https://github.com/typestack/class-transformer/issues/796 I suppose invalid, and it should be fixed in swc.

tonivj5:
@unlight if class-validator works perfect with tsc and fail with swc. I think it's a problem of swc handling decorators...

kdy1:
@tonivj5 Did you try importing `reflect-metadata`?

tonivj5:
@kdy1 yep! It doesn't work.

This code does not work:
```ts
import "reflect-metadata";
import { IsOptional, IsString } from "class-validator";

export class GetUsersOptionsInput {
  @IsString()
  @IsOptional()
  public query?: string;
}
```

This code does work:
```ts
import "reflect-metadata";
import { IsOptional, IsString } from "class-validator";

export class GetUsersOptionsInput {
  @IsString()
  @IsOptional()
  public query?: string = undefined;
}
```

I did a deeper research here https://github.com/swc-project/swc/issues/2117#issuecomment-906681274

kshahani:
[...]

Issue #2127, opened 2021-08-21
Setting a getter with Object.defineProperty not working within decorator


InsOpDe:
**Describe the bug**
If I set a getter for a property with `Object.defineProperty` within a decorator it returns `undefined` instead of the variable

**Input code**
```typescript
/* works */
test("without decorator", () => {
    const returnValue = "asdasd"
    class TestClass {
        public testProperty: Date;
    }

    Object.defineProperty(TestClass.prototype, "testProperty", {
        get: function () { return returnValue },
        enumerable: true,
        configurable: true,
    });
    const instance = new TestClass()
    expect(instance.testProperty).toBe(returnValue)
})

/* fails */
test("with decorator", () => {
    const returnValue = "asdasd"
    class TestClass2 {
        @deco public testProperty: Date;
    }
    function deco(target: any, key: string) {
        Object.defineProperty(target.constructor.prototype, key, {
            get: function () { return returnValue },
            enumerable: true,
            configurable: true,
        });
    }
    const instance = new TestClass2()
    expect(instance.testProperty).toBe(returnValue)
})

/* works */
test("with decorator without getter", () => {
    const returnValue = "asdasd"
    class TestClass2 {
        @deco public testProperty: Date;
    }
    function deco(target: any, key: string) {
        Object.defineProperty(target.constructor.prototype, key, {
            value: returnValue,
            enumerable: true,
            configurable: true,
        });
    }
    const instance = new TestClass2()
    expect(instance.testProperty).toBe(returnValue)
})
```

**Config**

```typescript
/* jest.config.js*/
module.exports = {
    testEnvironment: 'node',
    transform: {
        // "^.+\\.tsx?$": "ts-jest"
		// "^.+\\.(tsx?)$": ["@swc-node/jest", {
		// 	dynamicImport: true,
		// 	emitDecoratorMetadata: true,
		// 	esModuleInterop: true,
		// 	experimentalDecorators: true,
        //     keepClassNames: true,
		// 	// module: "commonjs",
		// 	// target: "es5"
		// }]
		"^.+\\.(tsx?)$": ["@swc/jest", {
            "jsc": {
                "parser": {
                    "syntax": "typescript",
                    "decorators": true,
                    "dynamicImport": false
                },
                "transform": {
                    "legacyDecorator": true,
                    "decoratorMetadata": true
                },
                "target": "es2016",
                "loose": false,
                "externalHelpers": false,
[...]

Issue #960, opened 2020-08-13
Property decorator break class inherent behavior


Brooooooklyn:
```ts
class Base {
  constructor() {
    this.action = new Subject()
  }
}

class Child extends Base {
  @DefineAction() action: Observable<void>
 
  callApi() {
    console.log(this.action) // undefined
  }
}
```

Seems caused by `defineProperty` in constructor in Child Class

kdy1:
How should this be hsndled? 
Does babel works for the case?

Brooooooklyn:
TypeScript works fine，action should be Subject here

Brooooooklyn:
https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAEQKbBmVBBa8wAUAlIgN6IBOqUIFSoksCBA+gIYA0iLARlxCXJUadRBEQBfSQChpEADZsAzksQAhZajLTEYhEqgUQ0OBWLbduqAAsYSgHRs8CRAF5EARh0yJshctUAYVt5ABNEVAAPKFQwUNUNJS1SbwABNAwsXCZCEiccgC5EMBAAWx5UCh1vCDZ5eWwABxhzFMs9MCU4eVR7eTgAcwIbO0dnMBIAeknEcFD0TFRQ719fOX0oMTdi1AB3RGCYMOI-e1r6ppaiIA

Click run and output **1** in Logs tab

kdy1:
Thanks for the example!

PR #3459, opened 2022-02-05
(fix) Decorators on class properties missing

williamtetlow:
**Description:**

**BREAKING CHANGE:**

https://github.com/swc-project/swc/issues/2117

### DIFF
--- a/crates/swc_ecma_transforms/tests/decorators.rs
+++ b/crates/swc_ecma_transforms/tests/decorators.rs
@@ -4467,42 +4467,42 @@ export let Product = _class = _dec6(((_class = class Product extends Timestamped
     configurable: true,
     enumerable: true,
     writable: true,
-    initializer: void 0,
+    initializer: null,
 }), _descriptor1 = _applyDecoratedDescriptor(_class.prototype, 'price', [
     _dec1
 ], {
     configurable: true,
     enumerable: true,
     writable: true,
-    initializer: void 0,
+    initializer: null,
 }), _descriptor2 = _applyDecoratedDescriptor(_class.prototype, 'type', [
     _dec2
 ], {
     configurable: true,
     enumerable: true,
     writable: true,
-    initializer: void 0,
+    initializer: null,
 }), _descriptor3 = _applyDecoratedDescriptor(_class.prototype, 'productEntityId', [
     _dec3
 ], {
     configurable: true,
     enumerable: true,
     writable: true,
-    initializer: void 0,
+    initializer: null,
 }), _descriptor4 = _applyDecoratedDescriptor(_class.prototype, 'orders', [
     _dec4
 ], {
     configurable: true,
     enumerable: true,
     writable: true,
-    initializer: void 0,
+    initializer: null,
 }), _descriptor5 = _applyDecoratedDescriptor(_class.prototype, 'discounts', [
     _dec5
 ], {
     configurable: true,
     enumerable: true,
     writable: true,
-    initializer: void 0
+    initializer: null
 }), _class)) || _class;
 "
 );
@@ -4532,7 +4532,7 @@ export class Product extends TimestampedEntity {
         configurable: true,
         enumerable: true,
         writable: true,
-        initializer: void 0
+        initializer: null
     }), _class)) || _class;
     "
 );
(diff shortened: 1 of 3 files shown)


kdy1:
If I remeber correctly, tgis is wrong

williamtetlow:
@kdy1 I had a look into this issue and how to fix it while keeping fix for #2127 

My commit isn't pulling into PR for some reason (will take a look) but here it is https://github.com/williamtetlow/swc/commit/396f753518e757177545bc6a505e56de96832904

For #2127 I compiled the same code with SWC, Babel, esbuild and TSC.

SWC - supported
Babel - not supported
TSC (esnext) - not supported
TSC (<esnext) - supported
esbuild - not supported

Looks like only SWC and TSC (lower than esnext) supports this at the moment all the others return `undefined` for the property defined in the decorator.

I moved you original fix to `_initializerDefineProperty`.

If there is no `initializer` for the property and the property already has a value that `!= void 0` then it will set it to the current value. Otherwise, set it to `void 0`.

I read the spec for decorators to see if there was a mention of how this should be handled but there wasn't anything.

williamtetlow:
I've pushed my local test bench for this issue which shows how each tool handles both issues https://github.com/williamtetlow/2117-2127-swc-testbench


Screenshot 2022-02-06 at 11 57 58

kdy1:
Oh... Then I think we should follow `esnext` behavior of tsc.

williamtetlow:
It looks like it's related to [`useDefineForClassFields` ](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#the-usedefineforclassfields-flag-and-the-declare-property-modifier).

williamtetlow:
It looks like it's related to [`useDefineForClassFields`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html%23the-usedefineforclassfields-flag-and-the-declare-property-modifier) which is for migrating to new TC39 standard version of class fields.

This is turned on by default for `esnext`.

The issue in https://github.com/swc-project/swc/issues/2127 is that the field is defined on `prototype.constructor` but by defining the field on the class too the original is hidden.

```js
((_class = class TestClass2 {
    constructor() {
      // defining here hides prototype.constructor.testProperty
      _initializerDefineProperty(this, "testProperty", _descriptor, this);
    }
  })
```

> To help mitigate the second issue, you can either add an explicit initializer or add a declare modifier to indicate that a property should have no emit.

Going forward tsc will define class fields by default (like swc is doing) but you can override this and not emit by adding `declare` modifier to a property.

```ts
class TestClass2 {
  @deco public declare testProperty: Date;
}
```

I've added this logic to swc in https://github.com/swc-project/swc/pull/3459/commits/813e3eb76b2835da28b6ac35a0e42a1c78e75f16#diff-8f03d343376a656cd03aec743c8cb8b60d7c1a4de29f48d940be0795c0931235R676

williamtetlow:
I'm not sure there is a way to have opt out logic for Javascript?

kdy1:
> Should we add setting or are we ok with making https://github.com/swc-project/swc/issues/2127 and https://github.com/swc-project/swc/issues/960 not work anymore when compiling Ecmascript to better replicate Typescript behaviour?

I think it's fine, and it's the right way

kdy1:
Is this PR ready?

williamtetlow:
@kdy1 yep it's ready 😄

crates/swc_ecma_transforms_base/src/helpers/_apply_decorated_descriptor.js
```diff
@@ -16,14 +16,6 @@ function _applyDecoratedDescriptor(target, property, decorators, descriptor, con
     desc.value = desc.initializer ? desc.initializer.call(context) : void 0;
     desc.initializer = undefined;
   }
-
-  var own = Object.getOwnPropertyDescriptor(target, property);
```
kdy1, line 20: Please apply the same things to helpers in https://github.com/swc-project/swc/tree/02c401830d49632e5fbea7280ca703d27c3917d9/packages/swc-helpers

kdy1: Oh, I see. 🤣

kdy1 (approved):
Thanks!

aboqasem:
When could this fix land in Next.js?

kdy1:
@aboqasem I'll make a PR today.

aboqasem:
@kdy1 Thank you for your effort!

JCMais:
Did this land on Next.js?

PR #3460, opened 2022-02-05
Refactor selector

alexander-akait:
**Description:**

Big selector refactor:
1. Better AST for attribute selector
2. Better AST for namespaces
3. Parse `||` combinator
4. Fix span in some cases
5. Use `parse<t>` instead functions
6. Simplify code in many places

Note - we still have couple todos, I will fix them late

**BREAKING CHANGE:**

yes

No


kdy1 (approved):
Thanks!

PR #3461, opened 2022-02-06
fix: respect errors from at rules

alexander-akait:
**Description:**

Bug `fix` https://github.com/vercel/next.js/issues/33946, some error messages look misleading, but let's fix them late

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

PR #3462, opened 2022-02-06
fix(es/minifier): Fix `codemirror`

kdy1:
**Description:**

 - https://github.com/vercel/next.js/discussions/30237?sort=new#discussioncomment-2114870

### DIFF
--- a/crates/swc_ecma_minifier/tests/compress/fixture/next/feedback-2/codemirror/input.js
+++ b/crates/swc_ecma_minifier/tests/compress/fixture/next/feedback-2/codemirror/input.js
@@ -1,0 +1,9849 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
+// Distributed under an MIT license: https://codemirror.net/LICENSE
+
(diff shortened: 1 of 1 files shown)


crates/swc_ecma_minifier/src/analyzer/mod.rs
```diff
@@ -468,12 +468,12 @@ where
 
         if let Expr::Ident(i) = e {
             if cfg!(feature = "debug") {
-                tracing::debug!(
-                    "Usage: `{}``; update = {:?}, assign_lhs = {:?} ",
-                    i,
-                    self.ctx.in_update_arg,
-                    self.ctx.in_assign_lhs
-                );
+                // tracing::debug!(
```
kwonoj, line 471: is this intended?

kwonoj (approved):

crates/swc_ecma_minifier/src/analyzer/mod.rs
kdy1: Yes, because it was too verbose.
I couldn't find the required log because of it.

PR #3463, opened 2022-02-07
fix(es/minifier): Fix `mapbox`

kdy1:
**Description:**

swc_ecma_minifier:
 - `mangler`: Handle the identifiers of class expressions correctly.

image

As shown above, name mangler assigns `i` to a class expression and uses it again inside the class, as the name of a parameter.
This results in an exception from `new i`.

---

This is a fixed version.

image

(I copy-pasted it to js beutifier)

 - https://github.com/vercel/next.js/discussions/30237?sort=new#discussioncomment-2109335


crates/swc_ecma_minifier/tests/terser/compress/harmony/class_name_can_be_preserved/output.js
```diff
@@ -1,4 +1,6 @@
 function x() {
-    (class Baz {});
-    class Foo {}
+    (class a {
```
kdy1, line 2: This change is due to we didn't implement the option of terser to keep class names.

kdy1:
I'll merge it as code changes are quite trivial

PR #3465, opened 2022-02-07
refactor: remove unused code

alexander-akait:
**Description:**

Refactor code - remove `AtSelector`, no grammar for `@` in selector spec, we use it before for parsing nested at-rules in `@page`, but now we refactor code and don't need it anymore

**BREAKING CHANGE:**

Yes

No


alexander-akait:
@kdy1 Looks like we forget to update latest commit when merge rules

kdy1 (approved):
Thanks!
I missed it

PR #3469, opened 2022-02-07
feat: parse `@counter-style`

alexander-akait:
**Description:**

feat - parse `@counter-style` at-rule https://developer.mozilla.org/en-US/docs/Web/CSS/@counter-style

**BREAKING CHANGE:**

Yes, new types

No

### DIFF
--- a/crates/swc_css_codegen/src/lib.rs
+++ b/crates/swc_css_codegen/src/lib.rs
@@ -86,6 +86,7 @@ where
             AtRule::Viewport(n) => emit!(self, n),
             AtRule::Document(n) => emit!(self, n),
             AtRule::ColorProfile(n) => emit!(self, n),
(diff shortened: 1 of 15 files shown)


crates/swc_css_codegen/src/lib.rs
```diff
@@ -834,7 +835,33 @@ where
         emit!(self, n.name);
         formatting_space!(self);
         punct!(self, "{");
-        self.emit_list(&n.block, ListFormat::NotDelimited)?;
```
alexander-akait, line 837: Fixed bug

alexander-akait, line 837: Fixed bug

kdy1 (approved):
Thanks.
I didn't know that this rule exists 🤣

PR #3473, opened 2022-02-07
respect inline config when swcrc is false

RiESAEX:
**Description:**
1. add a rc to auto detection jsx
2. only try to find and merge `.swcrc` when `Options.swcrc = true`
3. merge `TsConfig.dts` and `TsConfig.no_early_errors`

**BREAKING CHANGE:**
since the second poiint change the behavior of `process_js_with_custom_pass`, it may break a lot of things, like rust api, js api and cli. 
I took a brief look at cli, and it seemed use `swcrc=true` by default, so I guess it is fine

#3374

### DIFF
--- a/crates/swc/tests/projects.rs
+++ b/crates/swc/tests/projects.rs
@@ -898,3 +898,81 @@ fn opt_source_file_name_1() {
 
     assert!(map.contains("entry-foo"));
 }
+
+#[test]
+fn auto_detect_jsx() {
+    testing::run_test2(false, |cm, handler| {
+        let c = swc::Compiler::new(cm.clone());
+
+        let fm = cm
+            .load_file(Path::new("tests/projects/jsx/input.jsx"))
+            .expect("failed to load file");
+
+        c.process_js_file(
+            fm,
+            &handler,
+            &Options {
+                swcrc: true,
+                ..Default::default()
+            },
+        )
+        .expect("failed to process file");
+        Ok(())
+    })
+    .unwrap()
+}
+
+#[test]
+fn should_not_merge() {
+    testing::run_test2(false, |cm, handler| {
+        let c = swc::Compiler::new(cm.clone());
+
+        let fm = cm
+            .load_file(Path::new("tests/projects/merge/1/input.tsx"))
+            .expect("failed to load file");
+
+        c.process_js_file(
+            fm,
+            &handler,
+            &Options {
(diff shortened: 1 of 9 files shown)


kdy1 (changes requested):
Can you add some tests?

crates/swc/tests/projects.rs
```diff
@@ -898,3 +898,81 @@ fn opt_source_file_name_1() {
...
+
+        c.process_js_file(
+            fm,
+            &handler,
+            &Options {
+                swcrc: true,
+                config: Config {
+                    jsc: JscConfig {
+                        syntax: Some(Syntax::Typescript(TsConfig {
+                            tsx: false,
```
kdy1, line 964: I think the test is wrong?
`tsx` from `.swcrc` will override this `tsx: false`, and it's current behavior.

kdy1: I think the test should fail, so using a js api test system can be a solution.

kdy1 (changes requested):
Thanks!

crates/swc/tests/projects.rs
RiESAEX: when `swcrc: true`,it keeps current behaviror.

kdy1: Does the current code ignore inline config when swcrc does not exists?
I don't think it's the case as there are some js api tests.

RiESAEX: if swcrc does not exists it become false since `#[serde(default)]`,
them it will only use inline config and for field not exist in string, serde use `default::Default`.
the test of `dynamic_tsx.mjs` pass because we also adjust at it here, should I remove it?
https://github.com/swc-project/swc/blob/d065cb15bf4c98acb28ee8263487a7c04085626c/crates/node/src/transform.rs#L46

kdy1: I don't think so. If an extension is `.tsx`, I think `tsx` should be enabled.
But I'm not sure about other options.

kdy1: Hmm... Is there a reason `.tsx` is used for non-tsx files?

kdy1: If I remember correctly, we added automatic tsx because otherwise, we need to duplicate the whole config in `.swcrc`. (With different regex)

Issue #3424, opened 2022-02-02
Minification - Missing `reserved` mangle option


bogy0:
### Describe the feature

Do not mangle property names listed in the `reserved` array as you can see in [Terser's documentation](https://terser.org/docs/api-reference.html#mangle-options).

```
{
  "jsc": {
    "minify": {
      "mangle": { 
         reserved: ['firstLongName']
      }
    }
  }
}
```

### Additional context

In our codebase we have some reserved function names which later in our build process we try to match to post-process it. Without the `reserved` option we are not able to use SWC to minify our code.

We would be grateful for this feature or a workaround suggestion.

kdy1:
Fixed by https://github.com/swc-project/swc/pull/3476

PR #3476, opened 2022-02-07
(feat): Add `reserved` mangle option to minifier

williamtetlow:
**Description:**
Adds the `reserved` mangle option to the minifier config.

```js
{
  "jsc": {
    "minify": {
      "mangle": { 
         reserved: ['firstLongName']
      }
    }
  }
}
```

**BREAKING CHANGE:**

https://github.com/swc-project/swc/issues/3424


crates/swc_ecma_minifier/src/pass/mangle_names/preserver.rs
```diff
@@ -104,7 +116,7 @@ impl Visit for Preserver {
         n.visit_children_with(self);
 
         if let Pat::Ident(i) = n {
-            if self.should_preserve {
+            if self.should_preserve || self.is_reserved(&i.id) {
```
williamtetlow, line 119: Question: Do we think reserved identifiers should apply to private identifiers too?

crates/swc_ecma_minifier/src/option/mod.rs
```diff
@@ -60,6 +60,9 @@ pub struct MangleOptions {
 
     #[serde(default, alias = "safari10")]
     pub safari10: bool,
+
+    #[serde(default, alias = "reserved")]
+    pub reserved: Vec<String>,
```
kdy1, line 65: I think it's better to use `JsWord` here.

kdy1: You can read about it at https://rustdoc.swc.rs/swc_atoms/

crates/swc_ecma_minifier/src/pass/mangle_names/preserver.rs
```diff
@@ -24,6 +24,12 @@ pub(super) struct Preserver {
     in_top_level: bool,
 }
 
+impl Preserver {
+    fn is_reserved(&self, ident: &Ident) -> bool {
+        self.options.reserved.contains(&ident.sym.to_string())
```
kdy1, line 29: `&ident.sym.to_string()` needlessly allocates and drops the memory right away.
This will be fixed if you change type to the `JsWord`

crates/swc_ecma_minifier/src/pass/mangle_names/preserver.rs
kdy1: I don't think so, but I prefer following behavior of terser.

crates/swc_ecma_minifier/src/pass/mangle_names/private_name.rs
```diff
@@ -5,11 +5,15 @@ use swc_ecma_ast::*;
 use swc_ecma_utils::{ident::IdentLike, Id};
 use swc_ecma_visit::{as_folder, noop_visit_mut_type, Fold, VisitMut, VisitMutWith};
 
-pub fn private_name_mangler(keep_private_props: bool) -> impl Fold + VisitMut {
+pub fn private_name_mangler(
+    keep_private_props: bool,
+    reserved: Vec<String>,
```
kdy1, line 10: `JsWorrd` would be better.

crates/swc_ecma_minifier/src/pass/mangle_names/private_name.rs
```diff
@@ -18,6 +22,8 @@ struct PrivateNameMangler {
     private_n: usize,
 
     renamed_private: AHashMap<Id, JsWord>,
+
+    reserved: Vec<String>,
```
kdy1, line 26: Same here.

kdy1 (changes requested):
Thanks

crates/swc_ecma_minifier/src/pass/mangle_names/preserver.rs
williamtetlow: I checked the behaviour of terser and `reserved` and `properties.reserved` do not apply to private props so removed this logic

crates/swc_ecma_minifier/src/option/mod.rs
williamtetlow: Thanks for the link! 

Updated to use `JsWord`. I also updated the existing `properties.reserved` to use it too https://github.com/swc-project/swc/pull/3476/files#diff-b40213014998d084ab9d7c55540e0646fb34d56f60e8d400edf83457c78bb5c8R72

crates/swc_ecma_minifier/src/pass/mangle_names/preserver.rs
williamtetlow: resolved in c12701c9208b636d660ef23458d9aaf87706d2bc

crates/swc_ecma_minifier/src/pass/mangle_names/private_name.rs
williamtetlow: resolved in c12701c9208b636d660ef23458d9aaf87706d2bc

crates/swc_ecma_minifier/src/pass/mangle_names/private_name.rs
williamtetlow: resolved in c12701c9208b636d660ef23458d9aaf87706d2bc

kdy1 (approved):
Thank you!

PR #3477, opened 2022-02-07
feat: implement all `<dimension>` types

alexander-akait:
**Description:**

Refactor `<dimension>`, now we have all CSS types - https://www.w3.org/TR/css-values-4/#lengths and https://www.w3.org/TR/css-values-4/#other-units

We will use grammar parsing and for better minification - `1000ms` == `0.1s`, I will start to implement first transformations in `swc_css_minifier` when we merge it

**BREAKING CHANGE:**

Yes

No


crates/swc_css_parser/src/parser/value/mod.rs
```diff
@@ -637,7 +637,43 @@ where
         let span = self.input.cur_span()?;
 
         if !is!(self, Dimension) {
-            return Err(Error::new(span, ErrorKind::Expected("Dimension")));
+            return Err(Error::new(span, ErrorKind::Expected("dimension token")));
+        }
+
+        match cur!(self) {
+            Token::Dimension { unit, .. } => {
+                match &*unit.to_ascii_lowercase() {
```
kdy1, line 645: This code seems bit problematic, as we don't use value of `&*unit.to_ascii_lowercase()`.

crates/swc_css_parser/src/parser/value/mod.rs
```diff
@@ -637,7 +637,43 @@ where
...
+                    // <length>
+                    _ if is_length_unit(unit) => Ok(Dimension::Length(self.parse()?)),
```
kdy1, line 647: I think it should be`unit if is_length_unit(unit)`?

kdy1 (changes requested):

alexander-akait:
@kdy1 Done :+1:

kdy1 (approved):
Thanks!

Issue #3309, opened 2022-01-19
Support `swc_common::GLOBALS` in plugin


kwonoj:
### Describe the bug

Context: 

https://github.com/swc-project/swc/blob/f5af22ef716db76a39ea22cc88f3f92745fb4aad/crates/swc_common/src/syntax_pos/hygiene.rs#L66-L69

https://github.com/swc-project/swc/blob/f5af22ef716db76a39ea22cc88f3f92745fb4aad/crates/swc_common/src/syntax_pos/hygiene.rs#L99-L102

Similar to https://github.com/swc-project/swc/issues/3308, runtime should support `globals` in the plugin.

### Expected behavior

-

### Version

master TOT

kdy1:
We can create a special wrapper type which is `scoped-tls` if it's not a wasm build or `plugin-mode` is not enabled, and which calls `__emit_diagnostic` if it's wasm build and `plugin-mode` is enabled.
With this way, we can reduce work and provide identical API while testing/publishing.

PR #3478, opened 2022-02-08
feat(swc/plugin): support global HANDLER in plugin context

kwonoj:
**Description:**

Followup changes to https://github.com/swc-project/swc/issues/3308 to support better ergonomics around sharing host context.

This PR follows suggested discussion from https://github.com/swc-project/swc/issues/3309#issuecomment-1020968828. PR creates / exports global `HANDLER` to the plugin side by resembling existing scoped_thread_local's interface (specifically, `HANDLER.with`) with existing emitter in the plugin's side. Primary goal for this change is preserving user ergonomics around `HANDLER`: as test case updated, now either plugin or host would use `HANDLER.with` with provided `swc_common::Handler` for same interfaces. 

One of change to support those was lazy initialization of HANDLER in plugin's context via oncecell. This allows us to create a proper handler per each plugin's context. It requires having stub sync impl, however we know plugin will not require thread safety at all.

**BREAKING CHANGE:**
Pretty much unused, but `HostContext` is removed again.

- https://github.com/swc-project/swc/issues/3308

### DIFF
--- a/crates/swc_common/src/errors/diagnostic_builder.rs
+++ b/crates/swc_common/src/errors/diagnostic_builder.rs
@@ -27,7 +27,10 @@ use tracing::debug;
 #[derive(Clone)]
 pub struct DiagnosticBuilder<'a> {
     pub handler: &'a Handler,
+    #[cfg(not(feature = "plugin-mode"))]
     pub(crate) diagnostic: Box<Diagnostic>,
+    #[cfg(feature = "plugin-mode")]
+    pub diagnostic: Box<Diagnostic>,
     allow_suggestions: bool,
 }
 
--- a/crates/swc_plugin/src/context.rs
+++ b/crates/swc_plugin/src/context.rs
@@ -1,17 +1,21 @@
-use swc_common::{errors::Diagnostic, plugin::Serialized};
+use swc_common::plugin::Serialized;
 
 extern "C" {
     fn __emit_diagnostics(bytes_ptr: i32, bytes_ptr_len: i32);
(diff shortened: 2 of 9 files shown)


crates/swc_common/src/errors/diagnostic_builder.rs
```diff
@@ -27,7 +27,10 @@ use tracing::debug;
 #[derive(Clone)]
 pub struct DiagnosticBuilder<'a> {
     pub handler: &'a Handler,
+    #[cfg(not(feature = "plugin-mode"))]
     pub(crate) diagnostic: Box<Diagnostic>,
+    #[cfg(feature = "plugin-mode")]
+    pub diagnostic: Box<Diagnostic>,
```
kwonoj, line 33: Exposing this for https://github.com/swc-project/swc/pull/3478/files#diff-4912e1065d4d9582db46d3a9a3b1ab8319c874737534bcaa725ff35f8b75b0d9R18 only to let plugin send diagnostic created via Handler into host.

crates/swc_plugin/src/context.rs
```diff
@@ -1,46 +1,28 @@
...
-}
-
-/// A struct contains set of utility fn to access host environment inside
-/// of plugin. This is being used for certain operations such as sending
-/// diagnostics back to the host.
-pub struct HostContext {
-    pub diagnostics: Diagnostics,
-}
-
-impl HostContext {
```
kwonoj, line 34: I've retired hostcontext for now as its usage is practically empty for now.

kdy1 (approved):
Great work! LGTM.

PR #3479, opened 2022-02-08
fix(es/module): hoist export

### LINKED ISSUE #3439: Named exports not triggering hoisting
### Describe the bug

We just detected one of our compiled files failing when ran with the following error 

```
cannot access xxx before initialization
```

After a bit of researching, it seems that the code that triggered it was a named export being used before its initialization, this however did not fail with `esbuild` or `tsc`.

I managed to find one big difference in the outputs of `tsc` and `swc` with this test case

```js
// repro case
export { foo }
const foo = 123

// tsc
"use strict";
exports.foo = void 0;
const foo = 123;
exports.foo = foo;

// swc
"use strict";
exports.foo = void 0;
exports.foo = foo;
const foo = 123;
```

Both were tried with the following options
- target: es2019
- module: commonjs

Both transpilation outputs were taken from the playgrounds.

This looks like a bug to me?
Thanks for the time reviewing this!

### Expected behavior

It seems that `swc` is failing to hoist the const definition compared to `tsc` , expected behavior would be what `tsc` is doing

### Version

1.2.136

magic-akari:
**Description:**

**BREAKING CHANGE:**

- fix #3439


magic-akari:
Hi, @kdy1 .
There are some points that I'd like to confirm.

https://github.com/swc-project/swc/blob/484cd92959118d7d8d89b8db742c193b3b024829/crates/swc/src/config/mod.rs#L1058-L1067

Why `import_hoister ` is needed since it is `ModuleConfig::Es6`?

kdy1:
Oh, it's not required, as runtime will hoist it anyway.

magic-akari:
Not all issues have been fixed.

There are some case with `ExportDefaultDecl`. Should we hoist the it as well?

```javascript
foo = 1;
export default function foo() {}
foo = 2;
```

[babel output](https://babeljs.io/repl#?browsers=chrome%2045&build=&builtIns=false&corejs=false&spec=false&loose=false&code_lz=GYexAIF5wRgbgFAFMAeAHEAnALuAJksAIYCuANrsCQHYDG2AliNeKCABQCU4A3gL4I2UcACZEQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=true&targets=&version=7.17.1&externalPlugins=&assumptions=%7B%7D)
```javascript
"use strict";

  value: true
});
exports.default = foo;
exports.default = foo = 1;

function foo() {}

exports.default = foo = 2;

```

[swc output](https://play.swc.rs/?version=1.2.137&code=H4sIAAAAAAAAA0vLz1ewVTC05kqtKMgvKlFISU1LLM0pUUgrzUsuyczPU0jLz9fQVKiu5UoDqzSyBgDdFa2gMgAAAA%3D%3D&config=H4sIAAAAAAAAA0WMSwrDMAxE76K1t93kDj2EcJXg4B8aBWqM714bUrKSmPdmOp3wtHWqrBBdH1o2%2FtJG4hPDa6hGbmoz2jlChiNjPcSWgtdksRTITR2lkMPe1pIvqaoAD%2BJ8xL855lAqn2sFnazVeVcllXyCxjN0lwPet216yfgBCKaI9boAAAA%3D)
```javascript
"use strict";
    value: true
});
exports.default = foo;
foo = 1;
function foo() {}
exports.default = foo = 2;

```

crates/swc/src/config/mod.rs
```diff
@@ -1057,52 +1057,61 @@ impl ModuleConfig {
...
+                            base,
+                            root_mark,
+                            config,
+                            Some(scope),
+                        )
                     ))
                 }
             }
             Some(ModuleConfig::Umd(config)) => {
+                let base_pass = module_hoister();
```
kdy1, line 1090: Seems like this let can be moved to the `else` block?

kdy1 (changes requested):
Thanks!

crates/swc/src/config/mod.rs
magic-akari: Here is wrong.
But it should also be handled in the if clause, right?

kdy1: What do you mean by here is wrong?

magic-akari: I just updated it.

kdy1 (approved):
Thanks!

tmayr:
Wanted to say thanks for the quick response and fix to the bug!

PR #3481, opened 2022-02-08
feat(es/lint): dot-notation rule implementation

ArturAralin:
**Description:**

feat(es/lint): dot-notation rule implementation

**BREAKING CHANGE:**
No


crates/swc_ecma_lints/src/rules/dot_notation.rs
```diff
@@ -0,0 +1,189 @@
...
+use std::{
+    fmt::{self, Debug},
+    sync::Arc,
+};
+use swc_common::{collections::AHashSet, errors::HANDLER, sync::Lazy, SourceMap, Span};
+use swc_ecma_ast::*;
+use swc_ecma_visit::{noop_visit_type, Visit, VisitWith};
+
+static KEYWORDS: Lazy<AHashSet<&'static str>> = Lazy::new(|| {
+    let mut keywords: AHashSet<&'static str> = AHashSet::default();
```
kdy1, line 17: We already has a list for keywords.

kdy1: See: https://rustdoc.swc.rs/swc_ecma_ast/trait.IdentExt.html

crates/swc_ecma_lints/src/rules/dot_notation.rs
```diff
@@ -0,0 +1,189 @@
...
+        let dot_notation_config = config.get_rule_config();
+
+        Self {
+            expected_reaction: *config.get_rule_reaction(),
+            allow_keywords: dot_notation_config.allow_keywords.unwrap_or(true),
+            source_map,
+            pattern: dot_notation_config
+                .allow_pattern
+                .as_ref()
+                .map(|patter| Regex::new(patter).unwrap()),
```
kdy1, line 134: I think it's better to have a cache for regex.

kdy1: Also, can you improve the error message?
`unwrap()` will fail with a bad error message, so `expect()` with a good error message is preferable.

crates/swc_ecma_lints/src/rules/dot_notation.rs
```diff
@@ -0,0 +1,189 @@
...
+
+        self.emit_report(span, quote_type, prop_name);
+    }
+}
+
+impl Visit for DotNotation {
+    noop_visit_type!();
+
+    fn visit_member_prop(&mut self, member: &MemberProp) {
+        if let Some(prop) = member.as_computed() {
```
kdy1, line 175: `if let MemberProp::Computed(prop). = member`
would be better

crates/swc_ecma_lints/src/rules/dot_notation.rs
```diff
@@ -0,0 +1,189 @@
...
+            match prop.expr.as_ref() {
```
kdy1, line 176: Can you use `&*prop.expr` instead?

kdy1 (changes requested):
Thanks!

crates/swc_ecma_lints/src/rules/dot_notation.rs
ArturAralin: > I think it's better to have a cache for regex.

How I can cache regex in this case? Static mutable hash map?

kdy1: You can refer to https://github.com/swc-project/swc/blob/5e82640b42c11b07973b2849347570b643137018/crates/swc/src/config/mod.rs#L871-L905

ArturAralin: Thanks

crates/swc_ecma_lints/src/rules/dot_notation.rs
```diff
@@ -0,0 +1,143 @@
...
+            match &*prop.expr {
+                Expr::Lit(Lit::Str(lit_str)) => {
+                    let quote_type = resolve_string_quote_type(&self.source_map, lit_str).unwrap();
+
+                    self.check(prop.span, quote_type, &*lit_str.value);
+                }
+                Expr::Member(member) => {
+                    member.visit_children_with(self);
+                }
+                _ => {}
```
kdy1, line 139: I think we need `prop.visit_with(self)` is required.

kdy1 (changes requested):
Mostly LGTM. Thanks!

crates/swc_ecma_lints/src/rules/dot_notation.rs
ArturAralin: Why?

kdy1: Because `foo[call(my['var'])]` will not be handled otherwise.

ArturAralin: Oh.. I fixed it and add your example as test case

kdy1 (approved):
Thanks!

PR #3483, opened 2022-02-08
fix(ecma/parser): disallow opt chain as assign target

Austaras:
**Description:**

Only identifier and property access is valid assignment target

**BREAKING CHANGE:**
No


kdy1 (dismissed):
Thanks!

kdy1:
Can you run `UPDATE=1 cargo test` on the parser crate?

kdy1 (approved):

PR #3484, opened 2022-02-08
Refactor names

alexander-akait:
**Description:**

Refactor code:
- Rename `percent` to `percentage` according spec https://www.w3.org/TR/css-values-4/#percentages
- Rename `*Identifier` to `Ident`/`CustomIdent`/`DashedIdent` as in spec https://www.w3.org/TR/css-values-4/#textual-values
- Move some at-rules to own files
- `base.rs` is not contains base types for CSS (we need to move at-rules too, but it will be in future, after we finish refactor at-rules)

**BREAKING CHANGE:**

Yes

No


kdy1 (dismissed):

kdy1:
Can you rebase?

alexander-akait:
Yep, WIP

alexander-akait:
@kdy1 Done :+1:

kdy1 (approved):

Issue #3472, opened 2022-02-07
CSS parser errors when parsing asterisk before css property


skittles1412:
### Describe the bug

`parser.parse_all()` returns an `Err` value when parsing a property prefixed by an asterisk (it's a hack for IE 7 and below [source](https://stackoverflow.com/questions/1690642/purpose-of-asterisk-before-a-css-property)). This is not spec compliant but it should not return an error.

### Input code

```typescript
use swc_common::{input::StringInput, BytePos};
use swc_css_parser::{
    lexer::Lexer,
    parser::{Parser, ParserConfig},
};

fn main() {
    let data = "a { *font-size: 16px; }";
    let input = StringInput::new(data, BytePos(0), BytePos(u32::MAX));
    let config = ParserConfig {
        parse_values: false,
        allow_wrong_line_comments: false,
    };
    let lexer = Lexer::new(input, config);
    let mut parser = Parser::new(lexer, config);
    parser.parse_all().unwrap();
}
```

### Expected behavior

`parser.parse_all()` should not return an error. Instead, this hack should be reflected in the AST somehow.

### Actual behavior

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { inner: (Span { lo: BytePos(4), hi: BytePos(5), ctxt: #0 }, ExpectedButGot("\"}\"")) }'
```

### Version

swc_common = "0.17.3", swc_css_parser = "0.69.0"

alexander-akait:
@kdy1 Expected, because it is violate CSS spec, we can implement `ie` option for this, but I think it was bad for design, I want to say change `*` on `_`, i.e. violate many things in CSS:
```
.selector { !property: value; }
.selector { $property: value; }
.selector { &property: value; }
.selector { *property: value; }
.selector { )property: value; }
.selector { =property: value; }
.selector { %property: value; }
.selector { +property: value; }
.selector { @property: value; }
.selector { ,property: value; }
.selector { .property: value; }
.selector { /property: value; }
.selector { `property: value; }
.selector { ]property: value; }
.selector { #property: value; }
.selector { ~property: value; }
.selector { ?property: value; }
.selector { :property: value; }
.selector { |property: value; }
```

alexander-akait:
As you can see `@property` even violate nested at-rules, and `&property` violate nested CSS selectors (new spec)

kdy1:
I also think it should be at most recovered error. 
It's a syntax error anyway.

[...]

PR #3486, opened 2022-02-08
Improve recovery mode

alexander-akait:
**Description:**

Recovery mode and refactor:
- fixes https://github.com/swc-project/swc/issues/3472
- move stylesheet to base, because it is base structure of parsing https://www.w3.org/TR/css-syntax-3/#parse-stylesheet, we will have more them for recovery mode in future

**BREAKING CHANGE:**

No

No

### DIFF
--- a/crates/swc_css_parser/src/parser/base.rs
+++ b/crates/swc_css_parser/src/parser/base.rs
@@ -7,6 +7,23 @@ use crate::{
 use swc_common::Span;
 use swc_css_ast::*;
 
+impl<I> Parse<Stylesheet> for Parser<I>
+where
+    I: ParserInput,
+{
+    fn parse(&mut self) -> PResult<Stylesheet> {
+        let start = self.input.cur_span()?;
+        let rules = self.parse_rule_list(RuleContext { is_top_level: true })?;
+
+        let last = self.input.last_pos()?;
+
+        Ok(Stylesheet {
+            span: Span::new(start.lo, last, Default::default()),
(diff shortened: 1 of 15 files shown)


crates/swc_css_parser/src/parser/base.rs
```diff
@@ -137,31 +162,9 @@ where
...
-                                span: span!(self, span.lo),
-                                tokens,
-                            })
-                        }
-                    };
-
-                    declarations.push(prop);
+                    declarations.push(DeclarationBlockItem::AtRule(
+                        self.parse_at_rule(Default::default())?,
+                    ));
```
alexander-akait, line 167: No need recovery mode here, because `at-rules` parser itself handle recovery, i.e. `UnknownAtRule`

crates/swc_css_parser/src/parser/base.rs
```diff
@@ -188,9 +191,33 @@ where
                     declarations.push(prop);
                 }
 
-                _ => {
+                // TODO refactor me
+                tok!("}") => {
```
alexander-akait, line 195: Will be fixed when we union `Block` and `SimpleBlock`

alexander-akait:
Found more hacks, let's add them also

skittles1412:
I believe this passes `a{*b:c;}` but fails on `a{*b:c}` (notice the semicolon).

kdy1 (approved):
Thanks!

PR #3487, opened 2022-02-08
Refactor code according spec

alexander-akait:
**Description:**

Just test from postcss, found some cases where we should improve error recovery (no these test here, because they are broken), put in TODO list

**BREAKING CHANGE:**

No

No


alexander-akait:
Wrong branch

PR #3488, opened 2022-02-08
Test more from postcss

alexander-akait:
**Description:**

Just test from postcss, found some cases where we should improve error recovery (no these test here, because they are broken), put in TODO list

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

PR #3491, opened 2022-02-08
feat: urange syntax

alexander-akait:
**Description:**

Feat parse `<urange>` syntax https://www.w3.org/TR/css-syntax-3/#urange-syntax and no errors on this anymore

**BREAKING CHANGE:**

Yes, new types

No


crates/swc_css_parser/src/parser/value/mod.rs
```diff
@@ -1326,6 +1330,214 @@ where
     }
 }
 
+// <urange> =
```
kdy1, line 1333: I love this comment

crates/swc_css_parser/src/parser/value/mod.rs
```diff
@@ -1326,6 +1330,214 @@ where
...
+                            _ => {
+                                unreachable!();
+                            }
+                        };
+
+                        urange.push(question);
+                    }
+                } else {
+                    let question = match bump!(self) {
+                        Token::Delim { value } => value,
```
kdy1, line 1404: I think we should check if it's `?`?

kdy1 (changes requested):
Thanks!

crates/swc_css_parser/src/parser/value/mod.rs
alexander-akait: In future I will add more these comments, it allow to read code without looking on strange logic inside :star:

crates/swc_css_parser/src/parser/value/mod.rs
alexander-akait: Good catch, fixed

alexander-akait:
@kdy1 Fixed and rebased

kdy1 (approved):
Perfect! Thanks!

PR #3492, opened 2022-02-08
feat(swc/plugin): implement proxy for Mark::fresh

kwonoj:
**Description:**

Partial implementation for https://github.com/swc-project/swc/issues/3309, before applying same practices, would like to confirm this would bea feasiblee approach.

As similar to previous implementation with plugin_runtime, this PR implements proxy fn for the globals (Mark::fresh in here) which calls host fn when it's being called in plugin's context. Unlike other imported fn signature this doesn't do serialization or copy bytes across memories as Mark is wrapper to u32 value. Fn passes over values directly then each context construct Mark as needed.

**BREAKING CHANGE:**

- DUMMY_SP moved into swc_plugin::syntax_pos namespace.

- #3309

### DIFF
--- a/crates/swc_common/src/syntax_pos/hygiene.rs
+++ b/crates/swc_common/src/syntax_pos/hygiene.rs
@@ -61,15 +61,54 @@ struct MarkData {
     is_builtin: bool,
 }
 
+/// List of proxy calls injected by the host in the plugin's runtime context.
+/// When related calls being executed inside of the plugin, it'll call these
+/// proxies instead which'll call actual host fn.
+extern "C" {
+    // Instead of trying to copy-serialize `Mark`, this fn directly consume
+    // inner raw value as well as fn and let each context constrcuts struct
+    // on their side.
+    fn __mark_fresh_proxy(mark: i32) -> i32;
+}
+
 impl Mark {
     pub fn fresh(parent: Mark) -> Self {
-        HygieneData::with(|data| {
+        // Note: msvc tries to link against proxied fn for normal build,
+        // have to limit build target to wasm only to avoid it.
+        #[cfg(feature = "plugin-mode")]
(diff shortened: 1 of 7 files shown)


kwonoj:
Two questions

- is this acceptable approach?
- What's recommended pattern to make test cases for this?

crates/swc_common/src/syntax_pos/hygiene.rs
```diff
@@ -61,15 +61,44 @@ struct MarkData {
...
+                parent
+                    .as_u32()
+                    .try_into()
+                    .expect("Should able to convert mark into i32"),
+            )
+            .try_into()
+            .expect("Should able to convert i32 into mark")
+        });
+
+        #[cfg(not(feature = "plugin-mode"))]
```
kdy1, line 92: I think we need `all(not(feature = "plugin-mode"), not(target_arch = "wasm32"))`.

kdy1: It's required for testing while developement.

e.g. User build `swc_common` with `plugin-mode` (because the plugin code depends on `swc_plugin`) and it's native default target instead of wasm.

kdy1: (`cargo test`)

kdy1 (changes requested):
Other changes LGTM.

crates/swc_common/src/syntax_pos/hygiene.rs
kwonoj: I honestly would like to restrict users of `swc_plugin` always specify proper target for build / test and would not reach this path, but yes, that'll be somewhat destructive at this moment.

kdy1 (approved):

PR #3499, opened 2022-02-09
perf(es): Investigate perf issue

kdy1:
**Description:**

I don't like benchmark games, but the performance of swc in https://twitter.com/jarredsumner/status/1485397686224257027 seems weird.

image

- Rust-only benchmark takes 0.86 ms


kdy1:
Closing as outdated.

PR #3502, opened 2022-02-09
[WIP] Refactor code and Union `Block` and `SimpleBlock`

alexander-akait:
**Description:**

Refactor code and union `Block` and `SimpleBlock`.

Also now we have better recovery mode

**BREAKING CHANGE:**

yes

**Related issue (if exists):**

No

PR #3507, opened 2022-02-09
Refactor avoid tokens

alexander-akait:
**Description:**

Simplify AST by removing `Tokens` in favor `Vec<TokenAndSpan>` in most of case (we still have some places, they will be refactored in future). We don't need extra `Tokens` AST, because we can use `Vec<TokenAndSpan>` in this case. Also we need this to finish `Block` and `SimpleBlock`, grammar parse and error recovery

Let's merge after https://github.com/swc-project/swc/pull/3502

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

PR #3509, opened 2022-02-09
test: case for multiple keyframe selectors

alexander-akait:
**Description:**

Just test case for multiple keyframe selectors

**BREAKING CHANGE:**

No

No

### DIFF
--- a/crates/swc_css_codegen/tests/fixture/at-rules/keyframes/3/output.min.css
+++ b/crates/swc_css_codegen/tests/fixture/at-rules/keyframes/3/output.min.css
@@ -1,0 +1,1 @@
+@keyframes hahaha{0%,1%{color:red}100%{color:red}}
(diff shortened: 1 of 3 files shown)


kdy1 (approved):

crates/swc_css_codegen/tests/fixture/at-rules/keyframes/3/output.min.css
```diff
@@ -0,0 +1 @@
+@keyframes hahaha{0%,1%{color:red}100%{color:red}}
```
yisibl, line 1: `100%` can be `to`

alexander-akait: Yes, we don't support minimizing them currently, it is just codegen (i.e. not AST transformations), but yes, we will improve it in near future

PR #3515, opened 2022-02-10
feat(es/lint): add this handling support to no-alert rule

ArturAralin:
**Description:**

Check call "alert" on "this"

**BREAKING CHANGE:**
No


crates/swc_ecma_lints/src/rules/no_alert.rs
```diff
@@ -182,4 +206,28 @@ impl Visit for NoAlert {
             expr.visit_children_with(self);
         }
     }
+
+    fn visit_class(&mut self, class: &Class) {
+        self.inside_class = true;
```
kdy1, line 211: I think this will error if it's nested, as after calling `visit_class` for a nested class, it will be `false`.

kdy1: You can use `let old_inside_class = self.inside_class`.

crates/swc_ecma_lints/src/rules/no_alert.rs
```diff
@@ -182,4 +206,28 @@ impl Visit for NoAlert {
...
+
+        class.visit_children_with(self);
+
+        self.inside_class = false;
+    }
+
+    fn visit_object_lit(&mut self, lit_obj: &ObjectLit) {
+        self.inside_object = true;
```
kdy1, line 219: Same here

crates/swc_ecma_lints/src/rules/no_alert.rs
```diff
@@ -182,4 +206,28 @@ impl Visit for NoAlert {
...
+
+        lit_obj.visit_children_with(self);
+
+        self.inside_object = false;
+    }
+
+    fn visit_arrow_expr(&mut self, arrow_fn: &ArrowExpr) {
+        self.inside_arrow_fn = true;
```
kdy1, line 227: Same here

kdy1 (changes requested):
Sorry for late review. I was busy due to other tasks.

crates/swc_ecma_lints/src/rules/no_alert.rs
ArturAralin: You mean this case? 

```
class X {
  method() {
    class Y () {}
  }
}
```

kdy1: Not exactly. I meant 

```js
class X {
  method() {
    
  }
  
  method2() {
    class Y () {}
    this.alert()
  }
}

```

but your example may also break this.

kdy1 (approved):
Thanks!

PR #3517, opened 2022-02-10
feat(es/lint): add "allow" config for no-console rule

ArturAralin:
**Description:**

* add "allow" config for no-console rule
* highlight whole member expression in report

**BREAKING CHANGE:**
No


crates/swc_ecma_lints/src/rules/no_console.rs
```diff
@@ -12,44 +12,49 @@ const MESSAGE: &str = "Unexpected console statement";
...
-        _ => Some(visitor_rule(NoConsole::new(*rule_reaction, top_level_ctxt))),
+        _ => Some(visitor_rule(NoConsole::new(config, top_level_ctxt))),
     }
 }
 
 #[derive(Debug, Default)]
 struct NoConsole {
     expected_reaction: LintRuleReaction,
     top_level_ctxt: SyntaxContext,
+    allow: Option<AHashSet<String>>,
```
kdy1, line 32: Can you use `JsWord` instead?

kdy1: This value is compared to `JsWord`, and comparison/hash of `JsWord` is super-fast.
It's already hashed on creation, so comparison does not involve `memcmp` calls.

crates/swc_ecma_lints/src/rules/no_console.rs
```diff
@@ -60,7 +65,19 @@ impl NoConsole {
...
+    fn visit_member_expr(&mut self, member: &MemberExpr) {
+        if let Expr::Ident(ident) = member.obj.as_ref() {
+            match &member.prop {
+                MemberProp::Ident(Ident { sym, .. }) => {
+                    self.check(member.span, ident, &*sym);
+                }
+                MemberProp::Computed(ComputedPropName { expr, .. }) => {
+                    if let Expr::Lit(Lit::Str(Str { value, .. })) = expr.as_ref() {
+                        self.check(member.span, ident, &*value);
+                    }
```
kdy1, line 77: I think we need `else { expr.visit_with(self); }` as `console.log` can be used as a property of a computed member expression.

kdy1 (changes requested):
Thanks!

kdy1 (approved):
Hmm... I decided to approve this as my concern would not be true in real-world codes.
Thanks!

PR #3522, opened 2022-02-10
Feat minifier

alexander-akait:
**Description:**

Two first minifier plugins

**BREAKING CHANGE:**

No

No


alexander-akait:
I put bootstrap file here for tracking our progress, after we will finish most of them, I will refactor tests and move all of them to parser, other packages will load CSS fixtures from `swc_css_parser`, so we will have cross testing on minication and code generation and parsing

kdy1 (approved):
LGTM. Thanks!
We may need to merge visitors in the future if it's slow

PR #3531, opened 2022-02-11
refactor: union simple block and block

alexander-akait:
**Description:**

refactor union block and simple block

**BREAKING CHANGE:**

yes

No


kdy1 (approved):
Thanks!

Issue #3311, opened 2022-01-19
Plugin scaffolding support


kwonoj:
### Describe the feature

https://github.com/swc-project/swc/tree/main/tests/rust-plugins/swc_internal_plugin is a simple example how someone can write a plugin. 

But it is a moving target, which is constantly changing for internal integration test purpose. Also it is not good way to officially document / instruct a way to write plugin by pointing some internal test fixture.

There are 2 things roughly needed to write plugin - first, setting up package itself, then apply multivalue polyfill to enable multivalue without rustc's upstream support (which is not working fully yet).

We'd like to have simple, lightweight cli support to provide both capability for the plugin author.

kwonoj:
closed via #3536.

PR #3536, opened 2022-02-11
feat(swc/cli): plugin scaffolding command

kwonoj:
**Description:**

PR introduces initial cli features from native swc cli to create plugin project as well as few bikesheddings.

1. swc cli

Introducing a subcommand named `plugin`, provides utilities for the plugin. For now there is only one subcommand.

```
swc plugin new {name} --target-type=wasm32-wasi
```

One of debate was what should be main subcommand for the plugin creation. I considered below two, and choose path to use `plugin` as main subcommand.

```
swc new --type=plugin ...
swc plugin new ...
```

This is slighlty different to future subcommand swc will support, like `compile`, `bundle`, `minify` which is a verb to the action. However, if we choose `new` as main subcommand further plugin-scoped feature will be scattered around each verb-command. Choosing `plugin` as subcommand allows consolidating any other features.

Implementation of scaffolding is very thin and naive. Mostly, it mimics behavior of what `cargo new` does.

- have embedded string based template for the files we'll generate
- directly write files, do not rely on packages like `cargo-generate` or similar template scaffolding
- rely on system's `git` binary instead of trying to embed git / libgit.

This is mostly due to we only need to write 4-5 files at maximum which doesn't require any heavy templating. Decisions to not embed libgit is to avoid heavy dependencies as well as build time. This is not a firm decision, if we're going to expand usage of git / other vcs other than one invocation (`git init`) we may revisit this. 

Lastly, PR includes change to rename `[plugin_module]` macro to `[plguin_transform]`. This'll indicate macro will be used for the transform plugin more clearly.

**BREAKING CHANGE:**
- `plugin_module` is renamed to `plugin_transform`.

https://github.com/swc-project/swc/issues/3311


crates/swc_cli/src/commands/plugin.rs
```diff
@@ -0,0 +1,234 @@
...
+        write_ignore_file(path)?;
+
+        // Create `Cargo.toml` file with necessary sections
+        paths::write(
+            &path.join("Cargo.toml"),
+            format!(
+                r#"[package]
+name = "{}"
+version = "0.1.0"
+edition = "2018"
```
RiESAEX, line 144: why not use edition 2021 by default

kwonoj: No major reason. Unsure if it should be or not?

kdy1: I think 2021 would be better, as swc uses 2021.

crates/swc_cli/src/commands/plugin.rs
```diff
@@ -0,0 +1,234 @@
...
+        // Create entrypoint src file
+        let src_path = path.join("src");
+        create_dir_all(&src_path)?;
+        paths::write(
+            &src_path.join("lib.rs"),
+            r#"use swc_plugin::{ast::*, plugin_transform};
+
+pub struct TransformVisitor;
+
+impl VisitMut for TransformVisitor {
```
kdy1, line 189: We may reexport `swc_common::util::Take` as VisitMut requires it.

kwonoj: yes, added. Exposing / adding new one should be easy that we can do over iterations.

kdy1 (approved):
Thanks!

PR #3550, opened 2022-02-13
feat(es/lint): implement no-empty-pattern rule

ArturAralin:
**Description:**

no-empty-pattern rule implementation

**BREAKING CHANGE:**
No


crates/swc_ecma_lints/src/rules/no_empty_pattern.rs
```diff
@@ -0,0 +1,90 @@
...
+                    return;
+                }
+
+                elems.iter().for_each(|array_pat_elem| {
+                    if let Some(elem) = array_pat_elem {
+                        self.check(elem);
+                    }
+                });
+            }
+            _ => {}
```
kdy1, line 71: I think we need to handle `Pat::Assign` and `Pat::Rest`.

e.g. 

`[...[]] = arr`

crates/swc_ecma_lints/src/rules/no_empty_pattern.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+impl Visit for NoEmptyPattern {
+    noop_visit_type!();
+
+    fn visit_var_declarator(&mut self, var_declarator: &VarDeclarator) {
+        self.check(&var_declarator.name);
+    }
+
+    fn visit_function(&mut self, function: &Function) {
```
kdy1, line 83: I think `visit_param` is better than `visit_function` for this case.

crates/swc_ecma_lints/src/rules/no_empty_pattern.rs
```diff
@@ -0,0 +1,90 @@
...
+                });
+            }
+            _ => {}
+        }
+    }
+}
+
+impl Visit for NoEmptyPattern {
+    noop_visit_type!();
+
```
kdy1, line 78: Also, we may need to check catch paramters. (I'm not sure if the original rule catches it)

kdy1 (changes requested):

kdy1 (approved):
Thanks! I'll use this PR to the swc bot CI script, so please do not rebase :)

kdy1 (changes requested):
Oh... Can you rebase?
There were merge conflicts.

kdy1 (approved):
Thanks!

PR #3551, opened 2022-02-13
fix: semicolon is delim in new spec

alexander-akait:
**Description:**

Fix `;` is delim in new spec, example - https://www.w3.org/TR/css-values-4/#interpolate

**BREAKING CHANGE:**

yes, new types

no


crates/swc_css_parser/tests/errors/rome/invalid/fit-content/invalid-call/output.stderr
```diff
@@ -1,9 +1,3 @@
-error: Expected Declaration value
- --> $DIR/tests/errors/rome/invalid/fit-content/invalid-call/input.css:2:45
-  |
-2 |     grid-template-columns: fit-content("broken";
-  |                                                ^
-
```
alexander-akait, line 6: We will improve error reporting more after grammar parsing

PR #3552, opened 2022-02-13
refactor: remove `parse_values`

alexander-akait:
**Description:**

We don't use `parse_values` in parser, let's remove it, in future we will add grammar parsing and implement another options to enable/disable values/selectors/at-rules/functions parsing, currently this function just does nothing

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

Issue #3268, opened 2022-01-14
Unnecessary rest parameter transformation when targeting recent Safari versions


cecchi:
### Describe the bug

Safari has supported rest parameters since [version 10](https://caniuse.com/rest-parameters). When targeting recent versions of Safari (e.g. "Safari >= 15"), SWC transforms rest parameters using a for loop.

### Input code

```typescript
function foo(...args) {
  console.log(args)
}

foo(1,2,3)
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true,
  "env": {
    "targets": "Safari >= 15"
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.129&code=H4sIAAAAAAAAA0srzUsuyczPU0jLz9fQ09NLLEov1lSo5lJQSM7PK87PSdXLyU%2FXAIty1XJxgVQZ6hjpGGsCAMQsaGQ5AAAA&config=H4sIAAAAAAAAA0WOPQrDMAxGrxI0Z%2BmQDoH2Bp16AuHKwSX%2BQXJKg%2FHdK5uUbJK%2B9z1U4C0G5gIJWYjbJHvI%2BIUZyHgUwy5lGBXTk8VVqI6QkRfKDZFJszVGoSMdwbvg7N5MJvrEJHJGGJb1T1YV%2Bfja2qFA3hN14RXq6Th6Th4HmHnTncKnd%2FobqocnWmQ33G%2FDZYJaf9Mhm07WAAAA

### Expected behavior

The code should be untransformed:

```js
function foo(...args) {
    console.log(args);
}
foo(1, 2, 3);
```

### Actual behavior

```js
function foo() {
    for(let _len = arguments.length, args = new Array(_len), _key = 0; _key < _len; _key++){
        args[_key] = arguments[_key];
    }
    console.log(args);
}
foo(1, 2, 3);
```

### Version

1.2.129

RiESAEX:
because there is a bug in safari 14.
You can add `bugfixes: true` to env
https://bugs.webkit.org/show_bug.cgi?id=220517
https://github.com/babel/babel/issues/13916

kdy1:
@RiESAEX  Thank you!

PR #3556, opened 2022-02-14
fix(ecma/transform/compat): arrow parameters in class field

### LINKED ISSUE #3471: Arrow functions as class properties compile incorrectly
### Describe the bug

When using specific browser target to reduce how much compilation is being applied to the code, an edge case surfaces that incorrectly transpiles arrow functions as class properties, losing the correct reference to `this`. See the attached playground and you'll notice `_this.getBar(b)` is referencing `_this` which is not a valid reference.

### Input code

```typescript
class Foo {
  getBar(b) {
    return b * 2
  }

  getFoo = ({ b } = {}) => {
    return this.getBar(b)
  }
}

const f = new Foo()
console.log(f.getFoo({ b: 2 }))
```

### Config

```js
{
    env: {
      targets: '> 0.25% and last 2 versions, not dead, not ie <= 11',
      coreJs: 3,
      mode: 'entry'
    },
    jsc: {
      parser: {
        jsx: true,
        exportDefaultFrom: true
      },
      transform: {}
    }
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.133&code=H4sIAAAAAAAAA0vOSSwuVnDLz1eo5lJQSE8tcUos0kjSBPMUFIpSS0qL8hSSFLQUjIACtVwQNSDltgoa1UCJWiCjulZTwdYOVUtJRmaxHtw4sF6g7uT8vOIShTSgnrzUcpCtGppgsfycVL2c%2FHSNND2I6SCjrRSMFGo1NQHVZOk3oQAAAA%3D%3D&config=H4sIAAAAAAAAA0WO3QqDMAyFXyUEdidjE9yFbHuDPUSwUTq0lSbKRPrua53Du5Pz85EV39JgveJIQThkJYtT%2BmCN3AwkTbCjYpFqyWqpF44FKoWONVekSlnvvfCeFjhYZ9slkxo%2FjIFFjohc1%2F%2BbMYEGb6ZsrKjLyBvwhvFg7Dsrr72oYUo3u3nbbG8kPD7hci6rE5Az0JMolDBzEOudFOC8gmEyP2UZ7g%2B4XjHGL6hpzCH9AAAA

### Expected behavior

this.getBar(b) call should work!

### Actual behavior

Error is thrown `Uncaught TypeError: _this.getBar is not a function`, because `_this` is not pointing to the class instance.

### Version

1.2.133

### Additional context

_No response_

Austaras:
closes #3471

### DIFF
--- a/crates/swc_ecma_transforms_compat/src/es2015/parameters.rs
+++ b/crates/swc_ecma_transforms_compat/src/es2015/parameters.rs
@@ -27,6 +27,7 @@ struct Params {
     /// transform an arrow expression to a function expression.
     hoister: FnEnvHoister,
     in_subclass: bool,
+    in_prop: bool,
     c: Config,
 }
 
@@ -394,6 +395,25 @@ impl Params {
 impl VisitMut for Params {
     noop_visit_mut_type!();
 
+    // generally speaking, there won't be class field in here, but Safari 14.1
+    // still has bugs in parameters
+    fn visit_mut_class_prop(&mut self, prop: &mut ClassProp) {
+        prop.key.visit_mut_children_with(self);
+
+        let old_in_prop = self.in_prop;
+        self.in_prop = !prop.is_static;
+        prop.value.visit_mut_with(self);
+        self.in_prop = old_in_prop;
+    }
+
+    // same for private prop
+    fn visit_mut_private_prop(&mut self, prop: &mut PrivateProp) {
+        let old_in_prop = self.in_prop;
+        self.in_prop = !prop.is_static;
+        prop.value.visit_mut_with(self);
+        self.in_prop = old_in_prop;
+    }
+
     fn visit_mut_block_stmt_or_expr(&mut self, body: &mut BlockStmtOrExpr) {
         let old_rep = self.hoister.take();
 
@@ -492,9 +512,17 @@ impl VisitMut for Params {
                     _ => false,
                 });
 
+                let mut local_vars = None;
+
                 // this needs to happen before rest parameter transform
                 if need_arrow_to_function {
-                    f.visit_mut_children_with(&mut self.hoister);
+                    if !self.in_prop {
+                        f.visit_mut_children_with(&mut self.hoister)
+                    } else {
+                        let mut hoister = FnEnvHoister::default();
+                        f.visit_mut_children_with(&mut hoister);
+                        local_vars = hoister.to_stmt();
+                    }
                 }
 
                 let body_span = f.body.span();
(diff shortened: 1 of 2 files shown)


crates/swc_ecma_transforms_compat/src/es2015/parameters.rs
```diff
@@ -536,6 +564,29 @@ impl VisitMut for Params {
                             return_type: Default::default(),
                         },
                     });
+                    *e = match (self.in_prop, local_vars) {
+                        (true, Some(var_decl)) => Expr::Arrow(ArrowExpr {
```
kdy1, line 568: I think we should not generate an arrow function from here, as we apply `arrow` before `parameters`.

kdy1 (changes requested):
Thanks!
I think we may change the pass ordering

crates/swc_ecma_transforms_compat/src/es2015/parameters.rs
Austaras: A little bit context is needed. As discussed in #3268, Safari somehow still has bug for parameters until today(swc have `safari_id_destructuring_collision_in_function_expression` to fix it), so even if class properties is supported, a parameters pass is still needed. So this is only for Safari > 14.1 and `bugfixes: false`, where arrow function is supported perfectly.

[Babel](https://babeljs.io/repl#?browsers=safari%2014.1&build=&builtIns=false&corejs=3.6&spec=true&loose=false&code_lz=MYGwhgzhAECC0G8BQ1pmgXmgRmgamgApCA6MsASkwD40KVoAjTIskxqjW5ggFwAsAljAYReYXoODRpWQrJwAmAMyda0vkJEBfIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=true&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.17.2&externalPlugins=&assumptions=%7B%7D) behaves the same(you need to mannually close bugfixes), and there's no other way to capture `this` in class field

kdy1: But as this pass is applied after arrow, the emitted code will not work in browsers that only support es5.
I'm not talking about browers with support for arrows.

Austaras: For browser that doesn't support arrows, there's no chance they support class field, so this shall still work, but implicit dependency between passes is indeed a bad thing.

kdy1: Oh, I see.
My bad, sorry. I tried creating some example and found that class properties will not exist in ast in such cases. Thanks!

kdy1 (approved):

PR #3565, opened 2022-02-14
feat(es/lint): no-empty-function rule implementation

ArturAralin:
**Description:**
no-empty-function rule implementation.

I used different config structure compare with a eslint no-empty-function rule. Instead of using flat allows array I used allowing by function modifier type.

eslint config example
```
[
  "error",
  { "allow": ["asyncFunctions"] }
]
```

swc lint config
```
[
  "error",
  {
    "functions": ["async"],
    "methods": ["async"],
    "arrowFunctions": ["async"]
  }
]
```

It's a bit more redundantly but it gave more flexibility.

Also I added "considerComment" bool param to configure considering comments into functions

```
[
  "error",
  {
    "considerComment": false
  }
]
```

will report with following code

```
() => {
  // comment only
}
```

**BREAKING CHANGE:**
No


kdy1:
Can you rebase?

kdy1 (changes requested):
Can you add some typescript tests, especially for classes with a private constructor and constructors with parameter properties?
You can use `jsc.parser` to configure syntax.

ArturAralin:
Sure. I'll add ts tests

crates/swc/tests/errors/lints/no-empty-function/default/ts/output.swc-stderr
```diff
@@ -0,0 +1,30 @@
+error: Unexpected empty constructor pattern
+ 
+  |
+2 |   private constructor() {}
```
kdy1, line 4: I think `private consturctor`s are useful and we should not report an error for this.
Does `eslint` print an error for this?

ArturAralin: I can make option for this like

```
constructors: {
  allow: ["private"]
}
```

kdy1: Hmm... In my opinion, it should be allowed by default because it's a valid pattern and has a clear purpose.
I'm not sure about default of other linters, though

ArturAralin: It's really uncommon patten for me. If you think that it's should be ignored by default, I'll made it

kdy1: I found https://github.com/typescript-eslint/typescript-eslint/issues/1178 and hmm... I think error by default is fine, but I prefer to have an option to allow private constructors.

ArturAralin: Ok. I'll allow to turn off it

ArturAralin:
I added ts modifiers support (private/protected/public) for methods and constructors

ArturAralin:
@kdy1 any issues here?

kdy1:
No, I'm just busy.
Can you rebase?

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_lints

PR #3566, opened 2022-02-14
feat(swc/plugin_runner): improve resolver support for npm

### LINKED ISSUE #3548: Plugin template scaffold generates package.json for easier publish
### Describe the feature

https://github.com/swc-project/swc/discussions/3540#discussioncomment-2163346

When we create a plugin via template, we expect it'll be published as npm package in most cases. Having default package.json would be helpful. 

We could do

- create package.json by default 
- having a flag to indicate to create package.json.

For now I'm leaning on first approach, as said expect most of cases we expect plugin will be published as npm pkg and installed next to `@swc/core`.

_No response_

### Additional context

_No response_

kwonoj:
**Description:**

PR attempts to close #3548 to generate simple package.json.

Also change includes resolver changes for the plugin. Now plugin is a single packge to .wasm instead of having per-platform optionaldeps, instead of using custom resolver tries to reuse existing node_resolver. It seems to work reasonably, however might need second eye if it's right location to create resolver.

- closes https://github.com/swc-project/swc/issues/3548


kdy1 (approved):
Thanks!

PR #3568, opened 2022-02-15
feat(swc/plugin): support `context` for the plugin transform

kwonoj:
**Description:**

From discussion https://github.com/swc-project/swc/discussions/3540#discussioncomment-2163757. 

Plugin may need additional data other than config to plugin itself, mostly state around swc/core. Some examples like

- specific file name for current transform
- maybe whole swcrc config, not only for the plugin itself

or possibly more. This PR attempts to resolve those by forwarding new param named `context`.

There are few possible approaches to enable this.

1. expand plugin interface param signature
```
#[plugin_transform]
fn process(program: Program, plugin_config: String, filename: String, swcrc: String...);
```

Obviously, this is not preferrable as we'll continously expand signatures and each time we add new param it'll cause breaking changes.

2. Typed struct param
```
struct PluginContext {
 filename: String,
 other: ...
}

#[plugin_transform]
fn process(program: Program, plugin_config: String, context: PluginContext);
```

This is most clear interface signature. However, have same pitfall as first. Each time we add new properties into struct, it'll be a breaking change since our serialization does not support add-only / versioned

3. Context as JSON-string

Current approach. Safe to add new properties since it's string, plugin can try to pick up necessary properties for their use.

Though I picked 3rd, I'm currently bit debating 2 also - if we can add enough context properties in experimental phase, we can stabilize those to avoid frequent breaking changes. The part I'm not sure is how frequent we'll need to change this context information. Would like to hear thoughts, and update PR into second if needed.

**BREAKING CHANGE:**
Plugin will need to accept 3rd param in any case, it could be a string or a typed struct.

https://github.com/swc-project/swc/discussions/3540#discussioncomment-2163757.

### DIFF
--- a/crates/swc/src/config/mod.rs
+++ b/crates/swc/src/config/mod.rs
@@ -60,7 +60,11 @@ use swc_ecma_transforms_optimization::{inline_globals2, GlobalExprMap};
 use swc_ecma_visit::{Fold, VisitMutWith};
 
 use self::util::BoolOrObject;
-use crate::{builder::PassBuilder, plugin::PluginConfig, SwcComments, SwcImportResolver};
+use crate::{
+    builder::PassBuilder,
+    plugin::{PluginConfig, PluginContext},
+    SwcComments, SwcImportResolver,
+};
 
 #[cfg(test)]
 mod tests;
@@ -425,6 +429,15 @@ impl Options {
             NodeModulesResolver::new(TargetEnv::Node, Default::default(), true),
         );
 
+        let transform_filename = match base {
+            FileName::Real(path) => path.as_os_str().to_str().map(String::from),
+            _ => None,
+        };
+
+        let plugin_context = PluginContext {
+            filename: transform_filename,
+        };
+
         let pass = chain!(
             lint_to_fold(swc_ecma_lints::rules::all(LintParams {
                 program: &program,
@@ -457,7 +470,7 @@ impl Options {
                 ),
                 syntax.typescript()
             ),
-            crate::plugin::plugins(plugin_resolver, experimental),
+            crate::plugin::plugins(plugin_resolver, experimental, plugin_context),
             custom_before_pass(&program),
             // handle jsx
             Optional::new(
--- a/crates/swc/src/plugin.rs
+++ b/crates/swc/src/plugin.rs
@@ -11,13 +11,37 @@ use swc_ecma_loader::resolvers::{lru::CachingResolver, node::NodeModulesResolver
 use swc_ecma_transforms::pass::noop;
 use swc_ecma_visit::{noop_fold_type, Fold};
 
+/// A tuple represents a plugin.
+/// First element is a resolvable name to the plugin, second is a JSON object
+/// that represents configuration option for those plugin.
+/// Type of plugin's configuration is up to each plugin - swc/core does not have
+/// strong type and it'll be serialized into plain string when it's passed to
+/// plugin's entrypoint function.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(deny_unknown_fields, rename_all = "camelCase")]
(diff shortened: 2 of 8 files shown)


kwonoj:
Actually I'm slightly leaning on 2 at this moment. We'll need to figure out what properties we'll expose in that case.

crates/swc/src/plugin.rs
```diff
@@ -57,17 +80,22 @@ impl RustPlugins {
         if let Some(plugins) = &self.plugins {
             for p in plugins {
                 let config_json = serde_json::to_string(&p.1)
-                    .context("failed to serialize plugin config as json")?;
-                let config_json = Serialized::serialize(&config_json)?;
+                    .context("Failed to serialize plugin config as json")
+                    .and_then(|value| Serialized::serialize(&value))?;
+
+                let context_json = serde_json::to_string(&self.plugin_context)
```
kdy1, line 86: Is there a reason you selected json instead of `rkyv`?

kwonoj: I wanted this to be json-based string, instead of strictly typed one. PR body tried to explain those (2 vs. 3). 

While I'm in favor of strict typed with rkyv, this'll make some tricky cases like

```
struct PluginStruct { 
 filename: string
}

... some times later

// This'll be breaking change!
struct PluginContext {
  filename:String;
  another: String
}
```

just adding a new property makes whole swc as breaking major bump. It is same as our AST cases. In this PR, instead makes context as string-based JSON and plugin author to pick up required property only which allows to add new properties without breaking changes. Still, this doesn't prevent removing / changing existing property with breaking changes though.

As wrote in comment (https://github.com/swc-project/swc/pull/3568#issuecomment-1039709771) interfacewise I prefer to have strict typed. If we want to go for those route (and I think we can), we need to define type of PluginContext reasonably suffecient while we do experiment / testing.

Thoughts?

kdy1: I think you are right, thanks!
I didn't think about breaking changes.

kdy1 (dismissed):
Thanks!
Can you rebase it?

kwonoj:
Yup, was doing it already. 🚢

crates/swc/src/config/mod.rs
```diff
@@ -424,6 +428,9 @@ impl Options {
             40,
             NodeModulesResolver::new(TargetEnv::Node, Default::default(), true),
         );
+        let plugin_context = PluginContext {
+            source_file_name: source_file_name.clone(),
```
kdy1, line 432: This should use `base` instead.

kdy1: `source_file_name` is used to control sourcemap, and `base` is the filename.
It's named `base` because the parameter is introduced because of `paths` support, but I think it's a bad name.

kdy1 (changes requested):
I think we should rename variables.

crates/swc/src/config/mod.rs
kwonoj: yes, I was confused by name actually.

crates/swc/src/config/mod.rs
```diff
@@ -425,6 +429,15 @@ impl Options {
             NodeModulesResolver::new(TargetEnv::Node, Default::default(), true),
         );
 
+        let transform_filename = match base {
+            FileName::Real(path) => path.as_os_str().to_str().map(String::from),
+            _ => None,
+        };
+
+        let plugin_context = PluginContext {
+            filename: transform_filename,
```
kwonoj, line 438: simplified property name to `filename` as I expect there won't be another `file` we'll get confused. 

I choose to pickup only real filename to pass forward. Unsure if we want to forward some other values here instead.

kdy1 (approved):
LGTM.

PR #3574, opened 2022-02-15
feat(es/lints): Implement no-dupe-args

magic-akari:
**Description:**

Implement [no-dupe-args](https://eslint.org/docs/rules/no-dupe-args)

**BREAKING CHANGE:**


kdy1 (approved):
swc-bump:
 - swc_ecma_lints

magic-akari:
It should probably be set as a critical error.
In strict mode, all naming duplicates should be considered as errors.
In non-strict mode, named duplicates are only allowed in normal functions, while errors will be reported in arrow functions and class methods.

What do you think about it? @kdy1

kdy1:
Yeah I agree that it should be a critical error by default.

crates/swc_ecma_lints/src/rules/no_dupe_args.rs
```diff
@@ -0,0 +1,59 @@
...
+        });
+    }
+}
+
+impl Visit for NoDupeArgs {
+    noop_visit_type!();
+
+    fn visit_function(&mut self, f: &Function) {
+        let variables: Vec<(JsWord, Span)> = find_ids(&f.params);
+
```
kdy1, line 44: I think we should visit children

crates/swc_ecma_lints/src/rules/no_dupe_args.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn visit_arrow_expr(&mut self, arrow_fn: &ArrowExpr) {
+        let variables: Vec<(JsWord, Span)> = find_ids(&arrow_fn.params);
+
+        self.check(variables);
+    }
+
+    fn visit_constructor(&mut self, n: &Constructor) {
+        let variables: Vec<(JsWord, Span)> = find_ids(&n.params);
+
```
kdy1, line 56: Same here

crates/swc_ecma_lints/src/rules/no_dupe_args.rs
```diff
@@ -0,0 +1,59 @@
...
+    noop_visit_type!();
+
+    fn visit_function(&mut self, f: &Function) {
+        let variables: Vec<(JsWord, Span)> = find_ids(&f.params);
+
+        self.check(variables);
+    }
+
+    fn visit_arrow_expr(&mut self, arrow_fn: &ArrowExpr) {
+        let variables: Vec<(JsWord, Span)> = find_ids(&arrow_fn.params);
```
kdy1, line 49: Same here

kdy1 (changes requested):

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_lints

PR #3578, opened 2022-02-15
fix(plugins): Pass custom filenames to plugin context

williamtetlow:
**Description:**
We should pass custom filenames to plugin context.

**BREAKING CHANGE:**


kdy1 (approved):

kdy1 (approved):
swc-bump:
 - swc

PR #3579, opened 2022-02-15
feat: compress `@keyframes` content

alexander-akait:
**Description:**

feat - compress `@keyframes` content - keyframe name to string when it is possible and keyframe selector blocks

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_css_minifier

yisibl:
Nice work!

crates/swc_css_minifier/src/compress/keyframes.rs
```diff
@@ -0,0 +1,56 @@
...
+
+impl VisitMut for CompressKeyframes {
+    fn visit_mut_keyframes_rule(&mut self, keyframes_rule: &mut KeyframesRule) {
+        keyframes_rule.visit_mut_children_with(self);
+
+        match &keyframes_rule.name {
+            KeyframesName::Str(i)
+                if !matches!(
+                    &*i.value.to_lowercase(),
+                    "initial" | "inherit" | "unset" | "revert" | "default"
```
yisibl, line 19: Need add `revert-layer`,  spec: https://drafts.csswg.org/css-cascade-5/#revert-layer

Consider introducing the global CSSWideKeyword concept, so that the `<custom-ident>` CSS types can be used to make judgments.

- https://github.com/servo/servo/blob/6fced22e4784714540a2bf31b00c5f8149cbba87/components/style/properties/declaration_block.rs#L868
- https://github.com/servo/servo/blob/eef04b11e0ad5af35e9b7b23fe27b31a696123f0/components/style/properties/properties.mako.rs#L1028-L1048

alexander-akait: Yep, make sense, we can create custom function in `swc_css_utils` and put this logic there, so it can be reused in many places, added in my TODO list, thanks

PR #3582, opened 2022-02-16
refactor(swc/plugin): remove direct dep to oncecell

kwonoj:
**Description:**

Minor refactor to use reexport onceCell from swc_common instead of direct dependency.


kdy1 (approved):
swc-bump:
 - swc_plugin

PR #3583, opened 2022-02-16
fix(es/lints): Fix `duplicate-bindings`

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_ecma_lints
 - swc

PR #3584, opened 2022-02-16
refactor(node): provide typed configuration for plugins

kwonoj:
**Description:**

Let's provide typed configuration options for experimental plugin for easier configuration.

### DIFF
--- a/node-swc/src/types.ts
+++ b/node-swc/src/types.ts
@@ -492,7 +492,21 @@ export interface JscConfig {
 
   experimental?: {
     optimizeHygiene?: boolean,
-    keepImportAssertions?: boolean
+    keepImportAssertions?: boolean,
+    /**
+     * Specify the location where SWC stores its intermidiate cache files.
+     * Currently only transform plugin uses this. If not specified, SWC will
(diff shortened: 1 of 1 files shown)


node-swc/src/types.ts
```diff
@@ -492,7 +492,21 @@ export interface JscConfig {
...
   experimental?: {
     optimizeHygiene?: boolean,
-    keepImportAssertions?: boolean
+    keepImportAssertions?: boolean,
+    /**
+     * Specify the location where SWC stores its intermidiate cache files.
+     * Currently only transform plugin uses this. If not specified, SWC will
+     * create `.swc` directories.
+     */
+    cache_root: string;
```
kdy1, line 501: I think this should be `cacheRoot` instead?

kdy1 (changes requested):

node-swc/src/types.ts
kwonoj: 🤦 oh you're right

kdy1 (approved):

kdy1 (approved):
swc-bump:
 - swc

PR #3590, opened 2022-02-16
fix(es/lint): move to `kebab-case`

magic-akari:
**Description:**
Use `kebab-case` to get better compatibility with the existing eslint config.

**BREAKING CHANGE:**
The camelCase is no longer supported.


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_lints

PR #3593, opened 2022-02-16
feat(es/lint): support number as error level

magic-akari:
**Description:**
support number as error level for better compatibility with the existing eslint config.

**BREAKING CHANGE:**


kdy1 (commented):
I think we should have an alias for currently existing rules.

crates/swc_ecma_lints/src/config.rs
```diff
@@ -23,15 +23,41 @@ impl Default for LintRuleReaction {
...
+                _ => LintRuleReaction::Off,
+            },
+        }
+    }
+}
+
 #[derive(Debug, Clone, Default, Serialize, Deserialize)]
 pub struct RuleConfig<T: Debug + Clone + Serialize + Default>(
-    #[serde(default)] LintRuleReaction,
+    #[serde(default)] LintRuleLevel,
```
kdy1, line 54: I think we don't need to change the type.

kdy1: We can implement `Deserialize`, by calling `LintRuleLevel::deserailize` and converting it into `LintRuleReaction`.

kdy1 (changes requested):
Thank you!

crates/swc_ecma_lints/src/config.rs
magic-akari: Do you mean revert the [1fa6ef6](https://github.com/swc-project/swc/pull/3593/commits/1fa6ef608f34a6fd973fe8454b703f70cfcd52f6)?

kdy1: Oh, yes. I didn't know you tried it 🤣

magic-akari: Implementing the LintRuleReactionVisitor requires handling more details.
Adding the Enum wrapper give us a cleaner abstraction.
We can get the normalized `LintRuleReaction` directly through `get_rule_reaction` since we implement the `From<LintRuleLevel>`.

Which implementation method do you prefer? Deserialize or enum wrapper?

kdy1: You don't need to implement visitor, you can use derived implementation of `Deserialize` for `LintRuleLevel `.

There's similar code. https://github.com/swc-project/swc/blob/aa1090023530d128b573dfc6669e7a127345bb9f/crates/swc_estree_ast/src/class.rs#L160-L179

magic-akari: Sorry, I cannot get it.
We're implementing `LintRuleLevel::deserailize`, so we can't actually call ourself, right?

kdy1: Oh, my bad it's a typo. It's `LineRuleLevel`

You can declare an enum with 

```rust
#[derive(Debug, Clone, Copy, Deserialize)]
#[serde(untagged)]
enum LintRuleLevel {
    Str(LintRuleReaction),
    Number(u8),
}
```

can call `deserialize` of `LintRuleLevel`

magic-akari: I tried but failed.
https://github.com/swc-project/swc/blob/4c3f4a6669bf0102cf5ef4cac61b63e9a6793435/crates/swc_ecma_lints/src/config.rs#L26-L49

Note that the definition in `LintRuleReaction` is also present in `LintRuleLevel`.
We are implementing `Deserialize` for `LintRuleReaction` and calling `LintRuleLevel::deserialize` in it.
But it is recursive, so what we are implementing will also be called by `LintRuleLevel::deserialize`. We have actually constructed a spiral recursion.
Unless, we redefine an identical LintRuleReaction2 and map them to the LintRuleReaction.

kdy1: Oh, I see. Then I think we should go with your solution.

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_lints

PR #3596, opened 2022-02-16
test: verify no problems with custom property in at-rules

alexander-akait:
**Description:**

Just test

**BREAKING CHANGE:**

No

No


crates/swc_css_codegen/tests/fixture/at-rules/supports/2/output.min.css
```diff
@@ -0,0 +1 @@
+@supports(--foo: ){html{background:green}}
```
alexander-akait, line 1: Whitespace should be keep

kdy1 (approved):
Thanks!

---

swc-bump:
 - ''

crates/swc_css_codegen/tests/fixture/at-rules/supports/2/output.min.css
yisibl: Can we add more test cases?

https://github.com/evanw/esbuild/commit/9f5e8c8286feec89d3122e5bbbd4f4d6db3a8ebf#diff-6295ad03d759052960b811b40e9e900d036937c31e92dbc67b7a94abadd5a9c6R284-R290

alexander-akait: Yes, but I think there are no problems for this, because we always print ` ` (space) for custom properties if they are empty in codegen, if we can't parse something we keep tokens and print these tokens as is, but I will add test cases

alexander-akait: Also chrome has bug, because by spec `--foo:;` is valid - https://tomhodgins.com/demo/cssom/, open chrome and firefox and write:
```
:root {
  --foo:;
}
```

Also

yisibl: Yes, I asked the CSS experts to discuss it here: https://twitter.com/yisibl/status/1494143919441195011

alexander-akait: Yep, as you can see empty value is valid, there is `TODO` in `chromium` source code about this, anyway now we print ` ` (space) for empty custom properties to avoid problems, in future we will codegen based on browserslist, so when chrome fixes it and your browserslist will be above these versions, we will not print ` ` (space)

PR #3597, opened 2022-02-16
feat: remove empty qualified and at rules

alexander-akait:
**Description:**

Feat: remove empty qualified rules and at-rules

I think we should refactor `AtRule` AST and use only one AtRule structure with different block and prelude types, it will reduces logic for minified and codegen and do our AST is simple, I am working on it

**BREAKING CHANGE:**

No

No


crates/swc_css_minifier/src/compress/empty.rs
```diff
@@ -0,0 +1,86 @@
...
+            | Rule::AtRule(AtRule::Supports(SupportsRule { block, .. }))
+            | Rule::AtRule(AtRule::Page(PageRule { block, .. }))
+            | Rule::AtRule(AtRule::Viewport(ViewportRule { block, .. }))
+            | Rule::AtRule(AtRule::Document(DocumentRule { block, .. }))
+            | Rule::AtRule(AtRule::Property(PropertyRule { block, .. }))
+                if block.value.is_empty() =>
+            {
+                return false;
+            }
+            Rule::AtRule(AtRule::Layer(LayerRule { block, .. })) if block.is_some() => {
```
kdy1, line 28: I think it can be `Rule::AtRule(AtRule::Layer(LayerRule { block: Some(block), .. }))` instead.

kdy1 (changes requested):
Mostly LGTM. Thank you!

alexander-akait:
@kdy1 Done and rebased

kdy1 (approved):
LGTM.

---


swc-bump:
 - swc_css_minifier

PR #3600, opened 2022-02-16
fix: recovery mode for function

alexander-akait:
**Description:**

Fix recovery mode for `function` (regression in next.js)

Ideally we should refactor this logic more and consume components values firstly, when try to resolve to `Vec<Value>`, if failed keep them as is, if not failed apply `Vec<Value>` nodes, but it requires more refactor and can take a time (I will focus on it in this week, so we will not have such regressions in future)

No

No

### DIFF
--- a/crates/swc_css_parser/tests/recovery/delim-token/minus/output.json
+++ b/crates/swc_css_parser/tests/recovery/delim-token/minus/output.json
@@ -204,14 +204,32 @@
                 },
                 "value": [
                   {
-                    "type": "Number",
+                    "type": "PreservedToken",
                     "span": {
                       "start": 51,
                       "end": 52,
                       "ctxt": 0
                     },
-                    "value": 1.0,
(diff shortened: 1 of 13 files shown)


crates/swc_css_parser/tests/recovery/delim-token/minus/output.json
```diff
@@ -204,14 +204,32 @@
...
+                    "span": {
+                      "start": 52,
+                      "end": 53,
+                      "ctxt": 0
+                    },
+                    "token": {
+                      "WhiteSpace": {
+                        "value": " "
+                      }
+                    }
```
alexander-akait, line 232: Previously we lost `whitespaces` on tokens (it is not good), I found some more edge cases, and will it soon

kdy1 (approved):

kdy1 (approved):
swc-bump:
 - swc_css_codegen

Issue #3028, opened 2021-12-13
Pathing under --out-dir has extra segments


alexeagle:
### Describe the bug

When the input directory has multiple segments, swc appears to strip only the first. Remaining segments appear in the output tree, making it more nested than expected.

Example:
```
alexeagle@system76-pc:~/repros/swc-paths$ mkdir -p some/path/to/dir
alexeagle@system76-pc:~/repros/swc-paths$ cat > some/path/to/dir/file.js
const a = 1

alexeagle@system76-pc:~/repros/swc-paths$ npx swc some/path/to/dir --out-dir=some/other/path
Successfully compiled: 1 file with swc (8.46ms)
alexeagle@system76-pc:~/repros/swc-paths$ ls -R some/other/path
some/other/path:
path

some/other/path/path:
to

some/other/path/path/to:
dir

some/other/path/path/to/dir:
file.js
```

I don't think we expected the repeated `path/path` part of that resulting output directory.

Note, under bazel this results in output directories like `bazel-bin/examples/directory/minify/k8-fastbuild/bin/examples/directory/split_app/file1.js`

### Expected behavior

I would have expected that example to write output to `some/other/path/file.js`

### Version

@swc/cli: 0.1.53 @swc/core: 1.2.119

EastblueOkay:
I also encountered the same problem

ruiconti:
Can I work on this, @kdy1?

kdy1:
Of course, thank you!

reintroducing:
Just ran into this as well. The expected output is consistent with how Babel handles the same command but the extra directories are definitely being generated unexpectedly. @ruiconti @kdy1 has there been any headway on this one? i feel like this is a pretty large issue and causes extra code to have to be written to re-copy all the output into the expected dir and remove the extra generated dirs manually.

milovangudelj:
I think I have the same issue.

I'm building a component library for one of my projects and I have the component files under `./src/components`. When running `npx swc ./src/components -d ./dist` I expect all files to be compiled and placed in the destination folder as tsc would do, instead I get an extra `components` folder inside the output `dist` folder.

As a temporary fix I'm compiling everything in the root directory and then renaming the output `components` directory to `dist`. See `build:js` below
```jsonc
// package.json

{
  ... stuff ...
  "sripts": {
    ... other stuff ...
    "build": "yarn clean && yarn build:js && yarn build:types",
[...]

PR #3602, opened 2022-02-16
feat(swc/cli): initial compile command implementation

kwonoj:
**Description:**

We have `plugin` subcommand, now we can work on other subcommand. This is initial effort to port over `@swc/cli`'s compilation command features under `compile` subcommand. It doesn't cover all of existing cli args yet, only covers most basic execution paths.

There are a couple of challenges we need to deal in the future. First, output path calculation relates to the input. This is not unique to this cli, as we have known issues like https://github.com/swc-project/swc/issues/3028 already. However, it's something we need to take care of.

Second one is more tricky - `plugin` options for the js-written plugins. Previously we can rely on node.js runtime as cli runs on those, which new cli doesn't. I don't have great idea yet except either deprecate js plugin or spin node.js separately to get transformed output from those. Both are not great.

### DIFF
--- a/crates/swc_cli/Cargo.toml
+++ b/crates/swc_cli/Cargo.toml
@@ -14,6 +14,17 @@ path = "./src/main.rs"
 
 [dependencies]
 anyhow = "1.0.53"
-clap = {version = "3.0.14", features = ["derive", "wrap_help"]}
+clap = { version = "3.1.0", features = ["derive", "wrap_help"] }
+walkdir = "2"
+rayon = "1"
+swc = { version = "0.127.1", path = "../swc" }
+swc_common = { version = "0.17.5", path = "../swc_common" }
+relative-path = "1.6.1"
+serde = { version = "1", features = ["derive"] }
+serde_json = { version = "1", features = ["unbounded_depth"] }
+
+[dependencies.path-absolutize]
+version = "3.0.11"
+features = ["once_cell_cache"]
 
 [features]
--- a/crates/swc_cli/src/commands/compile.rs
+++ b/crates/swc_cli/src/commands/compile.rs
@@ -1,5 +1,214 @@
-use clap::Parser;
+use std::{
+    fs,
+    path::{Path, PathBuf},
+    sync::Arc,
+};
 
+use anyhow::Context;
+use clap::Parser;
+use path_absolutize::Absolutize;
+use rayon::prelude::*;
+use relative_path::RelativePath;
+use swc::{
+    config::{Config, Options},
+    try_with_handler, Compiler, TransformOutput,
+};
+use swc_common::{sync::Lazy, FilePathMapping, SourceMap};
+use walkdir::WalkDir;
+
+/// Configuration option for transform files.
 #[derive(Parser)]
-#[clap(author, version, about, long_about = None)]
-pub struct CompileOptions {}
+pub struct CompileOptions {
+    /// Path to a .swcrc file to use
+    #[clap(long = "config")]
+    config_file: Option<PathBuf>,
+
+    /// Files to compile
+    files: Vec<PathBuf>,
+
+    /// The output directory
+    #[clap(long)]
+    out_dir: Option<PathBuf>,
+
+    /// Specify specific file extensions to compile.
+    #[clap(long)]
+    extensions: Option<Vec<String>>,
+}
+
+static COMPILER: Lazy<Arc<Compiler>> = Lazy::new(|| {
+    let cm = Arc::new(SourceMap::new(FilePathMapping::empty()));
+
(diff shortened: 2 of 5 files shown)


crates/swc_cli/src/commands/compile.rs
```diff
@@ -1,5 +1,220 @@
...
+    let default = PathBuf::from(".");
+    let base = file_path
+        .parent()
+        .unwrap_or(&default)
+        .as_os_str()
+        .to_string_lossy();
+
+    let dist_absolute_path = out_dir.absolutize()?;
+
+    // These are possible combinations between input to output dir.
```
kwonoj, line 120: This is note for the future what we'll need to support for the path calculation.

crates/swc_cli/src/commands/compile.rs
```diff
@@ -1,5 +1,220 @@
...
+        let files = get_files_list(&self.files, &included_extensions, false)?;
+        let cm = COMPILER.clone();
+
+        if let Some(out_dir) = &self.out_dir {
+            fs::create_dir_all(out_dir)?;
+        }
+
+        files
+            .into_par_iter()
+            .try_for_each_with(cm, |compiler, file_path| {
```
kwonoj, line 202: I choose naive parallel iterator for processing files, maybe there's better / recommended practices?

crates/swc_cli/src/commands/mod.rs
```diff
@@ -17,9 +17,7 @@ pub enum Command {
 }
 
 #[derive(Parser)]
-#[clap(name = "SWC", version)]
-#[clap(global_setting(AppSettings::PropagateVersion))]
-#[clap(global_setting(AppSettings::UseLongFormatForHelpSubcommand))]
+#[clap(name = "SWC", version, propagate_version = true)]
```
kwonoj, line 20: https://epage.github.io/blog/2022/02/clap-31-a-step-towards-40/ and happy to see clap moving towards a modular approach.

crates/swc_cli/Cargo.toml
```diff
@@ -14,6 +14,17 @@ path = "./src/main.rs"
...
+clap = { version = "3.1.0", features = ["derive", "wrap_help"] }
+walkdir = "2"
+rayon = "1"
+swc = { version = "0.127.1", path = "../swc" }
+swc_common = { version = "0.17.5", path = "../swc_common" }
+relative-path = "1.6.1"
+serde = { version = "1", features = ["derive"] }
+serde_json = { version = "1", features = ["unbounded_depth"] }
+
+[dependencies.path-absolutize]
```
kdy1, line 26: Oh, I love this.

crates/swc_cli/src/commands/compile.rs
```diff
@@ -1,5 +1,220 @@
...
+        }
+
+        WalkDir::new(input_dir)
+            .into_iter()
+            .filter_map(|e| e.ok())
+            .map(|e| e.into_path())
+            .filter(|e| {
+                extensions.iter().any(|ext| {
+                    e.extension()
+                        .map(|v| v.to_str().unwrap_or("").eq(ext))
```
kdy1, line 71: `OsStr` implements `PartialEq<str>`. (https://doc.rust-lang.org/stable/std/ffi/struct.OsStr.html#impl-PartialEq%3Cstr%3E)

You can do `v == &**ext` instead.

crates/swc_cli/src/commands/compile.rs
```diff
@@ -1,5 +1,220 @@
...
+
+/// Calculate full, absolute path to the file to emit.
+/// Currently this is quite naive calculation based on assumption input file's
+/// path and output dir are relative to the same directory.
+fn resolve_output_file_path(out_dir: &Path, file_path: &Path) -> anyhow::Result<PathBuf> {
+    let default = PathBuf::from(".");
+    let base = file_path
+        .parent()
+        .unwrap_or(&default)
+        .as_os_str()
```
kdy1, line 115: I think `.display().to_string()` is better.

crates/swc_cli/src/commands/compile.rs
kdy1: The parallel iterator is the most ideal one if we have a list of files to process.

kdy1 (changes requested):
Thank you!

crates/swc_cli/Cargo.toml
kwonoj: Yes, this is not a complete solution but reasonably works well to cover some cases as far as I could check. Still, as written in path resolution logic I believe we'll need further improvement though.

kdy1 (approved):
LGTM. Thanks!

---

swc-bump:
 - swc_cli

PR #3603, opened 2022-02-16
fix(swc/types): fix experimental config options

kwonoj:
**Description:**

Fix mistakenly missing optional flag for plugins.


kdy1 (approved):
swc_bump:
  - swc

PR #3606, opened 2022-02-17
feat(api/rust): Improve error messages for wrong usages of scoped thread-local variables

kdy1:
**Description:**

`scoped-tls` is used various places in swc, but it does not provides meaningful error message on wrong usages.

```
cannot access a scoped thread local variable without calling set first
```

This is all of the error messages. This is not useful, so I created a new wrapper crate, which panics with

```
You should perform this operation in the closure passed to `set` of better_scoped_tls::tests::TESTTLS
```

Note that this wrapper is actually noop on release builds. It will still panic with a bad error message on release builds.


kdy1 (commented):
swc-bump:
 - swc_common
 - swc_ecma_transforms_base
 - swc_estree_ast

kwonoj (approved):

PR #3615, opened 2022-02-17
feat: minify easing functions

alexander-akait:
**Description:**

Feature - port https://github.com/cssnano/cssnano/tree/master/packages/postcss-normalize-timing-functions

**BREAKING CHANGE:**

No

No


crates/swc_css_minifier/src/compress/easing_function.rs
```diff
@@ -0,0 +1,145 @@
...
+                            && *second == 0.1
+                            && *third == 0.25
+                            && *fourth == 1.0
+                        {
+                            *value = Value::Ident(Ident {
+                                span: *span,
+                                value: "easy".into(),
+                                raw: "easy".into(),
+                            })
+                        } else if *first == 0.42
```
kdy1, line 50: Is this correct?

Comparison of a float to.`0.0` and `1.0` will work, but I'm not sure if it's fine to compare with `0.42`.

kdy1: Nevermind, I checked original source code and it uses string to compare the value

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_minifier

Issue #3598, opened 2022-02-16
Investigate performance issues for experimental plugin


kwonoj:
I wrote a fairly simple plugin to extract the first string literal argument from calls to `i18n.get`, in order to collect translation strings that are used.

My goal was to transform an input file into a JSON array of strings. Please disregard that it currently will emit a semicolon at the end of that array :-)

```rust
use std::vec::Vec;
use swc_plugin::{ast::*, plugin_transform, syntax_pos::DUMMY_SP};

pub struct I18NVisitor {
    strings: Vec<String>,
}

impl Default for I18NVisitor {
    fn default() -> I18NVisitor {
        I18NVisitor {
            strings: Vec::new(),
        }
    }
}

impl Visit for I18NVisitor {
    noop_visit_type!();

    fn visit_call_expr(&mut self, call_expr: &CallExpr) {
        if let Callee::Expr(expr) = &call_expr.callee {
            if let Expr::Member(member_expr) = &**expr {
                if let (Expr::Ident(obj_id), MemberProp::Ident(prop_id)) =
                    (&*member_expr.obj, &member_expr.prop)
                {
                    if &*obj_id.sym == "i18n" && &*prop_id.sym == "get" {
                        let args = &call_expr.args;
                        let arg0 = &args[0];
                        let expr = &arg0.expr;
                        if let Expr::Lit(lit_expr) = &**expr {
                            if let Lit::Str(str_lit) = lit_expr {
                                self.strings.push(str_lit.value.to_string());
                            }
                        }
                    }
                }
            }
        }
    }
}

#[plugin_transform]
pub fn process_transform(program: Program, _config: String, _context: String) -> Program {
    let mut visitor = I18NVisitor::default();

    program.visit_with(&mut visitor);

    let strs = visitor.strings.into_iter().map(|s| {
        Some(ExprOrSpread {
            spread: None,
            expr: Box::new(Expr::Lit(Lit::Str(Str {
                span: DUMMY_SP,
                has_escape: false,
                kind: StrKind::default(),
                value: JsWord::from(s),
            }))),
        })
    });
    let expr = Box::new(Expr::Array(ArrayLit {
        span: DUMMY_SP,
        elems: strs.collect(),
    }));
    let stmt = Stmt::Expr(ExprStmt {
        span: DUMMY_SP,
        expr,
    });
    let body = Vec::from([ModuleItem::Stmt(stmt)]);

    return Program::Module(Module {
        span: DUMMY_SP,
        body,
[...]

PR #3616, opened 2022-02-17
refactor(swc/plugin): revise module cache implementation

kwonoj:
**Description:**

This PR attempts to revise our plugin module cache. 

It looks like the current implementation is obviously flawed spends huge amount of time to initialize module. PR moves out cache into shared, static instance across, as well as more aggressively utilize in-memory cache for module instance.

- Cache compiled `Module` instance in-memory instead of bytecodes
Previously we keep compiled bytecode of plugin (`vec<u8>`) in its memory then create module instance from it. It doesn't necessarily have to be. Now we're keeping actual instance of `Module` in memory, instance creation occurs only once when we load cold bytes (either from `fs` or `fs_cache` bytecode)

- bytecode cache versioning
Pretty basic, but introduces versioning for cache root path for compiled byte cache stability. It's fairly safe to move bytecache around machine but it is not safe against different version of `wasmer`. It's bit manual process we need to update version string though.

**Testing**

I used below snippet (https://github.com/swc-project/swc/issues/3598#issuecomment-1042416762) against our `swc_internal_plugin` to see number differences.

```ts
const EXECUTE_COUNT = 10000;

...

const arr = [...Array(EXECUTE_COUNT).keys()];
obs.observe({ entryTypes: ['measure'], buffered: true });
let tasks = arr.map((value) => transform(`console.log(${value})`, {
  jsc: {
    experimental: {
      plugins: [[`${pluginPath}`, {}]]
    }
  }
}));

performance.mark("executeTask-start");
Promise.all(tasks).then(() => {
  performance.mark("executeTask-complete");
  performance.measure("executeTask", "executeTask-start", "executeTask-complete");
});
```

Which schedules 10,000 transforms in async with plugin init phase. Number may vary across devices.

**Before**
```
PerformanceEntry {
  name: 'executeTask',
  entryType: 'measure',
  startTime: 0,
  duration: 46487.8278
}
```

**after**
```
❯❯❯ hyperfine --warmup "2" "node test.js"
Benchmark 1: node test.js
  Time (mean ± σ):      1.335 s ±  0.046 s    [User: 0.003 s, System: 0.009 s]
  Range (min … max):    1.277 s …  1.434 s    10 runs

❯❯❯ node test.js

PerformanceEntry {
  name: 'executeTask',
  entryType: 'measure',
  startTime: 0,
  duration: 1429.4356
}
```

Which shows clear differences between two.

- https://github.com/swc-project/swc/issues/3598


crates/swc_plugin_runner/src/cache.rs
```diff
@@ -0,0 +1,153 @@
...
+        // if constructed Module is available in-memory, directly return it.
+        // Note we do not invalidate in-memory cache currently: if wasm binary is
+        // replaced in-process lifecycle (i.e devserver) it won't be reflected.
+        let in_memory_module = inner_cache.memory_cache.modules.get(&binary_path);
+        if let Some(module) = in_memory_module {
+            return Ok(module.clone());
+        }
+
+        let module_bytes =
+            std::fs::read(&binary_path).context("Cannot read plugin from specified path")?;
```
kwonoj, line 121: Unlike previous implementation, this doesn't try to locate in-memory bytecode cache anymore: since we expect in-memory cache should be available once after we read this file from fs / or fscache for compiled bytecode.

crates/swc_plugin_runner/src/cache.rs
```diff
@@ -0,0 +1,153 @@
...
+
+/// Version for bytecode cache stored in local filesystem.
+///
+/// This MUST be updated when bump up wasmer.
+///
+/// Bytecode cache generated via wasmer is generally portable,
+/// however it is not gauranteed to be compatible across wasmer's
+/// internal changes.
+/// https://github.com/wasmerio/wasmer/issues/2781
+const MODULE_SERIALIZATION_VERSION: &str = "v1";
```
kwonoj, line 23: So, it is reasonably safe to believe bytecode caches are portable (i.e share between CI machines). What we need to ensure is as wrote should invalidate cache when we bump up wasmer. 

The tricky part is when we `expire` those caches. We do not delete caches ever at the moment, so cache will indefinitely grow and users have to manage it on their own. Maybe we may not need to worry that much as other tools basically do the same.

crates/swc_plugin_runner/src/cache.rs
```diff
@@ -0,0 +1,153 @@
...
+/// Current it doesn't have any invalidation or expiration logics like lru,
+/// having a lot of plugins may create some memory pressure.
+#[derive(Default)]
+pub struct InMemoryCache {
+    modules: AHashMap<PathBuf, Module>,
+}
+
+#[derive(Default)]
+pub struct PluginModuleCache {
+    inner: OnceCell<Mutex<CacheInner>>,
```
kdy1, line 46: Why did you use `OnceCell`?

crates/swc_plugin_runner/src/cache.rs
```diff
@@ -0,0 +1,153 @@
...
+/// Bytecode cache generated via wasmer is generally portable,
+/// however it is not gauranteed to be compatible across wasmer's
+/// internal changes.
+/// https://github.com/wasmerio/wasmer/issues/2781
+const MODULE_SERIALIZATION_VERSION: &str = "v1";
+
+/// A shared instance to plugin's module bytecode cache.
+/// TODO: it is unclear how we'll support plugin itself in wasm target of
+/// swc, as well as cache.
+pub static PLUGIN_MODULE_CACHE: Lazy<PluginModuleCache> = Lazy::new(Default::default);
```
kdy1, line 28: We use it in `Lazy`, so `PluginModuleCache` does not need to have `OnceCell` internally

kdy1 (commented):
I'm just curious. It's fine if it's intentional.


---


swc-bump:

crates/swc_plugin_runner/src/cache.rs
kwonoj: What I wanted to achieve was setting value once with dynamic value (cache_root_path for fs), which can be set via onceCell::get. 

Maybe I should change into `PLUGIN_MODULE_CACHE: OnceCell<PluginModuleCache>` then remove inner's OnceCell?

kdy1: Ah, I see.
I think it may result in an error if a user invoke `npx swc` from a different directory than project root.

crates/swc_plugin_runner/src/cache.rs
```diff
@@ -0,0 +1,153 @@
...
+pub struct PluginModuleCache {
+    inner: OnceCell<Mutex<CacheInner>>,
+    /// To prevent concurrent access to `WasmerInstance::new`
+    instantiation_lock: Mutex<()>,
+}
+
+fn create_filesystem_cache(filesystem_cache_root: &Option<String>) -> Option<FileSystemCache> {
+    let mut root_path = if let Some(root) = filesystem_cache_root {
+        Some(PathBuf::from(root))
+    } else if let Ok(cwd) = current_dir() {
```
kdy1, line 54: This will use wrong directory if user manually invokes `npx swc` from non-root directory.
Maybe we need to use directory of `package.json` or `.swcrc`

kwonoj: `filesystem_cache_root` is a configuration option from `.swcrc` already, this is for the default path. 

We do try to create default from cwd and gracefully falls back to not create cache path if cwd is not accessible. npx is usually treated as one-off, on non-CI execution path which would be ok without fs cache support.

kdy1 (approved):
Thanks!

---

swc-bump:

PR #3623, opened 2022-02-18
Feat compress selector

alexander-akait:
**Description:**

Feat port of https://github.com/cssnano/cssnano/tree/master/packages/postcss-minify-selectors with better an+b compression.

One thing is missed, removing duplicated subclass selector, because we need to derive `Hash` and `Eq` traits, I want to it separably 

**BREAKING CHANGE:**

No

No


kdy1 (approved):
LGTM.
Thank you!


---

swc-bump:
 - swc_css_minifier

kdy1:
I think added tests from other PRs are making CI fail.
Can you update the test refs?

crates/swc_css_minifier/src/compress/selector.rs
```diff
@@ -0,0 +1,267 @@
...
+struct CompressSelector {}
+
+impl CompressSelector {}
+
+impl VisitMut for CompressSelector {
+    fn visit_mut_nth(&mut self, nth: &mut Nth) {
+        nth.visit_mut_children_with(self);
+
+        match &nth.nth {
+            // `2n+1` + `odd`
```
konradekk, line 18: this case could be simplified even more: `2n+1`, `2n-1`, `-2n+5` are all odd; you should check whether `(a == 2 || a == -2) && b % 2 == 1` for odds and with last clause `b % 2 == 0` for evens; this covers then the case with `b == 0` below too.

crates/swc_css_minifier/src/compress/selector.rs
```diff
@@ -0,0 +1,267 @@
...
+            }) if *b == 0 => {
+                nth.nth = NthValue::AnPlusB(AnPlusB {
+                    span: *span,
+                    a: *a,
+                    a_raw: a_raw.clone(),
+                    b: None,
+                    b_raw: None,
+                });
+            }
+            _ => {}
```
konradekk, line 73: you seem to lack cases for `a == 1` and `a == -1` too..

crates/swc_css_minifier/src/compress/selector.rs
alexander-akait: `-2n+5` is not `odd`
- `0` -> `-2 * 0 + 5` -> `5`
- `1` -> `-2 * 1 + 5` -> `3`
- `2` -> `-2 * 2 + 5` -> `1`
- `3` -> `-2 * 3 + 5` -> `-1`

crates/swc_css_minifier/src/compress/selector.rs
alexander-akait: `1` and `-1` covered by codegen, it is part of CSS serialization, `1` and `-1` can be safely removed without changing AST, i.e. `1n+4` == `n+4`

crates/swc_css_minifier/src/compress/selector.rs
alexander-akait: Also `2n+3` is not odd, but yes, I see some improvement

alexander-akait:
@kdy1 Ready for review :star:

kdy1 (approved):
LGTM.

crates/swc_css_minifier/src/compress/selector.rs
konradekk: well… -1 is also odd for me (as integer); that the very definition of oddness: yielding 1 from dividing (modulo) by 2.

this might be slightly different in various languages, not sure which is which in rust (see reminder vs. modulo operators, or various definition variants having different elements having same/different signs; cf. [wiki](https://en.wikipedia.org/wiki/Modulo_operation)).

PR #3630, opened 2022-02-19
feat(es/lint): no-loop-func rule implementation

ArturAralin:
**Description:**

no-loop-func rule implementation

**BREAKING CHANGE:**
No


crates/swc_ecma_lints/src/rules/no_loop_func.rs
```diff
@@ -0,0 +1,305 @@
...
+        block.visit_children_with(self);
+
+        self.scopes.pop();
+        self.scoped_unsafe_vars.remove(&block.span);
+    }
+
+    fn visit_for_stmt(&mut self, for_stmt: &ForStmt) {
+        self.loop_depth += 1;
+
+        self.inside_loop_decl = true;
```
kdy1, line 155: I think this value can be wrong for nested loops.

```js
for (;;) {
   for (;;);

   // At here, `inside_loop_decl` will be false.
   foo()
}
```

crates/swc_ecma_lints/src/rules/no_loop_func.rs
```diff
@@ -0,0 +1,305 @@
...
+
+        for_stmt.body.visit_children_with(self);
+
+        self.loop_depth -= 1;
+    }
+
+    fn visit_for_of_stmt(&mut self, for_of_stmt: &ForOfStmt) {
+        self.loop_depth += 1;
+
+        self.inside_loop_decl = true;
```
kdy1, line 179: Same here

crates/swc_ecma_lints/src/rules/no_loop_func.rs
```diff
@@ -0,0 +1,305 @@
...
+
+        for_of_stmt.body.visit_children_with(self);
+
+        self.loop_depth -= 1;
+    }
+
+    fn visit_for_in_stmt(&mut self, for_in_stmt: &ForInStmt) {
+        self.loop_depth += 1;
+
+        self.inside_loop_decl = true;
```
kdy1, line 194: Same here

crates/swc_ecma_lints/src/rules/no_loop_func.rs
```diff
@@ -0,0 +1,305 @@
...
+
+        for_in_stmt.body.visit_children_with(self);
+
+        self.loop_depth -= 1;
+    }
+
+    fn visit_while_stmt(&mut self, while_stmt: &WhileStmt) {
+        self.loop_depth += 1;
+
+        self.inside_loop_decl = true;
```
kdy1, line 209: Same here

crates/swc_ecma_lints/src/rules/no_loop_func.rs
```diff
@@ -0,0 +1,305 @@
...
+
+        while_stmt.body.visit_children_with(self);
+
+        self.loop_depth -= 1;
+    }
+
+    fn visit_do_while_stmt(&mut self, do_while_stmt: &DoWhileStmt) {
+        self.loop_depth += 1;
+
+        self.inside_loop_decl = true;
```
kdy1, line 223: Same here

kdy1 (changes requested):
Thanks!

crates/swc_ecma_lints/src/rules/no_loop_func.rs
ArturAralin: I think not. Firstly I check loop decl (1)

```
    fn visit_for_in_stmt(&mut self, for_in_stmt: &ForInStmt) {
        self.loop_depth += 1;

        self.inside_loop_decl = true;  // start of checking loop head

        for_in_stmt.left.visit_children_with(self);        // check head
        for_in_stmt.right.visit_children_with(self);

        self.inside_loop_decl = false; // end of checking loop head

        for_in_stmt.body.visit_children_with(self); // check body

        self.loop_depth -= 1;
    }

```

kdy1: Oh, then the problem is

```js
for (var a = () => {
  for (;;);
   // At here, `inside_loop_decl` will be false.
};;) {
   
}
```

but I don't expect people to write this code.
So it's fine I guess?

ArturAralin: I think yes

kdy1:
Can you rebase?

kdy1 (approved):
Thank you so much!


---


swc-bump:
 - swc_ecma_lints

PR #3632, opened 2022-02-19
improve experience for new plugin authors

jantimon:
**Description:**

This PR tries to improve two minor developer experience issues:

1.  `swc plugin new` is surprisingly fast!  
the instant feedback which also told us we should not forget about `rustup target add` made us believe something went wrong 😄
maybe we could add an emoji for future beginners?
2. to develop our first visitor the list of possible methods was extremely helpful  
what do you think about adding a link to it?


kwonoj (approved):
🚢

kdy1 (approved):
swc-bump:
 - ''

PR #3633, opened 2022-02-19
perf(es/minifier): Fix performance bug of the mangler

kdy1:
**Description:**

swc_ecma_minifier:
 - Use bi-directional hashmap instead of iterating all keys.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

Issue #3617, opened 2022-02-17
Improper unescaping of Unicode (ES5 and below)


todoa2c:
### Describe the bug

I have confirmed that the problem of improper unescaping of Unicode has been resolved in the following issue.
https://github.com/swc-project/swc/issues/1187

However, there is still a problem with improper unescaping when target is a string of ES5 or lower.

### Input code

```typescript
// a string of all valid unicode whitespaces
module.exports = '\u0009\u000A\u000B\u000C\u000D\u0020\u00A0\u1680\u2000\u2001\u2002' +
  '\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200A\u202F\u205F\u3000\u2028\u2029\uFEFF';
```

### Config

```json
{
    "jsc": {
        "target": "es5"
    }
}
```

### Expected behavior

```
// a string of all valid unicode whitespaces
module.exports = '\u0009\u000A\u000B\u000C\u000D\u0020\u00A0\u1680\u2000\u2001\u2002' + '\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200A\u202F\u205F\u3000\u2028\u2029\uFEFF';
```

### Actual behavior

```
// a string of all valid unicode whitespaces
module.exports = '\t\n\v\f\r \xa0    ' + '          　\u2028\u2029﻿';
```

### Version

1.2.141

todoa2c:
It worked as expected! Thank you very much!

PR #3636, opened 2022-02-19
fix(codegen): fix unicode escaping in es5

williamtetlow:
**Description:**
Fixes https://github.com/swc-project/swc/issues/3617

Babel has the logic `if target <= ES5 && original source string contains any unicode map all non-ascii to unicode`.

It's a bit implicit as the logic is across a few plugins and also through the options passed to `jsesc` (that depend on target and babel version) [but here babel sets](https://github.com/babel/babel/blob/0058b7fef4d028d2826645b28e3b448517cd979d/packages/babel-plugin-transform-literals/src/index.ts#L19.) `node.extra = undefined` if there's any unicode in string literal

[And here if](https://github.com/babel/babel/blob/df27d542ef3871c342b085e41206d21a86673c1d/packages/babel-generator/src/generators/types.ts#L204) `node.extra` exists it will return the string and otherwise process it with `jsesc` 

- I've replicated this logic in this PR.
- I've also added the logic for converting unicode characters `> 0xFFFF` to surrogate pairs to support ES5.
- I've updated the TSC test suite to match TSC and assert against unicode code points instead of characters

**BREAKING CHANGE:**

https://github.com/swc-project/swc/issues/3617


williamtetlow:
Just posting these links here for future reference

https://mathiasbynens.be/notes/javascript-escapes#unicode

https://mathiasbynens.be/notes/javascript-encoding#surrogate-formulae

crates/swc_ecma_codegen/src/lib.rs
```diff
@@ -3070,7 +3071,12 @@ fn unescape_tpl_lit(s: &str, is_synthesized: bool) -> String {
     result
 }
 
-fn escape_without_source(v: &str, target: EsVersion, single_quote: bool) -> String {
+fn escape_without_source(
+    v: &str,
+    target: EsVersion,
+    single_quote: bool,
+    emit_non_ascii_as_unicode: bool,
```
williamtetlow, line 3078: setting this to `true` is equivalent to disabling `jsesc` `minimal` option

https://github.com/mathiasbynens/jsesc#minimal

crates/swc_ecma_codegen/src/lib.rs
```diff
@@ -3134,25 +3156,34 @@ fn escape_with_source(
...
     let orig = match orig {
         Ok(orig) => orig,
         Err(v) => {
-            return escape_without_source(s, target, single_quote.unwrap_or(false));
+            return escape_without_source(s, target, single_quote.unwrap_or(false), false);
         }
     };
 
+    if target <= EsVersion::Es5 {
+        let emit_non_ascii_as_unicode = UNICODE_CODEPOINT.is_match(&orig);
```
kdy1, line 3175: The only concern about this PR I have is the performance of this line.
Do we really need regex?

williamtetlow: @kdy1 I think that's my JS brain defaulting to regex 😛  Simple string contains is enough, updated in [4a11bab](https://github.com/swc-project/swc/pull/3636/commits/4a11bab72e265ab7cdeecb1746262a6c48b317fb)

crates/swc_ecma_codegen/Cargo.toml
```diff
@@ -14,6 +14,7 @@ bitflags = "1"
 memchr = "2.4.1"
 num-bigint = {version = "0.2", features = ["serde"]}
 once_cell = "1.9.0"
+regex = "1"
```
kdy1, line 17: Can you remove this line?

kdy1 (changes requested):
It's mostly LGTM.

---

swc-bump:
  - swc_ecma_codegen

crates/swc_ecma_codegen/Cargo.toml
kdy1: Also, please revert `Cargo.lock`

williamtetlow: resolved in 0ef810b4a4d699c0ebf59a02d5cc139e0f41c062

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_codegen

PR #3638, opened 2022-02-20
feat(es/minifier): Improve name mangler

kdy1:
**Description:**

swc_ecma_minifier:
 - Treat `ClassMethod` as a scope.
 - Treat `Constructor` as a scope.
 - Consider the number of identifier usages.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #3641, opened 2022-02-20
improve swc-plugin package.json template

jantimon:
I had some ideas to improves the generated `package.json` of the `swc plugin new` command.

 1) package.json `scripts`
    -  
    
- `prepublishOnly` (will be run automatically on `npm publish`)
ensure that the very latest source is compiled.
- `test`
I thought running `cargo test` might be a better default than `"echo \"Error: no test specified\" && exit 1"`
what do you think?
- `dev`
maybe this is too much - what do you think?

<br>

 2) package.json `files`
    -  
    
The result will stay the same as before just with less config.

from [https://docs.npmjs.com/cli/v7/configuring-npm/package-json#files](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#files):

> Certain files are always included, regardless of settings:
> 
> package.json
> README
> LICENSE / LICENCE
> The file in the "main" field

you can try it with:

```
npm pack
```

swc-plugin-console-prefix

### DIFF
--- a/crates/swc_cli/src/commands/plugin.rs
+++ b/crates/swc_cli/src/commands/plugin.rs
@@ -190,20 +190,17 @@ target = "{}""#,
     "name": "{}",
     "version": "0.1.0",
     "description": "",
-    "main": "{}",
-    "scripts": {{
-        "test": "echo \"Error: no test specified\" && exit 1"
-    }},
-    "keywords": [],
     "author": "",
     "license": "ISC",
-    "files": [
-        "{}",
-        "README.md"
-    ]
+    "keywords": ["swc-plugin"],
+    "main": "{}",
+    "scripts": {{
+        "prepublishOnly": "cargo build --release"
+    }},
+    "files": []
 }}
 "#,
-                name, dist_output_path, dist_output_path
+                name, dist_output_path
             )
             .as_bytes(),
         )


kwonoj:
for `test` and `dev`, I'd suggest to leave it as-is. First, it is identical to what `npm init --yes` does for minimal scaffolding. Also I'm not sure if we just want to proxy `cargo` command for those.

General idea for scaffolding template is making as minimal as much as user configure anything else for their own.

jantimon:
@kwonoj - okay I removed `dev` and `test` from this PR 👍

I understand that `npm` is not a necessary proxy for `cargo`.  
My idea behind adding `test` anyway was rather to add some guidance for `js` devs seeing a `swc-plugin` on npm.  
Probably their first action (if they are new to rust) would be to check the `package.json` especially `main` and `scripts`.

What about moving the `test` idea to discussions?

kwonoj:
I think current feedback discussion is good enough to see if the current template causes unnecessary inconveniences for the plugin authors. There would be multiple ways to setup scripts for the template, and it is highly subjective to each individual - reason I prefer to have an absolute minimum required field for the template.

Especially, we do not have a custom command to run tests or build. It is straightforward `cargo` subcommand we expect plugin author should be familiarized since our template is rustlang based one. Do not think we have to hide / proxy it into node's specific norm by default. If ppl would like to do it it should be fine, but not as a de facto recommended from swc's core.

jantimon:
I didn't intent this change to be for plugin authors. It is only straightforward for devs who already used cargo before.
This was rather an idea to be more inclusive for npm users who like to make a first contribution to an oss swc plugin and have not used cargo before.

But I am totally fine with adding it and already removed it from this PR.  
What do you think about the other changes?

kwonoj (approved):

kdy1 (commented):
swc-bump:
 - swc_cli

PR #3642, opened 2022-02-20
feat(es): support TypeScript Instantiation Expression

g-plane:
**Description:**

This PR added support of TypeScript new feature: [Instantiation Expression](https://github.com/microsoft/TypeScript/pull/47607).


crates/swc_ecma_ast/src/expr.rs
```diff
@@ -147,6 +147,9 @@ pub enum Expr {
     #[tag("TsAsExpression")]
     TsAs(TsAsExpr),
 
+    #[tag("TsExpressionWithTypeArguments")]
+    TsExprWithTypeArgs(TsExprWithTypeArgs),
```
kdy1, line 151: Is this required?
Also, it does not seem like a good name.

g-plane: I think it's required. The name is borrowed from TypeScript and shortened. Or, do you have any suggestions?

kdy1: `Expr:: TsExprWithTypeArgs` does not seems like an instantiation expression. How about `TsInstantiation`?

g-plane: LGTM.

kdy1 (commented):
LGTM. Thank you!

---

swc-bump:
 - swc_ecma_ast --breaking

kdy1 (approved):

Issue #940, opened 2020-08-06
const enum compile time evaluate


Brooooooklyn:
https://www.typescriptlang.org/docs/handbook/enums.html#const-enums

kdy1:
It's not possible at the moment, as swc process file by file. babel also has same restriction.

Brooooooklyn:
I'm trying replace `ts-jest` in [vue-next](https://github.com/vuejs/vue-next) with `@swc-node/jest`, and I found that `const enum` would cause runtime error in test specs.

kdy1:
I think using spack's system can be a solution.

Something like
```js
bundle({
    constEnum: true,
    merge: false
})
```
?


Do you think it's acceptable?

(I'll provide better api)

Brooooooklyn:
I don't think spack satisfy with the `vue-next` use case.

And there are some other way to make `swc` working fine with `jest` in vue-next, I will do more research in future.

Note: `TypeScript` meet the same problem with the `preserveConstEnums` flag, so I think maybe I could try to fix it in `jest-runtime`.

kdy1:
I can modify spack act like tsc

a.ts -> a.js
b.ts -> b.ts

while handling const enums

Brooooooklyn:
I've not dig into `spack` yet... it's a bundle tool right?

`jest` transformer must transform files one by one, so I guess spack not suitable to use in `jest`

kdy1:
Hmm... 


If ts-jest handles custom enum by using typescript compiler, it can be done.

Inlining only const enum in a file-by-file manner is also possible, but it requires reading dependency file anyway. (tsc will also do it, because logically it is required)

mehrangta:
any updates on this ?

Austaras:
duplicate of #1584?

kdy1:
@Austaras #1584 is a better name in the point of view of implementors, but users will search it by "const enum" and will typically come here.

@mehrangta No, because this is a huge change which cannot be reverted.

kdy1:
As babel now supports const enum, I'll add this to the radar.

Austaras:
What's your opinion on babel's current implementation?

kdy1:
@Austaras It seems like a proper decision, considering the fact that babel only works on one file at a time.
I think matching the behavior of `tsc` is the best way to go.

LeonardSSH:
Any updates? 😅🙄

akshayr20:
Hey, thanks for the great work on above package. 
[...]

PR #3647, opened 2022-02-20
feat(es/typescript): inline enum

magic-akari:
**Description:**
Inline enum if possible.

**BREAKING CHANGE:**


kdy1:
Does this fix #940?

magic-akari:
Partially.
It does not work across files.

crates/swc_ecma_transforms_typescript/src/inline_enum.rs
```diff
@@ -0,0 +1,135 @@
...
+#[serde(rename_all = "camelCase")]
+#[serde(untagged)]
+pub enum InlineEnumConfig {
+    Const(EnumConst),
+    All(bool),
+}
+
+impl Default for InlineEnumConfig {
+    fn default() -> Self {
+        Self::All(true)
```
magic-akari: @kdy1 
I set "inline all as possible" as default option.
What do you think about it?

kdy1: I think it's good enough to be default

crates/swc_ecma_transforms_typescript/src/inline_enum.rs
```diff
@@ -0,0 +1,135 @@
...
+                    }
+                }
+            }
+        }
+
+        e.visit_mut_children_with(self);
+    }
+
+    fn visit_mut_assign_expr(&mut self, a: &mut AssignExpr) {
+        a.right.visit_mut_with(self);
```
kdy1: `a.left` can store expressions.

crates/swc_ecma_transforms_typescript/src/inline_enum.rs
```diff
@@ -0,0 +1,135 @@
...
+use serde::{Deserialize, Serialize};
+use swc_atoms::JsWord;
+use swc_common::collections::AHashMap;
+use swc_ecma_ast::{
+    AssignExpr, ComputedPropName, Expr, Id, Lit, MemberExpr, MemberProp, Module, Script, Str, Tpl,
+};
+use swc_ecma_utils::ident::IdentLike;
+use swc_ecma_visit::{as_folder, noop_visit_mut_type, Fold, VisitMut, VisitMutWith};
+
+pub type TSEnumLit = Rc<RefCell<AHashMap<Id, AHashMap<JsWord, Lit>>>>;
```
kdy1: Can you change this to a struct which hides the actual type?
The field can be like

```rust
pub(crate) values: Rc<RefCell<AHashMap<Id, AHashMap<JsWord, Lit>>>>
```

crates/swc_ecma_transforms_typescript/src/inline_enum.rs
magic-akari: There is a risk to make it default.
See https://www.typescriptlang.org/play?#code/KYOwrgtgBAglDeAoKsA0UBCiC+jEAo4BDAZyiJAE8BKAbQHIZ6BdKAXigEYBuA4sijVoAGVh06IgA

kdy1: If so, we have two options I guess.

The first one is adding a field to `Assumptions` and the second one is defaulting to `EnumConst::Const`.
I think the second one would be better at the moment

kdy1 (changes requested):
Thank you!

crates/swc_ecma_transforms_typescript/src/inline_enum.rs
Austaras: I prefer an assumption, we need to expose `Assumptions` in a near future anyway.

kdy1: Oh, I see. I think the assumption is good too.

Austaras: That would also need to change the options from `InlineEnumConfig` to `bool`

crates/swc_ecma_transforms_typescript/src/strip.rs
```diff
@@ -985,15 +1012,33 @@ where
...
+            let mut enums = self.ts_enum_lit.borrow_mut();
+
+            let k_v = enums.entry(e.id.to_id()).or_default();
+
+            members
+                .iter()
+                .filter(|(_, v)| matches!(v, Expr::Lit(..)))
+                .for_each(|(m, e)| {
+                    let member_key = m.id.as_ref().clone();
+                    let value = e.clone().expect_lit();
```
kdy1, line 1028: Can you add a test for enum which has a template literal in its initializer?

crates/swc_ecma_transforms_typescript/src/inline_enum.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+struct InlineEnum {
+    ts_enum_lit: TSEnumLit,
+    ts_enum_config: TSEnumConfig,
+
+    is_lhs: bool,
+}
+
+trait SkipInlineEnum {
```
kdy1, line 31: Is this trait required?

crates/swc_ecma_transforms_typescript/src/inline_enum.rs
```diff
@@ -0,0 +1,150 @@
...
+    fn skip(&self) -> bool;
+}
+
+impl SkipInlineEnum for InlineEnum {
```
kdy1, line 35: Seems like this is the only impl.
I think it can be a method of `InlineEnum`

kdy1 (changes requested):
Thank you!

crates/swc_ecma_transforms_typescript/src/inline_enum.rs
magic-akari: I had implemented this method on another pub struct while I want to hide it, so I wrote this trait.
I am going to fix it.

crates/swc_ecma_transforms_typescript/src/strip.rs
magic-akari: Done.
see:
https://github.com/swc-project/swc/blob/ac1f5f97f0d801608994b84ff7b51a194d3a2174/crates/swc/tests/fixture/ts-inline-enum/default/input/index.ts#L11-L18
https://github.com/swc-project/swc/blob/ac1f5f97f0d801608994b84ff7b51a194d3a2174/crates/swc/tests/fixture/ts-inline-enum/default/output/index.ts#L10-L17

crates/swc_ecma_transforms_typescript/src/inline_enum.rs
```diff
@@ -0,0 +1,144 @@
...
+    Module, Script, Str, Tpl,
+};
+use swc_ecma_utils::ident::IdentLike;
+use swc_ecma_visit::{as_folder, noop_visit_mut_type, Fold, VisitMut, VisitMutWith};
+
+use crate::TSEnumConfig;
+
+pub(crate) type TSEnumLit = Rc<RefCell<AHashMap<Id, AHashMap<JsWord, Lit>>>>;
+
+pub fn inline_enum(ts_enum_lit: TSEnumLit, ts_enum_config: TSEnumConfig) -> impl Fold + VisitMut {
```
kdy1, line 16: This function is exposing a crate-local type (`TSEnumLit`) in public api.

kdy1: Maybe it should be `pub(crate) fn`?

kdy1 (changes requested):
I found some tiny issues, thank you!

crates/swc_ecma_transforms_typescript/src/inline_enum.rs
magic-akari: Yes, `inline_enum` should not be public API

kdy1 (approved):
Thank you!
LGTM.
Sorry for the delay, I had lots of PRs to review.


---


swc-bump:
 - swc_ecma_transforms_typescript --breaking

PR #3649, opened 2022-02-20
feat: compress declaration

alexander-akait:
**Description:**

Compress few popular properties, I will add them more and more in future

**BREAKING CHANGE:**

No

No


crates/swc_css_minifier/src/compress/declaration.rs
```diff
@@ -0,0 +1,110 @@
...
+                                right.unwrap().clone(),
+                                bottom.unwrap().clone(),
+                            ];
+                        }
+                    }
+                }
+                "font-weight" => {
+                    declaration.value = declaration
+                        .value
+                        .iter()
```
kdy1, line 82: I think we can use `.take().into_iter()` and remove `.clone()` in the closure

kdy1 (changes requested):
Thank you!

crates/swc_css_minifier/src/compress/declaration.rs
alexander-akait: Done :+1:

kdy1 (approved):
LGTM. Thank you!

---

swc-bump:
 - swc_css_minifier

PR #3651, opened 2022-02-20
feat(es/lint): implement radix rule

ArturAralin:
**Description:**

Implemented radix rule.
Also `extract_arg_value_type` replaced with `extract_arg_val` util

**BREAKING CHANGE:**
No


crates/swc_ecma_lints/src/rules/radix.rs
```diff
@@ -0,0 +1,322 @@
...
+    rule::{visitor_rule, Rule},
+    rules::utils::{extract_arg_val, ArgValue},
+};
+
+const OBJ_NAMES: &[&str] = &["Number", "globalThis"];
+
+const MISSING_PARAMS_MESSAGE: &str = "Missing parameters";
+const REDUNDANT_RADIX_MESSAGE: &str = "Redundant radix parameter";
+const MISSING_RADIX_MESSAGE: &str = "Missing radix parameter";
+const INVLID_RADIX_MESSAGE: &str = "Invalid radix parameter, must be an integer between 2 and 36";
```
konradekk, line 23: typo

crates/swc_ecma_lints/src/rules/radix.rs
```diff
@@ -0,0 +1,322 @@
...
+
+            return;
+        }
+
+        match call_expr.args.get(1) {
+            Some(ExprOrSpread { expr, .. }) => {
+                match &extract_arg_val(expr.as_ref(), self.check_parens) {
+                    ArgValue::Ident => {}
+                    ArgValue::Number(radix) => {
+                        if !(2f64..=36f64).contains(radix) {
```
konradekk, line 168: it seems you skip checking for integers; as far as i know `radix` could be equal to `2.5f64` and still accepted.

crates/swc_ecma_lints/src/rules/utils.rs
```diff
@@ -41,3 +42,51 @@ pub fn resolve_string_quote_type(source_map: &Arc<SourceMap>, lit_str: &Str) ->
...
+        Expr::Lit(Lit::Str(Str { value, .. })) => ArgValue::Str(value.clone()),
+        Expr::Lit(Lit::Num(Number { value, .. })) => ArgValue::Number(*value),
+        Expr::Lit(Lit::Regex(Regex { exp, flags, .. })) => ArgValue::RegExp {
+            exp: exp.clone(),
+            flags: flags.clone(),
+        },
+        Expr::Tpl(Tpl { exprs, .. }) => {
+            if exprs.is_empty() {
+                // TODO: extract text from tpl
+                ArgValue::Str(js_word!(""))
```
kdy1, line 74: This seems wrong to me

crates/swc_ecma_lints/src/rules/utils.rs
```diff
@@ -41,3 +42,51 @@ pub fn resolve_string_quote_type(source_map: &Arc<SourceMap>, lit_str: &Str) ->
...
+                ArgValue::Other
+            }
+        }
+        Expr::TaggedTpl(TaggedTpl {
+            tpl: Tpl { exprs, .. },
+            ..
+        }) => {
+            if exprs.is_empty() {
+                // TODO: extract text from tpl
+                ArgValue::Str(js_word!(""))
```
kdy1, line 85: Same here

crates/swc_ecma_lints/src/rules/radix.rs
```diff
@@ -0,0 +1,322 @@
...
+        }
+    }
+
+    fn handle_call(&mut self, call_expr: &CallExpr) {
+        if let Some(callee) = call_expr.callee.as_expr() {
+            self.inside_callee = true;
+
+            callee.visit_with(self);
+
+            self.inside_callee = false;
```
kdy1, line 272: I think this can error for `foo()()`

crates/swc_ecma_lints/src/rules/radix.rs
```diff
@@ -0,0 +1,322 @@
...
+            self.obj = None;
+            self.prop = None;
+        }
+    }
+}
+
+impl Visit for Radix {
+    noop_visit_type!();
+
+    fn visit_expr(&mut self, expr: &Expr) {
```
kdy1, line 287: Isn't it better to declare `visit_callee`?

kdy1 (changes requested):
Sorry for delay. I had too many tasks to do

crates/swc_ecma_lints/src/rules/utils.rs
ArturAralin: I can read value from source map

kdy1: I mean, this will  make `parseInt(10, `foo`)`  treated as `parseInt(10, '')`

ArturAralin: Actually it's no matter for `radix` rule for now. Tpl handling required for rule `prefer_regex_literals`

kdy1: Oh... Seems like missed it while reviewing the PR for regex literals rule

kdy1 (commented):
Can you rebase?

ArturAralin:
@kdy1 I was update `extract_arg_val` behaviour. How tagged call will be extracted as String only if tagged call made on `String.raw`

kdy1 (approved):
LGTM. Thank you!

---

swc-bump:
 - swc_ecma_lints

PR #3653, opened 2022-02-21
fix(es/parser): Allow type arguments inside optional chaining

g-plane:
**Description:**

This PR fixes parser for parsing:

```typescript
fn?.<number>();
```

Note that this is *unrelated* to that "Instantiation Expression" feature, and current TypeScript has already allowed it: [TS playground](https://www.typescriptlang.org/play?#code/GYOw-AdAPCCuC2AjApgJwHwAoCUBuIA), so I fixed this in a separated PR.


kdy1 (approved):
LGTM. Thank you!

swc-bump:
 - swc_ecma_parser

PR #3654, opened 2022-02-21
feat(es/minifier): `const` => `let`

kdy1:
**Description:**

This will be merged after moving more logics to the pure optimizer.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #3655, opened 2022-02-21
perf(es/minifier): Parallelize `join_vars`

kdy1:
**Description:**

Move `join_vars` to the pure optimizer.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #3657, opened 2022-02-21
fix(es/ast): Fix definition of `Expr::TsInstantiation`

kdy1:
**Description:**

`type_args` of `TsInstantiation` is not optional, and it's an invalid state.
So I replaced `TsExprWithTypeArgs` with `TsInstantiation`.


kdy1 (commented):
swc-bump:
 - swc_ecma_ast --breaking

PR #3659, opened 2022-02-21
feat: system js

underfin:
System js is very important to esm loader at legacy broswer, babel is already implement it with `@babel/plugin-transform-modules-systemjs`.  I ported the feature and tests with swc, this is a hard work for review. Thanks for your work, hope get feedback :)


crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+struct SystemJs<R>
+where
+    R: ImportResolver,
+{
+    root_mark: Mark,
+    resolver: Option<(R, FileName)>,
+    config: Config,
+
+    declare_var_idents: Vec<Ident>,
+    export_map: HashMap<(JsWord, Option<SyntaxContext>), Vec<JsWord>>,
```
kdy1, line 30: Please use `AHashMap` instead.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+    export_names: Vec<JsWord>,
+    export_values: Vec<Expr>,
```
kdy1, line 32: I think this seems like a wrong design.

kdy1: Rust does **not** have GC

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+    SystemJs {
+        root_mark,
+        resolver: None::<(NoopImportResolver, _)>,
+        config,
+
+        declare_var_idents: vec![],
+        export_map: HashMap::new(),
+        export_names: vec![],
+        export_values: vec![],
+        export_identifier: quote_ident!("_export"),
```
kdy1, line 53: Why is this field required?

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+}
+
+pub fn system_js(root_mark: Mark, config: Config) -> impl Fold {
+    SystemJs {
+        root_mark,
+        resolver: None::<(NoopImportResolver, _)>,
+        config,
+
+        declare_var_idents: vec![],
+        export_map: HashMap::new(),
```
kdy1, line 50: Please use `Default::default()` instead

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+        export_names: vec![],
+        export_values: vec![],
+        export_identifier: quote_ident!("_export"),
+        context_identifier: quote_ident!("_context"),
```
kdy1, line 54: Why is this field required?

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+where
+    R: ImportResolver,
+{
+    SystemJs {
+        root_mark,
+        resolver: Some((resolver, base)),
+        config,
+
+        declare_var_idents: vec![],
+        export_map: HashMap::new(),
```
kdy1, line 79: Same here

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+        export_names: vec![],
+        export_values: vec![],
+        export_identifier: quote_ident!("_export"),
```
kdy1, line 82: Same here

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+        context_identifier: quote_ident!("_context"),
```
kdy1, line 83: Same here

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+        enter_fn: false,
+        root_fn_decl_idents: vec![],
+        module_item_meta_list: vec![],
+        import_ident_names: vec![],
+    }
+}
+
+struct ModuleItemMeta {
+    export_names: Vec<JsWord>,
+    export_values: Vec<Expr>,
```
kdy1, line 95: Seem wrong to me.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+    has_export_all: bool,
+    src: JsWord,
+    stmts: Vec<Stmt>,
```
kdy1, line 98: Can you rename this field? `stmts` is too general and I'm not sure which values it store.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+    fn replace_assign_expr(&mut self, assign_expr: AssignExpr) -> Expr {
+        match &assign_expr.left {
+            PatOrExpr::Expr(pat_or_expr) => match &**pat_or_expr {
+                Expr::Ident(ident) => {
+                    for (k, v) in self.export_map.iter() {
+                        if is_eq_sym((ident.sym.clone(), Some(ident.span.ctxt())), k) {
+                            let mut expr = Expr::Assign(assign_expr);
+                            for value in v.iter() {
+                                expr = Expr::Call(self.export_call(
+                                    value.clone(),
```
kdy1, line 134: This is super inefficient.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+                pat.visit_with(&mut VarCollector { to: &mut to });
+
+                match &**pat {
+                    Pat::Ident(ident) => {
+                        for (k, v) in self.export_map.iter() {
+                            if is_eq_sym((ident.id.sym.clone(), Some(ident.id.span.ctxt())), k) {
+                                let mut expr = Expr::Assign(assign_expr);
+                                for value in v.iter() {
+                                    expr = Expr::Call(self.export_call(
+                                        value.clone(),
```
kdy1, line 157: This is super inefficient

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+                                    arg: Box::new(Expr::Ident(ident.clone())),
+                                })),
+                                right: Box::new(Expr::Lit(Lit::Num(Number {
+                                    span: DUMMY_SP,
+                                    value: 1.0,
+                                }))),
+                            });
+                            for value in v.iter() {
+                                expr = Expr::Call(self.export_call(
+                                    value.clone(),
```
kdy1, line 222: This is super inefficient

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+        } else {
+            Expr::Update(update_expr)
+        }
+    }
+
+    fn add_export_name(&mut self, key: (JsWord, Option<SyntaxContext>), value: JsWord) {
+        let mut find = false;
+        for (k, v) in self.export_map.iter_mut() {
+            if is_eq_sym(key.clone(), k) {
+                v.push(value.clone());
```
kdy1, line 246: This is super inefficient

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+            let mut tos = vec![];
+            var_declartor.visit_with(&mut VarCollector { to: &mut tos });
+
+            for (sym, ctxt) in tos {
+                if var_declartor.init.is_none() {
+                    for (k, v) in self.export_map.iter_mut() {
+                        if is_eq_sym((sym.clone(), Some(ctxt)), k) {
+                            for value in v.iter() {
+                                self.export_names.push(value.clone());
+                                self.export_values.push(undefined_expr(DUMMY_SP));
```
kdy1, line 438: Use `undefined` instead. `Box<Expr>` typically better than `Expr`

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+                                        })),
+                                    }
+                                    .into_stmt(),
+                                );
+                            }
+                            Decl::Fn(fn_decl) => {
+                                self.export_names.push(fn_decl.ident.sym.clone());
+                                self.export_values.push(Expr::Ident(fn_decl.ident.clone()));
+                                self.add_export_name(
+                                    (fn_decl.ident.sym.clone(), Some(fn_decl.ident.span.ctxt())),
```
kdy1, line 877: Please use `fn_decl.ident.to_id()` instead

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1224 @@
...
+                                    self.export_names.push(js_word!("default"));
+                                    self.export_values.push(Expr::Class(class_expr));
+                                }
+                            }
+                            DefaultDecl::Fn(fn_expr) => {
+                                if let Some(ident) = &fn_expr.ident {
+                                    self.export_names.push(js_word!("default"));
+                                    self.export_values.push(Expr::Ident(ident.clone()));
+                                    self.add_export_name(
+                                        (ident.sym.clone(), Some(ident.span.ctxt())),
```
kdy1, line 936: Same here

kdy1 (changes requested):
Thank you!

crates/swc_ecma_transforms_module/src/system_js.rs
underfin: Can you point at it? I'm not clear =.=

crates/swc_ecma_transforms_module/src/system_js.rs
underfin: I'm not sure what resolve it, Can you help me ? I want to transform `var d = 3; export { d as e, d as f };` to `_export("f", _export("e", d = 3));`.

crates/swc_ecma_transforms_module/src/system_js.rs
kdy1: Sorry, I thought this was used to store `clone()`-d values at the moment of reviewing this line. 

I forgot removing this comment after reviewing the code below.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+        CallExpr {
+            span,
+            callee: Callee::Expr(Box::new(Expr::Ident(quote_ident!("_export")))),
+            args: vec![Expr::Lit(Lit::Str(quote_str!(name))).as_arg(), expr],
+            type_args: None,
+        }
+    }
+
+    fn fold_module_name_ident(&mut self, ident: Ident) -> Expr {
+        if ident.sym.eq("__moduleName") && ident.span.ctxt().outer().eq(&self.root_mark) {
```
kdy1, line 109: Please compare using `&*ident.sym == "__moduleName"`.

kdy1, line 109: Please use `ident.span.ctxt().outer() == self.root_mark` instead

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+
+impl<R> SystemJs<R>
+where
+    R: ImportResolver,
+{
+    fn export_call(&self, name: JsWord, span: Span, expr: ExprOrSpread) -> CallExpr {
+        CallExpr {
+            span,
+            callee: Callee::Expr(Box::new(Expr::Ident(quote_ident!("_export")))),
+            args: vec![Expr::Lit(Lit::Str(quote_str!(name))).as_arg(), expr],
```
kdy1, line 103: I think `quote_str!(name).as_arg()` will work and it's better for readability.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+}
+
+impl<R> SystemJs<R>
+where
+    R: ImportResolver,
+{
+    fn export_call(&self, name: JsWord, span: Span, expr: ExprOrSpread) -> CallExpr {
+        CallExpr {
+            span,
+            callee: Callee::Expr(Box::new(Expr::Ident(quote_ident!("_export")))),
```
kdy1, line 102: `quote_ident!("_export").as_callee()` would be better.

crates/swc_ecma_transforms_module/src/system_js.rs
kdy1: Oh, sorry. It's fine, as the type is `JsWord`.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+        }
+        Expr::Ident(ident)
+    }
+
+    fn replace_assign_expr(&mut self, assign_expr: AssignExpr) -> Expr {
+        match &assign_expr.left {
+            PatOrExpr::Expr(pat_or_expr) => match &**pat_or_expr {
+                Expr::Ident(ident) => {
+                    for (k, v) in self.export_map.iter() {
+                        if ident.to_id().eq(k) {
```
kdy1, line 124: Please use `==` instead of `.eq`. You may need to reference or dereference operands.

kdy1: I think.`ident.to_id() == *k` will work

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                _ => Expr::Assign(assign_expr),
+            },
+            PatOrExpr::Pat(pat) => {
+                let mut to = vec![];
+                pat.visit_with(&mut VarCollector { to: &mut to });
+
+                match &**pat {
+                    Pat::Ident(ident) => {
+                        for (k, v) in self.export_map.iter() {
+                            if ident.to_id().eq(k) {
```
kdy1, line 147: Same here

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+
+                        for to in to {
+                            for (k, v) in self.export_map.iter() {
+                                if to.eq(k) {
+                                    for _ in v.iter() {
+                                        exprs.push(Box::new(Expr::Call(
+                                            self.export_call(
+                                                to.0.clone(),
+                                                DUMMY_SP,
+                                                Expr::Ident(Ident::new(
```
kdy1, line 172: `Expr::Ident()` is not required for `.as_arg()`

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                                                    to.0.clone(),
+                                                    Span::new(BytePos(0), BytePos(0), to.1),
```
kdy1, line 174: Please change this to `DUMMY_SP.with_ctxt(to.1)`

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+            }
+        }
+    }
+
+    fn replace_update_expr(&mut self, update_expr: UpdateExpr) -> Expr {
+        if !update_expr.prefix {
+            match &*update_expr.arg {
+                Expr::Ident(ident) => {
+                    for (k, v) in self.export_map.iter() {
+                        if ident.to_id().eq(k) {
```
kdy1, line 200: Please use `==` instead.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                            let mut expr = Expr::Bin(BinExpr {
+                                span: DUMMY_SP,
+                                op: BinaryOp::Add,
```
kdy1, line 203: Please use `op!(bin, "+")` instead

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                                left: Box::new(Expr::Unary(UnaryExpr {
+                                    span: DUMMY_SP,
+                                    op: UnaryOp::Plus,
```
kdy1, line 206: Please use `op!(unary, "+")` instead

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                                    arg: Box::new(Expr::Ident(ident.clone())),
+                                })),
+                                right: Box::new(Expr::Lit(Lit::Num(Number {
```
kdy1, line 209: `right: 1.0.into()` will work.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+            }
+        } else {
+            Expr::Update(update_expr)
+        }
+    }
+
+    fn add_export_name(&mut self, key: Id, value: JsWord) {
+        let mut find = false;
+        for (k, v) in self.export_map.iter_mut() {
+            if key.eq(k) {
```
kdy1, line 239: Please use `==`

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                    (*export_values.remove(0)).as_arg(),
+                )
+                .into_stmt()],
+            _ => {
+                let mut props = vec![];
+                while !export_names.is_empty() {
+                    let sym = export_names.remove(0);
+
+                    props.push(PropOrSpread::Prop(Box::new(Prop::KeyValue(KeyValueProp {
+                        key: match Ident::verify_symbol(&sym).is_ok() {
```
kdy1, line 274: Please match like `Ok(..) => {}` instead of matching on `is_ok()`.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                            true => PropName::Ident(quote_ident!(sym)),
+                            false => PropName::Str(quote_str!(sym)),
+                        },
+                        value: export_values.remove(0),
+                    }))));
+                }
+                return vec![CallExpr {
+                    span: DUMMY_SP,
+                    callee: Callee::Expr(Box::new(Expr::Ident(quote_ident!("_export")))),
```
kdy1, line 283: `quote_ident!("_export").as_callee()` will work

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                    .as_arg()],
+                    type_args: None,
+                }
+                .into_stmt()];
+            }
+        }
+    }
+
+    fn build_module_item_export_all(&mut self, mut meta: ModuleItemMeta) -> Vec<Stmt> {
+        match meta.has_export_all {
```
kdy1, line 297: I think `if !meta.has_export_all {} else {}` is better, as it's a dedicated syntax.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                let export_obj = quote_ident!("exportObj");
+                let key_ident = quote_ident!("key");
+                let target = quote_ident!(local_name_for_src(&meta.src));
+                meta.setter_fn_stmts.push(Stmt::Decl(Decl::Var(VarDecl {
+                    span: DUMMY_SP,
+                    kind: VarDeclKind::Var,
+                    declare: false,
+                    decls: vec![VarDeclarator {
+                        span: DUMMY_SP,
+                        name: Pat::Expr(Box::new(Expr::Ident(export_obj.clone()))),
```
kdy1, line 311: `name: export_obj.clone().into()` will work, and it's recommended to do so.

(Actually I refactored whole repository to use it because core transform source code is what plugin developers will look when they don't know a solution)

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                })));
+                meta.setter_fn_stmts.push(Stmt::ForIn(ForInStmt {
+                    span: DUMMY_SP,
+                    left: VarDeclOrPat::VarDecl(VarDecl {
+                        span: DUMMY_SP,
+                        kind: VarDeclKind::Var,
+                        declare: false,
+                        decls: vec![VarDeclarator {
+                            span: DUMMY_SP,
+                            name: Pat::Expr(Box::new(Expr::Ident(key_ident.clone()))),
```
kdy1, line 327: Same here

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                            }))),
+                            right: meta.export_values.remove(0),
+                        }
+                        .into_stmt(),
+                    );
+                }
+                meta.setter_fn_stmts.push(
+                    CallExpr {
+                        span: DUMMY_SP,
+                        callee: Callee::Expr(Box::new(Expr::Ident(quote_ident!("_export")))),
```
kdy1, line 387: I think `.as_callee()` can be used instead.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                        args: vec![Expr::Ident(export_obj).as_arg()],
```
kdy1, line 388: `export_obj.as_arg()` will work.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+    fn hoist_var_decl(&mut self, var_decl: VarDecl) -> Option<Expr> {
+        let mut exprs = vec![];
+        for var_declartor in var_decl.decls {
+            let mut tos = vec![];
+            var_declartor.visit_with(&mut VarCollector { to: &mut tos });
+
+            for (sym, ctxt) in tos {
+                if var_declartor.init.is_none() {
+                    for (k, v) in self.export_map.iter_mut() {
+                        if (sym.clone(), ctxt).eq(k) {
```
kdy1, line 429: Please use `==` instead.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                            for value in v.iter() {
+                                self.export_names.push(value.clone());
+                                self.export_values.push(undefined(DUMMY_SP));
+                            }
+                            break;
+                        }
+                    }
+                }
+                self.declare_var_idents
+                    .push(Ident::new(sym, Span::new(BytePos(0), BytePos(0), ctxt)));
```
kdy1, line 439: Please use `DUMMY_SP.with_ctxt` instead

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                self.module_item_meta_list.insert(index, meta);
+            }
+            None => {
+                self.module_item_meta_list.push(m);
+            }
+        }
+    }
+
+    fn hoist_var_decl(&mut self, var_decl: VarDecl) -> Option<Expr> {
+        let mut exprs = vec![];
```
kdy1, line 421: I think type of this variable should be `Vec<Box<Expr>>`, to remove needlesss `memmove`s.

Note that you don't need to annotate type.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                            break;
+                        }
+                    }
+                }
+                self.declare_var_idents
+                    .push(Ident::new(sym, Span::new(BytePos(0), BytePos(0), ctxt)));
+            }
+
+            if let Some(init) = var_declartor.init {
+                exprs.push(Expr::Assign(AssignExpr {
```
kdy1, line 443: Please wrap argument with `Box::new()`.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+        if let VarDeclOrPat::VarDecl(mut var_decl) = var_decl_or_pat {
+            if var_decl.kind == VarDeclKind::Var {
+                let var_declartor = var_decl.decls.remove(0);
+                let mut tos = vec![];
+                var_declartor.visit_with(&mut VarCollector { to: &mut tos });
+
+                for to in tos {
+                    if var_declartor.init.is_none() {
+                        for (k, v) in self.export_map.iter_mut() {
+                            if to.eq(k) {
```
kdy1, line 470: Please use `==` instead

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                                for value in v.iter() {
+                                    self.export_names.push(value.clone());
+                                    self.export_values.push(undefined(DUMMY_SP));
+                                }
+                                break;
+                            }
+                        }
+                    }
+                    self.declare_var_idents
+                        .push(Ident::new(to.0, Span::new(BytePos(0), BytePos(0), to.1)));
```
kdy1, line 480: Same here, `DUMMY_SP` would be better.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                    ..assign
+                }
+                .fold_with(self);
+                self.replace_assign_expr(assign_expr)
+            }
+            Expr::Update(update) => self.replace_update_expr(update),
+            Expr::Call(call) => match call.callee {
+                Callee::Import(_) => Expr::Call(CallExpr {
+                    args: call.args.fold_with(self),
+                    callee: Callee::Expr(Box::new(Expr::Member(MemberExpr {
```
kdy1, line 572: `MemberExpr { }.as_callee()` will work.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                Expr::This(this_expr)
+            }
+            _ => expr.fold_children_with(self),
+        }
+    }
+
+    fn fold_fn_decl(&mut self, fn_decl: FnDecl) -> FnDecl {
+        let is_async = fn_decl.function.is_async;
+        if is_async {
+            self.enter_async_fn = true;
```
kdy1, line 611: As this does not preserve original state, the code like

```js
export async function foo() {
    async function bar() {
    }

    // enter_async_fn is false at here 
}

```

kdy1: You can store it in a variable like e.g. `old_enter_async_fn`and restore it before exiting the function.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1189 @@
...
+                        };
+
+                        let source_alias = local_name_for_src(&src);
+
+                        let mut setter_fn_stmts = vec![];
+
+                        for specifier in import.specifiers {
+                            match specifier {
+                                ImportSpecifier::Default(specifier) => {
+                                    self.import_ident_names.push(specifier.local.sym.clone());
```
kdy1, line 671: Seems like this will not work with identifier management system of swc.

You can read about it at https://rustdoc.swc.rs/swc_ecma_ast/struct.Ident.html

kdy1 (changes requested):
Thank you so much!

I didn't read the full code yet, as I have many tasks to do. (Sorry for it)

underfin:
I already replace some synax with util function. Thanks for your many advice :)

kdy1:
Did you install rust using rustup?
I think the verison of rustfmt can cause it.

underfin:
Default host: x86_64-apple-darwin
rustup home:  /Users/bytedance/.rustup

installed toolchains
--------------------

stable-x86_64-apple-darwin
nightly-2021-11-01-x86_64-apple-darwin
nightly-2021-11-15-x86_64-apple-darwin
nightly-2022-01-19-x86_64-apple-darwin
nightly-x86_64-apple-darwin (default)

active toolchain
----------------

nightly-x86_64-apple-darwin (default)
rustc 1.61.0-nightly (45e2c2881 2022-02-20)


This is my rust toolchain version.

kdy1:
Oh

```
active toolchain
----------------

nightly-2022-01-19-x86_64-apple-darwin (overridden by '/Users/kdy1/projects/master-swc/rust-toolchain')
rustc 1.60.0-nightly (9ad5d82f8 2022-01-18)

```


this is what I get. It seems like you did `rustup override`?

Can you try again after doing `rustup override unset`?

underfin:
I already change toolchain version as same as your version, but `cargo fmt --all --check` also hasn't error...  Run `rustfmt --check /path/to/systemjs.rs -v` also sucess without any changes.

kdy1:
Did you try `cargo +nightly-2022-01-19 fmt --all`?

underfin:
It also not work. I compaired rustup/toolchain/rustfmt version with ci, it is same as ci, only arch is different.

kdy1:
Then I'll run `rustfmt` instead, after reviewing the PR.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1134 @@
...
+            is_generator: false,
+            is_async: false,
+            type_params: Default::default(),
+            return_type: Default::default(),
+        };
+
+        Module {
+            body: vec![CallExpr {
+                span: DUMMY_SP,
+                callee: quote_ident!("System.register").as_callee(),
```
kdy1, line 1091: This is wrong. Use `member_expr!` instead

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1134 @@
...
+        }
+        function_stmts.append(&mut before_body_stmts);
+        function_stmts.push(return_stmt.into());
+
+        let function = Function {
+            params: vec![
+                Param {
+                    span: DUMMY_SP,
+                    decorators: Default::default(),
+                    pat: quote_ident!("_export").into(),
```
kdy1, line 1068: Also, is `quote_ident` correct? Wnat should happen if there's a variable named `_export`? I don't know well about system js

kdy1 (changes requested):

crates/swc_ecma_transforms_module/src/system_js.rs
underfin: Systemjs use `_export` let one module can export somethings. Eg. `export const a = 1` will be transformed to  `export('a', 1)`. You can read docs from here  https://github.com/systemjs/systemjs/blob/main/docs/system-register.md#format-definition.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1143 @@
...
+    R: ImportResolver,
+{
+    root_mark: Mark,
+    resolver: Option<(R, FileName)>,
+    config: Config,
+
+    declare_var_idents: Vec<Ident>,
+    export_map: AHashMap<Id, Vec<JsWord>>,
+    export_names: Vec<JsWord>,
+    export_values: Vec<Expr>,
```
kdy1, line 33: Please use `Vec<Box<Expr>>` instead.

kdy1: It will be faster because you have to box it anyway, and you can prevent `memmove` of a large type (`Expr`) by doing so.

kdy1: If clippy complains, uou can add `#[allow(clippy::vec_box)]` or `#![allow(clippy::vec_box)]` on `lib.rs`

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1143 @@
...
+    }
+
+    fn add_declare_var_idents(&mut self, ident: &Ident) {
+        self.declare_var_idents.push(ident.clone());
+    }
+
+    fn build_export_call(
+        &mut self,
+        export_names: &mut Vec<JsWord>,
+        export_values: &mut Vec<Expr>,
```
kdy1, line 243: Please use `&mut Vec<Box<Expr>>` instead

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1143 @@
...
+    ) -> Vec<Stmt> {
+        match export_names.len() {
+            0 => vec![],
+            1 => vec![self
+                .export_call(export_names.remove(0), DUMMY_SP, export_values.remove(0))
+                .into_stmt()],
+            _ => {
+                let mut props = vec![];
+                while !export_names.is_empty() {
```
kdy1, line 252: You can use 

```rust
for (sym, value) in export_names.drain(..).rev().zip(export_values.drain(..).rev()) {

}

```


this will be much more efficient

kdy1: Current code requires `memmove` of all elements per loop, but it's not required.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1143 @@
...
+                                )),
+                                right: Box::new(target.computed_member(key_ident)),
+                            }
+                            .into_stmt()],
+                        })),
+                        alt: None,
+                    })],
+                })),
+            }));
+            while !meta.export_names.is_empty() {
```
kdy1, line 350: Same here, please use for loop with a reversed iterator instead.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1143 @@
...
+                    left: PatOrExpr::Pat(Box::new(var_declartor.name)),
+                    right: init,
+                })));
+            }
+        }
+        match exprs.len() {
+            0 => None,
+            _ => Some(Expr::Seq(SeqExpr {
+                span: DUMMY_SP,
+                exprs: exprs.into_iter().collect(),
```
kdy1, line 436: This can be changed to `exprs`.

kdy1: `rustc` will optimize it into `exprs` as core has a specialization for this case, but we should not rely on it.

kdy1: (As it's an impl detail)

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1143 @@
...
+            .into_iter()
+            .map(|s| self.hoist_variables(s).fold_with(self))
+            .filter(|s| !matches!(s, Stmt::Empty(_)))
+            .collect();
+
+        // ====================
+        //  generate export call, here export_names is collected finished.
+        // ====================
+        if !self.export_names.is_empty() {
+            let mut export_names = self.export_names.drain(0..).collect();
```
kdy1, line 967: Please use `drain(..)` instead

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1143 @@
...
+            let mut export_values = self.export_values.drain(0..).collect();
```
kdy1, line 968: Please use `drain(..)` instead

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1143 @@
...
+            span: DUMMY_SP,
+            elems: vec![],
+        };
+
+        let mut dep_module_names = ArrayLit {
+            span: DUMMY_SP,
+            elems: vec![],
+        };
+
+        while !self.module_item_meta_list.is_empty() {
```
kdy1, line 983: Please use a for loop with a reversed iterator.

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -0,0 +1,1143 @@
...
+    base: FileName,
+    root_mark: Mark,
+    config: Config,
+) -> impl Fold
+where
+    R: ImportResolver,
+{
+    SystemJs {
+        root_mark,
+        resolver: Some((resolver, base)),
```
kdy1, line 76: Can you change this to use https://github.com/swc-project/swc/blob/ded7a8e0a96842e48a4aea7cca565b4c22b45da8/crates/swc_ecma_transforms_module/src/path.rs#L16-L22 instead?

This is important for reducing size of the binary.

kdy1 (changes requested):
Thank you so much and sorry for the late review. I had lots of PRs to review :)

underfin:
Thanks for your review :). I already resolve it, your reviews is important for me to learn rust. Thanks again: )

kdy1:
I'm glad to hear it and no problem :)

I always try to give meaningful information while reviewing swc (and next.js) PRs because I'm very grateful for those PRs

crates/swc_ecma_transforms_module/src/system_js.rs
kdy1: Seems like this issue still exists?

kdy1 (changes requested):
Thank you!

underfin:
Look like I forgot remove it ==.

kdy1 (approved):
Thank you!

---

swc-bump:

PR #3664, opened 2022-02-21
feat: compress deg to zero

alexander-akait:
**Description:**

feat - compress `<Angle>` with `0` value to `0`

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_css_minifier

yisibl:
This may not be very safe:

https://twitter.com/tabatkins/status/676297504237879296
> And Chrome/Edge's behavior is a bug. "0" (without a unit) can only be used for lengths, not angles or anything else.

yisibl:
Sorry, I misremembered, the specification was updated later, the angle value can omit the unit: https://github.com/cssdream/css-creating/issues/14

alexander-akait:
@yisibl Yes, but there is exception, it is legacy transfrom functions and they are supported `0` https://www.w3.org/TR/css-transforms-2/#three-d-transform-functions, we don't replace `0` in declarations

alexander-akait:
So we replace them on `0` only for these cases

yisibl:
In the current spec, CSS Gradients can be omitted as well. But I don't know when Firefox implemented it, so I think swc can be safer to leave it out for this case.

It's really a rabbit hole.

alexander-akait:
@yisibl https://tomhodgins.com/demo/cssom/, open firefox and write:
```
.class6{transform:rotateX(0)}
```

Works fine

yisibl:
Firefox 55 fixed “Zero (0) angle value without degree unit is not correctly interpreted in linear-gradient” https://bugzilla.mozilla.org/show_bug.cgi?id=1363292

image
https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient()

alexander-akait:
Yep, but we don't convert to `0` in `linear-gradient`, only in transform functions

yisibl:
That's great, thanks for the discussion.

PR #3665, opened 2022-02-21
fix(ecma/transform/compat): private properties cannot be initialized twice

### LINKED ISSUE #1743: Class Private Fields/Accessors/Methods must not be re-initialized on instance
**Describe the bug**

(Example copied from https://github.com/babel/babel/issues/13299)

**Input code**

(Run with Node)

```js
const assert = require("assert");

class Base {
  constructor(obj) {
    return obj;
  }
}

let counter = 0;
class Derived extends Base {
  #foo = ++counter;
  static get(obj) {
    return obj.#foo;
  }
}

const foo = {};
new Derived(foo);
assert.equal(Derived.get(foo), 1);

// This should throw an error, private fields must not be re-initialized
// on an object that already contains the field.
assert.throws(() => {
  new Derived(foo);
});
// ^ That should have thrown, but it doesn't.

// Must not re-initialize the field
assert.equal(Derived.get(foo), 1);
// But the counter is incremented (the initializer runs, but the field
// should not re-add)
assert.equal(counter, 2);
```

**Config**

**Current behavior**
The assertions fail after transpiling

**Expected behavior**
Behaviour doesn't change

**Version**
The version of @swc/core: 1.2.58

Austaras:
**Description:**

**BREAKING CHANGE:**

closes #1743


kdy1:
Does this fixes #1743?

Austaras:
> Does this fixes #1743?

Yes

kdy1 (approved):
LGTM. Thank you!

---

swc-bump:
 - swc_ecma_transforms_base
 - swc_ecma_transforms_compat

Austaras:
Do I need to manually rebase on main branch?

kdy1:
No, it will be automatically rebased.

I think the auto-rebase order should be reversed, though.
(I thought it will rebase the oldest PR first, but it does it in opposite order)

Issue #3523, opened 2022-02-10
Panic error when attempting to do spread operator


edwincoronado:
### Describe the bug

As part of some testing, when going through the [test262 suite](https://github.com/tc39/test262) and most specifically test case [obj-rest-not-last-element-invalid.js](https://github.com/tc39/test262/blob/main/test/language/statements/for-of/dstr/obj-rest-not-last-element-invalid.js), SWC is crashing with the following error (Detailed log when using `RUST_BACKTRACE=1 SWC_DEBUG=1 npx swc --sync <file>`):

```log
Panic: PanicInfo { payload: Any { .. }, message: Some(internal error: entered unreachable code: Object rest pattern should be removed by es2018::object_rest_spread pass), location: Location { file: "crates/swc_ecma_transforms_compat/src/es2015/destructuring.rs", line: 424, col: 52 } }
Backtrace:    0: <unknown>
   1: <unknown>
   2: <unknown>
   3: _rust_eh_personality
   4: _rust_eh_personality
   5: _rust_eh_personality
   6: _rust_eh_personality
   7: _wasmer_vm_f64_nearest
   8: _rust_eh_personality
   9: _rust_eh_personality
  10: _rust_eh_personality
  11: _rust_eh_personality
  12: _rust_eh_personality
  13: _rust_eh_personality
  14: _rust_eh_personality
  15: _rust_eh_personality
  16: _rust_eh_personality
  17: _rust_eh_personality
  18: _rust_eh_personality
  19: <unknown>
  20: __ZN6v8impl12_GLOBAL__N_123FunctionCallbackWrapper6InvokeERKN2v820FunctionCallbackInfoINS2_5ValueEEE
  21: __ZN2v88internal25FunctionCallbackArguments4CallENS0_15CallHandlerInfoE
  22: __ZN2v88internal12_GLOBAL__N_119HandleApiCallHelperILb0EEENS0_11MaybeHandleINS0_6ObjectEEEPNS0_7IsolateENS0_6HandleINS0_10HeapObjectEEESA_NS8_INS0_20FunctionTemplateInfoEEENS8_IS4_EENS0_16BuiltinArgumentsE
  23: __ZN2v88internalL26Builtin_Impl_HandleApiCallENS0_16BuiltinArgumentsEPNS0_7IsolateE

failed to handle: internal error: entered unreachable code: Object rest pattern should be removed by es2018::object_rest_spread pass

Stack backtrace:
   0: _rust_eh_personality
[...]

PR #3675, opened 2022-02-22
fix(ecma/parser): non last rest element should be error in object pattern

Austaras:
…tern

**Description:**

**BREAKING CHANGE:**

**Related issue (if exists):**
closes #3523

PR #3679, opened 2022-02-22
#3523: Raise syntax error when rest elt is not last

0xe:
**Description:**

Raises a syntax error when the rest element is not the last element.

For example,

```
var rest, b;

for ({...rest, b} of [{}
]) ;
```

will raise a syntax error like so:

```
error: Rest element is not the last
 --> <anon>:3:7
  |
3 | for ({...rest, b} of [{}
  |       ^^^^^^^
```

**BREAKING CHANGE:**

https://github.com/swc-project/swc/issues/3523


0xe:
closing this as it's dup of https://github.com/swc-project/swc/pull/3675

PR #3682, opened 2022-02-22
refactor: an plus b syntax

alexander-akait:
**Description:**

Refactor+ features:
- Refactor AST of `an+b` 
- Parse `:nth(2n + 1 of <selector_list>) {}`
- Parse `:not(<selector_list>) {}` selectors

**BREAKING CHANGE:**

Yes

No


crates/swc_css_ast/src/selector.rs
```diff
@@ -197,34 +197,32 @@ pub struct PseudoElementSelector {
...
 
-#[ast_node("Nth")]
-pub struct Nth {
-    pub span: Span,
-    pub nth: NthValue,
-    pub selector_list: Option<SelectorList>,
+#[ast_node]
+pub enum AnPlusB {
+    #[tag("Ident")]
+    Ident(Ident),
```
kdy1, line 216: Why is `Ident` variant required for `2n + 1` ?

kdy1 (changes requested):
Seems like CI for `swc_css_parser` is failing.

crates/swc_css_ast/src/selector.rs
alexander-akait: `An+B` can `even`/`odd` - https://www.w3.org/TR/css-syntax-3/#the-anb-type, so it can be as notation and can be ident

alexander-akait:
@kdy1 Fixed, forgot to send a test file :smile:

crates/swc_css_ast/src/selector.rs
```diff
@@ -197,34 +197,32 @@ pub struct PseudoElementSelector {
 
 #[ast_node]
 pub enum PseudoSelectorChildren {
-    #[tag("Nth")]
-    Nth(Nth),
+    #[tag("AnPlusB")]
+    AnPlusB(AnPlusB),
```
kdy1, line 201: Isn't `AnPlusBNotation` better here? The current type definition will have multiple representations of a single value.


In this case, it's `PseudoSelectorChildren::Ident` and `PseudoSelectorChildren::AnPlusB(AnPlusB::Ident)`

kdy1: I found that this is quite annoying while working on es transforms, due to `PatOrExpr`

kdy1 (changes requested):
Other codes looks fine, but I want to reduce mistakes

(I made lots of mistakes for ES ast because I didn't know anything)

crates/swc_css_ast/src/selector.rs
alexander-akait: @kdy1 Yep, I am thinking about it too, but here interesting case:
- `PseudoSelectorChildren::Ident` will have another values, for example `p:lang(en) {}`, `en` will be it
- `PseudoSelectorChildren::AnPlusB(AnPlusB::Ident)` is `even`/`odd`

It is not different in AST, but it is different on parser/visitor side, so if I want to visit all `AnPlusB` nodes (i.e. `Ident` or `AnPlusBNotation`), I will use `visit_an_plus_b` (same for `VisitMut`), you can see it in our minifier.

Also if somebody want to parse `an+b` notation, this will be used:
```rs
let an_plus_b: AnPlusB = parse_str(&"even", ParserConfig { ..Default::default() }, &mut errors).unwrap();
```

That is why we have two idents here.

kdy1 (approved):
LGTM. Thank you

---

swc-bump:
 - swc_css_ast --breaking

PR #3683, opened 2022-02-22
build(cargo): use lld for win x64 build

kwonoj:
**Description:**

This PR attempts to try new linker for the builds. As first step, replaces x64 windows build to use lld. Instead of using separate llvm installation for lld which can be verbose to setup this uses out-of-the box lld comes with rustc.

Consider this as an expriment, this may not improve performance drastically. Guess we can try & see.

**BREAKING CHANGE:**

- https://github.com/swc-project/swc/discussions/3637

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -49,7 +49,11 @@ rustflags = [
   "new-llvm-pass-manager=no",
 ]
 
+[target.x86_64-pc-windows-msvc]
+linker = "rust-lld"
+
 [target.aarch64-pc-windows-msvc]
+linker = "rust-lld"
 rustflags = [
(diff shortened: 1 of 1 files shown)


kwonoj:
note i686 doesn't do this - it fails to build.

kdy1 (approved):
swc-bump:
 - ''

yisibl:
Is zld helpful for swc? https://github.com/bevyengine/bevy/pull/312

kwonoj:
Zld and other (mold for example for linux) is in consideration, just couldn't spend time to setup CI & make it required for local dev. If there's poc shows meaningful gain we can definitely try.

PR #3684, opened 2022-02-22
build(package): expose new swc_cli as binary

kwonoj:
**Description:**
Actually expose new cli implementation when install `@swc/core`. 

Note I choose binary name to `swcx`. It doesn't have strong meaning (ne`x`t? e`x`perimental?), only to avoid conflict with binary name installed via existing `@swc/cli`. Once this one is matured enough, we can plan migrate & depreacte existing `@swc/cli`.


scripts/npm/android-arm64/package.json
```diff
@@ -12,6 +12,9 @@
         "swc.android-arm64.node",
         "swc"
     ],
+    "bin": {
```
kwonoj, line 15: this actually brings me a question: where do we use binary for `android`? and do we have to support /cli for this target?

kdy1: Hmm... I'm not sure. Maybe termux?

cc @Brooooooklyn Do you have any idea?

Brooooooklyn: Yes, seems like there are many termux users

kwonoj: That.. is surprising. I'm curious if we can reduce some of matrix including this target. If we say many, do we know how many are there?

kwonoj: (either cases, this PR won't deprecate anything yet)

kdy1 (approved):
LGTM. Thank you!

---

swc-bump:
 - ''

PR #3695, opened 2022-02-23
 fix(ecma/transform/compat): private getter/setter

### LINKED ISSUE #3618: Pirvate getter setter are wrongly transformed
### Describe the bug

swc would treat getter and setter as normal method, and generate two seperate weakset

### Input code

```typescript
class MyClass {
  get #a() {}
  set #a(x) {}
}
```

### Config

_No response_

### Playground link

https://play.swc.rs/?version=1.2.141&code=H4sIAAAAAAAAA0vOSSwuVvCtdAbT1VwKCumpJQrKiRqaCtW1QF4xhFcB5tYCAAb9JAAuAAAA&config=H4sIAAAAAAAAA0WOuw6DMAxFf6XyzEA7UIm5Kx9hBUOpyEO2qRqh%2FHsTHmKzrs899gofMdCuEJCFuEwSneIPWiBjUQxPQaHKWI6UF0oVKKOTwbMt%2BEwjmvgi4xnV8w5V0J9BR4o9KkI74CxbHXkkLQfkUd%2BbLJ%2B9FzqACuzkpiEWt%2FE2MIlcK3TjfJIpu6zvlxKsoDHQ5mwgXY6jN0l3gPt35L5bZ%2Fsk68G82Vu6PWtI6Q%2BymPZHFAEAAA%3D%3D

### Expected behavior

Only generate one WeakMap

### Actual behavior

_No response_

### Version

1.2.141

### Additional context

_No response_

Austaras:
**Description:**

**BREAKING CHANGE:**
would require a new version of `@swc/helpers`

closes #3618


Austaras:
I copied blindly from babel, which implementation is inefficient and [wrong](https://babeljs.io/repl#?browsers=chrome%2070&build=&builtIns=false&corejs=false&spec=false&loose=false&code_lz=MYGwhgzhAECC0G8BQ1oQKYBdoGIwAoAPASkRVWmAHsA7CKkdAOhCoHMjjyBfc6NrLgKlkqXqnI4ARtAC80AIwAmAMw0k5AGZUq-EX2iYAFgEsITPAGpLB68bMWpBg9ToNmrDvfPSuYpLxINOgA7nB6TNq6xEA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=true&targets=&version=7.17.6&externalPlugins=&assumptions=%7B%7D), I will start all over again.

Austaras:
This would be the final blockade before loose mode for class properties

kdy1 (commented):
Note: I didn't review it fully yet. Just to ensure that I don't accidentally break downstream crates.

---

swc-bump:
 - swc_ecma_parser --breaking

crates/swc_ecma_transforms_compat/src/es2022/class_properties/mod.rs
```diff
@@ -516,33 +523,10 @@ impl ClassProperties {
...
-                                args: vec![
-                                    class_ident.clone().as_arg(),
-                                    key.as_arg(),
-                                    value.as_arg(),
-                                ],
-                                type_args: Default::default(),
-                            }
-                            .into_stmt(),
-                        )
+                        extra_inits.insert(key, MemberInit::Prop(prop_span, value));
```
kdy1, line 527: Is the type of `key` `Expr?

kdy1: I think it's generally not safe to use it as a hash key.

crates/swc_ecma_transforms_compat/src/es2022/class_properties/mod.rs
```diff
@@ -384,20 +394,17 @@ impl ClassProperties {
 
         let has_super = class.super_class.is_some();
 
-        let mut typescript_constructor_properties = vec![];
-
-        let mut constructor_exprs = vec![];
+        let mut constructor_inits = MemberInitMap::default();
         let mut vars = vec![];
-        let mut extra_stmts = vec![];
+        let mut extra_inits = MemberInitMap::default();
```
kdy1, line 399: Can you add some comment about the reason we use an indexed hashmap here? Is a vector not enough?

kdy1 (changes requested):

crates/swc_ecma_transforms_compat/src/es2022/class_properties/mod.rs
Austaras: This is indeed problemtic, but since public proerties name could be str/num/bigint, I don't find better way.

kdy1: Yeah, I had thought about it and I concluded that it's fine because we parse `NaN` as `Expr::Ident`.
So this is not relevant at the moment of this transform is applied.

Austaras: We also need to preserve evaluation order of member init

crates/swc_ecma_transforms_compat/src/es2022/class_properties/mod.rs
Austaras: added

kdy1: Thanks!

kdy1 (approved):
Thank you so much!

Austaras:
@kdy1 you seems like to public `@swc/helper` before this pr merge

kdy1:
I published twice, once due to https://github.com/swc-project/swc/commit/72c9e6ca2b8186c4b297658bec1ce37912898d9f and once due to this PR

PR #3701, opened 2022-02-23
fix(parser): fix span of export declaration containing const enum

dsherret:
**Description:**

Fixes span of export declaration containing const enum.


kdy1 (approved):
Thank you!

---


swc-bump:
 - swc_ecma_parser

PR #3704, opened 2022-02-23
feature(lazy): Allow more lazy configurations, enabling local and ext…

gmalette:
# Description

(edited from Discord)

We've been liking lazy module loading a lot (it cut down or load times by over 30%), but I was wondering if we can push this even further.
Right now, there are 3 possible values for lazy:
- `false`: no lazy-loading at all
- `true`: external dependencies are lazy-loaded (not prefixed by `./` ?)
- `Array<string>`: Allow-list of paths that can be lazy loaded.

Our application contains many different packages that are not all required to perform any given task. At this time, we're running on `true`. In addition to having all external deps loaded lazily, we'd like either all local deps be lazy loaded, or an allow-list. 

With this change, I was thinking adding a new possible config value:

```
{
  all_external: bool,
  all_local: bool,
  allowed: Array<string>,
}
```

This create an ambiguity if `all_external: true, all_local: true, allowed: not_empty_list`, but for now it's the best I can think of.

# Questions

- [ ] How can I test these changes on my app before merging, to make sure it produces a statistically significant improvement?
- [ ] Do I need to update the documentation and/or other repos such as the npm package?


kwonoj:
I'd like to hear some more details around what exactly this proposes. Am I understanding correct this controls when to actually call cjs require per given import? If that's correct for me this sounds like somewhat micro optimization probably fits better for the plugin, or other userland ecosystem. Do we have similar precedents other tools provides in its core level?

gmalette:
> I'd like to hear some more details around what exactly this proposes

It adds more thoroughness to the cjs configuration with regards to lazy loading (in the `.swcrc` file). Right now you can only specify lazy loading either:
- all external dependencies (`module.lazy = true`)
- a list of dependencies (`module.lazy = ["path/to/file", "path/to/other_file"]`)

We very well could provide a list of all dependencies (local and external) for our app, but that would be tedious and error-prone. 

With these changes, at least we'd be able to specify that all external dependencies should be lazy-loaded, and then provide a list of local dependencies eligible for lazy loading too. 

My current hypothesis is that, barring circular dependencies issues, lazy-loading **all** dependencies will produce significant performance gains for many node projects.

> probably fits better for the plugin, or other userland ecosystem

I can definitely try a plugin, but I'd be curious what sets apart the currently-allowed configuration for `lazy` and the ones I'm proposing.

kwonoj:
> barring circular dependencies issues, lazy-loading all dependencies will produce significant performance gains for many node projects.

At my previous work, there was huge circular dependency chain in its import and just making lazy load everything brought up interesting side effect with failures. Lazy loading or other mechanics are possible way to workaround some of those, but not a definitive answer as well. 

Either cases, I guess that's somewhat arbitrary topic to discuss to this PR. Primary thing to figure out is if we want to this in core or not.

If you don't mind, would you give it a try with new experimental plugin? If you can make this work with plugin, it actually helps to answer question `How can I test these changes on my app before merging, to make sure it produces a statistically significant improvement?` : You can try with plugin / not to measure numbers with baseline `@swc/core`.

gmalette:
> try with new experimental plugin

Can def try! Can you point me towards some literature on how to use/write those new plugins?

kwonoj:
https://github.com/swc-project/swc/discussions/3540 would be a starting point.

gmalette:
Thanks I'll try that. In the meantime would you be open to this PR but without `all_local`? Essentially giving us the benefit of `module.lazy = true + module.lazy = Array<string>`

kwonoj:
I'm bit hesitant to open public api surface without deeper usecase consideration. I didn't deeply think about lazy option itself - probably discuss with @kdy1 what we'll actually want.

crates/swc_ecma_transforms_module/src/util.rs
```diff
@@ -62,11 +62,20 @@ const fn default_strict_mode() -> bool {
     true
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+pub struct LazyObjectConfig {
+    pub all_external: bool,
+    pub all_local: bool,
```
kdy1, line 69: Sorry, I missed the discord message because I was too busy.

kdy1: I think regex is enough, isn't it?
All local imports starts with './' or '../.

kdy1 (commented):
Thank you for trying to implement it yourself :P

crates/swc_ecma_transforms_module/src/util.rs
gmalette: Yes, a regex would work. Do you propose adding a new form of config, or reusing `Array<string>` to parse them as regexps? Reusing it could pose problems; I don't know how to differentiate between a regexp and plain strings

kdy1: I think `Lazy::Object(LazyObjectConfig)` where `LazyObjectConfig` has a field like 'pattern' will work

gmalette: Right yeah. A single “pattern” field? I’ll get right on that

gmalette: @kdy1 is this what you had in mind?

crates/swc_ecma_transforms_module/src/util.rs
```diff
@@ -59,11 +59,19 @@ const fn default_strict_mode() -> bool {
     true
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+pub struct LazyObjectConfig {
+    #[serde(with = "serde_regex")]
+    pub patterns: Vec<regex::Regex>,
```
gmalette, line 66: @kdy1 is this what you had in mind?

kdy1: Oh, sorry seems like I missed the email.

I think we should accept `Vec<JsWord>` and cache the regex based on the given string.

gmalette: Just making sure: you mean `patterns: Vec<JsWord>` instead of `patterns: Vec<regex::Regex>`? And then cache the transformation from `JsWord` to `Regex`?

If that's what you mean, are you fine with changing the failure mode from parse-time to run-time?

kdy1: Yes, but we don't need to change it to parse time.
I didn't get what you mean.

We create modules pass anyway (at parse time), and we can pass cached regex to it.

gmalette: What I meant is that right now, an error in the regex pattern in the `.swcrc` file would be caught at parse-time, when the file is parsed, because it's using `serde_regex`. If we allow `JsWord`in the config, and transform the `JsWord` into a regex while evaluating `is_lazy()`, it's switching the error to a different phase, at run-time.

> We create modules pass anyway (at parse time), and we can pass cached regex to it.

Oh! Can you show me how/where to do that? I thought your proposition was to replace it with something like:

```rust
pub struct LazyObjectConfig {
    patterns: Vec<JsWord>,
    regexes: Option<Vec<Regex>>,
}

impl LazyObjectConfig {
    // not sure exactly how to do this because is_lazy() doesn't receive `mut &self`, just `&self`.
    pub fn matches(&mut self, path: &str) -> Result<bool> {
        let regexes = self.regexes.get_or_insert_with(|| {
            let r = vec![];
            for pat in self.patterns {
                r.push(Regex::new(pat)?);
            }
            r
        })?;
    }
}
```

kdy1: No, it's not runtime vs parse time.
We have to convert it to `Arc<Regex>>`, because we have to pass it to module pass like common js.
I think you mean passing it to `common_js` pass as a `Vec<JsWord>`, but that's not what I mean.

gmalette: I don't follow what you're trying to say then. Where would you convert from `Vec<JsWord` to `Arc<Regex>`?

I'm sorry, I really don't have the full overview of the project, so I may be missing some important parts.

kdy1: While creating the config instance for module transform.

https://docs.rs/swc_ecma_transforms_module/0.84.2/swc_ecma_transforms_module/util/struct.Config.html

Especially, the `Lazy` type should be extended to accept `Vec<Arc<Regex>>` variant.

gmalette:
@kdy1 this PR is now using `CachedRegex`

kdy1 (approved):
LGTM. Thank you!

---

swc-bump:

PR #3707, opened 2022-02-23
feat: compress alpha value

alexander-akait:
**Description:**

part of https://github.com/cssnano/cssnano/tree/master/packages/postcss-convert-values

Ideally we should introduce `<alpha-value>` type, but I want to do it when we will implement color minification (in near future)

**BREAKING CHANGE:**

No

No


kdy1 (approved):
LGTM. Thanks!

---

swc-bump:
 - swc_css_minifier

PR #3722, opened 2022-02-24
feat(es/transforms/compat): loose mode for class_properties

Austaras:
**Description:**

**BREAKING CHANGE:**

would be break change for rust user


crates/swc/tests/tsc-references/privateNamesUnique-3_es5.1.normal.js
```diff
@@ -59,7 +59,7 @@ var A = function A() {
 var _foo = {
     writable: true,
     value: true
-} // error (duplicate)
+}// error (duplicate)
```
Austaras, line 62: I have no actual idea why this is happening, but I guess it's fine.

crates/swc_ecma_transforms_compat/src/es2022/class_properties/mod.rs
```diff
@@ -630,42 +656,78 @@ impl ClassProperties {
...
                         }
-                        (MethodKind::Method, true) => None,
+                        (MethodKind::Method, true) => {
+                            if self.c.private_as_properties {
+                                extra_inits.push(MemberInit::PrivMethod(PrivMethod {
+                                    span: prop_span,
+                                    name: weak_coll_var.clone(),
+                                    fn_name: fn_name.clone(),
+                                }));
+                                Some(Ident::dummy())
```
kdy1, line 701: Is this correct?

kdy1: Nevermind. I found that `private_as_properties` is used to check

kdy1 (changes requested):
Oh... I found that CI failed

Austaras:
oh, my bad

Austaras:
This would also require a new version of `@swc/helpers`

kdy1 (approved):
Thank you!

PR #3726, opened 2022-02-24
perf(es/parser): Reduce usage of generics to reduce binary size

kdy1:
**Description:**

swc_ecma_parser:
 - Reduce usage of generics.

---
t
---

# Binary size of `main`

- 54367296 bytes

# Binary size after 04dafe1d4ebaa42543260848fc68123a9f46bc1d

 - 54206672 bytes


kdy1 (commented):
swc-bump:
 - swc_ecma_parser

PR #3729, opened 2022-02-24
Feat parse more selectors

alexander-akait:
**Description:**

feat - parse `:is`/`:where`/`:matches`/`:has` (old legacy versions) + selectors for WebRTC + selectors for custom elements + lang/dir. 

These are almost all selectors known and in spec pseudo's at the moment.

If we can parse them - we can handle them and minify.

Also:
- fix codegen for combinators (we lose spaces in normal output)
- improve minify
- many tests

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_css_ast --breaking

Issue #3071, opened 2021-12-19
[Proposal] Support event tracing for profiling


kwonoj:
### Discussed in https://github.com/swc-project/swc/discussions/3065

<div type='discussions-op-text'>

<sup>Originally posted by **kwonoj** December 18, 2021</sup>
| Status  | Last Updated | Target version |
| ------------- | ------------- | ------------- |
| Accepted  | Dec.19.2021  | TBD |

# Summary

Expose a new flag, `--trace` or similar to emit profiling trace for SWC's internal behavior as trace event format.

# Motivation

Performance is one of the notable characteristics of SWC. While SWC itself needs to deliver consistent peak performance, there will be problems from end user's codebases. By providing trace capabilities from SWC's core itself user can have detailed diagnostics to their own problem, also it can be used for the tracing for observability.

Slack for example heavily utilizes chrome's tracing feature for internal perf analysis (https://slack.engineering/chrome-tracing-for-fun-and-profit/) for their desktop client to solve real world user performance issue cannot be reproducible internally.

## Explainer

[Trace Event format](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview) is JSON representation of trace event data can be processed via various tools, including trace viewer. Chrome uses it for its trace profiling (https://www.chromium.org/developers/how-tos/trace-event-profiling-tool) as well as other tools like Bazel (https://docs.bazel.build/versions/main/skylark/performance.html). Chrome has an embedded trace viewer for the visualization (chrome://tracing/).

# Detailed design

Thankfully, there are rust ecosystem allows us to adapt this easily. Moreover, SWC already have `tracing` (https://github.com/tokio-rs/tracing) internally. There is a layer `tracing_chrome` (https://github.com/thoren-d/tracing-chrome) on top of tracing subscriber which generates Event Trace Format JSON output from existing traces.

This PR (https://github.com/swc-project/swc/pull/3066) have quick & dirty poc demo which enables chrome traces with few dummy spans for `transform_sync`. Please refer generated JSON for the output format which can be visualized as below. 

![Screenshot from 2021-12-18 01-30-39](https://user-images.githubusercontent.com/1210596/146636403-bfe824aa-c2e8-4234-acdd-288a72a7d5e1.png)

As a consumer's point of view, I expect these kinds of pseudo workflow to generate traces for a given operation.

```
// run @swc/cli
npx swc ./file.js ...options --trace categories1, categories2
[...]

PR #3066, opened 2021-12-18
[DONOTMERGE]: chore(demo): trace profiler

kwonoj:
**Description:**

This is POC demo for https://github.com/swc-project/swc/discussions/3065 , should not be merged.

PR #3731, opened 2022-02-24
feat(node): experimental trace support in `@swc/core`

kwonoj:
**Description:**

This PR exposes new experimental interface `__experimental_registerGlobalTraceConfig` in `@swc/core`, allows generating event trace format based trace output for performance diagnose. Rust / node based CLI doesn't expose this flags yet.

The idea is pretty much same as https://github.com/swc-project/swc/pull/3066, PR adds some polishing & internal structures. Exposed interface itself may need some iteration, however it is exposed as `__experimental` for now. PR implements trace spans for `es2015::arrow`, generated trace confirms it is correct span with correct module (via `category`).

image

Notable changes in this PR is https://github.com/swc-project/swc/commit/40a96aeccf76632bf3b29a9e7895bf74398f4b0d. I did add instrumentation spans for all of default `visit_*`, `fold_*` via `swc_ecma_visit_macro` and then back it out. Codewise it worked, but in practical it creates too verbose traces makes trace output unusable. For example, with macro support running transform over 150 file created trace larger than **700MB**. It is nearly not possible to load on trace viewer as well.

In result, proposed approach in here is https://github.com/swc-project/swc/commit/9b2adee39e0cccfe5b3e66543b371a4619d72306

- for overridden visitor fn, add `#[instrument]` attr
- that's it!

Only thing to note is I choose to use `skip_all` to not to include fn param into trace. Those are typically AST nodes, large enough to increase trace size. 

- https://github.com/swc-project/swc/issues/3071


crates/node/src/print.rs
```diff
@@ -77,6 +79,8 @@ pub fn print(
 
 #[napi]
 pub fn print_sync(program: String, options: Buffer) -> napi::Result<TransformOutput> {
+    crate::util::init_trace_once(false, None)?;
```
kwonoj, line 82: This repeated call is to preserve existing behavior to support `ENV_LOG` without breaking changes.

I hope to declare breaking changes to this eventually, by letting users explicitly opt in by calling trace registration.

crates/node/src/util.rs
```diff
@@ -1,14 +1,68 @@
...
+use tracing_subscriber::{
+    prelude::__tracing_subscriber_SubscriberExt, util::SubscriberInitExt, EnvFilter,
+};
+
+// Hold FlushGuard for the trace until it flushes out completely.
+#[derive(Default)]
+struct TraceFlushGuard {
+    inner: Option<FlushGuard>,
+}
+static FLUSH_GUARD: OnceCell<Mutex<TraceFlushGuard>> = OnceCell::new();
```
kwonoj, line 27: I'm curious if there's better way to hold `FlushGuard` until a whole process lifecycle completes. We can't let this be dropped in `_register` or individual `transform*` interfaces.

crates/swc_ecma_transforms_compat/src/es2015/arrow.rs
```diff
@@ -69,6 +70,7 @@ struct Arrow {
 impl VisitMut for Arrow {
     noop_visit_mut_type!();
 
+    #[tracing::instrument(skip_all)]
```
kwonoj, line 73: Instead of manually creating span or use span! macro with hardcoded name, we'll use `#instrument` attr in general. We can always use span! or manual span for the cases we need additional customization but as a first step decorating all of our overridden visitor with default span would makes sense.

kdy1 (commented):
I think this can make swc marginally slower.

v1.2.145: https://github.com/swc-project/swc/runs/5316958723?check_suite_focus=true

Current branch: https://github.com/swc-project/swc/runs/5325695105?check_suite_focus=true

kwonoj:
Yes, there's overhead to enable tracings on runtime binary. That is also noted in the original proposal as well. It is not uncommon though, large runtimes like chromium enables trace by default as it gives rich data for the performances on each user's machine. Overhead is a cost for those tradeoffs

The other alternative is not user-friendly in my opinion - either users have to opt in a specific binary, or run a native profiler which involves non-trivial setup for normal users. That's where I came to think we may be able to bite those cost.

kwonoj:
Also per https://github.com/tokio-rs/tracing/issues/1070, it looks like explicitly opt in specific trace level reduces those overhead. Updated PR to use `trace` level only.

crates/node/src/util.rs
RiESAEX: do we really need to hold gurad ? I remember I even didn't bind a ident to it

kwonoj: Good q, per doc https://github.com/tokio-rs/tracing/blob/4365b3971b617784ca44a7be1f05358b2990056d/tracing-flame/src/lib.rs#L208 it says it'll flush writer when it's dropped.

The way I think is, napi-exported fns are not binded to process lifecycle and `trace_once_init()` will drop this guard if we don't hold it. At least, when I tried https://github.com/swc-project/swc/pull/3066 I confirmed it requires manual flush if I did not hold guard.

I think we need to ensure this won't be dropped in any way - just matter of how.

kdy1 (approved):
LGTM.
Thank you!

---

swc-bump:
 - swc_ecma_transforms_compat

PR #3732, opened 2022-02-24
fix(swc/plugin_macro): do not dupe free guest memory

kwonoj:
**Description:**

Peeking execution path, I realized I left over `_volatile` fn in macro when I updated return value logic from multivalue to calling imported fn from host. This change removed logic host allocates byte for returned value that guest no longer need manually free, instead rust will drop allocated serialized bytes accordingly.

https://github.com/swc-project/swc/issues/3598#issuecomment-1049252389


kdy1 (approved):
LGTM.

---

swc-bump:
 - swc_plugin_macro

Issue #3432, opened 2022-02-02
Fatal error `thread <unnamed> panicked` when attempting a shorthand property initializer


edwincoronado:
### Describe the bug

As part of some testing, when going through the [test262 suite](https://github.com/tc39/test262) and most specifically test case [cover-initialized-name.js](https://github.com/tc39/test262/blob/main/test/language/expressions/object/cover-initialized-name.js), SWC is crashing with the following error:

```log
thread '<unnamed>' panicked at 'internal error: entered unreachable code: assign property in object literal is invalid', crates/swc_ecma_transforms_compat/src/es2015/duplicate_keys.rs:66:33
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
failed to handle: internal error: entered unreachable code: assign property in object literal is invalid
Error: Failed to compile 1 file with swc.
    at Object.assertCompilationResult (/Users/edwin.coronado/Desktop/tmp/node_modules/@swc/cli/lib/swc/util.js:115:15)
    at files (/Users/edwin.coronado/Desktop/tmp/node_modules/@swc/cli/lib/swc/file.js:175:18)
    at async _default (/Users/edwin.coronado/Desktop/tmp/node_modules/@swc/cli/lib/swc/file.js:194:9)
```

Detailed log when using `RUST_BACKTRACE=1 SWC_DEBUG=1 npx swc --sync <file>`:
```log
Panic: PanicInfo { payload: Any { .. }, message: Some(internal error: entered unreachable code: assign property in object literal is invalid), location: Location { file: "crates/swc_ecma_transforms_compat/src/es2015/duplicate_keys.rs", line: 66, col: 33 } }
Backtrace:    0: <unknown>
   1: <unknown>
   2: <unknown>
   3: _rust_eh_personality
   4: _rust_eh_personality
   5: _rust_eh_personality
   6: _rust_eh_personality
   7: _wasmer_vm_f64_nearest
   8: _rust_eh_personality
   9: _rust_eh_personality
  10: _rust_eh_personality
  11: _rust_eh_personality
  12: _rust_eh_personality
  13: _rust_eh_personality
  14: <unknown>
  15: __ZN6v8impl12_GLOBAL__N_123FunctionCallbackWrapper6InvokeERKN2v820FunctionCallbackInfoINS2_5ValueEEE
  16: __ZN2v88internal25FunctionCallbackArguments4CallENS0_15CallHandlerInfoE
[...]

PR #3734, opened 2022-02-25
#3432: Raise syntax error on invalid shorthand assignment

0xe:
**Description:**

Handles invalid shorthand property initializers like `({ a = 1});`
https://github.com/tc39/test262/blob/main/test/language/expressions/object/cover-initialized-name.js

```
error: assignment property is invalid syntax
 --> <anon>:3:4
  |
3 | ({ a = 1 });
  |    ^
```

**BREAKING CHANGE:**

Closes https://github.com/swc-project/swc/issues/3432


crates/swc_ecma_parser/src/parser/object.rs
```diff
@@ -248,6 +248,7 @@ impl<I: Tokens> ParseObject<Box<Expr>> for Parser<I> {
 
             if eat!(self, '=') {
                 let value = self.include_in_expr(true).parse_assignment_expr()?;
+
```
0xe, line 251: will remove.

crates/swc_ecma_parser/src/parser/expr.rs
```diff
@@ -785,6 +785,23 @@ impl<'a, I: Tokens> Parser<I> {
                 }
             }
             return Ok(Box::new(Expr::Arrow(arrow_expr)));
+        } else {
+            // If there's no arrow function, we have to check there's no
+            // AssignProp in lhs to check against assignment in object literals
+            // like (a, {b = 1});
+            for expr_or_spread in paren_items.iter().cloned() {
```
kdy1, line 792: `cloned()` is bad for performance, as it clones the whole content.

crates/swc_ecma_parser/src/parser/expr.rs
```diff
@@ -785,6 +785,23 @@ impl<'a, I: Tokens> Parser<I> {
...
+                if let PatOrExprOrSpread::ExprOrSpread(e) = expr_or_spread {
+                    if let Expr::Object(o) = *e.expr {
+                        for p in o.props.into_iter() {
```
kdy1, line 795: I think `iter()` is enough here.

`Vec::iter` is used to iterate over a reference to vector.

kdy1: As it does not require ownership of the vector, you can remove `cloned()` above.

kdy1 (changes requested):
Thank you!

crates/swc_ecma_parser/src/parser/expr.rs
0xe: Thanks. I've removed cloned() and i'm accessing through reference.

PR #3740, opened 2022-02-25
feat: parse general enclosed

alexander-akait:
**Description:**

feat parse - https://www.w3.org/TR/css-conditional-3/#at-supports and https://www.w3.org/TR/css-conditional-4/#at-supports-ext

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_css_ast --breaking

Issue #3730, opened 2022-02-24
SWC doesn't enforce (TS1155): 'const' declarations must be initialized.


spencersaxey:
### Describe the bug

The given input code, `swc` compiles, and produces this output:
```
"use strict";
const why;
```
`tsc`, on the other hand, produces this error: 
error TS1155: 'const' declarations must be initialized.

### Input code

```typescript
const why;
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "decorators": true,
      "tsx": false
    },
    "transform": {
      "legacyDecorator": true,
      "decoratorMetadata": true
    },
    "target": "es2022",
    "loose": false,
    "externalHelpers": false,
    "keepClassNames": true
  },
  "module": {
    "type": "commonjs",
    "strict": true,
    "strictMode": true,
    "lazy": false,
    "noInterop": false
  },
  "isModule": true,
  "minify": false
}
```

### Playground link

https://play.swc.rs/?version=1.2.144&code=H4sIAAAAAAAAA0vOzysuUSjPqLTmAgCplY8ZCwAAAA%3D%3D&config=H4sIAAAAAAAAA0WPvW7DMAyE30Wzh8Kj13ZoB%2FcdCPkcKJFEgWSAuEHevRJsORv%2F7rvj013Vu%2BnpColCWqVbNnq4ydlWoF5CMTe4BZ6FjEXdZHLH4Ezr0UpR8aqNUNaVJTVCxIX89tUVXXAiZhgtZLQvmprkAquW0PFjHKtdZFYc%2BMHhYZBM8RuxoCU45jegfEZS%2FaUEPXGJl3tES9JeqFjPKXG%2BagWrSfDWI%2B3dzAv6JNLfdvIz%2F%2BTqzOX9aND5oO%2F3KeSwdsXrH6k%2FedRPAQAA

### Expected behavior

Expected this code not to compile, with an output similar to: `error TS1155: 'const' declarations must be initialized.`

### Actual behavior

The code compiles without errors.

### Version

1.2.144

PR #3742, opened 2022-02-25
fix(parser) Throw syntax error when const not inititialized in strict mode

williamtetlow:
**Description:**
fixes: https://github.com/swc-project/swc/issues/3730

Throw syntax error when const not initialized for Ecmascript and Typescript

**BREAKING CHANGE:**

https://github.com/swc-project/swc/issues/3730


crates/swc_ecma_parser/src/parser/stmt.rs
```diff
@@ -799,6 +803,14 @@ impl<'a, I: Tokens> Parser<I> {
                 // typescript allows `declare` vars not to have initializers.
                 if self.ctx().in_declare {
                     None
+                } else if *kind == VarDeclKind::Const && self.ctx().strict {
+                    let syntax_error = if self.syntax().typescript() {
```
williamtetlow, line 807: ❓ Do you normally consolidate errors like this that apply to ES and TS or do you keep them separate?

crates/swc_ecma_codegen/src/stmt.rs
```diff
@@ -42,13 +42,11 @@ mod tests {
     fn declaration_statement() {
         assert_min("var foo;", "var foo");
         assert_min("let foo;", "let foo");
-        assert_min("const foo;", "const foo");
```
williamtetlow, line 45: ❓ I removed these tests because they test against parsing module with strict mode as default which means these are invalid syntax now. I can change this if you want them to still be included?

crates/swc_ecma_parser/src/parser/stmt.rs
kdy1: I prefer consolidating

crates/swc_ecma_codegen/src/stmt.rs
kdy1: It should be removed, thanks!

crates/swc_ecma_parser/src/parser/stmt.rs
```diff
@@ -799,6 +803,14 @@ impl<'a, I: Tokens> Parser<I> {
...
+                        SyntaxError::TS1155
+                    } else {
+                        SyntaxError::ConstDeclarationsRequireInitialization
+                    };
+
+                    syntax_error!(self, span!(self, start), syntax_error)
```
kdy1, line 813: I think we can recover the error here.

`self.emit_err()` can be used to proceed parsing while reporting an error.

kdy1 (changes requested):
Thank you!

crates/swc_ecma_parser/src/parser/stmt.rs
williamtetlow: resolved in [48d5806](https://github.com/swc-project/swc/pull/3742/commits/48d58060388df760ca6904bf91148ee858211565)

crates/swc_ecma_parser/src/parser/stmt.rs
```diff
@@ -748,7 +748,11 @@ impl<'a, I: Tokens> Parser<I> {
         })
     }
 
-    fn parse_var_declarator(&mut self, for_loop: bool) -> PResult<VarDeclarator> {
+    fn parse_var_declarator(
+        &mut self,
+        for_loop: bool,
+        kind: &VarDeclKind,
```
kdy1, line 754: I think we don't need to use reference here, as `VarDeclKind` implements `Copy`.

kdy1 (changes requested):
Thank you!

crates/swc/tests/simple.rs
```diff
@@ -36,6 +36,36 @@ fn compile(src: &str, options: Options) -> String {
         .unwrap()
 }
 
+#[test]
+fn help() {
```
kdy1, line 40: Did you add this intentionally?

kdy1 (changes requested):
Thank you!

crates/swc/tests/simple.rs
williamtetlow: nope sorry that's the test I use for trying out issues locally 😄 rebased

kdy1 (dismissed):
Thank you!

---

swc-bump:
  - swc_ecma_parser

PR #3743, opened 2022-02-25
perf(es/modules): Reduce usage of generics

kdy1:
**Description:**

We don't need to store the code for transforming modules twice.
 

 - https://github.com/swc-project/swc/pull/3741

### DIFF
--- a/crates/swc/src/config/mod.rs
+++ b/crates/swc/src/config/mod.rs
@@ -1660,12 +1660,12 @@ impl Merge for HiddenTransformConfig {
     }
 }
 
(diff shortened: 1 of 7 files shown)


kdy1 (commented):
swc-bump:

kdy1 (commented):
Previously whole module transform code were monomorphized, and resuled in a bloated binary.

I turned on `share-generics` to get the correct size diff, and this patch makes binary (`swc_cli`) 0.3 MB smaller

kwonoj (approved):

PR #3744, opened 2022-02-25
feat(swc/plugin): support trace for plugin execution

kwonoj:
**Description:**

Finally, I can connect the pieces together. 🎉 

One of the major reason I tried to push forward runtime tracing support is to diagnost user-level execution. Plugin is an exact fit, I wanted to have some answers like `swc is being slow cause plugin xxxx taking too long`. I added initial trace for those in this PR.

image

This tells us

- What plugin module takes X time for total execution
- Serialization cost (loading, json serialization for config, context..)
- Actual plugin transformation cost

and we can add more gradually.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,5 @@ pkg/
 
 # Used to see output of babel.
 /lab
+
+*.mm_profdata
--- a/crates/swc/src/plugin.rs
+++ b/crates/swc/src/plugin.rs
@@ -68,6 +68,7 @@ struct RustPlugins {
 }
 
 impl RustPlugins {
+    #[tracing::instrument(level = "trace", skip_all, name = "apply_plugins")]
     #[cfg(feature = "plugin")]
     fn apply(&mut self, n: Program) -> Result<Program, anyhow::Error> {
         use std::{path::PathBuf, sync::Arc};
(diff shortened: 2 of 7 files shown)


crates/swc/src/plugin.rs
```diff
@@ -86,6 +87,13 @@ impl RustPlugins {
         // transform.
         if let Some(plugins) = &self.plugins {
             for p in plugins {
+                let span = tracing::span!(
```
kwonoj, line 90: Crateing manual span to track time to `serialize` json vs. actual plugin transform execution.

kwonoj:
This still doesn't able to explain plugin's internal stage. Since it's running under wasm runtime, I doubt we can have easy way to have integrated traces even.

crates/swc_ecma_transforms_compat/Cargo.toml
```diff
@@ -34,7 +34,7 @@ swc_ecma_transforms_classes = {version = "0.50.0", path = "../swc_ecma_transform
 swc_ecma_transforms_macros = {version = "0.3.0", path = "../swc_ecma_transforms_macros"}
 swc_ecma_utils = {version = "0.68.0", path = "../swc_ecma_utils"}
 swc_ecma_visit = {version = "0.54.0", path = "../swc_ecma_visit"}
-swc_trace_macro = {versio = "0.1.0", path = "../swc_trace_macro"}
+swc_trace_macro = {version = "0.1.0", path = "../swc_trace_macro"}
```
kwonoj, line 37: this seems typo, fixed in this PR.

.gitignore
```diff
@@ -52,4 +52,6 @@ pkg/
 *.tmp.js
 
 # Used to see output of babel.
-/lab
\ No newline at end of file
+/lab
+
+*.mm_profdata
```
kwonoj, line 57: unsure if it's intended but `npm run build:dev` always run `-- -Zself-profile`. Either case I don't expect we'll check in this profile ever.

kdy1 (approved):
LGTM.
Thanks!

---

swc-bump:
 - swc
 - swc_plugin_runner
 - swc_ecma_transforms_compat

PR #3746, opened 2022-02-25
feat(swc/cli): add trace configuration for compile

kwonoj:
**Description:**

This PR enables trace for the cli as well, via `--experimental-trace` boolean flag.

- https://github.com/swc-project/swc/issues/3071

### DIFF
--- a/crates/swc_cli/src/commands/compile.rs
+++ b/crates/swc_cli/src/commands/compile.rs
@@ -15,6 +15,8 @@ use swc::{
     try_with_handler, Compiler, TransformOutput,
 };
 use swc_common::{sync::Lazy, FileName, FilePathMapping, SourceMap};
+use tracing_chrome::{ChromeLayerBuilder, FlushGuard};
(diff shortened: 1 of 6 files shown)


crates/swc_cli/src/commands/compile.rs
```diff
@@ -259,6 +291,8 @@ impl super::CommandRunner for CompileOptions {
         }
 
         if let Some(stdin_input) = stdin_input {
+            let span = tracing::span!(tracing::Level::TRACE, "compile_stdin");
+            let _ = span.enter();
```
kdy1, line 295: This will immediately drop

kdy1 (changes requested):
Thanks!

crates/swc_cli/src/commands/compile.rs
kwonoj: thanks for catch, I forgot to update this along with non-stdin input spans.

PR #3747, opened 2022-02-25
feat(es/parser): Provide more api for reducing binary size

kdy1:
**Description:**

Descriptions are written to the code itself, using rustdoc.

---

 - Start: 24.8 MB
 - After 76c85275a6d6cbaeee46e12dcdcd360f316e0bb5: 22.5MB
 - After 119e7e6343e2fcb7f960ee7e4243729fc551eda0: 21.7MB

### DIFF
--- a/crates/swc/src/config/mod.rs
+++ b/crates/swc/src/config/mod.rs
@@ -42,7 +42,7 @@ use swc_ecma_minifier::option::{
 };
 #[allow(deprecated)]
 pub use swc_ecma_parser::JscTarget;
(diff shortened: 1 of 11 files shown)


kdy1 (commented):
We can't use `share-generics` because it has bad effect on performance.
But actually, there's a way to share generic code - instantiating all generics in a single crate.

This PR applies this technique to reduce the size of the binary.

kwonoj (approved):

kdy1 (commented):
swc-bump:
 - swc
 - swc_ecma_minifier
 - swc_ecma_parser
 - swc_ecma_transforms_base
 - swc_ecma_transforms_module
 - swc_ecma_transforms_react

PR #3752, opened 2022-02-26
fix: integer parsing

alexander-akait:
**Description:**

feat - new type `<integer>` - https://www.w3.org/TR/css-values-4/#integers

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
swc-bump:
 - swc_css_ast --breaking

PR #3753, opened 2022-02-26
refactor(css/ast): simplify code and CSS AST

alexander-akait:
**Description:**

Refactor code - simplify and reuse existing logic

**BREAKING CHANGE:**

No

No


crates/swc_css_ast/src/base.rs
```diff
@@ -52,26 +54,37 @@ pub struct SimpleBlock {
 
 #[ast_node]
 pub enum ComponentValue {
```
kdy1, line 56: Why does we unify every value into this enum?

kdy1 (approved):
LGTM. Thank you!

---

swc-bump:
 - swc_css_ast --breaking

PR #3765, opened 2022-02-27
feat(css/lints): Add CSS linter

g-plane:
**Description:**

This PR created a new crate `swc_css_lints` which is a CSS linter.

I referred [Stylelint](https://stylelint.io/), but `swc_css_lints` won't be a 100% port of Stylelint.

In this PR, I just implemented these rules:

- [block-no-empty](https://stylelint.io/user-guide/rules/list/block-no-empty)
- [at-rule-no-unknown](https://stylelint.io/user-guide/rules/list/at-rule-no-unknown)
- [no-empty-source](https://stylelint.io/user-guide/rules/list/no-empty-source)
- [declaration-no-important](https://stylelint.io/user-guide/rules/list/declaration-no-important)
- [keyframe-declaration-no-important](https://stylelint.io/user-guide/rules/list/keyframe-declaration-no-important)
- [no-invalid-position-at-import-rule](https://stylelint.io/user-guide/rules/list/no-invalid-position-at-import-rule)
- [selector-max-class](https://stylelint.io/user-guide/rules/list/selector-max-class)
- [color-hex-length](https://stylelint.io/user-guide/rules/list/color-hex-length)
- [color-no-invalid-hex](https://stylelint.io/user-guide/rules/list/color-no-invalid-hex)
- [unit-no-unknown](https://stylelint.io/user-guide/rules/list/unit-no-unknown)

I'm not going to implement all rules of Stylelint in this single PR, otherwise it will make the PR hard to be reviewed. Also, I split each rule implementation into a single commit, which should make PR review easier.

The rest rules of Stylelint will be implemented in the future, but I won't implement those ["Stylistic issues"](https://stylelint.io/user-guide/rules/list#stylistic-issues) rules, since they're deprecated.

**BREAKING CHANGE:**

New crate was created: `swc_css_lints`.


alexander-akait (approved):
Awesome, I am still improving AST, so some changes will be here in near future, but that is great work

g-plane:
I hope the parser can support comments. (Certainly, it shouldn't be stored in AST) Because, for example, the `no-empty-source` rule allows comments though there're no actual code, but currently the lexer of CSS parser just simply ignores comments, so we have no way to handle this case.

alexander-akait:
@g-plane 

> I hope the parser can support comments. (Certainly, it shouldn't be stored in AST) Because, for example, the no-empty-source rule allows comments though there're no actual code, but currently the lexer of CSS parser just simply ignores comments, so we have no way to handle this case.

Yep, it's on the roadmap, currently I focused on AST and minifier (and source map generation), then I will improve comments support, because we need it not only for lints

kdy1 (approved):
swc-bump:
 - ''

kdy1 (approved):

PR #3764, opened 2022-02-27
fix(es/typescript): transform `this` in TypeScript class

magic-akari:
**Description:**

**BREAKING CHANGE:**

**Related issue (if exists):**
- fix #3228

PR #3766, opened 2022-02-27
feat: unify class properties transform in ts and js

### LINKED ISSUE #3228: "this" inside arrow in a static class property is compiled to (void 0)
### Describe the bug

Using `this` inside a fat arrow function on a static class property does not work, because this is compiled to `(void 0)`;

### Input code

```typescript
export class StaticTestClass {
    static testProp = 'Hello world!';

    static testMethod = () => {
        console.log(this.testProp);
    }
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "target": "es2015",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": true,
  "env": {
    "targets": "node 12"
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.128&code=H4sIAAAAAAAAA0utKMgvKlFIzkksLlYILkksyUwOSS0ucQbzq7kUgKAYLKpQAhQOKMovULBVUPdIzcnJVyjPL8pJUVS35kJX5ptakpGfAlSooalgawc1BgSS8%2FOK83NS9XLy0zVKMjKL9WCGalqD1dRy1QIAgrT4spEAAAA%3D&config=H4sIAAAAAAAAA0VOSwqDMBC9SnlrF1Xoxjv0EEFHieTHTCyVkLt3IhZ3j%2Fcv2GTCWJAMC3FDcoRsvhiRj0QysU0ZHbIolXmnqtjwSlkdJMOzf6nqYhTCuBgn1MHbYJejdU3RJyaRWzJhdX9n1S4f570R5ZzTTo34GDZBvYuusJX35W5HOlD4nMHzjm4gxJke%2FYBaf%2FJ6D0jYAAAA

### Expected behavior

Expected the `testMethod` method to compile to:
```
StaticTestClass.testMethod = ()=>{
    console.log(StaticTestClass.testProp);
};
```

### Actual behavior

The `testMethod` method compiles to:
```
StaticTestClass.testMethod = ()=>{
    console.log((void 0).testProp);
};
```

### Version

1.2.128

### Additional context

Only seems to happen with typescript in combination with commonjs.

### LINKED ISSUE #3317: Expose `use_define_for_class_fields` in SWC config
### Describe the feature

When compiling a class with class fields using TypeScript syntax, SWC currently down-levels class fields no matter what version of JS is being targeted.

In other words, this:
```js
class Foo {
  foo = 'foo';
}
```

gets turned into this:
```js
class Foo {
    constructor(){
        this.foo = 'foo';
    }
}
```

no matter what config options you use.

However, the capability to avoid this down-leveling [already exists in SWC](https://github.com/swc-project/swc/pull/1478) (and [is even tested!](https://github.com/swc-project/swc/blob/366dc0eec3e19b4e608dc4df8eb728f6d710f9b7/crates/swc_ecma_transforms_typescript/tests/strip.rs#L3711-L3736)) but is unfortunately simply not exposed as a config option in the the SWC config itself.

Could you please provide a way to actually enable this feature in the SWC config?

_No response_

### Additional context

_No response_

### LINKED ISSUE #3389: Mobx - observable can't be applied
### Describe the bug

For mobx, class field need to be defined for that you need [this option](https://www.typescriptlang.org/tsconfig#useDefineForClassFields) in the tsconfig.json :

```json
{
    "compilerOptions": {
        "useDefineForClassFields": true,
    }
}
```

When I try to compile my application with swc I have this error : 

> Error: [MobX] Cannot apply 'observable' to 'User@2.elo': Field not found.

And if we follow what the answer in this [stackoverflow](https://stackoverflow.com/questions/67266810/error-mobx-cannot-apply-observable-to-storeuser-field-not-found) said that I need to use **useDefineForClassFields** which I already use.

### Input code

_No response_

### Config

_No response_

### Playground link

_No response_

### Expected behavior

That I can use **useDefineForClassFields** in my tsconfig

### Actual behavior

_No response_

### Version

1.2.135

### Additional context

_No response_

### LINKED ISSUE #3827: Type only import used in implements is not removed
### Describe the bug

title

### Input code

```typescript
import { Foo } from 'foo'

const a = class B implements Foo {
  
}
```

### Config

_No response_

### Playground link

https://play.swc.rs/?version=1.2.146&code=H4sIAAAAAAAAAx3KMQoAIQwEwD6v2M5X2Fj4DxEFwbhi0ol%2FP7ipZ%2BjmcVxkEg%2F9UBE6GUQqlzkKIuosZkgYumfTttz%2BfgWQ9wGsHqkbQgAAAA%3D%3D&config=H4sIAAAAAAAAAz2PSwrDMAxE76J1Fm0WXeQEvYZw5eDgH5IMMSF3r5yS7obRzBt0wCYOlgMqshAPJT0r7rCA9kriOFSFCVTMUm50TkC7EmeMb4qVWH6%2BRRiz%2BMJpUJjQ6SVa1pDIeNi0JNTg4DRILEXM9RhldJFXsjyQzI%2F5aYMp5OD7ILiSKpPIP5wwr%2FHuDtadve%2Fl08b9uF64oC%2Fb%2FAIU2c5e7AAAAA%3D%3D

### Expected behavior

Foo should be striped

### Actual behavior

_No response_

### Version

1.2.147

### Additional context

_No response_

Austaras:
**Description:**
And also expose an `use_define_for_class_fields` somewhere

**BREAKING CHANGE:**
It will break rust users

It won't be possible to use [[Set]] for pulibc fields and preserve private fields.

It will be a break change for ts users who want es2022/esnext output, but this is aligned with babel behaviour. And consider tsc doesn't have target env and would set `useDefineForPublicFields` to true in es2022, this shall also be aligned with tsc.

closes #3827, closes #3389, closes #3317, closes #3228, supersedes #3764


crates/swc/src/builder.rs
```diff
@@ -180,16 +180,19 @@ impl<'a, 'b, P: swc_ecma_visit::Fold> PassBuilder<'a, 'b, P> {
...
+            let assumptions = self.assumptions;
             Either::Right(chain!(
                 Optional::new(
                     compat::es2022::es2022(compat::es2022::Config {
                         class_properties: compat::es2022::class_properties::Config {
-                            private_as_properties: self.loose,
-                            constant_super: self.loose,
-                            set_public_fields: self.loose,
-                            no_document_all: self.loose
+                            private_as_properties: self.loose
```
konradekk, line 191: wouldn’t it be better to modify `assumptions` according to `loose` setting and pass down only the `assumptions`? (referring also to changes in `swc_ecma_preset_env`; something like this is already done in [config/mod.rs](https://github.com/swc-project/swc/pull/3766/files#diff-05ec577011f6096aac154c266838a96b40e857713f31a3441fc73baf1d90580e)?)

Austaras: Yes, `self.loose` here is redundant

Austaras:
One major difference would be [class exprs](https://play.swc.rs/?version=1.2.144&code=H4sIAAAAAAAAA0vOzysuUUhUsFVIzkksLlZwVKjm5VJQKC5JLMlMBosbGhnzctUCAN%2FM%2FacoAAAA&config=H4sIAAAAAAAAAxXJMQqAMAwF0Lv82UE79jahRFFEQn4GS8ndret7Axcb6kCIHxqoUJa1bFhg4lT%2Fj%2F0JeedFN2Xz02J%2BcNIuNzUzP6WpcwJIAAAA) where ts transform uses a sequence expr while js transform uses a iife.

This should be fixed.

crates/swc/src/config/mod.rs
```diff
@@ -314,18 +314,21 @@ impl Options {
...
+        let mut transform = transform.unwrap_or_default();
+
         // Do a resolver pass before everything.
         //
         // We do this before creating custom passses, so custom passses can use the
         // variable management system based on the syntax contexts.
         if syntax.typescript() {
+            assumptions.set_public_class_fields = !transform.use_define_for_class_fields;
+
+            assumptions.set_class_methods = true;
```
Austaras, line 326: I'm not quite sure if this line should be added. It would generate output more like tsc, but it's different from current behaviour

kdy1: I think it's better to mimic tsc for typescript users.

Austaras:
Ah, I've encountered with microsoft/typescript#45995

Austaras:
Here's some observeration about initialized order of parameter properties:
1. tsc would init parameter properties before normal properties in target before es2022, however in target es2022 and esnext tsc would preserve class properties, so they would be initialized first. As tsc doesn't follow semantic versioning, this may soon changes.
2. babel simply transform parameter properties into assignments and preserve class normal properties, so in any case normal properties would be initialized first
3. esbuild would transform parameter and noraml properties into assignments so the initialized order is preserved, and it doesn't support `useDefineForClassFields`

I originally would like to choose route 3, but consider private properties
```ts
class A {
  a = 123
  #a = 456
  constructor(public b) {
    console.log(this.#a)
  }
}
```
we cannot simply transform `a` into assignments, because it needs to be initialized before `#a`, but only if we bring whole `class_properties` in(and surprisingly, this is not impossible!), private properties could not be safely transformed. So now I choose to turn it into
```js
class A {
 a
  constructor(b) {
    this.b = b
    this.a = 123
    this.#a = 456
    console.log(this.#a)
  }
}
```
This would generate many useless boilerplate, but at least it's consisent and doesn't break current tests. Hopefully not many people would use private properties in TypeScript.

kdy1:
Hmm... Yeah I think your decision is right although I'm a bit confused by other tools.

Austaras:
Oh, and ts has same [bug](https://play.swc.rs/?version=1.2.144&code=H4sIAAAAAAAAA0vOzysuUUjOSSwuTi1WsFWIjuXiSssvUtDISS1RyAQKGFhn2hgCCW1tTYVqLgUgACtWcITyQCA6MxaoMhPMr%2BVCKEot1isoLc7QcNTkAgoTNBfolPycVL2c%2FHSNvNRymBFAw4HaATgF0YGoAAAA&config=H4sIAAAAAAAAA0WNOwrDMBBE7zK1isRFCt0hhxDK2sjox%2B4aIozuHsvYpBvm82bHKh52R3UsxENJy%2Bq%2BsNBWSTyHqugG6nghPWyS6TE9YRBLEYKdXRQySCGHuQ2AL6kyifwjl5d4N%2FvBSuWzDWM%2FP07ma3zcjGsX5H0VlTfqP4fM0UWsAAAA) I fixed in previous pr

crates/swc_ecma_transforms_typescript/tests/strip.rs
```diff
@@ -125,10 +133,14 @@ to!(
 
     abstract #test();
 }",
-    "class test {
-    #test() {
+    "var _test = new WeakSet();
+    class test {
+        constructor(){
+            _classPrivateMethodInit(this, _test);
```
kdy1, line 139: I think this is wrong?
`_test` is not defined.

kdy1 (changes requested):
I didn't review fully yet.
I will do it soon, but I think I found a bug

crates/swc_ecma_transforms_typescript/tests/strip.rs
kdy1: Oh sorry, I missed the `_test` above.

crates/swc/tests/tsc-references/classStaticBlock3_es2015.2.minified.js
```diff
@@ -1,3 +1,3 @@
 class C {
 }
-console.log(C.f1, C.f2, C.f3), console.log(C.f1, C.f2, C.f3), C.f1 = 1, C.f2 = 2, C.f3 = 3;
+C.f1 = 1, console.log(C.f1, C.f2, C.f3), C.f2 = 2, console.log(C.f1, C.f2, C.f3), C.f3 = 3;
```
kdy1, line 3:
```ts
const a = 1;

class C {
    static f1 = 1;

    static {
        console.log(C.f1, C.f2, C.f3)
    }

    static f2 = 2;

    static {
        console.log(C.f1, C.f2, C.f3)
    }

    static f3 = 3;
}

```

and output of original file is

```
1 undefined undefined
1 2 undefined
```

kdy1: Okay, it's correct.
Thanks!

kdy1 (approved):
Thank you so much!


---

swc-bump:

crates/swc/tests/tsc-references/classStaticBlock3_es2015.2.minified.js
Austaras: Yes, the original output is wrong

PR #3767, opened 2022-02-27
fix(ecma/transform/compat): class fields should be initialized last

Austaras:
**Description:**

So that code like
```js
class A {
  #a = this.#b;
  get #b() {}
}
```
would not break
**BREAKING CHANGE:**


kdy1 (approved):
LGTM. Thank you!


---

swc-bump:
 - swc_ecma_transforms_compat

PR #3769, opened 2022-02-27
feat(css/minifier): convert more length

alexander-akait:
**Description:**

feat - compress length, port length logic from https://github.com/cssnano/cssnano/tree/master/packages/postcss-convert-values

We don't convert `q`, because it is not widely supported, so we will implement it after support browserslist in our minifier

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---


swc-bump:
 - swc_css_minifier

PR #3773, opened 2022-02-28
mod(es/transforms/compat): use seq_expr instead of iife for class_expr

Austaras:
**Description:**

And also:
1. don't memorize computed method name if it's a literal
2. fix [this bug](https://play.swc.rs/?version=1.2.144&code=H4sIAAAAAAAAA0vOzysuUUjOSSwuTi1WsFWIjuXiSssvUtDISS1RyAQKGFhn2hgCCW1tTYVqLgUgACtWcITyQCA6MxaoMhPMr%2BVCKEot1isoLc7QcNTkAgoTNBfolPycVL2c%2FHSNvNRymBFAw4HaATgF0YGoAAAA&config=H4sIAAAAAAAAA0WNOwrDMBBE7zK1isRFCt0hh1iUtVHQj10ZIoTunsjYpBvm86bjrQ62o5Aoy1TaUqUPLNhFUie%2BVAyDSrJxnbYut%2BUOg5CzMuxKQdkg%2BuTXNgEuxyKs%2Bo8obeFqjh8r5tc%2BjY7aCh%2FMx%2Fy4GOfO6%2FMsVtl5fAGaNFAarAAAAA%3D%3D). This would require change cached computed keys variable from `var` to `let`, and it would cause thousands of snapshot changes

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_transforms_compat/src/es2022/class_properties/mod.rs
+++ b/crates/swc_ecma_transforms_compat/src/es2022/class_properties/mod.rs
@@ -11,8 +11,8 @@ use swc_ecma_transforms_classes::super_field::SuperFieldAccessFolder;
 use swc_ecma_transforms_macros::fast_path;
 use swc_ecma_utils::{
     alias_ident_for, alias_if_required, constructor::inject_after_super, default_constructor,
-    is_literal, prepend, private_ident, quote_ident, undefined, ExprFactory, ModuleItemLike,
-    StmtLike, HANDLER,
+    is_literal, prepend, private_ident, quote_ident, replace_ident, undefined, ExprFactory,
+    ModuleItemLike, StmtLike, HANDLER,
 };
 use swc_ecma_visit::{
     as_folder, noop_visit_mut_type, noop_visit_type, Fold, Visit, VisitMut, VisitMutWith, VisitWith,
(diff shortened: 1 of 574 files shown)


crates/swc_ecma_transforms_compat/src/es2022/class_properties/mod.rs
```diff
@@ -58,10 +59,93 @@ pub struct Config {
...
 }
 
+struct ClassExtra {
+    lets: Vec<VarDeclarator>,
+    vars: Vec<VarDeclarator>,
+    stmts: Vec<Stmt>,
+}
+
+impl ClassExtra {
+    fn new() -> Self {
```
kdy1, line 74: Isn't `#[derive(Default)]` better?
Are you going to add a field that does not have a good default?

kdy1 (approved):
Thank you!

I found a tiny issue, but I think it's fine if clippy is okay with it.

---

swc-bump:
 - swc_ecma_transforms_compat

kdy1 (changes requested):
Ohh... Can you rebase?

crates/swc_ecma_transforms_compat/src/es2022/class_properties/mod.rs
Austaras: No

kdy1 (approved):
Thank you!

PR #3777, opened 2022-02-28
test(css/parser): material design

alexander-akait:
**Description:**

test -  codegen + parser + small recovery test.

Found couple small bugs in recovery mode and codegen, so I added it to prevent regression when I will fix them

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - ''

PR #3779, opened 2022-02-28
test(es/minifier): Prepare porting more rules from `terser`

kdy1:
**Description:**

I'm currently looking for a good test case to work on.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #3784, opened 2022-02-28
fix(swc/cli): correctly expose swcx entrypoint

kwonoj:
**Description:**

This PR fixes entrypoint to `swcx` binary for npm. It was my incorrect understanding pointing to native binary (`swcx.exe` for example) creates symlink to `node_modules/.bin`. This isn't ideal as it requires warm up node.js process, but it allows easier experiment.

This doesn't mean we'll use this as permanent solution. We don't even have decided to use `@swc/core` as single pkg or use separate pkg for cli. It's something we'll need to think about.

Also: if someone see `node_modules/.bin` doesn't have `swcx` properly, it's likely due to https://github.com/npm/cli/issues/4308. I confirmed when it comes to scoped pkg there's no binary entrypoint being created.


kdy1 (approved):
Thank you!

---


swc-bump:
 - ''

PR #3787, opened 2022-02-28
docs(adr): adr-00002 for event trace profiling

kwonoj:
**Description:**

We have basic setups for trace now and started to expand it. Think it's good time to compile some of decision factors into adr for keeping references / history.

### DIFF
--- a/docs/adr/00001-support-transformation-plugin-native.md
+++ b/docs/adr/00001-support-transformation-plugin-native.md
@@ -2,9 +2,9 @@
 
 -   Status: accepted <!-- optional -->
 -   Deciders: @kdy1, @kwonoj <!-- optional -->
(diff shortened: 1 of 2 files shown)


docs/adr/00001-support-transformation-plugin-native.md
```diff
@@ -2,9 +2,9 @@
 
 -   Status: accepted <!-- optional -->
 -   Deciders: @kdy1, @kwonoj <!-- optional -->
--   Date: 2021-01-01 <!-- optional -->
+-   Date: 2022-01-01 <!-- optional -->
```
kwonoj, line 5: 😅 I didn't realize this

kdy1 (approved):
Thanks!

PR #3788, opened 2022-03-01
build(swc/plugin_runner): bump up wasmer@2.2

kwonoj:
**Description:**

Casual bump, notable changes are written in https://wasmer.io/posts/wasmer-2.2.

**BREAKING CHANGE:**

Expect this should not be a breaking changes, especially we have cache version bump to support this. 


crates/swc_plugin_runner/src/cache.rs
```diff
@@ -20,7 +20,7 @@ use wasmer_cache::{Cache as WasmerCache, FileSystemCache, Hash};
 /// however it is not gauranteed to be compatible across wasmer's
 /// internal changes.
 /// https://github.com/wasmerio/wasmer/issues/2781
-const MODULE_SERIALIZATION_VERSION: &str = "v1";
+const MODULE_SERIALIZATION_VERSION: &str = "v2";
```
kwonoj, line 23: maybe we should've set this like `v000001` and do the last digit increment? should we do it now?

kdy1 (dismissed):
LGTM. Thank you!

---

swc-bump:
 - swc_plugin_runner

kdy1 (approved):

PR #3789, opened 2022-03-01
feat(css/lints): Implement `selector-max-combinators` rule

g-plane:
**Description:**

Implemented [`selector-max-combinators`](https://stylelint.io/user-guide/rules/list/selector-max-combinators/) rule.

**BREAKING CHANGE:**

No.


crates/swc_css_lints/src/rules/mod.rs
```diff
@@ -53,6 +55,9 @@ pub fn get_rules(LintParams { lint_config }: &LintParams) -> Vec<Box<dyn LintRul
         (&rules_config.color_no_invalid_hex).into(),
     ));
     rules.push(unit_no_unknown((&rules_config.unit_no_unknown).into()));
+    rules.push(selector_max_combinators(
```
kdy1, line 58: Oh wait, I think this will make linter always run regardless of config.
I missed it in previous review.

g-plane: You're right, but I think we should fix it in another PR?

kdy1: Yeah, you are right, thanks!

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_css_lints

PR #3791, opened 2022-03-01
fix(swc/transforms/react): catch refresh directive comment widely

kwonoj:
**Description:**

This PR attempts to fix react-refresh comment directive `@refresh reset` behavior, catch it more widely if it's placed in top-level comment, or per-component level.

I applied pretty much brute way similar to https://github.com/swc-project/swc/blob/main/crates/swc_ecma_parser/tests/comments.rs#L86 by checking comments across span's hi/lo positions. 

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_transforms_react/src/refresh/mod.rs
+++ b/crates/swc_ecma_transforms_react/src/refresh/mod.rs
@@ -246,12 +246,37 @@ where
 
         let mut should_refresh = self.should_reset;
         if let Some(comments) = &self.comments {
-            if n.hi != BytePos(0) {
-                comments.with_leading(n.hi - BytePos(1), |comments| {
-                    if comments.iter().any(|c| c.text.contains("@refresh reset")) {
-                        should_refresh = true
-                    }
-                });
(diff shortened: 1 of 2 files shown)


crates/swc_ecma_transforms_react/src/refresh/mod.rs
```diff
@@ -246,12 +246,37 @@ where
...
-                        should_refresh = true
-                    }
-                });
+            comments.with_leading(n.hi - BytePos(1), |comments| {
+                if comments.iter().any(|c| c.text.contains("@refresh reset")) {
+                    should_refresh = true
+                }
+            });
+
+            if should_refresh {
```
kwonoj, line 255: I tried to early escape if any of comment have directives. However not much satisfied around code itself if repeated logics can be deduped or not.

kdy1: I think it's fine for now.

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_react

PR #3794, opened 2022-03-01
fix(es/minifier): Fix optimizations of terminating expressions

kdy1:
**Description:**

Previous logic has a bug related to assignments in return/throw arg.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #3797, opened 2022-03-01
refactor(swc/plugin_runner): better performing write_memory_view

kwonoj:
**Description:**

Minor perf improvement to plugin bootstrap around copying bytes into guest memory space. 

Previously we used `safe` operation which loops over alignedVec to set each guest ptr cell bytes. This change instead, use https://docs.rs/wasmer/latest/wasmer/struct.MemoryView.html#method.copy_from for the subarray we have allocated in guest memory space. This is unsafe operation due to this fn doesn't guarantee thread safety - but in our case our plugin runner itself guarantees thread safety in some sort (in short, we don't run single plugin across multiple thread).

**BREAKING CHANGE:**


kwonoj:
Other areas of perf improvement (will file individual issues later)

- module linking
- guest memory sharing between modules (this maybe achieved by above module linking instead)

kdy1 (approved):
I think this is fine, as all `MemoryView`s are thread-local in our case.

---

swc-bump:
 -

kwonoj:
> as all MemoryViews are thread-local in our case.

yes, reason to choose this without any thread sync primitives.

PR #3800, opened 2022-03-01
feat(es/codegen): skip empty args in new_expr

Austaras:
**Description:**

Terser does it (and have a bug).

This piece of code is a bit awkward, it's too complex to be in codegen and too simple to be a seperate pass. I put it in codegen simply because Terser does.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_codegen

Issue #3799, opened 2022-03-01
Filter out cranelift trace when trace is enabled 


kwonoj:
### Describe the bug

When plugin_runner compiles plugin for the first time, it hits cranelift compiler. Compiler have internal traces like

```
{"ph":"i","pid":1,"ts":22751.1,"name":"event C:\\Users\\kwono\\.cargo\\registry\\src\\github.com-1ecc6299db9ec823\\wasmer-compiler-cranelift-2.1.1\\src\\translator\\func_translator.rs:83","cat":"wasmer_compiler_cranelift::translator::func_translator","tid":7,"s":"p","args":{"[file]":"C:\\Users\\kwono\\.cargo\\registry\\src\\github.com-1ecc6299db9ec823\\wasmer-compiler-cranelift-2.1.1\\src\\translator\\func_translator.rs","[line]":83,"message":"translate(1623 bytes, u0:499(i64 vmctx, i32, i32, i32) windows_fastcall)"}},
```

which fills up our trace quickly - we should filter out those traces for our output.

### Expected behavior

-

### Version

latest

kdy1:
Fixed by https://github.com/swc-project/swc/pull/3801

PR #3801, opened 2022-03-01
fix(trace): filter cranelift, log events

kwonoj:
**Description:**

This PR filters out cranelift and other log records from trace generation. I feel there maybe better, effecient way - but couldn't find it for now. Will try to dig time to time. Also I realized release build doesn't generate any trace due to compile time loglevel flag. PR changes it to runtime level to allow trace to be collected.

- https://github.com/swc-project/swc/issues/3799

### DIFF
--- a/crates/node/src/util.rs
+++ b/crates/node/src/util.rs
@@ -11,7 +11,7 @@ use swc_common::{errors::Handler, sync::Lrc, SourceMap};
 use tracing::instrument;
 use tracing_chrome::ChromeLayerBuilder;
 use tracing_subscriber::{
-    prelude::__tracing_subscriber_SubscriberExt, util::SubscriberInitExt, EnvFilter,
+    filter, prelude::__tracing_subscriber_SubscriberExt, util::SubscriberInitExt, EnvFilter, Layer,
 };
 
 #[napi]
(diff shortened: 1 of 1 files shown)


kdy1:
> Also I realized release build doesn't generate any trace due to compile time loglevel flag

Oh, maybe that's the reason performance was not dropped

kdy1 (dismissed):
This seems to have some overhead considering https://github.com/swc-project/swc/runs/5373632744?check_suite_focus=true and https://github.com/swc-project/swc/runs/5373616644?check_suite_focus=true but I think it's small enough.

---

swc-bump:
 - ''

kdy1:
Oh. I found one problem.
The performance of minifier will be greatly affected, because it assumes `release_max_level_info`.

I'll merge this after fixing the minifier.

kwonoj:
@kdy1 Curious if this is still blocked / what to fix if that's the case?

kdy1:
Can you revert `release_max_level_info`?
This requires more investigation, because it breaks an important assumption

kwonoj:
OK, guess need some time to let release build have tracing features.

I'll file separate issue for tracking it.

kdy1 (dismissed):
swc-bump:
 - swc_cli

kwonoj:
ref: https://github.com/swc-project/swc/issues/3863

kdy1:
Oh.. Can you rebase?

kwonoj:
suuuure thing, done.

kdy1 (approved):

kdy1:
Why did you close this? Is there a better way?

kwonoj:
@kdy1 sorry, I accidentally deleted branch while cleaning up and it caused to cancel automerge. 😭

kdy1 (approved):

kdy1:
Totally understandable, I did same thing lots of time, because branch list view of `kdy1/swc` does not show active PR numbers

PR #3803, opened 2022-03-01
refactor(css/lints): Skip running rule if disabled

g-plane:
**Description:**

Skip running rules that are disabled in config.

**BREAKING CHANGE:**

It seems no.


kdy1 (approved):
LGTM. Thanks!

---

swc-bump:
 - swc_css_lints

PR #3805, opened 2022-03-01
fix(es/minifier): Preserve more variables

kdy1:
**Description:**

swc_ecma_minifier:
 - `unused`: Preserve more variables.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #3806, opened 2022-03-01
build(swcx): rename entrypoint to avoid artifact collision

kwonoj:
**Description:**

Gh action uses wide match to upload native binary artifacts

```
with:
    name: bindings-x86_6[4](https://github.com/swc-project/swc/runs/5377724303?check_suite_focus=true#step:13:4)-apple-darwin
    path: swc*
```

which makes swcx.js also hits those. PR renames entrypoint try to avoid those collision.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

PR #3809, opened 2022-03-01
fix(node-swc): Fix referencing wrong property name `expr` of `OptionalChainingExpression` in `Visitor`

### LINKED ISSUE #3808: Exception in `visitOptionalChainingExpression` when visiting program
### Describe the bug

When visiting AST with `Visitor.visitProgram()` and encountering an `OptionalChainingExpression`, it throws an exception when calling:

https://github.com/swc-project/swc/blob/f58b50bea7507def95b94b498ba9c5faf55df802/node-swc/src/Visitor.ts#L1097-L1100

Because the object `OptionalChainingExpression` doesn't have a property named `expr` (but `base`).

```repl
Welcome to Node.js v17.6.0.
Type ".help" for more information.
> require('@swc/core').parseSync('a?.b').body[0].expression
{
  type: 'OptionalChainingExpression',
  span: { start: 0, end: 4, ctxt: 0 },
  questionDotToken: { start: 1, end: 2, ctxt: 0 },
  base: {
    type: 'MemberExpression',
    span: { start: 0, end: 4, ctxt: 0 },
    object: { type: 'Identifier', span: [Object], value: 'a', optional: false },
    property: { type: 'Identifier', span: [Object], value: 'b', optional: false }
  }
}
```

So `visitExpression` throws.

### Input code

```typescript
`a?.b`
```

### Config

```json
N/A
```

### Playground link

_No response_

### Expected behavior

It should visit the `base` property.

### Actual behavior

It throws an exception.

### Version

1.2.146

### Additional context

_No response_

Menci:
**Description:**

Fix #3808.

#3808.


kdy1 (changes requested):
Can you change type definitions, too?

PR #3810, opened 2022-03-01
Refactor recovery mode

alexander-akait:
**Description:**

refactor test and improve error reporting messages

Some error tests were invalid - they even contains `.stderr` files (because we don't check it in our tests), some of them are valid (looks like they were copied somewhere old), some of them were moved to recovery (they still report errors, but also output AST, i.e. recovery mode) 

**BREAKING CHANGE:**

No

No


kdy1 (approved):
LGTM. Thanks!

---

swc-bump:
 - swc_css_parser --breaking

PR #3811, opened 2022-03-01
fix(swc/node): correctly close trace spans when completes

kwonoj:
**Description:**

**NEED UPDATE**

There was a bug in `swc/node` for flushing trace. In short, holding flushguard in static is not what we want. It won't drop ever, which causes flushguard teardown doesn't occur. PR changes direction to utilize `add_env_cleanup_hook` in `napi-rs`. This seems to work, but there seems some problem to be solved logistically.

I guess @Brooooooklyn may able to answer these questions in this PR:

- is it correct to use `add_env_cleanup_hook` to finalize some of operations to be expected to trigger when process' lifecycle completes?
- I was not able to use `#[napi]` macro to access `CallContext` (or `Env`). Is there correct configuration to enable those?
- If not, using `js_function` is correct to get callcontext?
- It looks like napi-rs autogenerate `binding.js` which skips module exports to fn declared with `js_function`. Any workaround to solve this?

**BREAKING CHANGE:**


Brooooooklyn:
> is it correct to use `add_env_cleanup_hook` to finalize some of operations to be expected to trigger when process' lifecycle completes?

Yes.

> I was not able to use `#[napi]` macro to access `CallContext` (or `Env`). Is there correct configuration to enable those?

You can inject `Env` directly into `fn`: https://napi.rs/docs/concepts/inject-env . So you don't need `#[js_function]` and `#[module_exports]`

kdy1 (approved):
Thank you!

swc-bump:
 - ''

PR #3815, opened 2022-03-02
feat(swc): Add comment preservation

### LINKED ISSUE #2964: Comments on types are stripped
### Describe the bug

Comments are stripped due to unknown reasons, against expectation.

### Input code

```typescript
// FILE bad.ts:

/* istanbul ignore next */
type Z = number;
x = 1;


// FILE good.ts
/* istanbul ignore next */
x = 1;
```

### Config

```json
{
  "jsc": {
    "transform": {}
  },
  "swcrc": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.117&code=H4sIAAAAAAAAAzXMOwrAIBBF0d5VvFoCklqykXQahjigY4hjPruPTdpz4ToLbhok9gzepZ4EoUdhnXEO%2Bh6EFQukl0inx7AuWy2FRKGJGzILTbiDbmkA4Y8h1otwBeGWzDMWszcfPlBa620AAAA%3D&config=H4sIAAAAAAAAA6tWys3My0yrVLJKS8wpTtVRyipOVrKqVipILCpOLQKxiivzShIrlKyUSioLUouTizILSpRqa2sBrzFAMDkAAAA%3D

### Expected behavior

`bad.ts` should emit:

```
/* istanbul ignore next */
x = 1;
```

but does emit:

```
x = 1;
```

dropping the istanbul comment

### Version

1.2.117

### Additional context

_No response_

dpchamps:
## Problem

Generated code will drop comments in a variety of situations. Some of these situations may include:

- Comments that precede types
- Comments that precede / trail nodes that are transformed / moved / removed

The JS Ecosystem is (unfortunately) reliant on maintaining comment continuity between source and compiled code. There are many existing tools that use comments as annotations for various functionality, the most entrenched probably being `istanbul` coverage annotations. Not maintaining good parity between source and compiled output causes adopting SWC for test runners to be non-tenable for a lot of customers.

## Solution

Implement a "comment fixer" visitor, `dropped_comments_preserver`, and introduce a feature flag `preserve_dropped_comments` into the jsc configuration object.

### Overview of changes

Given the way SWC internals track comments, the comment fixer makes a best effort to preserve comments that may have otherwise been dropped. The output may not be 1:1 with what babel or other transpilers output, but it will preserve all expected functionality from existing js tooling where possible.

When `preserve_dropped_comments` is enabled, all trailing and leading comments will be shifted to the first non-dummy span that exists in the ast. This guarantees that all existing comments will be transported to the compiled output, orphaned comments will _precede_ comments that would otherwise be preserved, and comments that would be preserved remain in a consistent position.

---

I tried my best to organize things as I thought they should be according to the current standards of the repo. Open to any and all suggestions 🙏  would _really_ like to support this feature as it is currently blocking us over here at Walmart from adopting the SWC jest transformer fully. The gains are too good to ignore.

**BREAKING CHANGE:**

**Related issue (if exists): #2964 **

---

Closes #2964


crates/swc_ecma_transforms_base/src/dropped_comments_preserver.rs
```diff
@@ -0,0 +1,55 @@
...
+    comments: Option<&'a dyn Comments>,
+    comment_position: u32,
+    is_first_span: bool,
+}
+
+impl VisitMut for DroppedCommentsPreserver<'_> {
+    noop_visit_mut_type!();
+
+    fn visit_mut_span(&mut self, span: &mut Span) {
+        if span.is_dummy() || self.is_first_span {
```
dpchamps, line 31: When rearranging comments, we have to stop and think what we should do with the module / script span. A comment at the top could be dropped, could be preserved in place, or could be transported to the next available span. Example of this ambiguity:

On the one hand, I could have something like this:

```typescript
/* Some doc block, this very clearly should remain at the top of the file */
type X = number;
...
```

Or I could have something like this:

```
/* istanbul ignore next-line */
type X = number;
const somethingFancy = (...x)  => x;
```

Because during transpilation code such as the second example will be shifted down to support utility functions, I opted to solve this ambiguity by assuming that the comment _must_ be moved to the next nearest existing span. This way at the very minimum tooling is supported.

kdy1 (changes requested):
Code changes looks fine, but I think we need some tests to ensure that we does not use comments of other modules.

Can you add some tests js test suite at https://github.com/swc-project/swc/tree/ce78344a7c3597559ef32b0ee727dc0c7512c492/node-swc/__tests__ ?

Also, please fix the clippy errors.

dpchamps:
@kdy1 happy to add the tests, will start early tomorrow pst. 

Could you help me understand what you mean by comments of other modules?

Lets say I call `transformSync(someSource)` with `preservedDroppedComments: true` 

I'd expect to see all comments within `someSource` to be shifted. Is there a case where there's an oppertunity for other comments to become present?

I'll also update the node-swc config types in the follow-up commit. 

Thanks!

kdy1:
SWC has a global storage for comments, and the storage is shared among files. And the storage uses `BytePos` to store/get comments.
The problem I'm worrying about is `BytePos` used incorrectly so the storage takes comments of other module, or stores comments in other module.


This may change in the future, but for now, we have to ensure that `BytePos` is used correctly.

crates/swc/src/config/mod.rs
```diff
@@ -1021,6 +1028,9 @@ pub struct JscConfig {
 
     #[serde(default)]
     pub lints: LintConfig,
+
+    #[serde(default)]
+    pub preserve_dropped_comments: bool,
```
magic-akari, line 1033: is it necessary? 
I suppose swc should always preserve comments.
And It's ok to remove Comments at the minify phase.

kdy1: It's because this patch tries to preserve comments on dropped nodes.

Btw, I think `preserve_all_comments` is better name. (Even if it does not preserve all comments)

crates/swc/src/config/mod.rs
```diff
@@ -362,7 +365,11 @@ impl Options {
 
         let regenerator = transform.regenerator.clone();
 
-        let preserve_comments = js_minify.as_ref().map(|v| v.format.comments.clone());
+        let preserve_comments = if preserve_all_comments {
```
dpchamps, line 368: `preserveAllComments` should override what is specified in minify

dpchamps:
@kdy1 lmk if this is what you had in mind: https://github.com/swc-project/swc/blob/8b971afa46f9359b7b7ac8fa5d843b35f3de4dd0/node-swc/__tests__/preserve_comments.mjs

Clippy errors fixed, unmarking this as a draft and requesting another pass. thanks!

crates/swc/src/config/mod.rs
dpchamps: @magic-akari I added this because it was requested here: https://github.com/swc-project/swc/issues/2964#issuecomment-996189885

@kdy1 : agreed, see https://github.com/swc-project/swc/pull/3815/commits/3ce4af74d690bfad24dc40f33526ee5b1b2d8fa5

kdy1:
Yes, that's what I meant.
Thank you!

crates/swc_ecma_transforms_base/src/dropped_comments_preserver.rs
```diff
@@ -0,0 +1,54 @@
...
+            self.is_first_span = false;
+            return;
+        }
+
+        let mut leading_comments = Vec::new();
+        let mut trailing_comments = Vec::new();
+
+        for idx in (self.comment_position..=span.lo.0).map(BytePos) {
```
kdy1, line 40: Oh, this will be extremely slow.

kdy1: It removes whole advantage of storing comments in a shared hash map.

kdy1 (changes requested):
I'm not sure about the way to fix this, so I'll think about it more.

crates/swc_ecma_transforms_base/src/dropped_comments_preserver.rs
dpchamps: @kdy1 Sorry, it was an extremely busy day for me. I'm just now sitting down and thinking about how to make this more this more performant. 

Reviewing the discord convo from last night, your suggestion to use `SingleThreadedComments` is unclear to me. I don't see a way to depart from the `SwcComments` data structure that is used top-level in `Compiler` without _replacing_ the compiler type.

This seems to be an appropriate way to solve this problem:

- Collect all candidate spans with a visitor
- iterate _exactly once_ over the `comments` entries, shifting orphaned comments to candidate spans

I understand the reasons that iterating over the concurrent hashmaps is undesirable, but reducing that iteration to one time outside of the visitor sounds like an acceptable concession. Do you agree?

If not, could you provide some guidance on how exactly you intend single threaded comments to be used.

kdy1: > because this needs to be implemented as a pass baked into existing compiler

No, it should not and it does not have to.
It will used inside `fn transform` and it will be dropped once `fn transform()` is finished.

dpchamps:
@kdy1 addressing the performance concerns and general feedback from our discord chat. Unmarking as draft and requesting another pass, thanks a ton!

crates/swc/src/dropped_comments_preserver.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// This maintains the highest fidelity between existing comment positions
+    /// of pre and post compiled code.
+    fn shift_leading_comments(&mut self, comments: &SingleThreadedComments) -> CommentEntries {
+        let mut existing_comments = self.collect_existing_comments(comments);
+
+        for span in self.known_spans.iter() {
+            let (comments_to_move, next_byte_positions): (CommentEntries, CommentEntries) =
+                existing_comments
+                    .drain(..)
+                    .partition(|(bp, _)| bp <= &span.lo);
```
kdy1, line 103: Please use `*bp <= span.lo` instead

crates/swc/src/dropped_comments_preserver.rs
```diff
@@ -0,0 +1,135 @@
...
+/// output. This can create problems in the JavaScript ecosystem, particularly
+/// around instanbul coverage and other tooling that relies on comment
+/// directives.
+///
+/// This transformer shifts orphaned comments to the next closest known span
+/// while making a best-effort to preserve the "general orientation" of
+/// comments.
+
+pub fn dropped_comments_preserver(
+    comments: Option<&SingleThreadedComments>,
```
kdy1, line 21: I think we don't need a lifetime here.

kdy1: `Comments` are cheap to clone, and it's part of the api.

crates/swc/src/dropped_comments_preserver.rs
```diff
@@ -0,0 +1,135 @@
...
+        existing_comments
+    }
+
+    /// These comments trail all known span lo byte positions.
+    /// Therefore, by shifting them to trail the highest known hi position, we
+    /// ensure that any remaining trailing comments are emitted in a
+    /// similar location
+    fn shift_trailing_comments(&mut self, remaining_comment_entries: CommentEntries) {
+        let last_trailing =
+            self.known_spans.iter().fold(
```
kdy1, line 122: Please use `.iter().copied()` instead.
`Span` is `Copy`, so we don't need indirection here.

kdy1 (changes requested):
Thank you!

dpchamps:
Feedback addressed https://github.com/swc-project/swc/pull/3815/commits/7b3d8e3c859148ef7923317093ee77f691b753de

kdy1 (approved):
Thank you!

---


swc-bump:
 - swc --breaking

PR #3818, opened 2022-03-02
feat(css/lints): Add `font-family-no-duplicate-names` rule

g-plane:
**Description:**

Implemented [`font-family-no-duplicate-names`](https://stylelint.io/user-guide/rules/list/font-family-no-duplicate-names/) rule.

**BREAKING CHANGE:**

No.


crates/swc_css_lints/src/rules/font_family_no_duplicate_names.rs
```diff
@@ -0,0 +1,167 @@
...
+    fn visit_declaration(&mut self, declaration: &Declaration) {
+        match &declaration.name {
+            DeclarationName::Ident(Ident { value, .. })
+                if value.eq_str_ignore_ascii_case("font-family") =>
+            {
+                declaration.value.iter().fold(
+                    AHashSet::with_capacity(declaration.value.len()),
+                    |mut fonts, item| {
+                        let font = match item {
+                            ComponentValue::Ident(Ident { value, .. }) => {
```
alexander-akait, line 43: There is potential problem, font can be sequence of ident, i.e. `font: Foo Bar, bar foo`, it is valid due legacy 

> Font family names other than generic families must either be given quoted as [strings,](https://www.w3.org/TR/CSS21/syndata.html#strings) or unquoted as a sequence of one or more [identifiers.](https://www.w3.org/TR/CSS21/syndata.html#value-def-identifier)

https://www.w3.org/TR/css-fonts-3/#family-name-value

kdy1 (approved):
swc-bump:
 - swc_css_lints

PR #3821, opened 2022-03-02
fix(css/codegen): semicolon

alexander-akait:
fix codegen with semicolon

**BREAKING CHANGE:**

No

No

### DIFF
--- a/crates/swc_css_codegen/tests/fixture/declaration/important/output.css
+++ b/crates/swc_css_codegen/tests/fixture/declaration/important/output.css
@@ -24,4 +24,4 @@ color: red !important;
 }
 }
 .foo {
-color: red !ie}
+color: red !ie;}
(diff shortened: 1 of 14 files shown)


crates/swc_css_codegen/tests/fixture/declaration/important/output.css
```diff
@@ -24,4 +24,4 @@ color: red !important;
 }
 }
 .foo {
-color: red !ie}
+color: red !ie;}
```
alexander-akait, line 27: Let's keep `;` for all invalid declarations

crates/swc_css_codegen/tests/fixture/list_of_declarations/3/output.min.css
```diff
@@ -0,0 +1,4 @@
+a.invert{margin:0;__ident__
+    border-radius: 7px;color:white;background:__ident__;__ident__
+    __ident__
+}
```
alexander-akait, line 4: We can compress white spaces here, but let's do it in other PR

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_parser

PR #3824, opened 2022-03-03
fix(es/hygiene): Fix renaming bugs

### LINKED ISSUE #3213: Variable being renamed unnecessarily is breaking code

### LINKED ISSUE #3823: Incorrect output when 2+ vars are named with a shared prefix and one var name is re-used across scopes
### Describe the bug

When there are variables whose names share the same prefix and where one variable name is shared between an inner and outer scope.

Reproduced here - https://github.com/bnoguchi/swc-bug-repro-1

To reproduce:

```bash
git clone https://github.com/bnoguchi/swc-bug-repro-1.git
npm i
npx swc index.js -o out.js
```

Then execute index.js and out.js, and notice that you have 

```bash
$ node index.js
{ aaa: 2, bbb: 3 }
```

```
$ node out.js
{ aaa: 'bbb', bbb: 3 }
```

If swc were working correctly, then we would expect to get the same result.
However, we get different results because `out.js` ends up conflating 2
variables.

### Input code

The contents of index.js:

```typescript
'use strict';

function print() {
  var value$4 = 2;
  var value$41 = value$4 + 100;
  var c = 1;
  if (true) {
    var value$42 = "bbb";
    if (value$42 === "bbb") c = 3;
  }

  [0, 1].map(function (value) {
    var value$4 = 2;
    var x = value ? value$4 : "";
    return x;
  });
  return {aaa: value$4, bbb: c};
}

console.log(print());
```

### Config

```json
/* empty */
```

### Playground link

https://play.swc.rs/?version=1.2.146&code=H4sIAAAAAAAAA21Q3QqDIBS%2B9ykOMUhZRG67SmIPMnZhYkNoGqYjiN599mMj2IUH%2FH7O93FS30vonVXCpQyhxmvhlNHQWaUdJjAigA%2B34bVenm5QwYUdIRqwyJ6BFkXkRSDo%2FFENYGe9XLcdzJegSeq6TtjCzMofU20cWVZdZ8mEwngUGdBn%2FuYd3vuutj8JsfEKDrEs3HdBCckWb6XzVsOwJJF5bsjIOS%2BjIYNQqgQxMRTqCKN708q8NS%2B8HY2wL%2BJYBH1WAQAA&config=H4sIAAAAAAAAA0WMSwrDMAxE76K1t93kDj2EcJXg4B8aBWqM714bUrKSmPdmOp3wtHWqrBBdH1o2%2FtJG4hPDa6hGbmoz2jlChiNjPcSWgtdksRTITR2lkMPe1pIvqaoAD%2BJ8xL855lAqn2sFnazVeVcllXyCxjN0lwPet216yfgBCKaI9boAAAA%3D

### Expected behavior

```javascript
'use strict';
function print() {
    var value$4 = 2;
    var value$41 = value$4 + 100;
    var c = 1;
    if (true) {
        var value$42 = "bbb";
        if (value$42 === "bbb") c = 3;
    }
    [
        0,
        1
    ].map(function(value) {
        var value$4 = 2;
        var x = value ? value$4 : "";
        return x;
    });
    return {
        aaa: value$4,
        bbb: c
    };
}
console.log(print());


//# sourceMappingURL=out.js.map
```

### Actual behavior

Notice how `value$4` and `value$42` from index.js are combined into one var in
out.js named `value$42`, instead of being kept as separate variables.

```javascript
'use strict';
function print() {
    var value$42 = 2;
    var value$41 = value$42 + 100;
    var c = 1;
    if (true) {
        var value$42 = "bbb";
        if (value$42 === "bbb") c = 3;
    }
    [
        0,
        1
    ].map(function(value) {
        var value$4 = 2;
        var x = value ? value$4 : "";
        return x;
    });
    return {
        aaa: value$42,
        bbb: c
    };
}
console.log(print());


//# sourceMappingURL=out.js.map
```

### Version

@swc/cli: 0.1.55 ; @swc/core: 1.2.147

### Clue 1

The issue seems to be related to the re-use of the var name `value$4` inside the Array.map callback. If we remove that callback or if we rename the var in the callback from `value$4` to `value$9`, then the issue disappears.

For example, the issue disappears if we change from:

```javascript
  [0, 1].map(function (value) {
    var value$4 = 2;
    var x = value ? value$4 : "";
    return x;
  });
```

to

```javascript
  [0, 1].map(function (value) {
    var value$9 = 2;
    var x = value ? value$9 : "";
    return x;
  });
```

### Clue 2

The issue seems to be related to the shared prefixes of the var names `value$4`, `value$41`, `value$42`. If we change `var value$41 = value$4 + 100;` to `var value$1 = value$4 + 100;`, then the issue disappears.

For example, the issue disappears if we change index.js to:

```javascript
'use strict';

function print() {
  var value$1 = 2;
  var value$2 = value$1 + 100;
  var c = 1;
  if (true) {
    var value$3 = "bbb";
    if (value$3 === "bbb") c = 3;
  }

  [0, 1].map(function (value) {
    var value$1 = 2;
    var x = value ? value$1 : "";
    return x;
  });
  return {aaa: value$1, bbb: c};
}

console.log(print());
```

### Clue 3

The issue seems to only occur when one of the variables that share a name prefix is declared and used inside of a block.

For example, the issue disappears if we change from:

```javascript
  var c = 1;
  if (true) {
    var value$42 = "bbb";
    if (value$42 === "bbb") c = 3;
  }
```

to 

```javascript
  var c = 1;
  var value$42 = "bbb";
  if (value$42 === "bbb") c = 3;
```

kdy1:
**Description:**

We 'reserve' all used symbols, so it cannnot result in an error

 - Closes #3213
 - Closes #3823


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base

jdalton:
This is still happening in 1.2.164.

It's happening with consts defined outside a try-catch and used inside a try-catch

kdy1:
@jdalton Please file a new issue

jdalton:
Input:
```js
const a = b;
try {
  foo(a);
} catch {}
```
Output:
```js
const a = b;
try {
  var a;
  a = a;
  foo(a);
} catch {}
```

jdalton:
@kdy1 Done! https://github.com/swc-project/swc/issues/4294 Thank you!

PR #3828, opened 2022-03-03
fix(css/parser): recovery mode

alexander-akait:
**Description:**

Fix recovery codegen

**BREAKING CHANGE:**

No

No


kdy1 (approved):
LGTM. Thank you!

---


swc-bump:
 - swc_css_parser

alexander-akait:
@kdy1 Fixed, just rewrite test, we should not write `;` at the end, when we have `Invalid` (i.e. token vector), now we handle these cases in parser/codegen and keep `;` (Semi) in AST

kdy1 (approved):

PR #3833, opened 2022-03-03
feat(css/lints): Implement `no-duplicate-at-import-rules` rule

g-plane:
**Description:**

Implemented [`no-duplicate-at-import-rules`](https://stylelint.io/user-guide/rules/list/no-duplicate-at-import-rules/).

**BREAKING CHANGE:**

No.


alexander-akait (commented):
Let's add tests with layer https://www.w3.org/TR/css-cascade-5/#at-import

g-plane:
Are we missing anything? Why do we mark this PR "work in progress"?

kdy1:
Because of https://github.com/swc-project/swc/pull/3833#pullrequestreview-899319493

I marked it as a draft (WIP) because I don't want to poll PRs

g-plane:
@alexander-akait Are there any differences for the code below?

```css
@import "url.css" layer;
```

```css
@import "url.css" layer();
```

```css
@import "url.css" layer(default);
```

g-plane:
@alexander-akait Can you answer the question above? Thanks.

alexander-akait:
@g-plane Sorry for delay:
```css
@import "url.css" layer;
@import "url.css" layer();
```

are the same, but

```css
@import "url.css" layer(default);
@import "url.css" layer(other);
```

Are not, for example you can have:
```
@import "normalize.css" layer(framework);
@import "normalize.css" layer(my-library);
```

But in other hand our rule about duplication in URLs, and maybe we should report here anyway, that's a good question for discussion, currently I think we should always report about duplications, maybe we will revisit it late (will be great to open an issue in stylelint too, so that we can decide on a more correct behavior).

So right now let's add tests to ensure we always report about duplication with layer 

https://www.w3.org/TR/css-cascade-5/#cascading

alexander-akait (dismissed):

g-plane:
@kdy1 This can be reviewed now.

crates/swc_css_lints/src/rules/no_duplicate_at_import_rules.rs
```diff
@@ -0,0 +1,78 @@
+use std::fmt::Display;
+
+use ahash::AHashSet;
```
kdy1, line 3: Please use `AHashSet` from `swc_common` instead.

kdy1: (For consisstency.)

kdy1: `ahash::AHashSet` and `ahash::AHashMap` are newtype structs, while those in `swc_common` is type alias for standard hashmap type.

kdy1 (changes requested):
Thank you!
I reviewed only rust code and I'll request review for css parts, as I don't know well about css

alexander-akait (approved):

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_css_lints

kdy1 (changes requested):
CI failed

kdy1 (approved):

kdy1:
Can you rebase?

kdy1 (changes requested):
There are some merge conflicts

kdy1 (approved):
LGTM. Thanks!

---

swc-bump:
 - swc_css_lints

yisibl:
> @alexander-akait Are there any differences for the code below?
> 
> ```css
> @import "url.css" layer;
> ```
> 
> ```css
> @import "url.css" layer();
> ```
> 
> ```css
> @import "url.css" layer(default);
> ```

`@import "url.css" layer();` is a wrong syntax. see https://github.com/w3c/csswg-drafts/pull/7601

alexander-akait:
@yisibl Can you open an issue?

yisibl:
> @yisibl Can you open an issue?

No problem: https://github.com/swc-project/swc/issues/5643

PR #3836, opened 2022-03-03
feat(css/parser): parse color functions

alexander-akait:
**Description:**

Parse colors functions as in spec +  improve minifier for alpha value

Things to improve (currently we use https://www.w3.org/TR/css-color-4/#funcdef-hwb):
1. `from` parsing in spec https://www.w3.org/TR/css-color-5/#relative-colors
2. `color-mix`/`color-contrast` parsing, but currently no one browser supports it, so we will do it late

Also we need it to improve minification of color functions

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_ast --breaking

PR #3841, opened 2022-03-03

kdy1:
**Description:**

swc_ecma_minifier:
 - `conditionals`: Preserve a conditional expression if test is not resolved.
 - `unsafe_arrows`: Optimize an anonymous function expression as an arrow expression.
 - `unsafe_methods`: Optimize a method property where the only statement is return to an arrow expression.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #3842, opened 2022-03-04
fix(css/lints): Allow empty `@layer` rule before `@import` rule

g-plane:
**Description:**

According to [spec](https://www.w3.org/TR/css-cascade-5/#at-import), this PR fixes `no-invalid-position-at-import-rule` rule, which allows:

```css
@layer {}

@import "x.css";
```

> Any [@import](https://www.w3.org/TR/css-cascade-5/#at-ruledef-import) rules must precede all other valid at-rules and style rules in a style sheet (ignoring [@charset](https://www.w3.org/TR/css-syntax-3/#at-ruledef-charset) and empty [@layer](https://www.w3.org/TR/css-cascade-5/#at-ruledef-layer) definitions).

**BREAKING CHANGE:**

No.

### DIFF
--- a/crates/swc_css_lints/src/rules/no_invalid_position_at_import_rule.rs
+++ b/crates/swc_css_lints/src/rules/no_invalid_position_at_import_rule.rs
@@ -36,7 +36,7 @@ impl Visit for NoInvalidPositionAtImportRule {
             }
 
             match rule {
-                Rule::AtRule(AtRule::Charset(..) | AtRule::Import(..)) => seen,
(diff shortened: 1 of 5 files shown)


alexander-akait (approved):

kdy1 (approved):
swc-bump:
 - swc_css_lints

yisibl:
This implementation does not match the actual browser parsing, and I submitted an issue at CSSWG https://github.com/w3c/csswg-drafts/issues/7566

PR #3847, opened 2022-03-04
perf(es/transform): Use `SingleThreadedComments` for transform

kdy1:
**Description:**

We now uses `SingleThreadedComments` for transforms and  minification.

**BREAKING CHANGE:**

API is changed


kdy1 (commented):
swc-bump:
 - swc_common
 - swc --breaking

PR #3855, opened 2022-03-04
feat(css/lints): Allow regex in "ignore list"

g-plane:
**Description:**

In Stylelint, some rules, for example [`at-rule-no-unknown`](https://stylelint.io/user-guide/rules/list/at-rule-no-unknown), allow regular expressions in configuration, and this PR implemented that feature.

**BREAKING CHANGE:**

API of `swc_css_lints` has breaking changes, because its function signature has changed.


alexander-akait (approved):
Looks good, thanks

kdy1 (changes requested):
Can you change to expose `anyhow::Error` instead of `regex::Error`?

g-plane:
Would it be better that using `thiserror`?

kdy1:
Yes, because all errors belong to a configuration error

crates/swc_css_lints/src/pattern.rs
```diff
@@ -0,0 +1,41 @@
+use regex::Regex;
+
+#[derive(Debug)]
+pub(crate) enum NamePattern {
+    Str(String),
+    Regex(Regex),
```
kdy1, line 6: Can you use [CachedRegex](https://docs.rs/swc_cached/latest/swc_cached/regex/struct.CachedRegex.html) instead?

kdy1: I made `CachedRegex` because the configuration is typically shared among files in a project.

kdy1 (changes requested):

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_css_lints --breaking

kdy1 (approved):

PR #3856, opened 2022-03-04
fix(lazy/export): Fix lazy + 'export { foo } from "bar"'

gmalette:
# Description

When using lazy dependency loading, along with `export { foo } from "bar"`, the generated code for the export currently uses a identifier property, not a function call (`_bar.foo`, instead of `_bar().foo`).

The first commit in this PR shows the problem.
The second commit fixes it.

This changes things slightly where previously `export { foo } from "bar"` was denied lazy-loading for `"bar"`. Was this for performance reasons? If it was only for performance, and if it's something we'd like to preserve, I'll introduce the necessary commit.

# Discussion

The original design of CommonJS + Lazy was different than the direction in which we're currently taking it, and I want to make sure that everyone understands, and that we're all aligned on the new design.

Originally, identifiers were denied lazy-loading on-the-fly, which led to many bugs, mostly due to the order of operations. It has since changed to use a preliminary pass using `LazyIdentifierVisitor`. The goal was to change the failure mode of `Lazy`. Before, it was able to produce faulty code. Now, the failure mode *should* be valid but unoptimized code (not always the case yet, as shown in this PR).

The way I see it, if we run `CommonJS` without `LazyIdentifierVisitor`, it should still produce valid code. `LazyIdentifierVisitor` should only be thought of as a performance optimization.

Therefore, `CommonJS` should not be making decisions as to what can and cannot be lazy loaded.


crates/swc_ecma_transforms_module/src/common_js.rs
```diff
@@ -633,23 +633,24 @@ impl Fold for CommonJs {
...
                                 };
 
-                                let is_reexport = export.src.is_some()
-                                    || scope
-                                        .idents
-                                        .contains_key(&(orig.sym.clone(), orig.span.ctxt()));
-
                                 drop(scope);
 
                                 let old = self.in_top_level;
```
gmalette, line 638: I don't know if setting `in_top_level` after this block remains relevant

kdy1 (dismissed):
LGTM. Thank you!

---


swc-bump:
 - swc_ecma_transforms_module

gmalette:
Can you stop the merge? Without the few lines removed it causes `swc` to fail immediately on my codebase. I have no idea why

gmalette:
@kdy1 can you give this another glance please? I rebased, changed nothing else

gmalette:
@kdy1 can you give this another glance please? I rebased, changed nothing else

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_module

PR #3865, opened 2022-03-05
test(swc): Use external helpers for reference testing

kdy1:
**Description:**

This make diff less verbose

---

cc @Austaras


kdy1 (commented):
swc-bump:
 - ''

PR #3871, opened 2022-03-05
fix(es/parser): allow `await` in non-strict mode

### LINKED ISSUE #3861: error: await isn't allowed in non-async function
### Describe the bug

In the poc.js, await is not a reserved word, but a parameter, but it is detected as await state, so the following error is reported：
```bash
error: await isn't allowed in non-async function
```

### Input code

```typescript
async function fn() {
  function g(x = await) {}
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript"
    },
    "target": "es5"
  },
  "module": {
    "type": "commonjs"
  },
  "isModule": false
}
```

### Playground link

https://play.swc.rs/?version=1.2.148&code=H4sIAAAAAAAAA0ssrsxLVkgrzUsuyczPU0jL09BUqOZSQIika1Qo2CoklidmlgBlarlqAeoUWrMyAAAA&config=H4sIAAAAAAAAAy2LQQqAMAwE%2F5KzVy%2F9g48INUpL05YkgkX6dw142t0Z9oGsEcIDHUVJvOmohjcEoMioUVI3mAsYyknmWFff3ParkB9s9C8hNuZWs7pMuv36wKI0X3tLUmhnAAAA

### Expected behavior

SWC shouldn't report the error.

### Actual behavior

_No response_

### Version

1.2.148

### Additional context

I run swc use cmd as follows
```bash
```

magic-akari:
**Description:**

**BREAKING CHANGE:**

- fix #3861

### DIFF
--- a/crates/swc_ecma_parser/src/parser/expr/ops.rs
+++ b/crates/swc_ecma_parser/src/parser/expr/ops.rs
@@ -319,7 +319,7 @@ impl<'a, I: Tokens> Parser<I> {
 
         if is!(self, "await") {
             let ctx = self.ctx();
-            if ctx.in_function && !ctx.in_async {
+            if ctx.in_function && !ctx.in_async && ctx.strict {
                 self.emit_err(self.input.cur_span(), SyntaxError::AwaitInFunction);
             }
             return self.parse_await_expr();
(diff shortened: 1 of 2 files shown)


kdy1 (approved):
Thank you!

---


swc-bump:
 - swc_ecma_parser

PR #3876, opened 2022-03-06
TEST: wasm plugin runtime for wasm

kdy1:
**Description:**

I can't install wasm-pack because I'm using arm64 mac... So I decided to use CI instead.

cc @kwonoj 
Let's see if build works.


kdy1:
Closing as it's verified.

Context: https://discord.com/channels/889779439272075314/889785438938726441/949863238504415303

PR #3878, opened 2022-03-06
perf(es/ast): Implement `Clone` without inline for some enums

philipc:
Otherwise each crate which does a clone gets its own copy of the
implementation. This has been done for enums which were identified
as causing bloat in the output binary. This reduces the binary size
and shouldn't negatively affect performance because these were too
large to be inlined anyway.

```
   text	   data	    bss	    dec	    hex	filename
-21101321	 520952	   2560	21624833	149f801	target/release/swc
+20233529	 520984	   2560	20757073	13cba51	target/release/swc
```

### DIFF
--- a/crates/ast_node/src/lib.rs
+++ b/crates/ast_node/src/lib.rs
@@ -171,11 +171,33 @@ pub fn ast_node(
     let mut item = Quote::new(Span::call_site());
     item = match input.data {
         Data::Enum(..) => {
-            if !args.is_empty() {
-                panic!("#[ast_node] on enum does not accept any argument")
+            struct EnumArgs {
+                clone: bool,
             }
+            impl parse::Parse for EnumArgs {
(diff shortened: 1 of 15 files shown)


philipc:
There's failures because I haven't updated `ast_node` usage for the whole project. If this PR is something you want, then I can fix those too.

kdy1 (dismissed):
Thank you so much!

I'll wait for binary size CI action, as this is a breaking change of `ast_node`
(I don't expect others to use it, but...)

kdy1:
I definitely want it, thank you!

philipc:
Maybe we could use an argument to `ast_node` to avoid the breaking change?

kdy1:
Oh, yeah it would be nice

kdy1 (approved):
Thank you so much!


---

swc-bump:
 - ast_node
 - swc_ecma_ast

kdy1 (approved):

PR #3880, opened 2022-03-06
feat(es/lint): support module/script mode in DuplicateBindings 

### LINKED ISSUE #3852: Duplicate Declaration Function should be Overwritten but get Duplicate binding Error
### Describe the bug

The function repeatedly declared in the JavaScript scope will be overwritten by the last function, however, the program can still run correctly, but the error of duplicate binding is reported during SWC conversion.

### Input code

```typescript
function sum ( i, j) {
  return i + j;
}

function sum(){
  return 0;
}

console.log(sum());
console.log(sum(1, 2));
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false,
      "dynamicImport": false,
      "privateMethod": false,
      "functionBind": false,
      "exportDefaultFrom": false,
      "exportNamespaceFrom": false,
      "decorators": false,
      "topLevelAwait": false,
      "importMeta": false
    },
    "transform": null,
    "target": "es5",
    "loose": false,
    "externalHelpers": false
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.148&code=H4sIAAAAAAAAA0srzUsuyczPUyguzVXQUMjUUcjSVKjmUlAoSi0pLcpTyFTQVsiy5qrl4kpDUqmhiaTEACydnJ9XnJ%2BTqpeTn64BVqFpjSFmqKNgBBQHAM2STN90AAAA&config=H4sIAAAAAAAAAy2LQQqAMAwE%2F5KzVy%2F9g48INUpL05YkgkX6dw142t0Z9oGsEcIDHUVJvOmohjcEoMioUVI3mAsYyknmWFff3ParkB9s9C8hNuZWs7pMuv36wKI0X3tLUmhnAAAA

### Expected behavior

After SWC conversion, the code does not change the semantics, and the last function with the same name overwrites the previous function

### Actual behavior

_No response_

### Version

1.2.148

### Additional context

I run swc use cmd as follows
```bash
```

magic-akari:
**Description:**

**BREAKING CHANGE:**

- fix #3852


kdy1 (approved):
LGTM. Thank you!


---

swc-bump:
 - swc_ecma_lints

crates/swc_ecma_lints/src/rules/duplicate_bindings.rs
```diff
@@ -12,41 +15,62 @@ pub fn duplicate_bindings() -> Box<dyn Rule> {
...
+    fn add(&mut self, id: &Ident, unique: bool) {
         match self.bindings.entry(id.to_id()) {
             Entry::Occupied(mut prev) => {
-                HANDLER.with(|handler| {
-                    handler
-                        .struct_span_err(id.span, "Duplicate binding")
-                        .span_note(*prev.get(), &format!("{} was declared at here", id.sym))
-                        .emit();
-                });
+                if unique || prev.get().unique {
```
kdy1, line 40: This is not correct, because `var` can override anything.

kdy1 (changes requested):
Sorry, I missed one logic bug.

crates/swc_ecma_lints/src/rules/duplicate_bindings.rs
magic-akari: It's not true in module.
I have tested this locally with the following code.

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
  </head>
  <body>
    <script>
      var x = 1;
      var x = 2;
      function x() {}
      function x() {} // ok
    </script>

    <script type="module">
      var foo = 1;
      var foo = 2; // ok

      function foo1() {}
      function foo1() {} // error
    </script>

    <script type="module">
      var bar = 1;
      function bar() {} // error
    </script>
  </body>
</html>
```

kdy1: Oh... So `three.js` should be treated as a script?

https://github.com/swc-project/swc/runs/5442747097?check_suite_focus=true

magic-akari: I agree. It should be treated as script.

kdy1 (approved):

PR #3886, opened 2022-03-06
feat(es/lint): implement yoda rule

ArturAralin:
**Description:**

implement yoda rule

**BREAKING CHANGE:**
No


kdy1:
This is also about code style, isn't it?
I'm not sure about this rule

ArturAralin:
Yes. This rule restrict using literals on left side (e.g. 10 === a)

alexander-akait:
@kdy1 Also you can always ping/assign me in an issue and I will give advice about rule/naming/options/etc

kdy1:
Thanks! I'll review this soon.

crates/swc_ecma_lints/src/rules/yoda.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn default() -> Self {
+        Self::Never
+    }
+}
+
+#[derive(Debug, Clone, Default, Copy, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct YodaConfig {
+    mode: Option<YodaConfigMode>,
+    except_range: Option<bool>,
```
kdy1, line 28: I think we need `#[serde(default)]` for each fields.

crates/swc_ecma_lints/src/rules/yoda.rs
```diff
@@ -0,0 +1,141 @@
...
+            if let (
+                Expr::Bin(BinExpr {
+                    op: op!("<") | op!("<="),
+                    ..
+                }),
+                Expr::Bin(BinExpr {
+                    op: op!(">") | op!(">="),
+                    ..
+                }),
+            ) = (bin_expr.left.as_ref(), bin_expr.right.as_ref())
```
kdy1, line 126: I think this will skip checking of nested binary expression like `(foo === 1) && (1 === foo)`

kdy1 (changes requested):
Thank you!

crates/swc_ecma_lints/src/rules/yoda.rs
ArturAralin: I was check it. It works correctly. I added your case to tests

kdy1 (approved):
Thank you!


---

swc-bump:
 - swc_ecma_lints

kdy1:
CI is failing.
Can you run `UPDATE=1 cargo test` from `./crates/swc` after pulling the changes (merge commits for rebasing)

kdy1 (approved):

PR #3887, opened 2022-03-06
feat(es/fixer): Remove extra parens around iife in expression position

Austaras:
**Description:**

Originally fixer would make
```js
a = function () {} ()
```
into
```js
a = (function () {}) ()
```
and since it's the last transformation, it must be fixed here
**BREAKING CHANGE:**


kdy1 (dismissed):
Thank you!

---

swc-bump:
 - swc_ecma_transform_base

Austaras:
Oh, I need to update remain tests

node-swc/src/types.ts
```diff
@@ -1,397 +1,345 @@
 export interface Plugin {
```
kdy1, line 1: This seems like an unintentional change.

kdy1 (changes requested):

kdy1 (approved):

PR #3888, opened 2022-03-06
feat(es/lint): add partial max-length rule implementation

ArturAralin:
**Description:**
Add partial max-length rule implementation.

Now implemented:
* max code length check
* max comment length (works for single line and multiline)
* tab width option
* ignoring trailing comments
* ignoring string (single line and multiline)
* ignoring template literals

Other functions I'll implement future. I need to relax after this rule =)

**BREAKING CHANGE:**

No


kdy1:
I think we should not have rule for code styles.

ArturAralin:
Why? Code style rules is useful by my mind

g-plane:
Strictly, rules about code style shouldn't be a part of linter. Though current ESLint supports many rules about code style, actually, very early version of ESLint doesn't have any rules about code style. Also, there're already some mature formatter tools: Prettier and dprint, etc.

alexander-akait:
Agree, let's avoid most of stylistics rules, but note - there are some differences between them, for example:
- `padding-line-between-statements`, prettier respects newlines between code structures from [original code](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAbOMAEBDTBeTARgAYBuAHSnSwCN8izLLrMx6SKpmNMATdsiAA0ICAAcYAS2gBnZKGwAnRRADuABSUI5KbKlXYAnnJE1F2MAGsMAZWwBbOABlJUOMgBmemXFPmrtmIWrgDmyDCKAK6+IHD2NHC8vIlO2FAhkdghcABiEIr22DBS6cgg2JEwEMIgABYw9qgA6rWS8DJBYHA22m2SAG5thmVgMiYgrj6KMOrmIYWe3jEAVjIAHjah6ACKkRDwi6g+IkGKU2U02AmoNWKKrjBNkrwwtcgAHMQnKj5N5mJlO5wKb9dwiACOe3gs3EOnKMgAtG5Eokaoo4JDJOjZlkFkgvEcYj57JJwlEiVs4Lt9u58UsRDArk8Xm8kAAmBnmSSoUIAYQg9jxsRkAFYapEfAAVK46AnHED9aIASSgyVgNjA9wkAEFVTYYIZ0IcfABfE1AA), this may be useful for more struct code style 
- `max-length` is fully unnecessary, because prettier (and most of other formatters) found better length and reformat code for better readable 

Anyway I think it should be not rules, it should be options for formatter

PR #3889, opened 2022-03-06
feat(es/lint): improve top level decls call

ArturAralin:
**Description:**
Extract top level declared idents only when rule is enabled. This small fix will be good for checking large files

**BREAKING CHANGE:**
No


kdy1 (approved):
LGTM. Thanks!

---

swc-bump:
 - swc_ecma_lints

PR #3890, opened 2022-03-07
feat(css/lints): Implement `custom-property-no-missing-var-function` rule

g-plane:
**Description:**

Implemented [`custom-property-no-missing-var-function`](https://stylelint.io/user-guide/rules/list/custom-property-no-missing-var-function) rule.

**BREAKING CHANGE:**

No.


crates/swc_css_lints/src/rules/custom_property_no_missing_var_function.rs
```diff
@@ -0,0 +1,57 @@
...
+    ctx: LintRuleContext<()>,
+
+    in_declaration: bool,
+    in_declaration_value: bool,
+    in_var_function: bool,
+}
+
+impl Visit for CustomPropertyNoMissingVarFunction {
+    fn visit_declaration(&mut self, declaration: &Declaration) {
+        self.in_declaration = true;
```
kdy1, line 27: I think we should preserve old state, as declaration can be nested.

kdy1: https://docs.rs/swc_css_ast/latest/swc_css_ast/enum.ComponentValue.html#variant.DeclarationOrAtRule
https://docs.rs/swc_css_ast/latest/swc_css_ast/enum.DeclarationOrAtRule.html
https://docs.rs/swc_css_ast/latest/swc_css_ast/struct.Declaration.html

kdy1: `ComponentValue::DeclarationOrAtRule(DeclarationOrAtRule::Declaration(Declaration {}))`

crates/swc_css_lints/src/rules/custom_property_no_missing_var_function.rs
```diff
@@ -0,0 +1,57 @@
...
+        declaration.visit_children_with(self);
+        self.in_declaration = false;
+    }
+
+    fn visit_component_value(&mut self, component_value: &ComponentValue) {
+        self.in_declaration_value = self.in_declaration;
```
kdy1, line 33: Same here

crates/swc_css_lints/src/rules/custom_property_no_missing_var_function.rs
```diff
@@ -0,0 +1,57 @@
...
+        component_value.visit_children_with(self);
+        self.in_declaration_value = false;
+    }
+
+    fn visit_function(&mut self, function: &Function) {
+        self.in_var_function = function.name.value.eq_str_ignore_ascii_case("var");
```
kdy1, line 39: Same here, as `Function` contains `ComponentValue`

kdy1 (changes requested):
Thanks!

kdy1:
@alexander-akait Can you review CSS parts?

crates/swc_css_lints/tests/rules/fail/custom-property-no-missing-var-function/input.css
```diff
@@ -0,0 +1,7 @@
+a { color: --foo; }
+a { color: env(--foo); }
+a { color: color(--foo 0% 0% 0% 0%); }
+a { --foo: red; color: --foo; }
+@property --foo {} a { color: --foo; }
+:root { --foo: pink; }  a { color: --foo, red; }
+:root { --bar: 0; } a { color: --foo(--bar); }
```
alexander-akait, line 7: Can we also add test for `background-color: var(--foo, var(--bar))` (i.e. `var` in `var`) and also in other functions, for example - `width: calc(var(--variable-width) + 20px);`

g-plane: For inside other functions like `width: calc(var(--variable-width) + 20px)`, this is already included.

alexander-akait (approved):

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_css_lints

PR #3891, opened 2022-03-07
perf(es/minifier): Add a fast-path for export merger

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #3893, opened 2022-03-07
fix(ecma/transform/typescript): remove import used in interface and type signature

Austaras:
**Description:**
The other part of #3827

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

I didn't get it fully, but it works :P

---

swc-bump:

Austaras:
`Visit` is only used to strip type only import, which means it only need to know if a import is referred in a value context, so the impl can be simplified to skip all idents in type context.

Issue #3863, opened 2022-03-05
`@swc/core` release binary does not support tracing


kwonoj:
### Describe the bug

Context: https://github.com/swc-project/swc/pull/3801#issue-1155154875

Due to we have compile time flag to disable all tracing logs, release build does not generate actual trace. Need to resolve internal blockers to enable tracings in the release build.

### Expected behavior

-

### Version

latest

kwonoj:
@kdy1 Mind share what exactl part swc relies on those flag and possible path forward for fix?

PR #3904, opened 2022-03-07
feat(trace): complete initial trace support for transforms

kwonoj:
**Description:**

This wraps up initial trace spans for the transforms. May need to add addtional spans for some specific paths, but that's something we need to figure out with testing / feedbacks.

https://github.com/swc-project/swc/issues/3863 is still blocker to enable user testing for this. Once it's resolved, user should be able to try with release build of `@swc/core`. Other features like minifier, bundle, etcs would need traces as well but it looks like we'd better to stablize (especially resolving #3863) before try to add more spans.

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_transforms_compat/src/es3/member_expr_lits.rs
+++ b/crates/swc_ecma_transforms_compat/src/es3/member_expr_lits.rs
@@ -1,6 +1,7 @@
 use swc_ecma_ast::*;
 use swc_ecma_utils::is_valid_ident;
 use swc_ecma_visit::{noop_fold_type, Fold, FoldWith};
+use swc_trace_macro::swc_trace;
 
 /// babel: `transform-member-expression-literals`
 ///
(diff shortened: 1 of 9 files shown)


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_transforms_compat

crates/swc_ecma_transforms_compat/src/es3/member_expr_lits.rs
```diff
@@ -19,12 +19,14 @@ use swc_ecma_visit::{noop_fold_type, Fold, FoldWith};
...
 /// obj["var"] = "isKeyword";
 /// ```
+#[tracing::instrument(level = "trace", skip_all)]
 pub fn member_expression_literals() -> impl Fold {
     MemberExprLit
 }
 #[derive(Default, Clone, Copy)]
 struct MemberExprLit;
 
+#[swc_trace]
```
kdy1, line 29: Please add import for this.

kwonoj:
🤔 wasm/test fails with swc_trace macro, lemme peek

kdy1 (changes requested):
CI catched it

crates/swc_ecma_transforms_compat/src/es3/member_expr_lits.rs
kwonoj: 🤦 Thanks for the catch.

kdy1 (approved):

PR #3905, opened 2022-03-07
Feat more things in prefixer

alexander-akait:
**Description:**

prefix more properties +  bug fixes

Also I want to move all inlined tests to `fixtures`, because it is more easy to look and review, but I will do it separately  

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_stylis

PR #3909, opened 2022-03-07
fix(es/module): export update expr for bigint

magic-akari:
**Description:**

For bigint,
`ident ++` cannot be transformed into `ident = ident + 1`

```javascript
export let foo = 1n;

foo++;
```

**BREAKING CHANGE:**


crates/swc_ecma_transforms_module/tests/common_js.rs
```diff
@@ -2470,7 +2470,7 @@ let diffLevel = 0;
 exports.diffLevel = diffLevel;
 
 function diff() {
-  if (!(exports.diffLevel = diffLevel = +diffLevel - 1)) {
+  if (!(--diffLevel, exports.diffLevel = diffLevel)) {
```
kdy1, line 2473: Is this the correct change?
I checked babel but babel does not do this

magic-akari: Babel's behavior is not right, I will open an issue later.
The current behavior of this PR is equivalent to the (fixed) babel behavior with `assumptions.pureGetters=true`.

magic-akari: This case looks like it could be changed to `exports.diffLevel = --diffLevel`.
I'll give it a try.

kdy1 (dismissed):
LGMT. Thasnk!

---

swc-bump:
 - swc_ecma_transforms_module

magic-akari:
It seems that the export order is not stable.

kdy1 (approved):
LGTM.

---

swc-bump:
 - swc_ecma_transforms_module

PR #3912, opened 2022-03-07
chore(react): add example for transforms_react

ruiramos:
**Description:**

Adds an example script to the `swc_ecma_transforms_react` that
demonstrates how to parse and transform a JavaScript file that uses React and JSX.

Also adds a test jsx file, and the `swc` package to the dev-dependency
list of `swc_ecma_transforms_react` as that's needed to run the example
script.

### DIFF
--- a/crates/swc_ecma_transforms_react/Cargo.toml
+++ b/crates/swc_ecma_transforms_react/Cargo.toml
@@ -35,3 +35,4 @@ swc_ecma_transforms_compat = {version = "0.78.0", path = "../swc_ecma_transforms
 swc_ecma_transforms_module = {version = "0.89.0", path = "../swc_ecma_transforms_module"}
 swc_ecma_transforms_testing = {version = "0.68.0", path = "../swc_ecma_transforms_testing"}
 testing = {version = "0.18.0", path = "../testing"}
+swc = {version = "0.145.1", path = "../swc"}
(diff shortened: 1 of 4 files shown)


crates/swc_ecma_transforms_react/Cargo.toml
```diff
@@ -35,3 +35,4 @@ swc_ecma_transforms_compat = {version = "0.78.0", path = "../swc_ecma_transforms
 swc_ecma_transforms_module = {version = "0.89.0", path = "../swc_ecma_transforms_module"}
 swc_ecma_transforms_testing = {version = "0.68.0", path = "../swc_ecma_transforms_testing"}
 testing = {version = "0.18.0", path = "../testing"}
+swc = {version = "0.145.1", path = "../swc"}
```
kdy1, line 38: This will block publishing.

kdy1: Please remove this

kdy1 (changes requested):
I found a serious issue

crates/swc_ecma_transforms_react/Cargo.toml
ruiramos: Hey thanks for the quick feedback. 
I can remove that however I believe that you won't be able to run the example file without it, as it will be missing imports from that crate (`Compiler` and `Options`)

kdy1: Compiler is not require to use this crate

kdy1:
Closing as it's not being worked on

PR #3914, opened 2022-03-08
feat(css/lints): Implement `property-no-vendor-prefix` rule

g-plane:
**Description:**

Implemented [`property-no-vendor-prefix`](https://stylelint.io/user-guide/rules/list/property-no-vendor-prefix) rule.

**BREAKING CHANGE:**

No.

### DIFF
--- a/crates/swc_css_lints/src/rules/property_no_vendor_prefix.rs
+++ b/crates/swc_css_lints/src/rules/property_no_vendor_prefix.rs
@@ -1,0 +1,57 @@
+use serde::{Deserialize, Serialize};
+use swc_css_ast::*;
+use swc_css_visit::{Visit, VisitWith};
+
+use crate::{
+    dataset::strip_vendor_prefix,
+    error::ConfigError,
+    pattern::NamePattern,
+    rule::{visitor_rule, LintRule, LintRuleContext},
+};
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PropertyNoVendorPrefixConfig {
+    ignore_properties: Option<Vec<String>>,
+}
+
+pub fn property_no_vendor_prefix(
+    ctx: LintRuleContext<PropertyNoVendorPrefixConfig>,
+) -> Result<Box<dyn LintRule>, ConfigError> {
+    let ignored = ctx
+        .config()
+        .ignore_properties
+        .clone()
+        .unwrap_or_default()
+        .into_iter()
+        .map(NamePattern::try_from)
+        .collect::<Result<_, _>>()?;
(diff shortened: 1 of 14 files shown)


crates/swc_css_lints/src/rules/property_no_vendor_prefix.rs
```diff
@@ -0,0 +1,57 @@
...
+    ctx: LintRuleContext<PropertyNoVendorPrefixConfig>,
+
+    ignored: Vec<NamePattern>,
+}
+
+impl Visit for PropertyNoVendorPrefix {
+    fn visit_declaration(&mut self, declaration: &Declaration) {
+        if let DeclarationName::Ident(Ident { value, .. }) = &declaration.name {
+            match strip_vendor_prefix(&value.to_lowercase()) {
+                Some(unprefix) if self.ignored.iter().all(|pat| !pat.is_match(unprefix)) => {
```
alexander-akait, line 46: hm, stylelint has another logic here due https://github.com/stylelint/stylelint/blob/main/lib/rules/property-no-vendor-prefix/index.js#L57, i.e. stylelint only reports known properties where no prefixes are needed, we report on all prefixed properties, right now logic for prefixes in `swc_stylis` (we want to rename create to `swc_css_prefixer`) and I think we can put logic about known prefixes there and here reuse it, so when new properties will added this rule will know about new property.

You can help us and start it. What do you think?

g-plane: That's a nice idea, but how can I help?

alexander-akait: @g-plane I want to finish test refactor and bug fixes today in prefixer logic (currently logic in `swc_stylis`), the next tasks:
- rename it to `swc_css_prefixer`
- implement public API to get all known prefixable properties (and more, for selectors/at-rules/etc), it should be not hard
- implement browserslist support

I will ping you in PR when I finish it so you can help with renaming, public API and browserslist support

thanks for help

kdy1:
What's the state of this PR?
Is this ready to be merged?

g-plane:
Maybe no.

alexander-akait:
No, I am continue working on some required stuff, sorry for delay

alexander-akait:
@g-plane Let's keep open, sorry for delay, was small busy other tasks, we should finish it and return to this

alexander-akait:
@g-plane Sorry for big delay, really a lot of tasks, I have finished, now we have public https://github.com/swc-project/swc/blob/main/crates/swc_css_prefixer/src/prefixer.rs#L92 so you can use it

g-plane:
Sorry I don't have enough time to go on. I'm closing now.

PR #3916, opened 2022-03-08
add Cargo.lock to .gitignore

tamaroning:
**Description:**
.gitignore generated by `swc plugin new ...` lacks Cargo.toml

**BREAKING CHANGE:**
none

none


kdy1:
I will never accept this PR.
SWC even uses cargo-crev

kdy1:
Btw, why are you trying this?

kdy1:
Ah, sorry. I misread it because there was no decription.
But I'll not accept this, as it should not be ignored.

PR #3917, opened 2022-03-08
feat(es/parser): only allow abstract member in abstract class

Austaras:
**Description:**
And also don't allow abstarct prop to have initializer and abstarct member to have body

**BREAKING CHANGE:**


kdy1 (approved):
LGTM. Thank you!

---

swc-bump:
 - swc_ecma_parser

PR #3918, opened 2022-03-08
feat(es/transforms/base): remove extra parens around iife in stmt

Austaras:
**Description:**

mainly to reduce
```js
function foo() {
  return (function () {})()
}
```

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---


swc-bump:
 - swc_ecma_transforms_base

PR #3922, opened 2022-03-09
feat(es/parser): Relax MSRV requirement

kdy1:
**Description:**

Seems like latest rustc has some regression.
So I reverted rustc upgrade, but it requires some fix of SWC.

 - https://github.com/vercel/next.js/pull/35133

 - https://github.com/vercel/next.js/runs/5474698598?check_suite_focus=true


kdy1 (commented):
swc-bump:
 - swc_ecma_parser

PR #3925, opened 2022-03-09
feat(es/minifier): remove pure calls

Austaras:
**Description:**

remove call/new/tagged tempalte with `/*#__PURE__*/` comments. However
```js
class A extends B {}
```
still cannot be removed in es5 target. It's because we use exact same ident for the `var A` and `function A`
```js
var A = function(B) {
    "use strict";
    _inherits(A, B);
    var _super = _createSuper(A);
    function A() {
        _classCallCheck(this, A);
        return _super.apply(this, arguments);
    }
    return A;
}(B);
```
However, we cannot change any of them. If we change `var A` to private_ident, all reference outside class would be broken. And if we set `function A` to private_ident, function name would unnecessarily change.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_minifier

PR #3927, opened 2022-03-09
fix(es/module): correct return value of suffix update expression

magic-akari:
**Description:**

A regression is introduced in #3909.
This pull request correct the return value of exported suffix update expression.

**BREAKING CHANGE:**


kdy1 (approved):
LGTM. Thank you!

---

swc-bump:
 - swc_ecma_transforms_module

PR #3933, opened 2022-03-09
Fix todos

alexander-akait:
**Description:**

Fix bugs and handle more values in properties.

**BREAKING CHANGE:**

No

No


crates/swc_stylis/tests/fixture/animation/output.css
```diff
@@ -40,8 +40,7 @@ animation-direction: alternate-reverse;
 -moz-animation: rotation-reverse;
 -o-animation: rotation-reverse;
 animation: rotation-reverse;
--webkit-animation: rotation-reverse reverse;
--moz-animation: rotation-reverse reverse;
--o-animation: rotation-reverse reverse;
```
alexander-akait, line 45: We should not prefix them, because reverse is part of new spec and will not work in prefixes

kdy1 (approved):
LGTM. Thanks!

---

swc-bump:
 - swc_css_utils
 - swc_stylis

PR #3937, opened 2022-03-10
fix(es): Don't create `.swc` if not required

kdy1:
**Description:**

**BREAKING CHANGE:**


kdy1 (commented):
swc-bump:
 - swc

PR #3939, opened 2022-03-10
feat(es/transform/compat): add pure comments for class fields keys

### LINKED ISSUE #3718: Private name generated weak map/set should be marked as pure
### Describe the bug

So compressor could remove them

### Input code

```typescript
class A {
    #b = 123
}
```

### Config

_No response_

### Playground link

https://play.swc.rs/?version=1.2.144&code=H4sIAAAAAAAAA0vOSSwuVnBUqOZSAALlJAVbBUMjY65aAMgGJ8gYAAAA&config=H4sIAAAAAAAAA21TQZLjIAz8C%2BccdvYwhzxgq%2BawtU%2BgCAiHrAGXJDJxpfL3FbaD7WRvqOkWUkvc1YWsOt7VYJAA64nGxOamjgpsNGQxDKwOQhPIm57goAYMV8Pwlf6cLmBZHRkLPA6KDXbAVUk%2Ff3x8iqrPmaDpYkjBj%2FUNm%2BOAQFTPIioRElPjGcT8TXPagzrl3INJb7E2pENi6ABXqc19bwYCfTXYFPU1g4Fy2kGFwekB87CVJxc45CRhAx0Yp212ayMuoPQdrrAhSR4hJZLi9qCDU%2Bm66u38NFxNX8S%2BFt8mK%2BTRpjvnQKx9SfQCLdXOwhmaG11YwWsELpienEsOaefFXwCpszdEycRN%2FRPuZRZ7pv8PKyQvg%2BSxoTLltaoEnTSnQ%2FCtw1o1IAdoJARXLNQO7Zp8AXcNUXCgwXuxu4npO7A9b6ricYDs273YZvw6hDnUz6Xao3UP3sBfUhZv57Hg0fD5FaMxnnL%2FliICn7N7g8UJzq8gygrfhle0JAfiM7jNRaEaLr8tmtS1XXtUILtSgftkyPQNP9Vj%2FXbPHaHfC3FK9A8VB8qvAgQAAA%3D%3D

### Expected behavior

should output empty string

### Actual behavior

output `new WeakMap()`

### Version

1.2.144

### Additional context

_No response_

Austaras:
**Description:**

**BREAKING CHANGE:**

closes #3718


kdy1:
Does this close https://github.com/swc-project/swc/issues/3718?

kdy1 (approved):
swc-bump:
 - swc_common
 - swc_ecma_transforms_compat --breaking

Austaras:
Oh, comments for dummy span needs to be preserved for other tools to consume.

crates/swc_common/src/syntax_pos.rs
```diff
@@ -423,6 +434,20 @@ impl Span {
             }
         }
     }
+
+    /// Dummy span, both position are extremely large numbers so they would be
+    /// ignore by sourcemap, but can still have comments
+    pub fn dummy_with_cmt() -> Self {
+        GLOBALS.with(|globals| {
+            globals.dummy_cnt.fetch_add(1, Ordering::SeqCst);
```
kdy1, line 442: Can't we use the return value of this?
We can add `1` after fetching it.

crates/swc_common/src/syntax_pos.rs
```diff
@@ -237,6 +242,12 @@ impl Span {
     /// Returns `true` if this is a dummy span with any hygienic context.
     #[inline]
     pub fn is_dummy(self) -> bool {
+        self.lo.0 == 0 && self.hi.0 == 0 || self.lo.0 >= DUMMY_RESERVE
+    }
+
+    /// Returns `true` if this is a dummy span with any hygienic context.
+    #[inline]
+    pub fn is_dummy_with_cmt(self) -> bool {
```
kdy1, line 250: I think this name is the reverse of the actual impl?

How about `is_dummy_ignoring_cmt`?

crates/swc_common/src/syntax_pos.rs
kdy1: Seems like you missed this?

Austaras: I don't know much about atomics, but if two threads fetch_add at same time, wouldn't they get same return value?

kdy1: No, fetch-add is a single atomic instruction and if `atomic` here means all operations of multiple threads are synchornized, and the unit is `fetch + add`.

So other thread cannot perform anything about this data (in memory) between `fetch` and `add`

kdy1 (approved):
Thanks!

PR #3946, opened 2022-03-10
feat(common): Implement a pretty error reporter

### LINKED ISSUE #2860: Pretty-formatted diagnostics
### Describe the feature

Context: https://discord.com/channels/889779439272075314/889780921002246154/912939004159221790

Current diagnostics output from swc is good. But I'm curious if we have prettier output similar to what rust cli, or some other cli offer. Something looks like https://twitter.com/_lrlna/status/1462797187276939268 can be a good reference example. 

Actually above error diagnostics is from https://github.com/zkat/miette - if we could adapt it without heavy modification it may be a good way to go. Unfortunately we may need features like https://github.com/zkat/miette/issues/64 as far as know to utilize it. 

This issue is tracking purpose to casually look for possible approaches to achieve diagnostics report. It should be fine to use some external libs or if we build it for ourselves. @kdy1 suggested 

> we may implement ansi Handler and result would be similar

as possible alternative we can consider.

_No response_

### Additional context

_No response_

kdy1:
**Description:**

 - Closes #2860.

# Preview

image


kdy1 (commented):
swc-bump:
 - swc --breaking
 - testing --breaking

PR #3952, opened 2022-03-10
chore: refine error report text

magic-akari:
**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc/tests/errors/lints/const-assign/1/output.swc-stderr
+++ b/crates/swc/tests/errors/lints/const-assign/1/output.swc-stderr
@@ -1,12 +1,12 @@
-error: Cannot reassign to a variable declared with `const`
- 
-  |
-5 | a = 600;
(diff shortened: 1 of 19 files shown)


kdy1:
How do you think about https://github.com/swc-project/swc/pull/3946?
Or alternatively, updating emitter writer to match recent version of rustc

kdy1 (approved):
LGTM, regardless of the new error reporter.

Thank you!

---

swc-bump:
 - swc_common
 - swc_ecma_lints

magic-akari:
I didn't know you moved to the miette.
In the last two days I tried to follow the latest rustc but did not manage to migrate.
Rust has its own complexity and using a library like miette is a better choice.

kdy1:
Migration to miette will be gradual, as it requires lots of snapshot updates to rust-side users of swc

PR #3961, opened 2022-03-11
fix(es/parser): Parse if-else chain iteratively

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
  - swc_ecma_parser

PR #3969, opened 2022-03-11
chore(es/parser): Add some inline attributes

kdy1:
**Description:**

I tried delegating deallocation to another thread, but it resulted in a degraded performance because deallocating something from the other thread requires more atomic operations.
The doc of mimalloc says it's single CAS but it was slow enough


kdy1 (commented):
swc-bump:
 - swc_ecma_parser

Issue #3934, opened 2022-03-09
Support plugins in `@swc/wasm-*`


kwonoj:
### Describe the feature

The stories around supporting plugins in SWC (https://github.com/swc-project/swc/discussions/3540) did not cover wasm host version of the SWC, namely `@swc/wasm` and `@swc/wasm-web`. Currently, wasm host version of SWC lacks of _any_ kind of plugin support including existing JS based one. This issue tracks to support plugins universally across any runtime we support, including wasm host version.

The scope of plugin support is limited to the new, experimental plugin written in wasm. I do not have conclusions if it's technically possible to support JS written plugin or not but in either case we do not aim to support it in here. 

Plugin support in wasm host have working proof of concept at https://github.com/kwonoj/swc/tree/feat-plugin-js-wasm-host nowhere close to ready to be included in core. There are lots of problems need to be solved separately, however it proves main concept of let internal webassembly runtime we use can `borrow` host runtime's webassembly compiler if it runs on the supported platform. Notably, this means web brower & node.js.

These are known subtasks to achieve initial working experimental support.

- [ ] https://github.com/swc-project/swc/issues/3591 : to support wasm32-unknown-unknown plugin.
- [ ] Fix memory access issue in `PluginTransformTracker` : Somehow accessing guest memory ref fails only with PluginTransformTracker's init logic, works fine with raw poc code.
- [ ] Solve compilation config to support `wasmer` and `wasmer/js` selectively
- [ ] Non-prod ready but support loading wasm binary dynamically
- [ ] Wasm32-wasi plugin support story: likely this is upstream (browser / node) runtime issue.

### Babel plugin or link to the feature description

https://github.com/swc-project/swc/discussions/3540

kwonoj:
Upstream issues:

- https://github.com/wasmerio/wasmer/issues/2752
- https://github.com/servo/string-cache/issues/257
- https://github.com/wasmerio/wasmer/issues/2816

magic-akari:
> - [ ] Wasm32-wasi plugin support story: likely this is upstream (browser / node) runtime issue.

Does this "story" refer to https://github.com/WebAssembly/component-model ?

kwonoj:
> Does this "story" refer to [WebAssembly/component-model](https://github.com/WebAssembly/component-model) ?

[...]

PR #3977, opened 2022-03-11
build(cargo): setup plugin_runner features

kwonoj:
**Description:**

**This doesn't work currently**, opening up PR early to discuss & resolve build failures, or determine if we need some other way to workaround this instead of relying on cargo's workspace feature selection.

Basic structures of feature selection we'd like to have is

```
node
├─ swc/plugin
│  ├─ swc_plugin_runner/default
│  │  ├─ wasmer/sys

swc_cli
├─ swc/plugin
│  ├─ swc_plugin_runner/default
│  │  ├─ wasmer/sys

wasm
├─ swc/plugin_host_wasm_env
│  ├─ swc_plugin_runner/wasm
│  │  ├─ wasmer/js
```

Native entrypoint (swc_cli, node) wants to pick up wasmer/sys, and wasm target wants to pick up wasmer/js instead. However this looks like hitting https://github.com/rust-lang/cargo/issues/4463 and cargo attempts to unify all of the features at once for wasmer.

https://github.com/swc-project/swc/issues/3934

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -887,8 +887,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7fcd999463524c52659517fe2cea98493cfe485d10565e7b0fb07dbba7ad2753"
 dependencies = [
  "cfg-if 1.0.0",
+ "js-sys",
  "libc",
  "wasi 0.10.2+wasi-snapshot-preview1",
+ "wasm-bindgen",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


kdy1:
```
wasm
├─ swc/plugin_host_wasm_env
│  ├─ swc_plugin_runner/wasm
│  │  ├─ wasmer/js
```

I think this approach will not work.
Instead, we should enable it using `---features` flag only while publishing the `wasm ` packages.

kwonoj:
> we should enable it using --features flag only while publishing the wasm packages.

This is part I was not sure bit - do you mean via cli flag for wasm-pack? and by default `crates/wasm` try to compile with `wasmer/sys`?

kwonoj:
This'll need revisit after https://github.com/swc-project/swc/pull/3978 goes in.

PR #3978, opened 2022-03-11
build(cargo): set workspace default members

kwonoj:
**Description:**

Possible approaches for https://github.com/swc-project/swc/pull/3977 : 

It's non-fancy workaround. Basically, omit offending package (node / wasm) from default workspace build selection. `cargo build` skips those and we have to manually run each build scripts (`npm run build` for /node, `scripts/build.sh` for /wasm). To get the right binary those scripts are needed anyway, however I'm not sure about implications like `cargo check` _may_ miss to run over those packages. 

The other approach is having better build system to control build scripts instead of cargo. Given the scope of the work we need to achieve here, that's something I'd like to defer for now unless it shows visible gains for us.


kwonoj:
Intentionally mark as draft as this is depends on #3977.

kwonoj:
Thinking this bit more, I came to feel this is probably easiest way to resolve cargo feature selection issue (https://github.com/swc-project/swc/pull/3977)

- default cargo build won't break due to specific conflicting feature selection (wasmer/*)
- we have custom build scripts for those specific targets already (`npm run build` / `scripts/build.sh`) anyway
- does not involve heavier solutions like custom scripts, build systems, or other tooling for now.

Thoughts? concerns?

kdy1:
I agree, it's the simplest way I can think of.

kwonoj:
Cool, will rebase & convert to pr.

kdy1 (approved):
swc-bump:
 - ''

PR #3980, opened 2022-03-11
build(cargo): skip unnecessary package for the benchmark

kwonoj:
**Description:**

There are few packages we don't need to build run benchmark, updated gh workflow to skip it.

Also, not sure if this is intended or not - but there are failing benchmark cases with `--all`. PR attempts to fix it to run the benchmark properly.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -58,7 +58,7 @@ jobs:
           override: true
 
       - name: Run benchmark
-        run: cargo bench --all --exclude swc_plugin | tee output.txt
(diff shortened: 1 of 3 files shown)


.github/workflows/bench.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           override: true
 
       - name: Run benchmark
-        run: cargo bench --all --exclude swc_plugin | tee output.txt
+        run: cargo bench --workspace --exclude swc_plugin --exclude wasm --exclude swc_cli --exclude node | tee output.txt
```
kwonoj, line 61: `all` -> `workspace`

```
--all
Deprecated alias for --workspace
```

kdy1 (approved):
Thank you!


---

swc-bump:
 - ''

PR #3984, opened 2022-03-12
perf(es/minifier): Use fxhash for integers

kdy1:
Description:

 - `Id` is treated as integer, as `JsWord` and `SyntaxContext` are both integer.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #3985, opened 2022-03-12
perf(common): Use `fxhash` everywhere

kdy1:
Description:

To avoid a breaking change, I created a cargo feature that can be used to change hasher.

This leads to about 8% perf improvements on m1 max macbook pro 64gb


kdy1 (commented):
swc-bump:
 - swc_common
 - swc_node_base

PR #3988, opened 2022-03-12
perf(es/minifier): Merge hashmap for scoping before checking

kdy1:
**Description:**

Currently, we check for child scopes by doing a recursive call, like https://github.com/swc-project/swc/blob/2ad0d801a2a9d84ae2c6df2d15ff8a215f296eec/crates/swc_ecma_minifier/src/pass/mangle_names/analyzer/scope.rs#L112-L114

But this is slow because it results in an enormous amount of calls.

We can optimize this by merging information of child scopes beforehand.

This makes huge difference, so I decided to include this in v1.2.154


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier
 - swc_fast_graph --breaking
 - swc_graph_analyzer --breaking

PR #3989, opened 2022-03-12
perf(es/minifier): refactor base54

magic-akari:
**Description:**

**BREAKING CHANGE:**


crates/swc_ecma_minifier/tests/mangle/reuse/3/output.js
```diff
@@ -1267,9 +1267,9 @@ class aa extends Array {
         };
     }
 }
-class ba {
+class ab {
```
kdy1, line 1270: Is this intentional?
Just curious, as the previous impl may be wrong.

magic-akari: Yes.
It generate 
a, b, c ... z, A, B, C ... Z, $, _, 0,1, ... 9,
and
aa, ab, ac ...

In previous implementations, it seems that numbers were never used at the end.

magic-akari:
Some of the usize values exceed the wasm limit.
Does this mean that we will never reach those values?

kdy1:
I think so?
I'm sure, the process will be killed before we reach it.

kdy1:
Oh then we can use smaller integer type

magic-akari:
Should I remove those values for WASM, or remove them for all targets?
The max safe value for WASM is 4294967295, the max u32.

kdy1:
I think both are fine, but I prefer removing from both target.

crates/swc_ecma_minifier/src/util/base54.rs
```diff
@@ -1,46 +1,198 @@
...
+        x <<= 6;
+        x += reverse_base54(c);
+    }
+
+    ret += x;
+
+    ret
+}
+
+fn reverse_base54(c: &u8) -> usize {
```
kdy1, line 67: I think accepting `u8` will be better for performance.

kdy1 (changes requested):
Thank you!

I think this is a quite novel idea.

crates/swc_ecma_minifier/src/util/base54.rs
magic-akari: The `decode` and `reverse_base54` functions are currently not used anywhere except for testing.

kdy1: I see, then it's fine.
Thanks for the explanation

crates/swc_ecma_minifier/tests/mangle/reuse/3/output.js
magic-akari: Oh, numeric endings are possible in previous implementation.
Now it is easier to implement `encode` and `decode` after changing the rules.

crates/swc_ecma_minifier/src/util/base54.rs
```diff
@@ -1,46 +1,209 @@
...
+    98312762,   // false
+    321674951,  // super
+    335053132,  // throw
+    385347706,  // while
+    4211585658, // delete
+];
+
+// u32::MAX is 4294967295
+
+// const RESERVED_U64: [usize; 13] = [
```
Brooooooklyn, line 117: Are these commented codes necessary? If not, please delete them.

crates/swc_ecma_minifier/src/pass/compute_char_freq.rs
```diff
@@ -19,7 +19,7 @@ pub(crate) struct CharFreqInfo {
 pub(crate) fn compute_char_freq(_m: &Module) -> CharFreqInfo {
```
magic-akari, line 19: @kdy1 
CharFreqInfo change the `BASE54_DEFAULT_CHARS` order which will break the `is_reserved` function

kdy1: Ah...
Then I think it's fine to remove `CharFreqInfo` for now, as it does not have enough logic to be useful.

magic-akari: Would using CharFreqInfo achieve better gzip results?
If so, I'd prefer to keep it. 
I can generate the reserved list at runtime.

kdy1: Not yet at the moment, seems like we need full codegen or smart visitor to make the compression better.

kdy1 (approved):
LGTM. Thank you!

---

swc-bump:
 - swc_ecma_minifier

Issue #2630, opened 2021-11-02
Option `jsc.loose` don't actually passed to transformers when used in cli or playground


Austaras:
### Describe the bug

run following code in node

### Input code

```typescript
const swc = require('@swc/core')
swc
  .transform(
    `
for (const a of []) {}
`,
    {
      jsc: {
        target: 'es5',
        parser: {
          syntax: 'ecmascript'
        },
        loose: true
      }
    }
  )
  .then(output => {
    console.log(output.code)
  })
```

### Playground link

https://play.swc.rs/?version=1.2.106&code=H4sIAAAAAAAAA0vLL1LQSM7PKy5RSFTIT1OIjtVUqOZSUAAJ5eek6uXkp2skanLVAgD%2B%2FpMnKAAAAA%3D%3D&config=H4sIAAAAAAAAA0WNOwoDMQxE76LabVL4DjmEcLSLg39IWogxvnus4JBOzLx5GvCSAH5AQxZiu6QXxTd4oJBRAsem4Ba2ogOT0HSgyCepIXJbXapVCLzyRQ5yLPHoJgo1NyaRvVsVljPRT7M8uT4vCwZob%2FT13WH%2BHXsX5bFBezE%2F4BzHoLQAAAA%3D

### Expected behavior

Should generate loose code, but didn't

### Version

master

Austaras:
duplicate of #2216

PR #3990, opened 2022-03-12
fix(swc): merge `jsc.loose` and `jsc.assumptions`

Austaras:
**Description:**

Before this change, if swc is called from `transform` node api, `jsc.loose` and `jsc.assumptions` in second argument would be discarded. This would likely make loose switch in playground usable too.

**BREAKING CHANGE:**

maybe #2630


kdy1 (approved):
Thank you!

---
swc-bump:
 - swc

PR #3993, opened 2022-03-13
perf(es/minifier): Use `Vec<u8>` as a buffer for `base54`

kdy1:
Description:

We are only using ASCII characters so we can avoid utf8 logics by using `Vec<u8>` as a buffer and converting it into `String` at the end.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #3994, opened 2022-03-13
test(es): Add an execution test system

kdy1:
Description:

As we have a way to verify that input is behaving identically, we create a huge matrix of compile options and process the input with all of those options.

This system can be used to check for fixed issues easily.
For first, someone can add issues using a unix hidden directory, which starts with `.`. Those tests are ignored by `testing::fixture`.
Then, when something is changed, one can run `./scripts/check-issues.sh`. If a test succeeds, it means it's fixed and the starting dot (`.`) of the path to the test input file is removed.


kdy1 (commented):
swc-bump:
 - swc

PR #3996, opened 2022-03-13
feat(es/minifier): Compress more comparisons

kdy1:
**Description:**

I moved some code to the stateful optimizer as we need `ProgramData` due to the removal of member expressions.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #3997, opened 2022-03-13
fix(css/codegen): fix source maps

alexander-akait:
**Description:**

Fixes:
- fix source map for string and string like tokens and values
- fix mapping in some cases
- refactor code
- test

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Wow...
Thank you so much!

---

swc-bump:
 - swc_css_codegen --breaking

PR #4003, opened 2022-03-14
test(*): Print diagnostics to stderr while testing

kdy1:
**Description:**

This will make debugging stack overflow and parser issues easier because it writes to stderrr right away, instead of holding a buffer for it.


kdy1 (commented):
swc-bump:
 - testing

PR #4007, opened 2022-03-14
fix(common): Don't use `None` as source while generating source maps

### LINKED ISSUE #4006: Source maps error on static class methods
### Describe the bug

Running the following command:

```
swc src/app.ts -o lib/app.js --no-swcrc --source-maps -C jsc.target=es2019 -C jsc.parser.syntax=typescript
```

I get a failure with this message in the console:

```
Error: original.line and original.column are not numbers -- you probably meant to omit the original mapping entirely and only map the generated position. If so, pass null for the original mapping instead of an object with empty or null values.
    at SourceMapGenerator._validateMapping (C:\Users\Guy\Projects\app\node_modules\source-map\lib\source-map-generator.js:271:15)
    at SourceMapGenerator.addMapping (C:\Users\Guy\Projects\app\node_modules\source-map\lib\source-map-generator.js:107:12)
    at C:\Users\Guy\Projects\app\node_modules\@swc\cli\lib\swc\file.js:54:25       
    at Array.<anonymous> (C:\Users\Guy\Projects\app\node_modules\source-map\lib\source-map-consumer.js:382:19)
    at mapping_callback (C:\Users\Guy\Projects\app\node_modules\source-map\lib\wasm.js:64:52)
    at wasm://wasm/0002f8d6:wasm-function[30]:0x3cac
    at C:\Users\Guy\Projects\app\node_modules\source-map\lib\source-map-consumer.js:387:30
    at Object.withMappingCallback (C:\Users\Guy\Projects\app\node_modules\source-map\lib\wasm.js:95:11)
    at BasicSourceMapConsumer.eachMapping (C:\Users\Guy\Projects\app\node_modules\source-map\lib\source-map-consumer.js:371:16)
    at concatResults (C:\Users\Guy\Projects\app\node_modules\@swc\cli\lib\swc\file.js:52:26)
```

This seems to happen specifically for static class properties per the code example below.

### Input code

```typescript
import {html, css, LitElement} from 'lit';

export class App extends LitElement {
  name: string;

  static styles = css`p { color: blue }`;

  static properties = {
    name: { type: String },
  };

  constructor() {
    super();
    this.name = 'Somebody';
  }

  render() {
    return html`<p>Hello, ${this.name}!</p>`;
  }
}
customElements.define('app', App);
```

### Config

```json
<none>
```

### Playground link

_No response_

### Expected behavior

Should compile with correct source maps.

### Actual behavior

_No response_

### Version

CLI 0.1.55, Core 1.2.154

### Additional context

_No response_

kdy1:
**Description:**

I used https://pastcompute.github.io/ScriptMapper/ to dump source map, and found that line/column becomes null if file is `None`. Seems like the rust crate `sourcemap` is mature enough to handle sourcemaps with those entries, but many other sourcemap libraries, including the one used by `@swc/cli` expects non-null value for the line and the column.

 - Closes #4006


kdy1 (commented):
swc-bump:
 - swc_common

PR #4011, opened 2022-03-14
fix(swc): comment preservation needs to explicitly sort spans

dpchamps:
## Description:

Spans in the comment fixer visitor need to be sorted. Spans are typically collected in an ordered fashion which will produce expected results, but certain code will produce unordered span collection. JSX / TSX seems to exacerbate this problem.

### Solution:

Sort spans by lo byte positions before adjusting comments.

Included a test fixture to demonstrate that this fixes the general problem.


kdy1 (approved):
Thank you!

---

swc-bump:
 -  swc

PR #4012, opened 2022-03-14
refactor(plugin_runner): split modules, introduce feature target

kwonoj:
**Description:**

- https://github.com/swc-project/swc/issues/3934

This PR is internal refactoring to `plugin_runner` to support wasm host based target (`@swc/wasm`).

Notable changes are 

- plugin_runner now have `embedded_runtime`, `native_runtime` features. Embedded stands for existing plugin runner where we need to bring wasmer's runtime. Native means we'll use host runtime as-is, such as browser / node.js's v8 webassembly.
- `swc` supports those new features via `plugin_embedded_runtime` / `plugin_native_runtime` features. Existing `plugin` is renamed into `plugin_embedded_runtime`.
- `plugin_runner` extracts few logics into internal modules as well as branches to each features. Currently `native_runtime` is empty.
- `apply_js_plugin` is now renamed to `apply_transform_plugin` to reflect what we do exactly. Further, we can possibly have other entrypoint like `apply_xxx_plugin` for css, or else.


.github/workflows/cargo-lints.yml
```diff
@@ -50,4 +50,5 @@ jobs:
 
       - name: Check
         run: |
-          cargo clippy --all
+          cargo clippy
+          cargo clippy -p node
```
kwonoj, line 54: with `wasmer/js`, it was not possible to run build / check against non-wasm target to `crates/wasm` anymore - specifically in this case wasmer doesn't bring its transitive deps for non-working target (https://github.com/wasmerio/wasmer/blob/master/lib/api/Cargo.toml#L58-L62). Not sure if there's way to override, but also it feels unreasonably complex work to bring few workflows.

kwonoj:
lint failure comes from

```
  Checking swc_ecma_transforms_optimization v0.102.0 (/home/runner/work/swc/swc/crates/swc_ecma_transforms_optimization)
error: useless conversion to the same type: `simplify::dce::Data`
   --> crates/swc_ecma_transforms_optimization/src/simplify/dce/mod.rs:410:21
    |
410 |         self.data = data.into();
    |                     ^^^^^^^^^^^ help: consider removing `.into()`: `data`
    |
note: the lint level is defined here
   --> crates/swc_ecma_transforms_optimization/src/lib.rs:1:9
```

.github/workflows/cargo-lints.yml
syrusakbary: Just read this. Happy to accept contributions to improve this! :)

crates/swc_plugin_runner/src/load_plugin.rs
```diff
@@ -0,0 +1,152 @@
...
+                        "__mark_is_builtin_proxy" => mark_is_builtin_fn_decl,
+                        "__mark_set_builtin_proxy" => mark_set_builtin_fn_decl,
+                        "__mark_is_descendant_of_proxy" => mark_is_descendant_of_fn_decl,
+                        "__mark_least_ancestor" => mark_least_ancestor_fn_decl,
+                        "__syntax_context_apply_mark_proxy" => syntax_context_apply_mark_fn_decl,
+                        "__syntax_context_remove_mark_proxy" => syntax_context_remove_mark_fn_decl,
+                        "__syntax_context_outer_proxy" => syntax_context_outer_fn_decl
+                    }
+                }
+            };
```
syrusakbary, line 144: Rather than if/else you can probably chain the WASI imports along with your custom ones.
Basically, creating your imports:

```rust
                imports! {
                    "env" => {
                        "__set_transform_result" => set_transform_result_fn_decl,
                        "__emit_diagnostics" => emit_diagnostics_fn_decl,
                        "__mark_fresh_proxy" => mark_fresh_fn_decl,
                        "__mark_parent_proxy" => mark_parent_fn_decl,
                        "__mark_is_builtin_proxy" => mark_is_builtin_fn_decl,
                        "__mark_set_builtin_proxy" => mark_set_builtin_fn_decl,
                        "__mark_is_descendant_of_proxy" => mark_is_descendant_of_fn_decl,
                        "__mark_least_ancestor" => mark_least_ancestor_fn_decl,
                        "__syntax_context_apply_mark_proxy" => syntax_context_apply_mark_fn_decl,
                        "__syntax_context_remove_mark_proxy" => syntax_context_remove_mark_fn_decl,
                        "__syntax_context_outer_proxy" => syntax_context_outer_fn_decl
                    }
                }
```

And then detecting if it's a WASI module, and chaining the WASI imports to it with:

https://docs.rs/wasmer/latest/wasmer/trait.ChainableNamedResolver.html#method.chain_front

(that would make the code probably a bit simpler)

kwonoj: @syrusakbary Thanks for the tip. Does this https://github.com/swc-project/swc/pull/4012/files#diff-b1fd8abab92d7f7db02dccde2eb021904f6e215ab0fb78553e197ef65fd7a058R102-R118 looks correct? seems it's working as expected at least.

.github/workflows/cargo.yml
```diff
@@ -118,7 +120,7 @@ jobs:
           - crate: swc
             os: ubuntu-latest
             check: |
-              cargo hack check --feature-powerset --no-dev-deps
+              cargo hack check --feature-powerset --no-dev-deps --exclude-features plugin_native_runtime
```
kdy1, line 123: Please modify https://github.com/swc-project/swc/blob/5800737752ec8a2befc7010290247b319e671ac3/scripts/github/tests.yml instead

crates/swc/Cargo.toml
```diff
@@ -27,8 +27,11 @@ concurrent = [
 ]
 debug = ["swc_ecma_visit/debug"]
 node = ["napi", "napi-derive"]
-plugin = [
-  "swc_plugin_runner",
+plugin_embedded_runtime = [
```
kdy1, line 30: Why is this required?

I think user of the `swc` crate can depend on `swc_plugin_runner/embedded_runtime` or `swc_plugin_runner/native_runtime` instead?

kdy1 (changes requested):
Thanks!

.github/workflows/cargo.yml
kdy1: You can create matrix by running https://github.com/swc-project/swc/blob/5800737752ec8a2befc7010290247b319e671ac3/scripts/github/create-matrix.sh

crates/swc_plugin_runner/src/load_plugin.rs
syrusakbary: Yup, looks correct!

crates/swc/Cargo.toml
kwonoj: This is probably I'm making some mistakes if you felt bit odd. My thought was each package have deps to `swc` can select features without transitive deps feature selection, also swc itself uses own feature selection like https://github.com/swc-project/swc/pull/4012/files#diff-8987cc214ab0c7d63b833d31785ecb48d89d1bcc204d42508c26ecd72becc7b9R72 .

.github/workflows/cargo.yml
kwonoj: I tried to ran `create-matrix.sh` and nothing happens. What should be exactly updated to make sure script works?

kdy1: You should copy output to the matrix to this file

kwonoj: Output doesn't seem to correct to me. Looks like script requires jq / yq, but after installing both I'm getting

```
...
parse error: Invalid numeric literal at line 1, column 6
parse error: Invalid numeric literal at line 1, column 6
parse error: Invalid numeric literal at line 1, column 6
- crate: testing_macros
  os: ubuntu-latest
parse error: Invalid numeric literal at line 1, column 6
parse error: Invalid numeric literal at line 1, column 6
parse error: Invalid numeric literal at line 1, column 6
- crate: wasm
  os: ubuntu-latest
parse error: Invalid numeric literal at line 1, column 6
parse error: Invalid numeric literal at line 1, column 6
parse error: Invalid numeric literal at line 1, column 6
```

kdy1: Are you using windows?
Interesting as it specifies `bash` requirement... I don't think `bash` parse script differently based on the OS

kwonoj: No, it was linux and managed to made work - script requires to be executed cwd under `scripts`, does not allow `./scripts/create-matrix.sh` from swc's root. It wasn't clear as there wasn't way to see error outputs.

crates/swc/Cargo.toml
kdy1: I think this is not required. I'll check using a fork

kdy1: I think https://github.com/swc-project/swc/pull/4027 will work, except `swc_cli` on platforms not supported by `cranelift`. We should exclude them via `--no-default-features`

kwonoj:
Superseded by https://github.com/swc-project/swc/pull/4035.

Issue #2385, opened 2021-10-09
In non-strict mode, the output code does not contain helpers correctly 


autoxbc:
**Input code**
```js
const swc = require('@swc/core');
const text = `
	const obj = { v: 42 };
	with(obj)
		console.log(typeof v);
`;
const { code } = swc.transformSync( text , { isModule: false } );
console.log(code);
```
node.js output
```js
var obj = {
    v: 42
};
with (obj)console.log(typeof v === "undefined" ? "undefined" : _typeof(v));
```
**Expected behavior**
The output code should contain helpers, such as **_typeof**

**Version**
The version of @swc/core: 1.2.95

PR #4025, opened 2022-03-15
fix(es/helpers): Inject helpers for scripts

kdy1:
**Description:**

This PR adds support for helpers in script mode. For now, we does not support `externalHelpers` for scripts, as there is no standard for importing other files from a script.

 - Closes https://github.com/swc-project/swc/issues/2385


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base

Issue #3006, opened 2021-12-10
Block scoping of const is broken when transpiling to ES5


damiantew:
### Describe the bug

Attached code incorrectly transpiles in SWC target ES5:
6 is logged 7 times instead of each number logged once as per const scoping behaviors in ES6.
Wrapping with the async function appears to be necessary to trigger the error

### Input code

```typescript
async function a() {
    for (const number_in_a_sequence of Array.from(new Array(7), (_,i) => i)) {
        setTimeout(() => console.log(number_in_a_sequence), 10)
    }
}
a()
```

### Config

```json
default playground configuration
```

### Playground link

https://play.swc.rs/?version=1.2.119&code=H4sIAAAAAAAAA22OzQrCMBCE74W%2Bwxw3UIqePCn4Dt5DDBsJtLuYH0oR393UgieHuQwD34zLq3iEKr5EFTgyePUdmoImkFfJBVLnOycbxTqb%2BVlZPEMDrim5dQxJZxJe9kgnM4DsEA3OF0Tz423KXG5xZq2F6NtvfJ14nPRB%2F1Ya63gwO%2BDdd83t4Qe5CcJstAAAAA%3D%3D&config=H4sIAAAAAAAAA0WMTQrEIAxG75K1286id5hDBCctFv9IUhgR714tlu7C915ehUMsrBUyshCPS0pU%2FMMKZAOKZZcVTNf6tKEXagYUeScdiiyd%2BZSEJjUQXHRbGSWbQmYSeRHG3T9m66GQfucYKmjJdAc%2F0N7G%2FHPynaLySe0Cj9ke9LUAAAA%3D

### Expected behavior

each number from 0 to 6 should be logged once as per const binding behaviors in ES6.

### Version

1.2.119

kwonoj:
master seems to be fixed, need double confirm once new version's published.

kdy1:
I investigated this, and the cause was that `regenerator` pass not considering block scoping of `const number_in_a_sequence`.

<img width="842" alt="image" src="https://user-images.githubusercontent.com/29931815/158376327-740128d5-0434-4133-8b66-2d3cef42504a.png">

PR #4029, opened 2022-03-15
fix(es/compat): Change pass ordering of `es2015`

kdy1:
**Description:**

We should apply `regenerator` after `block_scoping` because `regenerator` does not know how to handle `const`s.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/3006


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_compat

Issue #4032, opened 2022-03-15
 WARN import rewriter: failed to resolve: index not found


kshahani:
### Describe the bug

Recently, the logs when starting SWC are flooded by this message. The only thing I can think of that has changed is that I've updated to Big Sur over the weekend.

There are no paths to indicate where this issue is occurring, and I haven't found much other information.

It doesn't seem to break the build, so just being able to suppress these warnings would be great.

### Config

```json
{
  "jsc": {
    "loose": true,
    "target": "es2020",
    "externalHelpers": true,
    "keepClassNames": true,
    "parser": {
      "syntax": "typescript",
      "decorators": true
    },
    "transform": {
      "legacyDecorator": true,
      "decoratorMetadata": true
    },
    "baseUrl": "./src",
    "paths": {
      "~*": ["./*"]
    }
  },
  "module": {
      "type": "commonjs"
  },
  "sourceMaps": true
}
```

### Expected behavior

No log flood.

### Version

1.2.110

kdy1:
Oh, https://github.com/swc-project/swc/blob/cc2aa361fa6fbd6819966d709c6b24938b3dd07b/crates/node/src/util.rs#L52 seems wrong.
You can invoke swc with env `SWC_LOG=error` as a workaround.

PR #4033, opened 2022-03-15
fix(node-swc): Disable logging of `warn` level by default

kdy1:
**Description:**

As swc started using logging in more places, we should adjust logging levels.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/4032


kdy1 (commented):
swc-bump:
 - ''

PR #4035, opened 2022-03-15
refactor(plugin/runner): Split modules and introduce feature flags

kwonoj:
**Description:**

Trying to base https://github.com/swc-project/swc/pull/4027 , questions in comments.


crates/wasm/scripts/build.sh
```diff
@@ -1 +1 @@
-wasm-pack build --debug --scope swc -t nodejs
+wasm-pack build --debug --scope swc -t nodejs --features plugin
```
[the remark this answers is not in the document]
kwonoj: @kdy1 if I understood correctly does it have lot meaning to be able to compile `wasm` target natively? most of code which actually will use `swc` (`swc/plugin`) will be guarded by wasm build target and it'll be mostly just noop.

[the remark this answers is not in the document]
kdy1: >this is part I'm not fully getting now: does build script to wasm need to opt in features manually for https://github.com/swc-project/swc/pull/4035/files#diff-5b743c71dfc992501e49511ae0f681b46af76bac40bec59caf09667b4dd2b12cR19?

Yes, otherwise it breaks cargo operations.

> if that's the case, essentially wasm package need to be written to be able to compile with native target which doesn't do thing and then actual codebase for wasm target?

Maybe, but I think enabling `wasmer/default` `wasmer-wasi/default` from `devDependencies` would be enough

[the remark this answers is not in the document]
kdy1: > if I understood correctly does it have lot meaning to be able to compile wasm target natively? 


`buildable-by-default` is important for development.


> most of code which actually will use swc (swc/plugin) will be guarded by wasm build target and it'll be mostly just noop.

I didn't get it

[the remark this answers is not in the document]
kwonoj: > Maybe, but I think enabling wasmer/default wasmer-wasi/default from devDependencies would be enough

the main question I have is this:

```
//wasm/lib.rs

#[target(not(wasm32))]
fn transform_xx() {
  // noop since we can't refer swc, which depends on swc/plugin relies on plugin_runner's wasm32 target path
}

#[target(not(wasm32))]
fn transform_xx() {
  // actual code
}
```

`cargo build`, and `cargo check` will work but it's nearly noop as it won't be able to check any actual code under right target we need to build. The part I'm curious is if it's worth to try to keep `crates/wasm`  to be built in this way, while it only spends build time. Any reason not to exclude it as https://github.com/swc-project/swc/pull/4012 attempts?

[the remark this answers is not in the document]
kdy1: The important command is `cargo check` and `cargo test`.
But `cargo check` is important because of ide support, and `rust-analyzer` uses `cargo check` with test config, so it follows `devDependencies`.
`cargo test` also uses `devDependencies`, so it would be fine.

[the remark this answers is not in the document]
kwonoj: It looks like my point wasn't clearly delivered 😅 but it won't block immediate work at least. When I have some time I'll try to split this topic into different one for better claritity. Anyway, for now I'll proceed PR with suggested approaches.

package.json
```diff
@@ -56,9 +56,12 @@
         "prepare": "husky install && git config feature.manyFiles true && node ./crates/swc_ecma_preset_env/scripts/copy-data.js",
         "artifacts": "napi artifacts --dist scripts/npm",
         "prepublishOnly": "tsc -d && napi prepublish -p scripts/npm --tagstyle npm",
-        "build": "tsc -d && napi build --platform --cargo-name node --js ./node-swc/src/binding.js --dts ./node-swc/src/binding.d.ts -p node --release",
-        "build:dev": "tsc -d && napi build --platform --cargo-name node --js ./node-swc/src/binding.js --dts ./node-swc/src/binding.d.ts -p node",
+        "pack": "wasm-pack",
         "build:ts": "tsc -d",
+        "build:wasm": "npm-run-all \"pack -- build ./crates/wasm --scope swc {1} -t {2} --features plugin\" --",
```
kwonoj, line 61: Leaves existing `script/build.sh` as-is (for now), while makes it easy to local dev run wasm build.

`npm run build:wasm -- --debug|--release nodejs|web` places argument into placeholder.

tsconfig.json
```diff
@@ -51,6 +51,8 @@
     /* Experimental Options */
     // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
     // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
+    "incremental": true, /* Save .tsbuildinfo files to allow for incremental compilation of projects. */
```
kwonoj, line 54: small improvement for local dev, `tsc` runs incremental build to compile faster.

crates/swc_cli/Cargo.toml
```diff
@@ -12,6 +12,14 @@ version = "0.26.1"
 name = "swc"
 path = "./src/main.rs"
 
+[features]
+default = ["plugin"]
+plugin = [
+  "swc/plugin",
+  "wasmer/default",
```
kdy1, line 19: This will break `wasm/plugin`

kdy1: It should be `wasm/plugin` => `swc/plugin` && `wasmer/js-default`

crates/swc_cli/Cargo.toml
```diff
@@ -20,17 +28,18 @@ rayon = "1"
...
-swc = {version = "0.152.0", path = "../swc"}
+swc = {version = "0.152.0", path = "../swc", features = ["plugin"]}
 swc_common = {version = "0.17.5", path = "../swc_common"}
 swc_trace_macro = {version = "0.1.0", path = "../swc_trace_macro"}
 tracing = "0.1.32"
 tracing-chrome = "0.4.0"
 tracing-futures = "0.2.5"
 tracing-subscriber = {version = "0.3.9", features = ["env-filter"]}
 walkdir = "2"
+wasmer = {version = "2.2.0", optional = true}
```
kdy1, line 39: This should have `default-features = false`

kdy1: Otherwise publish action will fail

crates/swc_cli/Cargo.toml
```diff
@@ -20,17 +28,18 @@ rayon = "1"
 relative-path = "1.6.1"
 serde = {version = "1", features = ["derive"]}
 serde_json = {version = "1", features = ["unbounded_depth"]}
-swc = {version = "0.152.0", path = "../swc"}
+swc = {version = "0.152.0", path = "../swc", features = ["plugin"]}
```
kdy1, line 31: `swc/plugin` should be optionally enabled.

kdy1 (changes requested):

crates/swc_cli/Cargo.toml
kwonoj: Swc cli shouldn't enable wasmer/js, isn't it?

crates/swc_cli/Cargo.toml
kwonoj: Cli should always enable plugin - are you saying it should be built same as wasm does by manually opt into features? Isn't it ok to rely on wasmer/sys as we already did?

crates/swc_cli/Cargo.toml
kdy1: Oh, sorry. this comment is my mistake.

crates/swc_cli/Cargo.toml
kdy1: This is also wrong.
I thought this is `swc`

crates/swc_cli/Cargo.toml
kdy1: No, 

> Cli should always enable plugin

This is simply not possible, because `cranelift` does not support some platforms. We have to build cli for those platforms anyway.

kwonoj: How do we build `/node` currently for those platforms? Does it mean current `@swc/core` doesn't have plugin for those platforms?

kdy1: Yes, those platform lacks plugin support.

kwonoj: ugh, what a headache. Updated default to not to enable plugin by default (https://github.com/swc-project/swc/pull/4035/files#diff-f3d6e8402f7bbe6d2b9d1ce9ca08cc1b4c8406bdae6dc77414e22d65b61a601eR16). 

May need CI side update to enabling this, probabaly with separate PR as current one's not easy to swallow already.

kdy1 (approved):
LGTM. Thank you!


---

swc-bump:

kdy1 (approved):

PR #4038, opened 2022-03-15
fix(es/modules): Use correct span for `ExportAll`s

kdy1:
**Description:**

Now modules passes use correct span for `export * from './src'`.

 - Closes https://github.com/swc-project/swc/issues/4044


kdy1 (commented):
swc-bump:

Issue #4040, opened 2022-03-16
Declaration merging of export defaults not working


tamaroning:
### Describe the bug

The following code containing multiple export defaults seems not to work while Babel compiles it.

### Input code

```typescript
export default interface A　{ a: number };
export default interface B　{ b: number };
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.151&code=H4sIAAAAAAAAA0utKMgvKlFISU1LLM0pUcjMK0ktSktMTlVwfNzQUK2QaKWQV5qblFqkUGvNlYpLrZNCtUISkkoAhpd28FUAAAA%3D&config=H4sIAAAAAAAAA0WNSwrDMAxE76J1ts0id%2BghjKsEF%2F%2FQKBBjfPfawSU7aebpqdIXlrZK2QhYxoQS1Vy0kZbMsOKy0kKKHu3Gg1tfjBysHWG8eudTAs92oeCi28sw2RSyMPBUJh7%2BT7YuCulzjqDez27hSu1xzDuH9wRVTm4%2FOkwQUrUAAAA%3D

### Expected behavior

The input code compiles.

### Actual behavior

Emits an error
```
error: Duplicate export
 --> input.ts:2:1
  |
2 | export default interface B { b: number };
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
note: default was exported at here
 --> input.ts:1:1
  |
1 | export default interface A　{ a: number };
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

### Version

1.2.151

tamaroning:
Thanks for your quick fix :)

PR #4041, opened 2022-03-16
fix(es/lints): Fix false-positive of `duplicate-exports`

kdy1:
**Description:**

 - Closes https://github.com/swc-project/swc/issues/4040


kdy1 (commented):
swc-bump:
 - swc_ecma_lints

PR #4042, opened 2022-03-16
build(tsconfig): set tsbuildinfo filename

kwonoj:
**Description:**

Minor fix to tsconfig's incremental build as it tries to traverse upward.


kdy1 (approved):
Thank you!

---
swc-bump:
 - ''

PR #4047, opened 2022-03-16
fix(es/helpers): should export `default` for classPrivateFieldLooseKey

troy351:
**Description:**

Should export `default` for `classPrivateFieldLooseKey`

I should have noticed this on previous PR 

**BREAKING CHANGE:**
No


kdy1 (approved):
swc-bump:
 - ''

PR #4056, opened 2022-03-16
fix(swc/plugin_runner): make build pass for wasm32

kwonoj:
**Description:**

This PR attempts to fix build issues with wasm target when plugin feature is enabled. The main reason build fails was due to `wasmer-cache` - even though cargo marked it as non-wasm32 deps it is still being picked for wasm32 target build dep graph. wasmer-cache internally have deps top wasmer with `sys` enabled (https://github.com/wasmerio/wasmer/blob/master/lib/cache/Cargo.toml#L14) which makes feature selection conflict we can't control. It may need upstream change probably. For now trying to workaround by opting in fs cache features.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

---

swc-bump:

PR #4058, opened 2022-03-16
feat(swc): embed target triple string

kwonoj:
**Description:**

PR attempts to embed static string to its compilation target triple in case consumer need / want to know what exact binary they loaded. Technically if we follow logics for `bindings.js` does / or use pkg resolver by following transitive deps this could be reconstructable, but this provides faster / deterministic way to get those value instead.

As doc comment says (with `@ignore`) I'd like to propose this as _undocumented_ api and allow to break anytime if internal representation of rust build changes. I expect it's unlikely happen frequently but also I guess usecase of this value would be quite thin.


kwonoj:
Marking as `draft` to get conclusion if this is really needed or not. /cc @Brooooooklyn .

Brooooooklyn:
Looks good to me, I think returning the target string from binary is a good choice for us.

Brooooooklyn (dismissed):

kdy1 (approved):
LGTM. Thanks!

---

swc-bump:
 - ''

Issue #4015, opened 2022-03-14
`declare` modifier not allowed for code already in an ambient context


kaname-png:
### Describe the bug

I have upgraded to the latest version of @swc/core and now I get this error.

I have not updated any libraries or code, it is still the same as it was with the previous version of @swc/core.

### Config

```json
{
  "sourceMaps": true,
  "jsc": {
    "externalHelpers": false,
    "baseUrl": "./src",
    "target": "es2020",
    "parser": {
      "syntax": "typescript",
      "decorators": true,
      "dynamicImport": true
    },
    "paths": {
      "#arguments/*": ["arguments/*"],
      "#commands/*": ["commands/*"],
      "#listeners/*": ["listeners/*"],
      "#preconditions/*": ["preconditions/*"],
      "#decorators": ["lib/decorators/index.ts"],
      "#api/*": ["lib/api/*"],
      "#lib/*": ["lib/*"],
      "#root/*": ["*"]
    },
    "transform": {
      "legacyDecorator": true,
      "decoratorMetadata": true
    }
  },
  "module": {
    "type": "es6"
  }
}
```

### Expected behavior

A successful compilation as well as it happened with the previous version I had of `@swc/core@1.2.147`.

### Actual behavior

Compilation error.

```
PS E:\Proyectos\Personal\Naeko\naeko> yarn build
src/tsconfig.json
src/languages/en-US/api/cdn.json
src/languages/en-US/api/guild.json
 WARN import rewriter: failed to resolve: index not found
src/languages/en-US/api/main.json
src/languages/en-US/api/user.json
 WARN import rewriter: failed to resolve: index not found
 WARN import rewriter: failed to resolve: index not found
 WARN import rewriter: failed to resolve: index not found
 WARN import rewriter: failed to resolve: index not found
src/languages/en-US/commands/anime.json
... more files

  × `declare` modifier not allowed for code already in an ambient context
    ╭────
 32 │ declare module 'discord.js' {
    · ───────
    ╰────

  × `declare` modifier not allowed for code already in an ambient context
    ╭────
 38 │ declare module '@sapphire/pieces' {
    · ───────
    ╰────

  × `declare` modifier not allowed for code already in an ambient context
    ╭────
 75 │ declare module '@sapphire/framework' {
    · ───────
    ╰────

  × `declare` modifier not allowed for code already in an ambient context
     ╭────
 118 │ declare global {
     · ───────
     ╰────


Caused by:
    0: failed to process input file
[...]

PR #4060, opened 2022-03-17
fix(es/parser): Allow nested ambient context in `.d.ts` files

kdy1:
**Description:**

 - Closes https://github.com/swc-project/swc/issues/4015


kdy1 (commented):
swc-bump:
 - swc_ecma_parser

PR #4071, opened 2022-03-17
feat(es/ast): Add `raw` to `Str`

alexander-akait:
**Description:**

feat - add `raw` property for `StringLiteral`

Also:
- fix JSX with only `\n` in props
- improve perf for normal printing strings + print them as they were written by developer
- simplify `TemplateLiteral` - do not store `StringLiteral` (use `JsWord`), we just duplicate AST node, but it is unnecessary

~~Note - there is another improvement and simplification in `TemplateLiteral`, but I will do it in another PR~~ We need do it here to better rewrite code
Also we should add `raw` for numbers too, so we will support linting `1_000_000` ad also will print numbers as they were written by developer

@ArturAralin after this we can simplify `quotes` rule, no need to look at original code (it is slow)

**BREAKING CHANGE:**

Yes

No

Also:
- we have duplicate `IsDirective` in parser and `swc_ecma_utils`


crates/swc_ecma_utils/src/macros.rs
```diff
@@ -31,8 +31,7 @@ macro_rules! quote_str {
         $crate::swc_ecma_ast::Str {
             span: $span,
             value: $s.into(),
-            has_escape: false,
-            kind: Default::default(),
+            raw: $s.into(),
```
kdy1, line 34: Doesn't we need `'` or `"`?

crates/swc/tests/fixture/issue-1227/case-1/output/index.js
```diff
@@ -1 +1 @@
-var foo = "\xa0";
+var foo = "\u{a0}";
```
kdy1, line 1: This is wrong as the target is es5.

crates/swc_ecma_ast/src/lit.rs
```diff
@@ -149,80 +139,31 @@ pub struct Str {
...
 }
 
 impl Take for Str {
     fn dummy() -> Self {
         Str {
             span: DUMMY_SP,
             value: js_word!(""),
-            has_escape: Default::default(),
-            kind: Default::default(),
+            raw: Some("\"\"".into()),
```
kdy1, line 153: I think this should be `None`

crates/swc_ecma_ast/src/expr.rs
```diff
@@ -757,21 +757,42 @@ pub struct TplElement {
...
 }
 
 impl Take for TplElement {
     fn dummy() -> Self {
         TplElement {
             span: DUMMY_SP,
             tail: Default::default(),
-            cooked: Take::dummy(),
-            raw: Take::dummy(),
+            cooked: Some("".into()),
```
kdy1, line 773: Same here.

kdy1 (changes requested):
Thank you so much!

crates/swc_ecma_parser/src/parser/typescript.rs
```diff
@@ -670,10 +667,7 @@ impl<I: Tokens> Parser<I> {
                 TsEnumMemberId::Str(Str {
                     span,
                     value: v.to_string().into(),
-                    has_escape: false,
-                    kind: StrKind::Normal {
-                        contains_quote: false,
-                    },
+                    raw: Some(format!("\"{}\"", v).into()),
```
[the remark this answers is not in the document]
kdy1: Is having this better?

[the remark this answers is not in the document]
alexander-akait: @kdy1 Hm, do you mean

> You can manually use String for better performance, but I don' think it will have a huge didfference.
(I mean String::push & String::push_str)

Just create `String::new()`? I missed that because it is recovery mode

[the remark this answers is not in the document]
alexander-akait: I think in real code this will not happen... but I can refactor to using `String`

crates/swc_ecma_minifier/tests/terser/compress/conditionals/cond_7/output.js
```diff
@@ -1,10 +1,10 @@
 var x, y, z, a, b;
 x = 2;
 x = 2;
-x = "foo";
-x = "foo";
-condition(), (x = 20);
-z || condition(), (x = "fuji");
-x = (condition(), "foobar");
+x = y ? "foo" : "foo";
```
kdy1, line 4: This should not be modified

crates/swc_ecma_minifier/tests/terser/compress/concat-strings/concat_1/output.js
```diff
@@ -3,4 +3,4 @@ var b = "foo1" + x() + "2boo";
 var c = 1 + x() + 2 + "boo";
 var d = 1 + x() + 2 + 3 + "boo";
 var e = 1 + x() + 2 + "X3boo";
-var f = "\x00360\x008\x00";
+var f = "\x00360\x008\0";
```
kdy1, line 6: `crates/swc_ecma_minifier/tests/terser/**` should not be updated

crates/swc_ecma_minifier/tests/compress/fixture/next/feedback-util-promisify/chunks/pages/_app-72ad41192608e93a/output.js
```diff
@@ -1997,7 +1997,7 @@
                         }
                         function stylizeWithColor(r, t) {
                             var e = inspect.styles[t];
-                            return e ? "\x1b[" + inspect.colors[e][0] + "m" + r + "\x1b[" + inspect.colors[e][1] + "m" : r;
+                            return e ? "[" + inspect.colors[e][0] + "m" + r + "[" + inspect.colors[e][1] + "m" : r;
```
kdy1, line 2000: Seems like a wrong change?

crates/swc_ecma_minifier/tests/compress/fixture/projects/wmr/archive-1/chunks/meta-tags.c6b3dbd5/output.js
```diff
@@ -1,7 +1,7 @@
 import { m } from "../index.f66dda46.js";
 import { u as useLang, a as useTitleTemplate, b as useTitle, c as useMeta, d as useLink } from "./hoofd.module.6c5395cb.js";
 function MetaTags() {
-    return useLang("nl"), useTitleTemplate("%s | \u{1F4AD}"), useTitle("Welcome to hoofd"), useMeta({
+    return useLang("nl"), useTitleTemplate("%s | 💭"), useTitle("Welcome to hoofd"), useMeta({
```
kdy1, line 4: I think this can be problematic

crates/swc_ecma_minifier/tests/terser/compress/object/property_with_unprintable/output.js
```diff
@@ -1,23 +1,23 @@
...
+    "\0": "foo",
+    get "\0" () {
         return "bar";
     },
-    set "\x00\x01" (foo){
+    set "\0" (foo){
         save(foo);
     },
-    *"\x00\x01" () {
+    *"\0" () {
```
kdy1, line 9: This seems problematic

kdy1 (changes requested):

crates/swc_ecma_minifier/tests/compress/fixture/projects/wmr/archive-1/chunks/meta-tags.c6b3dbd5/output.js
alexander-akait: hm, it was original code, so we keep them as is, original code:
```
useTitleTemplate("%s | 💭")
```

Because target is not es5, it will not be problem

crates/swc_ecma_minifier/tests/compress/fixture/next/feedback-util-promisify/chunks/pages/_app-72ad41192608e93a/output.js
alexander-akait: In original code it is `ESC` character, so we keep them as is for non minified output

crates/swc_ecma_minifier/tests/terser/compress/object/property_with_unprintable/output.js
alexander-akait: Sam here in original code we have `SOH` character, so we just print it as is

crates/swc_ecma_minifier/tests/compress/fixture/next/feedback-util-promisify/chunks/pages/_app-72ad41192608e93a/output.js
kdy1: Oh, I see it's not rendered correctly in diff views.
I found that it's identical in file view, thanks!

crates/swc_ecma_codegen/src/lib.rs
```diff
@@ -505,55 +505,40 @@ where
...
+        srcmap!(node, true);
 
-                if self.cfg.minify && !is_invalid_use_strict {
-                    let value = get_quoted_utf16(&node.value, self.wr.target(), false, false, true);
+        if &*node.value == "use strict"
+            && node.raw.is_some()
+            && node.raw.as_ref().unwrap().contains('\\')
+        {
+            self.wr
+                .write_str_lit(DUMMY_SP, node.raw.as_ref().unwrap())?;
```
kdy1, line 516: I think we need `srcmap!(node, false)` here

kdy1 (changes requested):

kdy1 (dismissed):
LGTM.
Thank you so much for the hard work.


---
swc-bump:
 - swc_ecma_ast --breaking

kdy1 (approved):

PR #4082, opened 2022-03-18
Update line number start to 0

### LINKED ISSUE #4081: New miette-based error reporter reporting wrong line
### Describe the bug

The new error reporting is reporting "n + 1" instead of "n" line. I.e.: if error is in line 1, it's reporting line 2.

Example output:
```log
  × Unterminated string constant
   ╭────
 2 │ const x = 'aaaa
   ·           ─────
   ╰────
```

### Input code

```typescript
const x = 'aaaa
```

### Config

_No response_

### Playground link

_No response_

### Expected behavior

Correct line should be reported

### Actual behavior

"n + 1" line is getting reported instead of "n"

### Version

1.2.158

### Additional context

_No response_

edwincoronado:
**Description:**
This fixes issue #4081 by starting line number at 0 instead of 1 since Miette increments the line.

Closes #4081


kdy1 (changes requested):
This is wrong place to fix

kdy1 (changes requested):
You should modify miette adapter instead

edwincoronado:
Updating failing tests.

edwincoronado:
@kdy1 do you know if there is a way of generating the `*.stderr` files programmatically? After running this in root directory:

`find . -name '*stderr' | wc -l`

I see that there's a total of 1084 files. Would be very time consuming to fix all manually.

kdy1:
You can run `UPDATE=1 cargo test`
It's in contributing.md

2022년 3월 19일 (토) 오후 4:09, Edwin Coronado ***@***.***>님이 작성:

crates/swc_error_reporters/src/lib.rs
```diff
@@ -109,7 +109,7 @@ impl SourceCode for MietteSourceCode<'_> {
             _cm: self.0,
             data: src,
             span: convert_span(span),
-            line: loc.line,
+            line: loc.line - 1,
```
kdy1, line 112: Can you use `saturating_sub` instead?

kdy1 (changes requested):
If you want, I can update test references for you

crates/swc_error_reporters/src/lib.rs
kdy1: I'll update/merge this PR as this is required for other PRs

kdy1 (approved):
swc-bump:
 - swc_error_reporters

kdy1 (dismissed):

PR #4090, opened 2022-03-19
fix(es/compat): Visit generated codes from `arrow`

kdy1:
**Description:**

**Related issue (if exists):**

 - Closes https://github.com/swc-project/swc/issues/4088

PR #4093, opened 2022-03-19
fix(es/compat): Revert wrong fix of `arrow`

kdy1:
Reverts swc-project/swc#4090


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_compat

PR #4096, opened 2022-03-19
feat(es/minifier): Pass `ProgramData` to the pure optimizer

kdy1:
This is groundwork for more parallelism


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4113, opened 2022-03-21
fix(es/parser): let should be reserved word in strict mode

Austaras:
**Description:**

**BREAKING CHANGE:**


kdy1 (approved):
LGTM. Thank you!

---

swc-bump:
 - swc_ecma_parser

Issue #2948, opened 2021-12-02
error with Unicode escape


RiESAEX:
### Describe the bug

`"\u{Dc00}" // Invalid unicode code point`

### Expected behavior

compile to `"\u{Dc00}"`

### Version

1.2.117

### Additional context

I found this in 
/crates/swc_ecma_parser/test/typescript/tsc/es6/unicodeExtendedEscaptes/unicodeExtendedEsacpesInString11

jimberlage:
I tried fixing this in this commit:

https://github.com/jimberlage/swc/commit/6c6ef84621fd8624f373518ff53d9dab224a063f

It adds a test and uses the u32 value directly.  It looks like fixing this specific test might require something like Simon Sapin's [WTF-8](https://simonsapin.github.io/wtf-8/), which feels like something that's better left for a maintainer to decide.  But feel free to use the test and the code I have!

kdy1:
@jimberlage I thought about this, but I failed to find a clear answer. I'll postpone this a bit. (Because API change is too big)

PR #4115, opened 2022-03-21
fix(ecma/parser): consume surrogates  

alexander-akait:
**Description:**

Fix:

```js
const string98 = "\u{D800}";
```

Will be finished after https://github.com/swc-project/swc/pull/4071/, because we need logic from there

**BREAKING CHANGE:**

No, because we should revisit how we are storing strings, it is just workaround to parse them and keep in `raw`

 - Closes https://github.com/swc-project/swc/issues/2948


crates/swc_ecma_transforms_optimization/src/simplify/expr/mod.rs
```diff
@@ -1438,21 +1438,23 @@ where
...
-    if !s.contains("\\\0") {
+    if !s.contains("\\ud") && !s.contains("\\uD") {
         return Cow::Owned(s.chars().nth(idx).unwrap().to_string());
     }
 
     let mut iter = s.chars().peekable();
 
     while let Some(c) = iter.next() {
-        if c == '\\' && iter.peek().copied() == Some('\0') {
+        println!("FUCK {:?}", c);
```
kdy1, line 1448: 🤣

crates/swc_ecma_parser/tests/test262-error-references/fail/ca2716d236c027cd.js.stderr
```diff
@@ -2,5 +2,5 @@
   x Legacy octal escape is not permitted in strict mode
    ,-[$DIR/tests/test262-parser/fail/ca2716d236c027cd.js:1:1]
  1 | function hello() { 'use strict'; ({ "\1": 42 }); }
-   :                                      ^^
+   :                                        ^
```
kdy1, line 5: Is this span change intentional?

kdy1 (changes requested):

kdy1 (approved):
LGTM. Thank you!

---

swc-bump:
 - swc_ecma_codegen --breaking
 - swc_ecma_parser --breaking

kdy1 (approved):

PR #4116, opened 2022-03-22
fix(es/minifier): Consider side effects while lifting sequences

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4123, opened 2022-03-22
feat(swc/wasm): initial plugin support interface

kwonoj:
**Description:**

This is the initial groundwork for #3934. Mainly, PR attempts to introduce interface to `transformSync` in wasm to accept another param for the plugin's module bytes.

I'm not a huge fan of current approach, honestly. But also wasm-* target is pretty limited to access external system resources - for example, @swc/wasm cannot access filesystem also can't easily use remote-ish interfaces since it implies node.js target.

For now, I'll focus for the primary goal to enable end-to-end workflow to make plugin work in the wasm target. Once it's done, I can come up with proposal for consolidating all interface stablization.

- relates with https://github.com/swc-project/swc/issues/3934

### DIFF
--- a/crates/swc/src/plugin.rs
+++ b/crates/swc/src/plugin.rs
@@ -41,12 +41,12 @@ pub struct PluginContext {
     pub env_name: String,
 }
 
+#[cfg(all(feature = "plugin", not(target_arch = "wasm32")))]
 pub fn plugins(
     resolver: CachingResolver<NodeModulesResolver>,
     config: crate::config::JscExperimental,
     plugin_context: PluginContext,
 ) -> impl Fold {
-    #[cfg(feature = "plugin")]
     {
         RustPlugins {
             resolver,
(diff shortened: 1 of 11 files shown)


crates/swc/src/plugin.rs
```diff
@@ -41,24 +41,29 @@ pub struct PluginContext {
...
         }
     }
+}
 
-    #[cfg(not(feature = "plugin"))]
-    {
-        noop()
-    }
+#[cfg(all(feature = "plugin", target_arch = "wasm32"))]
+pub fn plugins(config: crate::config::JscExperimental, plugin_context: PluginContext) -> impl Fold {
```
alexander-akait, line 60: It seems that plugins API prepared only for JS, perhaps we should reconsider some things so that in the future it will be more extensible and be able to use CSS/HTML/Markdown/etc, just notes

kwonoj: You're correct, current plugin is completely tailored for *transform* only, even among the js execution paths. For example, there's no *parse* plugin interfaces.

I'm trying to move each step forward, current state is making every runtime able to load plugin & execute at least. We started from transform, so completing it resolves major exceution paths (load, execute wasm plugin). From there we need to revisit / expand interfaces to conform with remaining possible paths we want to support transform.

Does it sounds ok to you?

alexander-akait: Sounds good :+1:

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc
 - swc_cli
 - swc_plugin_runner

PR #4124, opened 2022-03-22
fix[]

alexander-akait:
**Description:**

**BREAKING CHANGE:**


crates/swc_ecma_codegen/tests/test262/12752899d5c5eb00.js
```diff
@@ -1,3 +1,3 @@
-`\n\r\\\t\\
+`\n\r\b\v\t\f\
```
alexander-akait, line 1: Fixed bugs

alexander-akait:
For `mapgl` we just replace `\n` (4 bytes) on real newline character (2 byte) (because it is allowed)

crates/swc_ecma_codegen/tests/fixture/template-literal/output.min.js
```diff
@@ -0,0 +1,15 @@
+const template_literal1=`test${"test"}test${"test"}`;const template_literal2=``;const template_literal3=` `;const template_literal4=`string text`;const template_literal5=`string text line 1
```
alexander-akait, line 1: Weird github behavior on source file, we have good tests on all characters and github things it is binary file

Vannevelj: Is it possible there's a NUL character at the end of the file in the console.logs? Github marks a file as binary if it finds one of those within the first 8000 bytes: https://github.com/git/git/blob/9c9b961d7eb15fb583a2a812088713a68a85f1c0/xdiff-interface.c#L187-L193

This is what VS Code shows me: 
image

kdy1: I think this is the only concern.
Other things look good to me

alexander-akait: @Vannevelj @kdy1 Just for tests, in real code it should not be happen

kdy1: Ah, thank you!

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_codegen

Issue #4017, opened 2022-03-15
Implement swcx compile to feature parity to @swc/cli


kwonoj:
### Describe the feature

Related with https://github.com/swc-project/swc/discussions/3859.

We want to establish initial feature parity to existing @swc/cli to plan out migration with possible breaking changes.

These are initial flags we'd like to have. 

- [ ] `config`
- [x] `config_file`
- [ ] `filename`
- [ ] `env_name`
- [ ] `ignore`
- [ ] `source_maps` (https://github.com/swc-project/cli/blob/d60fa69498e970ec045b194bf13a31c9ed6ce413/src/swc/file.ts#L78)
- [ ] `source_maps_target`
- [ ] `source_file_name`
- [ ] `source_root`
- [ ] `watch`
- [ ] `out_file` (https://github.com/swc-project/swc/pull/4016)
- [x] `out_dir`
- [x] `extensions`
- [x] `files`
- [x] `experimental_trace`
- [x] `trace_out_file`

RiESAEX:
just found `filename` is Implemented in #4016

Bruce-Hopkins:
I'm looking to contribute to this project, but it seems all the flags here have already been implemented according to the [docs](https://swc.rs/docs/usage/cli#--config-file).

kwonoj:
It is not, please see above list with checkboxes.

Bruce-Hopkins:
> It is not, please see above list with checkboxes.

Oh, I'm sorry, I got confused with the JS cli tool. I will open a PR hopefully soon to add one of these.

HosseinAgha:
Hi, just wanted to report that looks like the commonjs target outputs es6.

realtimetodie:
- [x]  source_maps 

was resolved in https://github.com/swc-project/swc/pull/6708

realtimetodie:
- [x]  source_maps 
- [x]  source_maps _target

were resolved in https://github.com/swc-project/swc/pull/6708

realtimetodie:
- [x]  source_maps 
- [x]  source_maps_target

were resolved in https://github.com/swc-project/swc/pull/6708

realtimetodie:
I'm working on the implementation of the `--config` option.

alexeagle:
Another TODO item: document the CLI options on the docsite somewhere. Right now tools like Bazel have to recommend that users run `swcx --help` to see how to use it.

Aghassi:
[...]

PR #4125, opened 2022-03-22
feat(swc/cli): implements few compile flags

kwonoj:
**Description:**

Update few more flags for the compile command.

https://github.com/swc-project/swc/issues/4017

### DIFF
--- a/crates/swc_cli/src/commands/compile.rs
+++ b/crates/swc_cli/src/commands/compile.rs
@@ -7,6 +7,7 @@ use std::{
 
 use anyhow::Context;
 use clap::Parser;
+use glob::glob;
 use path_absolutize::Absolutize;
 use rayon::prelude::*;
 use relative_path::RelativePath;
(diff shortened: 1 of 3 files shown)


kwonoj:
Side note, I haven't thought deeply but implementing `config` seems non trivial work since we can't dynamically access rust struct's property with string key indexer as if we did in js object. I spent some time for it, and trying couple of approaches to make it work.

crates/swc_cli/src/commands/compile.rs
```diff
@@ -271,6 +253,36 @@ struct InputContext {
...
 #[swc_trace]
 impl CompileOptions {
+    fn build_transform_options(&self, file_path: &Option<&Path>) -> anyhow::Result<Options> {
+        let base_options = Options::default();
+        let base_config = Config::default();
+
+        let config_file = if let Some(config_file_path) = &self.config_file {
+            let config_file_contents = fs::read(config_file_path)?;
+            serde_json::from_slice(&config_file_contents)
+                .map_err(|e| anyhow::anyhow!("{}", e))
```
kdy1, line 263: Is this required?
I think `.context("")` is enough to convert an error to `anyhow::Error`

kwonoj: yes, looks like old oversight.

kdy1 (approved):
LGTM. Thanks!

---

swc-bump:
 - swc_cli

kdy1 (approved):

PR #4129, opened 2022-03-23
test(es): Organize tests

kdy1:
**Description:**

We move fixture tests from e.g. `issue-3900` to `issues-3xxx/issue-3900` so we can easily find tests.


kdy1 (commented):
swc-bump:
 - ''

PR #4135, opened 2022-03-23
feat(css/minifier): remove `@charset` at-rule when it is possible

alexander-akait:
**Description:**

Remove `@charset` when stylesheet contains only ASCII characters

partial port of https://github.com/cssnano/cssnano/blob/master/packages/postcss-normalize-charset/

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thank you!


---

swc-bump:
 - swc_css_minifier

kdy1 (approved):

PR #4139, opened 2022-03-23
feat(es/lint): use str raw instead of source map access

ArturAralin:
**Description:**

use str raw instead of source map access

**BREAKING CHANGE:**

No


kdy1 (approved):
LGTM. Thanks!


---

swc-bump:
 - swc_ecma_lints

PR #4142, opened 2022-03-24
feat(es/lints): Improve error messages

kdy1:
**Description:**

I wanted to check if CST is required to create a good linter, but it seems like the current api is enough for good error messages.


kdy1 (commented):
swc-bump:
 - swc_ecma_lints

PR #4150, opened 2022-03-24
fix: make errorConfig.filename default to true

### LINKED ISSUE #3950: offending filename is not listed on compilation error
### Describe the bug

Reopening #3931 with repro information: 

The swc cli does not output the offending source filename when there is a compilation error. This can make it very hard to locate the problem on a large codebase.

The tool should emit the standard <filename:linenum:colnum> next to the errors, which makes the lines clickable in many terminals and editors.

I encountered this with issue with #3929.  

### Input code

Given foo.ts with this content:

```typescript
function fn1({x,y}: {x: string, y: string}): string {
    return x + y
}

function fn2({x,y}: {x: string, y: string}): string {
   const fn3: ({x,y}: {x: string, y: string}) => string = fn1
   return fn3({x,y})
}
```

Then run `swc` (with `@swc/core@1.2.151`) as follows: 

```sh
> npx swc foo.ts --out-dir=build

```

### Config

```json
(none)
```

### Playground link

_No response_

### Expected behavior

Errors should list  offending files in the `<filename:linenum:colnum>` format.

### Actual behavior

swc output:
```
error: Cannot reassign to a variable declared with `const`

  |
5 | function fn2({x,y}: {x: string, y: string}): string {
  |               ^
  |
note: x was declared here

  |
6 |    const fn3: ({x,y}: {x: string, y: string}) => string = fn1
  |                 ^

error: Cannot reassign to a variable declared with `const`

  |
5 | function fn2({x,y}: {x: string, y: string}): string {
  |                 ^
  |
note: y was declared here

  |
6 |    const fn3: ({x,y}: {x: string, y: string}) => string = fn1
  |                   ^


Error: Failed to compile 1 file with swc.
    at Object.assertCompilationResult (/<snip>/node_modules/@swc/cli/lib/swc/util.js:115:15)
    at files (/<snip>/node_modules/@swc/cli/lib/swc/file.js:175:18)
    at async _default (/<snip>/node_modules/@swc/cli/lib/swc/file.js:194:9)
```

Note that `foo.ts` is never mentioned anywhere in the output. Here in this case there is only one file so it's not an issue, but I would get the same output if this was run against the full codebase with 1000s of files.

### Version

1.2.151

### Additional context

_No response_

RiESAEX:
**Description:**

make errorConfig.filename default to true.
this will make @swc/cli always output error with filename, except with `-C error.filename=false`.
when using @swc/cli the value of this field in .swcrc is ignored, since swc use this value before read .swcrc

**BREAKING CHANGE:**

Closes #3950

### DIFF
--- a/crates/swc/src/config/mod.rs
+++ b/crates/swc/src/config/mod.rs
@@ -1291,11 +1291,15 @@ fn default_jsonify_min_cost() -> usize {
     1024
 }
 
-#[derive(Debug, Default, Clone, Copy, Serialize, Deserialize)]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 #[serde(deny_unknown_fields, rename_all = "camelCase")]
 pub struct ErrorConfig {
-    #[serde(default = "true_by_default")]
     pub filename: bool,
+}
+impl Default for ErrorConfig {
+    fn default() -> Self {
+        ErrorConfig { filename: true }
+    }
 }
 
 #[derive(Debug, Default, Clone, Serialize, Deserialize)]
--- a/crates/swc_ecma_codegen/src/macros.rs
+++ b/crates/swc_ecma_codegen/src/macros.rs
@@ -100,6 +100,7 @@ macro_rules! semi {
     };
 }
 
+///
 /// - `srcmap!(true)` for start (span.lo)
 /// - `srcmap!(false)` for end (span.hi)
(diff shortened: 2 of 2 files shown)


kdy1 (approved):
Thank you!
---
swc-bump:
 - swc

kdy1 (approved):

PR #4151, opened 2022-03-24
feat(swc/plugin_runner): implement initial load for wasm32

kwonoj:
**Description:**

This was unexpectedly bit tricky due to small bumps across various places. Anyway, this PR covers basic `load` execution paths for the plugin with native runtime (wasm host). It still uses current experimental `resolver` approach which accepts redundant bytes copy between js -> wasm. 

Also, there are couple of issues blocks actual plugin executions yet: 
- wasi binary init fails
- transformexecutor need to be updated to work with wasmer::js
- string-cache upstream fix needed

- https://github.com/swc-project/swc/issues/3934


crates/swc/src/config/mod.rs
```diff
@@ -486,7 +486,8 @@ impl Options {
                 env_name: self.env_name.to_owned(),
             };
 
-            crate::plugin::plugins(experimental, plugin_context)
+            swc_plugin_runner::cache::init_plugin_module_cache_once();
```
kdy1, line 489: I think we should check if `plugins` is not empty

kdy1: Otherwise, we create a directory named `.swc` even though the user does not use plugin.

kdy1 (changes requested):

crates/swc/src/config/mod.rs
kwonoj: this is for the wasm target, where `init_plugin_module_cache_once` does not have access to filesystem and its implementation actually do not create any file - only creates hashmap for in memory cache.

kdy1: Oh, I see. Thanks!

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_plugin_runner
 - swc

PR #4153, opened 2022-03-24
feat(css/stylis): more prefixes

alexander-akait:
**Description:**

More prefixes + resolve TODOs + remove already resolved TODOs

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks! We may need to create some functions because of wasm, but I don't think it's required for now.

---
swc-bump:
 - swc_stylis

crates/swc_stylis/tests/fixture/width-size/output.css
```diff
@@ -87,16 +87,21 @@ p {
...
   grid-template: min-content/ fit-content(10px) max-content;
+  grid-template: -webkit-max-content 1fr -webkit-max-content -webkit-max-content/ -webkit-max-content 1fr;
   grid-template: max-content 1fr max-content max-content/ max-content 1fr;
 }
 .grid-template-columns {
+  grid-template-columns: minmax(100px, -webkit-min-content);
   grid-template-columns: minmax(100px, min-content);
 }
 .grid-auto-columns {
+  grid-auto-columns: -webkit-min-content -webkit-max-content;
```
yisibl, line 104: It seems that Gird has never implemented `min/max-content` with a prefix in the browser?

alexander-akait: Grid properties are not handle right now, it is TODO in this file

yisibl: OK

Issue #4148, opened 2022-03-24
Compile error on type and ts redefinitions


tmayr:
### Describe the bug

Cant import a type from another file with the same name as a definition in the current file unless you use it, getting `is defined multiple times`

When the type has the same name in the same file, it works fine
When the type is being imported, and used, it works fine

I would expect same name values for types to work properly when imported or in the same file.

### Input code

```typescript
//a.ts
import { Foo } from "./b";
// import type { Foo } from "./b"; has the same behavior
const Foo = "hey";


```js
// b.ts
export type Foo = string;
```
```


### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true,
      "decorators": true
    },
    "target": "es2019"
  },
  "module": {
    "type": "commonjs"
  }
}
```

### Expected behavior

Declarations and Types can have the same name when imported.

### Actual behavior

```bash
  × the name `Foo` is defined multiple times
   ╭─[1:1]
 1 │ import { Foo } from "./b";
   ·          ─┬─
   ·           ╰── previous definition of `Foo` here
 2 │ const Foo = "hey";
   ·       ─┬─
   ·        ╰── `Foo` redefined here
   ╰────
```

### Version

1.2.160

XantreGodlike:
```
Error: 
@preact-signals/utils:build:   x the name `Uncached` is defined multiple times
@preact-signals/utils:build:     ,-[D:\Projects\open-source\preact-signals\packages\utils\src\$.ts:17:1]
@preact-signals/utils:build:  17 |  * <div>{$(() => sig.value * 10)}</div>
@preact-signals/utils:build:  18 |  * ```
@preact-signals/utils:build:  19 |  */
@preact-signals/utils:build:  20 | declare class Uncached<T> extends Signal<T> {
@preact-signals/utils:build:     :               ^^^^|^^^
@preact-signals/utils:build:     :                   `-- previous definition of `Uncached` here
@preact-signals/utils:build:  21 |   constructor(accessor: Accessor<T>);
@preact-signals/utils:build:  22 |   get value(): T;
@preact-signals/utils:build:  23 |   peek(): T;
@preact-signals/utils:build:  24 |   valueOf(): T;
@preact-signals/utils:build:  25 |   toString(): string;
@preact-signals/utils:build:  26 |   /** @internal */
@preact-signals/utils:build:  27 |   _a(): T;
@preact-signals/utils:build:  28 | }
[...]

PR #4163, opened 2022-03-25
fix(es/lint): ignore type only import

magic-akari:
**Description:**

Note: 

```typescript

import type { Foo } from "./b"; // fixed in this PR
import { type Foo } from "./b"; // it is OK

import { Foo } from "./b"; // cannot be fixed unless we have multifile mode

```

**BREAKING CHANGE:**

- #4148 partial fixed


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_lints

PR #4168, opened 2022-03-27
feat(es/lint): implement no-obj-calls rule

ArturAralin:
**Description:**

Implement no-obj-calls rule

**BREAKING CHANGE:**
No


kdy1 (approved):
Thank you!


---

swc-bump:
 - swc_ecma_lints

PR #4181, opened 2022-03-29
refactor(swc/cli): alias plugin build with wasm target instead of

kwonoj:
override

**Description:**

PR replaces plugin scaffolding config to not override build to specific target, instead introduce aliased one. This is due to overriding target defeats all other commands (i.e cargo `test` doesn't work out of the box). 

Name of new alias will likely change.

**BREAKING CHANGE:**

New plugin template created via cli now need to specify target for `cargo build` or use new alias.


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_cli

kwonoj:
Yes, probably.

PR #4186, opened 2022-03-29
chore(tests): add back span tests

dsherret:
Accidentally removed during the pretty error PR, which is understandable haha :)

PR #4187, opened 2022-03-29
feat(parser): allow parsing files with missing close braces

dsherret:
**Description:**

This is a small, but huge improvement for swc to get ASTs in spite of syntax errors. Essentially, it allows parsing even if there's a missing close brace in a section of the code.

For example, the following will now parse, but have two emitted diagnostics for missing `}`:

```ts
class MyClass {
  prop: string;

  myMethod() {
    console.log(5);
```
Part of what's outlined in #1178

**BREAKING CHANGE:**

A missing close brace token will no longer be a fatal parsing error, but will instead emit diagnostics.

### DIFF
--- a/crates/swc_ecma_parser/tests/span/js/decl/class-no-close-brace.js
+++ b/crates/swc_ecma_parser/tests/span/js/decl/class-no-close-brace.js
@@ -1,0 +1,3 @@
+class Test {
+
+  method() {}
--- a/crates/swc_ecma_parser/src/parser/class_and_fn.rs
+++ b/crates/swc_ecma_parser/src/parser/class_and_fn.rs
@@ -157,7 +157,16 @@ impl<'a, I: Tokens> Parser<I> {
                     ..p.ctx()
                 })
                 .parse_class_body()?;
-            expect!(p, '}');
+
+            if p.input.cur().is_none() {
+                let eof_text = p.input.dump_cur();
+                p.emit_err(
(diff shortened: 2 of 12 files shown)


crates/swc_ecma_parser/tests/span/js/decl/class-no-close-brace.js
```diff
@@ -0,0 +1,3 @@
+class Test {
+
+  method() {}
```
dsherret, line 3: Waiting on #4186 in order to update the .js.spans below.

kdy1: Thank you!

Actually, I tried error recovery several time and failed, but I think I can do it if I try in some gradual way...
The problem was an infinite loop in some test cases

kdy1 (commented):
Code changes LGTM.
Thank you!

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_parser

crates/swc_ecma_parser/tests/span/js/decl/class-no-close-brace.js
dsherret: This is resolved now.

I think what would be good for error handling is to look at what TypeScript does for certain things and work incrementally on it. TypeScript seems to work well in the editor.

PR #4197, opened 2022-03-30
fix(es/compat): use if/else clause for blocking scope

### LINKED ISSUE #4196: `break` and `continue` statements not executing due to incorrect `for` loop body extraction with `switch` statement when target is `es5`
### Describe the bug

Loop body is extracted into `_loop` function and any `break` or `continue` statements are rewritten to instead `return` `'break'` or `'continue'` string. Returned value from the extracted loop body is used in `switch` clause to `break` or `continue` the loop execution, but it instead `breaks` or `continues` the `switch` clause.

Reproduction code seems to be when the loop has both `break` and `continue` statements and when a function inside the loop references the loop variable (`i`). When any of these conditions are not met code compiles correctly.

Targets higher than `es5` do not cause this issue.

Babel had similar [issue](https://github.com/babel/babel/issues/11800) with the same broken `switch` clause.

### Input code

```typescript
for (let i = 0; i < 2; i++) {
  
  if (i === 0) continue
  if (i === 1) break

  [1].forEach((_) => {
    console.log(i)
  })

}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.157&code=H4sIAAAAAAAAA0vLL1LQyEktUchUsFUwsAZSNgpGQEpbW1OhmpdLQQGEM9MUNIDytkAVmgrJ%2BXklmXmlqagShpoKSUWpidm8XCDxaMNYvbT8ItfE5AwNjXhNBVs7iGEKIN3F%2BTmpejn56RqZmiCxWk2QnloAfi%2Br6okAAAA%3D&config=H4sIAAAAAAAAA0WMTQrEIAxG75K1286id5hDBCctFv9IUhgR714tlu7C915ehUMsrBUyshCPS0pU%2FMMKZAOKZZcVTNf6tKEXagYUeScdiiyd%2BZSEJjUQXHRbGSWbQmYSeRHG3T9m66GQfucYKmjJdAc%2F0N7G%2FHPynaLySe0Cj9ke9LUAAAA%3D

### Expected behavior

`break` and `continue` statements correctly `break` or `continue` the `for` loop instead of affecting the `switch` clause.

### Actual behavior

Loop body is extracted into `_loop` function and then executed with
```javascript
for(var i = 0; i < 2; i++){
    var _ret = _loop(i);
    switch(_ret){
        case "break":
            break;
        case "continue":
            continue;
    }
}
```
which will not `break` or `continue` the `for` loop but instead will `break` or `continue` the `switch` clause.

### Version

1.2.157

### Additional context

I encountered this issue when upgrading `Next.js` to `12.0.9`. Versions up to `12.1.2` also have this issue. Reverting from `swc` to `babel` compilation (by adding `babel.config.js` file to project directory) or downgrading `Next.js` to `12.0.8` solves the issue.

magic-akari:
**Description:**

- and support break/continue with label.

**BREAKING CHANGE:**

- fix #4196

### DIFF
--- a/crates/swc_ecma_transforms_compat/src/es2015/block_scoping/mod.rs
+++ b/crates/swc_ecma_transforms_compat/src/es2015/block_scoping/mod.rs
@@ -282,8 +282,6 @@ impl BlockScoping {
                     })),
                 ];
 
-                let use_switch = flow_helper.has_break && flow_helper.has_continue;
-
                 let check_ret = if flow_helper.has_return {
                     // if (_typeof(_ret) === "object") return _ret.v;
                     Some(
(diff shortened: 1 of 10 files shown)


kdy1:
Seems like a fix for same issue as https://github.com/swc-project/swc/pull/4198?

magic-akari:
> Seems like a fix for same issue as #4198?

Ah 🤣

crates/swc_ecma_transforms_compat/src/es2015/block_scoping/mod.rs
```diff
@@ -717,8 +689,8 @@ impl Visit for InfectionFinder<'_> {
 
 #[derive(Debug)]
 struct FlowHelper<'a> {
-    has_continue: bool,
-    has_break: bool,
+    continue_stmts: AHashSet<Option<Ident>>,
```
kdy1, line 692: I think we should use `AHashSet<Option<JsWord>>` instead.
This will error if there are multiple `continue foo`

kdy1:
Thank you so much, but I decided to merge https://github.com/swc-project/swc/pull/4198

PR #4204, opened 2022-03-30
feat(es/minifier): Introduce concept of epoch in the analyzer

kdy1:
**Description:**

 - Concept of `epoch` in analyzer

Rust does not have GC, so we can't store the value of expression without cloning, if we are not sure it will be used exactly only once.
But this is problematic for code like

```js
"use strict";
const id = (x)=>x
;
const func_bag = {
    func: function() {
        return void 0 === this ? "PASS" : "FAIL";
    }
};
func_bag.func2 = function() {
    return void 0 === this ? "PASS" : "FAIL";
};
var x;
console.log((x = func_bag.func)());
var x;
console.log((x = func_bag.func2)());

```

because the analyzer has the same restriction. (It's rust)
We can workaround this by attaching epoch to all accesses.


kdy1:

Issue #4182, opened 2022-03-29
Support optional comments access from plugin transform


kwonoj:
### Describe the feature

While I would not expect this is general usecase, it is sometimes inevitably plugin's transform would like to access comment from original `Program`. https://github.com/swc-project/swc/issues/4114 would be an example, where comments serve as directives to the instrumentation. There are possibly other cases, as it is quite common to use comments as directives in JS ecosystem. 

Comments are detached from AST, so it doesn't come together when transform fn receives `Program`. Still, host's plugin runner have access to reference to the comments already can pass into plugin. The main challenge here is let plugin access those lazily enough, instead of eagerly serialize & copy over to plugin's memory space. 

Due to current design of how we store comments, we can't directly attach it into some sort of global storage (tls) or attach into HostEnvironment for the wasm (https://github.com/wasmerio/wasmer/discussions/2830). The end goal is plugin access same interface to the host (`impl Comment`) and it serves as proxy to pass / mutate comments to the host without serialize & copy over whole.

PR #4206, opened 2022-03-31
feat(swc/plugin): implements comments:get_leading()

kwonoj:
**Description:**

This is first part implementation for https://github.com/swc-project/swc/issues/4182 , adding necessary foundations to implements `comments::get_leading()`.

There are some numbers of notable changes with bikeshedding which I left comments for some explanations. These are short list of changes:

- introducing thread-local `COMMENTS`, available via `swc_plugin_comments`. This is intended to be imported via `swc_plugin_runner` only, as COMMENTS is tied to lifecycle of reference to singlethreaded comments.
- There is no way to hold non-static lifecycle refernces with existing scoped tls implementation, resolved via `scoped_tls_hkt`. This implementation does not gaurantees lifetime to the comments itself, but we know it'll live until plugin completes transform.
- To return `Vec<Comments>` to the plugin, introduced hostenvironment to attach guest's `alloc` and host directly allocates guest memory. This is not a standard practice still since it needs to return ptr with its length, which can be avoided if guest allocates memory deterministically.
- Bikeshed to the plugin interface as it started to grow params. Mainly it consolidates all of context / metadata into single struct.

**BREAKING CHANGE:**

Plugin transform interface is changed to `transform_xxx(program: Program, metadata: TransformPluginProgramMetadata)`.

- https://github.com/swc-project/swc/issues/4182


crates/swc_plugin_runner/src/imported_fn/mod.rs
```diff
@@ -0,0 +1,136 @@
+//! Functions for syntax_pos::hygiene imported into the guests (plugin) runtime
+//! allows interop between host's state to plugin. When guest calls these fn,
+//! it'll be executed in host's memory space.
+/*
+* Below diagram shows one reference example how guest does trampoline between
+* host's memory space.
+* ┌───────────────────────────────────────┐    ┌─────────────────────────────────────────────┐
```
kwonoj, line 7: small diagram to understand which one lives where, as I myself keep forgetting this 😅

crates/swc_plugin_runner/src/host_environment.rs
```diff
@@ -0,0 +1,29 @@
...
+/// transform.
+/// ref: https://docs.wasmer.io/integrations/examples/host-functions#declaring-the-data
+pub struct HostEnvironment {
+    #[wasmer(export)]
+    pub memory: wasmer::LazyInit<Memory>,
+    pub transform_result: Arc<Mutex<Vec<u8>>>,
+    /// Attached imported fn `__alloc` to the hostenvironment to allow any other
+    /// imported fn can allocate guest's memory space from host runtime.
+    #[wasmer(export(name = "__alloc"))]
+    pub alloc_guest_memory: LazyInit<NativeFunc<u32, i32>>,
```
kwonoj, line 18: wasmerEnv will init this to imported fn `__alloc`, which we declared in plugin's memory space.

crates/swc_plugin_runner/src/imported_fn/comments.rs
```diff
@@ -0,0 +1,70 @@
...
+                        let leading_comments = comments.get_leading(BytePos(byte_pos));
+                        if let Some(leading_comments) = leading_comments {
+                            let serialized_leading_comments_vec_bytes =
+                                Serialized::serialize(&leading_comments)
+                                    .expect("Should be serializable");
+
+                            let serialized_bytes_len =
+                                serialized_leading_comments_vec_bytes.as_ref().len();
+
+                            let (allocated_ptr, allocated_ptr_len) = write_into_memory_view(
```
kwonoj, line 30: Since `get_leading` returns `Option<Vec<Comment>>`, plugin cannot pre-allocate size of the return value similar to others. Instead, it preallocates `CommentsVecPtr(ptr, len)` to let host writes ptr to the vec for the return value. In the host, it allocates return value's memory directly via hostEnvironment::alloc, then writes its ptr with length to guest's commentsVecPtr.

crates/swc_plugin_comments/src/plugin_comments_proxy.rs
```diff
@@ -0,0 +1,130 @@
...
+                serialized_comments_vec_ptr_raw_ptr,
+                serialized_comments_vec_ptr_raw_len
+                    .try_into()
+                    .expect("Should able to convert ptr length"),
+            )
+            .expect("Should able to deserialize CommentsVecPtr")
+        };
+
+        // Using CommentsVecPtr's value, reconstruct actual Vec<Comments>
+        Some(unsafe {
```
kwonoj, line 81: Once `__get_leading_comments_proxy` returns, we know it contains ptr to the `Vec<Comments>`.

crates/swc_plugin_comments/src/host_comments_storage.rs
```diff
@@ -0,0 +1,32 @@
...
+// TODO: This storage does not support mutable yet:
+// https://docs.rs/scoped-tls-hkt/latest/scoped_tls_hkt/#mutable-higher-kinded-types
+#[derive(Copy, Clone)]
+#[cfg(feature = "plugin-rt")]
+pub struct HostCommentsStorage<'a> {
+    pub inner: Option<&'a swc_common::comments::SingleThreadedComments>,
+}
+
+#[cfg(feature = "plugin-rt")]
+scoped_tls_hkt::scoped_thread_local!(
```
kwonoj, line 26: This is the way we hold references to the comments to scoped tls.

crates/swc/src/plugin.rs
```diff
@@ -44,12 +44,14 @@ pub struct PluginContext {
 #[cfg(feature = "plugin")]
 pub fn plugins(
     resolver: Option<CachingResolver<NodeModulesResolver>>,
+    comments: Option<&'_ swc_common::comments::SingleThreadedComments>,
```
kdy1, line 47: I think we can use `Option<swc_common::comments::SingleThreadedComments>` instead, as all `Comments` impls are cheap to clone.

kdy1 (changes requested):

crates/swc/src/plugin.rs
kwonoj: I thought we wanted to keep reference without clone all time. Using cloned would make things lot easier without new scoped tls to hold references :lolsob:

kwonoj: By the way, if we clone comments how can we make mutable work for the comments?

kdy1: We use `Rc<RefCell<T>>`.

https://github.com/swc-project/swc/blob/24ce355d77e083957225721a3f3c6fd77ccc8abb/crates/swc_common/src/comments.rs#L349-L357

kwonoj: oh right, forgot that completely, which was main reason I couldn't attach into hostenvironment.

kdy1: You can create a wrapper type for comments and do `unsafe impl Send for WrappedComments` + `unsafe impl Sync for WrappedComments` because we are sure that it's alive, and it's not multi threaded.

kwonoj: That was the way we created `PseudoScopedKey` for the `GLOBALS` in the plugin context. It is only able to be imported in plugin context (with target_arch support). In this case I was bit hesitant, as `SingleThreadComment` is being used elsewhere than the plugin's context. Even though we think it's likely not happen, any import site can attempt to use SingleThreadedComment with Send + Sync if we create wrapped type for it. Choose explicit approach in this PR instead.

crates/swc/src/config/mod.rs
```diff
@@ -466,7 +466,17 @@ impl Options {
                 swc_plugin_runner::cache::init_plugin_module_cache_once(&experimental.cache_root);
             }
 
-            crate::plugin::plugins(Some(plugin_resolver), experimental, plugin_context)
+            let comments = if let Some(comments) = comments {
```
kdy1, line 469:
```rust
            let comments = if let Some(comments) = comments {
                Some(comments.clone())
            } else {
                None
            };
```

I think this is same as `let comments = comments.clone()`?

crates/swc/src/config/mod.rs
```diff
@@ -487,7 +497,12 @@ impl Options {
             };
 
             swc_plugin_runner::cache::init_plugin_module_cache_once();
-            crate::plugin::plugins(None, experimental, plugin_context)
+            let comments = if let Some(comments) = comments {
```
kdy1, line 500: Same here

crates/swc_plugin_comments/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+authors = ["강동윤 <kdy1997.dev@gmail.com>", "OJ Kwon <kwon.ohjoong@gmail.com>"]
+description = "Internal sharable storage for the comments between host to the plugin"
+name = "swc_plugin_comments"
+version = "0.1.0"
+edition = "2021"
+publish = false
```
kdy1, line 7: Is this intentional?
If we do this, user can't add `swc_plugin` to their deps.

kdy1 (changes requested):

crates/swc/src/config/mod.rs
kwonoj: No, since comments is `Option<&'a SingleThreaded..>`, `comments.clone()` gives same instead of `Option<SingleThreaded..>` But also realized it can be simplifed to `cloned()` instead. Updated.

crates/swc_plugin_comments/Cargo.toml
kwonoj: No, probably mistake. Thanks for the catch, updated.

crates/swc_plugin_comments/src/plugin_comments_proxy.rs
```diff
@@ -0,0 +1,130 @@
+#[cfg(feature = "plugin-mode")]
+use swc_common::plugin::Serialized;
+#[cfg(feature = "plugin-mode")]
+use swc_common::{
+    comments::{Comment, Comments},
+    BytePos,
+};
+
+extern "C" {
```
kdy1, line 9: I think we need `#[cfg(target = "wasm32")]`

kdy1: It's required for testing, if the dependency on `swc_plugin` enables this code path

crates/swc_plugin_comments/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[features]
+# swc/core, which runs plugin
+plugin-rt = []
+# actual wasm plugin binary
+plugin-mode = []
+
+
+[dependencies]
+better_scoped_tls = "0.1.0"
```
kdy1, line 18: This should have the `path`, too.
`Cargo.lock` contains two entries for `better_scoped_tls` because of this

kdy1 (changes requested):

kdy1 (dismissed):
Thank you so much!

---

swc-bump:
 - swc_common
 - swc_plugin --breaking

PR #4207, opened 2022-03-31
fix(es/minifier): Fix evaluation of IIFEs

kdy1:
**Description:**

 - We now remap parameters using a new syntax context. This is required because we may copy functions, and it can result in duplicated variables with respect to the span hygiene.
 - We now visit the sequence expression generated by inlining. This is to improve performance in some cases.
 - We now remove variables from variable statements if it's identical to a parameter declared in the function.
 - IIFE evaluation is now aware of `this` being different for member expressions.
 - We now have more logic for detecting simple functions.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

crates/swc/tests/tsc-references/EnumAndModuleWithSameNameAndCommonRoot_es2015.2.minified.js
```diff
@@ -1,4 +1,3 @@
-var enumdule;
 !function(enumdule) {
```
kdy1, line 1: Oh... This is a bug.

PR #4209, opened 2022-03-31
fix(swc): init custom trace subscriber once

kwonoj:
**Description:**

When fix dropping flush guard global static was removed, and also removed init once logic. PR fixes to init once only, subsequent call do nothing.

**BREAKING CHANGE:**


kdy1 (approved):
LGTM. Thanks!

---

swc-bump:
 - ''

PR #4216, opened 2022-04-01
feat(es/minifier): Improve inliner and evaluator

kdy1:
**Description:**

 - We now inline more function declarations, in case those are simple enough.
 - `pure`: We now evaluate non-top-level IIFEs if it does not have a parameter and an argument.
 - `pure`: We now apply `ignore_return_value` to the argument of `void`.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

kdy1 (commented):
reset ci

PR #4217, opened 2022-04-01
perf(es/minifier): Add fast-path for sequences pass

kdy1:
**Description:**

We don't dig into `b` needlessly.

 - https://github.com/swc-project/swc/pull/4216: This is extracted from #4216 to make reviewing easier


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4218, opened 2022-04-01
feat(ecma/ast): added `raw` property to `BigInt`

alexander-akait:
**Description:**

feat - added `raw` property for BigInt:
- lints now have better AST to understand how it was written
- allow to print BigInt as is 
- better estree compatibility AST 

We will add them same for `Number` and `RegExp` (we have TODO in code for this)

**BREAKING CHANGE:**

Yes

No


crates/swc_ecma_parser/src/lexer/number.rs
```diff
@@ -274,13 +313,17 @@ impl<'a, I: Input> Lexer<'a, I> {
         }
 
         let raw_str = raw.0.take().unwrap();
+        // Remove number separator from number
+        let raw_number_str = raw_str.replace('_', "");
```
alexander-akait, line 317: There is another solution - create the second buffer and store number without number separator and with, but don't think it will be effective here

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_ast --breaking

kdy1 (approved):

PR #4221, opened 2022-04-01
fix(ecma/ast): fix eq ignore span

alexander-akait:
**Description:**

bug fix

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_ast

PR #4222, opened 2022-04-01
Update plugin lib.rs template

vsumner:
**Description:**
This PR updates the plugin template to match the changes from https://github.com/swc-project/swc/pull/4206

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_cli/src/commands/plugin.rs
+++ b/crates/swc_cli/src/commands/plugin.rs
@@ -213,7 +213,7 @@ prepublish = "build --target {}""#,
         create_dir_all(&src_path)?;
         fs::write(
             &src_path.join("lib.rs"),
-            r#"use swc_plugin::{ast::*, plugin_transform};
+            r#"use swc_plugin::{ast::*, plugin_transform, TransformPluginProgramMetadata};
 
 pub struct TransformVisitor;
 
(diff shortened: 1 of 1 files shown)


crates/swc_cli/src/commands/plugin.rs
```diff
@@ -243,7 +243,7 @@ impl VisitMut for TransformVisitor {
 /// important steps manually need to be performed like sending transformed
 /// results back to host. Refer swc_plugin_macro how does it work internally.
 #[plugin_transform]
-pub fn process_transform(program: Program, _plugin_config: String, _context: String) -> Program {
+pub fn process_transform(program: Program, _metadata: TransformPluginProgramMetadata, _context: String) -> Program {
```
kwonoj, line 246: `_context: String` also need to be removed - now all metadata is included in `TransformPluginProgramMetadata`.

kwonoj (commented):
Thanks for catching this!

crates/swc_cli/src/commands/plugin.rs
vsumner: Thanks! I forgot to port that from my project.

kwonoj (approved):

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_cli

PR #4229, opened 2022-04-03
Feat plugin mutable comment

kwonoj:
**Description:**

PR implements remaining comments interfaces along with some refactor to reduce repeated codes for the implementations. Also separate hostenvironment for each impl to not to expose too much details for other features.

**BREAKING CHANGE:**

- closes https://github.com/swc-project/swc/issues/4182

### DIFF
--- a/crates/swc_plugin_comments/src/plugin_comments_proxy.rs
+++ b/crates/swc_plugin_comments/src/plugin_comments_proxy.rs
@@ -6,9 +6,22 @@ use swc_common::{
     BytePos,
 };
 
-#[cfg(target = "wasm32")]
+#[cfg(target_arch = "wasm32")]
 extern "C" {
+    fn __copy_comment_to_host_env(bytes_ptr: i32, bytes_ptr_len: i32);
+    fn __add_leading_comment_proxy(byte_pos: u32);
+    fn __add_leading_comments_proxy(byte_pos: u32);
+    fn __has_leading_comments_proxy(byte_pos: u32) -> i32;
+    fn __move_leading_comments_proxy(from_byte_pos: u32, to_byte_pos: u32);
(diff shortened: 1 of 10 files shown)


crates/swc_plugin_comments/src/plugin_comments_proxy.rs
```diff
@@ -81,54 +111,135 @@ impl Comments for PluginCommentsProxy {
...
+}
+
+#[cfg(feature = "plugin-mode")]
+#[cfg_attr(not(target_arch = "wasm32"), allow(unused))]
+impl Comments for PluginCommentsProxy {
+    fn add_leading(&self, pos: BytePos, cmt: Comment) {
+        self.allocate_comments_buffer_to_host(&cmt);
+        #[cfg(target_arch = "wasm32")]
+        unsafe {
+            __add_leading_comment_proxy(pos.0);
```
kdy1, line 129: How does this work? We don't pass the pointer of the host memory we allocated above?

kwonoj: we do, https://github.com/swc-project/swc/pull/4229/files#diff-5d0fdaffb501760b81888c7b39468d14561cb4f509dd056f2f68ebdfaf3ead93R45

right above does the job. this is same way to set tranaform result, using hostenv's attached buffer guest copies value into host.

kdy1: Yeah, but I mean I'm not sure how `__add_leading_comment_proxy` works without `&pointer_to_comment_in_host_memory`

kdy1: Ah, I found `mutable_comment_buffer`

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_common
 - swc_plugin --breaking

PR #4230, opened 2022-04-03
fix(es/minifier): Remap variables on IIFE evaluation

kdy1:
**Description:**

 - We now check if there's pending inlining before remapping variables.
 - We now use ad-hoc type for remapping.
 - We now inline before remapping to prevent interference with inlining.

 - https://github.com/swc-project/swc/pull/4216

This PR is extracted from #4216 because it may take more time


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4232, opened 2022-04-03
fix(es/parser): support TS1274

magic-akari:
**Description:**
variance  modifier can only appear on a type parameter of a class, interface or type alias

**BREAKING CHANGE:**

- https://github.com/microsoft/TypeScript/pull/48240#issuecomment-1085938523


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_parser

Issue #4146, opened 2022-03-24
Comments in .swcrc files


jasongerbes:
### Describe the bug

The [Configuring SWC docs](https://swc.rs/docs/configuration/swcrc#compilation) contain [several examples](https://swc.rs/docs/configuration/compilation#jsctarget) of `.swcrc` files with comments.

However, SWC cannot be used with a `.swcrc` file that contains any comments.

### Config

```json
{
  "jsc": {
    // Disable es3 / es5 / es2015 transforms
    "target": "es2016"
  }
}
```

### Expected behavior

Ideally, `.swcrc` should use `JSONC` instead of `JSON` to allow the use of comments.

### Actual behavior

SWC fails, with an error like:

```
Caused by:
    0: failed to read swcrc file (svg-sprite-generator/src/index.ts)
    1: failed to deserialize .swcrc (json) file: syntax error: 3:9
    2: key must be a string at line 3 column 9
failed to process input file
```

### Version

0.1.52

jasongerbes:
Thank you @aircloud and @kdy1 :partying_face:

PR #4236, opened 2022-04-03
feat(swc): support JSONC(json comments) for .swcrc

aircloud:
**Description:**

support JSONC(json comments) for .swcrc, as described in this issue: https://github.com/swc-project/swc/issues/4146

>In fact, I really need this feature, especially in the case of team collaboration.
>So, I investigated how to add this little feature quickly and easily.

I went to check the discussion of JSONC in libraries such as `serde_json`. There are many discussions, TL;DR, and there is a solution, it is recommended to use [json_comments](https://crates.io/crates/json_comments), for details, please refer to [this change](https://github.com/dtolnay/request-for-implementation/pull/36) and [the issue]( https://github.com/dtolnay/request-for-implementation/issues/24)

I also think this change is relatively controllable and can continue to use the `serde_json`, so the overall scope of changes is small, and this part will not affect the performance at all.

I submitted a minimal change, and I also think this can continue to be reduced to one parse, there seems to be no problem.(Removing `\u{feff}` doesn't seem to have any side effects, don't know why it is separated)

If the main maintainer thinks it can continue to be simplified, I am also happy to modify it or modify in another PR later(if one PR only does one thing)

**BREAKING CHANGE:**

None

* fix  https://github.com/swc-project/swc/issues/4146


kdy1 (approved):
LGTM. Thank you!

---

swc-bump:
 - swc

PR #4240, opened 2022-04-04
feat(html): initial html/svg lexer

alexander-akait:
**Description:**

Feature - HTML/SVG lexer/tokenizer.

It is initial version - so there are many rooms to fix/improve:

- no AST/tree-builder currently
- Some TODOs
- many places to improve perf
- it is young so bugs exist
- no `raw`/`quote`/etc fields to better codegen (i.e. output as was written)
- some part of code under `unused` because it will be used in future
- more tests

Based on https://html.spec.whatwg.org/multipage/parsing.html

**BREAKING CHANGE:**

No

No


.gitignore
```diff
@@ -27,6 +27,7 @@ node_modules/
 
 # Flamegraph
 *.html
+!crates/swc_html*/**/*.html
```
alexander-akait, line 30: Not sure where do we generate htmls files, so using it, ideally we should ignore html files where we generate them

kdy1: It was used some profiling like `cargo build --timings`

crates/swc_html_parser/src/parser/tests.rs
```diff
@@ -0,0 +1 @@
+
```
kdy1, line 1: I think this files is not required?

crates/swc_html_parser/src/tests.rs
```diff
@@ -0,0 +1 @@
+
```
kdy1, line 1: Same here

kdy1 (changes requested):

kdy1 (approved):
swc-bump:
 - ''

PR #4243, opened 2022-04-04
fix(tests): rename all `inclidesStr` to `includesStr`

dev-itsheng:
**Description:**

When I read the code of swc, I found that many places in the test code use the `inclidesStr` variable to define a string of `'includes'`, I tried to find the reason for this definition, but I didn't find it, so I think it may be a spelling bugs and I fixed them.

**BREAKING CHANGE:**

No


kdy1:
Can you sign the CLA?

dev-itsheng:
image

Yes.

kdy1 (approved):
Thanks!

---

swc-bump:
 - ''

PR #4244, opened 2022-04-04
fix(es/parser) Report more rest errors

### LINKED ISSUE #3676: Trailing comma following rest element should be an error
### Describe the bug

swc would throw error on
```js
const [...rest, ] = a
```
but not in for in/of

### Input code

```typescript
for ([...rest, ] of a) {}

for ({...rest, } of b) {}
```

### Config

_No response_

### Playground link

https://play.swc.rs/?version=1.2.143&code=H4sIAAAAAAAAA0vLL1LQiNbT0ytKLS7RUYhVyE9TSNRUqK7l4koDSVXDpWpBUkkgKQAoNV3qNAAAAA%3D%3D&config=H4sIAAAAAAAAA0WOSw6DMAxEr1J5zaKtEAvW3XIIKxhKlZ9sUxWh3L0hgNhZkzcvs8JHDLQrRGQh3i5ZvOIPWtAlkhieokIFKjlSninlm9HLENhtuKURzfIiExg18A5V0J9BR4o9KkI7oJVSRx5J8wckz%2FujyXIbgtABVOAmPw3L5jbBRSaR6wn9aE8yZZcL%2FbwFa1lbnA2ky3H0JukOcF9H%2Fls6ZUnWg3lzcHSra0jpD2XbA54UAQAA

### Expected behavior

Should throw error `Unexpected trailing comma after rest element`

### Actual behavior

Compiles normally

### Version

1.2.143

### Additional context

_No response_

hanneslund:
**Description:**

I've tried to make reporting of CommaAfterRestElement and NonLastRestParam/TS1014 errors consistent.

I used `self.emit_err` but I noticed that `syntax_error!` is used in other NonLastRestParam errors, which one is preferred here?

closes #3676


Austaras:
`emit_err` is preferred, as it could report multiple error on one pass.

alexander-akait:
We have the same problem for codegen https://github.com/swc-project/swc/blob/main/crates/swc_ecma_codegen/src/lib.rs#L2172, if we will have AST without source code (for example from cache, stored as JSON) we will have problems with code generation, we have many workarounds in codebase for trailing commas, they are all not good solution.

So I think we need introduce `trailing_comma` (and store span value?) like do babel (and other parsers) https://astexplorer.net/#/gist/e217c98d3fe03741d03aa0bbe48e7047/48e674e14ccc25908e30d443566917096c827c88

And so we can solve more problems in this PR, because direction is right

alexander-akait:
We have the same problem for codegen https://github.com/swc-project/swc/blob/main/crates/swc_ecma_codegen/src/lib.rs#L2172, if we will have AST without source code (for example from cache, stored as JSON) we will have problems with code generation, we have many workarounds in codebase for trailing commas, they are all not good solution.

So I think we need introduce `trailing_comma` (and store span value?) like do babel (and other parsers) https://astexplorer.net/#/gist/e217c98d3fe03741d03aa0bbe48e7047/48e674e14ccc25908e30d443566917096c827c88

And so we can solve more problems in this PR, because direction is right

hanneslund:
@alexander-akait 
Are you thinking something like this?
```rust
pub struct ArrayLit {
    pub span: Span,
    pub trailing_comma: Option<Span>, // <---

    pub elems: Vec<Option<ExprOrSpread>>,
}
```

alexander-akait:
@hanneslund Yes, like do babel, it will allow to improve codegen too

crates/swc_ecma_parser/src/parser/mod.rs
```diff
@@ -52,6 +55,8 @@ struct State {
     potential_arrow_start: Option<BytePos>,
 
     found_module_item: bool,
+    /// Start position of an AST node and the span of its trailing comma.
+    trailing_commas: HashMap<BytePos, Span>,
```
kdy1, line 59: Please use `FxHashMap` from `rustc_hash` or `AHashMap` from `swc_common` intead.

kdy1 (changes requested):
Thank you!

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_parser

PR #4245, opened 2022-04-04
feat(ecma/ast): add `raw` to `NumberLiteral`

alexander-akait:
**Description:**

`raw` to `NumberLiteral`

**BREAKING CHANGE:**

Yes

No


crates/swc_ecma_parser/src/parser/typescript.rs
```diff
@@ -2046,16 +2046,30 @@ impl<I: Tokens> Parser<I> {
...
                 let lit = self.parse_lit()?;
                 let lit = match lit {
-                    Lit::Num(num) => TsLit::Number(Number {
-                        span: num.span,
-                        value: -num.value,
-                    }),
+                    Lit::Num(Number { span, value, raw }) => {
+                        let mut new_raw = String::from("-");
+                        let orig_raw = match raw {
+                            Some(raw) => raw,
```
kdy1, line 2061: Please use `push_str` instead.

crates/swc_ecma_parser/src/parser/typescript.rs
```diff
@@ -2046,16 +2046,30 @@ impl<I: Tokens> Parser<I> {
...
+                            _ => value.to_string().into(),
```
kdy1, line 2062: This line should be a call to `write!` macro.

kdy1: Current code allocates, and acquire global lock in `string_cache`, but we don't need these operations.

kdy1 (changes requested):

alexander-akait:
@kdy1 Done :+1:

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_ast --breaking

kdy1 (approved):

Issue #4233, opened 2022-04-03
Minification bug with throw error and preserveAllComments


smvv:
### Describe the bug

Enabling `preserveAllComments` and `compress` gives incorrect output when re-throwing an error:
```
!function() {
    try {} catch (e) {
        throw // hi
        baz(), e;
    }
}();
```

Notice the single line comment and newline between the throw statement and `baz()`.

### Input code

```typescript
function foo() {
  try {
  } catch (e) {
    // hi
    baz()
    throw e
  }
}

foo()
```

### Config

```json
{
  "sourceMaps": true,
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true,
      "dynamicImport": true,
      "decorators": true
    },
    "preserveAllComments": true,
    "transform": null,
    "target": "es2017",
    "loose": false,
    "externalHelpers": false,
    "keepClassNames": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": false,
        "conditionals": false,
        "dead_code": false,
        "directives": false,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": false,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true
      }
    }
  }
}
```

### Playground link

[...]

PR #4254, opened 2022-04-05
fix(es/codegen): Handle comments on the argument of `throw`

kdy1:
**Description:**

 - Closes https://github.com/swc-project/swc/issues/4233


kdy1 (commented):
swc-bump:
 - swc_ecma_codegen

PR #4256, opened 2022-04-05
Export swc_ecma_utils from swc_plugin as swc_ecma_utils::utils

vsumner:
**Description:**
This PR exports `swc_ecma_utils` from `swc_plugin` as `utils`. 

The main reason for this change is to allow the  `swc_ecma_utils::ident::IdentLike` trait to be applied to `swc_plugin::ast::Ident`. 

So that I can use `.to_id()` like `default_specifier.local.to_id()`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3852,6 +3852,7 @@ dependencies = [
  "swc_common",
  "swc_ecma_ast",
  "swc_ecma_quote",
+ "swc_ecma_utils",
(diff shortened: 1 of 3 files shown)


kwonoj (approved):
Sounds ok to me?

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_plugin

PR #4259, opened 2022-04-05
fix(html/parse): support case of cdata

alexander-akait:
**Description:**

Support `CDATA` and `cDaTa` in commnets + found bug - we don't reemit errors in parser (I will fix it)

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_html_parser

Issue #4178, opened 2022-03-28
1.2.161 regression with typed arrow function


dpchamps:
### Describe the bug

Bumping to 1.2.161 has surfaced a regression, which causes the compiler to throw for valid input in a specific context.

### Input code

```typescript
export const x = (a: number) => {
  const y = a && b ? (z: number): string => "hello" : null
}
```

### Config

```json
{
  "sourceMaps": true,
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "transform": {
      "react": {
        "runtime": "automatic"
      }
    },
    "target": "es2022"
  },
  "minify": false
}
```

### Playground link

https://play.swc.rs/?version=1.2.161&code=H4sIAAAAAAAAA0utKMgvKlFIzs8rLlGoULBV0Ei0UsgrzU1KLdJUsLVTqOZSgEpWAiUTFdTUFJIU7BU0quCqrBSKS4oy89JBqpUyUnNy8pUUQJI5OVy1AGbPJwleAAAA&config=H4sIAAAAAAAAAyWLSwoDIRAF7%2FLWswguvUMO0UgbDOOH7hZGBu8eneyKouqG1i6B39QU3qTzga8G%2BBuNRFk26ShGFzxsNNYgqRkOmF7%2FYy4WKhqr5J0LU7AHerGUeY3UrWayFDB3TfLhVYDVvZzDUjmVFAd8pFN5%2FgCKU2KVlgAAAA%3D%3D

### Expected behavior

For the compiler to emit code as expected

### Actual behavior

The compiler throws, looks the context expects an lambda token instead of a type.

```

  [31mx[0m Expected '=>', got ':'
   ,-[[36;1;4minput.tsx[0m:2:3]
 [2m2[0m | const y = a && b ? (z: number): string => "hello" : null
   : [31;1m                              ^[0m
   `----
```

### Version

1.2.161

williamtetlow:
I can take a look at this

dpchamps:
@kdy1 @williamtetlow this regression was only partially fixed in #4199 

[...]

PR #4199, opened 2022-03-30
fix(ecma_parser): Update arrow fn parsing in TS to test if direct child of braceless arrow fn

williamtetlow:
**Description:**
Fixes https://github.com/swc-project/swc/issues/4178 by checking if we are the child of a braceless arrow fn

**BREAKING CHANGE:**

**Related issue (if exists):**
Fixes https://github.com/swc-project/swc/issues/4178

PR #4264, opened 2022-04-06
fix(ecma_parser): Set `is_direct_child_of_braceless_arrow_function` to `false` when inside fn block

williamtetlow:
**Description:**
There was case missed in https://github.com/swc-project/swc/pull/4199

We need to make sure to set `is_direct_child_of_braceless_arrow_function=false` when an arrow function returns an arrow function that has a fn block.

**BREAKING CHANGE:**

fixes https://github.com/swc-project/swc/issues/4178


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_parser

PR #4268, opened 2022-04-07
feat(swc/plugin): implement dummy_with_cmt proxy

kwonoj:
**Description:**

Implements proxy for `Span::dummy_with_cmt` for the plugin.

**BREAKING CHANGE:**


crates/swc_common/src/syntax_pos.rs
```diff
@@ -199,6 +200,10 @@ pub struct MultiSpan {
     span_labels: Vec<(Span, String)>,
 }
 
+extern "C" {
```
kdy1, line 203: I think we need `#[cfg]`

kwonoj: isn't this guarded via https://github.com/swc-project/swc/pull/4268/files#diff-f2c8fa127c4fc658c5b278fa9d6084d2b8c45070d1fc7515191d2afa981bbd8bR446?

this is same as existing, i.e https://github.com/swc-project/swc/blob/main/crates/swc_common/src/syntax_pos/hygiene.rs#L78=

kdy1: Ah, I see, thanks!

kdy1 (approved):
Thanks!

---


swc-bump:
 - swc_common
 - swc_plugin_runner

kdy1 (approved):

PR #4270, opened 2022-04-07
build(cargo): align swc_common bump to include dummy_cmt

kwonoj:
**Description:**

It looks like swc_plugin does not contain latest bits from swc_common to support. Also exposes `Span` from `swc_plugin`.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_common
 - swc_plugin

PR #4271, opened 2022-04-07
feat(es/minifier): Introduce spread-only inliner

kdy1:
**Description:**

 - We don't mark a variable as `used_as_arg` if it's a spread argument.


kdy1:
Seems like I should focus on other rules before this

PR #4273, opened 2022-04-07
fix(es/minifier): Fix handling of `eval`

kdy1:
**Description:**

 - `analyzer`: Fix detection of `eval`.
 - Prevent removal of variables if `eval` is used.
 - Make name mangler work even if `eval` exists.

 - https://github.com/vercel/next.js/issues/35920

---

# Investigation

Seems like `createeElement` issue is already fixed, but there is another error.

image

## More investigation

Seems like a scoping issue.

image

This `e` should have properties with module numbers as the key, but seems like it's overriden

image

## 3

I think this is the combination of

 - ast optimizer changed the scope of some `e` (same symbol, different syntax context)
 - `eval` make name mangler noop
 - terser reuses identifiers, so there are many `e`


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4277, opened 2022-04-07
fix(es/module): TLA in dynamic import

### LINKED ISSUE #4253: Nested `await` in a dynamic import is  not handled
### Describe the bug

SWC is generating invalid code for nested `await` statements on the same line which is resulting in syntax errors when running the generated code 

### Input code

```typescript
let pipeline = await import(await resolve(file))
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript"
    },
    "target": "es2018"
  },
  "module": {
    "type": "commonjs"
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.157&code=H4sIAAAAAAAAA8tJLVEoyCxIzcnMS1WwVUgsT8wsUcjMLcgvKtGAcIpSi%2FNzylI10jJzUjU1AV3i8sMwAAAA&config=H4sIAAAAAAAAA0WMSwrDMAxE76K1F21WJXfoIYSjBAf%2FkBSoMb57bXDISmLmzatwioW1QkYW4vFJiYo%2FWIFsQLHssoLpWI929ELNgCIfpAOR5fX%2B9NqnJDQBA8FFt5chsylkJpGnwnj4m2zdFdJ2jaCCltzvmIQUT4H2iObYyXfSyhe1P0Sd89u9AAAA

### Expected behavior

Nested `await` statements on the same line does not result in a syntax error.

### Actual behavior

Outputs an `await` statement inside a regular function causing a syntax error.

The actual generated code:
```js
let pipeline = await Promise.resolve().then(function() {
    return _interopRequireWildcard(require(await resolve(file)));
});
```

### Version

"@swc/cli": "^0.1.57",  "@swc/core": "^1.2.163",  "@swc/register": "^0.1.10",

### Additional context

_No response_

Austaras:
**Description:**

I initially thought check if it contains await expr would be enough, but later I realize it may cause evalution order problem...

**BREAKING CHANGE:**

closes #4253

### DIFF
--- a/crates/swc/tests/fixture/issues-1xxx/issue-1682/case1/output/index.ts
+++ b/crates/swc/tests/fixture/issues-1xxx/issue-1682/case1/output/index.ts
@@ -6,8 +6,8 @@ exports.getPackage = getPackage;
 var swcHelpers = require("@swc/helpers");
 var _path = require("path");
 async function getPackage() {
-    const pkg = await Promise.resolve().then(function() {
-        return swcHelpers.interopRequireWildcard(require((0, _path).join(process.cwd(), 'package.json')));
+    const pkg = await Promise.resolve(`${(0, _path).join(process.cwd(), 'package.json')}`).then(function(s) {
+        return swcHelpers.interopRequireWildcard(require(s));
     });
     return pkg.default || pkg;
 }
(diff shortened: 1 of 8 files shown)


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_module

PR #4279, opened 2022-04-07
feat(swc/plugin_runner): allow cwd fs access from wasi plugin

kwonoj:
**Description:**

This PR enables `wasm32-wasi` target plugin to have limited filesystem access to cwd, under the `/cwd` alias.

Since this relies on wasi binding to syscall, practically only `@swc/core` would enable actual filesystem support. Wasm, or other runtime environment's sandbox will still limit access.

Primary reason for exposing this is for the cases plugin need some side effect to the filesystem: for some cases, plugin 1. may read from arbitary metadata other than precompiled value (node_modules, or package.json, etcs) 2. may need to emit output other than transformed program. 

Still, this is implicitly enabled to cwd only and does not expose any copnfigurable value yet. As noted in comments, we may need some feedback if we really want to make this as public interface.

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_plugin_runner/src/load_plugin.rs
+++ b/crates/swc_plugin_runner/src/load_plugin.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::{env, sync::Arc};
 
 use anyhow::{Context, Error};
 use parking_lot::Mutex;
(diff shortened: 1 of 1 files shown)


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_plugin_runner

PR #4280, opened 2022-04-07
feat(swc/plugin): supports copy, clone for the comments proxy

kwonoj:
**Description:**

Minor improvement to allow copy, clone for the proxy. Currently it's overly restrictive to pass references only within plugin.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_plugin_comments

PR #4281, opened 2022-04-08
feat(html/parser): implement raw fields

alexander-akait:
**Description:**

Allow to write tokens as they were written + allow to write formatter and linter for it

Also bugs fixes

HTML entities is not supported yet in raw

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thank you!

---


swc-bump:
 - swc_html_ast --breaking

Issue #4282, opened 2022-04-08
Support `Arc<Sourcemap>` in the plugin


kwonoj:
### Describe the feature

Similar to https://github.com/swc-project/swc/issues/4182. 

https://github.com/swc-project/swc/issues/4114 reveals several features need to be supported in the plugin's metadata, sourcemap is one of them to locate Source location from given codes.

### Additional context

Blocking issue for the https://github.com/swc-project/swc/issues/4114.

kwonoj:
I need to dig bit, but I presume this would be similar to how comments work with proxy in the plugin runtime.

PR #4289, opened 2022-04-09
feat(swc/common): allow loc serializable

kwonoj:
**Description:**
- https://github.com/swc-project/swc/issues/4282
As an intial step to support sourcemap, attempt to make `Loc` serializable. There are 2 types (url, pathbuf) doesn't support serialization out of the box requires to write naive string-based conversion.

**BREAKING CHANGE:**


kdy1 (approved):
LGTM. Thank you!

---

swc-bump:
 - swc_common
 - swc_plugin
 - swc_plugin_runner

PR #4290, opened 2022-04-09
refactor(swc/plugin_comments): expand package for generic proxy

kwonoj:
**Description:**

This is bikeshedding for sourcemap support. Originally `plugin_comments` were introduced to contain proxy to comments only and then we are expanding other proxies, like sourcemap. PR renames package more generic and add setup for sourcemap proxy support.

Further we may consolidate all of proxies altogether such as hygiene's `mark` for the globals.

**BREAKING CHANGE:**

If anyone used `plugin_comments` individually, package name has changed. I do not think that's the case though.


kdy1 (approved):
Thanks!

---

swc-bump
 - swc
 - swc_plugin
 - swc_plugin_runner

kdy1 (approved):

Issue #4276, opened 2022-04-07
Undefined variable when using template literals 


AGawrys:
### Describe the bug

There's an undefined instead of the actual string when using template literals.
Seems to happen on most new versions, and not on some older ones. For example, in version 1.2.67, this occurs
`var foo='foo-';console.log(foo);var A="".concat(foo,"bar");var B="".concat(foo,"bar2");console.log("test",foo,A,B);`

Which looks more correct

### Input code

```typescript
const foo = 'foo-';

console.log(foo)
const A = `${foo}bar`;
const B = `${foo}bar2`;

console.log("test", foo, A, B)
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2015",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": false,
        "conditionals": false,
        "dead_code": false,
        "directives": false,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": true,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": true,
        "side_effects": true,
        "switches": false,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true
      },
      "mangle": {
        "toplevel": true,
        "keep_classnames": false,
        "keep_fnames": false,
        "keep_private_props": false,
        "ie8": false,
        "safari10": false
      }
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": true,
  "isModule": true
}
```

### Playground link

[...]

PR #4292, opened 2022-04-09
fix(es/minifier): Fix inlining of uninitialized variables

kdy1:
**Description:**

 - Closes https://github.com/swc-project/swc/issues/4276


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4293, opened 2022-04-09
feat(ecma/ast) Add trailing commas to AST

hanneslund:
**Description:**
@alexander-akait mentioned that adding trailing commas to the AST would make it possible to improve codegen.

So far I've added trailing_comma to ArrayLit and ObjectLit. The other AST nodes that comes to mind:
- ArrayPat
- ObjectPat
- ArrowExpr
- Constructor
- ClassMethod
- PrivateMethod
- Function
- CallExpr
- ImportNamedSpecifier?
- ExportNamedSpecifier?

Does all of these make sense?

**BREAKING CHANGE:**
- trailing_comma added to ObjectLit
- trailing_comma added to ArrayLit

### DIFF
--- a/crates/swc_bundler/src/bundler/chunk/computed_key.rs
+++ b/crates/swc_bundler/src/bundler/chunk/computed_key.rs
@@ -107,6 +107,7 @@ where
             span: DUMMY_SP,
             arg: Some(Box::new(Expr::Object(ObjectLit {
                 span: DUMMY_SP,
+                trailing_comma: None,
                 props: take(&mut export_visitor.return_props),
             }))),
         });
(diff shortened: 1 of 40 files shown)


kdy1:
Hmm... I'm not sure if this is a good idea.

cc @alexander-akait Do we really need this?
My main concern is possible bugs related to this and performance drop.

cc @dsherret 
Does this help simplifying the formatter code?

dsherret:
@kdy1 no, it's not necessary for the formatter. I just collect the tokens and create some helper methods over swc's AST to make it really easy.

alexander-akait:
@dsherret hello, can you share? Currently we have https://github.com/swc-project/swc/blob/main/crates/swc_ecma_codegen/src/lib.rs#L2194 (so our codegen is some slow and potential buggy if we will use AST instead source code), the simple solution is adding it to AST (like do babel,  note - it will be useful for our linter too)

PR #4295, opened 2022-04-09
test: add specs for node loader

JulienKode:
## Description

The goal of this PR is to:
- Implement basic hoisting for monorepo 
- add specs for the node loader

This add the two following spec:
- basic import (currently pass)
- hoisting (currently do not pass)

## Current issue with relative `Path`

When you use `Path` and you are at `.` when you ask the parent it will not give you `../` resulting in a `None` state.

## New approach 

I do not know if this new approach is the best way to handle it as I'm new with rust.

We are now using absolute path so we can always get the `parent` except when we are at the root, and we are using `pathdiff` (which is used by rust itself https://github.com/rust-lang/rust/issues/67108) to get the relative path with:
- the path where we find the node_modules
- the current directory 

There are still a lof of unwrap in this first draft, so I might need to do some additional changes 

**BREAKING CHANGE:**

No breaking change in terms of API 

**Related (if exists):**

- https://github.com/rust-lang/rust/issues/67108
- https://github.com/rust-lang/rust/blob/e1d0de82cc40b666b88d4a6d2c9dcbc81d7ed27f/src/librustc_back/rpath.rs#L116-L158
- https://stackoverflow.com/questions/39340924/given-two-absolute-paths-how-can-i-express-one-of-the-paths-relative-to-the-oth


kdy1 (changes requested):
Please use https://github.com/kdy1/swc/tree/node-resolver
You can make a PR towards it, if you prefer it

JulienKode:
> 

I'll do an update based on this PR, I want to see the spec fail on this PR to make sure, the setup and expectation are right

kdy1:
Tests look good to me.

JulienKode:
close in favour of https://github.com/swc-project/swc/pull/4327

PR #4303, opened 2022-04-11
node-swc: add `isTypeOnly` memebr to missing interfaces

yorkie:
**Description:**

Type definitions(`isTypeOnly`) are not getting updated at #2309, this PR tries to fix that.

**BREAKING CHANGE:**

No breaking change, just a bug fix for typescript usage.

 #2309

**Workaround without this**

```ts
type PatchedNamedExportSpecifier = {
  isTypeOnly: boolean,
} & NamedExportSpecifier;

// some code...
(specifier as unknown as PatchedNamedExportSpecifier).isTypeOnly
```


kdy1 (approved):
Thanks

swc-bump:
 - ''

PR #4318, opened 2022-04-12
fix(css/prefixer): handle at-rules and don't generate unnecessary prefixes

alexander-akait:
**Description:**

Don't generate unnecessary prefixes + handle at-rules.

We have some more places to improve, but I want to refactor code and union logic in `add_declaration!`, so we don't need duplicate properties

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_css_prefixer

PR #4326, opened 2022-04-13
feat(es/parser): Support 'extends' clause to 'infer' type

g-plane:
**Description:**

Support the new TypeScript syntax below:

```ts
type X3<T> = T extends [infer U extends number] ? MustBeNumber<U> : never;
```
Reference: https://github.com/microsoft/TypeScript/pull/48112


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_parser

PR #4327, opened 2022-04-13
feat(ecma/loader): handle hoisting and add specs for node loader 

JulienKode:
## Description

The goal of this PR is to:
- Implement basic hoisting for monorepo 
- add specs for the node loader

This add the two following spec:
- basic import (currently pass)
- hoisting (currently do not pass)

## Current issue with relative `Path`

When you use `Path` and you are at `.` when you ask the parent it will not give you `../` resulting in a `None` state.

## New approach 

I do not know if this new approach is the best way to handle it as I'm new with rust.

We are now using absolute path so we can always get the `parent` except when we are at the root, and we are using `pathdiff` (which is used by rust itself https://github.com/rust-lang/rust/issues/67108) to get the relative path with:
- the path where we find the node_modules
- the current directory 

There are still a lof of unwrap in this first draft, so I might need to do some additional changes 

**BREAKING CHANGE:**

No breaking change in terms of API 

**Related (if exists):**

- https://github.com/rust-lang/rust/issues/67108
- https://github.com/rust-lang/rust/blob/e1d0de82cc40b666b88d4a6d2c9dcbc81d7ed27f/src/librustc_back/rpath.rs#L116-L158
- https://stackoverflow.com/questions/39340924/given-two-absolute-paths-how-can-i-express-one-of-the-paths-relative-to-the-oth


crates/swc_ecma_loader/src/resolvers/node.rs
```diff
@@ -294,14 +309,17 @@ impl NodeModulesResolver {
         base_dir: &Path,
         target: &str,
     ) -> Result<Option<PathBuf>, Error> {
-        let mut path = Some(base_dir);
+        let absolute_path = to_absolute_path(base_dir).unwrap();
+        let mut path = Some(Path::new(absolute_path.to_str().unwrap()));
```
kdy1, line 313: This is supper inefficient and wrong.
Please use `Some(&*absolute_path)` instead

kdy1 (changes requested):
Thanks!

crates/swc_ecma_loader/src/resolvers/node.rs
```diff
@@ -294,14 +309,17 @@ impl NodeModulesResolver {
         base_dir: &Path,
         target: &str,
     ) -> Result<Option<PathBuf>, Error> {
-        let mut path = Some(base_dir);
+        let absolute_path = to_absolute_path(base_dir).unwrap();
```
kdy1, line 312: This also can be an issue.
Please use `to_absolute_path(base_dir)?` instead

crates/swc_ecma_loader/src/resolvers/node.rs
```diff
@@ -391,7 +409,11 @@ impl Resolve for NodeModulesResolver {
                         .and_then(|p| self.wrap(p))
                 } else {
                     self.resolve_node_modules(base_dir, target)
-                        .and_then(|p| self.wrap(p))
+                        .and_then(|path| {
+                            let current_directory = current_dir().unwrap();
```
kdy1, line 413: Same here in this closure

crates/swc_ecma_loader/src/resolvers/node.rs
JulienKode: This point directly to the `inner` in the struct ?

crates/swc_ecma_loader/src/resolvers/node.rs
JulienKode: 4bcc044b013a76a1974e93506d7b1ece577cfd5a should fix the issue

crates/swc_ecma_loader/src/resolvers/node.rs
JulienKode: 4bcc044b013a76a1974e93506d7b1ece577cfd5a should fix the issue

crates/swc_ecma_loader/src/resolvers/node.rs
```diff
@@ -391,7 +409,11 @@ impl Resolve for NodeModulesResolver {
                         .and_then(|p| self.wrap(p))
                 } else {
                     self.resolve_node_modules(base_dir, target)
-                        .and_then(|p| self.wrap(p))
+                        .and_then(|path| {
+                            let current_directory = current_dir()?;
+                            let relative_path = diff_paths(path.unwrap(), current_directory);
```
kdy1, line 414: Can you remove this `unwrap()` too?

JulienKode: yes but with `?` it does not work

JulienKode: Yes b909630b183a5d7f9454dbb632faed79079b7f3f should fix the issue, but I need to add the context case for the error

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_loader

PR #4330, opened 2022-04-14
fix(css/prefixer): fix more bugs

alexander-akait:
**Description:**

Fixed:
- don't add unnecessary prefixes
- avoid invalid prefixes in prefixed at-rule/rule
- fix perf
- fix prefixes in nested at-rules/rule
- and big code refactor

Autoprefixer mostly ported, currently I want to focus:
1. Adding browserslist option to generate only necessary prefixes
2. Remove outdated prefixes based on browserslist (after first)
3. Finish small TODOs

Hope I will finish it on this/next week

**BREAKING CHANGE:**

No

No

### DIFF
--- a/crates/swc_css_prefixer/src/prefixer.rs
+++ b/crates/swc_css_prefixer/src/prefixer.rs
@@ -459,15 +459,17 @@ impl VisitMut for Prefixer {
         for mut n in take(&mut n.rules) {
             n.visit_mut_children_with(self);
 
-            new.append(&mut self.added_rules);
-            new.push(n);
-        }
+            let mut added_rules = take(&mut self.added_rules);
 
-        // TODO only for added_rules
-        new.dedup_by(|a, b| match (a, b) {
-            (Rule::AtRule(a), Rule::AtRule(b)) => a.eq_ignore_span(b),
-            _ => false,
-        });
+            new.append(&mut added_rules);
+            new.push(n);
+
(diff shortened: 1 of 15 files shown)


crates/swc_css_prefixer/src/prefixer.rs
```diff
@@ -512,121 +460,123 @@ impl VisitMut for Prefixer {
         for mut n in take(&mut n.rules) {
             n.visit_mut_children_with(self);
 
-            new.append(&mut self.added_rules);
+            for mut n in take(&mut self.added_top_rules) {
```
kdy1, line 463: I think we don't need clone here

crates/swc_css_prefixer/src/prefixer.rs
```diff
@@ -512,121 +460,123 @@ impl VisitMut for Prefixer {
...
+                new.push(n.1);
+
+                self.rule_prefix = old_rule_prefix;
+            }
+
             new.push(n);
         }
 
+        // Avoid duplicate prefixed at-rules
+        new.dedup_by(|a, b| match (a, b) {
```
kdy1, line 479: This can be an issue if it's not sorted.
`dedup_by` only deduplicates if it's consecutive

kdy1 (changes requested):
Thanks!

kdy1 (approved):
swc-bump:
 - swc_css_prefixer

PR #4332, opened 2022-04-14
fix(es/parser): Make TS InstantiationExpr parsing more permissive

g-plane:
**Description:**

This PR allows:

```ts
const x4 = f<true>
if (true) {}
```

Previously it's an error.

References:

- https://github.com/microsoft/TypeScript/issues/48654
- https://github.com/microsoft/TypeScript/pull/48659


crates/swc_ecma_parser/src/parser/expr.rs
```diff
@@ -1904,4 +1904,21 @@ impl<'a, I: Tokens> Parser<I> {
             }
         }
     }
+
+    fn is_start_of_left_hand_side_expr(&mut self) -> PResult<bool> {
+        Ok(is_one_of!(
+            self, "this", "super", "null", "true", "false", Num, BigInt, Str, '`', '(', '[', '{',
+            "function", "class", "new", '/', "/=", IdentRef
```
kdy1, line 1911: Does `/=` start an expression?

kdy1: Also I'm not sure about `/`

kdy1 (changes requested):

crates/swc_ecma_parser/src/parser/expr.rs
g-plane: > Does `/=` start an expression?

I'm not sure, and I just refer to TypeScript's implementation.

> Also I'm not sure about `/`

It's regular expression, I think. It's also referred from TypeScript.

kdy1: For `/`, we use `Regex` instead.
https://docs.rs/swc_ecma_parser/latest/swc_ecma_parser/token/enum.Token.html#variant.Regex

kdy1: I think our lexer is much more smarter than the one of tsc.
To lex regex, we need the context of parser.
So there are

kdy1: So can you remove `/` and `/=` from the list?

g-plane: Sure, I will remove them later.

kdy1 (commented):
Thank you!

---

swc-bump:
 - swc_ecma_parser

kdy1 (approved):

PR #4335, opened 2022-04-15
feat(es/minifier): Improve sequential inliner

kdy1:
**Description:**

 - We now skip more expressions.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4337, opened 2022-04-15
refactor(bindings): Rename crates

kdy1:
**Description:**

This is a groundwork for `@swc/css` and `@swc/html`.


crates/binding_core_node/Cargo.toml
```diff
@@ -4,7 +4,7 @@ build = "build.rs"
 edition = "2021"
 exclude = ["artifacts.json", "index.node"]
 license = "Apache-2.0"
-name = "node"
+name = "binding_core_node"
```
kwonoj, line 7: nit: do we need `core` in the name? `bindings_node`, or `bindings_wasm` would work?

kwonoj (approved):

kdy1:
@Brooooooklyn Does napi support publishing multiple node packages from single repository?

crates/binding_core_node/Cargo.toml
kdy1: Not required, but I selected it because `@swc/css` can use `bindings_css_node` and `bindings_css_wasm`

kwonoj: it was just naming related questions, no major opinions since in any case actual binding for the npm pkg won't use this name.

alexander-akait (dismissed):
Looks good for me

Brooooooklyn:
> Does napi support publishing multiple node packages from single repository?

Sure, I'll send a pr tomorrow based on yours

kdy1:
Thank you!

kdy1 (commented):
swc-bump:
 - ''

PR #4338, opened 2022-04-15
fix(es/minifier): Inline before cloning

kdy1:
**Description:**

 - We now inlines everything before calling `.clone()`.

 - https://github.com/vercel/next.js/issues/36127#issuecomment-1099794402


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

Issue #3067, opened 2021-12-18
Output mjs file extension


stefee:
### Describe the feature

A configuration option to output mjs files instead of js files (when using module type es6). 

### Babel plugin or link to the feature description

https://babeljs.io/docs/en/babel-cli#set-file-extensions

### Additional context

https://github.com/swc-project/swc/discussions/2953

https://github.com/babel/babel/pull/9144

snowyu:
Why has this merged feature not been released for two years?

kdy1:
It's not merged

Tjerk-Haaye-Henricus:
Will this ever be solved :D I would really need this feature.

wintercounter:
When I transpile to modules, my dynamic imports will stop working in Node, because it either requires an extension to be defined, or the file must have `.mjs` extension. A simple `import('./myModule')` will fail without having an extension defined.

himself65:
Landed in https://github.com/swc-project/cli/pull/286

Should close this issue

himself65:
No, after I tried, even the output file is mjs, but the code is still like `import './xxx.js'`

shimarulin:
In https://github.com/swc-project/swc/issues/8742#issuecomment-2040392007 I describe possible workaround. You also need to use the `.mts` extension of the sources to import them as `.mjs`.

 `tsconfig.json'

⬇⬇⬇

```json
{
  "compilerOptions": {
    "module": "NodeNext",
    "moduleResolution": "NodeNext"
    "...": "..."
  }
}

```

```ts
// ======
// SOURCE
// Import from './pkgs.mts'  to './ctx.mts'
// ======
import { ProjectPackages, getProjectPackages } from "~/pkgs.mjs";

export interface Ctx {
  packages: ProjectPackages | null
}

//...
```

⬇ swc --out-file-extension mjs ... ⬇

```js
import { getProjectPackages } from "./pkgs.mjs";
//...
```

josh-i386g:
@shimarulin thanks that works

```json

  "scripts": {
    "swc-watcher": "npx swc src --out-dir dist --strip-leading-paths --watch --out-file-extension mjs",
    "swc-runner": "node --watch-path=./dist/ ./dist/index.mjs"
  }
```

sirzeta:
I have a problem with ESM (mjs) modules, when i use something like this:

```typescript
// src/bootstrap.mjs
import 'reflect-metadata'
import {RegisterRoutes}      from './routes.mjs'
[...]

PR #4343, opened 2022-04-16
feat(cli): support set output file extension

RiESAEX:
**Description:**

add --out-file-extension and --keep-file-extension for native swc cli
**BREAKING CHANGE:**

#3067 #4017

### DIFF
--- a/crates/swc_cli/src/commands/compile.rs
+++ b/crates/swc_cli/src/commands/compile.rs
@@ -84,10 +84,6 @@ pub struct CompileOptions {
     #[clap(group = "input")]
     files: Vec<PathBuf>,
 
-    /// Preserve the file extensions of the input files.
-    #[clap(long, group = "file_extension")]
-    keep_file_extension: bool,
-
     /// Use a specific extension for the output files
     #[clap(long, group = "file_extension")]
     out_file_extension: Option<String>,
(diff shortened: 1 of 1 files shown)


crates/swc_cli/src/commands/compile.rs
```diff
@@ -84,6 +84,14 @@ pub struct CompileOptions {
     #[clap(group = "input")]
     files: Vec<PathBuf>,
 
+    /// Preserve the file extensions of the input files.
+    #[clap(long, group = "file_extension")]
+    keep_file_extension: bool,
```
kwonoj, line 89: what's the use case for this?

RiESAEX: I'm not sure, I add this just because babel has it. https://github.com/babel/babel/pull/6221
may be only `--out-file-extension` is enought?

kwonoj: Let's not 1:1 port features unless we are sure if it's needed. 

Do we have usecases for `out-file-extension` as well?

RiESAEX: to output .mjs file.

For `---output-file-extension`, maybe someone has both .js and .mjs in their codebase and wants swc to keep the output extension?

kwonoj: I feel we may hit some breaking changes for this later, but for now having out-file-extension only seems enough to get some feedbacks.

RiESAEX: removed

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_cli

PR #4345, opened 2022-04-16
fix(es/minifier): preserve vars in const `IfStmt`

Austaras:
**Description:**

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

---


swc-bump:
 - swc_ecma_minifier

PR #4352, opened 2022-04-17
perf(es/minifier): Split frequently used functions

kdy1:
This improve performance a bit because of cpu cache.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4353, opened 2022-04-17
chore(repo): Make git commit faster

kdy1:
**Description:**

We now uses `fd` instead of `find`


kdy1 (commented):
swc-bump:
 - ''

PR #4356, opened 2022-04-17
feat(es/minifier): Implement more evaluation rules

kdy1:
**Description:**

 - We now handle `array.join` in parallel optimizer.
 - We now handle `[].join` of template literals.
 - We now optimize a template literal as a string addition, in unsafe mode.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

kdy1 (commented):
reset ci

PR #4358, opened 2022-04-17
feat(html): first working prototype

alexander-akait:
**Description:**

First prototype, there are a lot of places for improvements.

Want to finish small basic things for first merge and then improve step by step

**BREAKING CHANGE:**

Yes

No


crates/swc_html_codegen/src/lib.rs
```diff
@@ -43,6 +43,154 @@ where
...
+
+        write_str!(self, n.span, &end_tag);
+    }
+
+    #[emitter]
+    fn emit_text(&mut self, n: &Text) -> Result {
+        let mut text = String::new();
+
+        for c in n.value.chars() {
+            let new_value = match c {
```
kdy1, line 169: I think this line is inefficient, as we don't need allocation here.
We can directly push to `text`

kdy1 (changes requested):

kdy1 (commented):
Can you split some code?
It's nearly impossible to review

kdy1 (approved):
Thank you!

---


swc-bump:
 - swc_html_ast --breaking

PR #4362, opened 2022-04-18
chore(node/types): Add `unsafe` to compress options

xnuk:
(as minification documentation https://swc.rs/docs/configuration/minification says)


kdy1 (approved):
Thanks!

---

swc-bump:
 - ''

PR #4364, opened 2022-04-18
feat(swc/plugin_proxy): support lookup_char_pos for sourcemap

kwonoj:
**Description:**

Implements `lookup_char_pos` for the sourcemap proxy.

This PR includes one notable changes around deserialization - adding new `deserialize` interface for `Fallible`. This is due to `SourceMap` struct contains shared pointer. It requires to satisfiy `SharedSerializeRegistry`, which should be `Fallible` and cannot be `Infallible`. 

All of existing struct used Infallible as default deserializer, but in case of new deserialze fn it'll use `SharedDeserializeMap` as deserializer instead.

In practical usecase, this won't make huge difference. Caller need to opt in corresponding deserialization depends on the character of the struct and it is easy to find out since it won't be able to compile. I feel ergonomics, interface around `Serialized` struct might need some touch later.

**BREAKING CHANGE:**

- https://github.com/swc-project/swc/issues/4282

### DIFF
--- a/crates/swc_plugin_proxy/src/source_map/plugin_source_map_proxy.rs
+++ b/crates/swc_plugin_proxy/src/source_map/plugin_source_map_proxy.rs
@@ -1,1 +1,32 @@
-pub struct PluginSourceMapProxy {}
+#[cfg(feature = "plugin-mode")]
+use swc_common::{BytePos, Loc};
+
+#[cfg(feature = "plugin-mode")]
+#[cfg_attr(not(target_arch = "wasm32"), allow(unused))]
+use crate::memory_interop::read_returned_result_from_host_fallible;
+
+#[cfg(target_arch = "wasm32")]
+extern "C" {
+    fn __lookup_char_pos_source_map_proxy(byte_pos: u32, allocated_ret_ptr: i32) -> i32;
+}
+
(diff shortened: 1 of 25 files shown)


crates/swc_plugin_proxy/src/source_map/plugin_source_map_proxy.rs
```diff
@@ -1 +1,32 @@
...
+}
+
+#[cfg(feature = "plugin-mode")]
+#[derive(Debug, Copy, Clone)]
+pub struct PluginSourceMapProxy;
+
+/// Subset of SourceMap interface supported in plugin.
+/// Unlike `Comments`, this does not fully implement `SourceMap`.
+#[cfg(feature = "plugin-mode")]
+impl PluginSourceMapProxy {
```
kwonoj, line 20: As commented unlike `Comment` this doesn't implement `SourceMap` fully. There are some looks like redundant for the plugin, such as `new` / `load_*`, etcs. 

While we expand this and eventually all of impl seems needed can change this trivially.

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_common
 - swc_plugin_proxy --breaking

PR #4365, opened 2022-04-19
fix(es/minifier): Apply name mangler as much as we can

kdy1:
**Description:**

## Investigation

image

Name mangler creates `var g` and `let g`

 - https://github.com/vercel/next.js/discussions/30237#discussioncomment-2587218


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4366, opened 2022-04-19
fix(es/minifier): Fix default values of `toplevel`

### LINKED ISSUE #2591: IIFE prevents minification?
**Describe the bug**

**Input code**

```js
(() => {
	var LONGNAME = 1;
	console.log(LONGNAME);
})();
```

**Config**

```js
const fs = require("fs");
const { minify } = require("@swc/core");
// const { minify } = require("terser");

let config = {
	sourceMap: { filename: "dist/index.js" },
	toplevel: false,
	module: false,
};

minify(fs.readFileSync("index.js", "utf8"), config).then(({ code, map }) => {
	console.log(code);
});

```

**Expected behavior**
Same minified output as terser:
```js
console.log(1);
```

(Most importantly that the variable names are at least renamed, better inlined, and ideally the now redundant IIFE is removed)

**Current behavior**
Only whitespace is removed
```js
(()=>{var LONGNAME=1;console.log(LONGNAME)})()
```

**Version**
The version of @swc/core: both 1.2.105 and 1.2.106

kdy1:
**Description:**

 - Closes #2591

### DIFF
--- a/crates/swc/src/builder.rs
+++ b/crates/swc/src/builder.rs
@@ -12,7 +12,7 @@ use swc_common::{
     FileName, Mark, SourceMap,
 };
 use swc_ecma_ast::{EsVersion, Module};
-use swc_ecma_minifier::option::MinifyOptions;
+use swc_ecma_minifier::option::{terser::TerserTopLevelOptions, MinifyOptions};
 use swc_ecma_parser::Syntax;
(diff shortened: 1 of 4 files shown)


kdy1 (commented):
swc-bump:
 - swc
 - swc_ecma_minifier

PR #4372, opened 2022-04-19
chore(es/minifier): Setup correct benchmark

kdy1:
**Description:**

For antd, `swc/minify` was 7 passes while `minify/full` was 2 passes


kdy1 (commented):
swc-bump:
 - swc

PR #4384, opened 2022-04-20
fix(swc/plugin_macro): align deps for the plugin

kwonoj:
Description:

There's one another missing pieces for the plugin. Not sure why this happens bit frequently for the plugin's transitive deps.


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_plugin_macro
 - swc_plugin

PR #4394, opened 2022-04-21
fix(es/codegen): Allow non-ascii character in more places

kdy1:
**Description:**

 - https://github.com/vercel/next.js/issues/36151


kdy1 (commented):
swc-bump:
 - swc_ecma_codegen

PR #4402, opened 2022-04-22
feat(html/codegen): print unquoted value when it is possible

alexander-akait:
**Description:**

In some cases we can print attributes without quotes, so let's use it for minified

**BREAKING CHANGE:**

No

No


kdy1 (commented):
Thanks!

---

swc-bump:
 - swc_html_codegen

kdy1:
Can you rebase?

alexander-akait:
@kdy1 Done :+1:

kdy1 (approved):

PR #4404, opened 2022-04-23
fix(es/parser): Reject invalid regex flags

g-plane:
**Description:**

Reject unknown regex flags and duplicated regex flags.

Those cases should be considered "Syntax Error": [spec](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-primary-expression-regular-expression-literals-static-semantics-early-errors).

### DIFF
--- a/crates/swc_ecma_parser/src/parser/expr.rs
+++ b/crates/swc_ecma_parser/src/parser/expr.rs
@@ -339,10 +339,10 @@ impl<'a, I: Tokens> Parser<I> {
                             },
                         );
                         if flags_count.remove(&'\u{0000}').is_some() {
-                            syntax_error!(self, span, SyntaxError::UnknownRegExpFlags);
+                            self.emit_err(span, SyntaxError::UnknownRegExpFlags);
                         }
                         if let Some((flag, _)) = flags_count.iter().find(|(_, count)| **count > 1) {
-                            syntax_error!(self, span, SyntaxError::DuplicatedRegExpFlags(*flag));
+                            self.emit_err(span, SyntaxError::DuplicatedRegExpFlags(*flag));
                         }
                         return Ok(Box::new(Expr::Lit(Lit::Regex(Regex { span, exp, flags }))));
                     }
--- a/crates/swc_ecma_parser/src/parser/expr.rs
+++ b/crates/swc_ecma_parser/src/parser/expr.rs
@@ -1,6 +1,6 @@
 use either::Either;
 use swc_atoms::js_word;
-use swc_common::{ast_node, util::take::Take, Spanned};
+use swc_common::{ast_node, collections::AHashMap, util::take::Take, Spanned};
 
 use super::{pat::PatType, util::ExprExt, *};
 use crate::{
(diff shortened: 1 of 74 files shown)


crates/swc_ecma_parser/src/parser/expr.rs
```diff
@@ -325,11 +325,26 @@ impl<'a, I: Tokens> Parser<I> {
...
+                                let key = match flag {
+                                    'g' | 'i' | 'm' | 's' | 'u' | 'y' => flag,
+                                    _ => '\u{0000}', // special marker for unknown flags
+                                };
+                                map.entry(key).and_modify(|count| *count += 1).or_insert(1);
+                                map
+                            },
+                        );
+                        if flags_count.remove(&'\u{0000}').is_some() {
+                            syntax_error!(self, span, SyntaxError::UnknownRegExpFlags);
```
kdy1, line 342: I think we can use `emit_err` instead?

crates/swc_ecma_parser/src/parser/expr.rs
```diff
@@ -325,11 +325,26 @@ impl<'a, I: Tokens> Parser<I> {
...
+                        }
+                        if let Some((flag, _)) = flags_count.iter().find(|(_, count)| **count > 1) {
+                            syntax_error!(self, span, SyntaxError::DuplicatedRegExpFlags(*flag));
```
kdy1, line 345: Same here

crates/swc_ecma_parser/tests/tsc/parserRegularExpressionDivideAmbiguity3.json
```diff
@@ -52,7 +52,7 @@
                 "ctxt": 0
               },
               "pattern": "regexp",
-              "flags": "a"
+              "flags": "i"
             },
```
kdy1, line 56: This should not be changed

crates/swc_ecma_parser/tests/tsc/parserRegularExpressionDivideAmbiguity3.ts
```diff
@@ -1 +1 @@
-if (1) /regexp/a.foo();
\ No newline at end of file
+if (1) /regexp/i.foo();
```
kdy1, line 1: Same here

kdy1 (changes requested):

crates/swc_ecma_parser/tests/tsc/parserRegularExpressionDivideAmbiguity3.ts
g-plane: Test will fail. How should I do?

kdy1: I think we need something like https://github.com/swc-project/swc/blob/82069d180463193116ce2d1502934fa824ea8509/crates/swc/tests/tsc.rs#L16-L32 ?

(As these tests are copied from tsc periodically, it's better to ignore)

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_parser

kdy1 (approved):

PR #4406, opened 2022-04-23
fix(es/parser): Reject invalid `new.target`

g-plane:
**Description:**

`new.target` is only allowed in function declarations, function expressions or classes.

[Babel REPL](https://babel.dev/repl#?browsers=&build=&builtIns=false&corejs=3.6&spec=false&loose=false&code_lz=HYUw7gdALghgTgcxFIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=false&targets=&version=7.17.9&externalPlugins=&assumptions=%7B%7D)


kdy1:
Seems like there are some test cases which is rejected after this PR?

g-plane:
It should work now, some tests failed to install dependencies.

g-plane:
@kdy1 Can you review this?

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_parser

kdy1 (approved):

PR #4407, opened 2022-04-23
feat(html/minifier): Remove `script` `type` attribute

g-plane:
**Description:**

```html
<script type="text/javascript"></script>
```

to

```html
<script></script>
```

cc @alexander-akait


alexander-akait (commented):
Let's add tests where type is not text/javascript + with type="module".

Also there are other types https://github.com/terser/html-minifier-terser/blob/master/src/htmlminifier.js#L155, make sense to remove them too

g-plane:
Those types shouldn't be removed, I think. For example, whether there's `type="module"` or not can introduce different behaviors.

alexander-akait:
@g-plane Yep, `type="module"` should be keeped, just to avoid regressions

alexander-akait:
But `application/javascript` is unnecessary (and other), there is logic https://github.com/terser/html-minifier-terser/blob/master/src/htmlminifier.js#L555

g-plane:
I think it shouldn't.

From [WHATWG HTML spec](https://html.spec.whatwg.org/multipage/scripting.html#script-processing-model):

the step 8 of **prepare a script**:

> if the script element has neither a type attribute nor a language attribute, then let the script block's type string for this script element be `"text/javascript"`.

Only `text/javascript` can be omitted.

alexander-akait:
@g-plane Other types are legacy, if we look at current spec we even don't found these types, they are depcrecated and no used in modern code, but I think it is safe to remove them right now, but yes, we should keep `type` for other, because for example it can be `text/babel` (for runtime in babel) and not only

alexander-akait (approved):
Thanks, looks good

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_html_minifier

kdy1 (approved):

g-plane:
@kdy1 Can you re-run CI? Some failed to install dependencies.

PR #4411, opened 2022-04-23
test(es/minifier): Use `_` instead of `-` for terser tests

kdy1:
**Description:**

This is required for the interactive postpone script I made


kdy1 (commented):
swc-bump:
 - ''

PR #4413, opened 2022-04-23
fix(es/minifier): remove last break in last `BlockStmt`

### LINKED ISSUE #4412: minify switches in 1.2.171 does not remove a break statement inside a block
### Describe the bug

The `switches` [PR](https://github.com/swc-project/swc/pull/4260) has a bug since it does not remove a `break` statement that's inside a block. See minimal repro code below.

### Input code

```typescript
export function foo(arg) {
	switch (arg) {
		case ENUM_VALUE: {
			const {data} = arg
			call(data)
			break
		}
		default:
			break
	}
}
```

### Config

```json
{
  "sourceMaps": true,
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true,
      "dynamicImport": true,
      "decorators": true
    },
    "preserveAllComments": true,
    "transform": null,
    "target": "es2017",
    "loose": false,
    "externalHelpers": false,
    "keepClassNames": false,
    "minify": { "compress": true }
  }
}
```

### Playground link

_No response_

### Expected behavior

The following output is generated:

```ts
export function foo(arg) {
    if (arg === ENUM_VALUE) {
        let { data  } = arg;
        call(data);
        break; // <-- remove this break
    }
}
```

Notice the break statement that causes a syntax error.

### Actual behavior

_No response_

### Version

1.2.171

### Additional context

_No response_

Austaras:
**Description:**

**BREAKING CHANGE:**

closes #4412


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_minifier

PR #4416, opened 2022-04-24
fix(es/parser): Reject `'use strict'` with non-simple params list in TS

g-plane:
**Description:**

This PR lets it check `'use strict'` with non-simple parameters list in TypeScript. Previously it only checks in JavaScript.


kdy1 (approved):
Thanks!

---
swc-bump:
 - swc_ecma_parser

PR #4419, opened 2022-04-24
fix(es/minifier): Don't drop arguments to an IIFE without body

kdy1:
**Description:**

 - `ignore_return_value`: Fix assign property.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4420, opened 2022-04-24
fix(es/minifier): Preserve more side effects

kdy1:
**Description:**

 - Improved fixer.
 - Make `expr_simplifier` preserve more side effects.


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base

PR #4422, opened 2022-04-24
fix(html/parser): fix foreign element parse

alexander-akait:
**Description:**

Fix parsing svg/mathml + svg can contain mathml and mathml can contains svg :exploding_head: 

**BREAKING CHANGE:**

No

no


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_html_parser

PR #4423, opened 2022-04-24
fix(html/parse): parsing entity

alexander-akait:
**Description:**

Add test to ensure we compress html entities by default + fix typo in condition to avoid errors

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thank you!

---
swc-bump:
 - swc_html_parser

PR #4427, opened 2022-04-25
feat(html/minifier): compress default values of attributes

alexander-akait:
**Description:**

Compress default values.

There are two limitation (so they were not added in default):
1. Some attributes have `default` value in spec, but some browsers don't respect spec and use own values (Safari like it), so they were not added
2. Legacy attributes (like `align`/etc), the same problems for different browsers

In any case, there can be many them, I tried to collect the most popular

**BREAKING CHANGE:**

No

No

### DIFF
--- a/crates/swc_html_minifier/src/lib.rs
+++ b/crates/swc_html_minifier/src/lib.rs
@@ -64,6 +64,86 @@ impl Minifier {
         BOOLEAN_ATTRIBUTES.contains(&name)
     }
 
+    fn is_default_attribute_value(
+        &self,
+        namespace: Namespace,
+        tag_name: &str,
+        attribute_name: &str,
+        attribute_value: &str,
+    ) -> bool {
+        matches!(
+            (
+                namespace,
+                tag_name,
+                attribute_name,
+                attribute_value.to_ascii_lowercase().trim()
+            ),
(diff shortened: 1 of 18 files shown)


crates/swc_html_minifier/src/lib.rs
```diff
@@ -70,6 +150,19 @@ impl VisitMut for Minifier {
         n.visit_mut_children_with(self);
 
         n.children.retain(|child| !matches!(child, Child::Comment(comment) if !self.is_conditional_comment(&comment.data)));
+        n.attributes.retain(|attribute| match &*attribute.name {
+            _ if attribute.value.is_some()
```
kdy1, line 154: I think we don't need if guard here

kdy1: We can directly return the negated boolean for if-guard

kdy1 (changes requested):

alexander-akait:
@kdy1 Rebased and refactored

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_minifier

alexander-akait:
@kdy1 Rebased and tests fixed, ready to merge

kdy1 (approved):

kdy1 (commented):
reset ci

PR #4428, opened 2022-04-25
fix(es/minifier): Fix analysis of update expressions

kdy1:
**Description:**

image

This is now treated as an update of `o`


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4434, opened 2022-04-25
feat(es/optimization): Accept top level mark from simplifiers

kdy1:
**Description:**

With this system we can avoid additional analysis


kdy1 (commented):
swc-bump:

PR #4436, opened 2022-04-25
feat(es/resolver): Use different syntax context for unresolved refs

### LINKED ISSUE #2956: `resolver` should use different mark for top level bindings and unresolved identifiers

kdy1:
**Description:**

 - We don't need a list of bindings to check if an identifier is unresolved.
 - Checking if an identifier is unresolved is now one CPU instruction.
   - Previously it was one hashmap operation.
 - This PR also improves performance, by removing hashmaps mentioned above.

- Closes #2956


kdy1 (commented):
swc-bump:

kdy1 (commented):
swc-bump:
 - swc_common

kdy1:
image


These are actual numbers for performance improvements

PR #4437, opened 2022-04-25
fix(html/parser): parse bug

alexander-akait:
**Description:**

Because HTML parser logic use the state machine logic, we can't bump and get the next token (because parser can change state of tokenizer), otherwise we will parse token in invalid state

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser

kdy1 (approved):

PR #4438, opened 2022-04-25
fix(html/parser): parsing bug with nested tags

alexander-akait:
**Description:**

We should not pop last element, otherwise we will ahve broken AST

**BREAKING CHANGE:**

No

No


kdy1 (changes requested):
CI failed

alexander-akait:
@kdy1 Rebased

kdy1 (approved):

PR #4441, opened 2022-04-26
refactor(html/minifier): Reorganize minification steps

g-plane:
**Description:**

Re-organize minification steps into different modules to increase maintainability.

cc @alexander-akait

### DIFF
--- a/crates/swc_html_minifier/src/lib.rs
+++ b/crates/swc_html_minifier/src/lib.rs
@@ -3,198 +3,25 @@
 use swc_html_ast::*;
 use swc_html_visit::{VisitMut, VisitMutWith};
 
-static BOOLEAN_ATTRIBUTES: &[&str] = &[
-    "allowfullscreen",
-    "async",
-    "autofocus",
-    "autoplay",
-    "checked",
-    "compact",
-    "controls",
-    "declare",
-    "default",
-    "defaultchecked",
-    "defaultmuted",
-    "defaultselected",
-    "defer",
-    "disabled",
-    "enabled",
-    "formnovalidate",
-    "hidden",
-    "indeterminate",
-    "inert",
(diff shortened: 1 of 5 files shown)


alexander-akait (commented):
I did not do this because it is bad for performance and HTML minifier will be small (there are not many things to compress) - less visitors = better perf, but let's ask @kdy1

kdy1:
I qlso prefer less visitor

2022년 4월 26일 (화) 오전 9:32, Alexander Akait ***@***.***>님이 작성:

g-plane:
Ok, but can we split them into different functions while still keeping one visitor?

kdy1:
Yeah, I think it's better for the minfier.
Actually es minifier does same thing

alexander-akait:
Yes, we can, but currenty we have very many functions :smile: Make sense to revisit it late when we implemented collapsing whitespaces/removing optional tags (i.e. we can remove `</p>`/`</html>`/`</body>` because HTML spec allow to not close them)/ and maybe some more. But HTML minifier will not be big because there are not many things to compress, I am thinking we will finish minifier in near future

kdy1:
I think @alexander-akait is right

g-plane:
> currenty we have very many functions

It looks there're four functions, currently, or am I missing anything?

My motivation is to re-organize them as soon as possible, which avoid messing the single `lib.rs` file, and if there're much logic it will be harder to re-organize.

alexander-akait:
@g-plane There is refactor https://github.com/swc-project/swc/pull/4427 to reduce them

g-plane:
Should I wait for that PR or just close this PR?

alexander-akait:
Let's wait, I will implement more compress features and we will revisit it

PR #4442, opened 2022-04-26
fix(html/parsing): formatting element

alexander-akait:
**Description:**

Fix parsing formatting element, currently it is only for non broken/non optional formatting element HTML (as most developers write), I will continue after merge, want to od it step by step to avoid regressions

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser

PR #4457, opened 2022-04-27
fix(html/parser): fosting logic for text nodes

alexander-akait:
**Description:**

Bugfix

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---


swc-bump:
 - swc_html_parser

PR #4461, opened 2022-04-27
fix(html/parser): parser and codegen bugs

alexander-akait:
**Description:**

There are two fixes:
1. Lexer bug - we don't emit temporary buffer to restore original context
2. Avoid escaping in some elements because they content should be outputed as is

**BREAKING CHANGE:**

No

No


alexander-akait:
@kdy1 Rebased and fixed too

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_codegen
 - swc_html_parser

kdy1 (approved):

PR #4468, opened 2022-04-28
fix(es/minifier): Disable some passes for `hoist_props`

kdy1:
**Description:**

 - This PR fixes recent version of react libraries.
 - Disabled passes will be re-enabled after fixing the alias analyzer.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4470, opened 2022-04-28
feat(es/minifier): Consider cost of functions for inlining

kdy1:
**Description:**

 - We now treat call expression as 'cheap enough to inline'.

 - https://github.com/swc-project/swc/issues/2934#issuecomment-1111192298


kdy1:
cc @Austaras 
I implemented simple ones based on your analysis, thank you!

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

kdy1:
Oh, we need some sort of cost calculation.
I'll split this PR

PR #4471, opened 2022-04-28
feat(es/minifier): Drop more returns

kdy1:
**Description:**

 - We now drop more useless returns.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4475, opened 2022-04-29
feat(es/minifier): Reuse variable names

kdy1:
**Description:**

 - We now reuse variable names if it's only used by an initializer of a variable declaration.


kdy1:
@Austaras Can you describe more about reusing variable names in https://github.com/swc-project/swc/issues/2934#issuecomment-1111192298 ?
I tried implementing it but seems like the size difference is too small.

One thing I found from diffing is reuse of variables in non-first statements, like 


```js
// lots of code

var a = b + b;

for (var d = init;) {
    use(d)
}

```

became

```js
// lots of code

var a = b + b;

for (b = init;) {
    use(b)
}

```


but, I'm curious if there was more renaming. (I didn't analyze fully yet)

Austaras:
Seems like it's not really about initializer, but about reference
```js
function hello(name) {
  console.log('Hello, ' + name);
  var b = window.a
  console.log(b === b)
}
```
be turned into
```
function hello(a) {
  console.log("Hello, " + a);
  a = window.a;
  console.log(a === a);
}
```
so `a` is not referred after `var b` declaration, so it can be reused

kdy1:
Thank you!

kdy1:
Closing in favor of https://github.com/swc-project/swc/pull/5781

PR #4478, opened 2022-04-29
feat(es/minifier): Drop pure `new` calls

kdy1:
**Description:**


Austaras:
This should be under `unsafe` flag -- as we cannot know if global constructor is modified

kdy1:
I'm not sure about it.

I think opt-out is fine in cases like this

Austaras:
Well `terser` put them under `unsafe` and disable it by default -- but IMO that's a bad design. `unsafe` sounds too scary, but today few people actually modify global object.

kdy1:
It sounds scary, but it's fine for more than 99.99% of users

kdy1:
Or `untouched_globals`
I don't have a good solid idea about the name

Austaras:
Yeah enable it by default is good, but unsafe by default sounds a bit awkward... what about `origin_globals` or `pristine_globals`

kdy1:
Oh, I love `pristine_globals`

PR #4491, opened 2022-04-30
feat(es/parser): Reject private field called `constructor`

g-plane:
**Description:**

Reject the code below:

```js
class C {
    #constructor() {}
}
```


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_parser

Issue #4387, opened 2022-04-20
Setting `jsc.minify: {}` removes comments regardless of `minify: false` and `jsc.parser.preserveAllComments: true`


xnuk:
### Describe the bug

There's a case that comment should be preserved like [GreaseMonkey user scripts](https://wiki.greasespot.net/Metadata_Block#Syntax). And, um, yes its metadata block wrapped in between `// ==UserScript==` and `// ==/UserScript==`, not block comments. swc did remove this anyway. After digging some hours, I found `jsc.minify: { ... }` (not `minify: false`) removes comments inside of my code.

### Input code

```typescript
// This code will make my senpai notice me
document.write([
  'uwu',
  'senpai',
  'notice',
  'me',
].join(' '))
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "preserveAllComments": true
    },
    "minify": {},
    "target": "es2022"
  },
  "module": {
    "type": "es6"
  },
  "minify": false
}
```

### Playground link

https://play.swc.rs/?version=1.2.168&code=H4sIAAAAAAAAAyXLwQqAIBAE0HtfMTcTIn%2BmW3QQW2irdSMV6e8rPM1jmHEO08YJQVdC5fOE%2BIMgDxLFyzOiZg5fQd2qoQjFPNabM%2FVzB5hSixl%2BtHVzuzTLn8u4K8fewFj7AijUiKdxAAAA&config=H4sIAAAAAAAAA02MMQ7DIBAE%2F7I1RUThgi7KSxA5R1gHRndHFGT57zGp0q00O3Ng04RwoEVRkrl0VIsfBFAqUZPkZnBoQhd%2F0535sZdC1RRhjax0OpRc8zoQamd2sCgvshlQf%2FMe87A%2FO9Os22j0Qwv%2BRJNO5xfwYBTvjAAAAA%3D%3D

### Expected behavior

```javascript
// This code will make my senpai notice me
document.write([
    'uwu',
    'senpai',
    'notice',
    'me', 
].join(' '));
```

### Actual behavior

```javascript
document.write([
    'uwu',
    'senpai',
    'notice',
    'me', 
].join(' '));
```

### Version

`@swc/core`: 1.2.170

### Additional context

- Removing `jsc.minify` property or setting as `null` fixes the problem.
- Same thing occurs on `jsc.parser.syntax === 'typescript'`.
[...]

PR #4492, opened 2022-04-30
fix(node/types): Types for `jsc.minify.format`

xnuk:
Uses helper type `ToSnakeCase` to support snake_case property usage.

### Related issue (if exists):
- #4387
- https://github.com/swc-project/website/commit/7b643838779db47c9329c0efba3b2435eef92584

### Maybe out of scope:
- Is there some tools for auto-generating these stuffs?
- I want to use Prettier for this file *badly*.


kdy1:
Yeah, I think we need to configure some formatter for ts/js files too

kdy1 (approved):
Thanks!

---

swc-bump:
 - ''

PR #4496, opened 2022-05-02
fix(es/decorators): Make legacy decorator identical to `tsc`

### LINKED ISSUE #2081: swc should not allow using undefined as decorator
I made a mistake while fixing #1362
But it's not urgent, because it is about reporting no error for wrong code, not breaking correct code.

### LINKED ISSUE #2655: Getters and setters applied through a decorator don't get called when decorated property is also initialized.
### Describe the bug

If you create a decorator that applies a getter and setter to the decorated property, and then also initialize (e.g. assign a value to the decorated property in the class definition), the resulting class, transpiled by `swc`, will never call the getters and setters for that property. The same code transpiled by `tsc` will work as expected.

### Input code

```typescript
const testDecorator = <T extends {}>(target: T, key: keyof T) => {
    const privateField = Symbol();
    // We define getters and setters for the property on the prototype of the class
    // A real application might use this to intercept changes to the decorated property.
    // This is a simplified version of a pattern used by the @microsoft/fast-elements library.
    Reflect.defineProperty(target, key, {
        get: function () {
            console.log(`called getter for property ${key}.`)
            return (target as any)[privateField]
        },
        set: function (newValue) {
            console.log(`called setter for property ${key} with newValue ${newValue}.`)
            return (target as any)[privateField] = newValue
        }
    })
    console.log("called testDecorator!");
  };

class TestClass {
    @testDecorator
    testProp = "hello"
}

const instance = new TestClass();
// SWC console output:
//    "called testDecorator!"
// TSC console output:
//    "called testDecorator!"
//    "called setter for property testProp with newValue hello"

instance.testProp = "goodbye";
// SWC console output:
//    <nothing>
// TSC console output:
//    "called setter for property testProp with newValue goodbye."

console.log("testProp is now", instance.testProp);
// SWC console output:
//    "testProps is now goodbye"
// TSC console output:
//    "called getter for property testProp."
//    "testProps is now goodbye"
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "decorators": true
    },
    "target": "es2016"
  }
}
```

### Playground link

[see here for current swc output of the above example](https://play.swc.rs/?version=1.2.106&code=H4sIAAAAAAAAA51SwU7DMAy99ytMxWGVpu6%2BMQQCcUZQwQEhLaRuG5ElVeIyqqn%2Fjpu2Y0NCQuRQOXb83vOrpTWegNDTLUrrBFkHa7jIAD8JTe5h313OSLgSaQnZHN6xXfYfW0CWwPoS9hHwkQGmdupDEN4p1DmjPLbbN6tnySo8WSzgGSHHQhkEhiN0HoTJwY9xwdRUIaPYGh21YM10J0ttjcCkfUJq4f2EeQ0OhQZR11pJQYqbtqqsCBqP%2FFp5IAvKMIPEmkBWwpQYkj1UPgyN%2BYE1TQPyAxYaJaWD3vuxODoRbJiPo%2FcnuFM0Rgb%2BWXJUmtyxGlNty9lGCq2Zb3AgDH0Y%2BHzPuF26SU66HVLjGHXgBtG71iYvx2a%2FHhq6%2BSH0p6IM7p6EbvAv4vyv4mCnqIIJi5NT%2BC%2FZvCRT%2F%2FcEIeqS6Ke4eBR3sqxncVivbhVFYS0g4%2BpNiIYxr06eh1Sf6f8os8cVam3jqPsCdXSR1AgDAAA%3D&config=H4sIAAAAAAAAA0WOQQ7DIAwE%2F%2BJzDm0POfCHPgIRJyICjGxHaoT4ewGlzc3a3Rm5wC4OTIFsWZD7JWdS%2BwEDemYUxz4rTKDSotUGwQkWdMRWiQWM8oG11ZY31AahvB7PuQGBSPCPRJ%2F8ena9o5gZRe7Kpi38lrW5Ii1HD8r4YDhnqLfj4ry8r%2BH44QtyiB%2BMygAAAA%3D%3D).

### Expected behavior

I'd expect the getters and setters defined by the decorator to be called, just like they are in the code transpiled by `tsc`.

### Version

1.2.106

### Additional context

You can see [`tsc` output of the above example here](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAYwgOwM7xsdARYSoCGM0cAvHADwAqcomyAJqnAN4C+AfABQwFQDmwGAC44VADRwA1sACeombIgAzMQEoyHVgCg4exCnRwwUAJYA3IsABip4ABsGZOAGVZAWwBGEe1zUBuXX0AemC4AHVgOAZgZVNkKMEYTChmAkY4VCEU5mUSGAALKJMIMGBYWTgUOELiqAhiGFkyqtVaxHsCVFQgvVC4AEE4KGACezgCMDB7UwQiU2r3U34C+ABXLJqC02ZiOHiUhGAweAQC9MFdiC2omPwrJxKyioA6F964ACVY+zwYF5icQSAAV6s8mjw+ElJIpJCwPvo4ElRMo1sgEDAFsg4H4dIj8fokGgfMAXvYIPwuAADOb2X5OJIpOB5KDGMHlJpwAAkLEUbBeVLUCIJwyEayg2MhAiEEzSyFkagA2iYLFZbA4GABdYV6NjiHWZIQotEYrE4hIAdwAamM1sANPCRfiiagSWSKdTafTDclysySE8OZUeXy4BbTIU4JabfY7dyWNHbcB+YKDYiRjBxZLeNL4F0JvKlSrLJh1Y5Nc5E7HgAa2B82ELES63eTKQAiL3AJyYHB4aBEaAAQjbASCbEC2gQnW6YiwMAAwtPmI79AABHswXD3YhQD4b0GlZxtop0iBt7RsIA).

I also have [a reproduction in this repo](https://github.com/astegmaier/swc-decorators-test).

The root cause appears to be the fact that typescript will transpile the property initializer into a simple assignment in the constructor - which invokes the getters and setters defined on the prototype - like this:

```
class TestClass {
    constructor() {
        this.testProp = "hello";
    }
}
```
...while SWC will ultimately do the initialization through a call to `Object.defineProperty([instance], "testProp", [descriptor])`, which bypasses the prototype (see swc playground link above for details).

This bug is the root cause of an issue that blocks the use of `swc` with the `@microsoft/fast-elements` library. The `@observer` ([docs](https://www.fast.design/docs/fast-element/defining-elements)) and `@attr` ([docs](https://www.fast.design/docs/fast-element/observables-and-state)) decorators in this library follow a pattern similar to this example. You can see a reproduction of those more-complicated real-life examples in [the repo I linked to above](https://github.com/astegmaier/swc-decorators-test).

### LINKED ISSUE #3035: _class is not defined when using with default Nest.js Logger
### Describe the bug

Somehow when you use `nestjs-pino` or just a regular `Logger` implementation of Nest js and pass down there a class like `UserService.name` it fails with weird `_class is not defined` error.

### Input code

```typescript
-
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "decorators": true,
      "dynamicImport": false
    },
    "transform": {
      "legacyDecorator": true,
      "decoratorMetadata": true
    },
    "target": "es2018"
  },
  "module": {
    "type": "commonjs"
  },
  "sourceMaps": true
}
```

### Playground link

https://gist.github.com/RIP21/e9d1ac60fab076ec6c2b01dd17d8a855

### Expected behavior

It shouldn't error out. The comparison between TSC output and SWC is pretty big and some weird `_class || _class || _class` noted there in `swc` output that may be the root cause of this issue. 

### Version

1.2.120

### Additional context

Above is mentioned the `gist` with all the code that you can download and just run `npm install` `npm run start` to get the error.

### LINKED ISSUE #3342: Defining a field on the prototype in a class field decorator makes the field unwritable
### Describe the bug

If in a class field decorator like so:

```
class Foo {
   @d
   member = 4
}
```

I define a property on `Foo`s prototype using: 
`Object.defineProperty(Foo.prototype, "member", {get() {...} set() {...}, ...}`

Then, that property definition will be overwritten by `_applyDecoratedDescriptor`, with a descriptor that includes `writable: false`. 

This behavior differs from `tsc` where the property is not overwritten. 

(note, this bug occurs whether `legacyDecorators` set to true or false. It doesn't work in either case)

This bug was originally found when using the [`lit`](https://lit.dev) library from Google and has been reduced to the following:

### Input code

```typescript
/* 
  Proxy any reads and writes to this member. We'll store the value at some private symbol
  and read or write to it when the member is accessed.
*/
const d = (proto: any, name) => {
  let self = proto.constructor;
  const key = typeof name === "symbol" ? Symbol() : `__${name}`;

  Object.defineProperty(self.prototype, name, {
    get(): any {
      return (this as { [key: string]: unknown })[key as string];
    },
    set(value: unknown) {
      const oldValue = (this as {} as { [key: string]: unknown })[
        name as string
      ];
      (this as {} as { [key: string]: unknown })[key as string] = value;
    },
    configurable: true,
    enumerable: true,
  });
};

class Test {
  @d
  member = 4;
}

console.log(new Test().member);
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "decorators": true,
      "loose": true
    },
    "target": "es2022"
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.133&code=H4sIAAAAAAAAA5VSy07DMBC85ytGCImkqtILp0YB%2FgAkEBwQat1k24Y6dmU7LVGVf2dtRxXlgrgk1u7sPNaeTZAAT0Z%2F9RCqhyFRWz7VOJrGkYXTcNvGoqV2RSbHG91ICeu0IW4QDkJ2BOFgdUvYm%2BYgHMH27UpLZvZMnhPaREZP2Dgct6TCfOQFK4iqImupzpPJLKm0sg41SqR7o52ee3dTKNFShvIOJ%2BaWxKok1wwKmDwMma5icwX3I8eOega4fk96HeZRliWuosMr3OM5nNIMcywXi%2BuTxwzLImGGx9UnVS6vad0o4iXtybg%2B9Zp5UPSs0dQ0OAI25NIsmB0L4PiuMwppWKOwOOGdPc15h6ZRm485OrVT%2BqgwZL7hIWOrCAzDNPwsM4dlnweys0RMqmX9Gm6j%2FCE2%2FCU5UiDu5iw%2BlkcP%2BA%2FjZQh2E2xfhGHD62bTGbGSnIfvjGKdVNfSr%2BqQFcnA11FJYS1eiJP63A81f8bXU%2BKWMUl4NVpSLvUmVXQM4DTLIyorvgHkLTWQ7QIAAA%3D%3D&config=H4sIAAAAAAAAAyWMQQqAMAwE%2F7JnD9JjfxNqFKXYko2glP7dirfZYZmGgwmxoYpR7SM%2Bp8uNCH%2BqMtleHROcQ62SqRMWTcXEixHR7Roml0L9Rx9fsU19FJRhDgG9v5KoSVNnAAAA

### Expected behavior

This code should log `4`.

In general, the decorator application logic for class fields should not define a property if a property is already defined.

### Actual behavior

This code fails to set the value of the field during object construction, because the property is marked as not writable and thus logs `undefined`.

### Version

1.2.133

### Additional context

This bug was likely introduced with the (admittedly kinda ironic PR): [fix(es/helpers): Fix decorators #3105](https://github.com/swc-project/swc/pull/3105/).

In typescript, class fields aren't defined as properties at all (at least by default). If you enable `useDefineForClassFields`, (which is more inline with the spec) you will *still* have a behavior discrepancy. `swc` will still log undefined, but `tsc` will log `4`. However, the property description from the decorator gets overwritten in the constructor with a more mundane:

```
Object.defineProperty(this, "member", {
  enumerable: true,
  configurable: true,
  writable: true,
  value: 4
});
```

meaning our custom setter and getter will never run.

Long story short though, this bug results in projects that build successfully with `tsc` but not with `swc` and thus should probably be fixed.

### LINKED ISSUE #3521: Using static property from a class in decorators on the same class
### Describe the bug

The output code when run throws "ReferenceError: Cannot access 'FormulaRuleEntity' before initialization". The same code built using tsc is valid.

The output code using tsc looks like this
``` javascript
var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {
    var c = arguments.length, r = c < 3 ? target : desc === null ? desc = Object.getOwnPropertyDescriptor(target, key) : desc, d;
    if (typeof Reflect === "object" && typeof Reflect.decorate === "function") r = Reflect.decorate(decorators, target, key, desc);
    else for (var i = decorators.length - 1; i >= 0; i--) if (d = decorators[i]) r = (c < 3 ? d(r) : c > 3 ? d(target, key, r) : d(target, key)) || r;
    return c > 3 && r && Object.defineProperty(target, key, r), r;
};
var FormulaRuleEntity_1;
const Test = (name) => {
    return (target, key) => {
    };
};
let FormulaRuleEntity = FormulaRuleEntity_1 = class FormulaRuleEntity {
};
FormulaRuleEntity.NAME = 'cat';
FormulaRuleEntity = FormulaRuleEntity_1 = __decorate([
    Test(FormulaRuleEntity_1.NAME)
], FormulaRuleEntity);
//# sourceMappingURL=index.js.map
```

### Input code

```typescript
const Test = (name: string): any => {
    return (target: any, key: string) => {
        // stuff
    };
}

@Test(FormulaRuleEntity.NAME)
class FormulaRuleEntity {

    static readonly NAME = 'cat';

}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "decorators": true
    },
    "target": "es2022",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.138&code=H4sIAAAAAAAAA2VOvQrCMBDeA3mH25qC2L2lokPddBBf4IhpCaZXSC5DkL67iQUR%2FIYb7vuVQi8UGO4mnx4U4WxaCOwtTXULSAn6A7ykgAxvOHoCxegnwx92B0%2BTvoYfbUHTZCKO4%2FZZOylWKaQ4ljJ1XvwcHd6iMwOx5bS%2Fni5DLYV2GAL80SV4CwqMbHVeg4%2BFXILiy9srjVx1RZRb3o8lVxnYAAAA&config=H4sIAAAAAAAAA0WOQQ7DIAwE%2F%2BJzDhWHHvhDH4GIE1EBRrYjFSH%2BXoLS5mbt7ozc4C0ebIPiWJDPS2pW9wELWguK51AUFlAZ0eai4AIremKnxAJW%2BcA%2Basc76oBQzMOYAUQiwT%2BSQg5bPfWeUmEUuSuX9%2Fhb9uFKtB5n0OYH0%2FmEfjsuLsjrGs4fvucDfqLKAAAA

### Expected behavior

The output code should not throw "ReferenceError: Cannot access 'FormulaRuleEntity' before initialization"

### Actual behavior

throw "ReferenceError: Cannot access 'FormulaRuleEntity' before initialization"

### Version

1.2.138

### Additional context

_No response_

### LINKED ISSUE #3592: Decorator-based dependency-injection not working (NestJS / TypeORM)
### Describe the bug

Fields decorated like this:

```ts
  @InjectRepository(User)
  private userRepo: Repository<User>;
```
are not being defined when `swc` is used

### Input code

Freshly created NestJS project with minimal reproduction:

https://github.com/vith/nest-swc-test

```console
$ pnpm run start
...
this.userRepo is present
```

```console
$ pnpm run start:swc
...
this.userRepo is undefined
```

https://github.com/vith/nest-swc-test/blob/main/src/other/other.service.ts is the main file for the issue reproduction

### Config

```json
{
  "jsc": {
    "baseUrl": ".",
    "paths": {
      "src/*": ["src/*"]
    },
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "decorators": true,
      "dynamicImport": false
    },
    "target": "es2020",
    "externalHelpers": true,
    "transform": {
      "legacyDecorator": false,
      "decoratorMetadata": true
    }
  },
  "module": {
    "type": "commonjs"
  }
}
```

### Playground link

_No response_

### Expected behavior

expect `this.userRepo is present` to be logged

### Actual behavior

`this.userRepo is undefined` is logged

### Version

'@swc/core': 1.2.140

### Additional context

also tried ` "legacyDecorator": true,`; same result

### LINKED ISSUE #4283: _dec is not a function
Hi,
I'm encountering an error using swc-loader as typescript compiler : **_dec is not a function** in a project Angular.

I have this service : 

immagine

used in 

```
@NgModule({
  imports: [
    BrowserModule,
    UpgradeModule,
    HttpClientModule
  ],
  providers: [
    TestService
  ]
})
```

and this configuration in webpack :

`loader: "swc-loader", options: {
								jsc: {
									externalHelpers: true,
									parser: {
										syntax: "typescript",
                                        decorators: true
									},
                                    transform: {
                                        legacyDecorator: true,
                                        decoratorMetadata: true
                                    }
								}
							}`

but during compilation I get the following error : 

immagine

What identifies the error **_dec is not a function**?

Additional NOTES

I compile the file **test.service.ts** with **npx swc** :

import * as angular from 'angular';
import { HttpClient, HttpParams } from '@angular/common/http';
import { Inject, Injectable } from '@angular/core';
var _class;
var _dec = typeof Reflect !== "undefined" && typeof Reflect.metadata === "function" && Reflect.metadata("design:paramtypes", [
    typeof HttpClient === "undefined" ? Object : HttpClient
]), _dec1 = typeof Reflect !== "undefined" && typeof Reflect.metadata === "function" && Reflect.metadata("design:type", Function), _dec2 = function _dec2(target, key) {
    return Inject(HttpClient)(target, undefined, 0);
}, _dec3 = Injectable();
export var TestService = _class = _dec3(_class = _dec2(_class = _dec1(_class = _dec((_class = /*#__PURE__*/ function() {
    "use strict";
    function TestService(http) {
        swcHelpers.classCallCheck(this, TestService);
        this.http = http;
    }
    var _proto = TestService.prototype;
    _proto.getData = function getData() {
        var _this = this;
        return swcHelpers.asyncToGenerator(regeneratorRuntime.mark(function _callee() {
            var opts, response;
                    case 0:
                        opts = {
                            params: new HttpParams({
                                fromString: "test=1"
                            })
                        };
                        _ctx.next = 3;
                        return _this.http.get("https://service_test", opts);
                    case 3:
                        response = _ctx.sent;
                    case 4:
                    case "end":
                        return _ctx.stop();
                }
            }, _callee);
        }))();
    };
    return TestService;
}()) || _class) || _class) || _class) || _class) || _class;
angular.module("prova").service("TestService", TestService);

kdy1:
**Description:**

# TODOs
 - export default decl
 - export decl

# Investigation

image

It works if we change

```js
_initializerDefineProperty(this, "testProp", _descriptor, this);
```

to

```js
this.testProp = 'hello';
```

 - Closes #2081
 - Closes #2655
 - Closes #3035
 - Closes #3342
 - Closes #3521
 - Closes #3592
 - Closes #4283


kdy1 (commented):
swc-bump:

kdy1 (commented):
reset ci

PR #4504, opened 2022-05-02
fix(html/parser): `select` parsing

alexander-akait:
**Description:**

just bug fix

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser

PR #4505, opened 2022-05-03
fix(html/parser): fux bug with unwrapping element without parent

alexander-akait:
**Description:**

Bug fix + tests

**BREAKING CHANGE:**

No

No


alexander-akait:
Rebased too

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_html_parser

PR #4506, opened 2022-05-03
fix(html/parser): parsing table with table

alexander-akait:
**Description:**

Bug fix

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser

Issue #4107, opened 2022-03-21
@swc/core behavior inconsistency between macOS and Linux


roland-reed:
### Describe the bug

When requiring two different versions of @swc/core (for example, two dependencies requires two different versions of @swc/core) in Linux, an error is thrown, but not in macOS.

The error message in Linux is:

```plain
Error: /dir_to_home/swc-inconsistency/node_modules/.pnpm/@swc+core-linux-x64-gnu@1.2.147/node_modules/@swc/core-linux-x64-gnu/swc.linux-x64-gnu.node: cannot allocate memory in static TLS block
    at Object.Module._extensions..node (internal/modules/cjs/loader.js:1122:18)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/dir_to_home/swc-inconsistency/node_modules/.pnpm/@swc+core@1.2.147/node_modules/@swc/core/binding.js:181:45)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
```

I test it in macOS 12.3 with Node.js 14.18.1 and Debian 9 with Node.js 14.16.1.

I create a reproduce repo in https://github.com/roland-reed/swc-inconsistency.

### Expected behavior

It should not throw error in Linux when requiring two different versions of @swc/core.

### Version

@swc/core@1.2.156

kdy1:
Pleaae use latest version of `@swc/core`

roland-reed:
> Pleaae use the latest version of `@swc/core`

Is this a known issue? I have tested with 1.2.159 and 1.2.158, error is still be thrown.

kdy1:
Your log message prints a different version

roland-reed:
```plain
Error: /dir_to_home/workspace/swc-test/swc-inconsistency/node_modules/.pnpm/@swc+core-linux-x64-gnu@1.2.158/node_modules/@swc/core-linux-x64-gnu/swc.linux-x64-gnu.node: cannot allocate memory in static TLS block
[...]

PR #4514, opened 2022-05-03
fix(node): tls allocation issue by updating linux gnu build image

Brooooooklyn:
Fix https://github.com/swc-project/swc/issues/4107

https://github.com/Brooooooklyn/swc/runs/6286599487?check_suite_focus=true passed

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -14,7 +14,7 @@ jobs:
       ${{ !contains(github.event.head_commit.message, 'chore: ') }}
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Cache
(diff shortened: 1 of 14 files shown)


Brooooooklyn:
@kdy1 ready

kdy1:
Seems like the integration test is failing? https://github.com/swc-project/swc/runs/6285669053?check_suite_focus=true

kdy1:
I'm not sure why it's faiiling, though

Brooooooklyn:
Hmmm, maybe because I upgrade the `node@14` to `node@16` in the integration test. Should we revert the upgrade?

kdy1:
I think it's better to update integration test (if it uses node 16 in recent version)
I'll modify this PR

kdy1 (approved):
swc-bump:
 - ''

PR #4518, opened 2022-05-04
fix(es/compat): Revert a wrong patch about private properties

kdy1:


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_compat

PR #4525, opened 2022-05-04
fix(es/compat): memorize computed props in logical assignment

### LINKED ISSUE #4521: Left-hand side expressions of assignments are evaluated twice
### Describe the bug

In code like `o[getKey()] ??= "someValue"`, `getKey()` should be evaluated just once.

Instead, it is evaluated twice if the value is not set. If `getKey()` is not pure, this causes problems.

### Input code

```typescript
o[fn()] ??= true;
o[fn()] ||= true;
o[fn()] &&= true;
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.176&code=H4sIAAAAAAAAA22OQQuCQBCF7%2Fsr3m0UQ%2FKaLB06CnnpFh3KJllbdkDXg6j%2FvU0pCJrTg3nfx7PsYaCxzZV69K7yRhxq9gUPUYxRIVwlrhPLqZU6ovWH6mot3ynOl0bLvm8dksTkalbqDXhI0I6zkvNHd8E0afi253ztfKXlreEqDOlogzWnTx66SOK0EeMi2lAMrTUoI%2BxBZUHYgY7lCSGGET%2B2g%2FTOc7vqzMJl%2F6gXl5lWVfwAAAA%3D&config=H4sIAAAAAAAAA0WMTQrEIAxG75K1286id5hDBCctFv9IUhgR714tlu7C915ehUMsrBUyshCPS0pU%2FMMKZAOKZZcVTNf6tKEXagYUeScdiiyd%2BZSEJjUQXHRbGSWbQmYSeRHG3T9m66GQfucYKmjJdAc%2F0N7G%2FHPynaLySe0Cj9ke9LUAAAA%3D

### Expected behavior

The playground code compiled and executed should set the object key `1`, and call `getKey()` just once.

### Actual behavior

The key `2` is set, and `getKey()` is called twice.

### Version

1.2.176

### Additional context

Both Firefox and Node.js evaluate the left hand side of the assignment just once when using these operators directly.

It is also clear from the spec that it should be that way.

> 1\. Let lref be the result of evaluating [LeftHandSideExpression](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#prod-LeftHandSideExpression).
> [...]
> 6\. Perform ? [PutValue](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-putvalue)(lref, rval).

https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-assignment-operators

Austaras:
**Description:**

**BREAKING CHANGE:**

closes #4521


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_ast
 - swc_ecma_transforms_compat

kdy1 (approved):

PR #4526, opened 2022-05-04
fix(html/parser): bug with optional end li

alexander-akait:
**Description:**

Fix bug with optional end tag in `li`

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser

PR #4536, opened 2022-05-05
Preserve class expression identifiers if keepClassNames is set to true

### LINKED ISSUE #4529: mangle keep_classnames fails when class is not exported in es6 module
### Describe the bug

When using module type `es6`, the keepClassNames only seem to take effect if the class is exported.

### Input code

```typescript
@decorator
export class ImportantName {} 

console.log(ImportantName.name)
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "decorators": true
    },
    "target": "es2019",
    "minify": {
      "compress": {
        "keep_classnames": true
      },
      "mangle": {
        "keep_classnames": true
      }
    },
    "keepClassNames": true
  },
  "module": {
    "type": "es6"
  },
  "minify": true,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.177&code=H4sIAAAAAAAAA1WOvQ7CMBCD9zyFR1iaV2BlYeEJrtdTiUhy1V34qSrenVZ0YbFs2dLnGGFS9CmQ96TWQHWAi4AzuaNSESTfNwP6GYXqmCXECH%2BlxrdUR7g%2BjAVtnlZRXNnS1NDb2jnS6ojv4TQIq1FTCzvrxziXLVBtlw22fBACa3XN0mUdD391tx06hi%2Blgw9ztgAAAA%3D%3D&config=H4sIAAAAAAAAA3WOywrCQAxF%2FyXrWagLwW5d6y9ImMYy2nmQpOBQ%2Bu9OqlBcuLsk5yR3hod46GYoyEJsSWpSfEEHWguJ51AUHPTkM6NmFuiUJ1ocKPJA2kCSw25%2FalAMKdyrHfE5FiYRy0%2BicvMjiiSMtPkR0zDSf6Ihtjjb%2FPpr5n76mNZxbXCEZftvmIMgly%2B3am%2FmE1rn6wAAAA%3D%3D

### Expected behavior

The class name should not be mangled when using es6 and the class is not exported.  It should respect the keepClassName settings

### Actual behavior

Class name is mangled

### Version

1.2.177

### Additional context

We are attempting to use SWC to minify code containing Catalyst Components, which use a `@controller` decorator to register the class as a web component.  The decorator is responsible for all of the work, so the class is not actually reference anywhere and is not typically exported from the file.  Catalyst uses the class name to determine what the custom element name should be, which is why we need to persist it during minification.

koddsson:
**Description:**

Hello! This is our attempt to fix #4529. I think this is the correct fix but having to update `crates/swc_ecma_minifier/tests/terser/compress/harmony/class_name_can_be_preserved/output.js` made me think that maybe we need to guard the code better?

My thought is that `keepClassNames` should always keep any class name but maybe there are nuances that I'm missing.

I'm happy to be mentored to get this PR mergable if it's not already. I like Rust but I haven't had many chances to actually work on projects using rust.

**BREAKING CHANGE:**

I don't think this is a breaking change unless we consider the changes described in the diff of `crates/swc_ecma_minifier/tests/terser/compress/harmony/class_name_can_be_preserved/output.js` to be breaking.

Fixes #4529


kdy1 (approved):
Thank you!
I think this is a correct fix, and maybe I made a mistake while updating test references

---

swc-bump:
 - swc_ecma_minifier

PR #4537, opened 2022-05-05
fix(html/parser): cdata parsing

alexander-akait:
**Description:**

Bug fix, maybe we can improve it in future (improve perf), and get `adjusted_current_node` when it is necessary, but it is will require more refactor, so let's postpone it

**BREAKING CHANGE:**

No

No


kdy1 (approved):
LGTM

---

swc-bump:
 - swc_html_parser

PR #4138, opened 2022-03-23
fix(ecma_parser) Allow arrow func without parenthesisised param in conditional for TS

williamtetlow:
**Description:**
Fixes https://github.com/swc-project/swc/issues/2174, https://github.com/swc-project/swc/issues/4067

In https://github.com/swc-project/swc/pull/3685#issuecomment-1050701591 we removed TS support for arrow function without parenthesis on single param to match the behaviour of TSC.

As of https://github.com/microsoft/TypeScript/pull/47550 TSC supports this syntax.

This PR adds support for syntax.

**BREAKING CHANGE:**

**Related issue (if exists):**
 - Fixes https://github.com/swc-project/swc/issues/2174
 - Fixes https://github.com/swc-project/swc/issues/4067

Issue #4067, opened 2022-03-17
Regression: Cannot parse arrow fn exprs in conditional expr cons and alt in object literals in TypeScript


dsherret:
### Describe the bug

Regression of #2174

https://github.com/swc-project/swc/pull/3685#issuecomment-1050701591

This syntax no longer parses, yet it is valid JavaScript. ~~Maybe the TS compiler can't figure out how to parse it right now~~(Edit: it can figure this out now), but it works in V8.

https://astexplorer.net/#/gist/d8417911fb6ea22a87b63d441b2a0e87/20b56c10cf6e34b3e7ddc60e9328702e09c930fd

cc @williamtetlow

### Input code

```typescript
const x = {
    prop: true
        ? fn => ({ })
            : fn => true,
};
```

### Playground link

https://play.swc.rs/?version=1.2.151&code=H4sIAAAAAAAAA0vOzysuUahQsFWo5uVSAIKCovwCK4WSotJUCB8E7BXS8hRs7RQ0qhVqNRHCIGAFlQJp0OHlqrUGAGjLrV1RAAAA&config=H4sIAAAAAAAAA0WOQQ6DMAwE%2F%2BIz1%2FaQP%2FCIKBiUipDIa6RGiL%2FXoaG9edfeWR%2F0QiB3UPECljahburf5EhrYQSJRWkghVmzX8GnCS8Lq50wHrZbcwaTU9l5oBS3ONcGCjkVYaDnbOW3ZeUbY5yUp70Zx9V18Z7Gg0oM%2Bot95Zgn%2Fn9wt%2FSTiLGj2hPnBwXfhU7WAAAA

### Expected behavior

Should parse.

### Actual behavior

```
error: Expected ':', got ','
 --> input.ts:4:25
  |
4 |             : fn => true,
  |                         ^



Caused by:
    0: failed to process js file
    1: failed to process js file
    2: Syntax Error
```

### Version

1.2.151

kdy1:
I think the patch for #2174 was removed for some reasons, but I don't remember the PR removed it.

dsherret:
It was in #3685. I think it's reasonable that this doesn't parse in TS given that the TS compiler can't figure it out, but it should still at least parse in JS.

williamtetlow:
@dsherret this does parse as JS. The playground you shared has language set to Typescript


see

[...]

PR #4540, opened 2022-05-05
fix(es/parser): Revert "Allow arrows without parenthesized param in typescript ( #4138)"

williamtetlow:
**Description:**
Context: https://github.com/swc-project/swc/pull/4138#issuecomment-1117976098

Typescript compiler has reverted https://github.com/microsoft/TypeScript/pull/47550 in https://github.com/microsoft/TypeScript/pull/48940 and plan to release this change in 4.8 instead of 4.7 so we should revert support until then.

**BREAKING CHANGE:**

- breaks https://github.com/swc-project/swc/issues/4067


williamtetlow:
@dsherret fyi ^

kdy1 (approved):
Thank you!


---

swc-bump:
 - swc_ecma_parser

PR #4541, opened 2022-05-05
feat(swc/plugin): make pluginsourcemap as sourcemapper

kwonoj:
**Description:**

This PR mostly reexports some of types from source_map, and also makes `PluginSourceMapProxy` as `SourceMapper`.

It is possible to write plugin's visitor as isomorphic to be used between wasm / host side custom pass, but in that case visitor need to accept few types works differently between those 2 context: mainly all of the proxy impl. For the comments, it already implements `trait Comment` but in case of sourcemap its proxy impl is strictly typed so bit awkward to define types to be used commonly. Instead, this PR makes sourcemap to use sourcemapper trait as similar to comments.

```
fn create_visitor(source_map: ${here})
```

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!


---

swc-bump:
 - swc_plugin
 - swc_plugin_proxy

kdy1 (approved):

PR #4302, opened 2022-04-11
feat(es/parser): typeof on #private Fields

magic-akari:
**Description:**

**BREAKING CHANGE:**
- swc_ecma_ast
- swc_ecma_codegen
- swc_ecma_parser
- swc_ecma_transforms_proposal
- swc_ecma_transforms_typescript
- swc_ecma_visit
- swc_estree_ast
- swc_estree_compat

**Related issue (if exists):**
- close #4287
- https://github.com/microsoft/TypeScript/pull/47696

Issue #4287, opened 2022-04-08
TypeScript - Support `typeof` on private fields


dsherret:
### Describe the feature

https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#typeof-on-private-fields

```ts
class Container {
    #data = "hello!";

    get data(): typeof this.#data {
        return this.#data;
    }

    set data(value: typeof this.#data) {
        this.#data = value;
    }
}
```

### Additional context

Currently errors:

https://play.swc.rs/?version=1.2.157&code=H4sIAAAAAAAAA0vOSSwuVnDOzytJzMxLLVKo5uVSAALllMSSRAVbBaWM1JycfEUla14uiER6aokCSE5D00qhpLIgNT9NoSQjs1gPogGqGwSKUktKi%2FKQJK0hcrUwk4phJpUl5pSmYjFNE9k4JEtsFcA6EObVAgCfa7ZexQAAAA%3D%3D&config=H4sIAAAAAAAAA0WOQQ6DMAwE%2F%2BIz1%2FaQP%2FCIKBiUipDIa6RGiL%2FXoaG9edfeWR%2F0QiB3UPECljahburf5EhrYQSJRWkghVmzX8GnCS8Lq50wHrZbcwaTU9l5oBS3ONcGCjkVYaDnbOW3ZeUbY5yUp70Zx9V18Z7Gg0oM%2Bot95Zgn%2Fn9wt%2FSTiLGj2hPnBwXfhU7WAAAA

PR #4542, opened 2022-05-05
Revert "feat(es): Support `typeof` with private fields (#4302)"

williamtetlow:
**Description:**
Context:https://github.com/swc-project/swc/pull/4302#issuecomment-1117984506

Typescript compiler has reverted https://github.com/microsoft/TypeScript/pull/47696 in https://github.com/microsoft/TypeScript/pull/48959 so we should revert support until then.

**BREAKING CHANGE:**

- breaks https://github.com/swc-project/swc/issues/4287


williamtetlow:
@dsherret also fyi 😄

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_ast --breaking

PR #4543, opened 2022-05-05
fix(html/parser): newline in textarea

alexander-akait:
**Description:**

Fix bug with newlines in textarea.

Ideally we should always output `\n` in AST and keep original newline in `raw` property, but let's postpone it, we have TODO for this.

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_html_parser

PR #4557, opened 2022-05-07
fix(es/minifier): skip getter/setter when collpasing vars

### LINKED ISSUE #4515: "[variable name] is not defined" when using SWC minifier
### Describe the bug

When using the SWC minifier, it produces output that causes an error in the browser:
  
Screen Shot 2022-05-03 at 9 09 12 AM

I have a small repro [in this Gist](https://gist.github.com/nolanlawson/3e5a6a081d23e8dfb87b453b641b468e). To repro, run:

    git clone git@gist.github.com:3e5a6a081d23e8dfb87b453b641b468e.git repro
    cd repro
    npm i
    npm start

Then navigate your browser to `localhost:3000`. (I assume you're using Node.js v16 and npm v8.)

I am using the SWC minifier in Node.js like so:

```js
  minifiedCode = swc.minifySync(transpiledCode, {
    mangle: true,
    compress: true,
  }).code;
```

If I switch to the equivalent code in Terser, the same error does not occur. In the repro, you can use the environment variable `USE_TERSER` to see the working (non-SWC) version:

```
USE_TERSER=1 npm run build
npm start
```

Screen Shot 2022-05-03 at 9 12 15 AM

### Input code

```typescript
// The code is too long to include here. Please see the repro link above.
```

### Config

```json
# N/A, I am not using .swrc
```

### Playground link

_No response_

### Expected behavior

A runtime error in the browser should not occur.

### Actual behavior

A runtime error in the browser occurs.

### Version

1.2.175

### Additional context

_No response_

Austaras:
**Description:**

**BREAKING CHANGE:**

closes #4515


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_minifier

PR #4566, opened 2022-05-07
fix(htm/parser): recovery attributes on html and body

alexander-akait:
**Description:**

Bug fix, it is not ideally, but we have `TODO` about optimize and I improve it in the next PRs, when start to work on parsing `document_fragment`

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_html_parser

PR #4568, opened 2022-05-08
refactor(swc/plugin): align deps

kwonoj:
**Description:**

This is minor bikeshedding for the swc_plugin's deps to use reduced number of deps instead of having individual for the reexport.

This won't make any major differences, but helps to import a visitor aims to be used for the plugin & custom pass both:

```
// visitor
- swc_ecma_script
- ...

// plugin
- swc_plugin
- visitor

// swc/custom-pass
- swc_ecma_script
- visitor

```

plugin / swc both will work by having matching version of swc_ecmascript only, instead of each (ast, visit, etcs) to avoid conflict. This is what I found while try to align versions for the https://github.com/vercel/next.js/pull/36692.

**BREAKING CHANGE:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3937,10 +3937,8 @@ version = "0.50.1"
 dependencies = [
  "swc_atoms",
  "swc_common",
- "swc_ecma_ast",
  "swc_ecma_quote",
- "swc_ecma_utils",
- "swc_ecma_visit",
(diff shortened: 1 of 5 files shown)


kdy1 (approved):
swc-bump:
 - swc_ecmascript
 - swc_plugin

PR #4577, opened 2022-05-08
refactor(html/parser): resolve todo 

alexander-akait:
**Description:**

Resolve todo to avoid memory leak, before it was broken, but I fixed it so let's unccoment

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser

PR #4580, opened 2022-05-09
fix(es/minifier): Preserve more identifiers if `eval` or `with` exists

kdy1:
**Description:**

**BREAKING CHANGE:**


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

kdy1 (commented):
reset ci

PR #4581, opened 2022-05-09
fix(es/minifier): Don't create an invalid labeled statement

kdy1:
**Description:**

 - Don't create an invalid labeled statement.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4584, opened 2022-05-09
feat(es/lint): implement prefer-const without options

ArturAralin:
**Description:**
Here is base implementations of prefer-const rule. Options I'll implement future

**BREAKING CHANGE:**

No


crates/swc_ecma_lints/src/rules/prefer_const.rs
```diff
@@ -0,0 +1,257 @@
...
+            var_decl
+                .decls
+                .iter()
+                .for_each(|var_decl| self.collect_decl_pat(var_decl.init.is_some(), &var_decl.name))
+        }
+
+        var_decl.visit_children_with(self);
+    }
+
+    fn visit_assign_expr(&mut self, assign_expr: &AssignExpr) {
```
kdy1, line 216: I think we should visit children of this node

crates/swc_ecma_lints/src/rules/prefer_const.rs
```diff
@@ -0,0 +1,257 @@
...
+            var_meta.postinitialized = true;
+        }
+    }
+
+    fn consider_mutation(&mut self, pat: &Pat, destructuring_assign: bool) {
+        match pat {
+            Pat::Ident(BindingIdent { id, .. }) => {
+                self.consider_mutation_for_ident(id, destructuring_assign);
+            }
+            Pat::Array(ArrayPat { elems, .. }) => elems.iter().for_each(|elem| {
```
kdy1, line 155: Please use flatten instead

crates/swc_ecma_lints/src/rules/prefer_const.rs
```diff
@@ -0,0 +1,257 @@
...
+            }),
+            Pat::Object(ObjectPat { props, .. }) => {
+                props.iter().for_each(|prop| match prop {
+                    ObjectPatProp::KeyValue(KeyValuePatProp { value, .. }) => {
+                        self.consider_mutation(value.as_ref(), true);
+                    }
+                    ObjectPatProp::Assign(AssignPatProp { key, .. }) => {
+                        self.consider_mutation_for_ident(key, true);
+                    }
+                    _ => {}
```
kdy1, line 168: I'm not sure, but we may need to consider rest patterns

kdy1 (changes requested):
Thanks!

crates/swc_ecma_lints/src/rules/prefer_const.rs
ArturAralin: Yes, you're right. eslint rule considers rest declaration

crates/swc_ecma_lints/src/rules/prefer_const.rs
ArturAralin: instead of `for_each`?

kdy1: I mean `iter().flatten()`

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_lints

kdy1 (approved):

PR #4588, opened 2022-05-09
feat(css/parser): add method to return error message

g-plane:
**Description:**

This can allow downstream developers to retrieve error message.


alexander-akait:
I think we will need the same for HTML, because now it is copy/paste from CSS parser :+1:

g-plane:
@alexander-akait I think it can be done in another PR.

g-plane:
ping @kdy1 .

kdy1 (approved):
swc-bump:
 - swc_css_parser

PR #4593, opened 2022-05-09
fix(es/minifier): Fix handling of `NaN`

kdy1:
**Description:**

 - `expr_simplifier` now uses `NaN#unresolved` instead of f64 with `NaN` value.


kdy1 (commented):
reset ci

PR #4598, opened 2022-05-10
fix(es/fixer): Handle `new yield`

kdy1:
**Description:**

This is part of minifier stabilization.


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base

PR #4602, opened 2022-05-10
chore(ci): Use nextest and partition tests

kdy1:
- We now cache files related to cargo. Previously this was impossible because swc has lots of modules and it resulted in a cache eviction everytime.
 - We now partition tests.


kdy1:
I found a better way to speed up ci

PR #4603, opened 2022-05-10
chore(es/minifier): Add a CLI crate

kdy1:
**Description:**

I'll use this to debug the minifier.


kdy1:
I'll use different repo due to cycle deps

PR #4611, opened 2022-05-11
feat(html/*): refactor parser and codegen

alexander-akait:
**Description:**

For parser:
- Added `parse_document`, `parse_document_fragment` (for partial parsing) + we have `parse_as`
- Removed parse fucntion for `&str`, I think we don't need because we have `cm.new_source_file(FileName::Anon, html_str)` (we support parsing `&str` for CSS parser, I think we should drop them too)
- TODOs

For codegen:
- fix many bugs
- test codegen for `recovery` + `html5lib_tests` (we have couple `exclude`, because currently we don't support `raw` for some AST nodes and can't print code as is, I will improve this)
- feat - `scripting_enabled` option like for parser

+ TODOs for better testing - I want to send them in the separate PR for better review

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_codegen
 - swc_html_parser

PR #4613, opened 2022-05-11
fix(html/parser): small improve

alexander-akait:
**Description:**

Reduce size, we don't need some fields + fixed maximum longest entity

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - ''

PR #4616, opened 2022-05-11
feat(common): Reserve `BytePos(0)` for dummy spans

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_common --breaking

PR #4617, opened 2022-05-11
test(es): Add a test for merging configurations

kdy1:
**Description:**

I got a report but seems like `module` is correctly merged.
I decided to add it as a test anyway


kdy1 (commented):
swc-bump:
 - ''

PR #4623, opened 2022-05-11
feat(html/parser): Add a method to get error message

g-plane:
**Description:**

Just like #4588 , but for HTML parser.

Also, marked lifetime as `'static` in CSS parser to address lifetime issue.


alexander-akait (approved):
Looks good, I found we use `msg` name for `swc_ecma_parser` and `message` for CSS/HTML, I thinm we should consider to get one name

Also `swc_ecma_parser` use `into_diagnostic` instead `to_diagnostics`

So I think we should be the same to better DX

But let's do it after merge

@kdy1

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_parser
 - swc_html_parser

PR #4626, opened 2022-05-11
feat(html/parser): improve public API

alexander-akait:
**Description:**

Remove generic version - we don't need alias + reduce binary size

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser --breaking

PR #4627, opened 2022-05-11
fix(html/parser): small improves in parser

alexander-akait:
**Description:**

Remove TODOs, because they should be unreachable - if somebody faced with it, it should be considered as bug

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser

PR #4629, opened 2022-05-11
fix(html/codegen): update default codegen

alexander-akait:
**Description:**

Improve default codegen, still not perfect (we need more `raw`), but better than we have

**BREAKING CHANGE:**

No

No


kdy1 (changes requested):
CI failed

alexander-akait:
@kdy1 Fixed

kdy1 (approved):
swc-bump:
 - swc_html_codegen

PR #4633, opened 2022-05-12
fix(html/parser): fix newlines

alexander-akait:
**Description:**

Use `\n` for new lines like in spec, also prepare code for `raw` (where we will store `\r` or `\n` or `\r\n`)

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---


swc-bump:
 - swc_html_parser

kdy1 (changes requested):
Can you rebase? CI failed

PR #4638, opened 2022-05-12
fix(es/codegen): Fix codegen of array literals with no elements

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_ecma_codegen

PR #4645, opened 2022-05-12
fix: ensure parameter type is resolved to parent when name is same

### LINKED ISSUE #4610: Scope analysis fails for parameter with type identifier with same name
### Describe the bug

Given the following code:

```ts
import type { Foo } from "./foo.ts";

function _bar(...Foo: Foo) {
  console.log(Foo);
}
```

### Expected behavior

```ts
import { Foo__1 } from "./foo.ts";
function _bar__1(...Foo__2: Foo__1) {
    console.log(Foo__2);
}
```

### Actual behavior

Started doing this in the most recent upgrade for us:

```ts
import { Foo__1 } from "./foo.ts";
function _bar__1(...Foo__2: Foo__2) {
    console.log(Foo__2);
}
```

### Version

swc_ecma_transforms_base 0.82.5 (main)

I looked through the code a bit in resolver/mod.rs, but on first glance I'm not sure what's going wrong.

dsherret:
**Description:**

Fixes some usages of parameter types. swc does not still properly handle some other scenarios, but not a big deal because it's super edge case and I will open an issue later.

Closes #4610


crates/swc_ecma_transforms_base/tests/ts-resolver/deno/lint/705/output.ts
```diff
@@ -1,4 +1,7 @@
 import { Foo__1 } from "./foo.ts";
-function _bar__1(...Foo__2: Foo__2) {
```
dsherret, line 2: I think the test suite could use improvement if regressions like this can happen.

Maybe instead of using numbers here the tests should instead allow using unique values (which may be numbers) to link identifiers instead of the context number. So for the test to pass, the identifiers in the output just have to be the same instead of having to have a specific syntax context number. This would also reduce a lot of churn in the test output.

kdy1: I'm not against it, but it sounds like a lot of work

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_base

kdy1 (approved):

PR #4647, opened 2022-05-12
fix: template AST

alexander-akait:
**Description:**

Fix `<template>` AST - store children in `DocumentFragment` like in spec

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_codegen
 - swc_html_parser

PR #4649, opened 2022-05-13
refact(es/react): proper ident scope for react refresh

Austaras:
**Description:**

**BREAKING CHANGE:**


crates/swc_ecma_transforms_react/src/refresh/hook.rs
```diff
@@ -122,15 +120,17 @@ impl<'a> HookRegister<'a> {
...
             let ident = match &hook {
                 HookCall::Ident(ident) => Some(ident),
                 HookCall::Member(Expr::Ident(ident), _) => Some(ident),
                 _ => None,
             };
-            if ident
-                .and_then(|id| self.scope_binding.get(&id.sym))
-                .is_none()
+            if !ident
+                .map(|id| current_mark.is_descendant_of(id.span.ctxt().outer()))
```
kdy1, line 132: How does this work?

crates/swc_ecma_transforms_react/src/refresh/tests.rs
```diff
@@ -1382,7 +1384,8 @@ test!(
                     emit_full_signatures: true,
                 }),
                 t.cm.clone(),
-                Some(t.comments.clone())
+                Some(t.comments.clone()),
+                unresolved_mark
```
kdy1, line 1388: Seems like an inconsistent mark?

kdy1 (changes requested):
Thanks! 
But I think this can error for files which has decorator and jsx.

decorator pass injects some variables and this pass will error in that case.

Austaras:
Could you give an example?

kdy1:
Can you try the code below?

```typescript
class Ns {

}
namespace Ns {
    export const FOO = 1;
    
}

const a = (a) => {
    const useE = useEffect;
    return function useFoo() {
      useE(() => console.log(a), []);
      return useState(123);
    };
}
```

It does not have a decorator, but `typescript::strip` also injects some variables.

Austaras:
It generates
```js
class Ns {
}
(function(Ns1) {
    var FOO = Ns1.FOO = 1;
})(Ns || (Ns = {}));
const a = (a1)=>{
    var _s = $RefreshSig$();
    const useE = useEffect;
    return _s(function useFoo() {
        _s();
        useE(()=>console.log(a1)
        , []);
        return useState(123);
    }, "useE{}\nuseState{(123)}", false, function() {
        return [
            useE
        ];
    });
};
```
which seems good enough. If I get what you mean, seems like I need to ignore injected variables?

kdy1:
Oh, you are right, my bad. I used playground to check if input injects `var`, but it's not done by `typescript::strip`.

Can you try

```ts


namespace Ns {
    export const FOO = 1;
    
}

const a = (a) => {
    const useE = useEffect;
    return function useFoo() {
      useE(() => console.log(a), []);
      return useState(123);
    };
}
```

this instead?

And yeah, those should be ignored, but we may attach syntax context to the scope node alternatively

Austaras:
> but we may attach syntax context to the scope node alternatively
This is definitely better

Austaras:
I've played with giving block syntax context but encountered other problems, maybe it could be done in another pr.

kdy1 (approved):
Thank you!


---

swc-bump:
 - swc_ecma_transforms_react

PR #4650, opened 2022-05-13
refactor(common): Change reserved `BytePos` to `u32::MAX`

kdy1:
**Description:**

---

cc @bartlomieju @dsherret 

Let's see which crate breaks


kdy1:
I decided not to do this as this causes some regression in source map.
I tried patching codes for sourcemap generation but I got no luck

dsherret:
@kdy1 does that indicate that some code is fragile because it is not explicitly relying on spans being dummy? Will that be fixed and this solution still explored?

kdy1:
Yeah, I think those codes should be updated to use `is_dummy` instead.
I may retry this in the future, but I'm not sure

kdy1:
Postponing as `SourceMap` is written under assumption `BytePos(0)` is dummy.
It compares `BytePos` using `cmp::max` or `cmp::min`

PR #4653, opened 2022-05-13
feat(dbg-swc): Add command to compare minifier

kdy1:
**Description:**

I want to make swc minifier emit smallest output, so I implemented a command to ensure it


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #4654, opened 2022-05-13
feat(dbg-swc): Add auto-reducer for the minifier

kdy1:
**Description:**

This command automatically reduces the input.


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #4658, opened 2022-05-13
feat(html/parser): support `iframe_srcdoc`

alexander-akait:
**Description:**

Support `iframe_srcdoc`, i.e.

https://www.w3schools.com/tags/att_iframe_srcdoc.asp

```
<iframe srcdoc="<p>Hello world!</p>" src="demo_iframe_srcdoc.htm"></iframe>
```

Before today I don't known about this features :confused: Currently we don't parse it by default for better perf and because it has very rare usage, but we add ability to do it because it is part of spec

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_codegen
 - swc_html_parser

alexander-akait:
@kdy1 Rebased :+1:

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser

PR #4663, opened 2022-05-14
fix(html/parser): improve error reporting

alexander-akait:
**Description:**

Better error reporting. Ideally we should avoid `UnexpectedToken` because this message is misleading. It is first part of refactor(I think I will finish it using 3 PRs) + many tests

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser

PR #4666, opened 2022-05-15
perf(es/parser): Optimize for cpu code cache

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_common
 - swc_ecma_parser

Issue #1202, opened 2020-11-04
import.meta.url does not get transpiled when using module type commonjs


TomasHubelbauer:
**Describe the bug**

I'm looking for a transpiler that will convert ESM import/export to CJS require and will also convert `import.meta.url` to something like `__filename` or similar.

With SWC config below I was able to transpile ESM import/export syntax to CJS. However, `import.meta.url` has stayed unchanged.

**Input code**

```js
#!/usr/bin/env node

import fs from 'fs';
import path from 'path';

void async function() {
  console.log(import.meta.url);
}()
```

**Config**
```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "dynamicImport": true,
      "importMeta": true
    },
    "target": "es2015"
  },
  "module": {
    "type": "commonjs",
    "strict": false,
    "strictMode": false,
    "lazy": false,
    "noInterop": true
  }
}
```

**Expected behavior**

Since `import.meta` is not valid in CJS I'd expect to see it transpiled. Not sure how exactly this should work but `import.meta.url` would make sense to me to be transpiled to `__filename` or `'file:///' + __filename`.

**Version**

```
  "dependencies": {
    "@swc/cli": "^0.1.27",
    "@swc/core": "^1.2.37"
  }
```

**Additional context**

`import.meta.url` remains intact in the output code.

kdy1:
Postponing as I couldn't decide the correct value.
There's no 'correct' value for this.

TomasHubelbauer:
@kdy1 what do you think about `"file:///" + __filename`? I wrote a patching script to convert ESM to CJS and I used this there, curious about possible issues or shortcomings.

dlindenkreuz:
@kdy1 You might want to take a look at https://github.com/javiertury/babel-plugin-transform-import-meta

From their docs:
> Transforms import.meta for nodejs environments. This plugin replaces any occurrence of import.meta.url.
> ```
> console.log(import.meta.url);
> ```
> And outputs code compatible with CommonJS
> ```
> console.log(require('url').pathToFileURL(__filename).toString());
> ```

PR #4670, opened 2022-05-15
fix(ecma-transform-module): transform import.meta.url correctly

djgrant:
**Description:**

Given the following config:

```js
{
  jsc: { parser: { importMeta: true } }, 
  compiler: { type: "commonjs" }
}
```

`import.meta.url` expressions are incorrectly transformed to `require("url").pathToFileURL(__filename).toString().url`.
 
This expression will always be evaluated to `undefined`.

**BREAKING CHANGE:**

`import.meta.url` is now transformed to:

```diff
- require("url").pathToFileURL(__filename).toString().url
+ require("url").pathToFileURL(__filename).toString()
```

#1202


kdy1 (approved):
Can you sign the cla?

PR #4682, opened 2022-05-16
feat(html/minifier): compress characters in `head`

alexander-akait:
**Description:**

After long investigation I found removing whitespaces from `<head>` (from first level, i.e. for element which parent is `<head>`) element is safe:
- We can have only whitespaces characters in `<head>` (i.e. '\x09' | '\x0A' | '\x0C' | '\x0D' | '\x20') 
- Any other text nodes with non whitespaces characters moved to body
- By default most of packages uses https://github.com/kangax/html-minifier or https://github.com/terser/html-minifier-terser/ (uses `terser` instead `uglify-js` for JS) and set http://perfectionkills.com/experimenting-with-html-minifier/#collapse_whitespace to `true` (i.e. collase multiple whitespaces to one) and it is safe in most of cases except cases described here http://perfectionkills.com/experimenting-with-html-minifier/#collapse_whitespace, but these cases related only to reprezentation elements (like `span`/`p`/`textarea`/etc)
- Also you can have only certain tags in `<head>` (`meta`/`link`/`script`/`style`) and they have no visible representation, another tags moved to body like text nodes (note - you can have `<template>` in `<head>`, but we keep whitespaces inside `<template>`)

So due these points above I don't see any unsafe things in removing whitespaces in `<head>` on first level

**BREAKING CHANGE:**

No

No


alexander-akait:
@kdy1 Rebased

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_minifier

kdy1 (approved):

PR #4683, opened 2022-05-16
test(html/codegen): broken html/svg

alexander-akait:
**Description:**

Check broken cases from https://github.com/kangax/html-minifier/issues/45 to ensure we don't need extra options for minifier, because we already parse and print it in recovery, so you will get valid HTML output

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - ''

PR #4685, opened 2022-05-16
perf(html): Setup benchmarks

alexander-akait:
**Description:**

Initial benches.

Should I update something in CI?

Also no lexer benches currently - added `TODO` for public API (implement iterator), because it can be useful.

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!
CI is already passing 👍 

---

swc-bump:
 - ''

Issue #4401, opened 2022-04-22
exported `let` variable not mutated when exported after mutating function


YellowKirby:
### Describe the bug

Given a module with 2 exports: a mutable `let` variable and a function that mutates that variable. If the variable is exported _after_ the function, then calling that function does not mutate the variable if the output module format is CommonJS/AMD/UMD. This does not seem consistent with the ESM output behavior.

Changing the order of the export so that the variable is exported before the function changes (fixes?) this behavior.

### Input code

```typescript
// Bug here? does NOT mutate
export const setFoo = () => {
  foo = "mutated!";
};
export let foo;
setFoo();

// -------------------------------

// Expected behavior: mutation!
export let bar;
export const setBar = () => {
  bar = "mutated!";
};
setBar();
```

### Output code

```js
"use strict";
Object.defineProperty(exports, "__esModule", {
    value: true
});
exports.setBar = exports.bar = exports.foo = exports.setFoo = void 0;
const setFoo = ()=>{
    foo = "mutated!"; // I think this should be `exports.foo = foo = "mutated!"`
};
exports.setFoo = setFoo;
let foo;
exports.foo = foo;
setFoo();
let bar;
exports.bar = bar;
const setBar = ()=>{
    exports.bar = bar = "mutated!";
};
exports.setBar = setBar;
setBar();
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

[...]

PR #4688, opened 2022-05-17
fix(es/modules): Fix re-assignment of the export variable in export default function

wre232114:
**Description:**

The swc_ecma_transforms_modules crate do not transform the following code as expected：

**Input**
```js
export var format;
export function defaultLocale(definition) {
  locale = formatLocale(definition);
  format = locale.format;
  return locale;
}
format = '123';
```

The expected generated code as below, which is the same as the output of babel:
**Expected Output**
```js
"use strict";
  value: true
});
exports.default = defaultLocale;
exports.format = void 0;
function defaultLocale(definition) {
    locale = formatLocale(definition);
    exports.format = format = locale.format;
    return locale;
}
var format;
exports.format = format;
exports.format = format = '123';
```

the re-assignment of `format` in defaultLocale should be transformed to `exports.format = format = local.format`. current output:

**Current Unexpected Output**
```js
"use strict";
  value: true
});
exports.default = defaultLocale;
exports.format = void 0;
function defaultLocale(definition) {
    locale = formatLocale(definition);
    format = locale.format;
    return locale;
}
var format;
exports.format = format;
exports.format = format = '123';
```

A test case is added for this case.

**BREAKING CHANGE:**

- Closes https://github.com/swc-project/swc/issues/4401


kdy1 (changes requested):
Can you run `UPDATE=1 cargo test` in `./crates/swc`?
You may need to run it multiple time.

wre232114:
@kdy1 updated

crates/swc_ecma_transforms_module/src/common_js.rs
```diff
@@ -213,6 +213,18 @@ impl Fold for CommonJs {
                     for ident in found {
                         exports.push(ident.sym.clone());
                     }
+
+                    if let Decl::Var(VarDecl { decls, .. }) = decl {
+                        for var_decl in decls {
+                            if let VarDeclarator {
```
kdy1, line 219: I think we should use `find_pat_ids` here instead.

kdy1: Can you add a test case for exporting something with an array pattern or an object pattern?

crates/swc_ecma_transforms_module/src/util.rs
```diff
@@ -161,6 +161,17 @@ pub struct Scope {
 }
 
 impl Scope {
+    /// insert exported bindings without duplication
+    pub(crate) fn insert_exported_bindings(&mut self, ident: &Ident) {
+        let bindings = self
+            .exported_bindings
+            .entry((ident.sym.clone(), ident.span.ctxt()))
+            .or_default();
+        if !bindings.contains(&(ident.sym.clone(), ident.span.ctxt())) {
```
kdy1, line 170: Please use `ident.to_id()` instead

kdy1: `(ident.sym.clone(), ident.span.ctxt())` is old version of `ident.to_id()`

crates/swc_ecma_transforms_module/src/util.rs
```diff
@@ -161,6 +161,17 @@ pub struct Scope {
...
+            bindings.push((ident.sym.clone(), ident.span.ctxt()));
```
kdy1, line 171: Same here

crates/swc_ecma_transforms_module/src/common_js.rs
wre232114: ok

crates/swc_ecma_transforms_module/src/util.rs
wre232114: ok

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_module

wre232114:
The `Performance regression check` failed due to Network issue. Could you rerun it, thank you

PR #4690, opened 2022-05-17
fix: don't use reserved start position in SourceMap::with_file_loader

dsherret:


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_common

PR #4693, opened 2022-05-18
test(es): Add a test for a fixed issue

### LINKED ISSUE #2728: spack: Minifier corrupts code that uses direct `eval`
### Describe the bug

Code that uses direct `eval` can access variables in any enclosing scope. This constraint isn't correctly handled by spack. There are two separate bugs:

* Names of variables in enclosing scopes are incorrectly minified when they should not be
* Variables in enclosing scopes are incorrectly removed when they could still be accessed

I'm using the following `spack.config.js` file:

```js
module.exports = {
  mode: 'production',
  options: {
    jsc: {
      minify: {
        compress: true,
        mangle: {
          topLevel: true,
        },
      },
    },
  },
  entry: {
    web: 'input.js',
  },
  output: {
    path: '.',
    name: 'output.js',
  },
}
```

### Input code

```typescript
function fn(bar) {
  const foo = 64206..toString(16)
  return eval(bar)
}
console.log(fn('foo + bar'))
```

### Config

_No response_

### Playground link

_No response_

### Expected behavior

That code should print the following:

```
facefoo + bar
```

However, when bundled with spack that code crashes instead:

```
$ node output.js
undefined:1
foo + bar
^

ReferenceError: foo is not defined
```

That's because spack generates the following code:

```js
function a1(a) {
    64206..toString(16);
    return eval(a);
}
console.log(a1("foo + bar"));
```

### Version

@swc/cli: 0.1.51, @swc/core: 1.2.108

### Additional context

_No response_

kdy1:
**Description:**

 - Closes #2728

### DIFF
--- a/crates/swc/tests/fixture/issues-2xxx/2728/input/.swcrc
+++ b/crates/swc/tests/fixture/issues-2xxx/2728/input/.swcrc
@@ -1,0 +1,10 @@
+{
+    "jsc": {
+        "minify": {
+            "compress": true,
+            "mangle": {
+                "topLevel": true
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


kdy1 (commented):
swc-bump:
 - ''

PR #4708, opened 2022-05-19
fix(es/config): Add `jsc.output.charset`

### LINKED ISSUE #4697: A standalone `jsc.transform.charset` configuration
### Describe the feature

Currently, swc will automatically escape any non-ASCII characters into Unicode sequences when (and only when) `target` was set to ES3 and ES5.

I want to force enable the escape of non-ASCII characters even the target engines support. Generally, it would reduce the after-gzipped size.

_No response_

### Additional context

_No response_

kdy1:
**Description:**

 - Closes #4697


kwonoj:
what if we make this option name as generic enough like other tools (`charSet`) but accepts two only (`ascii | utf8?`) so we can extend it later?

kwonoj:
Yes, probably slim chance. Utf16? Utf32? Not sure if someone ever want this 🤔

PR #4711, opened 2022-05-19
Don't add a newline after an arrow function expression (codegen)

kevinbarabash:
**Description:**

`emit_block_stmt_or_expr` was adding a newline after expressions if cfg.minify was false.  Modifying this function to never add a trailing new line improves the codegen output for arrow function expressions by avoiding unecessary newlines.

As an example, `(a, b) => a + b;` was being emitted as:
```
(a + b) => a + b
;
```

After this change output will look like `(a, b) => a + b;`.

**Test Plan:**

`cargo test -p swc_ecma_codegen --all-features`


kevinbarabash:
There are some test failures I need to fix.

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_codegen

kdy1 (approved):

PR #4713, opened 2022-05-19
fix(css/parser): Allow `var()` call in color functions

g-plane:
**Description:**

This PR fixes that using `var()` in color functions. For example:

```css
a {
  color: rgb(var(--a) var(--b) var(--c));
}
```


alexander-akait:
@g-plane yep, it is known bug, but I think this solution is not good too, because you can have:
```css
a {
  color: rgb(var(--a) var(--b));
}
```

And it is valid syntax too, ideally we should have `try_parse_variable_function<T: Fn(...) -> Type>(&mut self) { ... }`, so we will try `var`/`env` firstly, if it is not `function` token try to parser using our logic, also if last was `var`/`env` and it is `)` token - finish function parsing.

Note - `env` function is valid here too, currently browsers supported only `Dimension` env (i.e. screen spaces in `px`), but by spec it can be any value like `var`.

Also make sense to look how chrome/firefox/etc do in their source code

g-plane:
Where is the `try_parse_variable_function` should be used?

alexander-akait:
@g-plane On any argument in function

g-plane:
It sounds we need to add it in many places, not somewhere once. Am I right?

alexander-akait:
@g-plane Yep, but here only add it for color functions, and do it for another function late

g-plane:
How can I fix the failures in codegen and minifier?

kdy1:
`UPDATE=1 cargo test` in each directory

crates/swc_css_parser/tests/fixture/value/color/input.css
```diff
@@ -17,6 +17,9 @@
     color: rgb(48% 100% 100% / 50%);
     color: rgb(48% none 100% / 50%);
     color: rgb(48% 100% 100% / none);
+    color: rgb(var(--red), var(--green), var(--blue));
+    color: rgb(var(--red) var(--green) var(--blue));
+    color: rgb(var(--red) var(--green) var(--blue) / var(--alpha));
```
alexander-akait, line 22: Can we add `rgb(var(--foo))` and `rgb(var(--foo) var(--bar))`?

crates/swc_css_parser/tests/fixture/function/calc/span.rust-debug
```diff
@@ -11462,36 +11342,6 @@
...
- 88 | width: round(to-zero, var(--width), 50px);
-    :                       ^^^^^^^^^^^^
-    `----
-
-  x CalcValue
-    ,-[$DIR/tests/fixture/function/calc/input.css:88:5]
- 88 | width: round(to-zero, var(--width), 50px);
-    :                       ^^^^^^^^^^^^
-    `----
-
```
alexander-akait, line 11494: Strange change

alexander-akait (commented):
Looks good, let's add couple tests and check why calc tests are chaged

crates/swc_css_codegen/tests/fixture/packages/bootstrap_5_1_3/output.min.css
```diff
@@ -1 +1 @@
...
...r{border-radius:0}.modal-fullscreen-xl-down .modal-body{overflow-y:auto}.modal-fullscreen-xl-down .modal-footer{border-radius:0}}@media(max-width:1399.98px){.modal-fullscreen-xxl-down{width:100vw;max-width:none;height:100%;margin:0}.modal-fullscreen-xxl-down .modal-content{height:100%;border:0;border-radius:0}.modal-fullscreen-xxl-down .modal-header{border-radius:0}.modal-fullscreen-xxl-down .modal-body{overflow-y:auto}.modal-fullscreen-xxl-down .modal-footer{border-radius:0}}.tooltip{position:absolute;z-index:1080;display:block;margin:0;font-family:var(--bs-font-sans-serif);font-style:normal;font-weight:400;line-height:1.5;text-align:left;text-align:start;text-decoration:none;text-shadow:none;text-transform:none;letter-spacing:normal;word-break:normal;word-spacing:normal;white-space:normal;line-break:auto;font-size:.875rem;word-wrap:break-word;opacity:0}.tooltip.show{opacity:.9}.tooltip .tooltip-arrow{position:absolute;display:block;width:.8rem;height:.4rem}.tooltip .tooltip-arrow::before{position:absolute;content:"";border-color:transparent;border-style:solid}.bs-tooltip-top,.bs-tooltip-auto[data-popper-placement^=top]{padding:.4rem 0}.bs-tooltip-top .tooltip-arrow,.bs-tooltip-au
```
alexander-akait, line 1: There is regression, we don't need space between `*`

g-plane: I haven't changed anything about codegen, not sure why this is changed.

crates/swc_css_parser/tests/fixture/function/calc/span.rust-debug
g-plane: I think this change is similar to that in AST output: it isn't a `CalcSum` any more, instead, it's `Function` directly. Previously, it's a `CalcSum` that wraps a `Function`.

alexander-akait: It should 

> it's a CalcSum that wraps a Function

https://www.w3.org/TR/css-values-4/#calc-syntax 

That is why we have https://github.com/swc-project/swc/pull/4713#discussion_r878707487 problem

alexander-akait: Looks like we need to use `try_parse_variable` in calc functions too, regarding to syntax `var` is general syntax for any `calc-value`

crates/swc_css_codegen/tests/fixture/packages/bootstrap_5_1_3/output.min.css
kdy1: I think it means the AST generated by the parser is diffent from the previous version

crates/swc_css_parser/tests/fixture/function/calc/span.rust-debug
g-plane: It's already used in math functions: https://github.com/g-plane/swc/blob/67c4aa1efbe86aa440927a8661bf56ebfb7432ac/crates/swc_css_parser/src/parser/value/mod.rs#L267 .

kdy1 (commented):
Code style looks good.

cc @alexander-akait Do you think this change is correct?

---

swc-bump:
 - swc_css_parser

alexander-akait (commented):
There is regression `padding-right:calc(var(--bs-gutter-x) * 0.5);`, it should be `padding-right:calc(var(--bs-gutter-x)*.5);`, we need to fix AST for calc here, same for other changes for math functions

crates/swc_css_codegen/tests/fixture/packages/material-components-web_13_0_0/output.css
```diff
@@ -6919,19 +6919,31 @@ svg.mdc-button__icon {
...
 }
 @media (max-width: 599px) {
   .mdc-layout-grid--fixed-column-width {
     width: 368px;
-    width: calc(var(--mdc-layout-grid-column-width-phone, 72px) * 4 + var(--mdc-layout-grid-gutter-phone, 16px) * 3 + var(--mdc-layout-grid-margin-phone, 16px) * 2);
+    width: calc(
+                var(--mdc-layout-grid-column-width-phone, 72px) * 4 +
+                var(--mdc-layout-grid-gutter-phone, 16px) * 3 +
+                var(--mdc-layout-grid-margin-phone, 16px) * 2
+        );
```
alexander-akait, line 6946: This (and above) should not changed

alexander-akait (changes requested):
We still have problems with calc (math functions)

g-plane:
@alexander-akait Are there any tips about fixing it? I'm not clear about this.

alexander-akait:
@g-plane Let's don't use `try_parse_variable_function` in math functions like we use it for color functions and intead move it in `CalcValue`

https://github.com/swc-project/swc/blob/87a089b1571176a90546eb0482ee59910b331aa7/crates/swc_css_parser/src/parser/value/mod.rs#L2800

https://github.com/swc-project/swc/blob/87a089b1571176a90546eb0482ee59910b331aa7/crates/swc_css_parser/src/parser/value/mod.rs#L2836

I have ideas how to refactor it and avoid this complex logic for `var`/`env`, but it needs more time, so let's solve the current problem using my solutions above

g-plane:
Can we just fix for color functions in this PR? I mean, let's fix for math functions in the future.

alexander-akait (approved):
@kdy1 Looks good, no regression more

@g-plane Yep, let's improve it in other PRs, also I will look at souce good of chrome and mozilla how they parse variables, so we will have the same logic

kdy1 (approved):
Thank you!

kdy1 (approved):

PR #4719, opened 2022-05-20
fix(html/parser): unclosed td/th

alexander-akait:
**Description:**

Do not report on unclosed `td`/`th` when it is possible, it was typo when I did implementation

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!


---

swc-bump:
 - swc_html_parser

PR #4720, opened 2022-05-20
fix(html/parser): error reporting on `image` element

alexander-akait:
**Description:**

Bug fix, we don't need clone here, just rename tag, otherwise we will report on ack unclosed element

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser

PR #4727, opened 2022-05-20
feat(swc/transform_module): support local scoped amd require

### LINKED ISSUE #4718: AMD generated code misses local require-function
### Describe the bug

With AMD there is the global and local require-function ([more here](https://github.com/amdjs/amdjs-api/blob/master/require.md)). The latter knows the url of the current module and has a `toUrl`-util enabling modules to create urls of related resources. Think of this as the ancestor of `import.meta.url`. This doesn't seem to be supported with SWC because the `define` functions it generates don't spell out the "require" special import

### Input code

```typescript
function fff() {
	require.toUrl("");
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2020"
  },
  "module": {
    "type": "amd"
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.186&code=H4sIAAAAAAAAA0srzUsuyczPU0hLS9PQVKjm4ixKLSzNLErVK8kPLcrRUFLStOaqBQAX%2B2D7JgAAAA%3D%3D&config=H4sIAAAAAAAAA0WOMQ6DMAxF7%2BKZATFygw49hBUMSpXEkW2kRoi7N0GhbNbz%2F88%2B4KMO5gMyipK0SUsy%2FMIMVjKpE58NBjCtaMWgNMBCjgWNRWE22RspCaN3r5hZrNOzllA2sqoincZprJrArPQXRZ%2F8WtpRxzELqT4rTFu4k2d1RV72Bo7rr%2BrEuMD5OHrP67sHrx9%2B9tLJVOAAAAA%3D

### Expected behavior

The `define`-call's dependency-array should contain the special "require"-module (similar to the special "exports"-module) and also the callback should be define a `require` argument so that module use their local require-call, not the global one

```
define(["require"], function(require) {
    "use strict";
    function fff() {
        require.toUrl("");
    }
});
```

See TypeScript Playground: https://www.typescriptlang.org/play?jsx=0&module=2#code/JYWwDg9gTgLgBAbwIYF84DMoRHA5E3AKEPQFcA7AYxmAnI3XQAoBKRQgSCgFMBHU4DwB0MCAFUoAGyYAiGSwDchFEA

### Actual behavior

The define-call should depend on require, either always or by knowing if any dependency onto `require` exists

### Version

1.2.186

### Additional context

Thanks for this great project! Related to https://github.com/microsoft/vscode/issues/150025

kwonoj:
**Description:**
PR attempts to close https://github.com/swc-project/swc/issues/4718. Following logics to `exports`, PR attempts to inject local scoped `require` for the modules.

This PR mimics what tsc does (https://www.typescriptlang.org/play?jsx=0&module=2#code/KYDwDg9gTgLgBAMwK4DsDGMCWEVwEYQQAUAlHAN4C+QA) which shadows global cjs `require` entirely, instead of trying to apply private ident to existing `require` inside of module. This also reduces mutation to all of idents to callexpr, and we can short-curcuit visitor as soon as find first ident to callexpr.

**BREAKING CHANGE:**


kwonoj:
same as https://github.com/swc-project/swc/pull/4726, original PR was coming from incorrect branch.

crates/swc_ecma_transforms_module/src/amd.rs
```diff
@@ -100,6 +141,11 @@ impl Fold for Amd {
         let mut emitted_esmodule = false;
         let mut has_export = false;
         let exports_ident = self.exports.0.clone();
+        // We'll preserve local scoped `require` ident as amd's local require ident
+        // shadows global one
+        let scoped_local_require_ident = local_scoped_require_visitor
+            .require_ident
+            .unwrap_or(private_ident!("require"));
```
kdy1, line 148: This should be `unwrap_or_else`

kdy1 (changes requested):

crates/swc_ecma_transforms_module/src/amd.rs
kwonoj: Updated!

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_module

Issue #4724, opened 2022-05-20
Numeric separators are not transpiled in ECMAScript < 2021


webschik:
### Describe the bug

Numeric separators are not removed from the transpired code even if I set ES5 as a target.

### Input code

```typescript
const foo = 1_000;
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": false
}
```

### Playground link

https://play.swc.rs/?version=1.2.187&code=H4sIAAAAAAAAA0vOzysuUUjLz1ewVTCMNzAwsAYAekUtKRIAAAA%3D&config=H4sIAAAAAAAAA0WMTQrEIAxG75K1286id5hDBCctFv9IUhgR714tlu7C915ehUMsrBUyshCPS0pU%2FMMKZAOKZZcVTNf6tKEXagYUeScdiiyd%2BZSEJjUQXHRbGSWbQmYSeRHG3T9m66GQfucYKmjJdAc%2F0N7G%2FHPynaLySe0Cj9ke9LUAAAA%3D

### Expected behavior

The input code should be transpiled into 

```js
var foo = 1000;
```

### Actual behavior

Numeric separators stay unchanged, but the other syntax like `const` declaration gets traspiled:

```js
var foo = 1_000;
```

### Version

1.2.164+

### Additional context

It seems like a regression issue, version `1.2.163` gives a correct result.

_No response_

kdy1:
Seems like a regression related to spans...

PR #4732, opened 2022-05-21
feat(es/compat): transform numeric separator

magic-akari:
**Description:**

**BREAKING CHANGE:**

- #4724


kdy1:
I already fixed it and published a new version.

magic-akari:
> I already fixed it and published a new version.

I prefer to fix it in `es/compat` for a few reasons.

- leave it without change when target >= es2021
- minimal change. Just remove `_` and keep the original scientific notation, hexadecimal, etc.

kdy1:
Visitors are too costly for this case.
We can handle it in codegen, but not with visitor

kdy1:
Everything you mentioned it possible with codegen.
Visitor is really a waste

PR #4744, opened 2022-05-23
refactor(es/minifier): Make `rust-analyzer` faster

kdy1:
**Description:**

 - We now use a separate crate for testing the minifier. This makes IDE much faster.


kdy1 (commented):
swc-bump:
 - ''

Issue #4567, opened 2022-05-07
Base class initialized after class extending Base causing ReferenceError


smvv:
### Describe the bug

`Base` is initialized after the class is extending Base. This results in the javascript runtime error:
```
ReferenceError: Cannot access 'Base' before initialization
```

### Input code

```typescript
const Base = getBase('')

export default class Foo extends Base {}
```

### Config

```json
{
  "sourceMaps": true,
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true,
      "dynamicImport": true,
      "decorators": true
    },
    "preserveAllComments": true,
    "transform": null,
    "target": "es2017",
    "loose": false,
    "externalHelpers": false,
    "keepClassNames": false,
    "minify": null
  },
  "minify": true,
  "module": {
    "type": "commonjs"
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.178&code=H4sIAAAAAAAAA0vOzysuUXBKLE5VsFVITy0BsTTU1TW5uFIrCvKLShRSUtMSS3NKFJJzEouLFdzy8xVSK0pS81KKIZqqawHwwk9CQgAAAA%3D%3D&config=H4sIAAAAAAAAA0WPO27EMAxE78LaRTZNAHXBNkmR3IHQjgNvKFEg6cDGwnePjP115ADzHnkh19kyvrg5pbAZA509U7pQY3PYPvlagxdKFGuDZ5ta0EDhy71xWiuXKX%2BWphaPEFmNQ%2B0G3gZqho78w7vIUUtBjYc0jKuPaoVSnUV6wPaDDiP468vhrQtF1UFpZPFewBKwyvIBadgdt%2FwXaEdh928ueMZlqtO4XuHbc73Ki55mwf7q%2FmF35n6d1rPTtv0Do8IEIiIBAAA%3D

### Expected behavior

Typescript has Base initialized before the class extends it: [playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAQgQwgUxgXhgc2VRKAUA5IQJQBQZyAHgA4gBOsAJsgGYICuANrMF0hBgAxECBjUoyME0F5UAbwC+QA).
```js
const Base = getBase('');
export default class Foo extends Base {
}
```

### Actual behavior

The actual output is:
```js
"use strict";
Object.defineProperty(exports,"__esModule",{value:true});
exports.default=void 0;
class Foo extends Base{}
exports.default=Foo;
const Base=getBase("")
```

[...]

PR #4745, opened 2022-05-23
Split export statements in commonjs to fix hoisting

smvv:
Hi @kdy1 I would like to contribute to SWC and this PR could solve issue #3456.

However, I'm new to Rust (but have experience in JS compiler tech), and I would like to request early feedback regarding the PR.

For example, I would like to request feedback regarding:
1. the overall approach of splitting export declarations and moving those to the bottom.
2. what should we do with import statements? Can we freely move those to the top?
3. `todo()` statements left in the code.
4. I see new `_default` var declarations in the updated test files. I think this comes from `crates/swc_ecma_transforms_module/src/common_js.rs` and I've made a change to avoid the extra vars when `ModuleItem::ModuleDecl(ModuleDecl::ExportDefaultExpr(expr))` is used. Should I also look into the other var declarations, or can we postpone this to another issue/PR?
5. When should I use `DUMMY_SP` exactly as `span`?
 

**Description:**

This transformation will re-ordering the export declarations. The goal is to solve the ReferenceError problem when exports are referencing identifiers declared later in the code.

**BREAKING CHANGE:**

I would like to make the PR non-breaking, but it could be a breaking change as of now.

Note: I still want to remove all `dbg!()` and other commented code.

#4567


kdy1:
> 1. the overall approach of splitting export declarations and moving those to the bottom.

We can store it in a vector and append the content of the vector to the actual statements.

> 2. what should we do with import statements? Can we freely move those to the top?

Not sure. I don't know the specification.

> 3. `todo()` statements left in the code.

This is the exact thing bothering me these days. Technically it's a task of minifier, but we can't know if `exports.foo` in `exports.foo = void 0;` is side-effect-free.

I'm considering some new techniques, but I have to polish it and write a RFC.

> 4. I see new `_default` var declarations in the updated test files. I think this comes from `crates/swc_ecma_transforms_module/src/common_js.rs` and I've made a change to avoid the extra vars when `ModuleItem::ModuleDecl(ModuleDecl::ExportDefaultExpr(expr))` is used. Should I also look into the other var declarations, or can we postpone this to another issue/PR?

It can be postponed.

5. When should I use `DUMMY_SP` exactly as `span`?
 
If the compiler generate a span and in the case it does not come from user's source code, you should use `DUMMY_SP`.



Btw, thank you for the PR!

magic-akari:
@kdy1  I am working on the esbuild style module transformation.

kdy1:
@magic-akari Cool!
Thank you and feel free to open a draft PR if you want.

I typically do it to prevent duplicate works, but not opening is fine anyway

crates/swc_ecma_transforms_module/src/hoist.rs
```diff
@@ -19,10 +20,124 @@ impl CompilerPass for ModuleHoister {
...
+                            expr: Box::new(Expr::Ident(ident)),
+                        }));
+                    extra.push(export);
+                }
+
+                ModuleItem::ModuleDecl(ModuleDecl::ExportNamed(NamedExport {
+                    // TODO specifiers,
+                    src: None,
+                    type_only: false,
+                    asserts: None,
```
kdy1, line 117: I think this bound is not required?

crates/swc_ecma_transforms_module/src/hoist.rs
```diff
@@ -19,10 +20,124 @@ impl CompilerPass for ModuleHoister {
...
+                })) => {
+                    let expr = match decl.clone() {
+                        DefaultDecl::Class(expr) => Decl::Class(ClassDecl {
+                            class: expr.class,
+                            ident: expr.ident.unwrap_or(private_ident!("_default")),
+                            declare: false,
+                        }),
+                        DefaultDecl::Fn(expr) => Decl::Fn(FnDecl {
+                            function: expr.function,
+                            ident: expr.ident.unwrap_or(private_ident!("_default")),
```
kdy1, line 91: This invokes `private_ident` regardless of `expr.ident`.

kdy1: Please use `unwrap_or_else` instead

crates/swc_ecma_transforms_module/src/hoist.rs
```diff
@@ -19,10 +20,124 @@ impl CompilerPass for ModuleHoister {
...
+                //     need_hoist_up = true;
+                // }
+                ModuleItem::ModuleDecl(ModuleDecl::ExportDefaultDecl(ExportDefaultDecl {
+                    decl,
+                    ..
+                })) => {
+                    let expr = match decl.clone() {
+                        DefaultDecl::Class(expr) => Decl::Class(ClassDecl {
+                            class: expr.class,
+                            ident: expr.ident.unwrap_or(private_ident!("_default")),
```
kdy1, line 86: Same here

kdy1 (changes requested):
Thank you, and sorry for the late review

kdy1:
I think this can be closed, thanks to the new module transforms

PR #4748, opened 2022-05-23
fix(es/minifier): Consider parameters while detecting pure calls

kdy1:
**Description:**

**BREAKING CHANGE:**


kdy1 (commented):
swc-bump:
 - swc_ecma_utils

PR #4749, opened 2022-05-23
perf(common): Add more methods to `Spanned`

kdy1:
**Description:**

 - Lots of code only needs `lo` or `hi`, but they call `span()`, which is inefficient in some cases.
 - This PR only adds a trivial amount of optimization. I may add full optimization in future.
 - This PR improves codegen performance of `ExprOrSpread` and `Option<ExprOrSpread>`.


kdy1 (commented):
swc-bump:
 - swc_common
 - swc_ecma_codegen
 - swc_ecma_parser

PR #4754, opened 2022-05-23
fix(es/compat): use define for non-spread props

### LINKED ISSUE #4631: Bug: Object getter is statically evaluated when used with object spread
### Describe the bug

Object getter is statically evaluated when used with object spread

### Input code

```typescript
const b = {}
const a = {
	...b,
	get c() {
		console.log('get c')
	}
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.181&code=H4sIAAAAAAAAA0vOzysuUUhSsFWoruVKBnMSQRwuTj09vSQdLs701BKFZA1NkAgnSD4%2FJ1UvJz9dQx0soa7JxVnLVcsFALEMklZHAAAA&config=H4sIAAAAAAAAA0WMTQrEIAxG75K1286id5hDBCctFv9IUhgR714tlu7C915ehUMsrBUyshCPS0pU%2FMMKZAOKZZcVTNf6tKEXagYUeScdiiyd%2BZSEJjUQXHRbGSWbQmYSeRHG3T9m66GQfucYKmjJdAc%2F0N7G%2FHPynaLySe0Cj9ke9LUAAAA%3D

### Expected behavior

`a.c; a.c` should print `get c` twice

### Actual behavior

`a.c; a.c` should print nothing, `get c` is printed once statically.

See [babel playground](https://babeljs.io/repl/#?browsers=&build=&builtIns=entry&corejs=3.6&spec=true&loose=false&code_lz=MYewdgzgLgBARjAvDA3gXwFCkrAhk1DASADoy4AaYgcwFNZgAKASkKKOwhABtaTuQ1RgHI6DYc2KZMQA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=false&fileSize=false&timeTravel=true&sourceType=module&lineWrap=true&presets=env%2Ctypescript&prettier=true&targets=&version=7.17.11&externalPlugins=&assumptions=%7B%7D)

### Version

1.2.181

### Additional context

SWC output:

```javascript
    if (key in obj) {
            value: value,
            enumerable: true,
            configurable: true,
            writable: true
        });
    } else {
        obj[key] = value;
    }
    return obj;
}
function _objectSpread(target) {
    for(var i = 1; i < arguments.length; i++){
        var source = arguments[i] != null ? arguments[i] : {};
        var ownKeys = Object.keys(source);
        if (typeof Object.getOwnPropertySymbols === "function") {
            ownKeys = ownKeys.concat(Object.getOwnPropertySymbols(source).filter(function(sym) {
                return Object.getOwnPropertyDescriptor(source, sym).enumerable;
            }));
        }
        ownKeys.forEach(function(key) {
            _defineProperty(target, key, source[key]);
        });
    }
    return target;
}
var b = {};
var a = _objectSpread({}, b, {
    get c () {
        console.log("get c");
    }
});

```

Babel output:

```javascript
"use strict";

function ownKeys(object, enumerableOnly) {
  var keys = Object.keys(object);
  if (Object.getOwnPropertySymbols) {
    var symbols = Object.getOwnPropertySymbols(object);
    enumerableOnly &&
      (symbols = symbols.filter(function (sym) {
        return Object.getOwnPropertyDescriptor(object, sym).enumerable;
      })),
      keys.push.apply(keys, symbols);
  }
  return keys;
}

function _objectSpread(target) {
  for (var i = 1; i < arguments.length; i++) {
    var source = null != arguments[i] ? arguments[i] : {};
    i % 2
      ? ownKeys(Object(source), !0).forEach(function (key) {
          _defineProperty(target, key, source[key]);
        })
      : Object.getOwnPropertyDescriptors
      ? Object.defineProperties(
          target,
          Object.getOwnPropertyDescriptors(source)
        )
      : ownKeys(Object(source)).forEach(function (key) {
          Object.defineProperty(
            target,
            key,
            Object.getOwnPropertyDescriptor(source, key)
          );
        });
  }
  return target;
}

  if (key in obj) {
      value: value,
      enumerable: true,
      configurable: true,
      writable: true
    });
  } else {
    obj[key] = value;
  }
  return obj;
}

var b = {};

var a = _objectSpread(
  _objectSpread({}, b),
  {},
  {
    get c() {
      console.log("get c");
    }
  }
);

```

Austaras:
**Description:**

So, for following code
```js
({...a, b , ...c, c})
```

At first, I tried to copy babel's behavior, which would be
```js
_objectSpread(
  _objectSpread(
    _objectSpread({}, a),
    {},
    {
      b
    },
    c
  ),
  {},
  {
    c
  }
);
``` 

which seems a bit redundant, but it's necessary in babel's case: babel's helper `_objectSpread2`(swc copied `_objectSpread`) would only apply `getOwnPropertyDescriptors` for even targets, so it need an extra empty object for every non spread props, which doesn't seem to be optimal.

So I checkout esbuild and it would do
```js
var __spreadProps = (a, b) => __defProps(a, __getOwnPropDescs(b));

// project/index.ts
__spreadProps(__spreadValues(__spreadValues(__spreadProps(__spreadValues({}, a), { b }), c), d), { c });
```
which is more concise, but doesn't support loose.

So I took the third way, which introduces a new helper `objectSpreadProps` which is alike to `__spreadProps` and let `_objectSpread` remain the same.

This would result in an extra helper, but could reduce generated size and improve performance. I don't if it's the right thing to do.

**BREAKING CHANGE:**

closes #4631

### DIFF
--- a/packages/swc-helpers/src/_object_spread_props.js
+++ b/packages/swc-helpers/src/_object_spread_props.js
@@ -1,0 +1,30 @@
+function ownKeys(object, enumerableOnly) {
+  var keys = Object.keys(object);
+  if (Object.getOwnPropertySymbols) {
+    var symbols = Object.getOwnPropertySymbols(object);
+    if (enumerableOnly) {
+      symbols = symbols.filter(function (sym) {
+        return Object.getOwnPropertyDescriptor(object, sym).enumerable;
+      });
+    }
+    keys.push.apply(keys, symbols);
+  }
+  return keys;
+}
+
+export default function _objectSpreadProps(target, source) {
+  source = source != null ? source : {}
+  if (Object.getOwnPropertyDescriptors) {
+    Object.defineProperties(target, Object.getOwnPropertyDescriptors(source));
+  } else {
+    ownKeys(Object(source)).forEach(function (key) {
+      Object.defineProperty(
+        target,
+        key,
+        Object.getOwnPropertyDescriptor(source, key)
+      );
+    });
+  }
+
+  return target;
+}
--- a/crates/swc/src/builder.rs
+++ b/crates/swc/src/builder.rs
@@ -232,7 +232,8 @@ impl<'a, 'b, P: swc_ecma_visit::Fold> PassBuilder<'a, 'b, P> {
                     compat::es2018(compat::es2018::Config {
                         object_rest_spread: compat::es2018::object_rest_spread::Config {
                             no_symbol: assumptions.object_rest_no_symbols,
-                            set_property: assumptions.set_spread_properties
+                            set_property: assumptions.set_spread_properties,
+                            pure_getters: assumptions.pure_getters
                         }
                     }),
                     should_enable(self.target, EsVersion::Es2018)
--- a/crates/swc/tests/exec.rs
+++ b/crates/swc/tests/exec.rs
@@ -74,52 +74,49 @@ fn create_matrix(entry: &Path) -> Vec<Options> {
     })
     .matrix_bool()
     .matrix_bool()
-    .matrix_bool()
     .into_iter()
-    .map(
-        |((((target, syntax), minify), source_map), external_helpers)| {
-            // Actual
-            Options {
-                config: Config {
-                    jsc: JscConfig {
-                        syntax: Some(syntax),
-                        transform: None.into(),
-                        external_helpers: external_helpers.into(),
-                        target: Some(target),
-                        minify: if minify {
-                            Some(JsMinifyOptions {
-                                compress: BoolOrDataConfig::from_bool(true),
-                                mangle: BoolOrDataConfig::from_bool(true),
-                                format: Default::default(),
-                                ecma: Default::default(),
-                                keep_classnames: Default::default(),
-                                keep_fnames: Default::default(),
-                                module: Default::default(),
-                                safari10: Default::default(),
-                                toplevel: Default::default(),
-                                source_map: Default::default(),
-                                output_path: Default::default(),
-                                inline_sources_content: Default::default(),
-                                emit_source_map_columns: Default::default(),
-                            })
-                        } else {
-                            None
-                        },
-                        ..Default::default()
+    .map(|(((target, syntax), minify), source_map)| {
+        // Actual
+        Options {
+            config: Config {
+                jsc: JscConfig {
+                    syntax: Some(syntax),
+                    transform: None.into(),
+                    external_helpers: false.into(),
+                    target: Some(target),
+                    minify: if minify {
+                        Some(JsMinifyOptions {
+                            compress: BoolOrDataConfig::from_bool(true),
+                            mangle: BoolOrDataConfig::from_bool(true),
+                            format: Default::default(),
+                            ecma: Default::default(),
+                            keep_classnames: Default::default(),
+                            keep_fnames: Default::default(),
+                            module: Default::default(),
+                            safari10: Default::default(),
+                            toplevel: Default::default(),
+                            source_map: Default::default(),
+                            output_path: Default::default(),
+                            inline_sources_content: Default::default(),
+                            emit_source_map_columns: Default::default(),
+                        })
+                    } else {
+                        None
                     },
-                    module: Some(ModuleConfig::CommonJs(Default::default())),
-                    minify: minify.into(),
                     ..Default::default()
                 },
-                source_maps: if source_map {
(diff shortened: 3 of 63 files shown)


kdy1:
@kwonoj I think we should double-check this

Austaras:
Yes, I'm also not quite sure about the correctness of this pr...

kwonoj:
Initial glimpse looks ok, but not sure what test can ensure this other than we currently have in here.

kdy1:
I'll try adding lots of execution tests for object rest/spread.
Let's see if it works...

kdy1:
I updated the base branch as https://github.com/swc-project/swc/pull/4788 is merged

kdy1:
Seems like we need `yarn link` for helpers...

Austaras:
Or just don't `external_heleprs`

kdy1:
Oh, I think it would be better.

Can you patch https://github.com/swc-project/swc/blob/bebdb8a8974d14e76f921a140fa04d49cb1ba438/crates/swc/tests/exec.rs#L80?

Austaras:
Done.

packages/swc-helpers/src/_object_spread_props.js
```diff
@@ -0,0 +1,30 @@
+function ownKeys(object, enumerableOnly) {
+  var keys = Object.keys(object);
+  if (Object.getOwnPropertySymbols) {
```
kdy1, line 3: https://caniuse.com/mdn-javascript_builtins_object_getownpropertysymbols

Hmm, it's not supported by ie11

packages/swc-helpers/src/_object_spread_props.js
```diff
@@ -0,0 +1,30 @@
...
+      });
+    }
+    keys.push.apply(keys, symbols);
+  }
+  return keys;
+}
+
+export default function _objectSpreadProps(target, source) {
+  source = source != null ? source : {}
+  if (Object.getOwnPropertyDescriptors) {
```
kdy1, line 17: But well, this is supported by ie 11 so it's fine.

kdy1: https://caniuse.com/?search=getOwnPropertyDescriptor

packages/swc-helpers/src/_object_spread_props.js
```diff
@@ -0,0 +1,30 @@
...
+    Object.defineProperties(target, Object.getOwnPropertyDescriptors(source));
+  } else {
+    ownKeys(Object(source)).forEach(function (key) {
```
kdy1, line 20: I'm not sure if a browser without `Object.getOwnPropertyDescriptors` will have `Array.forEach`

packages/swc-helpers/src/_object_spread_props.js
kdy1: Do we really need this check?

kdy1 (commented):
I'm simply not sure.
Btw, CI is passing, great work!

packages/swc-helpers/src/_object_spread_props.js
Austaras: It's `getOwnPropertyDescriptors` and IE 11 doesn't support that.

packages/swc-helpers/src/_object_spread_props.js
Austaras: It's for getting object symbol keys, and IE11 doesn't have symbol to begin with

packages/swc-helpers/src/_object_spread_props.js
kdy1: Ah you are right, thank you!

packages/swc-helpers/src/_object_spread_props.js
Austaras: Oh, that would be a problem

Austaras: Oh, IE support `Array#forEach` starting with 9

kdy1:
Looks good, but I'll try some more verification

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_compat --breaking

Issue #4456, opened 2022-04-27
Import/Export order of cjs


danieloprado:
### Describe the bug

Is not possible to export destructuring in nextjs with swc, it throw that reference is undefined.

### Input code

```typescript
import '.';

import emotionStyled from '@emotion/styled';

import { breakpoinstUtils } from '@eduzz/houston-tokens/variables/breakpoints';

export { keyframes, cx, cx as clsx } from '@emotion/css';

export interface IStyledProp {
  className?: string;
}

export const breakpoints = breakpoinstUtils;

const styled = emotionStyled;
export default styled;
```

### Config

```json
{
  "test": ".*\\.tsx?$",
  "exclude": [".*.js$", ".*.map$", "node_modules", "build"],
  "module": {
    "type": "commonjs"
  },
  "jsc": {
    "externalHelpers": true,
    "target": "es5",
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "transform": {
      "react": {
        "runtime": "automatic"
      }
    },
    "keepClassNames": true
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.172&code=H4sIAAAAAAAAA0utKMgvKlGoVkhUqFVIK8rPVVAvSS0uSVW3BgCgr0PxGgAAAA%3D%3D&config=H4sIAAAAAAAAA0WMSwrDMAxE76K1t93kDj2EcJXg4B8aBWqM714bUrKSmPdmOp3wtHWqrBBdH1o2%2FtJG4hPDa6hGbmoz2jlChiNjPcSWgtdksRTITR2lkMPe1pIvqaoAD%2BJ8xL855lAqn2sFnazVeVcllXyCxjN0lwPet216yfgBCKaI9boAAAA%3D

### Expected behavior

The require imports should be before the exports. 

### Actual behavior

TypeError: undefined is not an object (evaluating '_css.keyframes')

### Version

1.2.172

### Additional context

"@swc/cli": "0.1.57"
"@swc/core": "1.2.172"
"@swc/helpers": "0.3.8"

magic-akari:
Cannot reproduce this issue.
Could you provide a minimal sample code?

kdy1:
Closing as more information required.
If this is still an issue, please open a new issue with required forms filled

PR #4758, opened 2022-05-23
refactor(es/module): new module transformer

### LINKED ISSUE #1135: Broken codegen of class with decorated method when using `jsc.module.lazy`
**Describe the bug**

Edit: I've used `module` as an example of an external module below, but I've now realized that it might be confused with the `module` Node builtin. The issue exists no matter what name the external module has.

Codegen breaks when all of the conditions below are met:

- `jsc.module.type` has to be set to `commonjs`.
- `jsc.module.lazy` has to be set to `true`.
- A base class has to be imported from an external module.
- Another class has to extend the base class.
- The subclass has to have a decorated method, decorated with a decorator that is a static property on the base class.

**Input code**

```ts
import { Class } from 'module';

class MyClass extends Class {
  @Class.Decorator()
  async method() {}
}
```

**Config**
```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "decorators": true
    },
    "target": "es2020",
    "transform": {
      "legacyDecorator": true
    }
  },
  "module": {
    "type": "commonjs",
    "lazy": true
  }
}
```

**Emitted code**

```js
"use strict";
var _module = require("module");
function _applyDecoratedDescriptor(target, property, decorators, descriptor, context) {
    var desc = {
    };
    Object.keys(descriptor).forEach(function(key) {
        desc[key] = descriptor[key];
    });
    desc.enumerable = !!desc.enumerable;
    desc.configurable = !!desc.configurable;
    if ("value" in desc || desc.initializer) {
        desc.writable = true;
    }
    desc = decorators.slice().reverse().reduce(function(desc, decorator) {
        return decorator(target, property, desc) || desc;
    }, desc);
    if (context && desc.initializer !== void 0) {
        desc.value = desc.initializer ? desc.initializer.call(context) : void 0;
        desc.initializer = undefined;
    }
    if (desc.initializer === void 0) {
        Object.defineProperty(target, property, desc);
        desc = null;
    }
    return desc;
}
var _class, _dec;
let MyClass = ((_class = function() {
    class MyClass extends _module().Class {
        async method() {
        }
    }
    return MyClass;
}()) || _class, _dec = _module.Class.Decorator(), _applyDecoratedDescriptor(_class.prototype, "method", [
    _dec
], Object.getOwnPropertyDescriptor(_class.prototype, "method"), _class.prototype), _class);



```

**Expected behavior**

The emitted code includes `class MyClass extends _module().Class`. Note the `_module()` call, which shouldn't happen, as `_module` is assigned the required module which isn't callable. This leads to a runtime error (`TypeError: _module is not a function`).

**Version**
The version of @swc/core: `1.2.35`

**Additional context**

This issue doesn't exist when `jsc.module.lazy` is set to `false`.

### LINKED ISSUE #4063: target 'amd' module with 'decorate' and 'class'  seems output  wrong code.
### Describe the bug

tscode with `decorators` and `class`

when target module is 'amd'

it's seems output the wrong order code lines:

```js
 var _class, _descriptor;
    _exports.default = _class1; 
    function state() {}
    class Controller {
    }
    let _class1 = ((_class = class extends Controller {
        onChange() {}
        constructor(...args){
            super(...args);
```

```ts
     _exports.default = _class1;
                       ^

ReferenceError: Cannot access '_class1' before initialization
    at /Users/bencode/work/swc-test/out.js:46:24
    at define (/Users/bencode/work/swc-test/out.js:2:3)
    at Object.<anonymous> (/Users/bencode/work/swc-test/out.js:5:1)

```

### Input code

```typescript
function state() {

}

class Controller {

}

export default class extends Controller {
  @state
  isTest = false

  onChange() {

  }
}
```

### Config

```json
{
  "module": {
    "type": "amd"
  },
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "decorators": true
    },
    "target": "es2016"
  }
}
```

### Playground link

_No response_

### Expected behavior

the exports

### Actual behavior

_No response_

### Version

1.2.156

### Additional context

_No response_

### LINKED ISSUE #4313: Helper `_interopRequireDefault` does not handle `export default`
### Describe the bug

I have files ts with `export default` like this one:
```ts
export default {
  foo: 'bar'
};
```

And its compiled into this:
```js
"use strict";
exports.default = void 0;
var _default = {
    foo: 'bar'
};
exports.default = _default;
```
This leads to a bug in files where is this file imported. As you can see in playground example. `_config.default.cwd` is output from swc compiler. In runtime this ends with error, where cwd prop will be undefined, becouse imported is actualy `_config.default.default.cwd`.

### Input code

_No response_

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "decorators": false
    },
    "target": "es2021"
  },
  "module": {
    "type": "commonjs",
    "strict": true,
    "noInterop": false
  },
  "sourceMaps": false,
  "inlineSourcesContent": false
}
```

### Playground link

https://play.swc.rs/?version=1.2.165&code=H4sIAAAAAAAAAy2OMQvCMBSE9%2FyKt6WFkq7S4urgYIeCe0hearRNQt6zIuJ%2FN1q3%2B%2B7g7vySYmYwMTg%2FgctxAalUu7HshSiKGF4CIOgFO2DPMzYFV8zkY%2BggaXPTE543%2FkYWyWSf%2BIdv2MNxHE4q6UxYOVIZtT34GcdnMFXSfCkOxXnFaptV5mGbcqP9N6srxSDrYt3Z7WRd9%2BIDwiE8s7cAAAA%3D&config=H4sIAAAAAAAAA0WNOw7CQAxE7%2BI6BaRMS0VBxQlWG4M2ytorjyMRRdwdB%2FHprDczzxtNyDRs1JKBbb%2BwiqcHDeRrY2QrzakjR6BbmsEdjZzVkqvhg56RJ7uzx4rRH%2FojBao6LjPvyt0UUdZaVSaEDm4lR91tCaHoWZxN298HXSzzJTX83haZi%2FD1HeCksRD%2FDl6uvNGOyAAAAA%3D%3D

### Expected behavior

Wen js file have `exports.default` swc should not use `_interopRequireDefault`. Or `_interopRequireDefault` should return `obj.default` if default prop exists.

```js
function _interopRequireDefault(obj) {
    return obj && (obj.__esModule || obj.hasOwnProperty('default')) ? obj : {
        default: obj
    };
}
```

### Actual behavior

_No response_

### Version

1.2.161

### Additional context

_No response_

### LINKED ISSUE #4534: Transpilation of import * as var is undefined when importing same file twice
### Describe the bug

When Importing the same file twice with import * under a different variable name, the first import is no longer defined.
The JS code throws with the following error:

```
exports.A = A;
            ^

ReferenceError: A is not defined
```

### Input code

```typescript
import * as A from "./A";
import * as B from "./A";;

export {
  A, B
};
```

### Config

_No response_

### Playground link

https://play.swc.rs/?version=1.2.177&code=H4sIAAAAAAAAA8vMLcgvKlHQUkgsVnBUSCvKz1VQ0tN3VLLmykSScUKWsebiSq0Ay1VzKSg46ig4cdVacwEABM1SQkkAAAA%3D&config=H4sIAAAAAAAAA0WNSwrDMAxE76J1tt3kDj2ESZXg4B8aBWqM7165OGQnzedNoxMbrY2KE7CMCzWp%2B9JKWgtjE1%2BUFlKYtLsA7vY4OVgtwniZF3IGT3eh6JPf6yBtORZh4LFcOsKd7AaK%2BXMNof3HDGiVmNMJ6g9olj3eM61ycf8BnwbyVboAAAA%3D

### Expected behavior

Both A and B being exported

### Actual behavior

The JS code throws with the following error:

```
exports.A = A;
            ^

ReferenceError: A is not defined
```

### Version

Latest

### Additional context

_No response_

### LINKED ISSUE #4567: Base class initialized after class extending Base causing ReferenceError

### LINKED ISSUE #4866: Wrong completion value when destructing on exported value
### Describe the bug

The completion value for destructing assign on exported bindings are wrong.

### Input code

```typescript
export let a, b
console.log([a, b] = [1, 2])
```

### Config

```json
This translation is wrong on all targets: including CommonJS, AMD, SystemJS, and UMD.
```

### Playground link

https://play.swc.rs/?version=1.2.196&code=H4sIAAAAAAAAA0utKMgvKlHISS1RSNRRSOLlSs7PK87PSdXLyU%2FXiAYJxSrYKkQb6igYxWoCAMtZwJQtAAAA&config=H4sIAAAAAAAAA0WMSwoEIQxE75K1i8Fl32EOEex04%2BCPJMKIePdRcOhViqqX1%2BEjDo4OBVmIV5KWFL9wALmI4tgXBTOxWV0YhIYBRb5JFyL2Ze2cQ85CGzAQffJXWzKXY2ESeSZMd%2FiTY7piPusqOmgr80KNJ4zHsf%2B8vDeoXGn8AN7HW6K4AAAA

### Expected behavior

The completion value should be `[1, 2]`

image

### Actual behavior

The completion value is `2`

### Version

1.2.196

### Additional context

The correct translation should be

```js
console.log(([
    [a, b] = [1, 2], // store completion value here
    exports.a = a,
    exports.b = b
][0])); // read it out by [0] here
```

### LINKED ISSUE #4981: assignment lost when transform `export var a = 1, b = 2`
### Describe the bug

just see code below. all assign statement lost after swc transform

### Input code

```typescript
export var a = 1, b = 2;

var c = 3;
export { c }
```

### Config

_No response_

### Playground link

https://play.swc.rs/?version=1.2.203&code=H4sIAAAAAAAAA0utKMgvKlEoSyxSSFSwVTDUUUgCUkbWXFwgoWQg29iaKxWiqBrIrwUAZkJC1TEAAAA%3D&config=H4sIAAAAAAAAA02OQQ6DMAwE%2F%2BIzh7ZSL%2Fyhj4iCaYPiJLJNVYr4ex2IRE%2B2vLPeXWESD%2F0KxbEg102WpO4DPaAnJ55DUegMs9PoouDWgTp%2BolZEbpfr3eSYs2ADOqCQwrjUZz5TYRQ5JZeeEf9esUsyZqZKMzqvxzKa6wW98mymAd8YcyFMepzMmIsGCt%2FWOVCJe%2BSubqZTHuYatIIuxWatQjlNAttZsJUK8mj0bv8BHPD9kxUBAAA%3D

### Expected behavior

```js
"use strict";

  value: true
});
exports.b = exports.a = void 0;
var c = 2;
var a = 1,
    b = c;
exports.b = b;
exports.a = a;
```

assignment preserved

### Actual behavior

```js
"use strict";
    value: true
});
exports.c = exports.b = exports.a = void 0;
var a, b;
exports.a = a;
exports.b = b;
var c;
exports.c = c;
```

all asignment lost

### Version

1.2.203

### Additional context

_No response_

### LINKED ISSUE #4984: Use before define in default export class static members
### Describe the bug

Hey there, it looks like swc is reordering `export default class` declarations in certain target configs, which places it before `const` declarations that it depends on in its class body, leading to use-before-define errors.

This shows up only in certain target/env configurations. I've provided a config with env targets `{ node: "14" }`, but this also shows up with a standard target: `target: "es2022"`.

### Input code

```typescript
const VALUE = 'hello';

export default class MyClass {
    static MEMBER = VALUE;
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "target": "es2017",
    "loose": true,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": true,
  "sourceMaps": "inline",
  "env": {
    "targets": {
      "node": "14"
    }
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.203&code=H4sIAAAAAAAAA0vOzysuUQhz9Al1VbBVUM9IzcnJV7fm4kqtKMgvKlFISU1LLM0pUUjOSSwuVvCtdAbT1VwKQFBckliSmazg6%2Brr5BoE1Aw2xJqrFgBJIWBsUwAAAA%3D%3D&config=H4sIAAAAAAAAA0WOUQqEMAxE75JvP9ZlYcE7eIhSo1TapCR1WZHefVMV9m%2BSybzJAat6GA7IThSlKd2puC8MUPaM6iXkAh0UtVWRDatpJwsWu0B9Pvq3uZFZ8fI7SIHCvDeU55QFVWGYXdRmOVoi3mM1VOJpa4vjbDOkRRLTqlD%2FoDscdLyvrx7lTTyOLhsfAsVAaK8gfU7c%2BaM2STw1cv%2BCWusPPNJy0vEAAAA%3D

### Expected behavior

I expect the declaration of `const VALUE` to appear before the declaration of `class MyClass` (or assignment of its static `MEMBER`)

### Actual behavior

The generated output is incorrectly:

```
"use strict";
    value: true
});
exports.default = void 0;
class MyClass {
    static MEMBER = VALUE;
}
exports.default = MyClass;
const VALUE = 'hello';
```

Where `VALUE` is declared after it is used

### Version

1.2.203

### Additional context

_No response_

### LINKED ISSUE #4988: Export import missing in list of exported symbols
### Describe the bug

The export import construct does always get exported, see sample

### Input code

```typescript
import * as testUtils from 'vs/base/test/common/testUtils';

export import flakySuite = testUtils.flakySuite;
```

### Config

```json
{
  "$schema": "http://json.schemastore.org/swcrc",
  "exclude": "\\.js$",
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "decorators": true
    },
    "target": "es2020",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "amd",
    "noInterop": true
  },
  "minify": false
}
```

### Playground link

https://play.swc.rs/?version=1.2.203&code=H4sIAAAAAAAAA%2BPKzC3ILypR0FJILFYoSS0uCS3JzClWSCvKz1VQLyvWT0osTtUHiesn5%2Bfm5ufpw9WoW3NxpVaANUPNSMtJzK4MLs0sSVWwRZilhxC2BgAQsceubgAAAA%3D%3D&config=H4sIAAAAAAAAA0WOTQ6DQAhG70JcNmpceoPewc1kxL84wwQw1RjvXqzR7sjj%2Bx7skIkfMDioYVBNdVFMQjG%2FoCgx5sR9IR%2FPHl6Aq5%2BXFi3dNPkkmaFJPNQ7JMeCfE6yRXWrRXRLKJ7HpBZTMdS5WfAFLXpiZ3KBWnnBw9aOe1QroVRlVVphJhJ8KmGMY7edek8hMYr8Vy728508zBWoXU6w%2Fz4wpwutCSO9oyJTeo7e0qv5BTnH4YgMAQAA

### Expected behavior

The `flakySuite` symbol should be exported

### Actual behavior

Nothing gets exported

### Version

v1.2.203

### Additional context

related to https://github.com/microsoft/vscode/issues/150025

### LINKED ISSUE #5012: ReferenceError: Cannot access 'CommentContentEditRecord' before initialization
### Describe the bug

This is still an issue:

```TS
import type { ObjectId } from "bson";
import { Record } from "immutable";
import { DefaultDate, DefaultObjectId } from "../utils/immutable/defaultValues";

type TCommentContentEditRecord = {
  editorId: ObjectId;
  bodyAsPlainText: string;
  editedAt: Date;
};

const CommentContentEditRecord = Record<TCommentContentEditRecord>({
  editorId: DefaultObjectId,
  bodyAsPlainText: "",
  editedAt: DefaultDate,
});

export default class CommentContentEdit extends CommentContentEditRecord {}
```

Produces:

```TS
"use strict";
    value: true
});
exports.default = void 0;
var _immutable = require("immutable");
var _defaultValues = require("../utils/immutable/defaultValues");
class CommentContentEdit extends CommentContentEditRecord {
}
exports.default = CommentContentEdit;
const CommentContentEditRecord = (0, _immutable).Record({
    editorId: _defaultValues.DefaultObjectId,
    bodyAsPlainText: "",
    editedAt: _defaultValues.DefaultDate
});
```

I'm getting:
```
ReferenceError: Cannot access 'CommentContentEditRecord' before initialization
```

Because the `class` goes defined before the Record.

### Input code

_No response_

### Config

_No response_

### Playground link

https://play.swc.rs/?version=1.2.204&code=H4sIAAAAAAAAA32QT2%2FCMAzF75X6HayemFTRO2WTEOzAaQgh7mljpqAkRokrgap%2Bd9I%2FsK3rdkry7Dz%2FnpW5kGPg2wWhho%2FijCVvJTRwcmQgKTzZJI8j1bfVsMeS3FddGVOxKDT%2BaNrgSVSaN4IxfTx%2BWc%2FnWcVK%2B%2Bzpkcm%2B9Sh0hb51jKMO7LAmY9DymiyH410qHjheoY4jAAwKua1cPBPkrVyQvK38TgtlD3jlBXh2yn7mjy8oV0FsMYPUdPNKsp7hn3n9Zfkn0ttsRDTKn06CJUk6gvq2wsD20sHhtVvwsCYotfB%2BghWCJVo5VRpi1M0dwPANiPgBAAA%3D&config=H4sIAAAAAAAAA0WOQQ6DMAwE%2F%2BIzh4pj%2FtBHRMGgIBJHXiMVIf5eg2i5Wbs7I%2B80I1HYqUUF63lhqxY%2FFMi2xkiam1FHBo%2FGuIA7GjiJRhMFBdOVD6%2BjTmwOMfpX3zuwiID%2FSMk1j9upT1KaMvBUsU7Lb3m4q8iwnsF%2BfeBOR4rUGXQ8ohvOeN%2Fr65Evi%2FIVsc8AAAA%3D

### Expected behavior

That the const definition goes before the class definition.

### Actual behavior

_No response_

### Version

^1.2.204

### Additional context

_No response_

### LINKED ISSUE #5042: `export import assign` not handled
### Describe the bug

The TypeScript export import assign is not transformed.

### Input code

```typescript
export import foo = require("foo");

foo();
```

### Config

```json
{
  "minify": false,
  "jsc": {
    "parser": {
      "syntax": "typescript"
    },
    "target": "es2022",
    "loose": false,
    "externalHelpers": true
  },
  "isModule": true,
  "module": {
    "type": "commonjs"
  }
}
```

### Playground link

[playground](https://play.swc.rs/?version=1.2.205&code=H4sIAAAAAAAAA1NQUFBIrSjILypRyMwFU2n5%2BQq2CkWphaWZRakaSkCukqY1FxdQHUhKQ9MaANaEsRwzAAAA&config=H4sIAAAAAAAAAz2MSwoDIRAF7%2FLWLoJLT5BNDiGmJzj4o7sHRsS7x0xCdu8VVA3kWOLW4TafhAx2CXADzbMQf5b0ov6Eg%2FZGEjg2xTRQzy%2FShUnszVoYpFqF%2Fh06lbj4dKfUiAVO%2BaDlRXnU55HoCwzy742rv3qh5lzLLpjzDUGEVr2dAAAA)

### Expected behavior

```JavaScript
const foo = require("foo");
exports.foo = foo;
foo();
```

### Actual behavior

```JavaScript
"use strict";
export import foo = require("foo")
foo();
```

### Version

1.2.205

### Additional context

_No response_

magic-akari:
**Description:**

- [x] CJS
- [x] AMD
- [x] UMD
- [ ] System.js
- [x] dynamic import
- [x] import.meta.url
- [x] lazy require
- [x] ~~top level await (report error)~~ leave it whithout error
- [x] handle `import =`/`export =`
- [x] support nodenext

**BREAKING CHANGE:**
- swc_ecma_transforms_module
- swc_ecma_preset_env

-  #3940
- [x] fix #1135
- [x] fix #4063
- [x] fix #4313
- [ ] not sure  #4456
- [x]  fix #4534
- [x]  fix #4567
- [x] fix #4866 
- [x]  fix #4981
- [x] fix #4984 
- [x] fix #4988
- [x] fix #5012
- [x] fix #5042


crates/swc_ecma_transforms_module/src/cjs.rs
```diff
@@ -0,0 +1,289 @@
...
+                }
+                Specifier::ExportStar => {
+                    should_re_export = true;
+                }
+            });
+
+            // require("mod");
+            let import_expr = self
+                .resolver
+                .make_require_call(self.unresolved_mark, src, src_span);
```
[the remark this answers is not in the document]
magic-akari: OK, `resolver.make_require_call` will handle the swc helpers.

crates/swc_ecma_transforms_module/tests/fixture/common/issue-4988/3/output.amd.js
```diff
@@ -0,0 +1,9 @@
+define([
+    "require",
+    "foo"
+], function(require, _foo) {
+    "use strict";
+    _foo.bar = 1;
+    module.exports = _foo;
```
magic-akari, line 7: Expect `export = expr` to transform to `return expr`.
For similar reasons to preserve `import =`, we should preserve `export =`.

kdy1: In AMD? I think it makes sense.

magic-akari: > In AMD? I think it makes sense.

`import=` implies nointerop.

---
`export = ` will overwrite all the exports.

In AMD, we should return it in the end. [TS REPL](https://www.typescriptlang.org/play?jsx=1&module=2&sourceMap=true#code/JYWwDg9gTgLgBAMwhRUIjgIiRTBuAKFEljgCMBDKOAXjigFMBHAV2EYApNKpMBKQgRwcBBHiMEAbBvEoAvWnACMhBgA8S8OvMHzFAJjxA).

It's similar in [CJS](https://www.typescriptlang.org/play?jsx=1&module=1&sourceMap=true#code/JYWwDg9gTgLgBAMwhRUIjgIiRTBuAKFEljgCMBDKOAXjigFMBHAV2EYApNKpMBKQgRwcBBHiMEAbBvEoAvWnACMhBgA8S8OvMHzFAJjxA).

---

I think we can allow this option in both AMD and CJS module transforms.

kdy1: Yeah I think you are right

magic-akari:
module_transform_flow_chart

crates/swc/tests/fixture/deno/paths/issue-2724/output/src/index.ts
```diff
@@ -1,8 +1,8 @@
 "use strict";
-var _interop_require_wildcard = require("@swc/helpers/lib/_interop_require_wildcard.js").default;
+var _interopRequireWildcardMjs = require("../../../../../../../../../packages/swc-helpers/src/_interop_require_wildcard.mjs");
```
magic-akari, line 2: there is something strange.
swc generated wrong path on my machine. but the correct code  on CI.

I don't know how does the `importResolver` work.

magic-akari: image
I checked out the main branch, and run into the same issue on my machine(MacBook Pro (13-inch, M1, 2020)).
Do you have any clues? @kdy1

kdy1: Can you try running yarn unlink @swc/helpers?

magic-akari: I tried but no luck.
nothing linked to `@swc/helpers`, no global package in yarn.
I cleaned the yarn cache and nothing changed.

magic-akari:
There are still some old test cases to migrate.
But I think it is ready to review. @kdy1 

I will open new a PR to implement System.js and UMD.

crates/swc/tests/fixture/issues-1xxx/1423/case1/output/index.js
```diff
@@ -1,5 +1,5 @@
-"use strict";
 var _necessary = require("necessary");
+"use strict";
```
kdy1, line 2: This seems like a wrong change?

kdy1 (changes requested):
I think `use strict` should be the first line

crates/swc_ecma_transforms_module/tests/fixture/common/ignore-dynamic/1/.input.js
```diff
@@ -0,0 +1,11 @@
...
+
+function foo() {
+    await import("foo");
+
+    callback(() => import("foo"));
+}
+
+import("side-effect");
+
+await import("awaited");
```
magic-akari, line 11: I got an error
```
top level await requires target to es2017 or higher and topLevelAwait:true for ecmascript
```

How to set the parser target with `test_fixture`? @kdy1

crates/swc/tests/fixture/issues-1xxx/1423/case1/output/index.js
magic-akari: You are right. I will fix it

crates/swc_ecma_transforms_module/tests/fixture/common/ignore-dynamic/1/.input.js
kdy1: You can change https://github.com/swc-project/swc/blob/29522b60a53e3c3dcd7078cc7abd4b9615ee6461/crates/swc_ecma_transforms_testing/src/lib.rs#L163 to use https://docs.rs/swc_ecma_parser/latest/swc_ecma_parser/struct.Parser.html#method.new_from with `EsVersion::latest()`.


(This is my mistake)

crates/swc/tests/fixture/issues-1xxx/1859/case2-ts/output/index.ts
```diff
@@ -2,29 +2,6 @@
...
-    if (key in exports && exports[key] === _databaseConfig[key]) return;
-    Object.defineProperty(exports, key, {
-        enumerable: true,
-        get: function() {
-            return _databaseConfig[key];
-        }
-    });
-});
+var _interopRequireWildcardMjs = require("@swc/helpers/lib/_interop_require_wildcard.js").default;
+_re_export(exports, require("./app.config"));
```
kdy1, line 6: I think we need to import this from somewhere?

kdy1 (changes requested):

crates/swc/tests/fixture/module/ignore-dynamic/1/output/index.js
```diff
@@ -1,6 +1,6 @@
 "use strict";
-var _interop_require_wildcard = require("@swc/helpers/lib/_interop_require_wildcard.js").default;
-var _foo = _interop_require_wildcard(require("foo"));
+const _interopRequireWildcardMjs = require("@swc/helpers/lib/_interop_require_wildcard.js").default;
+const _foo = _interop_require_default(require("foo"));
```
kdy1, line 3: We need definition of `_interop_require_default`

crates/swc/tests/fixture/issues-4xxx/4897/output/index.ts
```diff
@@ -1,19 +1,26 @@
...
     });
-    _exports.boo = boo;
-    _exports.Foo = void 0;
+    function _export(target, all) {
+        for(var name in all)Object.defineProperty(target, name, {
+            get: all[name],
+            enumerable: true
+        });
+    }
+    _export(exports, {
```
kdy1, line 15: We need a helper for this

crates/swc_ecma_transforms_module/Cargo.toml
```diff
@@ -42,4 +43,5 @@ swc_ecma_loader = { version = "0.30.0", path = "../swc_ecma_loader", features =
 ] }
 swc_ecma_transforms_compat = { version = "0.104.0", path = "../swc_ecma_transforms_compat" }
 swc_ecma_transforms_testing = { version = "0.91.0", path = "../swc_ecma_transforms_testing" }
+swc_ecma_transforms_typescript = { version = "0.124.0", path = "../swc_ecma_transforms_typescript" }
```
kdy1, line 46: This is wrong.
It will block publishing

crates/swc_ecma_transforms_module/tests/.old.amd.rs
```diff
@@ -0,0 +1,1341 @@
+use swc_common::{chain, Mark};
```
kdy1, line 1: Can you delete these files?

magic-akari: I will delete it after all test cases are migrated.

crates/swc/tests/fixture/issues-4xxx/4897/output/index.ts
magic-akari: The `_export` is always inlined if needed.
Do you prefer a helper for it?

kdy1: No I missed the declaration

crates/swc/tests/fixture/issues-2xxx/2281/case1/output/index.js
```diff
@@ -1,4 +1,7 @@
-var _func = require("./src/func");
+Object.defineProperty(exports, "__esModule", {
```
kdy1, line 1: This test case have `noInterop: true`

magic-akari: If there is nothing to export, should we omit the `__esModule` ?

magic-akari: tsc always emit if the source is ESM module (noInterop: false)
But Babel will omit it if nothing to export.

magic-akari: I will revert some commits to follow the Babel behaviour.

crates/swc_ecma_transforms_base/src/feature.rs
```diff
@@ -0,0 +1,253 @@
...
+    BugfixTransformSafariBlockShadowing, // TODO
+
+    /// `bugfix/transform-safari-for-shadowing`
+    BugfixTransformSafariForShadowing, // TODO
+
+    /// `bugfix/transform-v8-spread-parameters-in-optional-chaining`
+    BugfixTransformV8SpreadParametersInOptionalChaining, // TODO
+}
+
+pub type FeatureSet = Rc<RefCell<AHashSet<Feature>>>;
```
kdy1, line 159: I'm a bit worried about `Rc<RefCell<T>>`.

kdy1: This type is not used by compat transforms, right?

kdy1: I'm asking this because some passes transform file in parallel.

kdy1: Is `Rc<RefCell<T>>` really required?

kdy1 (commented):
It's just a question, not a change request

crates/swc/tests/fixture/issues-1xxx/1213/output/index.js
```diff
@@ -2,16 +2,18 @@
...
+const _interopRequireDefault = require("@swc/helpers/lib/_interop_require_default.js").default;
+const _foo = _interopRequireDefault(require("foo"));
+class OK {
     constructor(){
         console.log(_foo.default);
     }
 }
-exports.default = NotOK;
-class OK {
+class NotOK {
```
kdy1, line 16: Seems like order of `Ok` and `NotOk` is changed?

kdy1: Nevermind, this was a bugfix

crates/swc_ecma_transforms_base/src/feature.rs
magic-akari: I need to modify it in build function (preset or ES version target) and use it in module pass.
See 
https://github.com/swc-project/swc/blob/058e09f4d65d13cf4563c3f77f9285088e4ee6eb/crates/swc_ecma_preset_env/src/lib.rs#L71-L77

And
https://github.com/swc-project/swc/blob/fb299ce89f94e5ba2abf82b1e11b1c54fb42e92b/crates/swc/src/builder.rs#L198-L203

Is it possible to just use `&mut `?

kdy1: I think so? It can accept &mut AHashSet instead.

magic-akari: I will try the bitflag which is easy to copy and modify.

magic-akari: Done.

crates/swc_ecma_transforms_base/src/feature.rs
```diff
@@ -0,0 +1,246 @@
+#![allow(non_upper_case_globals)]
+use bitflags::bitflags;
+use swc_ecma_ast::EsVersion::{self, *};
+
+bitflags! {
+    #[derive(Default)]
+    pub struct FeatureFlag: u64 {
+        /// `transform-template-literals`
+        const TemplateLiterals = 1 << 1;
```
magic-akari, line 9: oh, I should start at 0 .

crates/swc_ecma_transforms_module/tests/fixture/common/strict/export-3/output.amd.js
```diff
@@ -0,0 +1,6 @@
+define([
+    "require",
+    "foo"
+], function(require, _foo) {
+    "use strict";
```
kdy1, line 5: Isn't this module?
We don't have `__esModule` for empty modules?

magic-akari: OK, I checked the Babel output and find it has `__esModule`.
The conclusion is that if there is an any export statement, empty binding or not, it should output the `__esModule`, right?
I will fix this soon.

kdy1: I think so?
`tsc` also emits it. https://www.typescriptlang.org/play?module=1#code/KYDwDg9gTgLgBAbwL5A

magic-akari: If there is nothing to export, but some import statements, tsc and babel behave differently.

I tested some examples of importing empty files in native ESM.
### I think as long as it's an esm module it should be always marked as `__esModule`.


In native ESM module environment, we cannot import default or named binding from empty module file.
An error will be thrown out.

We should mimic this behaviour when transfroming ESM module.
If there is no `__esModule ` mark, the imported default will be empty object `{}` which is not expected.

With `__esModule ` mark, we will get undefined instead.
It's not as perfect as native esm error report but better than empty object.

kdy1: Yeah I agree, we should have `__esModule`

magic-akari: Is it possible for a non-esm to be passed into the cjs pass?

kdy1: I don't think so, because it's not a module and `cjs` pass is for modules

magic-akari:
It seems there is something wrong with the sourcemap.
The test result jumps between correct and broken every time I run the test.

crates/swc/tests/stacktrace/case-1/input/_exec.js

Correct:
Screenshot 2022-06-27 at 09 18 44

Broken:
image

@kdy1 Do you have any idea about it?

kdy1:
Possibbly `AHashMap`?

magic-akari:
I think it is caused by the output file being transformed again as input file.
There are two `__esModule` marks here.

It's hard to detect if there is already a `__esModule` mark.
Any way, Maybe it should be conservative and add `__esModule` mark only when necessary.

magic-akari:
https://github.com/swc-project/swc/blob/f41c82ac67717afb7f5deef5ed306e3d2b4a07dd/crates/swc/tests/source_map.rs#L164-L168

It is configed as module here.

magic-akari:
I have try to change it into IsModule::Unknown, but failed with this error.
```
Cannot use script with externalHelpers; There's no standard way to import other modules
```

OK, do not emit `__esModule` if there is no import nor export.

crates/swc/tests/tsc-references/ambientDeclarationsExternal_es2015.1.normal.js
```diff
@@ -6,6 +6,3 @@ const imp1 = require('equ');
 const imp3 = require('equ2');
 var n = imp3.x;
 var n;
-//@Filename: decls.ts
-// Ambient external module with export assignment
-export { };
```
magic-akari, line 11: Regression. should not delete this.

crates/swc/tests/tsc-references/duplicateExportAssignments_es2015.1.normal.js
```diff
@@ -20,5 +20,5 @@ function y() {
 var x = 5;
 var y = "test";
 var z = {};
-module.exports = z;
+module.exports = x;
```
magic-akari, line 23: It's a bugfix.
tsc only count first `export =`.

crates/swc_ecma_transforms_module/tests/fixture/common/misc/import-const-throw/output.amd.js
```diff
@@ -0,0 +1,23 @@
...
+    "bar",
+    "baz"
+], function(require, exports, _foo, _bar, _baz) {
+    "use strict";
+    Object.defineProperty(exports, "__esModule", {
+        value: true
+    });
+    _foo = _interopRequireDefault(_foo);
+    _bar = _interopRequireWildcard(_bar);
+    Foo = 42;
```
kdy1, line 14: I think this will not throw an error if we don't have `use strict`.
But I think the output should throw regardless of `strict` if original source code throws

crates/swc/tests/tsc-references/importCallExpressionInExportEqualsAMD_es2015.1.normal.js
```diff
@@ -1,9 +1,5 @@
-import _async_to_generator from "@swc/helpers/src/_async_to_generator.mjs";
-// @filename: index.ts
-module.exports = _async_to_generator(function*() {
-    const something = yield import("./something");
-});
 // @module: amd
 // @target: esnext
 // @filename: something.ts
```
magic-akari, line 3: There are a lot of confusing changes here.
Since swc test kit cannot handle the `@filename` now, the contents of multiple files are treated as one file.

crates/swc_ecma_transforms_module/tests/fixture/common/misc/import-const-throw/output.amd.js
magic-akari: I moved the reassign check into the lint pass.
- #5033

crates/swc_ecma_transforms_module/tests/fixture/common/lazy/issue-3081/2/output.cjs
```diff
@@ -0,0 +1,16 @@
...
+    const data = require("lib");
+    _lib = function() {
+        return data;
+    };
+    return data;
+}
+function myFn() {
+    _lib().fn();
+}
+class MyClass extends _lib().Klass {
```
kdy1, line 15: This seems not ideal, although I think it's fine for now.

crates/swc_ecma_transforms_module/tests/fixture/common/issue-4988/2/input.ts
```diff
@@ -0,0 +1,3 @@
+import * as testUtils from "vs/base/test/common/testUtils";
+
+export import flakySuite = testUtils.flakySuite;
```
kdy1, line 3: I think we should return/export something for this syntax

magic-akari: OK, It should be handled by TS strip pass.

magic-akari: This issue is about the test config.
There is no TS strip pass in the swc_ecma_transforms_module tests.

In actual use, the strip pass of TS is always  before the module pass, so there is no such issue.
I should move this test case to swc integration test.

crates/swc_ecma_transforms_module/tests/fixture/common/issue-4898/3/output.amd.ts
```diff
@@ -0,0 +1,27 @@
+define([
+    "require"
+], function(require) {
+    "use strict";
+    Object.defineProperty(exports, "__esModule", {
+        value: true
+    });
+    _assert(true);
+    module Foo {
```
kdy1, line 9: Seems like this should be moved to integration test, too?

magic-akari:
It's really a hard work to review it. Thank you.

crates/swc_ecma_transforms_module/src/amd.rs
```diff
@@ -1,962 +1,481 @@
...
+        let mut stmts = Vec::with_capacity(link.len());
+
+        let mut export_obj_prop_list = export
+            .into_iter()
+            .map(|((key, span), ident)| (key, span, ident.into()))
+            .collect();
+
+        link.into_iter().for_each(
+            |(src, LinkItem(src_span, link_specifier_set, mut link_flag))| {
+                let is_swc_detault_helper =
```
kdy1, line 308: I found a typo

kdy1 (changes requested):

crates/swc_ecma_transforms_typescript/src/import_export_assign.rs
```diff
@@ -0,0 +1,226 @@
...
+        config,
+        export_assign: None,
+        found_import_assign: false,
+        found_export_assign: false,
+    })
+}
+
+struct ImportExportAssign {
+    unresolved_mark: Mark,
+    config: TSImportExportAssignConfig,
```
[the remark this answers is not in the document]
magic-akari: And the TSEnumConfig?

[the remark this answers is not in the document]
kdy1: Oh, yeah I think we should rename it too

kdy1 (approved):
Thank you so much again!

Let's bump `swc_ecma_transforms_base` so error of the bump bot does not affect users of swc.

---



swc-bump:

PR #4759, opened 2022-05-23
fix(swc/transforms): preserve referenced imports for namespace

### LINKED ISSUE #4481: Transpilation of import * as var is undefined in namespace usage
### Describe the bug

Following code is transpiled faulty:
```
import * as I from "./ABC";

/* eslint-disable @typescript-eslint/no-namespace */
export namespace IE {
    export import A = I.A;
    export import C = I.C;
    export import D = I.D;
    ...
}
```
To
```
(function(I1) {
    var A = I.A;
    I1.A = A;
    var C = I.C;
    I1.C = C;
    var D = I.D;
    I1.D = D;
    ...
})(IE || (exports.IE = IE = {}));
```

`VM2679:1 Uncaught ReferenceError: I is not defined`
The var `I` doesnt exist in the transpiled file at all anymore except within the assignation part in the IIFE.

### Input code

_No response_

### Config

```json
{
                    test: /\.ts(x?)$/,
                    use: 
                        {
                            loader: "swc-loader",
                            options: {
                                sync: true,
                                jsc: {
                                    parser: {
                                        syntax: "typescript",
                                        tsx: true,
                                        dynamicImport: true
                                    },
                                    loose: false,
                                    externalHelpers: true,
                                    keepClassNames: false,

                                },
                                module: {
                                    type: "commonjs",
                                    strictMode: false,
                                    lazy: true
                                }
                            }
                        }
                }
```

### Playground link

_No response_

### Expected behavior

The import * as VAR variable should still be defined, as it is used in the namespace code. But it seems to be gone.

### Actual behavior

The var is not existing causing is not defined error.

### Version

1.2.173

### Additional context

_No response_

kwonoj:
- fix(swc/transform): preserve import references to namespace reexports
- test(swc/transforms): update test cases

**Description:**

This PR attempts to fix #4481 which loses reference to imports if reference occurred in namespace as reexports. When module_items manually visit importdecl visitor hasn't completed to visit inner idents to update references, in result import gets stripped eagerly. PR changes logic to ensure reference if namespace scoped ident reference is non type-only reexports.

**BREAKING CHANGE:**

#4481

### DIFF
--- a/crates/swc_ecma_transforms_typescript/src/strip.rs
+++ b/crates/swc_ecma_transforms_typescript/src/strip.rs
@@ -1658,6 +1658,17 @@ where
                     match name {
                         TsEntityName::Ident(ref i) => {
                             entry.maybe_dependency = Some(i.clone());
+                            // Eagerly update referenced idents for the concrete exports
+                            // resolves https://github.com/swc-project/swc/issues/4481, if we
+                            // know reference exists & reexports.
+                            // when referenced idents are scoped in ts namespace, its references
+                            // is updated _after_ visiting import decl which strips out imports
+                            // already.
+                            if n.is_export && !n.is_type_only {
+                                let entry =
+                                    self.scope.referenced_idents.entry(i.to_id()).or_default();
+                                entry.has_concrete = true;
+                            }
                             break;
                         }
                         TsEntityName::TsQualifiedName(ref q) => name = &q.left,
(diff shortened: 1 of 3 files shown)


kdy1 (approved):
Thank you!

---

swc-bump:

kdy1 (approved):

PR #4762, opened 2022-05-23
feat(html/parser): improve public api

alexander-akait:
**Description:**

Implement iterator + bench them + remove unused methods from `ParserInput` + remove config for lexer (lexer for HTML doesn't need options) + method for set last start tag

**BREAKING CHANGE:**

Yes

No


crates/swc_html_parser/src/lexer/mod.rs
```diff
@@ -8,67 +8,9 @@ use swc_html_ast::{AttributeToken, Token, TokenAndSpan};
...
-            attribute_start_position: None,
-            character_reference_code: None,
-            temporary_buffer: None,
-            is_adjusted_current_node_is_element_in_html_namespace: None,
-            doctype_keyword: None,
-            last_emitted_error_pos: None,
-        }
-    }
-}
-
```
alexander-akait, line 71: Just move it below to be align with `impl`

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser --breaking

PR #4763, opened 2022-05-23

kdy1:
# Documentation for future readers

I can't remember correctly, but IIRC, it was about invoking IIFE containing trivial calls like 

```js
(function iife() {
	a() ? b() : c()
})()

```

but at the time of writing PR, it broke for some reason if I allowed other kinds of expressions.

---

**Description:**


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

kdy1 (commented):
reset ci

PR #4764, opened 2022-05-23
test(html/parser): enable error reporting tests

alexander-akait:
**Description:**

Now we handle all errors in HTML (you can use our parser as HTML validator :smile: ) and these tests show it, anyway there 
are some problems in `html5lib-tests`, out it under TODO - let's uncomment then they will be fixed

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!


---

swc-bump:
 - swc_html_parser

PR #4766, opened 2022-05-23
fix(html/codegen): output textarea and pre in foreign context

alexander-akait:
**Description:**

Fix textarea and pre in foreign context

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_codegen

PR #4767, opened 2022-05-24
feat(es/helpers): Import only used helpers

### LINKED ISSUE #2745: When externalHelpers is true, only import the helpers that are being used instead of the entire namespace
### Describe the feature

When using swc in its default configuration, all helpers except for `regenerator-runtime` are being inlined as they're needed.
For example, when targeting `es5`, the following code:

```js
typeof "";
```

Will be transpiled into:

```js
var _typeof = function(obj) {
    return obj && typeof Symbol !== "undefined" && obj.constructor === Symbol ? "symbol" : typeof obj;
};
_typeof("");
```

This adds the smallest possible footprint to the compiled output, which is great.

However, in cases where the `externalHelpers` option are set to to `true` in the options for `jsc`, this will be transpiled into:

```js
swcHelpers.typeOf("");
```

This leads to the entirety of the `@swc/helpers` module being imported as a namespace, and referenced via property access expressions. If this file is flattened, for example via a bundler, this is not tree-shakeable and will produce a file with 2209 lines of code when bundled with Rollup.

[Here's the output](https://gist.github.com/wessberg/d09e820b7c8b8bb1ca7dc50bde3194b6)

When using external helpers, its quite often to make it possible to avoid duplicating them across files in a bundle, but for many cases, inlining them will actually still produce the smallest possible output due to not having to include each and every helper unnecessarily. This is unfortunate, and I hope you'll consider changing it.

### How Babel does it

Since swc is seen and positioned in the ecosystem as a drop-in replacement for Babel, and since Babel has been around for a while, it's always useful to compare these things with how it is treated there.

In Babel, when using the `@babel/plugin-transform-runtime`, plugin, helpers are referenced via import statements, importing from the `@babel/runtime` module, and _only_ the specific helpers being used are imported. In the same example as before, here's the output when the `helpers` option are set to `true` with `@babel/plugin-transform-runtime` (when targeting ES modules):

```js
import _typeof from "@babel/runtime/helpers/esm/typeof";

_typeof("");
```

That adds the smallest possible footprint to the compiled output, including across files where the `_typeof` helper may be referenced multiple times.

`@babel/plugin-transform-runtime` transforms helpers into imports from `@babel/runtime`, which may serve as inspiration here

### Additional context

_No response_

kdy1:
**Description:**

 - Closes #2745


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base
 - swc_ecma_transforms_module

PR #4770, opened 2022-05-24
feat(html/codegen): omit end tags

alexander-akait:
**Description:**

HTML allows to omit start and end tag in some cases, so we can use it for minification.

This is PoC, actually we can simply it once we have ability to get parent node in emit/visitor (I think we will have the same logic if we implemented it), currently I just create additional method and move logic to basic function.

If solution is ok feel free to merge and then I will implement it for all possible tags in HTML.

**BREAKING CHANGE:**

No

No


crates/swc_html_codegen/tests/fixture.rs
```diff
@@ -425,6 +425,7 @@ fn test_indent_type_option(input: PathBuf) {
     );
 }
 
+// TODO minified verification
```
alexander-akait, line 428: Plus TODO for cross testing on minification

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_codegen

PR #4772, opened 2022-05-24
fix(es/resolver): var xx defined behind the catch(xx) block does not recognize scope correctly

wre232114:
**Description:**

the resolver folder does not recognize the following code's scope correctly.

**Input**
```js
try {
    console.log(111);
} catch (Ic) {
    throw Ic;
}

var jc, Ic;
(Ic = jc).PV = "page_view";
```
the `var Ic` appears behind the `catch(Ic)`.

**Expected Output**
```js
try {
    console.log(111);
} catch (Ic__3) {
    throw Ic__3;
}
var jc__1, Ic__1;
(Ic__1 = jc__1).PV = "page_view";
```
the Ic used after `var Ic` should have the same mark as `var Ic`.

**Current Unexpected Output**
```js
try {
    console.log(111);
} catch (Ic__3) {
    throw Ic__3;
}
var jc__1, Ic__1;
(Ic = jc__1).PV = "page_view";
```
Currently `Ic` is treat as `unresolved mark` .

two related test cases added.

**BREAKING CHANGE:**

**Related issue (if exists):**

PR #4773, opened 2022-05-24
fix(es/resolver): fix more edge case mentioned in #4772

wre232114:
**Description:**

see description in #4772 and the test case of this PR. More edge case handled.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

---


swc-bump:
 - swc_ecma_transforms_base

kdy1 (changes requested):
Can you run `./scripts/update-list.sh` and `UPDATE=1 cargo test` in `./crates/swc_ecma_minifier`?
Seems like there was a wrong passing test

wre232114:
ok

kdy1 (approved):

PR #4780, opened 2022-05-24
feat(html/codegen): omit start and end tags when it is possible

alexander-akait:
**Description:**

Reduce size of HTML when it is possible

**BREAKING CHANGE:**

No

No


alexander-akait:
@kdy1 Rebase, will be great to review it too

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_codegen

PR #4783, opened 2022-05-24
feat(html/minifier): sort classes

alexander-akait:
**Description:**

Sort classes for better gzip

https://github.com/terser/html-minifier-terser#sorting-attributes--style-classes marks it as unsafe, but it is full safe for valid HTML, `html-minifier-terser` supports custom parsing, i.e. you can use `<?php ?>` and any other tags, but we don't support it right now (and maybe never, need feedback, but it is exotic usage and compression non valid HTML is generally unsafe).

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---


swc-bump:
 - swc_html_minifier

PR #4788, opened 2022-05-25
fix(es/minifier): Consider side effects while removing an object spread

kdy1:
**Description:**

I was adding tests to verify https://github.com/swc-project/swc/pull/4754 but I found a bug of minifier


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4789, opened 2022-05-25
test(es): Make execution tests faster

kdy1:
**Description:**

 - We now cache the execution result if possible.


kdy1 (commented):
swc-bump:
 - ''

Issue #4790, opened 2022-05-25
Minifying seems to be 3x as slow since 1.2.181


mohd-akram:
### Describe the bug

I have a build that takes 89 seconds when using 1.2.179, and 253 seconds in 1.2.181. It's similarly slow with the latest version.

### Expected behavior

Minifying should be fast.

### Version

1.2.192

kdy1:
https://github.com/swc-project/swc/blob/1ac313a382780ad99114eacc8af5a8c66d0ac3ab/CHANGELOG.md#12181---2022-05-09


What's your environment? Is it linux?

mohd-akram:
Yes, Fedora 36.

`Linux my-pc 5.17.8-300.fc36.x86_64 #1 SMP PREEMPT Mon May 16 01:00:37 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

mohd-akram:
Looks like it was indeed flagged in the jemalloc commit - https://github.com/swc-project/swc/commit/af91094b2db0809f02e6f0a6f07b021d12aa1512.

kdy1:
No, benchmark action is flaky due to noisy neighbors

mohd-akram:
To such an extent though? 2.31 ratio.

kdy1:
Oh... 2.31 is not expected.
I was going to profile on my device because I though https://github.com/swc-project/swc/commit/3521ce09e7a2c3168f17491196f6fbc835637380 can be the cause, but 2.31 is way too much


## Profile result

Device: `m1 macbook pro 64gb ram`


Current perf:
```

Gnuplot not found, using plotters backend
Benchmarking es/minify/libraries/antd
Benchmarking es/minify/libraries/antd: Warming up for 3.0000 s

Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 7.7s.
Benchmarking es/minify/libraries/antd: Collecting 10 samples in estimated 7.6780 s (10 iterations)
Benchmarking es/minify/libraries/antd: Analyzing
es/minify/libraries/antd
                        time:   [755.37 ms 756.77 ms 758.01 ms]
                        change: [-0.0054% +0.4395% +0.8395%] (p = 0.07 > 0.05)
                        No change in performance detected.
Benchmarking es/minify/libraries/d3
Benchmarking es/minify/libraries/d3: Warming up for 3.0000 s
[...]

PR #4794, opened 2022-05-25
Fix jemalloc

Brooooooklyn:
**Description:**

**BREAKING CHANGE:**


kdy1 (changes requested):
We should use a better allocator for other linux, too.

See https://github.com/swc-project/swc/issues/4790

Brooooooklyn:
after this PR  jemalloc will be applied to `linux-x64-gnu` and `linux-aarch64-gnu`:

`[target.'cfg(all(target_os = "linux", target_env = "gnu", any(target_arch = "x86_64", target_arch = "aarch64")))'.dependencies]`

kdy1:
Oh, I see. Thank you!
I-ll mergr this PR tomorrow

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_node_base

PR #4796, opened 2022-05-25
perf(html/parser): improve lexer perf

alexander-akait:
**Description:**

Avoid clone token, because we need only name

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser --breaking

PR #4805, opened 2022-05-26
feat(es/compat): Don't create vars for computed class methods if possible

kdy1:
**Description:**

 - https://discord.com/channels/889779439272075314/889780890098610176/979284345435086898


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_compat

Austaras:
Wait, this is needed for evaluation order, consider following
```js
class A {
  [b()] () {}
  [a()] = 213
}
```
if we only have class properties transformed, evaluation order would be broken

crates/swc_ecma_transforms_compat/src/es2022/class_properties/mod.rs
```diff
@@ -516,6 +514,19 @@ impl<C: Comments> ClassProperties<C> {
             private: &self.private,
         });
 
+        let should_create_vars_for_method_names = class.body.iter().any(|m| match m {
```
kdy1, line 517: @Austaras We check for such cases here.
PR title is bit misleading

PR #4806, opened 2022-05-26
fix(es/minifier): Don't create names which is same as an unresolved var

kdy1:
**Description:**

 - `https://github.com/vercel/next.js/discussions/30237#discussioncomment-2819480`


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4807, opened 2022-05-26
fix(es/modules): Fix lazy option of common js

### LINKED ISSUE #4799: esm-style exports and imports don't get lazy loaded
### Describe the bug

When transpiling with lazy: true, dependencies that are declared with export are not being lazyfied.
The same is done with imports. They should also be lazy but instead are handled like below the export is done with the direct require. 

### Input code

_No response_

### Config

```json
{
                    test: /\.ts(x?)$/,
                    use: 
                        {
                            loader: "swc-loader",
                            options: {
                                sync: true,
                                jsc: {
                                    parser: {
                                        syntax: "typescript",
                                        tsx: true,
                                        dynamicImport: true
                                    },
                                    loose: false,
                                    externalHelpers: true,
                                    keepClassNames: false,

                                },
                                module: {
                                    type: "commonjs",
                                    strictMode: false,
                                    lazy: true
                                }
                            }
                        }
                }
```

### Playground link

_No response_

### Expected behavior

Code that looks like this 

```
export {
    createP
} from './St';

```
gets transpiled to 

```
Object.defineProperty(exports, "createP", ({
  enumerable: true,
  get: function get() {
    return _St().createP;
  }
}));

function _St() {
  var data = __webpack_require__("../path/to/St");

  _St = function _St() {
    return data;
  };

  return data;
}
```

### Actual behavior

Code that looks like this 

```
export {
    createP
} from './St';

```
gets transpiled to 

```
Object.defineProperty(exports, "createP", ({
  enumerable: true,
  get: function get() {
    return _st.createP;
  }
}));

var st = __webpack_require__ ("./path/to/St");
```

### Version

1.2.192

### Additional context

_No response_

kdy1:
**Description:**

 - Closes https://github.com/swc-project/swc/issues/4799

### DIFF
--- a/crates/swc_ecma_transforms_module/src/common_js.rs
+++ b/crates/swc_ecma_transforms_module/src/common_js.rs
@@ -69,13 +69,13 @@ pub fn common_js_with_resolver(
 
 struct LazyIdentifierVisitor {
     scope: Rc<RefCell<Scope>>,
-    top_level_idents: AHashSet<JsWord>,
+    denied_sources: AHashSet<JsWord>,
 }
 impl LazyIdentifierVisitor {
     fn new(scope: Rc<RefCell<Scope>>) -> Self {
         LazyIdentifierVisitor {
             scope,
-            top_level_idents: Default::default(),
+            denied_sources: Default::default(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_module

PR #4815, opened 2022-05-26
test(html/codegen): avoid newline at the end

alexander-akait:
**Description:**

Vaoind adding `\n` at the end if test file because `\n` is part of DOM (TextNode)

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thank you!

---

swc-bump:
 - testing

PR #4823, opened 2022-05-27
test(es/minifier): Add snapshots for the mangler

kdy1:
**Description:**

I found a way to make name mangler parallel, but I want to ensure that we don't break anything.


kdy1 (commented):
swc-bump:
 - ''

PR #4825, opened 2022-05-27
fix(html/codegen): more cases for omitting

alexander-akait:
**Description:**

~~It is bug in spec~~, they missed these cases (due adoption algorithm), I will send a fixes to HTML spec too

+ fix bugs for comments in `html` and document fragment
+ reorder tests
+ remove already solved

Report in HTML spec - https://github.com/whatwg/html/issues/7962, based on answer, we should not apply tag omitting for non confirming HTML documents (i.e. parsed with error), I think we should mark it as limitation and provide option to disable it on codegen (I will implement it in the next PR), but I think it is normal by default to omit them, because developer should prefer to have valid HTML document, instead trying to minify broken HTML and expect from browser works fine (note - behaviour in some browsers are different, so you can't rely on it too)

And I found bug in spec - XSS, just put:
```html
<svg></p><style><a title="</style><img src onerror=alert(1)>">
```

and open Chrome :exploding_head: 

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_html_codegen

PR #4829, opened 2022-05-28
feat(bundler): Experiment JS interop

kdy1:
**Description:**

**BREAKING CHANGE:**


crates/swcpack_node_interop/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+    asset::{AssetLoader, AssetLoaderContext, AssetProcessor},
+    esm::{EsModule, EsmPreprocessor, EsmPreprocessorContext, EsmProcessor},
+    file::FileLoader,
+    resource::{Res, ResourceIdGenerator},
+    Timer,
+};
+
+mod helper;
+
+pub struct JsAssetLoader {
```
alexander-akait, line 20: Maybe `EcmaAssetLoader`, because we use `ecma` everywhere for `js`? Same for `Js` everywhere to be align with our naming

kdy1: It's confusing as we have esm loader

PR #4831, opened 2022-05-28
feat(node): Split css binding

kdy1:
**Description:**

To ease version management, I created https://github.com/swc-project/css


kdy1 (commented):
swc-bump:
 - ''

PR #4846, opened 2022-05-30
fix(html/parser): fix error span for lexer

alexander-akait:
**Description:**

Unfortunately we can't use `html5lib_tests` to check lexer's error position, in some tests position is offset, in some column (and you can't undestand it from tests), in some tests codes are not the same as in spec (I double check it and also check it using w3 validator, our parser and w3 validator have the same code and position, but html5tests_lib not, I think it is just bug on their side), so I added tests and improve logic for error reporting (w3 validator has the same logic).

**BREAKING CHANGE:**

No

No


alexander-akait:
@kdy1 Rebased

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser

Issue #4841, opened 2022-05-29
Minifier: `if_return` removes return


mischnic:
### Describe the bug

`if_return` incorrectly removes a return (and adds `void void ...` which is unneeded).

### Input code

```typescript
global._processChunk = function _processChunk(chunk, async, newReq) {
  if (chunk) {
    if (!async) {
      return true;
    }
    var newReq = this.write();
    newReq.callback = callback;
    return;
  }
  cb();
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2021",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": false,
        "booleans": false,
        "booleans_as_integers": false,
        "collapse_vars": false,
        "comparisons": false,
        "computed_props": false,
        "conditionals": false,
        "dead_code": false,
        "defaults": false,
        "directives": false,
        "drop_console": false,
        "drop_debugger": false,
        "ecma": 5,
        "evaluate": false,
        "expression": false,
        "global_defs": {},
        "hoist_funs": false,
        "hoist_props": false,
        "hoist_vars": false,
        "ie8": false,
        "if_return": true,
        "inline": 0,
        "join_vars": false,
        "keep_classnames": false,
        "keep_fargs": false,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": false,
        "module": false,
        "negate_iife": false,
        "passes": 1,
        "properties": false,
        "pure_funcs": [],
        "pure_getters": false,
        "reduce_funcs": false,
        "reduce_vars": false,
        "sequences": false,
        "side_effects": false,
        "switches": false,
        "top_retain": [],
        "toplevel": false,
        "typeofs": false,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_symbols": false,
        "unsafe_undefined": false,
        "unused": false
      },
      "mangle": true
    },
    "transform": {
      "legacyDecorator": true
    },
    "externalHelpers": true
  },
  "module": {
    "type": "es6"
  },
  "minify": true,
  "isModule": true
}
```

### Playground link

[...]

PR #4847, opened 2022-05-30
fix(es/minifier): Make `if_return` to not drop required returns

kdy1:
**Description:**

 - Closes https://github.com/swc-project/swc/issues/4841


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4853, opened 2022-05-31
feat(wasm): Add json error

kdy1:
**Description:**

I'm not sure if it's generally useful, so I added it only to wasm.
But I think we need some `experimental` flag here.

 - https://github.com/swc-project/swc-playground/pull/19#issuecomment-1140815636


kwonoj:
Am I reqding correct this makes error emits into json formatted insted of current format?

kdy1:
Yeah, but I'm not sure about this api.

cc @g-plane

kwonoj:
Yes, this would be useful if someone want to consume swc as separate step for build pipeline and reads its output as machine-readable. Main q is do we have actual usecase.

kdy1:
We have one known usecase at the moment. SWC playground can show errors with better formatting

kwonoj:
I'm in for having as experimental for those case.

g-plane:
Should we pass those options as the third argument in those `parseSync`, `transformSync` function? Two reasons:

- Keep the second argument be focusing on SWC config, nothing else.
- If we have more WebAssembly-specific options in the future, we can reuse that options, instead of mixing the second argument.

kdy1:
Sounds reasonable, thanks!

g-plane:
SWC config's `experimental` property can be used everywhere, not only WebAssembly. Will it confuse those non-WebAssembly users?

kwonoj:
I expect if we'll have this, at least native cli would want same (as experimental). Stdout with structured output is quite common usecase.

g-plane:
Sounds reasonable.

kdy1 (commented):
swc-bump:
 - swc --breaking

PR #4860, opened 2022-06-01
fix(html/parser): fix end span for closed tags

alexander-akait:
**Description:**

Fix most of cases, there are some edge cases (when HTML is invalid, I will fix it late).

Calculation span is tricky in HTML, because you can have:
```html
<div>test <p>test</p><p>test</div>
```

There are elements which can omit ending tag, also HTML can move some elements into another parent and we need recalculate span, also elements and characters after `</body>` and `</html>` move into `<body>` and more things.

Spent the whole day to looking at AST/Span/DOM tests and chec them :tired_face: 

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser

PR #4865, opened 2022-06-02
test(html/parser): refactor

alexander-akait:
**Description:**

Need merge after https://github.com/swc-project/swc/pull/4860

I move each test to own directory and use the same file names for test files as in `fixture` tests, we need it for `span` testing (and more), it will allow to catch remain bugs with span and also catch regressions in future

I had to initially make such a structure, but then I didn’t think so deeply about cross testing with html5tests_lib

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - ''

PR #4869, opened 2022-06-02
perf(html/parser): lexer

alexander-akait:
**Description:**

Refactor code + improve perf

**BREAKING CHANGE:**

No

No


crates/swc_html_parser/src/lexer/mod.rs
```diff
@@ -160,7 +160,6 @@ where
     cur: Option<char>,
     cur_pos: BytePos,
     last_token_pos: BytePos,
-    last_emitted_error_pos: Option<BytePos>,
```
alexander-akait, line 163: We don't need it anymore, because we validate character only one now

kdy1 (approved):
Thanks!

---
swc-bump:
 - swc_html_parser

PR #4879, opened 2022-06-04
refactor(swc/ecma_codegen): support sourcemapper

kwonoj:
**Description:**

This PR looses type requirement to `SourceMap` in codegen's emitter to accept traits (SourceMapper + SourceMapperExt) instead.

Primarily this allows to any arbitary implementation to sourcemap to work with codegen's emitter - such as plugin can use emitter with `PluginSourceMapProxy` instead of host implementation. To achieve this, there were some adjustment as well such as bubbling up listformat to common package.

I've tried suggested approach as well, which makes `Emitter` accepts `Option<Sourcemap>` to allow to emit without sourcemapper implementation. It sort works, quickest, but there are inconsistencies when sourcemap is available vs. not vs. supplying `default` to the emitter for the actual output results. Seems like few paths rely on the sourcemapper creates those differences. 

PR aims to create `identical` behavior for the any sourcemapper implementation argument, which brings consistency when using emitter in any context (host vs plugin). This is bit important if a plugin visitor want to use codegen / emitter internally to ensure it'll behave same either if it's being used in custom pass, or inside of a plugin context.

**BREAKING CHANGE:**


kwonoj:
Updated PR with missing commits to support snap_to_snippet properly.

crates/swc_common/src/source_map.rs
```diff
@@ -35,8 +39,10 @@ pub use crate::syntax_pos::*;
 use crate::{
     collections::AHashMap,
     errors::SourceMapper,
+    list::ListFormat,
```
kdy1, line 42: We should not have `ListFormat` in `swc_common`

kdy1 (changes requested):
Please move it to `swc_ecma_ast`, if it's really required

kwonoj:
Ok, tried to adjust export to use ecma_ast instead. I hoped if there's way to preserve location of listformat, but it doesn't look like I can cut import chain easily to move sourcemapperext only.

kdy1:
Why is `SourceMapperExt` aware of `ListFormat`? Sounds like there's a fundamental issue

kwonoj:
I don't have good historical context how it is composed - only thing I can say is currently sourcemappext / spanext both are tightly copuled with listformat for most of the implementation. (https://github.com/swc-project/swc/pull/4879/files#diff-e55f8cb04960b61fbd39aa6bcfa4cbecfb736780d09c2a7b4659a3376fec3a7eR25)

kwonoj:
(Clarifying : I didn't change implementation, only location's changed to allow use in proxy as well.)

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_common
 - swc_ecma_ast --breaking
 - swc_ecma_proxy --breaking

PR #4880, opened 2022-06-04
feat(html/minifier): remove spaces between nodes in HTML

alexander-akait:
**Description:**

Removing spaces inside `html`, i.e. between `<head>` and `<body>` is also safe, can't found cases when you will have unsafe modifications  

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_minifier

PR #4885, opened 2022-06-05
refactor(html/codegen): avoid token and span

alexander-akait:
**Description:**

Remove unused code

**BREAKING CHANGE:**

Theoretically yes, but we don't have TokenAndSpan in our AST, so should be safe

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_ast --breaking

kdy1 (approved):

PR #4889, opened 2022-06-06
feat(html/codegen): improve entity compression

alexander-akait:
**Description:**

Here:
- move HTML entities to reuse it in parser/codegen
- fix perf bug in lexer (it was typo)
- minify `&` when it is possible, logic from https://github.com/terser/html-minifier-terser 

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_html_minifier

PR #4901, opened 2022-06-07
perf(html/parser): improve

alexander-akait:
**Description:**

Improve perf:
- avoid unencessary memory allocation
- fix perf for `Text`
- update fields in `data` and avoid creation extra buffers
- refactor and simplify code
- rename `value` to `data` as in spec

2-3x imrovement here

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_ast --breaking

Issue #4896, opened 2022-06-07
RuntimeError: Unreachable code should not be executed with "export * import"


jrieken:
### Describe the bug

Disabling `noInterop` and using `export * as foo from '...'` causes a runtime error. 

### Input code

```typescript
export * as icons from 'vs/workbench/contrib/welcomeGettingStarted/browser/gettingStartedIcons';
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2021",
    "loose": false
  },
  "module": {
    "type": "amd",
    "noInterop": true
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.197&code=H4sIAAAAAAAAA1XMMQ6AIAwAwN1XdCNx6QN8gHH2BYAViUJN24jPV0fXG47uk8WgB6%2BQI1eFVbiAuxQbyx6oxg1fN8kBGx2RC41klmuazYvRgkG4KQmmH09f5obuASkqlHRhAAAA&config=H4sIAAAAAAAAA0WMOw6EMAwF7%2BKaYqHMDfYYFjxWoBBHtpFAKHdfp6J7n9E8tNtM6aHKatCe7C7OFyXyu8Jm3arTQG4xrZwNLQrrDx4IbPpMY9xZxPAChyxnRrd1SYB8LEEV%2BRaHSqXkeqK1P3IDBQF%2FAAAA

### Expected behavior

Don't crash

### Actual behavior

```
RuntimeError: Unreachable code should not be executed (evaluating 'wasm.transformSync(retptr, ptr0, len0, addHeapObject(opts), addHeapObject(experimental_plugin_bytes_resolver))')

<?>.wasm-function[rust_panic]@[wasm code]
<?>.wasm-function[std::panicking::rust_panic_with_hook::hdd08feebb16a5d0c]@[wasm code]
<?>.wasm-function[std::panicking::begin_panic_handler::{{closure}}::hb1a3fff83235176b]@[wasm code]
<?>.wasm-function[std::sys_common::backtrace::__rust_end_short_backtrace::h35976c7c7e728385]@[wasm code]
<?>.wasm-function[rust_begin_unwind]@[wasm code]
<?>.wasm-function[core::panicking::panic_fmt::h9f028a3b92cbae68]@[wasm code]
[...]

PR #4904, opened 2022-06-08
fix(es/modules/amd): Support `export * as foo from 'foo'`

kdy1:
**Description:**

 - Closes https://github.com/swc-project/swc/issues/4896


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_module

PR #4913, opened 2022-06-09
feat(es/lint): add default-case-last rule implementation

ArturAralin:
**Description:**

default-case-last rule implementation

**BREAKING CHANGE:**
No


kdy1 (changes requested):
Thank you!

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_lints

PR #4916, opened 2022-06-09
feat(html/minifier): compress whitespaces

alexander-akait:
**Description:**

Compress whitespaces, there are three modes:
1. `all` - remove all wtitespaces, very agressive, fits only if you don't really on `display: inline` (and other) CSS property or you use JS creation of DOM
2. `smart` - based on `display` property, i.e. we keep spaces before/after `display: inline`, fits in most of case
3. `conservative` - only collapse multiple whitespaces in one, we don't remove whitespaces at all, very safe

Also I added many tests.

There are improvements/bugfixes
- Improve `smart` mode and remove trailing/leading whitespace if prev/next text node already remove them (Added TODO)
- Handle better `inline-block`/etc elements (I want to migrate tests from https://github.com/terser/html-minifier-terser/blob/master/tests/minifier.spec.js to fix all cases)

**BREAKING CHANGE:**

Yes

No


crates/swc_html_minifier/src/lib.rs
```diff
@@ -434,13 +459,128 @@ impl Minifier {
...
+            _ => WhitespaceMinificationMode {
+                collapse: false,
+                destroy_whole: false,
+                trim: false,
+            },
+        }
+    }
+
+    fn is_all_whitespace(&self, data: &str) -> bool {
+        for c in data.chars() {
```
kdy1, line 537: I think this can be `!data.chars().any(is_whitespace)` instead

kdy1 (changes requested):

crates/swc_html_minifier/src/lib.rs
alexander-akait: @kdy1 Done :+1:

alexander-akait:
@kdy1 Rebased

PR #4921, opened 2022-06-10
perf(css/parser): improve lexer perf

alexander-akait:
**Description:**

Here
- benches for lexer/parser
- Implement `Iter` for `TokensInput` and for `Lexer` (now we can remove `next` method in favor iter, I will do it in next PR)
- improve perf 40-50% (based on criterion)

**BREAKING CHANGE:**

No

No


crates/swc_css_parser/src/parser/input.rs
```diff
@@ -6,7 +6,7 @@ use swc_css_ast::{Token, TokenAndSpan, Tokens};
 use super::PResult;
 use crate::error::{Error, ErrorKind};
 
-pub trait ParserInput {
+pub trait ParserInput: Clone + Iterator<Item = TokenAndSpan> {
     type State: Debug;
 
     fn next(&mut self) -> PResult<TokenAndSpan>;
```
kdy1, line 12: I think we can safely remove this method

crates/swc_css_parser/src/parser/input.rs
```diff
@@ -109,7 +109,7 @@ where
         }
 
         if self.cur.is_none() {
-            let res = self.input.next();
+            let res = ParserInput::next(&mut self.input);
```
kdy1, line 112: So we don't need fully-qualified syntax.

crates/swc_css_parser/src/parser/input.rs
```diff
@@ -6,7 +6,7 @@ use swc_css_ast::{Token, TokenAndSpan, Tokens};
 use super::PResult;
 use crate::error::{Error, ErrorKind};
 
-pub trait ParserInput {
+pub trait ParserInput: Clone + Iterator<Item = TokenAndSpan> {
```
kdy1, line 9: Extending `Iterator` is enough

kdy1 (changes requested):

crates/swc_css_parser/src/parser/input.rs
alexander-akait: @kdy yes, I write it above, I just want to remove it in other PR + do buffer refactor for perf reasons, so I postpone it

crates/swc_css_parser/src/parser/input.rs
alexander-akait: Done

crates/swc_css_parser/src/lexer/mod.rs
```diff
@@ -31,13 +35,94 @@ where
...
         let start_pos = input.last_pos();
+
         Lexer {
             input,
             cur: None,
             cur_pos: start_pos,
             start_pos,
             last_pos: None,
             config,
+            buf: Rc::new(RefCell::new(String::with_capacity(256))),
```
kdy1, line 46: Do we need `Rc` here?

kdy1 (approved):
Thank you!

---

swc-bump:
 - ''

PR #4925, opened 2022-06-10
fix(es/minifier): Reserve more symbols on `eval`

kdy1:
**Description:**
image

`j` points other variable after mangling.

 - `https://github.com/vercel/next.js/discussions/30237#discussioncomment-2893721`


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #4932, opened 2022-06-10
feat(es/minifier): new Array

magic-akari:
**Description:**

**BREAKING CHANGE:**


crates/swc_ecma_minifier/src/compress/pure/array.rs
```diff
@@ -0,0 +1,28 @@
...
+impl Pure<'_> {
+    /// new Array(...) -> Array(...)
+    pub(super) fn remove_array_new(&mut self, e: &mut Expr) {
+        match e {
+            Expr::New(NewExpr {
+                span,
+                callee,
+                args,
+                type_args,
+            }) if callee.as_ident().map(|x| &x.sym) == Some(&js_word!("Array")) => {
```
ArturAralin, line 17: Possibly here need to check ident syntax context equality to unresolved_syntax_context, to prevent remove new form this case

```javascript
class Array {}

new Array(); // here Array is not a global Array object
```

magic-akari: Can't figure out why I'm getting the wrong result.
```JavaScript
new Array();
```

```JavaScript
class Array {}
new Array();
```

In both cases I get Mark(0) as their syntax context.
But the unresolved is Mark(1).

@kdy1 
Is there something wrong with the test case, or with my code?

```rust
span.ctxt.outer() == self.marks.unresolved_mark
span.ctxt == self.expr_ctx.unresolved_ctxt
```
I have tried both of the above methods. No luck so far.

crates/swc_ecma_minifier/src/compress/pure/misc.rs
```diff
@@ -230,6 +230,49 @@ impl Pure<'_> {
...
+                    // https://262.ecma-international.org/12.0/#sec-array-constructor
+                    "Array",
+                    // https://262.ecma-international.org/12.0/#sec-function-constructor
+                    "Function",
+                    // https://262.ecma-international.org/12.0/#sec-regexp-constructor
+                    "RegExp",
+                    // https://262.ecma-international.org/12.0/#sec-error-constructor
+                    "Error",
+                    // https://262.ecma-international.org/12.0/#sec-aggregate-error-constructor
+                    "AggregateError",
```
Austaras, line 257: Object too

Austaras: And [native errors](https://tc39.es/ecma262/#sec-native-error-types-used-in-this-standard)

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_utils
 - swc_ecma_minifier

PR #4933, opened 2022-06-10
feat(es/lint): support ignoreReadBeforeAssign for prefer-const rule

ArturAralin:
**Description:**

support ignoreReadBeforeAssign for prefer-const rule

**BREAKING CHANGE:**
No


crates/swc_ecma_lints/src/rules/prefer_const.rs
```diff
@@ -11,15 +12,21 @@ use crate::{
 // todo: implement option destructuring: all | any
 // https://eslint.org/docs/rules/prefer-const#destructuring
 
-// todo: implement option ignoreReadBeforeAssign
-// https://eslint.org/docs/rules/prefer-const#ignorereadbeforeassign
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PreferConstConfig {
+    ignore_read_before_assign: Option<bool>,
```
kdy1, line 18: I think we need `#[serde(default)]`

kdy1: And it can be `bool` instead of `Option<bool>`.

kdy1 (changes requested):
Thank you!

crates/swc_ecma_lints/src/rules/prefer_const.rs
ArturAralin: Sorry for late fix. I was busy

kdy1:
Sorry for the delay.
We need some more discussion about linter

ArturAralin:
> We need some more discussion about linter

With me?

kdy1:
Ah, my bad. 
I meant the core team

ArturAralin:
@kdy1 have any solution?

kdy1 (approved):
Thank you and sorry for the dealy

---



swc-bump:
 - swc_ecma_lints

PR #4936, opened 2022-06-10
feat(es/lint): implement no-await-in-loop rule

ArturAralin:
**Description:**

no-await-in-loop rule implementation

**BREAKING CHANGE:**
No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_lints

PR #4943, opened 2022-06-11
Don't use `get_unchecked` after having set the len to 0

Nilstrieb:
**Description:**

This code first sets the `len` to 0 in case of panics, and then uses `get_unchecked` to access the elements in the vec.
This is not allowed, since `get_unchecked` must only be used with indexes that are in bounds respective to the length.
Use `as_ptr().add()` instead, which does not care about the length.

Found by debug assertions in `std`  when using `-Zbuild-std`.


kdy1 (approved):
Thank you!

---


swc-bump:
 - swc_common

PR #4945, opened 2022-06-11
perf(common): Add `Atom` type

kdy1:
**Description:**

See https://github.com/swc-project/swc/issues/4946 for the context.


kdy1 (commented):
swc-bump:
 - swc_atoms

PR #4947, opened 2022-06-11
fix(es/minifier): turn `Array` `Object` call into literal

Austaras:
**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/pure/misc.rs
+++ b/crates/swc_ecma_minifier/src/compress/pure/misc.rs
@@ -230,11 +230,171 @@ impl Pure<'_> {
         }
     }
 
+    /// `new RegExp("([Sap]+)", "ig")` => `/([Sap]+)/gi`
+    fn optimize_regex(&mut self, args: &mut Vec<ExprOrSpread>, span: &mut Span) -> Option<Expr> {
+        if args.is_empty() || args.len() > 2 {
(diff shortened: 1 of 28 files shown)


crates/swc_ecma_minifier/src/compress/pure/misc.rs
```diff
@@ -230,12 +230,172 @@ impl Pure<'_> {
...
+        })))
+    }
+
+    /// Array() -> []
+    fn optimize_array(&mut self, args: &mut Vec<ExprOrSpread>, span: &mut Span) -> Option<Expr> {
+        if args.len() == 1 {
+            if let ExprOrSpread { spread: None, expr } = &args[0] {
+                match &**expr {
+                    Expr::Lit(Lit::Num(num)) => {
+                        if num.value <= 11_f64 && num.value >= 0_f64 {
```
magic-akari, line 312: why `11`?
It seems that 0 ~ 6 is reasonable.

Austaras: https://github.com/terser/terser/blob/master/lib/compress/index.js#L2086

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_ast
 - swc_ecma_codegen
 - swc_ecma_minifier

PR #4948, opened 2022-06-11
fix(css/parser): buffer and recovery

alexander-akait:
**Description:**

Here:
1. Migrate on `Iter`
2. Improve recovery mode fro broken comments
3. Avoid emit error on `// Comment` content

We need refactor tests, I will do it after this

**BREAKING CHANGE:**

No

No


crates/swc_css_parser/src/parser/input.rs
```diff
@@ -206,21 +198,23 @@ impl<'a> ParserInput for TokensInput<'a> {
     fn reset(&mut self, state: &Self::State) {
         self.idx = state.idx;
     }
+
+    fn take_errors(&mut self) -> Vec<Error> {
+        vec![]
```
alexander-akait, line 203: Because TokensInput can have lexer errors, we don't parse source code here

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_parser --breaking

PR #4949, opened 2022-06-11
refactor(html/parser): remove unused code

alexander-akait:
**Description:**

We don't peek somthing in html, so we don't need extra chech

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser

PR #4950, opened 2022-06-11
feat(html/parser): self-closing flag

alexander-akait:
**Description:**

Implement `is_self_closing` for better compress SVG and MathML syntax in HTML.

Note - `is_self_closing` is `true` only when you have slash at end, i.e. `<br />`, some parsers emits `<br>` with always `is_self_closing: true`, but it is wrong, `br` is void element, self closing is XHTML syntax https://dev.w3.org/html5/html-author/#self-closing-tag

Many changes due updating snapshots

**BREAKING CHANGE:**

Yes

No


alexander-akait:
@kdy1 Rebased and fixed

kdy1 (approved):
Thanks!

---


swc-bump:
 - swc_html_ast --breaking

PR #4952, opened 2022-06-13
perf(es/hygiene): Make `hygiene` faster

kdy1:
**Description:**

We don't use multi-thread because it's slower.


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base
 - swc_ecma_minifier

PR #4954, opened 2022-06-13
feat: support css-modules select-text parse and codegen 

zhusjfaker:
feat: support css-modules select-text parse and codegen 

**Description:**

css-modules include  ":local and :global" features  is important for modern front-end project. css-loader actually used
"post-css", now I ported the feature and tests with swc, this is a hard work for review. Thanks for your work, hope get feedback :)

**BREAKING CHANGE:**

nothing


crates/swc_css_modules/tests/fixture.rs
```diff
@@ -0,0 +1,170 @@
...
+    let target_code = r#"
+.x .fixture_demo3_16401825995056589139_b > .fixture_demo3_16401825995056589139_c  {
+  width: 20px;
+}
+  "#;
+    assert_eq!(
+        simple_compare(target_code.to_string()),
+        simple_compare(res3)
+    );
+}
```
alexander-akait, line 170: Let's avoid inline tests and use snapshots here, because it will be hard to maintance (if we improve or change logic we need manually change all hashes, it is really uncomfortable)

crates/swc_css_modules/src/hash.rs
```diff
@@ -0,0 +1,40 @@
...
+    fn generate_hash_by_content(content: &str) -> String {
+        let mut hasher = DefaultHasher::new();
+        content.hash(&mut hasher);
+        hasher.finish().to_string()
+    }
+
+    ///
+    /// Generate the prefix corresponding to css_modules according to the css
+    /// file path and file content
+    pub fn generate_css_module_hash(abs_filepath: &str, content: &str) -> String {
```
alexander-akait, line 21: We should not use `content` - it is bad for perf and hash should not be related to content (if I add `color: red` declaration we should not change hash)

crates/swc_css_modules/src/hash.rs
```diff
@@ -0,0 +1,40 @@
...
+                perfix += "_";
+            }
+        }
+        perfix += path
+            .file_stem()
+            .unwrap()
+            .to_str()
+            .unwrap()
+            .replace('.', "_")
+            .as_str();
```
alexander-akait, line 36: We will have problems on windows here, we should change `\\` on `/`

crates/swc_css_modules/src/hash.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// file path and file content
+    pub fn generate_css_module_hash(abs_filepath: &str, content: &str) -> String {
+        let path = Path::new(abs_filepath);
+        let mut perfix = "".to_string();
+        if let Some(parent_path) = path.parent() {
+            if let Some(parent_dir) = parent_path.file_name() {
+                perfix += parent_dir.to_str().unwrap();
+                perfix += "_";
+            }
+        }
```
alexander-akait, line 29: We need the `context` option here, we can't know developer context

crates/swc_css_modules/src/css_modules.rs
```diff
@@ -0,0 +1,266 @@
...
+            res = value.to_string();
+        } else {
+            return Err("convert token has no matching correct type".to_string());
+        }
+        if !res.is_empty() {
+            Ok(res)
+        } else {
+            Err("convert token has error".to_string())
+        }
+    }
```
alexander-akait, line 70: We should avoid it, better to add the `css_modules` option for CSS parser and parse `:global`/`:local` content as selectors, it will be faster and more easy

crates/swc_css_modules/src/css_modules.rs
```diff
@@ -0,0 +1,266 @@
...
+        let lexer = Lexer::new(SourceFileInput::from(&*fm), config);
+        let mut parser = Parser::new(lexer, config);
+        let stylesheet = parser.parse_all().unwrap();
+        let mut visitor = ChildCssModuleComponent {
+            selector_list: None,
+        };
+        stylesheet.fold_with(&mut visitor);
+        let selector_list = visitor.selector_list.unwrap();
+        selector_list.children.get(0).unwrap().to_owned()
+    }
```
alexander-akait, line 88: It is very bad for perf, we should never do it

alexander-akait (changes requested):
Let's do it step by step:

1. Parser improvements - parser content of `:local`/`:global` (let's add the `css_modules` option for our parser)
2. When we should port https://github.com/css-modules/postcss-modules-local-by-default/ (i.e. add `:local` where it necessary to avoid extra walking classes/id/attribute/et)
3. https://github.com/css-modules/postcss-modules-scope we need to handle `@keyframes` too https://github.com/css-modules/postcss-modules-scope/blob/master/src/index.js#L270

crates/swc_css_modules/src/css_modules.rs
zhusjfaker: sorry I don't know if swc can  provide me with a method to parse the string in global () directly into ast instead of str

zhusjfaker: :global(.a)  -> complex_selector

alexander-akait: @zhusjfaker We just add the `css_modules` option, I can do it today

crates/swc_css_parser/src/parser/selector/mod.rs
```diff
@@ -842,6 +841,11 @@ where
 
                             self.input.skip_ws()?;
                         }
+                        "global" | "local" => {
+                            self.input.skip_ws()?;
+                            let selector_list = self.parse()?;
+                            children.push(PseudoClassSelectorChildren::SelectorList(selector_list));
```
alexander-akait, line 847: I don't think we should use `SelectorList` here

alexander-akait:
Implemented, look at `swc_css_modules` crate and the `css_modules` option for parser, anyway thank you for the PR and sorry for delay

PR #4958, opened 2022-06-13
feat: add the `remove_empty_attributes` option

alexander-akait:
**Description:**

Added on comment - shorty - some very old/legacy libraries don't use DOM API and parser attributes from element using RegExps, so because empty attributes were removed regexp can be failed, so libraries will not work, anyway for modern application, it is safe to remove, so we keep it `true` by default

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_minifier

PR #4960, opened 2022-06-13
feat: add the `collapse_boolean_attributes` option

alexander-akait:
**Description:**

Collapsing attributes can be unsafe in some cases, ref http://perfectionkills.com/experimenting-with-html-minifier/#collapse_boolean_attributes, but I don't think it is good practice to write boolean attrbiutes in this style, so `true` by default

**BREAKING CHANGE:**

Yes

No


kdy1 (dismissed):
Thanks!

---
swc-bump:

alexander-akait:
@kdy1 Rebased, fixed

kdy1 (approved):

PR #4962, opened 2022-06-13
fix(es/helpers): Allow using external helpers with `type: module`

### LINKED ISSUE #4854: external helpers cannot be imported from projects with `type: "module"`
### Describe the bug

[This PR](https://github.com/swc-project/swc/pull/4767) changed the way how external helpers are imported, to import only the used external helpers, instead of the entire namespace, as described in this issue: https://github.com/swc-project/swc/issues/2745.

This causes a runtime bug, for example when using a rest spread (which uses the `_object_without_properties` helper). I'm not sure whether other helpers are affected, too, but I could reproduce this one. 

It compiles without complaints, but running the compiled file causes.

```
TypeError: _object_without_properties is not a function
    at file:///home/runner/SWC-helpers-import-broken/dist/index.js:6:29
    at ModuleJob.run (node:internal/modules/esm/module_job:185:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:281:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12)
```

This is fully **reproduced** here: https://replit.com/@nickbluey/SWC-helpers-import-broken#src/index.ts. Please fork the thing and run `npm run build` and `npm run start` to see the error message. (Not using SWC playground as it has no helpers and also this is a runtime issue).

I assume this has something to do with ESM/CJS interoperability, as the code in `@swc/helpers/lib_objectWithoutProperties.js` is CJS, but the consumer is ESM, and the exporting code is exporting like `exports.default = _objectWithoutProperties;`.

### Input code

```typescript
const obj = {
  a: 1,
  b: 2
};

const { a, ...other } = obj;
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript"
    },
    "target": "es2015",
    "externalHelpers": true
  }
}
```

### Playground link

_No response_

### Expected behavior

The generated code `import _object_without_properties from "@swc/helpers/lib/_object_without_properties.js";` should resolve to the function.

### Actual behavior

See error above.

### Version

v1.2.196

### Additional context

_No response_

kdy1:
**Description:**

We use `.mjs` for imports, and `common_js` pass modifies `.mjs` imports to common js requires with patched path

 - Closes #4854


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base
 - swc_ecma_transforms_compat

PR #4969, opened 2022-06-14
refactor(es/minifier): Use `Program` instead of `Module`

kdy1:
**Description:**

cc @alexander-akait 


kdy1 (commented):
swc-bump:

PR #4971, opened 2022-06-14
feat(html/codegen): added tag omission and self closing void elements options

alexander-akait:
**Description:**

Why we need the `tag_omission` option?

Not all robots/services/libraries/etc support tag omission, so we should provide an option to disable it

Why we need the `self_closing_void_elements` option?

Allow to print legacy code (i.r. not HTML5, although mot of modern browsers uses HTML5 spec to parse HTML)

By default `tag_omission` is `true` and `self_closing_void_elements` is `false` when `minify` is `true`. That is why we have `Option<bool>` type.

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_codegen --breaking

PR #4973, opened 2022-06-14
feat(html/minifier): compress CSS

alexander-akait:
**Description:**

Minify CSS.

We should minify `style` attribute too, but we need improve CSS parser, I will improve it

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_codegen

Issue #4953, opened 2022-06-13
swc minify bug with function defined in another function


wre232114:
### Describe the bug

see the code below, `function t()` is defined in another function which also has a parameter named `t`. swc minify removed the `function t()` which is not expected.

### Input code

```typescript
/* eslint-disable */
export default function(module, exports) {
  "use strict";
  !function(t, e) {
    e(exports)
  }(void 0, function(t) {
      "use strict";
      function vr(e, r, o) {
        try {
        } catch (t) {
        }
        function t() {
            return '123';
        }
        return t;
      }
      t.vr=vr;
  });
};
```

### Playground link

[...]

PR #4976, opened 2022-06-15
fix(es/resolver): Make catch parameter not disturb resolving

kdy1:
**Description:**

 - Closes https://github.com/swc-project/swc/issues/4953


kdy1 (commented):
swc-bump:
 - swc_ecma_transform_base

PR #4978, opened 2022-06-15
Fix module field in @swc/helpers package.json

devongovett:
Should have been updated to use `.mjs`.


kdy1 (approved):
Thank you!

---

swc-bump:
 - ''

PR #4979, opened 2022-06-15
Use .mjs extension in helper imports

devongovett:
The imports within the helpers themselves must also have extensions for them to work in node. You can test this by running `await import('@swc/helpers/src/index.mjs')` in a node repl.

One problem with the current approach of replacing `@swc/helpers/src/___.mjs` with `@swc/helpers/lib/___.js` in the CommonJS pass is that these relative imports inside the helpers won't be affected. Any idea on that @kdy1?


kdy1:
Hmm... Relative imports would be fine I guess?
If we load a common js one, it will use `required` for loading deps and esm one will use esm syntax for loading deps. Am I missing something?

devongovett:
Yeah so I guess we need to add something in the script that builds the SWC helpers package itself so the commonjs version doesn't have .mjs in its compiled requires.

kdy1:
Oh I see
Yeah you are right

devongovett:
Ok did it with a simple sed script. Hopefully that works ok.

devongovett:
Might be good to have a simple smoke test for the helpers too, like 

```js
await import('@swc/helpers/src/index.mjs') // can import
require('@swc/helpers/lib/index.js') // can require
```

but not sure where to put it

kdy1 (dismissed):
Thank you!

---

swc-bump:
 - ''

kdy1:
Oh smoke test would be useful.
We had one, but it was removed because we need to update helper package without publishing sometimes.

https://github.com/swc-project/swc/blob/4b340e2c1bfa00f19be5aecceb5c337487030046/crates/swc/tests/exec.rs#L85

This line is the problem, but I think it's better to have external helpers tested

kdy1:
I'll add one tomorrow

kdy1 (dismissed):

kdy1 (approved):

kdy1 (dismissed):

PR #4980, opened 2022-06-15
feat(html/minifier): option to set html5 doctype

alexander-akait:
**Description:**

Add option to force set HTML5 doctype, good for legacy systems, ported from https://terser.org/html-minifier-terser/

**BREAKING CHANGE:**

Yes

No

### DIFF
--- a/crates/swc_html_minifier/src/option.rs
+++ b/crates/swc_html_minifier/src/option.rs
@@ -5,7 +5,7 @@ use swc_cached::regex::CachedRegex;
 #[serde(rename_all = "camelCase")]
 #[serde(deny_unknown_fields)]
 pub struct MinifyOptions {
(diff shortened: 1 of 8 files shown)


crates/swc_html_minifier/src/option.rs
```diff
@@ -4,6 +4,8 @@ use serde::{Deserialize, Serialize};
 #[serde(rename_all = "camelCase")]
 #[serde(deny_unknown_fields)]
 pub struct MinifyOptions {
+    #[serde(default = "false_by_default")]
```
kdy1, line 7: We don't need this. `#[serde(default)]` is enough

kdy1 (changes requested):

alexander-akait:
@kdy Fixed

PR #4983, opened 2022-06-16
test(html): Remove files which are too large too verify

kdy1:
**Description:**

I consider this as a security risk.
cc @alexander-akait 
Let's use smaller files


kdy1 (commented):
swc-bump:
 - ''

PR #4986, opened 2022-06-17
fix(html/codegen): optional tags

alexander-akait:
**Description:**

Fixes:
- p with legacy element
- li only for menu/ol/li

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!


---

swc-bump:
 - swc_html_codegen

PR #4992, opened 2022-06-18
fix(es/parser): Fix bin expr parsed as instantiation expr

g-plane:
**Description:**

This PR fixes binary expression parsed as instantiation expression.

Input:

```ts
a<b>>c<d>
```

Before:

```ts
a > c
```

After:

```ts
a < b >> c
```

Note that expressions like `f<x> & g<y>` aren't valid any more, as [TypeScript does](https://www.typescriptlang.org/play?#code/GYHgHgfABAZFDmICeEg).

**BREAKING CHANGE:**

No.

**Related issue (if exists):**

Fixes #4990

PR #5000, opened 2022-06-19
fix(es/parser): Instantiation expr can be followed by line break or binary operator

g-plane:
**Description:**

Allow the code below:

```ts
const x2 = f<true>  // becomes instantiation expression
true;

f<x> && g<y>;
```

**BREAKING CHANGE:**

No.

- #4992 
- https://github.com/microsoft/TypeScript/pull/49353


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_parser

PR #5003, opened 2022-06-20
Some general refactoring in swc_cli

bcmyers:
**Description:**

Hi all. First time contributor here. 

I saw @kwonoj 's comment in this [issue](https://github.com/swc-project/swc/issues/1589#issuecomment-1054774477) and became interested in the port of `swc_cli` to rust. I had a couple of free hours today and so just went through trying to familiarize myself with the existing code a bit and perform a few changes. 

I mostly focused on the `compile` command.

I noticed a couple of small improvement opportunities if you guys like them:

* Avoid a few unnecessary allocations in places (You can get away with keeping pointers to paths in several places where before you were allocating new `PathBuf`s)
* Using buffered readers and writers when trying to read/write to files (this should be marginally faster)
* Generally adding a little more context to error messages
* Some small nits like changing `&Option<PathBuf>` in function signatures to what I think is the slightly more idiomatic `Option<&Path>`.
* A little cleanup of some control flow here and there

At some point, I'd like to try my hand at adding some new code fill in more of the missing pieces of the cli, but, today, I just started with some general tweaks to what's already there.

Looking forward to seeing if you're interested in any of these changes.

Best,
Brian

**BREAKING CHANGE:**

Previously, the user was only able to provide as command line arguments for `files`:
* One directory, or
* One or more files

I changed this so it is now OK for the user to provide,
* One or more directories, and/or
* One or more files

I'm not sure if you'll want this API change, but it's pretty easy to back out if you want to keep the existing semantics.

#1589

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -380,33 +380,42 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "3.1.0"
+version = "3.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e5f1fea81f183005ced9e59cdb01737ef2423956dac5a6d731b06b2ecfaa3467"
+checksum = "d53da17d37dba964b9b3ecb5c5a1f193a2762c700e6829201e645b9381c99dc7"
 dependencies = [
  "atty",
  "bitflags",
  "clap_derive",
+ "clap_lex",
  "indexmap",
- "lazy_static",
- "os_str_bytes",
+ "once_cell",
  "strsim",
  "termcolor",
  "terminal_size",
- "textwrap 0.14.2",
+ "textwrap 0.15.0",
 ]
 
 [[package]]
 name = "clap_derive"
-version = "3.1.0"
+version = "3.2.5"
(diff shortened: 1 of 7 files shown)


kwonoj:
> I changed this so it is now OK for the user to provide,

I'd like to request to take out these changes, for 2 reasons:

- First, this does not belong to the main purpose of PR to refactor existing implementations
- Secondly, there are deliberate decisions to not allow changed behavior due to confusions around base path calculation among multiple inputs, which need to be addressed in PR correctly.

bcmyers:
> > I changed this so it is now OK for the user to provide,
> 
> I'd like to request to take out these changes, for 2 reasons:
> 
>     * First, this does not belong to the main purpose of PR to refactor existing implementations
> 
>     * Secondly, there are deliberate decisions to not allow changed behavior due to confusions around base path calculation among multiple inputs, which need to be addressed in PR correctly.

Fair enough. Change in behavior removed. Now the old behavior is retained, i.e. user receives an error if they pass a directory plus 1 or more other paths.

PR #5004, opened 2022-06-20
feat(html/minifier): improve smart mode

alexander-akait:
**Description:**

Compress whitespaces better, especial in smart mode

Smart mode is almost finished, remains:
- we need to add the comment support to change display, like do prettier, i.e. `<-- block -->`/etc
- compress whitespaces for really hidden elements (i.e. `meta`/`link` and etc)

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---
swc-bump:
 - swc_ecma_minifier

PR #5011, opened 2022-06-20
fix(es/lint): allow duplicate function in non top level in module

### LINKED ISSUE #4907: False positives when checking for duplicate definitions

Austaras:
**Description:**

So, [per spec](https://tc39.es/ecma262/#sec-static-semantics-lexicallydeclarednames)

> At the top level of a [Module](https://tc39.es/ecma262/#prod-Module), function declarations are treated like lexical declarations rather than like var declarations.

and it's only an error at the top level of module

**BREAKING CHANGE:**

closes #4907


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_lints

PR #5014, opened 2022-06-20
test(es/parser): Add a test case for instantiation expr

g-plane:
**Description:**

Added a test case about instantiation expression.

- https://github.com/microsoft/TypeScript/issues/49551


kdy1 (approved):
Thanks!

---
swc-bump:
 - ''

PR #5016, opened 2022-06-21
fix(html/minifier): avoid optional attribute

alexander-akait:
**Description:**

Small tweak

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!


---

swc-bump:
 - swc_html_minifier

PR #5025, opened 2022-06-22
fix(css/parser): lexer errors

alexander-akait:
**Description:**

Add lexer errors for CSS parser.

We need to sort them in test because when we consume using tokens we run the lexer firstly and then the parser, but when we parser using string we lazy consume tokens, so errors have different order.

**BREAKING CHANGE:**

No

No


crates/swc_css_parser/src/lexer/mod.rs
```diff
@@ -198,7 +198,15 @@ where
         self.input.reset_to(self.cur_pos);
     }
 
-    fn read_token(&mut self) -> LexResult<Token> {
+    #[inline(always)]
```
kdy1, line 201: Can you remove this and instead add `#[cold]`?

kdy1 (changes requested):
Thank you!
`#[inline(always)]` makes debug build slower, and for error reporting functions, `#[cold]` is better

alexander-akait:
@kdy1 Fixed, also fixed `for swc_html_parser`

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_css_parser

PR #5031, opened 2022-06-24
feat(html/parser): improve API

alexander-akait:
**Description:**

- Allow to setup document and form element for document fragments
- Context element passed by reference, so no need to clone

I tried to extract tree builder as in TODO, but found less perf and more memory usage, because we need duplicate some things, so I think for perf reasons better to keep parser and tree builder together

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_codegen_macros --breaking

PR #5045, opened 2022-06-27
feat(html/minifier): add `normalize_attributes` to disable normalization

alexander-akait:
**Description:**

Allow to keep values as is

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

---

swc-bump:

PR #5061, opened 2022-06-28
Fix changelog script

stephenisau:
**Description:**
small typo!
Screen Shot 2022-06-28 at 2 47 59 PM

**BREAKING CHANGE:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     },
     "types": "./index.d.ts",
     "scripts": {
-        "changelog": "git cliff --output CHANGELOG.md",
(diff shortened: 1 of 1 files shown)


kwonoj:
Thanks, but it is not a typo:

We use https://github.com/orhun/git-cliff for the changelog generation, which enables `cliff` subcommand for git.
See examples with cliff installed: https://github.com/orhun/git-cliff#examples= .

kdy1:
It's not typo.
https://github.com/orhun/git-cliff

Issue #5060, opened 2022-06-28
Wasm plugins are not backward compatible


kwonoj:
### Describe the feature

This is by far the biggest blocker for the official release for the plugin feature. (I thought I filed this already somehow).

The main issue in here is reducing chances of breaking changes for the plugin. There are 3 possible cases which causes diverge between plugins to the host for serialization / deserialization, namely

- adding a new property into an existing ast struct
- removing a new property from an existing ast struct
- rename existing property from an ast struct

Each time this happens, the plugin will not be able to run on the newer versions of swc include those changes. These kind of breaking change itself is unavoidable unfortunately, but also we may be able to reduce the chances of it or at least avoid it happening unexpectedly.

We'll attempt to follow `versioned` struct approach suggested by underlying serialization mechanism (https://github.com/rkyv/rkyv/blob/master/examples/backwards_compat/src/main.rs) with compiler flags. When we need a possible breaking changes it'll be guarded by flags, providing backward compatibility as much as it can. Still worth mentioning this only applies to `adding` part - removing, or renaming is a clear breaking change we can't avoid to request plugin bumps up with the latest changes. At least, we hope those kind of operation won't happen that frequently.

### Additional context

https://github.com/swc-project/swc/discussions/3540

thesayyn:
FWIW, [capn](https://capnproto.org/) solves this is really good way, there's not really a good way to solve this problem other than inventing a clever encoding mechanism like this.

RobinClowers:
I'm just trying out SWC and this issue is pretty painful for my team. I was excited to find https://github.com/pmndrs/swc-jotai, but then discovered they are pretty far behind on the core version, so to get everything working, I would have to manually downgrade things until I find a set of versions that work together.

frattaro:
so if the issue with backwards compatibility is solved, does that mean swc plugins will no longer be under `experimental` in next?

kdy1:
Nope. Wasm plugins are still experimental

kdy1:
It'll be much more backward compatible with `v1.15.0`.

https://blog.swc.rs/2025-11-4-wasm-backward-compatibility

PR #5062, opened 2022-06-28
feat(swc/common): add versioned wrapper struct (#5060: Part 1)

kwonoj:
**Description:**

Related to #5060.

PR introduces a new struct `versioned`, which is a transparent type can be used to carry over different versioned struct. Also PR wraps ast with `bytecheck` to validate serialized byte with each versions.

**BREAKING CHANGE:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2994,6 +2994,7 @@ dependencies = [
 name = "swc_atoms"
 version = "0.2.12"
 dependencies = [
+ "bytecheck",
  "once_cell",
  "rkyv",
  "rustc-hash",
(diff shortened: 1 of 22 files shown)


crates/ast_node/Cargo.toml
```diff
@@ -13,6 +13,7 @@ bench = false
 proc-macro = true
 
 [dependencies]
+bytecheck = "0.6.8"
```
kdy1, line 16: `ast_node` is proc macro, so I don't think we need this

kwonoj: 👍

kdy1 (approved):
Thank you!

---


swc-bump:
 - swc_atoms
 - swc_common
 - ast_node --breaking

PR #5064, opened 2022-06-29
perf(atoms): More groundwork for replacing `JsWord` with `Atom`

kdy1:
**Description:**

**BREAKING CHANGE:**


kdy1 (commented):
swc-bump:
 - swc_atoms
 - ast_node --breaking

PR #5065, opened 2022-06-29
fix(es/module): handle this in class method

### LINKED ISSUE #5063: `this` in class method default parameter transpiles to `(void 0)`
### Describe the bug

Since `1.2.206`, when `module` is not set to `es6`, if a class with a method that has a default parameter references a class field with `this`, that reference will be transpiled to `(void 0)`, yielding an error when the code is run.

`this` is correctly retained when `module` is set to `es6` but is transpiled to `(void 0)` for all other values of `module`. Prior to `1.2.206`, `this` is retained regardless of the value of `module`.

### Input code

```typescript
class Foo {
    bar = 5
    getThing(a, b = this.bar) {
        return a + b;
    }
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2022",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.207&code=H4sIAAAAAAAAA0vOSSwuVnDLz1eo5lIAgqTEIgVbBVMwOz21JCQjMy9dI1FHIQkoWpKRWawHVKAJVQsCRaklpUV5CokK2gpJ1mDRWq5aLgCgdCWCVgAAAA%3D%3D&config=H4sIAAAAAAAAA0WMTQrEIAyF75K1i8Fl79BDBCctFv9IUqiIdx8Fh64S3vve1%2BASB1uDgizE85OaFB%2FYgFxEceyLghnYiA4MQt2AIp%2BkExH7sXbUIWehBRiIPvmjTpnLsTCJvBWmM%2FzJPlwxf%2B8ZNNBaxp2TmNMl0F%2FRGnvZF618U%2F8BElOxt70AAAA%3D

### Expected behavior

The transpiled version of the above input code should retain the reference to `this`:

```js
"use strict";
class Foo {
    bar = 5;
    getThing(a, b = this.bar) {
        return a + b;
    }
}
```

### Actual behavior

The reference to `this` is transpiled to `(void 0)`, such that an error is yielded if the method is called:

```js
"use strict";
class Foo {
    bar = 5;
    getThing(a, b = (void 0).bar) {
        return a + b;
    }
}
```

### Version

1.2.207

### Additional context

On the [the MDN page for default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters), one of the examples uses `this` in a default parameter position -- I don't think it's rare to do this.

magic-akari:
**Description:**

**BREAKING CHANGE:**

- fix #5063

### DIFF
--- a/crates/swc_ecma_transforms_module/src/module_ref_rewriter.rs
+++ b/crates/swc_ecma_transforms_module/src/module_ref_rewriter.rs
@@ -99,15 +99,11 @@ impl VisitMut for ModuleRefRewriter {
     }
 
     fn visit_mut_function(&mut self, n: &mut Function) {
-        n.params.visit_mut_with(self);
-
-        self.visit_mut_with_non_global_this(&mut n.body);
+        self.visit_mut_with_non_global_this(n);
     }
 
     fn visit_mut_constructor(&mut self, n: &mut Constructor) {
-        n.params.visit_mut_with(self);
-
-        self.visit_mut_with_non_global_this(&mut n.body);
+        self.visit_mut_with_non_global_this(n);
     }
 
     fn visit_mut_class_prop(&mut self, n: &mut ClassProp) {
--- a/crates/swc_ecma_transforms_module/src/module_ref_rewriter.rs
+++ b/crates/swc_ecma_transforms_module/src/module_ref_rewriter.rs
@@ -35,7 +35,7 @@ pub(crate) struct ModuleRefRewriter {
 
     pub lazy_record: AHashSet<Id>,
 
-    pub top_level: bool,
+    pub is_global_this: bool,
 }
 
 impl VisitMut for ModuleRefRewriter {
@@ -67,7 +67,7 @@ impl VisitMut for ModuleRefRewriter {
             }
 
             Expr::This(ThisExpr { span }) => {
-                if self.top_level {
+                if self.is_global_this {
                     *n = *undefined(*span);
                 }
             }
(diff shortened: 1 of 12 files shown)


crates/swc_ecma_transforms_module/src/module_ref_rewriter.rs
```diff
@@ -101,36 +101,42 @@ impl VisitMut for ModuleRefRewriter {
     fn visit_mut_function(&mut self, n: &mut Function) {
         n.params.visit_mut_with(self);
```
Austaras, line 102: param is not top level.
[babel](https://babeljs.io/repl#?browsers=chrome%2055&build=&builtIns=entry&corejs=3.6&spec=false&loose=false&code_lz=KYDwDg9gTgLgBAMwK4DsDGMCWEWIhACgEM4BeOGAC0wGcBKOAbzRxogBtgA6diAc2J0AvgCggA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.18.7&externalPlugins=&assumptions=%7B%7D)

magic-akari: Thanks

Austaras: visit_mut_class_method could be removed now.

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_module

PR #5069, opened 2022-06-29
feat(es/module): support node interop

magic-akari:
**Description:**

follow Babel's `importInterop`
- https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs#importinterop

**BREAKING CHANGE:**


kdy1 (commented):
Do we need `interop: "swc"`?
I know babel has one, but I'm not sure about adding one more brand.

(Really, just not sure, and this is not a change request)

magic-akari:
It is the default option. 
In most cases, the user does not need to write it explicitly when selecting this option.
Anyway, we need to name it.
It would be a little weird if we used `babel` as default name.

What name do you prefer?

kdy1:
I think `swc` is the best option if so, thank you!

crates/swc_ecma_transforms_module/tests/fixture/common/interop-node/export-from-1/output.cjs
```diff
@@ -0,0 +1,5 @@
+"use strict";
+Object.defineProperty(exports, "__esModule", {
+    value: true
+});
+_exportStar(require("foo"), exports);
```
kdy1, line 5: We don't need interopRequireWildcard here?

magic-akari: No, it's not necessary.

1. `export *` never reexport `default`.
2. import node-esm from node-esm, everything is as same as browser-esm.
3.  import cjs in node-esm, node will wrap all export ident in `default` <- this is different

The only thing we need special handling is importing cjs to node-esm.
We should wrap everything in `default` to mimic Node's behavior.

```JavaScript
// foo.cjs
exports.foo = function foo() {};

// bar.cjs
exports.bar = function bar() {};

// index.js
import foo from "./foo.cjs";
import * as bar from "./bar.cjs";

export * from "./foo.cjs"
export * from "./bar.cjs";

console.log(foo); 
// { foo: [Function: foo] }

console.log(bar);
// [Module: null prototype] {
//     bar: [Function: bar],
//     default: { bar: [Function: bar] }   <- Node will wrap all exports in `default`
//   }
```

But, we never reexport `default`,  and function bar is in both namespace and default.
it's ok just use `_exportStar` and `require` to reexport everything.

Let's see the Babel‘s transforms.
Input:
```JavaScript
import * as foo from "./foo.cjs"; // trigger `_interopRequireWildcard`

export * from "./foo.cjs"
export * from "./bar.cjs";

foo;
```

output:
```JavaScript
"use strict";

  value: true
});

var foo = _interopRequireWildcard(require("./foo.cjs"), true);

Object.keys(foo).forEach(function (key) {
  // omit...
});

var _bar = require("./bar.cjs");

Object.keys(_bar).forEach(function (key) {
// omit...
});

foo;
```

You can see that foo is wrapped by `_interopRequireWildcard` and bar is not, and there is no difference when re-exporting.

kdy1: Thank you for the explanation!

kdy1 (approved):
Thank you!


---

swc-bump:

PR #5071, opened 2022-06-29
feat(swc/common): allow to take inner from versioned (#5060: Part 2) 

kwonoj:
**Description:**

I made this PR intentionally short for the verification of idea (which means I'm not certain about this change and I may wrong 😅)

PR implements `take` for the `VersionedSerializable` to move out value after deserializaiton is done from transform. Pseudocodewise, it'll be used like below:

```
let versioned_program = VersionedSerializable::new(program);
let mut serialized = Serialized::serialize(&versioned_program)?;

//... run plugin transform

// deserialize, return actual inner program
return Serialized::deserialize(&serialized)
                    .map(|mut v: VersionedSerializable<Program>| v.take())
```

Main thing I'm unsure is implementing `Default` for the `Program` to use `mem::take` in here. It maybe harmless, but I'm not easily able to say default impl for the program is exactly right. Maybe 1. either default should not be implemented 2. or implement better default ?

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_common/src/plugin.rs
+++ b/crates/swc_common/src/plugin.rs
@@ -4,7 +4,7 @@
 //! `swc_common`.
 #![allow(unused)]
 
-use std::any::type_name;
+use std::{any::type_name, mem};
 
 use anyhow::Error;
 use bytecheck::CheckBytes;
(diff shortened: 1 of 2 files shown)


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_common
 - swc_ecma_ast

PR #5073, opened 2022-06-30
feat(visit): Add path-aware variants

kdy1:
# TODOs
 - [ ] Add a method to convert node to `AstKind`
 - [ ] Add a method to convert node to `AstNodeRef`
 - [ ] Adjust other codegen parts to use __ast_path 
 - [ ] Add cfg(feature = "path") to items. (I didn’t do this because it makes development harder)
 - [ ] Implement `swc_visit::AstKindPath`
 - [ ] Implement `swc_visit::AstNodePath`


docs/adr/00004-add-ast-path-info-to-visitor.md
```diff
@@ -0,0 +1,47 @@
...
+
+-   Help plugin authors writing plugins.
+
+## Considered Options
+
+-   [option 1] Stay with as-is.
+
+-   [option 2] Provide full information using `unsafe`.
+
+-   **[option 3] Provide small amount of information .**
```
kwonoj, line 37: It 'd be good if we could list a few of what kind of things will be exposed with this approach.

kdy1: 👍 
I'll add it tomorrow as I'm going to sleep

kwonoj: Sure thing, no hurries. Just came to think of one more thing :

- what we won't try to expose as it is fundamentally related to the constraint explained in context, like mutable parent node itself or similar.

Austaras:
Maybe this could be done in a non breaking way?

Add two function to `Visit/VisitMut` `add_parent_path` and `remove_parent_path` and default them to noop. If user would like to save parent path, they could add a field on their `Struct` and overrider these functions. Hopefully rust could optimize that cost away.

This would require a bit more boilerplate, but doesn't need an extra feature and all the problems.

kdy1:
@Austaras This is not a breaking change.
This is about creating more traits.


I want to use the feature flag because

 - Core transforms will not use path-aware visitor
 - Making it optional via feature flag will improve compile time. (release build takes more than 1 hour at the moment)

But well, I think your solution is better than creating a whole new trait. ~~~(Except the compile time in release mode)~~~


Oh well we can use feature flag in this case too

kdy1:
After thinking about more things, I found that your solution is better in all cases, thank you!

kdy1:
I found one problem.

For `dyn Visit`, two dynamic method calls for each visitor methods call seems like too much.
Any ideas?

Austaras:
> Core transforms will not use path-aware visitor

Well I'm currently writing something greatly in need of parent scope.

> For dyn Visit, two dynamic method calls for each visitor methods call seems like too much.

That won't happen -- https://godbolt.org/z/1aG8T84vK

kdy1:
@Austaras LLVM has devirtualizer so you need more code to avoid it.

See assembly of https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=311d925a04c1e683ccc1c23996116e4d

kdy1:
See https://godbolt.org/z/9WzGYaM3Y

Austaras:
Your example doesn't have a default impl for add_parent_path -- do you mean [this](https://godbolt.org/z/WE4GT59xM)? In Visit::method1, there's only call to print.

kdy1:
I meant the count of `callq`

kdy1:
> Your example doesn't have a default impl for add_parent_path -- do you mean [this](https://godbolt.org/z/WE4GT59xM)? In Visit::method1, there's only call to print.

In your example, the trait is devirtualized. `f3` has call to print

kdy1:
Oh it's same in my example

Austaras:
> LLVM has devirtualizer
https://doc.rust-lang.org/std/hint/fn.black_box.html could be used to avoid that

kdy1:
I decided to fo with `with_path` because main use case is `&dyn Visit`

kdy1 (commented):
swc-bump:
 - swc_visit

PR #5075, opened 2022-06-30
fix(html/minifier): meta element minification

alexander-akait:
**Description:**

Remove spaces around non display elements, i.e. `meta`/`link`/`style`/etc + fix svg

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_minifier

PR #5078, opened 2022-07-01
fix(es/parser): labelled function

Austaras:
**Description:**
This is from annex B 3.1

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_parser/src/parser/stmt.rs
+++ b/crates/swc_ecma_parser/src/parser/stmt.rs
@@ -1030,13 +1030,19 @@ impl<'a, I: Tokens> Parser<I> {
                     function:
                         Function {
                             span,
-                            is_generator: true,
+                            is_generator,
+                            is_async,
                             ..
                         },
                     ..
                 }) = f
                 {
-                    syntax_error!(p, span, SyntaxError::LabelledGenerator)
+                    if p.ctx().strict {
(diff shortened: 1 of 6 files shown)


crates/swc_ecma_parser/src/parser/stmt.rs
```diff
@@ -1030,13 +1030,19 @@ impl<'a, I: Tokens> Parser<I> {
...
+                            is_generator,
+                            is_async,
                             ..
                         },
                     ..
                 }) = f
                 {
-                    syntax_error!(p, span, SyntaxError::LabelledGenerator)
+                    if p.ctx().strict {
+                        syntax_error!(p, span, SyntaxError::LabelledFunctionInStrict)
```
kdy1, line 1041: I think we can use `emit_err` instead?

crates/swc_ecma_parser/src/parser/stmt.rs
```diff
@@ -1030,13 +1030,19 @@ impl<'a, I: Tokens> Parser<I> {
...
+                    }
+                    if is_generator || is_async {
+                        syntax_error!(p, span, SyntaxError::LabelledGeneratorOrAsync)
```
kdy1, line 1044: Same here

kdy1 (changes requested):
Thank you!

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_parser

PR #5080, opened 2022-07-01
fix(es/module): improve cjs lexer compatibility

magic-akari:
**Description:**

It's a little weird that [cjs-module-lexer](https://github.com/nodejs/cjs-module-lexer) doesn't recognize exports getter if `enumerable: true` is not the first one.
This pull request improve the compatibility with cjs-module-lexer.

https://runkit.com/embed/ez1qzzokityr
image

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_module

magic-akari:
Ah, node-api test failed. I will update the test cases.

PR #5081, opened 2022-07-01
fix(es/module): preserve exports order

magic-akari:
**Description:**

I don't know if anyone will rely on export order. But preserving the order is easy for swc because we have collected all the span.

**BREAKING CHANGE:**

- https://github.com/babel/babel/issues/9177


crates/swc/tests/fixture/issues-2xxx/2964/case-2/output/index.ts
```diff
@@ -10,8 +10,8 @@ function _export(target, all) {
 }
 _export(exports, {
     badIstanbul: ()=>badIstanbul,
```
magic-akari, line 12: https://github.com/swc-project/swc/blob/428ca6a650cc221d017245568b71e43c7c440fca/crates/swc/tests/fixture/issues-2xxx/2964/case-2/input/index.ts#L1-L8

badIstanbul's span is lost. 

This issue will disappear if `...pixelParams` is removed.
I supposed it is a bug related to `object_rest_spread`.

magic-akari: I will fix this soon in another pull request.

kdy1: Cool, thanks!

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_transforms_module

magic-akari:
It looks like something is wrong with setup-node.

PR #5085, opened 2022-07-01
feat(html/minifier): remove empty metadata elements

alexander-akait:
**Description:**

Remove empty metadata elements, they will not be display so we can remove them from + option for it

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_minifier

PR #5088, opened 2022-07-02
test(es/module): add test for module-string-names

magic-akari:
**Description:**

The `module-string-names` syntax is introduced in #3190 and #3195 .
The transformation across module types is implemented in #4758 .
This pull request adds unit test cases to make sure that transformation works well.

**BREAKING CHANGE:**

- #3021
- #4758


kdy1 (approved):
Thanks!

---

swc-bump:
 - ''

PR #5089, opened 2022-07-02
fix(es/module): indirect call for lazy require

magic-akari:
**Description:**

It is my mistake to skip the indirect call transforms if the  callee is lazy required.
It still needs to be transformed into indirect call to correct `this`.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_module

PR #5087, opened 2022-07-02
feat(es/module): add pure_annotations to interop functions

magic-akari:
**Description:**

**BREAKING CHANGE:**

**Related issue (if exists):**

PR #5091, opened 2022-07-02
feat(es/module): support AMD triple slash directive

### LINKED ISSUE #4122: Read amd-module directive to determine AMD module IDs
### Describe the feature

Hi! First off thanks so much for adding support for naming AMD modules via the moduleId field in the .swcrc! 

Our team wants to compile a TS project that uses --outDir though, which means that if we want to be able to produce a named AMD module per file we need to generate a .swcrc for each TS file in the project.

I'm wondering if it's possible to also enable swc to read the name from a `/// <amd-module />` directive in a source file to determine the module Id on a per-file basis? This is a pretty standard way of defining AMD module names.

Link to the description of this directive in the TS handbook: https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-amd-module-

Our team's also happy to work on an implementation if this is a feature that y'all agree makes sense to support!

_No response_

### Additional context

_No response_

magic-akari:
**Description:**

**BREAKING CHANGE:**

- blocked by #5087 
- close #4122


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_module

PR #5092, opened 2022-07-02
fix(es): block scoped function

### LINKED ISSUE #4882: Block scoped function isn't correctly transformed
### Describe the bug

In ES6 and after, function declaration is strictly block scoped.

### Input code

```typescript
if (false) {
  function foo() {}
}

console.log(foo())
```

### Config

_No response_

### Playground link

https://play.swc.rs/?version=1.2.196&code=H4sIAAAAAAAAA8tMU9BIS8wpTtVUqOZSUEgrzUsuyczPU0jLz9cACtVy1XJxJefnFefnpOrl5KdrgMU1AXK46TA2AAAA&config=H4sIAAAAAAAAA0WMQQ7DIAwE%2F%2BIz1174Qx9hUSciAoxsRypC%2FL2gJsrJ1u7sdDg0gO9QUZRkfdqK4Rc8UMioQWI1cBObkclJw4Gh7GSL0NesErMS%2BA2TkoMcS9zaEgXOVUj1qbDs6SbHFGX%2BnCvoYK3OuyaZy6EwHtE1jvq%2B6P%2F8B7PHho66AAAA

### Expected behavior

```js
if (false) {
    var _$foo = function _$foo() {};
}
console.log(foo());
```

### Actual behavior

```js
if (false) {
    var _$foo = function _$foo() {};
}
console.log(_$foo());
```

### Version

1.2.196

### Additional context

_No response_

Austaras:
**Description:**

This comes from es spec annex b 3.2

This also updates lint so that function declaration is compliant to spec -- that a function declaration is non lexical in top level of function or script, and lexical in all other case

**BREAKING CHANGE:**

closes #4882


Austaras:
It's total waste of time and energy to implement this, but babel has it so anyway

crates/swc_ecma_minifier/tests/terser/compress/blocks/issue_1672_if_strict/output.js
```diff
@@ -1,12 +1,10 @@
...
-    }
-) {
+switch(function() {
+    return xxx;
+}){
     case xxx:
         if (console.log("FAIL")) {
-            function xxx() {}
+            function xxx1() {}
         }
```
Austaras, line 8: This is because swc minifier test defaults to `Module`, but does this really matter?

crates/swc_ecma_minifier/tests/terser/compress/blocks/issue_1672_for_strict/output.js
```diff
@@ -1,12 +1,10 @@
 "use strict";
-switch (
-    function () {
-        return xxx;
-    }
-) {
+switch(function() {
```
kdy1, line 2: Seems like terser does not implement annex B?

Austaras: It does. Terser just don't rename ident unnecessarily

Austaras: try adding 'use strict' to following code in https://try.terser.org/
```js
switch (
    function () {
        return xxx;
    }
) {
    case xxx:
        for (; console.log("FAIL"); ) {
            function xxx() {}
        }
        break;
}
```
you could see function xxx() {} be removed or not

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_transforms_base

Issue #5097, opened 2022-07-03
The Reference Acting on the Outer Const Variable is not Triggered


clhiker:
### Describe the bug

In the poc.js, v0++ will trigger a ReferenceError and catch then print 'failure'.
But in the transpiled code v0 is transpiled to the var, and it can't be caught.

### Input code

```typescript
const v0 = {};
try {
    Object.defineProperty(v0, 'foo', { value: 'bar' });
    v0++;
} catch (e) {
    console.log('failure');
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6",
    "strict": false,
    "strictMode": false,
    "lazy": false,
    "noInterop": false
  },
  "minify": false
}
```

### Playground link

https://play.swc.rs/?version=1.2.208&code=H4sIAAAAAAAAAy2NQQrDIBQF957i7b4hIbiu9AztFYz9timSX6wRgnj3SOnsZ8bL9s0oBlfUZlVOB6pC57a82ef5wWHd%2BJ7kwykfupgJFERoQkVxcecLaHGJ0Ab784oZR6savMv%2BBc3Dv%2Bf7SCLPUZ6aglvjnpi6007Gm8iRggAAAA%3D%3D&config=H4sIAAAAAAAAA0WMSwrDMAxE76K1t93kDj2EcJXg4B8eB2KM716ZumQlMe%2FNdDphaeuUuUDK%2FNBi5Zs2EhsYtrhcyaim0c4eMgxVLofUqeClzKcEWdRQcNHtbS7ZFHIR4EEcD%2F83hw6F9Llm0Km2rHdWQoonaDxDq%2BzwXvav%2FgXU82csuwAAAA%3D%3D

### Expected behavior

The ReferenceError can still be caught after transpiling.

### Version

1.2.208

PR #5098, opened 2022-07-03
fix(es/lint): cover more const assign cases

magic-akari:
**Description:**

**BREAKING CHANGE:**

- related #5097


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_lints

PR #5104, opened 2022-07-04
fix(es/compat): preserve class name

### LINKED ISSUE #5102: keep class name when transform es6 to es5
### Describe the feature

when using babel to transform es6 class to es5, the class name will be reserved:
input:
`let C = class {}`
output:
```
var C = /*#__PURE__*/_createClass(function C() {
  "use strict";

  _classCallCheck(this, C);
});
```

but when using SWC, the class name will be changed to `_class`
```
var C = function _class() {
    "use strict";
    _classCallCheck(this, _class);
};
```
the class `C` be considerd as an anonymous class, and all anonymous class will be named as _class.
so, is it possible to make SWC works like babel?

_No response_

### Additional context

_No response_

Austaras:
**Description:**

**BREAKING CHANGE:**

closes #5102


kdy1 (approved):
Thank you!


---

swc-bump:
 - swc_ecma_transforms_compat

PR #5118, opened 2022-07-05
fix(es/minifier): Don't inline a callable expression if it's used as a reference

kdy1:
**Description:**

# Investigation

This is an issue of `collapse_vars`

image

This `return` fixes the issue.

---

image

I found exact condition

 - `https://github.com/vercel/next.js/discussions/30237?sort=new#discussioncomment-3079694`


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5128, opened 2022-07-06
feat(swc/common): wrap serialized struct with versioned (#5060: Part 3)

kwonoj:
**Description:**

This PR actually wraps up structs with `VersionedSerializable` for the plugins. `PluginSerializedBytes`'s serialize / deserialize interface now explicitly accepts `VersionedSerializable` only. Most changes are straightforward, with implementing `default` for few structs & couple of `copy` around diagnostic / comment to workaround moves. 

There is one item I need to revisit later (https://github.com/swc-project/swc/compare/main...kwonoj:swc:feat-versioned-serialized#diff-b2b47f82dd3bafcc79a1e0e6ec2ccb75624c020e295e2a65c03d01058b56da92R178) since `Result<T>` is not possible to use `VersionedSerializable::take()` due to lacks of default.

**BREAKING CHANGE:**

#5060

### DIFF
--- a/crates/swc/src/plugin.rs
+++ b/crates/swc/src/plugin.rs
@@ -80,7 +80,10 @@ impl RustPlugins {
         use std::{path::PathBuf, sync::Arc};
 
         use anyhow::Context;
-        use swc_common::{plugin::PluginSerializedBytes, FileName};
+        use swc_common::{
+            plugin::{PluginSerializedBytes, VersionedSerializable},
+            FileName,
+        };
         use swc_ecma_loader::resolve::Resolve;
 
         // swc_plugin_macro will not inject proxy to the comments if comments is empty
@@ -92,7 +95,8 @@ impl RustPlugins {
                 inner: self.comments.clone(),
             },
             || {
-                let mut serialized = PluginSerializedBytes::try_serialize(&n)?;
+                let program = VersionedSerializable::new(n);
+                let mut serialized = PluginSerializedBytes::try_serialize(&program)?;
 
                 // Run plugin transformation against current program.
                 // We do not serialize / deserialize between each plugin execution but
@@ -111,11 +115,19 @@ impl RustPlugins {
 
                         let config_json = serde_json::to_string(&p.1)
                             .context("Failed to serialize plugin config as json")
-                            .and_then(|value| PluginSerializedBytes::try_serialize(&value))?;
+                            .and_then(|value| {
+                                PluginSerializedBytes::try_serialize(&VersionedSerializable::new(
+                                    value,
+                                ))
+                            })?;
 
                         let context_json = serde_json::to_string(&self.plugin_context)
                             .context("Failed to serialize plugin context as json")
-                            .and_then(|value| PluginSerializedBytes::try_serialize(&value))?;
+                            .and_then(|value| {
+                                PluginSerializedBytes::try_serialize(&VersionedSerializable::new(
+                                    value,
+                                ))
+                            })?;
 
                         let resolved_path = self
                             .resolver
@@ -152,7 +164,7 @@ impl RustPlugins {
 
                 // Plugin transformation is done. Deserialize transformed bytes back
                 // into Program
-                serialized.deserialize()
+                serialized.deserialize().map(|mut v| v.take())
             },
         )
     }
@@ -162,7 +174,10 @@ impl RustPlugins {
         use std::{path::PathBuf, sync::Arc};
 
         use anyhow::Context;
-        use swc_common::{plugin::PluginSerializedBytes, FileName};
+        use swc_common::{
+            plugin::{PluginSerializedBytes, VersionedSerializable},
+            FileName,
+        };
         use swc_ecma_loader::resolve::Resolve;
 
         let should_enable_comments_proxy = self.comments.is_some();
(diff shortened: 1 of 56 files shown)


crates/swc/src/plugin.rs
```diff
@@ -152,7 +164,7 @@ impl RustPlugins {
 
                 // Plugin transformation is done. Deserialize transformed bytes back
                 // into Program
-                serialized.deserialize()
+                serialized.deserialize().map(|mut v| v.take())
```
kdy1, line 167: Why do we need  `.take()` in `map`?

kdy1 (changes requested):
Can't we add something like `into_inner()` instead of using `.take()`?

crates/swc_common/src/comments.rs
```diff
@@ -548,6 +548,16 @@ pub struct Comment {
     pub text: String,
 }
 
+impl Default for Comment {
+    fn default() -> Self {
+        Comment {
+            kind: CommentKind::Line,
+            span: DUMMY_SP,
+            text: "".into(),
```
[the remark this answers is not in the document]
kwonoj: Yes, this is oversight after I created it via copilot 😓

crates/swc/src/plugin.rs
kdy1: I checked it and I think we can add `into_inner()` to `VersionedSerializable`

kwonoj: That sounds legit to me.

kwonoj: Actually, I misread this: `serialized.deserialize()` returns `Result<VersionedSerializable<Program>, Error>` and we need to return `Result<Program, Error>` by taking out VersionedSerializable, so here we call `Result::map`. Implementing `into_inner`  to `VersionedSerializable` can resolve this without `map`?

kdy1: If it's a result, `map` is fine. I thought it was an option

crates/swc_common/src/syntax_pos.rs
```diff
@@ -1193,6 +1210,7 @@ pub struct LineCol {
     pub col: u32,
 }
 
+#[derive(Default)]
```
kdy1, line 1213: Same here

crates/swc_plugin_proxy/src/memory_interop/read_returned_result_from_host.rs
```diff
@@ -1,10 +1,11 @@
 #[cfg_attr(not(target_arch = "wasm32"), allow(unused))]
 use swc_common::plugin::{
     deserialize_from_ptr, deserialize_from_ptr_into_fallible, PluginSerializedBytes,
+    VersionedSerializable,
 };
 
 /// A struct to exchange allocated data between memory spaces.
-#[derive(rkyv::Archive, rkyv::Serialize, rkyv::Deserialize)]
+#[derive(rkyv::Archive, rkyv::Serialize, rkyv::Deserialize, Default)]
```
kdy1, line 8: Same here

crates/swc_plugin_proxy/src/memory_interop/read_returned_result_from_host.rs
```diff
@@ -85,11 +88,11 @@ where
 pub fn read_returned_result_from_host_fallible<F, R>(f: F) -> Option<R>
 where
     F: FnOnce(i32) -> i32,
-    R: rkyv::Archive,
+    R: rkyv::Archive + std::default::Default,
```
kdy1, line 91: Same here

kdy1 (changes requested):
Wrong `Default` impl should not exist.
We should use `Take::dummy()` instead

crates/swc_common/src/comments.rs
```diff
@@ -548,6 +548,16 @@ pub struct Comment {
     pub text: String,
 }
 
+impl Default for Comment {
```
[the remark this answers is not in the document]
kwonoj: default was added to satisfy mem::take in `VersionedSerializable::take`. if there's other way default is not required mandatoriliy.

I'm bit unsure if I understood the suggestion correctly, mind elaborate bit more? The way I understood is let each struct implement `Take::dummy()` instead of default, and VersionedSerializable should utilize it for the `take`. Not sure where `into_inner` plays in here.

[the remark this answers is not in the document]
kdy1: `into_inner` take ownership of `VersionedSerializable` so it does not require `take(&mut self.0.1)`.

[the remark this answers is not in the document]
kdy1: The `Default` constraint comes from it.

crates/swc_plugin_proxy/src/memory_interop/read_returned_result_from_host.rs
kwonoj: I'll take it out with other refactorings, but this is technically possible to be correctly one, isn't it? If we'll need default later somehow, I expect this struct can use default derive.

kdy1: Problem is that this is a pointer, as `0` is definitely not a good number for pointer

kwonoj: Yes, but for a new value / or default I'd expect 0 as invalid ptr number to be filled in. But in any case, that's somethign to discuss in the future and PR removed default impl.

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_common --breaking

PR #5134, opened 2022-07-07

kdy1:
**Description:**


kdy1:
image

I profiled it but I think path-based optimization will regress performance

PR #5147, opened 2022-07-08
refactor(swc/plugin_runner): refactor transform executor (#5060: Part 4)

kwonoj:
- refactor(swc/plugin_runner): refactor transform executor
- refactor(swc/plugin_macro): refine internal macro fn ident name

**Description:**

To introduce another call to plugin binary for schema version invalidation, PR refactors execution entrypoint bit to allow `TransformExecutor` can be used for subsequent calls.

Also amends internal swc_plugin_macro's ident name to clarify `transform` plugin. This'll probably break existing plugins and requires to bump up dependencies on each plugin's end. Something we should do before out of experimental flag.

**BREAKING CHANGE:**

#5060


crates/swc_plugin_runner/src/transform_executor.rs
```diff
@@ -102,14 +102,26 @@ impl TransformExecutor {
...
+     * Host should appropriately handle if plugin is not compatible to the
+     * current runtime.
+     */
+    pub fn is_transform_schema_compatible(&self) -> bool {
+        todo!("Not supported yet");
+    }
+
     #[tracing::instrument(level = "info", skip_all)]
     pub fn transform(
         &mut self,
```
kdy1, line 118: Can we take `self` instead?
As it's single-run only, it's better to take ownership.

kdy1 (commented):
Not sure, I'm just asking if it's possible

crates/swc_plugin_runner/src/transform_executor.rs
kwonoj: I'd like to see with further refactorings - I am under assumption I may need to touch few other places until implementing version schema. Once it's done I can determine if this is one-off and can take self completely.

PR #5148, opened 2022-07-08
feat(html/parser): `raw` for text

alexander-akait:
**Description:**

Support `raw` for future linting and allow to get original written value in codegen.

**BREAKING CHANGE:**

Yes

No


crates/swc_html_parser/tests/recovery/element/a/input.html
```diff
@@ -1 +1 @@
-<a><p></a></p>
\ No newline at end of file
+<
```
kdy1, line 1: Is this change intentional?

alexander-akait: Good catch, thank you

alexander-akait:
@kdy1 Fixed

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_ast --breaking

PR #5158, opened 2022-07-08
feat(swc/plugin_macro): export new interface for getting plugin schema version (#5060: Part 5) 

kwonoj:
**Description:**

This PR adds a new fn in plugin's macro side, returns schema version defined in `swc_common` at the moment of compilation. To support per-build time flags, also introduces features to the downstream pkg from top level (binding-*).

Changes are based on theoritical workflow I have in mind, which means this is possibly not correct and would require further refactorings. But having each step helps to setting up actual (manually) testable workflow.

Once this goes in, I'll pause bit and try to attempt refactor integration test for the plugin. I feel it's hitting tipping point to worth better ergonomics to test more than simple transformation execution. Probably will need node.js side testing as well for whole integrations.

**BREAKING CHANGE:**

#5060


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc
 - swc_cli
 - swc_common
 - swc_plugin
 - swc_plugin_runner
 - swc_plugin_macro

PR #5159, opened 2022-07-08
fix(es/module): Handle top level this

### LINKED ISSUE #5157: `this` gets wrongly compiled to `(void 0)` when inside a property definition of an object literal
### Describe the bug

When defining a property on a plain object, references to `this` will be replaced erroneously with `void 0`. 

Example: 
```javascript
const foo = { 
    get prop1() {
        return 1
    },
    get prop2() {
        return this.prop1 + 1
    }
}
```

This will be compiled to:
```javascript
const foo = {
    get prop1() {
        return 1
    },
    get prop2 () {
        return (void 0).prop1 + 1;
    }
};
```

Note `(void 0)` inside `prop2`. This is incorrect because `this` should be a valid reference here. 

### Input code

```typescript
const foo = { 
    get prop1() {
        return 1
    },
    get prop2() {
        return this.prop1 + 1
    }
}
```

### Config

```json
{
  "test": [".*.ts",".*.js"],
  "exclude": [".*\\.d\\.ts$", ".*\\.map$"],
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "decorators": false,
      "dynamicImport": false
    },
    "target" : "es2021"
  },
  "module": {
    "type": "commonjs"
  },
  "sourceMaps": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.128&code=H4sIAAAAAAAAA0utKMgvKlFIzkksLlYILkksyUwOSS0ucQbzq7kUgKAYLKpQAhQOKMovULBVUPdIzcnJVyjPL8pJUVS35kJX5ptakpGfAlSooalgawc1BgSS8%2FOK83NS9XLy0zVKMjKL9WCGalqD1dRy1QIAgrT4spEAAAA%3D&config=H4sIAAAAAAAAA0VOSwqDMBC9SnlrF1Xoxjv0EEFHieTHTCyVkLt3IhZ3j%2Fcv2GTCWJAMC3FDcoRsvhiRj0QysU0ZHbIolXmnqtjwSlkdJMOzf6nqYhTCuBgn1MHbYJejdU3RJyaRWzJhdX9n1S4f570R5ZzTTo34GDZBvYuusJX35W5HOlD4nMHzjm4gxJke%2FYBaf%2FJ6D0jYAAAA

### Expected behavior
```javascript
const foo = {
    get prop1() {
        return 1
    },
    get prop2 () {
        return this.prop1 + 1;
    }
};
```

### Actual behavior
```javascript
const foo = {
    get prop1() {
        return 1
    },
    get prop2 () {
        return (void 0).prop1 + 1;
    }
};
```

### Version

1.2.210

### Additional context

* I notice this bug has been introduced in 1.2.206.  
* The SWC playground link includes links of other usages of `this` which work correctly. It seems to be specifically only: 
  (1) inside a property definition; AND
  (2) if the property is defined on an object literal, as opposed to a class.

magic-akari:
**Description:**

**BREAKING CHANGE:**

- fix #5157


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_module

Austaras:
Once again, It's top level in accessor/private accessor's computed prop name

magic-akari:
> Once again, It's top level in accessor/private accessor's computed prop name

Thanks, I will check again.

PR #5160, opened 2022-07-09
fix(es/module): Handle top level this really

magic-akari:
**Description:**

Note:
I can't construct `PrivateName` test case that contains `this`. But technically it is indeed top level.

**BREAKING CHANGE:**

- fix https://github.com/swc-project/swc/pull/5159#issuecomment-1179478734


kdy1 (approved):
Thank you!

---
swc-bump:
 - swc_ecma_transforms_module

PR #5164, opened 2022-07-09
fix(es/module): keep TDZ for `export default` if possible

magic-akari:
**Description:**

I started to reconsider whether I should put the module pass before the compat pass.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

---

swc-bump:

PR #5170, opened 2022-07-10
fix(es/module): add missing span for `export default`

magic-akari:
**Description:**

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

---
swc-bump:
 - swc_ecma_transforms_module

Issue #2175, opened 2021-08-29
AST struct transfer, serialization, deserialization performance enhancement


overlookmotel:
As mentioned in #2123 and #1392, `parse()` is currently quite slow.

SWC's Rust parser is of course extremely fast, but the bottleneck is passing ASTs back into JS.

I've done some experiments, and believe I may have the beginnings of a solution to speed it up by 5x or more, enough so it out-performs Babel at least.

https://github.com/overlookmotel/swc-parse-test

![1000 lines](https://raw.githubusercontent.com/overlookmotel/swc-parse-test/master/doc/images/1000_lines_brief.png)

### The problem

There are 2 bottlenecks involved in transferring AST from Rust to JS:

1. Cost of passing AST from Rust to JS (as JSON string)
2. Cost of parsing JSON in JS (using `JSON.parse()`)

### Potential solution

I've attempted to alleviate these pinch points by outputting a buffer (`JsBuffer`) from Rust instead of a JSON string. The buffer is a binary serialization of the AST, and I've written a matching deserializer in JS.

This has 2 advantages:

1. `JsBuffer` can be created in main thread's memory so passing it from Rust to JS is zero cost.
2. AST can be encoded and deserialized more efficiently in binary form.

The results appear to be pretty good.

### Provisos

The implementation is only a proof of concept. So far it can only deal with `const x = 1;` statements. I see no reason why it shouldn't maintain it's speed when support for all AST node types is included, but I could be wrong.

It may also be naive. Until this weekend, I've never written a line of Rust before, so I may be totally missing a reason why it's not possible to integrate this approach into SWC.

The Rust code I've written is, for the same reason, quite embarrassingly bad. The JS code is also hacked together very quickly. So potentially it could get significantly faster with some optimization.

### Questions

* Does this seem like a promising direction?

* If `parse()` (and `print()`) can be made much faster, might this make it feasible to maintain support for plugins written in Javascript?

As mentioned above, I don't know Rust at all, so I have limited ability to help with an effort to flesh out these experiments into fully-fledged features. However, I would be very happy to help out on the JS side (the deserializer etc).

kdy1:
Yeah, if we can improve performance of passing ast, I expect the js plugin api to be really useful.

Brooooooklyn:
Maybe deserialize via https://github.com/rkyv/rkyv in the wasm and convert into JsObject will be more faster.

overlookmotel:
[...]

PR #5175, opened 2022-07-11
Fix TS types for AST nodes

overlookmotel:
**Description:**

This PR fixes various inaccuracies in the TS type definitions for AST nodes. They seem to have got out of sync over time.

These discrepancies were identified programmatically based on the model of the Rust types I've made for #2175. So this PR should be definitive and bring the TS types completely up to date.

NB Some of the changes just re-order fields and enum options. This doesn't make a difference to correctness of the types. But now they're all consistent with field order in the Rust types, which will maybe make it easier to spot when further changes are required in future.

n/a

### DIFF
--- a/node-swc/src/Visitor.ts
+++ b/node-swc/src/Visitor.ts
@@ -82,6 +82,7 @@ import {
   MethodProperty,
   Module,
   ModuleDeclaration,
+  ModuleExportName,
   ModuleItem,
   NamedExportSpecifier,
   NamedImportSpecifier,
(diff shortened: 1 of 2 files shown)


kdy1 (changes requested):
CI failed

overlookmotel:
Yes, the types are now correct, but their usage in `node-swc/src/Visitor.ts` is now wrong!

overlookmotel:
I've made changes to `Visitor.ts` to use the updated types, so TS check now passes.

There are likely a lot of inaccuracies remaining in `Visitor.ts`, but I'll try to fix them another time. At least the types are correct now.

kdy1 (approved):
Thank you!

Issue #5168, opened 2022-07-10
transform produces different result if called with AST


overlookmotel:
### Describe the bug

I'm not sure if this is a bug or not, but it's certainly strange.

```js
const { transformSync, parseSync } = require('@swc/core');

const code = 'x ??= true';
console.log( transformSync( code ) );
console.log( transformSync( parseSync( code ) ) );
```

Output:

```js
x !== null && x !== void 0 ? x : x = true;
x = x !== null && x !== void 0 ? x : true;
```

The first is more correct. If `x` is a global defined with a setter, the 2nd will trigger the setter even when `x` is not nullish - which `x ??= true` doesn't.

```js
Object.defineProperty( globalThis, 'x', {
  get(v) { return true; },
  set(v) { console.log('set!'); }
} );

x ??= true; // Nothing happens
x = x !== null && x !== void 0 ? x : true; // Logs "set!"
```

But mainly I raise this because it may indicate some other problem. Perhaps some transforms are being skipped, or the wrong target is being selected.

Any idea what might be causing this?

### Input code

```typescript
x ??= true
```

### Config

```json
n/a
```

### Expected behavior

Output in both cases should be `x !== null && x !== void 0 ? x : x = true;`.

### Version

1.2.211

overlookmotel:
Here's another one:

```js
const code = 'c ??= class {}';
console.log( transformSync( code ) );
console.log( transformSync( parseSync( code ) ) );
```

Output:

```js
function _classCallCheck(instance, Constructor) {
    if (!(instance instanceof Constructor)) {
        throw new TypeError("Cannot call a class as a function");
    }
}
c !== null && c !== void 0 ? c : c = function c() {
    "use strict";
    _classCallCheck(this, c);
};

function _classCallCheck(instance, Constructor) {
    if (!(instance instanceof Constructor)) {
        throw new TypeError("Cannot call a class as a function");
    }
}
c = c !== null && c !== void 0 ? c : function _class() {
    "use strict";
    _classCallCheck(this, _class);
};
```

The 2nd case is a regression of #5102.

magic-akari:
There is something wrong with the `deserialize_json`.

magic-akari:
There are two ways to get Rust side AST.

source code -> parse -> AST
AST JSON(parsed from source code) -> deserialize -> AST

Take the following code as an example.

```JavaScript 
x ??= input;
```

It will be parsed into the following AST.

```
AssignExpr {
    span: Span {..},
[...]

PR #5179, opened 2022-07-11
Fix JSON deserialization of AssignExpr

overlookmotel:
**Description:**

Fix for #5168. However:

1. Surely there's an easier way to do this?
2. I imagine there is a more idiomatic way to use `match { }` here:

https://github.com/overlookmotel/swc/blob/assign-expr-deserialize/crates/swc_ecma_ast/src/expr.rs#L568-L576

I would appreciate any pointers as to how to improve this.

#5168


kdy1 (changes requested):
Please remove images

magic-akari:
Is there a case where `PatOrExpr::Expr` is not `Expr::Ident`? @kdy1 
I wonder if `PatOrExpr` is redundant?

magic-akari:
It seems a lot of work to do if I remove the `PatOrExpr`, I will try it later.
And I don't have strong opinions on this pull request change. It's ok to fix #5168.

crates/swc_ecma_ast/src/expr.rs
```diff
@@ -471,6 +492,102 @@ impl Take for AssignExpr {
...
+                if op_field.is_none() {
+                    return Err(de::Error::missing_field("operator"));
+                }
+                if left_field.is_none() {
+                    return Err(de::Error::missing_field("left"));
+                }
+                if right_field.is_none() {
+                    return Err(de::Error::missing_field("right"));
+                }
+
```
kdy1, line 562: Can you use `.ok_or_else` instead for these 4 vars?

crates/swc_ecma_ast/src/expr.rs
```diff
@@ -471,6 +492,102 @@ impl Take for AssignExpr {
...
+            {
+                let mut span_field: Option<Span> = None;
+                let mut op_field: Option<AssignOp> = None;
+                let mut left_field: Option<PatOrExpr> = None;
+                let mut right_field: Option<Box<Expr>> = None;
+
+                while let Some(key) = map.next_key::<&str>()? {
+                    match key {
+                        "span" => {
+                            if Option::is_some(&span_field) {
```
kdy1, line 521: Also, please change this to `span_field.is_some()`

kdy1 (changes requested):
Thank you!

crates/swc_ecma_ast/src/expr.rs
overlookmotel: Done.

crates/swc_ecma_ast/src/expr.rs
overlookmotel: Done.

crates/swc_ecma_ast/src/expr.rs
```diff
@@ -471,6 +492,102 @@ impl Take for AssignExpr {
...
+                                return Err(de::Error::duplicate_field("span"));
+                            }
+                            span_field = Some(map.next_value()?);
+                        }
+                        "operator" => {
+                            if Option::is_some(&op_field) {
```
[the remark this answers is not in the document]
overlookmotel: Done.

crates/swc_ecma_ast/src/expr.rs
```diff
@@ -471,6 +492,102 @@ impl Take for AssignExpr {
...
+                                return Err(de::Error::duplicate_field("operator"));
+                            }
+                            op_field = Some(map.next_value()?);
+                        }
+                        "left" => {
+                            if Option::is_some(&left_field) {
```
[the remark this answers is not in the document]
overlookmotel: Done.

crates/swc_ecma_ast/src/expr.rs
```diff
@@ -471,6 +492,102 @@ impl Take for AssignExpr {
...
+                                return Err(de::Error::duplicate_field("left"));
+                            }
+                            left_field = Some(map.next_value()?);
+                        }
+                        "right" => {
+                            if Option::is_some(&right_field) {
```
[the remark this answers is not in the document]
overlookmotel: Done.

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_ast

overlookmotel:
I've made the changes requests and tidied up some other bits.

I'm still not completely happy with this:

```rust
if let PatOrExpr::Pat(ref pat) = left {
  if let Pat::Ident(ident) = &**pat {
    left = PatOrExpr::Expr(Box::new(Expr::Ident(ident.id.clone())));
  }
}
```

The `.clone()` call should be unnecessary - `ident.id` can just be moved from the old structure to the new. But I couldn't find a way to get the compiler to accept this, since `pat` is borrowed.

My first attempt was:

```rust
left = match left {
  PatOrExpr::Pat(pat) => match *pat {
    Pat::Ident(ident) => PatOrExpr::Expr(Box::new(Expr::Ident(ident.id))),
    pat => PatOrExpr::Pat(Box::new(pat)),
  },
  left => left,
}
```

This avoids cloning as it takes ownership, but has the worse consequences of unclear code and re-creating `PatOrExpr::Pat(Box::new(pat))` unnecessarily in the majority of cases.

Or am I worrying too much? Will the compiler compile away the `.clone()` call?

kdy1:
I think so?
Even if the compiler fails to optimize it, it will not affect performance in almost all cases.
`clone()` of `Ident` is typically very cheap.

overlookmotel:
Is there no way to tell compiler "give me ownership of `ident`, and make `left` and `pat` uninitialized"? Sorry, this has nothing to do with SWC specifically, I'm just still learning Rust. This feels like it'd be a common situation.

PR #5181, opened 2022-07-11
fix(css/parser): avoid skip whitespaces where unnecessary

alexander-akait:
**Description:**

Remove wrong skipping whitespaces + more comments from spec

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_parser

PR #5182, opened 2022-07-11
feat(html/minifier): compress more default attributes

alexander-akait:
**Description:**

Here:
1. Compress more default attributs
2. Move attributes to JSON file (easy to maintance and faster due hashing), SVG default attributes will be moved to `swc_svg_minify` in future, that is why I moved it to separate file

**BREAKING CHANGE:**

Yes

No


crates/swc_html_minifier/Cargo.toml
```diff
@@ -36,6 +36,7 @@ swc_html_ast = { version = "0.17.0", path = "../swc_html_ast" }
 swc_html_codegen = { version = "0.22.0", path = "../swc_html_codegen" }
 swc_html_visit = { version = "0.17.0", path = "../swc_html_visit" }
 swc_html_parser = { version = "0.20.0", path = "../swc_html_parser" }
+swc_html_utils = { version = "0.6.0", path = "../swc_html_utils" }
```
kdy1, line 40: Please patch `includes`

```toml
include = ["Cargo.toml", "src/**/*.rs"]
```

this will not work

kdy1 (changes requested):

crates/swc_html_minifier/src/lib.rs
```diff
@@ -470,351 +471,78 @@ impl Minifier<'_> {
         attribute_name: &str,
         attribute_value: &str,
     ) -> bool {
```
IWANABETHATGUY, line 473: The return type could be written as `Option<bool>`, so you could use the `?` in function body.

crates/swc_html_minifier/src/lib.rs
```diff
@@ -470,351 +471,78 @@ impl Minifier<'_> {
...
+                let default_attributes = if namespace == Namespace::HTML {
+                    &HTML_DEFAULT_ATTRIBUTES
+                } else {
+                    &SVG_DEFAULT_ATTRIBUTES
+                };
+
+                let attributes = match default_attributes.get(tag_name) {
+                    Some(element) => element,
+                    None => return false,
+                };
```
IWANABETHATGUY, line 518: So you could write the code like this `let attributes = default_attributes.get(tag_name)?`

crates/swc_html_minifier/src/lib.rs
```diff
@@ -470,351 +471,78 @@ impl Minifier<'_> {
...
+
+                let default_value = match attributes.0.get(attribute_name) {
+                    Some(default_value) => default_value,
+                    None => return false,
```
IWANABETHATGUY, line 522: Same here.

crates/swc_html_minifier/src/lib.rs
alexander-akait: But why? I feel fine with `bool`

kdy1 (approved):
Thank you!


---

swc-bump:
 - swc_html_utils --breaking

PR #5153, opened 2022-07-08
fix(source-map): better handling of multi-byte characters

dsherret:
**Description:**

Uses character lengths instead of byte lengths in some places.

**Related issue (if exists):**

Closes #5145

PR #5183, opened 2022-07-11
fix(css/codegen): source map generation

alexander-akait:
**Description:**

Fix source map generation with multi bytes characters + improve perf

**BREAKING CHANGE:**

No

https://github.com/swc-project/swc/pull/5153


kdy1 (approved):
Thanks!

---
swc-bump:
 - swc_css_codegen

PR #5184, opened 2022-07-11
fix(html/codegen): source map generation

alexander-akait:
**Description:**

Fix source map generation for multi byte - in real life not many it is very rare case, but can be useful + avoid new allocation

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thank you!


---

swc-bump:
 - swc_html_codegen --breaking

crates/swc_html_codegen/src/writer/basic.rs
```diff
@@ -221,3 +242,30 @@ where
         self.indent_level -= 1;
     }
 }
+
+fn compute_line_starts(s: &str) -> Vec<usize> {
```
IWANABETHATGUY, line 246: This implementation seems not right, a few cases.
```rs
let case1 = "";
let case2 = "test\r\n";
```

PR #5187, opened 2022-07-12
test(swc): refactor test fixtures

kwonoj:
**Description:**

PR adjusts fixture locations from top level `/tests` to per each test location accordingly.

Actually, the only fixture we need is `internal_plugin`, rest of them are either unused (srcmap), or redundant that we can use cmd directly instead. This is first step to introduce more actual `integration` test for the plugins.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - ''

PR #5192, opened 2022-07-13
test(swc): initial fixture setup for e2e plugin test

kwonoj:
**Description:**

For transform version schema, this PR try to introduce additional set of tests to test plugin fully end to end via node bindings. This is an initial effort to setup base fixture to build corresponding node bindings per each version (`v1`, `vtest`).

There is no actual test yet. There are lot additional setup need to be done. Splitting PR into small chunk to avoid unnecessarily large PR to review.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---
swc-bump:
 - ''

PR #5196, opened 2022-07-13
feat(html/parser): raw for comments

alexander-akait:
**Description:**

Raw for comments, it will allow to print comments as they were written in codegen

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thank you!


---

swc-bump:
 - swc_html_ast --breaking

PR #5203, opened 2022-07-14
ci(publish): skip e2e tests for unsupported targets

kwonoj:
**Description:**

Workaround attempt to skip e2e test for the targets does not have plugin support.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

---
swc-bump:
 - ''

PR #5213, opened 2022-07-14
fix(swc/cli): emit sourcemap with file inputs

### LINKED ISSUE #5171: swcx (pure rust CLI) doesn't write sourcemap file
### Describe the bug

Trying to use the pure rust CLI for performance under Bazel, thanks @kwonoj for the fix in #5124 but hit the next problem.

### Input code

```typescript
1
```

### Config

```json
{
  "sourceMaps": true
}
```

### Playground link

_No response_

### Expected behavior

should write a sourcemap file at the `--source-maps-target` path

### Actual behavior

no sourcemap file is written anywhere that I can find

### Version

1.2.211

### Additional context

Here's my repro:

```
$ wget https://registry.npmjs.org/@swc/core-linux-x64-gnu/-/core-linux-x64-gnu-1.2.211.tgz
$ tar -xzvf core-linux-x64-gnu-1.2.211.tgz
$ chmod u+x package/swc
$ echo 1 > input.ts
$ jq --null-input '.sourceMaps = true' > .swcrc
$ ./package/swc compile input.ts --source-maps true --out-file my.js --source-maps-target my.js.map
$ ls
core-linux-x64-gnu-1.2.211.tgz  input.ts  my.js  package
```

I can't find `my.js.map` anywhere, and can't find another way to invoke the CLI such that this file is written.

kwonoj:
**Description:**

- close #5171

Attempt to close #5171, emits sourcemap if inputs are file instead of dir.

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_cli/src/commands/compile.rs
+++ b/crates/swc_cli/src/commands/compile.rs
@@ -54,7 +54,7 @@ pub struct CompileOptions {
 
     /// Define the file for the source map.
     #[clap(long)]
-    source_maps_target: Option<String>,
+    source_map_target: Option<String>,
 
     /// Set sources[0] on returned source map
     #[clap(long)]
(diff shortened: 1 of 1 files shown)


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_cli

Issue #1686, opened 2021-05-12
Import statements are hoisted when target is CommonJs


petkaantonov:
**Describe the bug**

Import statements are hoisted when target is CommonJS:

```
import A from "a;
console.log(1);
import B from "a";
```

Becomes

```
require(...);
require(...);
console.log(1);
```

This is incorrect, require calls are normal functions and should have preserved order. Imports should only be hoisted when they are actual es6 imports. This is incompatible with tsc and probably others, tsc keeps the order as expected when output module is commonjs.

**Additional context**

It was fixed but then reverted in https://github.com/swc-project/swc/issues/1457

mischnic:
With native esm (in Node or in browsers), the following prints "a, b, index"

```js
// index.js
import "./a.js";
console.log("index");
import "./b.js";

// a.js
console.log("a");

// b.js
console.log("b");
```

And I'd argue that transpiling the code to CJS shouldn't change the behaviour of the code

petkaantonov:
- It is incompatible with tsc
- It sucks for commonjs users, remember we are forced to use import not because it is better syntax but because typescript types dont work with require
- If you are able to use esm, why compile code to cjs? and why expect the behavior be the same? They are incompatible module systems.

mischnic:
From my perspective, the real problem is that [tsc](https://www.typescriptlang.org/play?module=1&ssl=1&ssc=1&pln=3&pc=12#code/JYWwDg9gTgLgBAIgIYINwCgDGEB2BnCAGwFMA6QiAcwAoBGASg1ElkQCM0g) and [Babel](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.6&spec=false&loose=false&code_lz=JYWwDg9gTgLgBAIgIYINwCgDGEB2BnCAGwFMA6QiAcwAoBGASg1ElkQCM0g&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2&prettier=false&targets=&version=7.14.1&externalPlugins=)/the spec don't agree

I guess swc has to choose a side here.

kdy1:
This feature is useful for using some of node.js modules, which expect execution order of common js modules.
[...]

PR #5215, opened 2022-07-15
fix(es/module): Preserve order for `import foo = require("mod")`

magic-akari:
**Description:**

Import/Export should be hoisted, but Ts Import require should be kept in order.

Input code

```TypeScript
import A = require("a"); // print a
console.log(1);
import B = require("b"); // print b
```

Before this pull request:

```JavaScript
"use strict";
    value: true
});
const _a = require("a");
const _b = require("b");
console.log(1);
```

This pull request change:

```JavaScript
"use strict";
    value: true
});
const A = require("a"); // print a
console.log(1);
const B = require("b"); // print b
```

**BREAKING CHANGE:**

- #1686


kdy1 (approved):
Thank you!

---

swc-bump:
  - swc_ecma_transforms_module

PR #5216, opened 2022-07-15
test(swc): fix e2e tests on windows

kwonoj:
**Description:**

This was root cause i686 CI was hanging indefinitely. We do not run test anymore, but still this need to be fixed for some CI / local dev env.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---
swc-bump:
 - ''

PR #5223, opened 2022-07-16
fix(es/codegen): number and bigint literal

Austaras:
**Description:**

**BREAKING CHANGE:**


crates/swc_ecma_codegen/src/lib.rs
```diff
@@ -3623,7 +3628,7 @@ fn minify_number(num: f64) -> String {
 
     let mut original = printed.clone();
 
-    if num.fract() == 0.0 && num.is_sign_positive() && num <= (0x7fffffffffffffff_i64 as f64) {
+    if num.fract() == 0.0 && (i64::MIN as f64) <= num && num <= (i64::MAX as f64) {
         let hex = format!("{:#x}", num as i64);
```
Austaras, line 3632: Maybe there's a better way

crates/swc_ecma_codegen/src/lib.rs
```diff
@@ -3635,6 +3640,10 @@ fn minify_number(num: f64) -> String {
         original.replace_range(0..1, "");
     }
 
+    if original.starts_with("-0.") {
```
Austaras, line 3643: terser would parse `-0.5` as Unary ( Number (0.5)), so in here we need to treat negative number too

alexander-akait: Sounds like a bug somewhere, there are no negative numbers in JS, it should be always `Unary + Number`,

kdy1: @alexander-akait minifier generates negative numbers because it's much easier to match on

crates/swc_ecma_codegen/src/lib.rs
```diff
@@ -661,7 +661,14 @@ where
         self.emit_leading_comments_of_span(v.span, false)?;
 
         if self.cfg.minify {
-            self.wr.write_lit(v.span, &v.value.to_string())?;
+            let value = if v.value >= 10000000000000000_i64.into() {
```
kdy1, line 664: What's this value?

kdy1: Where does it come from?

kdy1 (changes requested):
Thank you and sorry for the late review

crates/swc_ecma_codegen/src/lib.rs
kdy1: Oh it's not a change request. It was just question

Austaras: 10000000000000000n toString ->
10000000000000000n (length 17)
10000000000000000n toString(16) ->
0x2386f26fc10000n (length 16)

So I ran a program to find the smallest bigint that hex representation is shorter than decimal and the result is 10000000000000000n

kdy1: Got it, thank you!

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_codegen

PR #5229, opened 2022-07-17
build(package): use explicit list for pkg files

kwonoj:
**Description:**

Minor refactor to pkg manifest to include explicit list of files instead of excluding via npmignore, removes redundant files like prettier config from pkg.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

---

swc-bump:
 - ''

Issue #4414, opened 2022-04-23
Compress would make output size larger when evaluating const expression


Austaras:
### Describe the bug

This would require implementing `terser` `size` method.

### Input code

```typescript
console.log(16 ** 16)
```

### Playground link

https://play.swc.rs/?version=1.2.170&code=H4sIAAAAAAAAA0vOzyvOz0nVy8lP1zA0U9DSUjA00wQAmFk5txUAAAA%3D&config=H4sIAAAAAAAAA22TzZLbIAzH38XnHLqdaQ%2F7AL3tMzAKCEdbAx4ksvFk8u4VthfbSW%2Fox19CX9y7T7bd%2B70bITPmeuIpCty69w5tALaZRulOKlPkYWB8nDqB3KNUCf%2F88fZLr4eUGFfBqQsUyU81mE1hzMhcz%2BpUAkbhpoOc05eakota55QGhPhiG2BDUbDHvLnaNAwwMpor5OZRX4NMnOIBFUFnxpzGvXt0JJSimg06BGdsclshjjJaoSvuRBpHRZE1uSN0eC59X5u4PI1XGAoINvs2t0IfbX6XRCzGl8hPaM12cVzQUuiqIm8ySsnxW%2FOZKB568RdR8xyAOULY5T9zr7M4Kv1%2FVBS9DlKmRnXKW1YRey3OEPlWYc0asxA2UUZXLNYK7RZ8hYeCmBwa9F7b3Zz5i8RedlnJNGLy7V7bBn4bwmKa76U60roHL%2FCPpiX7eaw8gFyeGU%2FhnIaXEAHlktwL1k5IeoZZV%2Fg2PtMSHWqf0e0uClezVqm%2FLUDs2649KkiuVHCfGzJ%2Fw9%2FdY%2Ft2S3eIP1bdHOcfmKDn2%2BoDAAA%3D

### Expected behavior

```js
console.log(16**16)
```

### Actual behavior

```js
console.log(18446744073709552e3)
```

### Version

1.2.170

kdy1:
But for the js engine, it's much faster to parse 18446744073709552e3 than 16**16

PR #5232, opened 2022-07-17
fix(es/optimiziation): size calc for number

### LINKED ISSUE #5202: minifier should optimize `-(-0.5 * 0.5)`
This is follow-up of https://github.com/swc-project/swc/pull/5201

Ideally it should be `0.25`, but at least it should be `0.5 * 0.5`

Austaras:
**Description:**

This issue raise from `expr_simplifier` wrong way of calculating size of `BinExpr`. It thinks the size of `BinExpr` is `right.span.hi` -  `left.span.lo` which seems to no problem at first glance. However, if both left and right is substitute with same variable, they would have same span, so only the size of inlined number literal is calculated. This raises further need for #4414, since there's now not only one size calculator in swc, but two, and both of them are broken.

**BREAKING CHANGE:**

closes #5202


kdy1 (approved):
Thanks!

---

swc-bump:

crates/swc_ecma_transforms_optimization/src/simplify/expr/tests.rs
```diff
@@ -786,7 +786,7 @@ fn test_fold_arithmetic() {
     fold("x = -1 % 3", "x = -1");
     fold_same("x = 1 % 0");
     fold("x = 2 ** 3", "x = 8");
-    fold("x = 2 ** -3", "x = 0.125");
+    // fold("x = 2 ** -3", "x = 0.125");
```
Austaras, line 789: Maybe add it back later

kdy1 (approved):

PR #5235, opened 2022-07-17
docs(transforms/hygiene): `hygiene` consists of four phases. -> hygiene` consists of three phases.

7213:
**Description:**

HI, hygiene's work should be three stages in terms of source code and follow-up comments, I think it should be a typo

**BREAKING CHANGE:**
no

no


7213:
sorry, this commit record is not clean

PR #5237, opened 2022-07-17
fix(testing): Allow space in file names

kdy1:
**Description:**

**BREAKING CHANGE:**


kdy1 (commented):
swc-bump:
 - testing_macros

PR #5242, opened 2022-07-18
feat(testing): Generalize test_name with regex

orionmiz:
**Description:**

This PR applies a regex that converts non-alphanumeric characters into underscores to test_name.

It resolves the error that occurs when the path of the fixture file contains special characters the existing source did not process individually.

**BREAKING CHANGE: X**

### DIFF
--- a/crates/testing_macros/src/fixture.rs
+++ b/crates/testing_macros/src/fixture.rs
@@ -107,7 +107,7 @@ pub fn expand(callee: &Ident, attr: Config) -> Result<Vec<ItemFn>, Error> {
         glob(&pattern).with_context(|| format!("glob failed for whole path: `{}`", pattern))?;
     let mut test_fns = vec![];
     // Allow only alphanumeric and underscore characters for the test_name.
-    let re = Lazy::new(|| Regex::new(r"[^A-Za-z0-9_]").unwrap());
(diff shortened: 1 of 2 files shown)


crates/testing_macros/src/fixture.rs
```diff
@@ -105,6 +106,8 @@ pub fn expand(callee: &Ident, attr: Config) -> Result<Vec<ItemFn>, Error> {
     let paths =
         glob(&pattern).with_context(|| format!("glob failed for whole path: `{}`", pattern))?;
     let mut test_fns = vec![];
+    // Allow only alphanumeric and underscore characters for the test_name.
+    let re = Lazy::new(|| Regex::new(r"[^A-Za-z0-9_]").unwrap());
```
kdy1, line 110: Please make this static

kdy1 (changes requested):

crates/testing_macros/src/fixture.rs
kdy1: https://docs.rs/once_cell/latest/once_cell/#lazy-initialized-global-data

kdy1 (approved):
Thank you!

---

swc-bump:
 - testing_macros

PR #5247, opened 2022-07-19
feat(css/prefixer): env option

alexander-akait:
**Description:**

Added `env` option to support browsers, like we do in `swc_emca_present_env` +  added many properties.
In the next PRs will add:
1. at-rules 
2. selectors
3. values

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_css_prefixer --breaking

PR #5253, opened 2022-07-19
feat(es/minifier): size calculator

Austaras:
**Description:**

**BREAKING CHANGE:**


crates/swc_ecma_minifier/tests/single-pass/1/output.js
```diff
@@ -6,8 +6,6 @@ var condString, exprAny1, exprBoolean1, exprNumber1, exprString1, exprIsObject1,
     "2",
     "3"
 ];
-condString.toUpperCase;
-array[1];
-condString.toUpperCase;
-array[1];
+condString.toUpperCase, foo(), array[1], foo();
```
Austaras, line 9: related issue tersr/terser#1227

We could mark function return literal as pure though

crates/swc_ecma_minifier/tests/projects/output/react-dom-17.0.2.js
```diff
@@ -5537,12 +5531,6 @@
             dispatch
         ];
     }
-    function updateState(initialState) {
```
Austaras, line 5540: This is actually problematic, it makes output size large.

It's because `initialState` doesn't actually be passed, so `cost_limit` is incorrectly calculated as 3 + 2.

crates/swc/tests/tsc-references/logicalNotOperatorWithStringType_es2015.2.minified.js
```diff
@@ -1,3 +1,6 @@
+function foo() {
```
kdy1, line 1: Shouldn't we add the cost for `function` + `()` + `{}`?

kdy1: If we decide to do so, we should also check the number of function calls (by multiplying them)

kdy1 (commented):
Just thoughts

crates/swc/tests/tsc-references/logicalNotOperatorWithStringType_es2015.2.minified.js
Austaras: Sure! I've already written a separate PR for this optimization, just some details needs to be polished.

crates/swc_ecma_minifier/src/util/size.rs
```diff
@@ -0,0 +1,355 @@
...
+    }
+}
+
+impl SizeWithCtxt for Pat {
+    fn size(&self, unresolved: SyntaxContext) -> usize {
+        match self {
+            Pat::Ident(id) => id.size(unresolved),
+            Pat::Array(a) => 2 + a.elems.size(unresolved),
+            Pat::Rest(r) => 3 + r.arg.size(unresolved),
+            Pat::Object(_) => todo!(),
```
kdy1, line 263: I think this should not panic

kdy1 (changes requested):

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_minifier

PR #5254, opened 2022-07-19
feat(swc/plugin): add experimental metadaata field

kwonoj:
**Description:**

This is part of stabilization effort for the plugin across versions. 

For now, we freely add new metadata property if needed. This is fine as we are under experimental and can break things, but will not be available once we stablize plugin interface. Unlike AST struct and other data crosses over runtime (host to plugin), `TransformPluginProgramMetadata` is constructed internally in plugin's memory space only and does not have any support for the schema evolution. If we add new property in this metadata and `@swc/core` uses it, plugin should be updated to use sdk including those change - and it's not backward compatible as well, so plugin will hit hard breaking changes between 2 host versions.

PR adds one permananet hashmap as `experimental`, which host can add / remove any necessary metadata while we  trying to validate its usecases. It doesn't support typed interface to avoid any breaking, so all of actual data is stringifed serialization data. While experimenting, plugin need to manually deseralize per-data based. Once we can confirm it's useful enough, it'll be included as actual property and we can declare breaking changes for the plugin from that point.

This may not be an ideal approach (especially lacks of typed deserialization support), but I expect it'll work reasonbly & we won't have frequent changes to the metadata struct anyway.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_plugin_macro --breaking

kdy1 (changes requested):
I found a merge conflict

kwonoj:
> I found a merge conflict

Yup, should be resolved I believe.

kwonoj:
```
Error: .github/workflows/cargo.yml (Line: 396, Col: 16):
Error: The template is not valid. .github/workflows/cargo.yml (Line: 396, Col: 16): hashFiles('**/Cargo.lock') couldn't finish within 1[2](https://github.com/swc-project/swc/runs/7422619375?check_suite_focus=true#step:34:2)0 seconds.
```

Seems this occurs intermittently but unsure why this started occurring recently.

kdy1:
Hmm.... CI instance is super slow?

kwonoj:
> Hmm.... CI instance is super slow?

Not sure honestly. Hashing lockfile taking more than 120 sec sounds fishy indeed though.

kdy1 (approved):

PR #5255, opened 2022-07-19
test(swc): test plugins with different schema versions (#5060 Part 6)

kwonoj:
**Description:**

Continuation of https://github.com/swc-project/swc/pull/5243, where actually establishes versioned schema compatiblity verification. Good new is, it works well. Not good, or somewhat annoying thing is current ergonomics is not elegant for the core devs. (Plugin authors or users are fine!)

First, let's start with compatibility table. `1` and `2` indicate **AST VERSION**, not actual major pkg versions of each.

|   |  plugin@1  | plugin@2  |
|---|---|---|
| @swc/core@1  |  ⭕  | ❌   | 
| @swc/core@2  | ⭕   | ⭕  | 

This is our desired behavior. Say we add a new AST property and it increases AST version to 2, we expect new `@swc/core` can seamleassly use plugin@1, 2 both without asking to rebuild. In case of old version of `@swc/core@`, it won't be able to read newly updated plugin.

The thing I originally tried to do was `@swc/core` itself can be selectively compiled based on the versions, so `rkyv` can choose specific type to be serialized / deserialized. But actually, rkyv was able to fluently serialize / deserialze backward-compatible types. To test those, PR updates tests with adding a dummy property to top-level AST (`Program`) to ensure compile time compatibility as well as runtime compatibility for desired 3 working cases. 

Problem is core's development ergonomics - since AST update is detached to AST version, this is the way we need when we add new property in the AST struct

1. Properly add new property into some of AST struct
2. Update / create new `plugin_transform_schema_v${new}` feature
3. Change default to `v${current}` to `v${new}`

Lastly, when we release a new breaking changes we can safely remove old `v*` flags except latest current one. (We'll ever increment only for the version, so we'll not reset current).

It's not great, but not too bad as well in my opinion. However, would love to hear thoughts. Once this is approved, I'll try to write down some instruction documentation for the compatibility table with workflows.

**BREAKING CHANGE:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
         "@babel/preset-typescript": "^7.13.0",
         "@babel/types": "^7.14.0",
         "@napi-rs/cli": "^2.10.0",
+        "@swc/core": "^1.2.218",
         "@swc/helpers": "^0.4.2",
         "@taplo/cli": "^0.3.2",
         "@types/jest": "^28.1.4",
--- a/crates/swc_ecma_ast/src/lib.rs
+++ b/crates/swc_ecma_ast/src/lib.rs
@@ -37,7 +37,7 @@ pub use self::{
     },
     list::ListFormat,
     lit::{BigInt, Bool, Lit, Null, Number, Regex, Str},
-    module::{Module, ModuleItem, Program, Script},
+    module::{Module, ModuleItem, Program, ReservedUnused, Script},
     module_decl::{
         DefaultDecl, ExportAll, ExportDecl, ExportDefaultDecl, ExportDefaultExpr,
         ExportDefaultSpecifier, ExportNamedSpecifier, ExportNamespaceSpecifier, ExportSpecifier,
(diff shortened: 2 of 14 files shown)


kwonoj:
Lastly, I do not expect we'll try to support `@swc/core@1` can run `plugin@2`, but I may wrong?

package.json
```diff
@@ -96,6 +96,7 @@
         "@babel/preset-typescript": "^7.13.0",
         "@babel/types": "^7.14.0",
         "@napi-rs/cli": "^2.10.0",
+        "@swc/core": "^1.2.218",
```
kdy1, line 99: Will this work with npm publish?

kwonoj: I haven't able to actually publish, but I guess it may ok since this is devdependencies. Unfortunately I don't have good way to verify it other than `publish --dry-run` (which seems to pass)

kdy1: Then I think it's fine. I didn't know node js supports cycle in dev deps, though.

kdy1 (approved):
Thank you!


---

swc-bump:
 - swc_ecma_ast --breaking

PR #5256, opened 2022-07-20
fix(html/minifier): don't break unknown attributes

alexander-akait:
**Description:**

Fix - don't change unknwo attributes +
+ remove legacy attributes which were removed from spec
+ collect all html elements for future improvements
+ more default attributes

**BREAKING CHANGE:**

Yes, due rename and refactor `swc_html_utils`

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_utils --breaking

PR #5261, opened 2022-07-21
feat(swc/plugin): actually pass experimental metadata from host

kwonoj:
**Description:**

PR actually passes experimental metadata from host to the plugin, even though it's empty hashmap for now.

There is a ugly commit https://github.com/swc-project/swc/commit/5ef7333fbf931aaf3e0bc7e317955fd84d7e1f3f to disable test-only ast changes and some test, for the reason our e2e test is now validly test against latest public and this PR is _technically_ breaking change and our test fails.

It is not great, probably we have to invent something works automatically in the future. For now, dealing with manual approach.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_ast --breaking
 - swc_plugin_macro --breaking

PR #5271, opened 2022-07-22
perf(es/ast): Use `Atom` in some places

kdy1:
**Description:**

As this is tradeoff of comparison/hash time and global lock, I decided to use `Atom` for only places I can be sure.


kdy1 (commented):
swc-bump:
 - swc_atoms --breaking

kdy1:
cc @overlookmotel Can you check if the mutex is still the blocker for the performance?

kwonoj (commented):
If we could have some 10,000ft level docs describe this it'd be helpful for the future, mainly

- the main difference between new (atom) vs legacy approach
- what atom exactly does & why it performs better for certain usecases
- judge case scenarios for the future code when to use atom vs. when to use others

(Yes, aware there is a comment https://github.com/swc-project/swc/pull/5271/files#diff-11618d446401b8dc9d8092ad6ad3703d4bf959007bc4c22770aedf82f476eb53R36, but for example how long is too long, etcs)

overlookmotel:
Argh! I'm too late to the party. You move so damn fast!

I had an idea about how it might be possible to combine the advantages of `JsWord` and the new `Atom` in a single entity.

In short: have a shared-across-threads static set (read only, so no mutex required) pre-populated with common words + a per-thread dynamic set (for words not in the static set). The static set could use perfect hashes, and the dynamic set not. Both could use the faster [FxHasher](https://doc.rust-lang.org/stable/nightly-rustc/rustc_data_structures/fx/struct.FxHasher.html).

However, work is completely crazy today so it'll have to wait until weekend for me to outline this idea further. I'll open an issue as soon as I have time.

kdy1:
I merged it after verifying the performance of other operations, but yeah, I only used `Atom` for places where we don't need perfect hashing

overlookmotel:
Quick thought: Since the `Atom`s are never shared across threads, couldn't they use `Rc` instead of `Arc`? My understanding is `Rc` can have less overhead.

kdy1:
Nope. they are moved across threads

PR #5294, opened 2022-07-25
fix(css/prefixer): generate at-rules based on env

alexander-akait:
**Description:**

Don't generate unnecessary `at-rules` based on `env`

**BREAKING CHANGE:**

No

No


crates/swc_css_prefixer/src/prefixer.rs
```diff
@@ -553,6 +553,16 @@ struct Prefixer {
     added_declarations: Vec<Declaration>,
 }
 
+impl Prefixer {
+    fn add_at_rule(&mut self, prefix: Prefix, at_rule: AtRule) {
```
kdy1, line 557: I think `at_rule` should be `&AtRule`

crates/swc_css_prefixer/src/prefixer.rs
```diff
@@ -594,85 +604,87 @@ impl VisitMut for Prefixer {
...
+                        Prefix::Ms,
+                        AtRule {
+                            span: DUMMY_SP,
+                            name: AtRuleName::Ident(Ident {
+                                span: DUMMY_SP,
+                                value: "-ms-viewport".into(),
+                                raw: None,
+                            }),
+                            prelude: n.prelude.clone(),
+                            block: original_simple_block.clone(),
```
kdy1, line 618: So we clone only if required

kdy1 (changes requested):
Thanks!

alexander-akait:
@kdy1 Fixed :+1:

kdy1 (approved):
Thanks!

---
swc-bump:
 - swc_css_prefixer

PR #5295, opened 2022-07-25
fix(es/codegen): extra space before private method

Austaras:
**Description:**

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!


---


swc-bump:
 - swc_ecma_codegen

PR #5305, opened 2022-07-26
fix(es/preset-env): don't add instance properties when there's static properties

Austaras:
**Description:**

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_preset_env

Issue #5281, opened 2022-07-24
Make `TransformPluginProgramMetadata::transform_context` lazy


kwonoj:
### Describe the feature

This isn't worth RFC in my opinion, a small proposal to `TransformPluginProgramMetadata`.

Currently global context metadata to the plugin (`transform_context`, `experimental`) is eagerly passed into plugin's context per each plugin execution. We limit surface of `transform_context` at this moment, but as it'll grow it can be a burden for the serialization / deserialization.

Instead, as same as other metadata does (`comments`, `source_map`) propose to change it to be lazely requested from plugin only if needed. Same as comments / source_map it doesn't have any memoization and each call will requires fn invocation to call host side serialization / deserialization, but these are mostly static context avoid those repeated calls fairly easy.

Proposed pseudo interface will looks like below:

```

enum TransformPluginContext {
  FilePath,
  ...
}

impl TransformPluginProgramMetadata {
  // Returns specific context metadata corresponding to the key.
  fn get_context(context_key: TransformPluginContext): String;
  // Returns all context metadata as hashMap
  fn get_context_map(): HashMap<TransformPluginContext, String>;
  // Returns experimental context 
  fn get_experimental_context(context_key: &Str): Option<String>;
  fn get_experimental_context_map(): HashMap<String, String>;
}
```

get_context* represents current `transform_context`. This is a typed interface with known set of enum key, in result gaurantee to return value instead of `Option<T>`. But it still doesn't support return type deserialization - it'll be returned as string either can be plain string or JSON-stringified depends on the type of the context for the simplicity of the interface.

get_experimental* represents newly added experimental property. This is untyped interface to allow non-semver breaking changes.

kwonoj:
This'll effectively makes to remove existing eager property (`transform_context` and `experimental`) as well.

Thoughts? @kdy1

kdy1:
I think this way is better than passing them to all plugins.
(Actually I considered suggesting something similar when reviewing PR for experimental metadata.)

kwonoj:
Let's do it then 🚢 :shipit: 

I hope this could be the last piece of breaking changes before stabilization (but.. who knows? ¯\_(ツ)_/¯)

PR #5310, opened 2022-07-26
feat(swc/plugin_runner): allow to read metadata lazily

kwonoj:
**Description:**

- closes https://github.com/swc-project/swc/issues/5281

This PR implements interfaces for `TransformPluginProgramMetadata` to be able to request metadata context lazily. This is a clear breaking change since the old eager, static values are gone. 

Initial interfaces are exposed as below. 

```
fn get_transform_plugin_config() -> Option<String>;
fn get_context(key: &TransformPluginMetadataContextKind) -> Option<String>;
fn get_experimental_context(key: &str) -> Option<String>;
```

`get_transform_plugin_config` returns given transform plugin's config. Since this is being deserialized from `serde_json::Value`, host pass whole stringified JSON as is. Once plugin receives it, deserialization is up to plugin as same as before.

`get_context` replaces `transform_context` values. This accepts known list of the indexer for the values (`TransformPluginMetadataContextKind`) for predefined values. Adding, or removing new values for this context will be a breaking change.

`get_experimental_context` replaces recently introduced experimental metadata. This'll be a placeholder we'll add / remove new metadata values without causing breaking change.

For getting context / expeirmental both, current interface requires to specify specific key for the specific value. We may need to have getting all context at once - but we can add new interface always once we have solid usecase, so decided to not to have those as first step.

All of these functions are lazy, as well as will involve communication between host to the plugin with serialization & deserialization. Plugin should own responsibility to memoize values to avoid redundant costs.

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc/src/config/mod.rs
+++ b/crates/swc/src/config/mod.rs
@@ -24,6 +24,7 @@ use swc_common::{
     collections::{AHashMap, AHashSet},
     comments::SingleThreadedComments,
     errors::Handler,
+    plugin::metadata::TransformPluginMetadataContext,
     FileName, Mark, SourceMap, SyntaxContext,
 };
 use swc_config::{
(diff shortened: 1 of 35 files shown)


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_common --breaking

PR #5312, opened 2022-07-26
fix(html/parser): span

alexander-akait:
**Description:**

Fix span for adoption agency algorithm

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!


---

swc-bump:
 - swc_html_parser

PR #5320, opened 2022-07-27
feat(html/minifier): compress title

alexander-akait:
**Description:**

Compress title element, from spec:

```
The title attribute must, on getting, run the following algorithm:

If the [document element](https://dom.spec.whatwg.org/#document-element) is an [SVG svg](https://svgwg.org/svg2-draft/struct.html#SVGElement) element, then let value be the [child text content](https://dom.spec.whatwg.org/#concept-child-text-content) of the first [SVG title](https://svgwg.org/svg2-draft/struct.html#TitleElement) element that is a child of the [document element](https://dom.spec.whatwg.org/#document-element).

Otherwise, let value be the [child text content](https://dom.spec.whatwg.org/#concept-child-text-content) of [the title element](https://html.spec.whatwg.org/multipage/dom.html#the-title-element-2), or the empty string if [the title element](https://html.spec.whatwg.org/multipage/dom.html#the-title-element-2) is null.

[Strip and collapse ASCII whitespace](https://infra.spec.whatwg.org/#strip-and-collapse-ascii-whitespace) in value.

Return value.
````

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_html_minifier

PR #5321, opened 2022-07-27
fix(css/parser): list of compoments values

alexander-akait:
**Description:**

Fix AST, because CSS is grammar free, list of components values are valid and may used for future syntax

It requires more fixes, but it is a start

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thank you!

CSS is much more complex than I initially thought...

---


swc-bump:
 - swc_css_ast --breaking

PR #5322, opened 2022-07-27
build(package): include fallback binding as dep

kwonoj:
**Description:**

This PR adds a step to publishing step,  for the targets we'd like to use `@swc/wasm` to have it as dependency. This removes friction to postinstall failures depends on the local setup. Once the time comes for the migration, we can simply remove native bindings, then leave pkg as-is so installation process won't change.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - ''

kwonoj:
@kdy1 fyi, CI failure seems to be https://github.com/swc-project/swc/pull/5324

kdy1 (approved):

PR #5324, opened 2022-07-27
test(swc): update test cases

kwonoj:
**Description:**

Update test cases with recent changes to the swc core to keep things up. This is a canonical example test fails if we hit some unexpected breaking change (which is good). For now, moving on by updating tests.

Actual failures looks like https://github.com/swc-project/swc/runs/7546769441?check_suite_focus=true.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - ''

PR #5169, opened 2022-07-10
refactor(es/module): SystemJS

magic-akari:
**Description:**

SystemJS is totally different from CJS/AMD/UMD and I can't define a getter on exports because it uses the `export()` function.
But I can use the same trick of class private field updater.
I can use getter/setter defined bindings to simplify update expressions.

**BREAKING CHANGE:**

**Related issue (if exists):**
- fix #4895

PR #5327, opened 2022-07-28
fix(es/utils): `contains_top_level_await`

magic-akari:
**Description:**

`contains_top_level_await` does not handle all cases.

**BREAKING CHANGE:**

- block #5169


kdy1 (dismissed):
Thanks!

---

swc-bump:
 - swc_ecma_utils

kdy1:
cc @kwonoj
Seems like the fix for the `Atom` rkyv issue was a breaking change for plugins?

kwonoj:
> Seems like the fix for the Atom rkyv issue was a breaking change for plugins?

Yes, I think so.

PR #5331, opened 2022-07-28
build(cargo): update packages

kwonoj:
**Description:**

Attempt to fix https://github.com/swc-project/swc/pull/5327#issuecomment-1197978046

Note PR should still fail - once new `swc_plugin` is published, there can be another PR to fix it actually.

Yes, again this is not a great experience (even though we hope this won't happen frequently) and we'll need to find better way (at least trying to fail before change lands on master). I'll think about something.

**BREAKING CHANGE:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4117,7 +4117,7 @@ dependencies = [
 
 [[package]]
 name = "swc_plugin"
-version = "0.88.1"
+version = "0.88.2"
 dependencies = [
  "swc_atoms",
  "swc_common",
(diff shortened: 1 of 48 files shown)


kdy1:
Seems like CI is still failing

kwonoj:
@kdy1 I stated it in PR:

> Note PR should still fail - once new swc_plugin is published, there can be another PR to fix it actually.

kdy1 (approved):
Oh, my bad

---

swc-bump:
 - ''

PR #5333, opened 2022-07-29
fix(es/minifier): Change default es version to `es5`

kdy1:
**Description:**

 - https://github.com/vercel/next.js/issues/39090


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5341, opened 2022-07-29
feat(swc/wasm): typed interface support for wasm bindings

kwonoj:
**Description:**

This PR improves type definition for `@swc/wasm` by mimicing existing type interface support to `@swc/core`. Unfortunately wasm-bindgen does not support to directly reference types from existing typescript or build-time macro to import it, so had to duplicate those types. Currently it's 100% duplicate, so there are some options can be passed to bindings even if it's not actually supported (i.e plugins). Something to improve further.

This helps to fix `@swc/core`'s fallback binding behavior to prevent pass incorrect params to the fallback bindings functions.

**BREAKING CHANGE:**

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -26,6 +26,9 @@ jobs:
           restore-keys: |
             cargo-release-${{ runner.os }}-${{ hashFiles('**/Cargo.lock') }}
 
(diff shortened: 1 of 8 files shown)


kdy1 (approved):
Thanks!

---
swc-bump:
 - ''

PR #5351, opened 2022-07-31
feat(swc/wasm): enable interning strings

kwonoj:
**Description:**

This PR enables interning string feature from wasm-bindgen (https://github.com/rustwasm/wasm-bindgen/blob/062aa5f70a1e8a170d0ae90123ebdeb3242ac782/CHANGELOG.md) . This is not well exposed in the docs somehow.

While it _sounds_ like a nice improvement, it is not clear if it'll give a noticeable gain for us. I felt it's worth trying if we haven't done this before.

I tried quick search but I may miss history if we did this & revert. If that's the case, please feel free to close.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - ''

PR #5352, opened 2022-08-01
feat(swc/wasm): expose async facade interfaces

kwonoj:
**Description:**

This PR implements, exposes counterpart of `*Sync` interfaces to `@swc/wasm`, as same as `@swc/core`. It'll allow `@swc/wasm` becomes near-complete drop in replacement to `@swc/core` as much as possible.

One important thing to note is, even though those are async interfaces there aren't any actual async implementation in place. In short, this is pretty much similar to JS-side facade wrap like `new Promise(() => swcSync())`, which only wraps sync interfaces to async. The only difference is it is exposed in bindgen level, so we don't need to come up with some custom binding wrapper to make things work.

Once there is support for the thread / async in wasm with standard support we'll update those internals accordingly. We do not expect to try some workaround approach like using bindgen with workers, etcs at this moment.

**BREAKING CHANGE:**

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -5,3 +5,4 @@ imports_granularity = "Crate"
 reorder_impl_items = true
 use_field_init_shorthand = true
 wrap_comments = true
+edition = "2018"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -217,6 +217,7 @@ dependencies = [
  "swc_plugin_runner",
  "tracing",
  "wasm-bindgen",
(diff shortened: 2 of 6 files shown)


kdy1 (approved):
Thanks!

---
swc-bump:
 - swc

kdy1 (changes requested):
There's a conflict

kwonoj:
@kdy1 yup, should be resolved by now.

kdy1 (approved):

PR #5356, opened 2022-08-01
refactor(swc): use async interfaces for fallback bindings

kwonoj:
**Description:**

We can use async intefaces for the fallback bindings now.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - ''

PR #5361, opened 2022-08-02
fix(es/minifier): Drop more comments

kdy1:
**Description:**

- https://github.com/vercel/next.js/discussions/30237?sort=new#discussioncomment-3303141


kdy1 (commented):
swc-bump:
 - swc

PR #5364, opened 2022-08-02
feat(swc/core): introduce swc_core package

kwonoj:
**Description:**

First step implementation for https://github.com/swc-project/swc/discussions/5244.

PR introduces new package `swc_core`, then move `swc_plugin`'s feature under the `swc_core` as first step. This is a breaking change to any plugins that we'll need to update references accordingly.

The idea of `swc_core` is simple, that `mostly reexport all the features, without having its own code as much as it can`.
I'd like to setup these principles for those, while it may / should change depends on how we'll going to make it work actually.

- single package to import : plugin author, or custom binary author should not need to import any `swc_*` dependencies other than `swc_core`. (We do not care about any other 3rd party deps though, like serde, etcs)
- no default features: There are too many conflicting features to be exposed (i.e plugin vs. host runtime) and it's not easy to bring common set of default features with its transitive deps. We'll start off from no default until we see valid usecase.
- Complete opt-in: all deps are trying to be optional as much, and correct feature opt in should enable those.
- Compile time validatable feature dependencies: if A needs A' or A'', better to validate in build time.
- Flattened dep tree: this is probably moon shot, but hoping to reduce circular references / complex transitive deps and making it flatten as much to optimize build.
- Transparent forward: instead of trying to re-categorize, attempt to reexport transparently. For example: `swc_common::*` -> `swc_core::common::*`.

This is probably not the complete `swc_plugin` support, may need to check with existing plugins to find out missing exports. Still it is a first step.

**BREAKING CHANGE:**

 https://github.com/swc-project/swc/discussions/5244


crates/swc_core/src/lib.rs
```diff
@@ -0,0 +1,32 @@
+#![cfg_attr(docsrs, feature(doc_cfg))]
+
+// Quote
+#[cfg(feature = "quote")]
```
kdy1, line 4: I think we need `#[cfg(any(docsrs, feature = "quote")]` along with `#[cfg_attr(docsrs, doc(cfg(feature = "quote")))]`

crates/swc_core/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+pub mod quote {
+    pub use swc_ecma_quote::*;
+}
+
+// Plugins
+#[cfg(feature = "__plugin_transform")]
```
kdy1, line 10: Same here

crates/swc_core/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+pub mod plugin;
+
+// ast exposed via swc_ecma_ast
+// TODO: Can dependency tree simplified
+// by swc_ecma_ast reexports swc_atoms?
+#[cfg(any(feature = "__ast"))]
```
kdy1, line 16: Same here

crates/swc_core/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+pub mod ast {
+    pub use swc_atoms::*;
+    pub use swc_ecma_ast::*;
+}
+
+// visit* interfaces
+#[cfg(any(feature = "__visit"))]
```
kdy1, line 23: Same here

crates/swc_core/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+pub mod visit {
+    pub use swc_ecma_visit::*;
+}
+
+// swc_common features
+#[cfg(any(feature = "__common"))]
```
kdy1, line 29: Same here

crates/swc_core/src/plugin.rs
```diff
@@ -0,0 +1,36 @@
+// #[plugin_transform] macro
+#[cfg(feature = "__plugin_transform")]
+pub use swc_plugin_macro::plugin_transform;
```
kdy1, line 3: Same here

crates/swc_core/src/plugin.rs
```diff
@@ -0,0 +1,36 @@
...
+/// allows to allocate memory from the host side.
+/// This should not be directly referenced.
+#[cfg(all(feature = "__plugin_transform", target_arch = "wasm32"))]
+pub mod memory {
+    pub use swc_plugin::allocation::*;
+}
+
+/// Global HANDLER implementation for the plugin
+/// for error reporting.
+#[cfg(feature = "__plugin_transform")]
```
kdy1, line 15: Same here

crates/swc_core/src/plugin.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// global context HANDLER in plugin's transform function.
+    pub static HANDLER: swc_plugin::pseudo_scoped_key::PseudoScopedKey<
+        swc_common::errors::Handler,
+    > = swc_plugin::pseudo_scoped_key::PseudoScopedKey {
+        inner: once_cell::sync::OnceCell::new(),
+    };
+}
+
+/// Plugin's environment metadata context.
+#[cfg(feature = "__plugin_transform")]
```
kdy1, line 26: Same here

crates/swc_core/src/plugin.rs
```diff
@@ -0,0 +1,36 @@
...
+pub mod metadata {
+    pub use swc_common::plugin::metadata::TransformPluginMetadataContextKind;
+    pub use swc_plugin_proxy::TransformPluginProgramMetadata;
+}
+
+/// Proxy to the host's data not attached to the AST, like sourcemap / comments.
+#[cfg(feature = "__plugin_transform")]
```
kdy1, line 33: Same here

kdy1 (changes requested):
Thanks!

crates/swc_core/src/lib.rs
```diff
@@ -0,0 +1,32 @@
+#![cfg_attr(docsrs, feature(doc_cfg))]
+
+// Quote
+#[cfg(any(docsrs, feature = "quote"))]
```
kdy1, line 4: We also need `cfg_attr`

kwonoj: is this (https://github.com/swc-project/swc/pull/5364/files#diff-640c564973380a758182cae3dc7d5ea4e31bc82aba03d3b3dfb383273bf93f1dR5) right? and should apply to other places if it's correct?

kdy1: Yes, you can verify by using `cargo doc`
(We specifies `--cfg docsrs` for rustdoc using cargo config toml file)

crates/swc_core/Cargo.toml
```diff
@@ -0,0 +1,87 @@
...
+  # Enable optional packages
+  "swc_common/plugin-mode",
+  "swc_plugin_proxy/plugin-mode",
+  "swc_plugin_macro",
+  "swc_plugin",
+  "once_cell",
+]
+
+## Common
+__common = ["swc_common"]
```
kdy1, line 71: I think this should depend on `swc_atoms` too.
(`swc_common` depends on `swc_atoms`

kdy1: Nevermind. User should depend on `__ast` anyway

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_plugin_macro --breaking

crates/swc_core/Cargo.toml
kwonoj: We don't need to list all the transitive deps to the top level for the specific feature, isn't it? We can easily add later once we find those dependencies around reexports.

PR #5342, opened 2022-07-30
fix(es/minifier): calc function body cost when inline function

Austaras:
**Description:**

A follow up of #5253.

**BREAKING CHANGE:**

**Related issue (if exists):**

Issue #4415, opened 2022-04-23
Evaluate expression when compress conditional statement


Austaras:
### Describe the bug

This would require implementing `terser` `evaluate` method.

### Input code

```typescript
let a = 'aaaa'

if (a === 'aaaa') {
	console.log(a)
}
```

### Playground link

https://play.swc.rs/?version=1.2.170&code=H4sIAAAAAAAAA8tJLVFIVLBVUE8EAnUursw0BQ0g3xYmoqlQzcWZnJ9XnJ%2BTqpeTn66RqMlVCwBln0kENQAAAA%3D%3D&config=H4sIAAAAAAAAA22TzZLbIAzH38XnHLqdaQ%2F7AL3tMzAKCEdbAx4ksvFk8u4VthfbSW%2Fox19CX9y7T7bd%2B70bITPmeuIpCty69w5tALaZRulOKlPkYWB8nDqB3KNUCf%2F88fZLr4eUGFfBqQsUyU81mE1hzMhcz%2BpUAkbhpoOc05eakota55QGhPhiG2BDUbDHvLnaNAwwMpor5OZRX4NMnOIBFUFnxpzGvXt0JJSimg06BGdsclshjjJaoSvuRBpHRZE1uSN0eC59X5u4PI1XGAoINvs2t0IfbX6XRCzGl8hPaM12cVzQUuiqIm8ySsnxW%2FOZKB568RdR8xyAOULY5T9zr7M4Kv1%2FVBS9DlKmRnXKW1YRey3OEPlWYc0asxA2UUZXLNYK7RZ8hYeCmBwa9F7b3Zz5i8RedlnJNGLy7V7bBn4bwmKa76U60roHL%2FCPpiX7eaw8gFyeGU%2FhnIaXEAHlktwL1k5IeoZZV%2Fg2PtMSHWqf0e0uClezVqm%2FLUDs2649KkiuVHCfGzJ%2Fw9%2FdY%2Ft2S3eIP1bdHOcfmKDn2%2BoDAAA%3D

### Expected behavior

```js
console.log('aaaa')
```

### Actual behavior

```js
let a="aaaa";"aaaa"===a&&console.log(a)
```

### Version

1.2.170

kdy1:
We have inliner,

[...]

PR #5365, opened 2022-08-02
fix(es/minifier): workaround for inline

Austaras:
**Description:**

This raises from the dicussion of #5342. Currently, swc cannot handle following code
```js
var a;
a = 10; console.log(a)
```
which is due to #4415, but `inline` pass would generate lots of such pattern of code.

So I propose an alternative. For
```js
function a(a, b) {
    return ((a % b) + b) % b;
}
console.log(a(10, 15));
```
instead of 
```js
var a, b
console.log(((a=10) % (b=15) + b) % b);
```
do
```js
var a = 10, b = 15
console.log(((a % b) + b) % b);
```
**BREAKING CHANGE:**


crates/swc_ecma_minifier/tests/projects/output/react-17.0.1.js
```diff
@@ -136,13 +136,14 @@
             case REACT_PROVIDER_TYPE:
                 return getContextName(type._context) + ".Provider";
             case REACT_FORWARD_REF_TYPE:
-                return outerType = type, innerType = type.render, wrapperName = "ForwardRef", functionName = innerType.displayName || innerType.name || "", outerType.displayName || ("" !== functionName ? wrapperName + "(" + functionName + ")" : wrapperName);
```
Austaras, line 139: Thish seems to be the only regression -- mainly due to not many people use `switch` today -- which only adds `4 + 1 - 1` byte

kdy1 (approved):
I think the idea was so clever, thank you!

---
swc-bump:
 - swc_ecma_minifier

Austaras:
Seems like some tests is stagnant, let me manually rebase this...

kdy1 (approved):

PR #5368, opened 2022-08-02
build(cargo): update toml formatting

kwonoj:
**Description:**

This is pure cosmetic changes, enable few options for the formatter (taplo) for toml formatting. Mostly I wanted to enable align entries, so I can easily compare each entry's values. See formatted Cargo.toml in PR as an example.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---
swc-bump:
 - ''

PR #5369, opened 2022-08-02
feat(swc/core): expand features, update ref to wasm / cli

kwonoj:
**Description:**
PR improves feature surface bit more, then wasm / cli actually started to import this. (Not completed)

Since I'm not intend to create huge dent to existing package structure immediately, this change are conservative & progressive - i.e proc macro with its deps, etcs -.  Hope to move forward to reach point all bindings refer swc_core only, then do internal refactor for the dependency tree more heavily.

**BREAKING CHANGE:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -211,15 +211,10 @@ dependencies = [
  "serde",
  "serde_json",
  "swc",
- "swc_common",
- "swc_ecma_lints",
- "swc_ecmascript",
- "swc_plugin_runner",
+ "swc_core",
  "tracing",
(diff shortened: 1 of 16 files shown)


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_plugin --breaking

kdy1 (changes requested):
There was a conflict

kwonoj:
@kdy1 ok,updated and resolved.

kdy1 (approved):

PR #5373, opened 2022-08-03
feat(es/minifier): Remove duplicate `var` declarations

kdy1:
**Description:**

TypeScript namespaces result in lots of duplicated variable declarations like
`var ts`


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5376, opened 2022-08-03
fix(es/base): Don't insert extra paren when left of BinExpr is Update

Austaras:
**Description:**

So following is valid js code. Impressive
```js
console.log(a--&&++b)
```

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---
swc-bump:
 - swc_ecma_transforms_base

PR #5387, opened 2022-08-04
feat(swc/core): expand features to support node bindings

kwonoj:
**Description:**

This PR improves `swc_core`, now most of node binding's deps are encapsulated into. There are 2 remaining (node_bundler, nodejs_common) that I may need to think bit about its organization.

I tried to simplify some features for high-level pkgs like swc, but there are still lot to go. For now, focused to making all the check passes.

**BREAKING CHANGE:**

https://github.com/swc-project/swc/discussions/5244


crates/swc_core/Cargo.toml
```diff
@@ -30,11 +41,29 @@ perf = ["swc_common/perf"]
 # refer binding_core_wasm for example: it has plugin_transform_host feature, but
 # it has to be a feature to be enabled by compile time flag to avoid cargo's feature collision,
 # while it also need to access 'common' exports by default for cargo check / or clippy.
-common = ["swc_common"]
+common = ["__common"]
+# TODO: internally concurrent / perf refers to the same feature (`parking_lot`)
+# - can we consolidate them?
+common_concurrent = ["__common", "swc_common/concurrent"]
```
kwonoj, line 47: q: what's difference between common/perf vs. common/concurrent? it looks like `/perf` enables specific for some hashmap, curious if `concurrent` can includes everything or it have dinstict purpose.

crates/swc_core/build.rs
```diff
@@ -12,6 +12,15 @@ compile_error!(
      run plugin, use 'plugin_transform_host_*' instead."
 );
 
+#[cfg(all(feature = "__plugin_transform", feature = "common_concurrent"))]
```
kwonoj, line 15: nit note: I wanted to have more strict checkers, but it'll make cargo's `check / ...` fail when it tries to merge mutually exclusive features. :/

crates/swc_core/src/lib.rs
```diff
@@ -54,3 +61,28 @@ pub mod plugin_runner {
...
+// swc_ecma_loader
+#[cfg(any(docsrs, feature = "loader"))]
+#[cfg_attr(docsrs, doc(cfg(feature = "loader")))]
+pub mod loader {
+    pub use swc_ecma_loader::*;
+}
+
+#[cfg(any(docsrs, feature = "allocator_node"))]
+#[cfg_attr(docsrs, doc(cfg(feature = "allocator_node")))]
+extern crate swc_node_base;
```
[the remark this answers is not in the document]
kdy1: I think this is fine, as we have a feature flag

crates/swc/Cargo.toml
```diff
@@ -1,34 +1,37 @@
...
 [lib]
 bench = false
-name = "swc"
+name  = "swc"
 
 [features]
-concurrent = ["swc_ecma_transforms/concurrent", "swc_common/concurrent", "swc_ecma_minifier/concurrent", "swc_ecmascript/concurrent"]
-debug = ["swc_ecma_visit/debug"]
+# TODO: this seems unused, confirm & remove later once imports to swc flattens out ecma_*
+#concurrent = ["swc_ecma_transforms/concurrent", "swc_common/concurrent", "swc_ecma_minifier/concurrent", "swc_ecmascript/concurrent"]
```
kdy1, line 20: Can you remove `(swc, concurrent)` from ci?

kwonoj: Oh right, thanks for catching. Updated.

crates/binding_core_wasm/src/lib.rs
```diff
@@ -75,7 +75,7 @@ pub fn minify_sync(s: JsString, opts: JsValue) -> Result<JsValue, JsValue> {
 
     try_with_handler(
         c.cm.clone(),
-        swc::HandlerOpts {
+        swc_core::base::HandlerOpts {
```
kdy1, line 78: I think `swc_error_reporters` deserves a own module, because it can be useful for someone does not need full swc

kwonoj: Yes, maybe I came to have some similar opinion around other dependency reorg. For now, pr focused to move things transparently first and then apply necessary changes on top of that. I'll note this as well.

crates/swc_core/Cargo.toml
kdy1: `concurrent`allows using `Lrc` and some types in multi-threaded context.
(i.e. They implements `Sync` and `Send`

crates/swc_core/build.rs
```diff
@@ -12,6 +12,15 @@ compile_error!(
...
+compile_error!("plugin transform cannot enable concurrent mode.");
+
+#[cfg(all(feature = "transforms", feature = "transforms_concurrent"))]
```
kdy1, line 18: Why are these features are mutually exclusive?

crates/swc_core/Cargo.toml
```diff
@@ -1,15 +1,15 @@
...
+license     = "Apache-2.0"
+name        = "swc_core"
+repository  = "https://github.com/swc-project/swc.git"
+version     = "0.2.0"
 
-[package.metadata.docs.rs]
-all-features = true
-rustdoc-args = ["--cfg", "docsrs"]
+  [package.metadata.docs.rs]
+  all-features = true
```
kdy1, line 11: We may need to modify this.
https://docs.rs/crate/swc_core/0.2.0/builds

crates/swc_core/build.rs
kwonoj: Not technically, more for the guide to the users. User do not needs to select transforms / transforms_concurrent both - enabling transforms_concurrent implies to enable transforms already. If they don't need concurrent, using transforms only should work.

crates/swc_core/Cargo.toml
kwonoj: Not sure what'd be good way to fix - tried to limit number of features for now.

crates/swc/Cargo.toml
kwonoj: Peeking this more, it looks like this is still needed for now until we reorganize swc to detach some of its transitive deps. Reenabled this + with CI test again and then will try revisit this later.

crates/swc_ecmascript/README.md
```diff
@@ -0,0 +1,3 @@
+# Deprecated
```
kwonoj, line 1: Digging dep tree, swc_ecmascript was not being referenced across the internal pkgs, so attempt to make one-last publish to include readme then improve swc_core for the external usecase.

kwonoj:
@kdy1 Ok, I think this PR's grown enough and would like to avoid additional changes on top of this to make review easier. If this could be approved & land soon, it'll be helpful to work on additional work for today. 🙏

crates/swc_core/build.rs
kdy1: But what if some of their dependency depend on one feature?

kwonoj: > some of their dependency

Do you mean if user level code have another direct deps to the transform? The whole purpose of swc_core is prevent that, user level code should not have another deps which can be a transitive deps to swc_core. If that happens, we'll see more troubles like version conflict before worrying about feature conflict.

kdy1: Hmm... Okay I got it

kdy1 (approved):
Thanks!


---

swc-bump:
 - swc_core --breaking

PR #5388, opened 2022-08-04
fix(es/base): Preserve paren expr for optional chaining expr

### LINKED ISSUE #5332: Optional chaining parentheses semantics lost
### Describe the bug

An optional chaining sequence short circuits the entire chain of operations, but does not cross the boundary of parentheses. Parentheses are lost when compiling an optional chain, so it behaves differently (which will almost always result in a `TypeError` but hey, it's what the user asked for).

Something to note is that `tsc`'s behavior is that when the parentheses only contain a cast is the parentheses are dropped. Not sure if that behavior is to be kept but something I discovered. If transpiling to <ES2020, tsc _keeps_ the parens and the semantics are different, which I found interesting and confusing:

`tsc` in: `(a?.b as string[]).c`
`tsc` out (ES2020): `a?.b.c;`
`tsc` out (ES2016): `(a === null || a === void 0 ? void 0 : a.b).c;`

`tsc` in: `(a?.b).c`
`tsc` out: `(a?.b).c`

### Input code

```typescript
(a?.b).c
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2020",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.220&code=H4sIAAAAAAAAA9NItNdL0tRLBgDvvqKQCAAAAA%3D%3D&config=H4sIAAAAAAAAA0WNSwrDMAxE76J1FiHL3KGHMK4SHPxD40CN8d0rF5fsJM2bp0YXLO2NshGwjAk1FvOhnUrNDCsuF1qoQE%2BH8eCui5GTiyKMbd1WjX1K4AksFFx0Rx0ym0IWBp7IxNP%2Fya6ukN73OLTfP3VqJaR4gfojmmWH16SL3Ny%2F3SauU70AAAA%3D

### Expected behavior

```js
"use strict";
(a?.b).c;
```

### Actual behavior

```js
"use strict";
a?.b.c;
```

### Version

1.2.220

### Additional context

_No response_

magic-akari:
**Description:**

**BREAKING CHANGE:**

- fix #5332


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_transforms_base

PR #5399, opened 2022-08-05
perf(es/minifier): Reduce recursion

kdy1:
For 998 files, difference was 8.843936416s vs 9.063045625s

The difference of 0.2 second was consistent


kdy1 (commented):
swc-bump:
 - swc_ecma_utils

PR #5400, opened 2022-08-05
feat(swc/core): expose quote macros

kwonoj:
**Description:**

This PR attempts to encapsultes `swc_ecma_quote` into `swc_core`. Now importing `swc_core::quote` will enable quote macros. 

Basically the approach took in here is bubbling up all the normal macro definition into swc_core from ecma_quote, effectively deprecating swc_ecma_quote. actual inner macro (quote_macros) updated import site to refer swc_core as well.

I'm not sure if this is a desired way - I tried to reexport from existing ecma_quote with preserving import site in ecma_quote_macros, but it doesn't look like able to resolve modules from swc_core when macro expand. I may wrong and happy to course correct.

**BREAKING CHANGE:**


crates/swc_core/src/lib.rs
```diff
@@ -3,9 +3,12 @@
 // Quote
 #[cfg(any(docsrs, feature = "quote"))]
 #[cfg_attr(docsrs, doc(cfg(feature = "quote")))]
-pub mod quote {
-    pub use swc_ecma_quote::*;
-}
+pub mod quote;
+
+/// Not a public interface.
```
kdy1, line 8: Can we have `#[doc(hidden)]`?

crates/swc_ecma_quote/README.md
```diff
@@ -0,0 +1,3 @@
+# `swc_core` replaces `swc_ecma_quote`
+
+Instead of directly importing `swc_ecma_quote`, `swc_core` provides equivalent `swc_core::quote` and `swc_core::quote_expr`.
```
kdy1, line 3: We can remove this in next version I think...
So the README for swc_ecma_quote show proper instruction

kdy1 (changes requested):
Thanks!

crates/swc_ecma_quote/README.md
kwonoj: Yes, that was the reason I added readme but didn't remove whole pkgs yet.

crates/swc_core/src/lib.rs
kwonoj: Thanks, updated.

kdy1 (approved):
Thank you!


---

swc-bump:
 - swc_ecma_quote_macros --breaking

kdy1 (approved):

kdy1 (approved):

PR #5403, opened 2022-08-05
test(es/minifier): Organize simple tests

kdy1:
These tests are reduced version of input where swc output size > terser output size


kdy1 (commented):
swc-bump:
 - ''

PR #5407, opened 2022-08-06
fix(es/codegen): emit extra dot when float is too large

### LINKED ISSUE #5397: Number to Fixed unexpected error
### Describe the bug

Hello everyone. 

I have a problem i am using the latest swc/core but when i use this kind of code it throws an error
`var hasToFixedBugs = (1000000000000000128).toFixed(0) !== '1000000000000000128'` it should keep the parenthesis and transform to string and then fixed to 0 but it removed and tries number to fixed 

```
var hasToFixedBugs = 1000000000000000128.toFixed(0) !== '1000000000000000128';
                     ^^^^^^^^^^^^^^^^^^^^

SyntaxError: Invalid or unexpected token
```

`npm ls`

```
├── @swc-node/register@1.5.1
├── @swc/cli@0.1.57
├── @swc/core@1.2.223
```

### Input code

_No response_

### Config

_No response_

### Playground link

_No response_

### Expected behavior

keep the parenthesis

### Actual behavior

_No response_

### Version

1.2.223

### Additional context

_No response_

Austaras:
**Description:**

The original implementation is not only slow, but also problematic when float is too large so toString would lose precision 

**BREAKING CHANGE:**

closes #5397


kdy1 (approved):
Thank you!

---
swc-bump:
 - swc_ecma_codegen

PR #5408, opened 2022-08-06
fix(swc/core): update references for quote

kwonoj:
**Description:**

Minor fix to quote macro to update reference accordingly to avoid compilation error.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_quote_macros
 - swc_core

PR #5409, opened 2022-08-06
fix(swc/utils): allow quote_ident works with swc_core

kwonoj:
**Description:**

This PR allows `quote_ident!` from utils macro works within swc_core, by resolving inner references expansion to its own transitive deps.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---
swc-bump:
 - swc_ecma_utils

Issue #5406, opened 2022-08-06
Plugin raises `heap_get_oob` on certain machines


kwonoj:
### Describe the bug

Recently on some machines plugin raises these error without clear stack where it happens

```
Error: failed to handle: failed to invoke plugin: failed to invoke `/Users/ojkwon/github/swc-coverage-instrument/target/wasm32-wasi/debug/swc_plugin_coverage.wasm` as js transform plugin at /Users/ojkwon/github/swc-coverage-instrument/target/wasm32-wasi/debug/swc_plugin_coverage.wasm

Caused by:
    0: RuntimeError: out of bounds memory access
    1: heap_get_oob
```

This is not all-time reproducible, but if some machine hits those machine will hit this always.

### Input code

```typescript
-
```

### Config

```json
-
```

### Expected behavior

-

### Version

latest

### Additional context

Per bisecting, https://github.com/swc-project/swc/pull/5062 seems causing the issue. It is unclear which specific change is offender, but could be a starting point to dig further.

kwonoj:
ref: https://github.com/rkyv/rkyv/issues/291

alekangelov:
WSL ubuntu works, Vanilla Ubuntu works, windows 11 works (all with SOME issues, but eventually ran through)

but M1 mac does not

kwonoj:
Latest swc_core / swc@core applied workaround this by disabling bytecheck for now. I'll close this, and track re-enabling via #5472.

liaoyinglong:
with `swc_core = { version = "0.23.29", }` still not work, in M1 mac
```text
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at 'failed to invoke plugin: failed to invoke plugin on 'Some("/Users/lyl/work/self/some-tools/node_modules/.pnpm/next@12.3.0_biqbaboplfbrettd7655fr4n2y/node_modules/next/dist/pages/_document.js")'

Caused by:
    0: failed to invoke `/Users/lyl/work/self/some-tools/packages/swc-plugin/dist/swc_plugin.wasm` as js transform plugin at /Users/lyl/work/self/some-tools/packages/swc-plugin/dist/swc_plugin.wasm
    1: RuntimeError: unreachable
    2: unreachable', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/swc-0.224.0/src/plugin.rs:216:14
```
---
```
[...]

PR #5414, opened 2022-08-07
fix(swc): temporarily disable bytecheck

kwonoj:
**Description:**

Reverts https://github.com/swc-project/swc/pull/5062 partially to amend https://github.com/swc-project/swc/issues/5406 as a workaround.

**This is not a correct fix**, since issue itself is quite tricky it requires time to investigate & create some reliable repro steps - meanwhile this PR unblocks existing users. This means potential delay to AST schema version support.

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_atoms/src/lib.rs
+++ b/crates/swc_atoms/src/lib.rs
@@ -54,7 +54,6 @@ include!(concat!(env!("OUT_DIR"), "/js_word.rs"));
     feature = "rkyv",
     derive(rkyv::Archive, rkyv::Serialize, rkyv::Deserialize)
 )]
-#[cfg_attr(feature = "rkyv", archive_attr(repr(C), derive(bytecheck::CheckBytes)))]
 pub struct Atom(#[cfg_attr(feature = "rkyv", with(crate::EncodeAtom))] Arc<str>);
 
 impl Atom {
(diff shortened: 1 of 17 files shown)


kwonoj:
This requires to release new `@swc/core` also.

crates/swc_atoms/src/lib.rs
```diff
@@ -54,7 +54,6 @@ include!(concat!(env!("OUT_DIR"), "/js_word.rs"));
     feature = "rkyv",
     derive(rkyv::Archive, rkyv::Serialize, rkyv::Deserialize)
 )]
-#[cfg_attr(feature = "rkyv", archive_attr(repr(C), derive(bytecheck::CheckBytes)))]
 pub struct Atom(#[cfg_attr(feature = "rkyv", with(crate::EncodeAtom))] Arc<str>);
```
kdy1, line 57: I think we can remove `crate::EncodeAtom`?

kdy1: Nvm, we can do it later.

kwonoj: If we're going to reintroduce bytecheck we'll still need it, is that correct? If that's the case I'd limit this revert in minimal, single scoped vhange so we can bring it back easily and also bisect easily later if needed.

kdy1 (approved):
Thanks!


---



swc-bump:
 - swc_atoms --breaking

crates/swc_atoms/src/lib.rs
kdy1: No it's workaround for https://github.com/rkyv/rkyv/issues/288

PR #5415, opened 2022-08-07
fix(swc/core): fix quote_macros dep

kwonoj:
**Description:**

Looks like bump script didn't work correctly for this PR? swc_ecma_quote_macros missing new versions for the fix.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_core
swc-bump:

PR #5421, opened 2022-08-08
fix(swc/cli): update plugin template

### LINKED ISSUE #5412: [swc_cli] Generated plugin template fails to compile
### Describe the bug

An SWC plugin template generated with `swc plugin new --target-type wasm32-wasi my-first-plugin` will not compile:

```
error[E0432]: unresolved imports `swc_plugin::ast`, `swc_plugin::plugin_transform`, `swc_plugin::TransformPluginProgramMetadata`
 --> src/lib.rs:1:18
  |
1 | use swc_plugin::{ast::*, plugin_transform, TransformPluginProgramMetadata};
  |                  ^^^     ^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `TransformPluginProgramMetadata` in the root
  |                  |       |
  |                  |       no `plugin_transform` in the root
  |                  could not find `ast` in `swc_plugin`
```

`swc_plugin` v0.89.0 seems to be incomplete. I was able to get things to compile after making the following changes:
1. Set the version of `swc_plugin` to be `"0.88.2"` in the `Cargo.toml`
2. Update the import for `swc_plugin` to be:

```rs
use swc_plugin::{ast::*, metadata::TransformPluginProgramMetadata, plugin_transform};
```

### Input code

_No response_

### Config

_No response_

### Playground link

_No response_

### Expected behavior

I expect to be able to compile the plugin that is generated via the CLI 

### Actual behavior

_No response_

### Version

SWC 0.88.0

### Additional context

_No response_

kwonoj:
**Description:**
- closes #5412, as well as updating swc/core bit more for the testing plugin.

**BREAKING CHANGE:**


crates/swc_cli/build.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use toml::Value;
+
+/// Creates a compile time constant to pick up latest known working version of
+/// swc_core for the plugin template.
+/// This'll ensure each version of published swc_cli will create a plugin
+/// template can be compiled correctly, then user can try to upgrade dependency
+/// as needed.
+fn main() {
+    let core_path = PathBuf::from(&env::var("CARGO_MANIFEST_DIR").unwrap())
```
kdy1, line 16: We don't need to parse `toml` file.
https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

kdy1 (changes requested):
Thanks!

crates/swc_cli/build.rs
kwonoj: It returns version of `self`, isn't it? for example, if I try `env::var("CARGO_PKG_VERSION")` it returns 0.89.0, instead of 0.5.0 for `swc_core`. The thing I'm trying to do is ref _another_ pkg (swc_core) to embed its version into template.

kdy1: Ah I see

kdy1: I think `swc_core` should export it instead
It would allow sharing it from various bindings

kwonoj: huh, yeah, good point. Let me revise changes here.

kwonoj: Ok, updated to let swc_core exposes version by self.

kdy1 (approved):
Thanks!

---
swc-bump:
 - swc_core
 - swc_cli

Issue #5395, opened 2022-08-04
Easier macro / templated fn to create custom SWC runtime binary


kwonoj:
### Describe the feature

While there are a lot of things to be improved though, path to writing a transform plugin is somewhat simplified via support of SWC upstream (`swc_core`'s transform plugin macro and utilities).

There is another pillar to instantiate custom transforms on the other hand by creating a custom binary instead of relying on `@swc/core`. This is somewhat less documented, and requires some boilerplate. to create necessary interfaces to work with. `next-swc` is a good example. For example, it have its own n-api interface definition like https://github.com/vercel/next.js/blob/d6cb795fc6842337e4ba836a399c8d97a8b3aea0/packages/next-swc/crates/napi/src/transform.rs which is somewhat similar to what `@swc/core` have : https://github.com/swc-project/swc/blob/c50b4f08aa30a0898a5a5a9d7f3ce4c275eaa454/crates/binding_core_node/src/transform.rs .  And the only reason next-swc has repeated codes is to apply a few custom transform plugins + few custom configurations to support it.

As we aim `swc_core` to be a de-facto SDK to write any kind of custom codes for the SWC either custom runtime or plugin, it'd be preferred if we could expose some public interface to the factory to those repeated codes. Ideally, if possible having some sort of macros accepts a number of custom transforms & configuration options. The same applies for the minify, bundle and other api surfaces SWC supports.

It's not a must goal for the `swc_core` for now, rather a item to think about. It may not possible to achieve this.

PR #5430, opened 2022-08-10
feat(binding_macros): wasm binding macro

kwonoj:
**Description:**

Partially implements https://github.com/swc-project/swc/issues/5395.

This PR attempts to explore possibility to `swc_core` exposes set of builder macro to create binding functions for the SWC's features. As a first step, PR replaces all existing wasm binding implementation into naive macro-based expansion.

Key takeaway for this approach is, in most cases if someone wants to customize SWC's behavior to create custom binary they'll likely need few customization points only. For example, `transform*`'s `before / custom_pass` is what ppl mostly need if they want custom SWC binary, but it still requires to repeat existing codes compare to SWC's binding those. Macro tries to accept those as customizable param, and then scaffold boilerplate instead.

`binding_core_wasm` in result gets thinner dependency tree - it only needs `swc_core` and `wasm-bindgen` (TBD to remove this even or not). Custom binaries like `next-swc` can be the same instead of keep redundant code inside.

This born from internal need of discussions and I think it has some usecases, but the way it's implemented maybe debatable. Reason I started from simpler wasm binding first.

**BREAKING CHANGE:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,16 +192,22 @@ dependencies = [
 name = "binding_core_wasm"
 version = "1.2.224"
 dependencies = [
+ "swc_core",
+ "wasm-bindgen",
+]
+
+[[package]]
+name = "binding_macros"
+version = "0.1.0"
+dependencies = [
  "anyhow",
  "console_error_panic_hook",
  "js-sys",
  "once_cell",
- "parking_lot_core",
- "path-clean",
- "serde",
- "serde_json",
- "swc_core",
- "tracing",
+ "swc",
+ "swc_common",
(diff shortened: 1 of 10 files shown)


kdy1 (approved):
Thank you!
I'm not sure how much people will build a custom binary, though


---

swc-bump:
 - swc_core --breaking

kwonoj:
> I'm not sure how much people will build a custom binary, though

Primary usecase (for now) is reduce repeated code between `next-swc`, and I guess that'll be suffecient win.

Once next-swc does bump for the SWC next time, we may able to try this.

PR #5431, opened 2022-08-10
refactor(binding_macros): update import references

kwonoj:
**Description:**

Minor update to import references. 

I'll leave `wasm_bindgen` macro requires direct dep due to proc_macro for now, it may require amount of work compare to gain. Since caller need to assign attribute explicitly if needed, it somewhat makes sense to make it as direct dep anyway.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - binding_macros

PR #5438, opened 2022-08-11
feat(swc/plugin): add few plugin hosts traces

kwonoj:
**Description:**

Add few more traces for the plugin's proxy runs on host side. We still can't instrument inside of plugin though.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_plugin_proxy
 - swc_plugin_runner

PR #5442, opened 2022-08-11
fix(swc/core): bump up swc_atoms

kwonoj:
**Description:**

Yet another pkg bump error causes `swc_core` compilation fails.

@kdy1 I feel this recently occurs lot more frequently, we may need some sort of tests to catch this before publish (or even after publish pkgs).

**BREAKING CHANGE:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -759,7 +759,7 @@ dependencies = [
 
 [[package]]
 name = "dbg-swc"
-version = "0.42.0"
+version = "0.42.1"
 dependencies = [
  "anyhow",
  "clap 3.1.0",
@@ -2926,7 +2926,7 @@ dependencies = [
 
 [[package]]
 name = "swc"
-version = "0.213.0"
+version = "0.213.1"
 dependencies = [
  "ahash",
  "ansi_term",
@@ -2982,7 +2982,7 @@ dependencies = [
 
 [[package]]
 name = "swc_atoms"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "bytecheck",
  "once_cell",
@@ -2995,7 +2995,7 @@ dependencies = [
 
 [[package]]
 name = "swc_bundler"
-version = "0.176.0"
+version = "0.176.1"
 dependencies = [
  "ahash",
  "anyhow",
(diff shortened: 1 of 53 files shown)


kdy1 (approved):
This is only problematic for users who update dependendencies, and they can update `swc_atoms` themselves.

We can verify this by not using `--no-verify` in auto publish script, but it will make everything extremely slow.

One possible solution is supporting `-D` of `cargo-mono` in the auto-bump script. It will patch all dependants.

---

swc-bump:
 - ''

kwonoj:
> This is only problematic for users who update dependendencies, and they can update swc_atoms themselves.

@kdy1  I don't think that is correct statement, the main issue is `swc_atoms` is *missing* version bump, so there's no way user can update their `swc_atoms` while other dependency relies on `swc_atoms` refers latest commits of the main branch in upstream repo.

The issue I hit is like this:

- swc_ecma_ast@0.xx.y
   - swc_atoms@ { version = 0.4.0, path = "../swc_atoms" }
  
This builds fine in swc's git repo, since resolution uses `path` to pick up the latest code. But user code will fail to compile, since it no longer have path resolution but picks up `0.4.0`, where it missing necessary changes. There should be a way to catch depedent dependencies are picking up right commit / changes when it tries to build against version based resolutions.

kdy1:
>@kdy1 I don't think that is correct statement, the main issue is swc_atoms is missing version bump, so there's no way user can update their swc_atoms while other dependency relies on swc_atoms refers latest commits of the main branch in upstream repo.

If so, I think this is a bug of the bump script. It occurs quite frequently.


I think one solution is supporting something like

```yml
swc-bump:
 - swc_atoms
   - swc_ecma_ast
     - swc_ecma_utils

```

where deps of `swc_ecma_utils` => `swc_ecma_ast` and `swc_ecma_ast` => `swc_atoms` is updated

kwonoj:
> I think this is a bug of the bump script. It occurs quite frequently. If the bump script worked correctly, user could update swc_atoms,

Yes, I think so - I wanted to point out that occurs somewhat frequently recently and we may need some checker.

At least if we can sure `swc_core` always builds (which is public package we want to sure it won't break), that'd be a starting point.

PR #5451, opened 2022-08-11
fix(swc): bump up pkgs

kwonoj:
**Description:**

Another attempt to fix code differences issue. This time compilation passes, but some esoteric runtime error thrown while using latest git as dep didn't cause problems. It was not easy to find out root cause, instead I tried to bisect changes between published pkg vs git tree then bumps up pkg have any source differences. 

I'm not sure if this'll fix the problem though - we'll need updated @swc/core and all the rust pkgs to retry those tests.

**BREAKING CHANGE:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2926,7 +2926,7 @@ dependencies = [
 
 [[package]]
 name = "swc"
-version = "0.213.1"
+version = "0.213.2"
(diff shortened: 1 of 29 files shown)


kdy1 (approved):
Thanks!
---

swc-bump:
 - ''

PR #5452, opened 2022-08-11
test(swc): trying to verify swc_core with local pkg

kwonoj:
**Description:**

This PR attempts to add another set of e2e test to ensure missing version bump in some transitive deps, which are not able to be caught in current CI as it uses path resolutions all the time.

Idea's straightforward, for the pkgs we'd like to ensure (`swc_core`), make a compile test to build against locally packed (`cargo package`) instead of path resolution. generated pkg will try to behave as close to actually published one without path resolution, so we can catch some of build time errors before publish.

Depends on usefulness plan to grow test more, especially some regression tests between @swc/core to plugin for the newly published pkg. This is bit headache with existing plugins (swc-coverage-instrument) and having test will help future regressions.

**BREAKING CHANGE:**


kdy1 (changes requested):
I think it should not be a js test.
It will cause headaches while publishing

kwonoj:
> It will cause headaches while publishing

Will it? theoritically this'll ensure test catches in PR level.

kdy1:
Yep, you can use `ci` branch to verify it. I was working on ci fix, but feel free to override it

kwonoj:
Before digging further, would like to hear workflow which'll break publishing.

The main goal of this PR is (if it works correctly)

- in each PR / or CI run, verify against soon-to-be-published pkg to verify compilation
- if bulid fails, we know some pkg bump is missing.

If using plain js is not right place, where would be a good place? I don't think where / how to execute it changes the way it works in this PR but I may wrong.

kdy1:
We run all js tests before publishing the node package.
It should be done using shell script, not jest-based testing system

kwonoj:
> publishing the node package.

But test doesn't do anything with node.js package, am I missing something?

kdy1:
No, `test doesn't do anything with node.js package` is the exact reason we don't need to run this test in all platforms

kdy1:
We don't need all platform for such task. Using linux is enough

kdy1:
Also, when creating another PR, please expand test matrix of node js integration test so we can catch issues while normal review process

kwonoj:
We're now using swc_core public version, so we can check this when we build binding_*.

PR #5455, opened 2022-08-12
Synchronize implementation of ECMA minimizers

### LINKED ISSUE #5454: Inconsistencies between ECMA and HTML mininfier
I was learning how to minify JS files from the [ECMA minifier CLI](https://github.com/swc-project/swc/blob/7437f53047853dad28d5b0274137a4167e74976c/crates/swc_ecma_minifier/src/cli/bin.rs). It uses all three transforms, the `resolver`, `hygiene` and `fixer`.

I discovered that the HTML minifier, [when it minimizes JS](https://github.com/swc-project/swc/blob/9d793f1cd3f2e23ba5b286783e0633a13690f1c2/crates/swc_html_minifier/src/lib.rs#L1452-L1571), only uses the `resolver` transform.

Is this intentional? Or a bug?

daxpedda:
**Description:**

This makes sure that all transforms are applied to all instances of ECMA minimization.
I also added `omit_trailing_semi`, I hope this is correct.

Fixes #5454.

### DIFF
--- a/crates/swc_ecma_minifier/examples/compress.rs
+++ b/crates/swc_ecma_minifier/examples/compress.rs
@@ -9,7 +9,7 @@ extern crate swc_node_base;
 use std::{env::args, fs, path::Path};
 
 use swc_common::{sync::Lrc, Mark, SourceMap};
-use swc_ecma_codegen::text_writer::JsWriter;
+use swc_ecma_codegen::text_writer::{omit_trailing_semi, JsWriter};
 use swc_ecma_minifier::{
     optimize,
     option::{ExtraOptions, MinifyOptions},
@@ -81,7 +81,7 @@ fn print<N: swc_ecma_codegen::Node>(cm: Lrc<SourceMap>, nodes: &[N], minify: boo
             },
             cm: cm.clone(),
             comments: None,
-            wr: Box::new(JsWriter::new(cm, "\n", &mut buf, None)),
+            wr: Box::new(omit_trailing_semi(JsWriter::new(cm, "\n", &mut buf, None))),
         };
 
         for n in nodes {
--- a/crates/swc_ecma_minifier/examples/minifier.rs
+++ b/crates/swc_ecma_minifier/examples/minifier.rs
@@ -5,7 +5,7 @@ extern crate swc_node_base;
 use std::{env::args, fs, path::Path};
 
 use swc_common::{sync::Lrc, Mark, SourceMap};
-use swc_ecma_codegen::text_writer::JsWriter;
+use swc_ecma_codegen::text_writer::{omit_trailing_semi, JsWriter};
 use swc_ecma_minifier::{
     optimize,
     option::{ExtraOptions, MangleOptions, MinifyOptions},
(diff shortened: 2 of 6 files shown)


crates/swc_html_minifier/src/lib.rs
```diff
@@ -1533,17 +1533,25 @@ impl Minifier<'_> {
             },
         );
 
+        swc_ecma_visit::VisitMutWith::visit_mut_with(
```
kdy1, line 1536: We should not apply this

crates/swc_html_minifier/src/lib.rs
```diff
@@ -1533,17 +1533,25 @@ impl Minifier<'_> {
...
-                cm.clone(),
-                "\n",
-                &mut buf,
-                None,
-            )) as Box<dyn swc_ecma_codegen::text_writer::WriteJs>;
-
-            wr = Box::new(swc_ecma_codegen::text_writer::omit_trailing_semi(wr));
+            let wr = swc_ecma_codegen::text_writer::JsWriter::new(cm.clone(), "\n", &mut buf, None);
+
+            let wr = Box::new(swc_ecma_codegen::text_writer::omit_trailing_semi(wr));
```
daxpedda, line 1554: Boxing it twice wasn't necessary.
Happy to clean this up by importing `JsWriter` and `omit_trailing_semi`? Was trying not to change too much.

crates/swc_ecma_minifier/examples/minifier.rs
```diff
@@ -58,7 +58,7 @@ fn main() {
         )
         .expect_module();
 
-        let output = output.fold_with(&mut fixer(None));
+        let output = output.fold_with(&mut hygiene()).fold_with(&mut fixer(None));
```
kdy1, line 61: Please revery this

crates/swc_ecma_minifier/examples/minify-all.rs
```diff
@@ -69,7 +69,7 @@ fn main() {
                         )
                         .expect_module();
 
-                        let output = output.fold_with(&mut fixer(None));
+                        let output = output.fold_with(&mut hygiene()).fold_with(&mut fixer(None));
```
kdy1, line 72: Same here

crates/swc_ecma_minifier/examples/compress.rs
```diff
@@ -59,7 +59,7 @@ fn main() {
         )
         .expect_module();
 
-        let output = output.fold_with(&mut fixer(None));
+        let output = output.fold_with(&mut hygiene()).fold_with(&mut fixer(None));
```
kdy1, line 62: Same here

crates/swc_ecma_minifier/examples/minifier.rs
daxpedda: So it should only apply fixer but not `hygiene`? I would like to understand why.

crates/swc_html_minifier/src/lib.rs
daxpedda: So it should only apply fixer but not `hygiene`? I would like to understand why.

daxpedda: Done.

kdy1 (changes requested):
We need `fixer`, and that was the issue

daxpedda:
Should I remove `omit_trailing_semi`?
Should I make it in a separate PR or is it undesired? If yes, I would like to know why, should I not use it in my own code? Same with `hygiene`.

kdy1:
No, hygiene and name mangler is playing same role, so it's not required for the minifier

daxpedda:
Ah! Alright, thanks!

Is the PR alright like that, or should I also remove `omit_trailing_semi`?

kdy1:
No, it's fine

PR #5459, opened 2022-08-12
build(cargo): downgrade rkyv

kwonoj:
**Description:**

Speculative fix.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!
---

swc-bump:
 - swc_common --breaking

kdy1 (changes requested):
Oh ci failed

crates/swc_common/src/plugin/serialized.rs
```diff
@@ -173,13 +173,11 @@ where
         .map_err(|_err| Error::msg("Failed to deserialize given ptr"))
 }
 
+/*
```
kwonoj, line 176: @kdy1 somehow compilation fails as soon as downgrade rkyv related with deserialization logic with versionedserializable struct. This maybe partially related with the issues we're digging - tried to disable it for now until we can reliably figure out plugin workflows.

kdy1: Interesting... But yeah we should disable it for now.

kdy1 (approved):

PR #5467, opened 2022-08-12
fix(css/ast): use list of components values for recovery

alexander-akait:
**Description:**

Better AST

**BREAKING CHANGE:**

Yes


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_ast --breaking

Issue #5410, opened 2022-08-07
Support CSS `@container`


yisibl:
### Describe the feature

Container Query is a very important new CSS feature that browsers are actively implementing. Currently, Chrome 105+ and Safari 16+ are shipping.

- https://caniuse.com/css-container-queries
- Spec https://drafts.csswg.org/css-contain-3/
- Disallow and/not/or from <container-name> https://chromium-review.googlesource.com/c/chromium/src/+/3698402

```css
@supports ( container-type: size ) {
  @container ( width <= 150px ) {
    #inner {
      background-color: skyblue;
    }
  }
}
```

alexander-akait:
Hm, but we already supported this, do you check this?

yisibl:
@alexander-akait It seems to resolve to `PreservedToken` in many places.
https://swc-css.netlify.app/?code=eJxzSM7PK0nMzEstUtBQKM9MKclQsLFVMDQ1KKhQ0FSoruXS11IoK1bQ0udyyE1NyUzErgoA8PIUcw

```css
@container ( width <= 150px ) {}
/* vs */
@media ( width <= 150px ) {}
```

yisibl:
That's great. 

I have a table documenting the differences between the different parsers, please help to see if there are any errors or omissions.

![image](https://user-images.githubusercontent.com/2784308/183908351-899a5596-e149-4e0b-a695-afa01b3e6da6.png)

alexander-akait:
Thank you for the table, `@font-feature-values` already in my todo (today/tomorrow), can you provide link on
- `@scroll-timeline`
- `@position-fallback`

yisibl:
Yes, see https://www.yuque.com/docs/share/593d42c7-14c7-4be4-8af2-fe15b1bdb889?#rNXU 
password: `rn5p`

- `@scroll-timeline`: seems to have been removed from the spec.
 - `@position-fallback`: Chrome is in the process of implementing it. https://bugs.chromium.org/p/chromium/issues/detail?id=1309178#c5

yisibl:
Add a special case:

> both of these are valid, with not being part of the condition rather than the name. If implementations are getting that wrong, we should at least create a WPT, and file bugs - and potentially clarify the spec.
> https://github.com/w3c/csswg-drafts/issues/7203#issuecomment-1183737711

```css
@container not (width <= 500px ) { … }
@container name not (width <= 500px) { … }
```

alexander-akait:
WIP in `@container`

[...]

PR #5468, opened 2022-08-12
feat(css/parser): parse @font-feature-values at-rule

alexander-akait:
**Description:**

Partial https://github.com/swc-project/swc/issues/5410
Spec: https://www.w3.org/TR/css-fonts-4/#font-feature-values

**BREAKING CHANGE:**

Yes


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_ast --breaking

Issue #5469, opened 2022-08-12
`swc_ecma_minifier` requires unstable to build


kwonoj:
### Describe the bug

This forces plugin, or `swc_core` depends binary to use nightly.

### Expected behavior

-

### Actual behavior

```
swc_ecma_transforms_compat/src/es2017/async_to_generator.rs:356:32

356 |                 if !in_iife && let Some(c) = &mut self.comments {
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #53667 <https://github.com/rust-lang/rust/issues/53667>
```

### Version

latest

kdy1:
cc @Austaras I think we should not use it for now.
I think a good MSRV would be 1.60

PR #5473, opened 2022-08-13
refactor(swc/transforms): remove unstable syntax

kwonoj:
**Description:**
- closes https://github.com/swc-project/swc/issues/5469

Personally I'm fine if we can use unstable syntax encapsulated in swc's codebase, but this leaks into user level for the ppl wants to write plugins or custom binary.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_transforms_compat

Issue #5471, opened 2022-08-13
Ineffecient PluginSourceMapProxy::lookup_char_* can leads into plugin execution fails


kwonoj:
### Describe the bug

When source input code is large enough, and then if `lookup_char_*` (i.e lookup_char_pos) is being called enough amount of times, it can drain plugin runtime's memory space to lead into plugin execution fails. These are somewhat subjective, as input / execution count both needs to be some amounts.

Peeking implementation bit, it looks like due to we entirely exchange whole struct to `Loc` per each request, specifically

```
pub struct Loc {
    /// Information about the original source
    pub file: Lrc<SourceFile>,
...
```

While host shares it as Arc, per each request plugin actually gets whole copy of the `SourceFile` which is fairly large. Instead of serialize - deserialize whole `Loc` object, proxy should try to share `SourceFile` per whole transformation and only gets rest of properties for the `Loc` from the host.

### Expected behavior

-

### Version

latest

kwonoj:
Other places: 

`span_to_lines`

PR #5475, opened 2022-08-13
fix(swc/plugin_runner): share SourceFile for the comments

kwonoj:
**Description:**

- closes https://github.com/swc-project/swc/issues/5471

PR improves sourcemap proxy to copy `SourceFile` only once into plugin and reuse it for subsequent calls. This works since plugin is per each transform cycle that we can expect single SourceFile will live.

**BREAKING CHANGE:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,7 +202,7 @@ dependencies = [
  "path-clean",
  "serde",
  "serde_json",
- "swc_core 0.7.10 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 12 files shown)


kwonoj:
**THIS REQUIRES TO UPDATE @swc/core and rust pkgs at once**, so let's defer this until we verify existing issues for the plugin has fixed.

kwonoj:
This is ready for review. @kdy1

kdy1 (approved):
Thanks!

---

swc-bump:
 - ''

PR #5482, opened 2022-08-14
refactor(swc/bindings): use public swc_core as interface

kwonoj:
**Description:**

This PR now changes how we build bindings - instead of relying on path to get the latest codes in the repo, it uses official, public swc_core instead. It provides numbers of benefits for us - first we can ensure latest public package won't break, also it removes concerns of binary differences between host to the other consumers. 

But also this changes the workflow we release bindings - we'll need to publish swc_core first, then release bindings_node from it. It's a small step though.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - ''

PR #5488, opened 2022-08-14
Derive a few common traits for simple public types

alvra:
This project contains three identical versions of two simple public enums: `IndentType` and `LineFeed`.

For a project I would like to be able to write tests with `assert_eq!()`, but that requires `Eq` and `Debug`. I thought it would make sense to also implement `Clone` and `Copy`.

PR #5489, opened 2022-08-14
fix(html_ast,css_ast): some types miss rkyv attributes

wre232114:
**Description:**
* fix that some types in `swc_html_ast` and `swc_css_ast` are not rkyv serializable.

**BREAKING CHANGE:**
None


wre232114:
@kdy1 please take a look, thanks!

crates/swc_common/src/syntax_pos/hygiene.rs
```diff
@@ -57,6 +57,10 @@ struct SyntaxContextData {
 
 /// A mark is a unique id associated with a macro expansion.
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
+#[cfg_attr(
+    feature = "rkyv",
+    derive(rkyv::Archive, rkyv::Serialize, rkyv::Deserialize)
+)]
```
kwonoj, line 63: If I recall correctly I beieve I intentionally did not enable serialization for this since its inner value (u32) can be exchanged between host - guest without any serialization.

kwonoj: ref: https://github.com/swc-project/swc/blob/34a0fac442bacad5ef0fabbccdaf2bfabb647da4/crates/swc_common/src/syntax_pos/hygiene.rs#L79-L81

kdy1:
https://github.com/swc-project/swc/pull/5488#issuecomment-1214365192

Please wait

crates/swc_common/src/syntax_pos/hygiene.rs
wre232114: OK, I'll revert it

wre232114:
Got it, thanks!

kwonoj:
It's somewhat late stage but curiousity, would you mind fill in some background for the reason would like to add rkyv support for these ast? these are not currently exposed elsewhere like plugin, makes thin usecases to have it.

wre232114:
I am building tools based on swc toolchain (parse/transformer/codegen for ecma, css and html), and needs to serialize these ast for caching. ecma ast is already rkyv serializable but others don't.

BTW, I think in the future css and html may support plugins and also need rkyv support.

kwonoj:
> , and needs to serialize these ast for caching. ecma ast is already rkyv serializable, but others don't.

That makes sense. But also I'd like to clarify one thing: being `rkyv`-serializable is not a **public** spec of SWC. SWC _may_ change internal mechanism to serialization / deserialization other than rkyv in the future. The only guarantee SWC's AST may provide is it'll be `serializable / deserializable`.

wre232114:
> future

That's fine, if there are better alternatives, I'll keep up with swc.

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_ast
 - swc_html_ast

kdy1 (approved):

PR #5495, opened 2022-08-15
fix(swc/plugin_proxy): correct compilation attr

kwonoj:
**Description:**

This seems to be caused while resolving merge conflict, ends up compilation fails on the plugin side.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - ''

PR #5498, opened 2022-08-15
feat(es/compat): Fix syntax context of transpiled class methods

kdy1:
**Description:**

Currently, once renamer decides to rename `dipose#0` as `dispose`, it uses same rule for different scopes, even if the other scope has `dispose#1` => `dispose`.

Under assumptions of the span-hygiene system, the input is wrong because there are multiple unrelated variables with the same symbol and the syntax context.

So I fixed `classes` pass instead of renamer.

 - https://github.com/vercel/next.js/issues/39360


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_compat

PR #5501, opened 2022-08-15
Revert 5494 feat swc core metadata git

kwonoj:
**Description:**

It is my mistake to understand way `vergen` works - in result, it makes not possible to publish plugin.

For now simply revert it and attempt to find new way instead.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---
swc-bump:
 - ''

PR #5503, opened 2022-08-15
fix(css/prefixer): handle supports

alexander-akait:
**Description:**

Fix `@supports` + future imporvements

**BREAKING CHANGE:**

No

No


crates/swc_css_prefixer/src/prefixer.rs
```diff
@@ -698,6 +701,67 @@ impl VisitMut for Prefixer {
         }
     }
 
+    fn visit_mut_supports_condition(&mut self, supports_condition: &mut SupportsCondition) {
+        let old_supports_condition = self.supports_condition.take();
+
+        self.supports_condition = Some(supports_condition.clone());
```
kdy1, line 707: This seems too inefficient. Is this unavoidable?

alexander-akait: @kdy1 Unfortunately not, we need to avoid duplication, so we need to know which supports conditions exist, the same for declarations (we are copying simple block)

kdy1: I see, then it's fine. Thank!

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_prefixer

PR #5506, opened 2022-08-16
refactor(bindings): separate bindings binary from workspace

kwonoj:
**Description:**

Now all the bindings binaries are built against public swc_core, and it causes version selection conflict between workspaces's path-resolved deps. PR takes a first step to refine this by separating out those binaries from the workspace.

There will be additional improvement for the overall build pipeline - for now, primary focus is just make it work to unblock releasing new `@swc/core`.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - ''

.github/workflows/cargo.yml
```diff
@@ -96,9 +96,7 @@ jobs:
             os: ubuntu-latest
           - crate: better_scoped_tls
             os: ubuntu-latest
-          - crate: binding_core_node
```
kdy1, line 99: I think this should be restoed to use test code below

.github/workflows/cargo.yml
```diff
@@ -96,9 +96,7 @@ jobs:
...
-            os: ubuntu-latest
-          - crate: binding_core_wasm
```
kdy1, line 101: Same here

.github/workflows/cargo.yml
```diff
@@ -128,13 +124,6 @@ jobs:
               cargo hack check --feature-powerset --no-dev-deps
           - crate: swc_cached
             os: ubuntu-latest
-          - crate: swc_cli
```
kdy1, line 131: Same here

kdy1 (changes requested):

kdy1 (approved):

kdy1 (approved):

kdy1 (approved):

PR #5515, opened 2022-08-16
feat(swc/core): implement engine diagnostics

kwonoj:
**Description:**

This PR is second attempt to https://github.com/swc-project/swc/pull/5501 . 

Relying on build script time git info is not something we want - since `build` can occur when binary try to import `swc_core`. What we need is pre-baked in constants when `swc_core` is published. PR still try to use vergen for some optional values, but for the git sha it uses a more direct approach to generate sha when publish script running.

**BREAKING CHANGE:**

### DIFF
--- a/.github/workflows/bot.yml
+++ b/.github/workflows/bot.yml
@@ -126,6 +126,7 @@ jobs:
       - name: Publish crates
         if: steps.bump.outcome == 'success' && steps.bump.conclusion == 'success'
         run: |
+          npx ts-node src/cargo/update-constants.ts
           cargo mono publish --no-verify
           cargo mono publish --no-verify
           cargo mono publish --no-verify
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -60,9 +60,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.58"
+version = "1.0.62"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bb07d2053ccdbe10e2af2995a2f116c1330396493dc1269f6a91d0ae82e19704"
+checksum = "1485d4d2cc45e7b201ee3767015c96faa5904387c9d87c6efdd0fb511f12d305"
 
 [[package]]
 name = "arbitrary"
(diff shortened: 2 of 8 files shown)


kdy1 (changes requested):
I think this will not work if I publish crates locally

kwonoj:
> I think this will not work if I publish crates locally

yes indeed. how frequent we do manual publish?

kdy1:
I don't have concrete number, but it's quite frequent

kwonoj:
I may be wrong, but I think there's no good way to satisfy all the cases if we want to physically embed some constants. Any suggestions? Or just accept it for some releases by bailing out and emitting default placeholder string instead?

kdy1:
We can generate the dummy file from build script instead.

> Or just accept it for some releases by bailing out and emitting default placeholder string instead

Yeah

kwonoj:
> dummy file from build script instead.

Vergen is essentially this and it reveals thing I didn't thought of : if plugin user want to use swc_core, it is built from cargo cache and not for the git repo so it failed to build it further.

kdy1:
Then we should remove git hash.
The crate version is enough

kwonoj:
This pr will serve git has as 'opportunistic', as it'll likely include but in case of local / manual it may not. 

Thoughts on making it as `Option` instead to reflect it better? Then it'll be an additional metadata but can miss sometimes.

kdy1:
Why is git hash required?
If we know the version, we can know know the git hash by looking at the version.

Note that this PR will break publishing as cargo really hates non-commited files

kdy1:
That's why we have data json files for preset-env commited in the repository.
I tried generating it from CI, but cargo rejected it.

kdy1:
Alternative way is to have it commited in the repository by storing it in the `prepare` script.
If we do so, crate publish CI action will update it everytime, although it would be one-off

kwonoj:
1. We had already know cases version does not include changes we 'thought' we included in. Basically version is non deterministic way to check snapshot of changes.
2. It still can be retrieved by look through git has as you said, but in combination of 1 above it requires amount of manual process each time. 

For the publish, current generated script creates one off dummy commit to remove dirty state but not oush it to origin to not to leak those into git history.

I think the need of hash maybe debatable, but for me it was helpful to bisect regressions in most cases.

kdy1 (approved):
Okay, sounds reasonable so let's try.

---

swc-bump:
 - swc_core

kdy1 (changes requested):
Threre's a conflict

kdy1 (approved):

PR #5516, opened 2022-08-16
build(cargo): remove redundant packages

kwonoj:
**Description:**
We can remove these packages as it is no longer referenced (and update readme's published).

**BREAKING CHANGE:**


kdy1 (changes requested):
We still need `swc_ecmascript`

kwonoj:
> We still need swc_ecmascript

Yes, will revert. Is there way we can announce & instruct migrate into swc_core for the users of swc_ecmascript?

kdy1:
We need to add more to `swc_core` before removing it.
See `lib.rs` of `swc_ecmascript`

kdy1:
About migration, we can do same thing like `swc_ecma_quote`

kwonoj:
Fwiw, swc_core already have ecma_quote support.

kdy1:
I don't mean `swc_core` lacks it. `swc_ecmascript` has much more deps

kdy1:
Closing as stale

PR #5517, opened 2022-08-16
fix(swc/plugin_runner): provide guest memory alloc for the sourcemap

kwonoj:
**Description:**

There was a missing memory export while migrating into wasmer@3. Unfortunately, this means we'll need another `@swc/core` with swc_core release.

**BREAKING CHANGE:**


kdy1 (approved):
swc-bump:
 - ''

PR #5526, opened 2022-08-17
feat(es/minifier): ignore recursive usage for function and pure call

Austaras:
**Description:**

This works suprisingly well for few lines of code

图片

And in the long run we could disable tsc minifier entirely as they would be mostly empty files.

**BREAKING CHANGE:**


crates/swc_ecma_minifier/tests/fixture/issues/vercel/004/output.js
```diff
@@ -1,9 +0,0 @@
-function ItemsList() {
```
kdy1, line 1: This was a test about dependency issue of sequential inliner and I think we should not remove the test

kdy1 (changes requested):
I'm not sure how should I review this PR properly 🤣

crates/swc_ecma_minifier/tests/fixture/issues/vercel/004/output.js
Austaras: ~~It's not removed, it's just compressed~~

Should we add an export here?

kdy1: I think it will work

kdy1 (approved):
Thank you!


---

swc-bump:
 - swc_ecma_minifier

Issue #5535, opened 2022-08-17
Plugin cannot lookup pos for the enums


kwonoj:
### Describe the bug

It _looks_ like when AST's being passed into plugin some of transformation's already done, such as `enum`. When input code contains enum, plugin receives ExprStmts and attempt to lookup original position fails with ` position 0 does not resolve to a source location`.

Not sure if this is intended design or not.

### Input code

```typescript
https://github.com/kwonoj/swc-sourcemap-test

input 

enum MessageSeverity {
  Warning = 1
}
```

plugin
```
    fn visit_mut_expr_stmt(&mut self, expr_stmt: &mut ExprStmt) {
        let span_hi_loc = self.source_map.lookup_char_pos(expr_stmt.span.hi);
    }
```
```


### Config

_No response_

### Playground link

_No response_

### Expected behavior

Not sure what's desired behavior - if plugin supposed to get the ast for the enums, or lookup pos should work consistently.

### Actual behavior

_No response_

### Version

latest

### Additional context

_No response_

kwonoj:
@kdy1 Thoughts on how to resolve this? it blocks coverage instrumentation for some typescript files.

kdy1:
Maybe we can pass the whole AST with typescript types to the plugin but then plugins using `noop_visit_mut` for optimization will be broken.

kwonoj:
Maybe plugin would silently ignore these, in most cases lookup source would be needed for the real codes only. In that case the issue is lookup_char traps if it fails to lookup - which plugin does not have way to gracefully handle this at all.

sam-goodwin:
> this plugin or custom transforms may accept BytePos(0)

In what cases will this happen? Can plugins rely on the spans correctly mapping back to the source?

kwonoj:
> In what cases will this happen?

If plugin receives already transformed AST from prior transform / or core transforms which have dummy spans.

sam-goodwin:
> If plugin receives already transformed AST from prior transform

Not precisely sure what you mean by this. Below is my configuration - I only have my plugin configured, but I'm transforming typescript to JS using the built-in SWC transform. Will the spans my plugin map correctly back to the source that created them? It's not clear to me if the ts->js transform runs before or after my plugin - I assume before.

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "dynamicImport": false,
      "decorators": false,
      "hidden": {
        "jest": true
      }
    },
    "transform": null,
    "target": "es2021",
[...]

PR #5536, opened 2022-08-17
[PoC] V2 API changes using cargo feature flag

### LINKED ISSUE #5527: Remove minification apis from `transform()`
I think this is a mistake.

We should provide `minify()` and `transform()` independantly

kdy1:
**Description:**

We can use multi-branch but I think it will really hurt rust-side users.
With multi-branch approach, we have to prevent some patches from being published to crates.io, but I think it's not acceptable for some rust-side users.

So I'm suggesting another approach. We can use a (potentially private) feature flag named `swc_v2` in js-surfacing crates like `swc` or `swc_node_bundler` to 'hide' some changes to node api.

This is clearly not additive, and that's why I think it should be a private feature.
As users are not expected to depend on `swc_v2`, enabling it from outside of SWC **may** break the whole dependency tree, but it's fine.

For PoC, I fixed #5527 only for v2.

I want to hear more about opinions about this approach

## Some constraints

 - rust users should not depends on `swc_v2`. This is not considered as a public api.

 - All code should be accessible witthout `swc_v2`. 
 
But altering behavior is fine.
Let's say, we want to `fix` #5535 in v2 and change the order of plugins to after `resolver` but before `decorator` or `typescript::strip`. In this case we can use `#[cfg(feature = "swc_v2")]` and `#[cfg(not(feature = "swc_v2")]` respectively.

### DIFF
--- a/crates/swc/Cargo.toml
+++ b/crates/swc/Cargo.toml
@@ -38,6 +38,9 @@ plugin_transform_schema_vtest = [
   "swc_plugin_runner/plugin_transform_schema_vtest",
 ]
 
+swc_v1 = []
+swc_v2 = []
+
 [dependencies]
 ahash = "0.7.4"
 anyhow = "1"
--- a/crates/swc/Cargo.toml
+++ b/crates/swc/Cargo.toml
@@ -1,19 +1,19 @@
 
 
 [package]
-authors       = ["강동윤 <kdy1997.dev@gmail.com>"]
-description   = "Speedy web compiler"
+authors = ["강동윤 <kdy1997.dev@gmail.com>"]
+description = "Speedy web compiler"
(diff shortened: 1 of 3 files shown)


kwonoj:
At this point, I think we'll probably need to try couple of different approaches to find what's best to suit our needs. Branch, features both have pros and cons and probably one approach will not fulfill most of our needs. The way I'm thinking is probably we'll use feature for the additive (i.e new things on top of sharable), and branch for the change / remove. Some of questions regarding feature-based major is some cases like if involved change requires to keep lot of repeated code with small amount of changes, or how to control deeply transitive dep's (i.e core -> A -> B -> C -> D and we need change D, whereas A / B/ C potentially have D as well), or something else.

Double clarify this is not an objection, just saying we probably need to collect real world challenges before make a firm decision on one end.

PR #5537, opened 2022-08-17
fix(css/prefixer): less duplication

alexander-akait:
**Description:**

Less duplication + more test + fix bug with `@import`

**BREAKING CHANGE:**

No

No

### DIFF
--- a/crates/swc_css_prefixer/src/prefixer.rs
+++ b/crates/swc_css_prefixer/src/prefixer.rs
@@ -568,13 +568,12 @@ impl Prefixer {
 impl VisitMut for Prefixer {
     fn visit_mut_stylesheet(&mut self, stylesheet: &mut Stylesheet) {
         let mut new_rules = Vec::with_capacity(stylesheet.rules.len());
-        let original_rules = stylesheet.rules.clone();
 
         for mut rule in take(&mut stylesheet.rules) {
             rule.visit_mut_children_with(self);
 
             for mut added_rule in take(&mut self.added_top_rules) {
-                let need_skip = original_rules
+                let need_skip = new_rules
                     .iter()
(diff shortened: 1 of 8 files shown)


crates/swc_css_prefixer/tests/fixture/duplicate-at-rule-3/output.css
```diff
@@ -1,3 +1,4 @@
+@-webkit-keyframes anim {}
```
alexander-akait, line 1: Expected, because the second `@-webkit-keyframes` after `@keyframes` (`autoprefixer` does the same), anyway we will remove it in future when we implemented removing outdated prefixes (after resolve remain TODOs)

crates/swc_css_prefixer/src/prefixer.rs
```diff
@@ -701,6 +695,41 @@ impl VisitMut for Prefixer {
         }
     }
 
+    fn visit_mut_import_prelude(&mut self, import_prelude: &mut ImportPrelude) {
+        import_prelude.visit_mut_children_with(self);
+
+        if !self.added_declarations.is_empty() {
+            if let Some(ImportPreludeSupportsType::Declaration(declaration)) =
+                import_prelude.supports.take()
```
kdy1, line 703: This will set `import_prelude.supports` regardless of the value being `Some(ImportPreludeSupportsType::Declaration(...))`.

kdy1: Is this intended?

alexander-akait: Yes, we do it only when we have `added_declarations` not empty, it means we have more declarations in `supports()` and should convert it to `SupportConditions`

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_css_prefixer

PR #5541, opened 2022-08-17
feat(dbg-swc): Improve minifier comparator

kdy1:
**Description:**

```
Total
  swc: 32898142 bytes
  terser: 32956139 bytes
  Size ratio: 0.998240176132283
swc produced smaller or equal output for 897 files out of 999 files, 89.79%
```


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #5543, opened 2022-08-18
feat(css/parser): make `keyword` optional

alexander-akait:
**Description:**

`keyword` is like `raw` required only for linter, but we don't need to spend time on it in transformations

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_ast --breaking

kdy1 (changes requested):
There's a conflict

alexander-akait:
@kdy1 Fixed :+1:

kdy1 (approved):
Thanks!

PR #5544, opened 2022-08-18
fix(css/codegen): media print

alexander-akait:
**Description:**

Fix print + tests

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_codegen

PR #5545, opened 2022-08-18
feat(es/minifier): Optimize more string property names

kdy1:
**Description:**

I found it while comparing outputs with terser.
image


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5546, opened 2022-08-18
fix(ci): adjust script path

kwonoj:
**Description:**

Other script have working dir to scripts/bot (https://github.com/swc-project/swc/compare/main...kwonoj:adjust-script-path?expand=1#diff-5e734feb9584888f1b2e8bfdd6906c62f34ec000b67e13f661992fdd9fe5e4ccR86), but publish doesn't. adjust path to correct it.

**BREAKING CHANGE:**

### DIFF
--- a/.github/workflows/bot.yml
+++ b/.github/workflows/bot.yml
@@ -126,7 +126,7 @@ jobs:
       - name: Publish crates
         if: steps.bump.outcome == 'success' && steps.bump.conclusion == 'success'
         run: |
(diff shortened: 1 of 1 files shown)


kdy1 (approved):

kdy1 (changes requested):
Oh I think it will fail because `npx ts-node` is not invoked from `scripts/bot`

kdy1 (approved):
swc-bump:
 - swc_core

PR #5548, opened 2022-08-18
feat(es/minifier): Optimize numeric property names

kdy1:
**Description:**

image

swc vs terser


kdy1 (commented):
swc-bump:
 - dbg-swc
 - swc_ecma_minifier

PR #5555, opened 2022-08-18
fix(es/fixer): Wrap yield expr in extends

magic-akari:
**Description:**

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---
swc-bump:
 - swc_ecma_transforms_base

Issue #5559, opened 2022-08-19
Minifier tasklist


kdy1:
- [ ] Boolean(true) => true

---

 - [ ] `RegExp` with a wrong flag should not be converted.

https://github.com/swc-project/swc/blob/86b5641d5ef46d1cb63019d29c17915b7caf9fa0/crates/swc_ecma_minifier/tests/terser/compress/issue_269/regexp/input.js#L5

---

 - [ ] Drop more unused assignments

https://github.com/swc-project/swc/blob/a0b0fd3e0d8a1084a0399e8b0345d66cfe835b0f/crates/swc_ecma_minifier/tests/terser/compress/drop_unused/chained_3/input.js#L1-L8

We can drop `a` because `c = b` follows it.

---

 - [ ] IIFE: Object literal

---

 - [ ] sequential inliner: typeof

---

 - [x] Swap more

<img width="1267" alt="image" src="https://user-images.githubusercontent.com/29931815/187595883-c18fc8cd-3988-4fa1-a79e-332935a3faa1.png">

---

 - [ ] `if_return` is not good as one of terser

<img width="1267" alt="image" src="https://user-images.githubusercontent.com/29931815/187593813-2f3dec6f-8355-49af-b49d-e63e3d2d3fd8.png">

<img width="1267" alt="image" src="https://user-images.githubusercontent.com/29931815/187594101-57306151-4db0-45e1-aa31-6956403ba33c.png">

---

 - [ ] Optimize parens

<img width="1210" alt="image" src="https://user-images.githubusercontent.com/29931815/185269061-cbbf9121-3280-44a3-92bc-3e8e1b7fdfdd.png">

iheyunfei:
Are these tasks up to date?

PS: 😯I wonder about the real-world use case of `Boolean(true) => true`.

kdy1:
Yes, those are up-to-date, and I don't remember the input file. 
I added it to the list because I faced it in a real-world input, but I expected myself to implement it in a near time 🤣

kdy1:
Closing as obsolete. I'll use TODO.txt of terser

PR #5563, opened 2022-08-19
feat(es/minifier): Improve invokation of IIFEs

kdy1:
**Description:**

 - https://github.com/swc-project/swc/issues/5559


Austaras:
This seems have too much edge case to consider, e.g.
```js
var x = 1
function foo(a) {
  console.log(!!a)
  a = 10
}

foo(x)
```

kdy1:
@Austaras Do you have any idea about the way to implement this optimization?

Austaras:
> @Austaras Do you have any idea about the way to implement this optimization?

In your example, terser simply refuses to inline that function call

kdy1:
Oh lol you are right

Austaras:
#5365 is not enough because at the time of inline, function args could be some very complex expression, but later they could be turned into literal. So in the end we still need to be able to compress
```js
var a
a = 10
console.log(a)
```

kdy1:
Closing as this seems this seems like a too risky optimization

kdy1:
I updated the tasklist issue

PR #5564, opened 2022-08-19
fix(es/parser): False positive of `YieldParamInGen`

### LINKED ISSUE #5556: 'yield' cannot be used as a parameter within generator
### Describe the bug

The code transpiled by swc report
```shell
'yield' cannot be used as a parameter within generator
```
But it is correct code and can be run by node and transpiled by babel

### Input code

```typescript
var app;
var ctx;
x = function (y = app => {
    return {
        schedule: {
            interval: '1m',
            type: 'all',
            disable: app.config.env === 'local'
        },
        *task(ctx) {
            const res = yield ctx.curl('http://www.api.com/cache', { contentType: 'json' });
            ctx.app.cache = res.data;
        }
    };
}) {
};
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6",
    "strict": false,
    "strictMode": false,
    "lazy": false,
    "noInterop": false
  },
  "minify": false
}
```

### Playground link

https://play.swc.rs/?version=1.2.239&code=H4sIAAAAAAAAA11PQY7CMAy89xW%2BpaxQKq6Nsq%2FYD5g0QHZDGiVuS4XydxxYLQtzsj32zHjGBBijamYuDF1UcwENhykYcmOAduWOedCfcG2AkSxNKfw2Fdmc7DB52%2F%2BbVbhANs3oexC7s9i%2BcLRG3hfo%2FRsxuIz7qsWe0ozh4I7Shhm01iD8aNCLv%2F3yPP0gzD8tx9%2B8hWCJTJw58xurs36oP0ozJd%2BKE1Hsu25ZFonRsdu5M8jPiC1c6yHZQF%2BPoN95DALKRr2Ks9Q9Z71iA7aRAxI%2Bt8q9KqopNVhRN%2Fzap%2BhuAQAA&config=H4sIAAAAAAAAA0WMSwrDMAxE76K1t93kDj2EcJXg4B8aBWqM714bUrKSmPdmOp3wtHWqrBBdH1o2%2FtJG4hPDa6hGbmoz2jlChiNjPcSWgtdksRTITR2lkMPe1pIvqaoAD%2BJ8xL855lAqn2sFnazVeVcllXyCxjN0lwPet216yfgBCKaI9boAAAA%3D

### Expected behavior

The code can be transpiled by swc normally.

### Actual behavior

_No response_

### Version

1.2.239

### Additional context

_No response_

magic-akari:
**Description:**

**BREAKING CHANGE:**

- fix #5556


crates/swc_ecma_parser/src/parser/expr/ops.rs
```diff
@@ -319,7 +319,7 @@ impl<I: Tokens> Parser<I> {
 
         if is!(self, "await") {
             let ctx = self.ctx();
-            if ctx.in_function && !ctx.in_async && ctx.strict {
+            if ctx.in_function && !ctx.in_async {
```
magic-akari, line 322: `await` should be report as error in "module environment" not "strict mode".
And it is reserved word in async function as well.

crates/swc_ecma_parser/src/parser/ident.rs
```diff
@@ -139,8 +138,8 @@ impl<I: Tokens> Parser<I> {
 
                 // It is a Syntax Error if the goal symbol of the syntactic grammar is Module
                 // and the StringValue of IdentifierName is "await".
-                Word::Keyword(Keyword::Await) if p.ctx().module => {
-                    syntax_error!(p, p.input.prev_span(), SyntaxError::ExpectedIdent)
+                Word::Keyword(Keyword::Await) if p.ctx().module | p.ctx().in_async => {
```
magic-akari, line 141: https://github.com/swc-project/swc/blob/03a2fd06ce548214cb75f51d0ea47795a24c6cce/crates/swc_ecma_parser/src/parser/ident.rs#L135-L137

@kdy1 Do you have any idea about the error is ignored in typescript mode?

kdy1: I think there was a test case in `tsc` test list. I don't remember the name, though.
So you may need to remove this check and run tsc test of the parser

magic-akari: I removed these lines and got more errors.

https://github.com/swc-project/swc/blob/e06d717c36572f8e45a07a2ec5cd32171fecdf32/crates/swc/tests/tsc.rs#L187-L192

Since we always test tsc files in module mode, we may get more errors than expected.

kdy1: I see, I think it's fine for now

kdy1 (approved):
Thank you!


---
swc-bump:
 - swc_ecma_parser --breaking

PR #5566, opened 2022-08-19
mod(es/compat): use seperate SyntaxContext for class name

Austaras:
**Description:**

Although this problem would disappear if user separate their transform and minifiy pass, but it's wired that for if some certain AST is generated and parsed back into another AST, their SyntyaxContext would change. Also the performance cost is negligible.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---
swc-bump:
 - swc_ecma_transforms_compat

PR #5567, opened 2022-08-19
fix(css/parser): parsing media query

### LINKED ISSUE #5615: Exclude `layer` keyword in media query parser
### Describe the bug

According to recent spec, the <media-type> production does not include the keywords `only, not, and, or, layer`.

- Chrome fix it in: https://chromium-review.googlesource.com/c/chromium/src/+/3839085
- WPT test: http://wpt.live/css/mediaqueries/mq-invalid-media-type-layer-001.html

The following syntax is illegal:

```css
@media not layer {
    div { background-color: red; }
}

@media layer {
    div { background-color: pink; }
}
```

alexander-akait:
**Description:**

From spec:

> <media-type> = [<ident>](https://drafts.csswg.org/css-values-4/#typedef-ident)

> The [<media-type>](https://drafts.csswg.org/mediaqueries-5/#typedef-media-type) production does not include the keywords [only](https://drafts.csswg.org/mediaqueries-5/#valdef-media-only), [not](https://drafts.csswg.org/mediaqueries-5/#valdef-media-not), and, or, and layer.

**BREAKING CHANGE:**

Yes

fixes #5615


kdy1 (changes requested):
Why is this required? `Ident` is much more convenient to use than `MediaType::Ident`

alexander-akait:
@kdy1 By spec yes, because we will can reuse it in future in other grammar, also ehre fix - https://github.com/swc-project/swc/issues/5615

alexander-akait:
@kdy1 Rebased and fixed, CI fail is not related to my changes...

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_css_ast --breaking

PR #5578, opened 2022-08-22
test(es/transforms): Copy execution tests from babel

kdy1:
We can execute lots of babel execution tests, so we do so.


kdy1 (commented):
swc-bump:
 - swc

Issue #5540, opened 2022-08-17
Integration tests should use locally patched version of crates


kdy1:
Currently, integration tests use published version of `swc_core`, which is pretty useless

PR #5591, opened 2022-08-22
feat(swc/plugin): provide swc_core diag from plugin

kwonoj:
**Description:**

This PR implements a new interface from plugin to expose it's core engine diag for `swc_core`. 
Host (plugin_runner) will call it immediately once plugin instance is ready. These values can be used to determine plugin's status / incompatiblity in runtime, in case if we need to branch execution depends on the semver major.

As noted in comment, this is considered as `handshake` and we have to extremely careful to not to break those interface - which means we need to reintroduce forward compat schema version again soon.

**BREAKING CHANGE:**

This is a breaking changes for the plugin. We need to release new `@swc/core` and `swc_core` both. I was thinking if we should make this as non-breaking, while techinically possible I decided to make one another breaking as this'll be core information we'll need anyway.

### DIFF
--- a/node-swc/e2e/plugins/plugins.schema.test.js
+++ b/node-swc/e2e/plugins/plugins.schema.test.js
@@ -63,7 +63,7 @@ const buildPlugin = async (feature) => {
 };
 
 describe("Plugins", () => {
-    describe("Transform AST schema versions", () => {
+    describe.skip("Transform AST schema versions", () => {
         const versionMatrix = [
             // {
             //     host: "plugin_transform_schema_v1",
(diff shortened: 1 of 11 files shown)


node-swc/e2e/plugins/plugins.schema.test.js
```diff
@@ -63,7 +63,7 @@ const buildPlugin = async (feature) => {
 };
 
 describe("Plugins", () => {
-    describe("Transform AST schema versions", () => {
+    describe.skip("Transform AST schema versions", () => {
```
kwonoj, line 66: this is due to https://github.com/swc-project/swc/issues/5540

kdy1 (approved):
Thank you!

---
swc-bump:
 - swc_plugin --breaking

PR #5595, opened 2022-08-23
fix(swc/core): update dependency version

kwonoj:
**Description:**

Another version mismatch fix between swc_common / swc_plugin_runner.

**BREAKING CHANGE:**


kdy1 (approved):
swc-bump:
 - swc

PR #5597, opened 2022-08-23
fix(es/helpers): Fix `_classStaticPrivateFieldSpecSet`

kdy1:
**Description:**

An import was wrong


kdy1 (commented):
swc-bump:
 - swc

PR #5605, opened 2022-08-23
fix(swc/core): properly exports swc_cached

kwonoj:
**Description:**

Fixing one another mistake 😭 

**BREAKING CHANGE:**


kdy1 (approved):
swc-bump:
 - swc

PR #5607, opened 2022-08-23
mod(es/minifier): relax function inline requirement

### LINKED ISSUE #5685: minifier should inline a simple function with a throw statement
### Describe the feature

image

swc vs terser

_No response_

### Additional context

_No response_

Austaras:
**Description:**

There's no need to check `use_above_decl` for function

**BREAKING CHANGE:**

closes #5685


Austaras:
Seems like currently `ref_count` is always equal to `usage_count`

kdy1:
Hmm... Then I think it's a bug of the analyzer.
It should be different in some cases...

kdy1:
Can you update the tests?

Austaras:
> Can you update the tests?

Not now. This pr needs a major overhaul.

kdy1:
I see, thanks

crates/swc_ecma_minifier/tests/terser/compress/functions/issue_2630_3/output.js
```diff
@@ -1,12 +1,6 @@
...
-    a = 1;
-(function () {
-    (function f1(a) {
-        f2();
-        --x >= 0 && f1({});
-    })(a++);
-    function f2() {
-        a++;
-    }
-})();
```
Austaras, line 11: As I mentioned before, terser cannot inline into conflict variable names, great work for swc!

crates/swc_ecma_minifier/src/compress/optimize/inline.rs
```diff
@@ -517,6 +518,18 @@ where
...
+                                    AliasConfig {
+                                        marks: Some(self.marks),
+                                        ignore_nested: false,
+                                    },
+                                ) {
+                                    if let Some(usage) = self.data.vars.get_mut(&i) {
+                                        usage.ref_count += 1;
+                                    }
+                                }
+
```
Austaras, line 532: As we only inline very simple functions, this shouldn't be a big performance penalty

crates/swc/tests/tsc-references/readonlyRestParameters.2.minified.js
```diff
@@ -1,19 +1,2 @@
...
-    ].concat(_to_consumable_array(args))), f1.apply(void 0, _to_consumable_array(args));
-}
-function f2() {
-    for(var _len = arguments.length, args = Array(_len), _key = 0; _key < _len; _key++)args[_key] = arguments[_key];
-    f0.apply(void 0, _to_consumable_array(args)), f1("abc", "def"), f1.apply(void 0, [
-        "abc"
-    ].concat(_to_consumable_array(args))), f1.apply(void 0, _to_consumable_array(args)), f2("abc", "def"), f2.apply(void 0, [
-        "abc"
-    ].concat(_to_consumable_array(args))), f2.apply(void 0, _to_consumable_array(args));
-}
```
Austaras, line 19: We shouln't rely on inling to remove multi recursive functions, this needs to be fixed in future.

Austaras: #5709

kdy1: Oh.. I'll check if a graph is really required once I merge this PR and https://github.com/swc-project/swc/pull/5699

Austaras: It's needed. Just not in this case.

kdy1 (approved):
Thank you!
I got the logic. Lefting a comment for (possible) future reference

Previously we check `used_above_decl` to inline from up to down. This solves the issue mentioned in a comment (https://github.com/swc-project/swc/pull/5607/files#r959710288), but it's not ideal. This PR adds 1 to `usage_count` for other identifiers to solve this issue more smartly

---

swc-bump:
 - swc_ecma_minifier

PR #5617, opened 2022-08-24
feat(swc/cli): enable lto by default for the plugins

kwonoj:
**Description:**

Let plugin scaffold apply optimizations as much as it can.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_cli

PR #5625, opened 2022-08-25
fix(es/fixer): Handle `async` in `for of`

magic-akari:
The following code is valid.
```JavaScript 
var async;
for ((async) of [1, 2]);
```

But it is invalid without parentheses.
```JavaScript 
var async;
for (async of [1, 2]);
```


magic-akari:
There are some issues in parser.

https://github.com/swc-project/swc/blob/14e7b5d76f0f64387f3a40f19c80bff3c20d11b2/crates/swc/tests/tsc-references/parserForOfStatement23_es2015.1.normal.js#L1-L11

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_transforms_base

PR #5637, opened 2022-08-26
feat(es/compress): Drop cons/alt when comparing with object/array literals

magic-akari:
It's safe to drop cons branch since `{} === []` is always false.
```JavaScript 
if ( {} === [] ) {
    console.log("cons")
}
```


kdy1 (changes requested):
Can you move this to the pure optimizer?
https://github.com/swc-project/swc/tree/fdd29c9fe57acdd7778e77b88907edc639bc7018/crates/swc_ecma_minifier/src/compress/pure

Methods in `swc_ecma_utils` are called frequently so this is too much logic to be in core utils

magic-akari:
Hmmm.
I think in reality no one probably writes code like that.
Is it worth implementing it?

kdy1:
Honestly I don't think so

PR #5640, opened 2022-08-26
fix(es/parser): Handle more `TS2406` error

magic-akari:
The following code is also invalid in TypeScript.

```TypeScript
1 >>= 2;
```


kdy1 (approved):
Thanks!

---
swc-bump:
 - swc_ecma_parser

PR #5650, opened 2022-08-28
fix(es/minifier): Fix optimization of template literals with escape

kdy1:
**Description:**

A template string literal is wrongly optimized if it has an escape.

 - https://github.com/vercel/next.js/issues/17056


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5661, opened 2022-08-29
feat(es/parser): add preserve jsx flag

MrToy:
**Description:**

Add preserveJsx flag to prevent transform jsx, such as [typescript jsx preserve](https://www.typescriptlang.org/docs/handbook/jsx.html#basic-usage), sometimes useful for subsequent process

with config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": "preserve"
    }
}
```

```jsx
// input.jsx
<a />
```

will be transform to 

```jsx
// output.jsx
<a/>
```

**BREAKING CHANGE:**

NO

https://github.com/swc-project/swc/discussions/2747


magic-akari:
I think it is better to extend the value of jsx than to add a new parameter.
Because `"jsx": false, "preserveJsx": true` doesn't seem to make sense.

The possible value
- true
- false
- "react" (same as true)
- "preserve"
- "react-native"
- "react-jsx"
- "react-jsxdev"

RiESAEX:
How about add a field to jsc.transform.react as it is controlling the behavior of transform_react.

kwonoj:
I recall there was some discussion around this but forget where it was. Baaically q was does swc need this - i.e in case of tsc it was for th case to chain with different transpiler. swc aims to be terminal transpiler, shouldn't need to chain separate transpiler after swc runs. 

what's the actual usecase in here, and is there way swc provides capabilities correctly instead?

MrToy:
> How about add a field to jsc.transform.react as it is controlling the behavior of transform_react.

The meaning of preserve flag is not to  perform the jsx transform process, should not be reflected in the package swc_ecma_transforms_react that handles the react process

kwonoj (changes requested):
@kdy1 my take on this change is this:

https://github.com/swc-project/swc/pull/5661#issuecomment-1230501407

I'd rather hold this off, and collect practical usecase first. It is a relatively straightforward change we can adapt anytime we want if we find the right usecases, while once add it expanded config / interfaces will be tricky to change / or remove later.

MrToy:
> @kdy1 my take on this change is this:
> 
> [#5661 (comment)](https://github.com/swc-project/swc/pull/5661#issuecomment-1230501407)
> 
> I'd rather hold this off, and collect practical usecase first. It is a relatively straightforward change we can adapt anytime we want if we find the right usecases, while once add it expanded config / interfaces will be tricky to change / or remove later.


I agree with you that should be careful with configuration changes, so we can think about it some more time.

In my use case, swc is used as a loader for my react-native bundler, I need the original jsx information to continue.

I hope swc is not limited to just being a terminal build tool, as I like to use it to do some bare builds, at least this feature is very helpful for me.

I'm looking forward to some use cases coming up, Thanks!

kwonoj:
> I like to use it to do some bare builds

Generally, our recommendation is use rust side packages if someone want to use swc as a building blocks for their own custom tools. It doesn't mean we won't expand js apis at all, but as already shared I'd like to approach very carefully.

Issue #4175, opened 2022-03-28
[css-minifier]Support `@keyframes: <string>`


yisibl:
### Describe the feature

[Chrome does not currently support](https://bugs.chromium.org/p/chromium/issues/detail?id=423424#c11) animation names with quotation marks, which is not compliant with the CSS [specification](https://www.w3.org/TR/css-animations-1/#typedef-keyframes-name). 

```
<keyframes-name> = <custom-ident> | <string>

For example, the following two @keyframes rules have the same name, so the first will be ignored:
@keyframes foo { ... }
@keyframes "foo" { ... }
```

If it is `@keyframes "foo" { ... }`, we can safely remove the quotation marks. This is not only Chrome compatible, but also reduces characters.

Also, it is recommended that users use `<custom-ident>` instead of `<string>` in linter

## Background

https://github.com/w3c/csswg-drafts/issues/118#issuecomment-1079828983

cc @alexander-akait

yisibl:
@alexander-akait  Yes, CSS-wide keywords must be kept in quotation marks.

yisibl:
> Also, it is recommended that users use <custom-ident> instead of <string> in linter

Is this linter already available in swc?

alexander-akait:
@yisibl No, we don't have a linter's rule for it right now, now we want to focus on parser/tranformations and mifications, anyway you can send a PR if you want it

PR #5663, opened 2022-08-29
fix(css/minifier): don't compress `None` for keyframes

alexander-akait:
**Description:**

We already compress keyframes and change `String` to `CustomIdent`

Here we just avoid compress `none`, because it is special word https://www.w3.org/TR/css-animations-1/#keyframes

> A [@keyframes](https://www.w3.org/TR/css-animations-1/#at-ruledef-keyframes) block has a name given by the [<custom-ident>](https://www.w3.org/TR/css-values-4/#identifier-value) or [<string>](https://www.w3.org/TR/css3-values/#string-value) in its prelude. The two syntaxes are equivalent in functionality; the name is the value of the ident or string. As normal for <custom-ident>s and <string>s, the names are fully [case-sensitive](https://html.spec.whatwg.org/multipage/infrastructure.html#case-sensitive); two names are equal only if they are codepoint-by-codepoint equal. The <custom-ident> additionally excludes the [none](https://www.w3.org/TR/css-animations-1/#valdef-animation-name-none) keyword.

**BREAKING CHANGE:**

No

Fixes https://github.com/swc-project/swc/issues/4175


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_minifier

PR #5664, opened 2022-08-30
feat(es/minifier): Remove redundant check

kdy1:
**Description:**

This check is redudant because we only need to check this while 'skipping' and we check it from `is_skippable_for_seq`.
This change is required to optimize
```js
var a = sideEffect();
a(), z();
```


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

Issue #4851, opened 2022-05-31
Regression in 1.2.189 -> 1.2.194 - type-only namespace is not defined.


bradzacher:
### Describe the bug

In the @typescript-eslint project we consume SWC via `@swc/jest`. We have the code/config as below.

In our current version (1.2.189) - our tests run amazingly!
However [dependabot attempted to bump the version to 1.2.194](https://github.com/typescript-eslint/typescript-eslint/runs/6661076466?check_suite_focus=true), and it our test suite to fail with the error `ReferenceError: AST is not defined`.

I believe this is because SWC is not emitting the `AST` namespace due to it being a type-only namespace.

I can't provide a true playground repro because v194 is not available on the playground at the time of filing.

### Input code

```typescript
import { TSESTree, AST_TOKEN_TYPES } from '../ts-estree';

namespace AST {
  export type TokenType = AST_TOKEN_TYPES;

  export type Token = TSESTree.Token;

  export type SourceLocation = TSESTree.SourceLocation;

  export type Range = TSESTree.Range;
}

export { AST };
```

### Config

```json
// consumed via @swc/jest
{
  jsc: {
    target: 'es2019',
    transform: {
      react: {
        runtime: 'automatic',
      },
    },
  },
}
```

### Playground link

https://play.swc.rs/?version=1.2.189&code=H4sIAAAAAAAAA22PPQvCMBCG9%2FsVt3XR%2BgOCg0MnRcXc4lRCOaVIPkgiKCX%2F3TQoWOt2vPc8L3e9dtZHHJBkI8kzL3AjqaXDttm3dD42EhNevNVY1fUqhiWHmKlKABilOTjV8WjgAIj8KGXx6RjJ3tjQOK1%2FG7P7h83c54a6BDNM2rvveGc7FXs74aebmXhS5srffAkEJIA3NpQfkoAX9KmwUxABAAA%3D&config=H4sIAAAAAAAAA0WOSwrDMAxE76J1Fm0XhfoOPYRwleDgH5ICNcZ3r21SshtmRk9TYRcLpkJGFuKhpETFLxjQkkksu6ywgEq3lA9qXSNvpL1B8rjdXz31KQmBWdELLRBcdGsZLJtCZhK5Ioyb%2FzdbZ4X0OYZR57vJfEK7GOedk%2FdZnBt%2BJYyKhrcAAAA%3D

### Expected behavior

Our tests should pass as they did with v1.2.189.

### Actual behavior

[...]

PR #4736, opened 2022-05-21
fix(es/typescript): Mark namespaces as concrete

kdy1:
**Description:**

**Related issue (if exists):**

 - Closes https://github.com/swc-project/swc/issues/2669
 - Closes https://github.com/swc-project/swc/issues/4393

PR #5670, opened 2022-08-30
fix(es/typescript): Mark namespace as concrete only if it is

kdy1:
**Description:**

 - https://github.com/swc-project/swc/pull/4736
 - Closes https://github.com/swc-project/swc/issues/4851


kdy1 (commented):
swc-bump:

Issue #5472, opened 2022-08-13
Enable bytechecked deserialization for the plugin


kwonoj:
### Describe the feature

We are currently doing unchecked deserialization (https://github.com/swc-project/swc/blob/main/crates/swc_common/src/plugin/serialized.rs#L124) for the plugin deserializations.

This was deliberated decision at the moment, however it exposes some of cons recently - as it implies, it is unsafe operation and can take down host randomly without having additional details. We could try `checked` instead to ensure host - plugin can exchange data as expected, and know it'll fail prior to actual serialization.

Unfortunately this is blocked by https://github.com/swc-project/swc/issues/5406, since this requires to enable `bytecheck` for each structs. We may carefully approach enabling bytecheck first to get there.

kwonoj:
https://github.com/swc-project/swc/pull/7280

PR #5686, opened 2022-08-31
feat(swc/atoms): introduce bytecheck rkyv version

kwonoj:
**Description:**

- https://github.com/swc-project/swc/issues/5472

We tried this once aggressively, then reverted. This time, trying to approach bit slow to avoid regressions as much.

Pluing execution with `swc_core` is now somewhat stablized. At least there's no esoteric runtime failure we can't dig enough. Using current state as base would like to enable bytechecked serialization. As first step, creating compile-time flag path to use updated version of rkyv. 

This could be a separate branch, but that requires separate publish pipeline for beta, or tagged version. It is not there yet, so trying to achieve it via compile time flags instead even though it requires some workaround for the cargo's version selection.

**BREAKING CHANGE:**


crates/swc_atoms/Cargo.toml
```diff
@@ -7,18 +7,25 @@ edition       = "2021"
...
 [dependencies]
-bytecheck    = { version = "0.6.9", optional = true }
-once_cell    = "1"
-rkyv         = { version = "=0.7.37", optional = true }
+bytecheck   = { version = "0.6.9", optional = true }
+once_cell   = "1"
+rkyv-stable = { package = "rkyv", version = "=0.7.37", optional = true }
+# This is to avoid cargo version selection conflict between rkyv=0.7.37 and other versions, as it is strictly pinned
+# cannot be merged.
+rkyv-latest  = { git = "https://github.com/rkyv/rkyv", package = "rkyv", tag = "v0.7.38", optional = true }
```
kwonoj, line 28: This is the trick: currently all of swc_* uses pinned rkyv version, specifying another causes version resolution conflict. Workaround by specifying git's tagged sha instead.

kdy1, line 28: I'm not sure if cargo publish will work with this.
Did you verify it?

kwonoj: ugh right, it needs version.

kwonoj: ok, this is tricky:


```
error: failed to verify package tarball

Caused by:
  failed to select a version for `rkyv`.
      ... required by package `swc_atoms v0.4.10 (/Users/ojkwon/github/swc/target/package/swc_atoms-0.4.10)`
  versions that meet the requirements `=0.7.37` are: 0.7.37

  all possible versions conflict with previously selected packages.

    previously selected package `rkyv v0.7.38`
      ... which satisfies dependency `rkyv-latest = "=0.7.38"` of package `swc_atoms v0.4.10 (/Users/ojkwon/github/swc/target/package/swc_atoms-0.4.10)`

  failed to select a version for `rkyv` which could resolve this conflict
```

fails when try to verify publish. I expect this'll occur same on the user (swc_core) as well. :/

kwonoj: Peeking bit, there's no way to escape (https://users.rust-lang.org/t/impact-of-pinning-dependencies/17634/16) with cargo. Probably as a shortcut, the quickest way is to publish dummy pkg doesn't have same pkg name instead of trying to alias. (Or branching, but that'll be more cumbersome for now)

kwonoj: I ended up using forked version to have different pkg name for the versions we want to test (0.7.38, 39).

crates/swc_atoms/Cargo.toml
```diff
@@ -7,18 +7,25 @@ edition       = "2021"
...
 [dependencies]
-bytecheck    = { version = "0.6.9", optional = true }
-once_cell    = "1"
-rkyv         = { version = "=0.7.37", optional = true }
+bytecheck   = { version = "0.6.9", optional = true }
+once_cell   = "1"
+rkyv-stable = { package = "rkyv", version = "=0.7.37", optional = true }
+# This is to avoid cargo version selection conflict between rkyv=0.7.37 and other versions, as it is strictly pinned
+# cannot be merged.
+rkyv-latest  = { package = "rkyv-test", version = "=0.7.38-test.1", optional = true }
```
kdy1, line 28: I don't think this will work for other crates depending on rkyv

kwonoj: once all pkgs have necessay changes with feature flags, it'll all rely on rkyv-test instead. mixing between rkyv / rkyv-test should not happen anytime anyway.

kdy1 (approved):
Let's see if it works...

Thanks!

---


swc-bump:
 - swc_core

kdy1 (changes requested):
Oh there's a conflict

kwonoj:
@kdy1 ok, tried to resolve conflict.

kdy1 (dismissed):

kdy1 (approved):

PR #5698, opened 2022-08-31
feat(es/minifier): Exclude `JSON` from alias list

kdy1:
**Description:**

I found that `JSON` is treated as an alias and it prevents proper minification


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5700, opened 2022-08-31
feat(es/minifier): Move variables declared with `let`

### LINKED ISSUE #5696: minifier should join variable declarations declared with `let`
### Describe the feature

image

swc vs terser

We do it for `VarDeclKind::Var` but not for `VarDeclKind::Let`

kdy1:
**Description:**

To simplify implementation, I decided to make `collapse_vars_without_init` accept a 'target var kind'.
We pass `let` and `var` respectively.

 - Closes #5696


kdy1:
Oh this is wrong because of scope

Issue #5694, opened 2022-08-31
sequential inliner should ignore side effects of nested functions


kdy1:
### Describe the feature

<img width="1302" alt="image" src="https://user-images.githubusercontent.com/29931815/187603339-4bf88add-931c-41c3-9b39-2b92e2c8ca62.png">

swc vs terser

# Input

```js

let updateFunction = (oldValue, ...functionArgs) => {
    let interceptedValue = value2(isControlled ? stateRef.current : oldValue, ...functionArgs);
    onChangeCaller(interceptedValue, ...args);
    if (!isControlled) return interceptedValue;
    return oldValue;
};
setStateValue(updateFunction);

```

PR #5702, opened 2022-08-31
feat(es/minifier): Ignore closure in initializer from sequential inliner

kdy1:
**Description:**

```js
let updateFunction = (oldValue, ...functionArgs) => {
    let interceptedValue = value2(isControlled ? stateRef.current : oldValue, ...functionArgs);
    onChangeCaller(interceptedValue, ...args);
    if (!isControlled) return interceptedValue;
    return oldValue;
};
setStateValue(updateFunction);
```

We can ignore the body of the arrow expression in the initializer because the function body is not evaluated before evaluating `setStateValue`.

 - Closes https://github.com/swc-project/swc/issues/5694


Austaras:
Should we make arrow function inline able? Seems like a more generic fix

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

kdy1:
@Austaras It's a more general fix (and I prefer it), but it will require much more analysis.
Also, this is not about inlining arrow functions. It's about just marking arrow functions as side-effect-free for sequential inliner

Issue #5691, opened 2022-08-31
minifier can convert a method property to an arrow function without checking es version


kdy1:
### Describe the feature

<img width="1267" alt="image" src="https://user-images.githubusercontent.com/29931815/187596848-fd1b3377-84b3-4799-924d-20a5455b76f8.png">

swc vs terser

The method property and the arrow function are both ES2015

PR #5703, opened 2022-08-31
feat(es/minifier): Remove useless ES version check

kdy1:
**Description:**

Method properties and arrow functions are both es2015, so we don't need to check ES version

 - Closes https://github.com/swc-project/swc/issues/5691


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5706, opened 2022-08-31
fix(swc/core): bump versions, separate css

kwonoj:
**Description:**

With latest `swc_core` there's some errors like

```
error[E0277]: the trait bound `Box<swc_ecma_ast::Expr>: From<swc_ecma_ast::Function>` is not satisfied
    --> /Users/ojkwon/.cargo/registry/src/github.com-1ecc6299db9ec823/swc_ecma_transforms_compat-0.128.2/src/es2015/generator.rs:1282:34
     |
1282 | ...                   .into(),
     |                        ^^^^ the trait `From<swc_ecma_ast::Function>` is not implemented for `Box<swc_ecma_ast::Expr>`
     |
     = help: the following other types implement trait `From<T>`:
               <Box<(dyn StdError + 'a)> as From<E>>
               <Box<(dyn StdError + 'static)> as From<&str>>
```

Guess it requires another missing version bump.

Also PR splits swc_core importing whole `swc_css` to importing individual.

**BREAKING CHANGE:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2898,7 +2898,7 @@ dependencies = [
 
 [[package]]
 name = "swc"
-version = "0.222.2"
+version = "0.222.3"
 dependencies = [
  "ahash",
  "ansi_term",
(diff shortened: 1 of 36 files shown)


kdy1 (approved):
Thank you!

---
swc-bump:
 -  swc_core

PR #5713, opened 2022-09-01
refactor(swc/core): adjust namespace

kwonoj:
**Description:**

I realize this bit late, there's duplicated names across different features. Namely, like `ecma_visit` vs `css_visit`, etcs. In case of css we already created namespace `swc_core::css`, but I didn't to it for `ecma_*` makes namespace hierarchy bit weird.

This PR makes breaking change by readjusting namespaces. Now `swc_core::ecma`, and `swc_core::css` contains features accordingly. It is breaking change, but I hope it's reasonably straightforward migration.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks! I adjusted rustdoc command.

---

swc-bump:
 - dbg-swc

kwonoj:
> Thanks! I adjusted rustdoc command

Thanks for catching!

Issue #5709, opened 2022-09-01
DCE must use graph


kdy1:
### Describe the feature

![image](https://user-images.githubusercontent.com/29931815/187861076-e2233e49-09c4-4847-bad4-5365f1d657a7.png)

We need graph to remove this.

I'll implement this after finishing https://github.com/swc-project/swc/pull/5699

Austaras:
I cannot quite get what you mean from the picture above -- but if you're referring to remove unused function that refers each other, I believe we could somewhat use `infects` field in `UsageInfo`

kdy1:
> remove unused function that refers each other

Yeah, it's exact. Sorry for a bad issue 🤣 

Problem is that DCE should use a separate analyzer. With the current pass limit, 2, DCE cannot remove enough items.
That's why I'm working to apply DCE multiple time without relying on `pass`. (https://github.com/swc-project/swc/pull/5699)

PR #5715, opened 2022-09-02
feat(es/minifier): Support cycles in DCE

kdy1:
**Description:**

We need graph to drop 
```js
function a() {
    b()
}
function b() {
    a()
}
```

 - Closes https://github.com/swc-project/swc/issues/5709


kdy1 (commented):
swc-bump:

crates/swc/tests/tsc-references/emitClassDeclarationWithThisKeywordInES6.2.minified.js
```diff
@@ -1,16 +1 @@
 //// [emitClassDeclarationWithThisKeywordInES6.ts]
-class B {
```
sokra, line 2: I feel like this renders these test cases useless and they should better run with DCE disabled for top level symbols

kdy1: Will do it tomorrow, thanks!

PR #5728, opened 2022-09-03
feat: update swc plugin template

magic-akari:
**Description:**

We should add the swc plugin template to the test. This will prevent forgetting to update the template when the plugin is updated.


kdy1 (changes requested):
Can you fix the benchmark action?
Seems like there's an issue with cargo feature unification.

I think `swc_plugin_testing` should be moved to `bindings` workspace

bindings/swc_cli/src/commands/plugin.rs
```diff
@@ -231,55 +238,7 @@ build-wasm32 = "build --target wasm32-unknown-unknown"
...
-    Default::default(),
-    |_| as_folder(TransformVisitor),
-    boo,
-    // Input codes
-    r#"console.log("transform");"#,
-    // Output codes after transformed with plugin
-    r#"console.log("transform");"#
-);"##
-                .as_bytes(),
+            include_bytes!("../../../../crates/swc_plugin_testing/src/template.rs"),
```
kdy1, line 241: This will not work if this crate is published

magic-akari: You are right. I forgot that crate can be installed and used separately.

magic-akari:
I did not managed to move it without build error.

PR #5730, opened 2022-09-03
test(es/tsc): Disable `toplevel` for tsc minifier tests

kdy1:
**Description:**

**BREAKING CHANGE:**


Austaras:
Why?

kdy1:
I started this because I thought too many tests are useless due to new minifier rules (too many blank files), but seems like it was not the case.

PR #5733, opened 2022-09-04
perf(es/minifier): Skip useless analysis if not required

kdy1:
**Description:**

We don't need to run full analyzer if both of options are false


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5739, opened 2022-09-04
chore(es/minifier): Fix example

kdy1:
**Description:**

This is to compare performance of real-world cases


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #5740, opened 2022-09-04
perf(bindings): Enable LTO

kdy1:
**Description:**

This will have noticable effect on the binary size and performance


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #5753, opened 2022-09-05
fix(es): Add paren_remover pass

### LINKED ISSUE #5652: Issue with `Parenthesized Expression`
### Describe the bug

IdentifierReference is allowed in the Parenthesized Expression.
There are a lot of cases involving `Parenthesized Expression` that are incorrectly transformed.

### Input code

```typescript
import { foo } from "foo";

foo();                       // ⭕️
(foo)();                     // ❌

bar?.bar();                  // ⭕️
(bar as any)?.bar();         // ❌

bar.bar ||= 1;               // ⭕️
(bar.bar as number) ||= 1;   // ❌

class Bar {
  #bar = 1;

  bar() {
    this.#bar++;             // ⭕️
    (this.#bar)++;           // ❌
    this.#bar = 1;           // ⭕️
    (this.#bar) = 1;         // ❌
  }
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2019"
  },
  "isModule": true,
  "module": {
    "type": "commonjs"
  }
}
```

### Playground link

[playground](https://play.swc.rs/?version=1.2.244&code=H4sIAAAAAAAAA8vMLcgvKlGoVkjLz1eoVUgrys9VUAKylay5uICUhqa1Anagr6%2FwaO3U9zv6uTSA6jRxKQQpm9vDxZWUWGSvBySwqkMyC6hEIbFYITGvUhNDPZJZICmFmhpbBUN041DN0oOal1eam5RapInQAjMrOSexuFjBCaiqmktBQRmkHKSCC8gB2w4WVlAoycgs1gPJamtb47AOxNOAq9NEVQi1DsUoBTTX4zQKVSHcqFquWi4ADHWK1r8BAAA%3D&config=H4sIAAAAAAAAAy2MSQ7CMBAE%2FzLiGCXADf%2BBR1hmsljxoumJIIr8d4bl1l2q7oNOCDMnT45m1Qo3DBEl9z8KLcJ9kWnAM0igjiICuYOqF7B8Evas%2FmVz3SsjyFLVNIWh0a%2FgZsXLxGoK43q%2B3MjQgnt5bCuTU9m4o%2FRvx%2FfG1FBSKjmCWnsD2rHcj6QAAAA%3D)

### Expected behavior

Parenthesized Expression should be transformed correctly.

### Version

1.2.244

### Additional context

_No response_

### LINKED ISSUE #5752: `delete` used with parentheses always incorrectly leads to TS2703
### Describe the bug

```typescript
let foo: { bar?: number } = { bar: 123 }
delete (foo.bar);
```

The above codes [worked in tsc](https://www.typescriptlang.org/play?#code/DYUwLgBAZg9jBcEDeEBGBDATgfkQOwFcBbVETCAXwgF5k0tEBGAJgGZKAoAExFDBAgAKWDAB0GTAEoA3EA), but [got the following output in swc](https://play.swc.rs/?version=1.2.247&code=H4sIAAAAAAAAA8tJLVFIy8%2B3UqhWSEossrdSyCvNTUotUqhVsIUIWSkYGhkr1HKlpOaklqQqaAAV6wGFNa0BJpypLzoAAAA%3D&config=H4sIAAAAAAAAA0WNSwrDMAxE76J1tt3kDj2ESZXg4B8aBWqM7165OGQnzedNoxMbrY2KE7CMCzWp%2B9JKWgtjE1%2BUFlKYtLsA7vY4OVgtwniZF3IGT3eh6JPf6yBtORZh4LFcOsKd7AaK%2BXMNof3HDGiVmNMJ6g9olj3eM61ycf8BnwbyVboAAAA%3D):

```
  x The operand of a delete operator must be a property reference.
   ,----
 2 | delete (foo.bar);
   :         ^^^^^^^
   `----


Caused by:
    0: failed to parse code
    1: error was recoverable, but proceeding would result in wrong codegen
    2: Syntax Error
```

### Input code

_No response_

### Config

_No response_

### Playground link

https://play.swc.rs/?version=1.2.247&code=H4sIAAAAAAAAA8tJLVFIy8%2B3UqhWSEossrdSyCvNTUotUqhVsIUIWSkYGhkr1HKlpOaklqQqaAAV6wGFNa0BJpypLzoAAAA%3D&config=H4sIAAAAAAAAA0WNSwrDMAxE76J1tt3kDj2ESZXg4B8aBWqM7165OGQnzedNoxMbrY2KE7CMCzWp%2B9JKWgtjE1%2BUFlKYtLsA7vY4OVgtwniZF3IGT3eh6JPf6yBtORZh4LFcOsKd7AaK%2BXMNof3HDGiVmNMJ6g9olj3eM61ycf8BnwbyVboAAAA%3D

### Expected behavior

Should not report any errors.

### Actual behavior

_No response_

### Version

1.2.247

### Additional context

I found that the TS2703 error is thrown by `swc_ecma_parser/src/parser/expr/ops.rs`:

```rust
match &*arg {
    Expr::Member(..) => {}
    Expr::OptChain(OptChainExpr {
        base: OptChainBase::Member(..),
        ..
    }) => {}
    _ => self.emit_err(unwrap_paren(&arg).span(), SyntaxError::TS2703),
}
```

Here the branches does not take `Paren`s into account.

Would love to open a PR for this if it is confirmed as a bug!

magic-akari:
**Description:**

**BREAKING CHANGE:**

- fix #5652 
- fix #5752


crates/swc/src/config/mod.rs
```diff
@@ -598,6 +599,7 @@ impl Options {
                 }),
                 syntax.decorators()
             ),
+            paren_remover(Some(Box::leak(Box::new(comments)) as _)),
```
magic-akari, line 602: I think this is wrong.
Do you have any idea about how to pass the `comments `? @kdy1

Austaras:
No, this is not enough -- it would only remove useless paren, but would preserve semantic meaningful paren.

crates/swc/src/config/mod.rs
kdy1: You can simply clone it. https://rustdoc.swc.rs/swc_common/comments/trait.Comments.html#implementation-notes

magic-akari: do you mean `paren_remover(comments.clone())`?
I got this error.

```
error[E0308]: mismatched types
   --> crates/swc/src/config/mod.rs:602:27
    |
602 |             paren_remover(comments.clone()),
    |             ------------- ^^^^^^^^^^^^^^^^ expected trait object `dyn Comments`, found struct `SingleThreadedComments`
    |             |
    |             arguments to this function are incorrect
    |
    = note: expected enum `std::option::Option<&dyn Comments>`
               found enum `std::option::Option<&SingleThreadedComments>`
```

magic-akari:
> No, this is not enough -- it would only remove useless paren, but would preserve semantic meaningful paren.

All parens will be removed except paren in `(a?.b).c`.

crates/swc/src/config/mod.rs
kdy1: I think casting in `Option.map` may work

magic-akari:
Test case failed with concurrent mode.
Do you have any idea? @kdy1

crates/swc_common/src/comments.rs
```diff
@@ -381,7 +381,13 @@ impl Comments for SingleThreadedComments {
     fn move_leading(&self, from: BytePos, to: BytePos) {
         let cmt = self.leading.borrow_mut().remove(&from);
 
-        if let Some(cmt) = cmt {
+        if let Some(mut cmt) = cmt {
+            // TODO: if let chain
+            if from < to {
+                if let Some(to_cmt) = self.leading.borrow_mut().remove(&to) {
```
magic-akari, line 387: I got wrong result with these changes in concurrent mode.

kdy1: Hmm... I have no idea why concurrent mode fails

kdy1: Not sure why is this logic required

magic-akari: I am trying to keep the order of comments.

magic-akari: See https://github.com/swc-project/swc/commit/877a3a5f87b3ba72d1be6107ef536443070ef397

magic-akari: I revert it and get the same error 😅.

kdy1: Seems like current CI failure is simple update issue?

magic-akari: No. I test it on my machine, the generated results changed randomly every time.
I suspect there is something wrong with `Comments` and concurrent mode.

kdy1: Ah... I'll investigate, thanks!

kdy1:
I think flaky tests are related to iteration order of `AHashMap` or something similar which uses `ahash::RandomState`

magic-akari:
SingleThreadedCommentsMap is using FxHashMap.

kdy1:
https://github.com/swc-project/swc/blob/fcf4deb5f0c04ce0e2ea588cf6e5a6176c14675d/crates/swc_ecma_transforms_base/src/fixer.rs#L39

`fixer` and `paren_remover` is using `AHashMap` (I think it should not, though)

crates/swc_ecma_transforms_base/src/fixer.rs
```diff
@@ -36,7 +37,7 @@ struct Fixer<'a> {
     ///
     /// Key is span of inner expression, and value is span of the paren
     /// expression.
-    span_map: AHashMap<Span, Span>,
+    span_map: FxHashMap<Span, Span>,
```
kdy1, line 40: I think we should use `IndexMap`. Iteration order of `fxhash` is deterministic but it's different from it of a vector.

kdy1 (changes requested):
Thank you and sorry for a late review

crates/swc_ecma_transforms_base/src/fixer.rs
magic-akari: I expect get the same result from any order of iteration.

for example.

```JavaScript 
// span1
(
    // span2
    (
        // span3
        expr
    )
)
```

We will remove the parens and merge all comments. 
And we get the map

```
span_map {
    span2 -> span1,
    span3 -> span2,
}
```

Whether you get [`span1 span2`, `span3`] and then `span1 span2 span3` , or [`span1`, `span2 span3`] and then  `span1 span2 span3`, they should be same.

So FxHashMap should be fine.

But I don't know why AHashMap doesn't work, maybe it will lose data in some cases.

kdy1: Oh you are right, thanks

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_common

PR #5757, opened 2022-09-05
fix(es/minifier): don't drop params in exported function

Austaras:
**Description:**

Terser has since fixed this bug and it's no longer needed.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_minifier

PR #5761, opened 2022-09-06
perf(es/minifier): Make `MultiReplacer` single-pass

kdy1:
**Description:**

We don't have to recurse fully. We can simply check `e` and children nodes


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5764, opened 2022-09-06
fix(es/tsc-test): Handle tsx

magic-akari:
**Description:**

**BREAKING CHANGE:**


crates/swc/tests/tsc-references/checkJsxChildrenProperty12.1.normal.js
```diff
@@ -1,7 +1,7 @@
 //// [file.tsx]
 //! 
-//!   x Expected '>', got '{'
-//!     ,----
-//!  14 | return <InnerButton {...this.props} />
-//!     :                     ^
-//!     `----
+//!   x Import assignment cannot be used when targeting ECMAScript modules. Consider using `import * as ns from "mod"`, `import {a} from "mod"`, `import d from "mod"`, or another module format instead.
+//!    ,----
+//!  2 | import React = require('react');
```
magic-akari, line 5: Maybe we should set `CommonJS` as default Module config.

magic-akari: Yes.

https://github.com/swc-project/swc/blob/390b1d12e30a30b4791f493e34bfc8b6a790c85d/crates/swc/tests/tsc.rs#L112-L114

and

https://github.com/swc-project/swc/blob/390b1d12e30a30b4791f493e34bfc8b6a790c85d/crates/swc/tests/tsc.rs#L175-L177

kdy1: I'm fine with changing the default

magic-akari: Hmmm. too many files changed. should not do it in this pull request.

crates/swc/tests/tsc-references/jsxParsingError2.1.normal.js
```diff
@@ -2,37 +2,36 @@
 //// [file.tsx]
 //// [Error1.tsx]
 //! 
-//!   x Unexpected eof
+//!   x Unexpected token `div`. Expected jsx identifier
```
magic-akari, line 5: confusing error message. We can improve it.

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc

PR #5770, opened 2022-09-06
fix(es/minifier): use seperate scope for class

Austaras:
**Description:**
This would fix two issues --
```js
export class A {
   foo() {
      var o = this
      return class {
        x = console.log(o)  
      }
   }
}
```
and
```js
export var a

(a = new Foo())
```

**BREAKING CHANGE:**


crates/swc/tests/vercel/full/next-33088/output/index.js
```diff
@@ -14,8 +14,8 @@ export default function i() {
         function t() {
             return (t = r(function() {
                 var r, t;
-                return e(this, function(e) {
-                    switch(e.label){
+                return e(this, function(r) {
```
kdy1, line 17: This seems wrong.

crates/swc/tests/vercel/full/next-33088/output/index.js
```diff
@@ -24,10 +24,10 @@ export default function i() {
                         case 1:
                             return [
                                 4,
-                                (r = e.sent()).json()
+                                r.sent().json()
```
kdy1, line 27: `r = e` is only true if this line is evaluated

crates/swc/tests/vercel/full/utf8-1/output/index.js
```diff
@@ -29,17 +29,17 @@ function p() {
                 case 2:
                     return [
                         4,
-                        (e = t.sent()).json()
+                        e.sent().json()
```
kdy1, line 32: Same here

kdy1 (commented):
Thanks!

crates/swc/tests/vercel/full/next-33088/output/index.js
Austaras: r and e doesn't get referenced anywhere, so assignment is removed. You can repplicate this in terser, swc currently only needs more passes to remove unused vars.

kdy1: I see. My bad, thank you!

kdy1 (approved):
Thank you!
I'll merge https://github.com/swc-project/swc/pull/5772 after merging this.

---

swc-bump:
 - swc_ecma_minifier

PR #5771, opened 2022-09-06
test(swc): e2e for the published plugins

kwonoj:
**Description:**

This PR enables naive tests to verify swc_core compats with existing plugins. It doesn't cover when intended breaking changes occur though. Probably we'll need some manual way to toggle test on-off for those, but at least this will prevent accident breaking changes when we try to bump up swc_core for the bindings.

Given there is an issue https://github.com/swc-project/swc/issues/5540, probably this test need to be moved into `bindings/` eventually. For now uses existing test setup with other tests.

**BREAKING CHANGE:**


kdy1 (commented):
This test will not fail on the problematic PR.
Instead, it will fail on version bump.

Maybe we should change this to rust-based test for now

kwonoj:
> will fail on version bump

yes, that's current intention.

> Maybe we should change this to rust-based test for now

I don't think this will work unfortunately, as we don't have to way to create rust pkg(swc_core) as reproducible to be published. The regressions we experienced were only found when we used published swc_core, while we couldn't with locally built bindings using path based dependency resolution.

I hope to improve things further, but at least this test will let us know before publish @swc/core if something goes unexpected.

kdy1 (approved):
I see, thank you!

---

swc-bump:
 - dbg-swc

PR #5773, opened 2022-09-07
perf(es/minifier): Make passes more parallel

kdy1:
**Description:**

This had noticable effect on real-world apps.

## main

```
Using 999 files
Took 11.499062833s
```

## After parallelizing tree shaker

```
Using 999 files
Took 10.688992916s
```


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base

PR #5779, opened 2022-09-07
fix(html/parser): potential bug in spec

alexander-akait:
**Description:**

I found a bug in HTML spec, so I started to fix it in spec and our parser + tests for `html5tests-lib` for other parser

Original issue - https://github.com/whatwg/html/issues/7922

So we are showring potential fix in spec/parser here (and no problems with other tests). Let's merge when HTML spec marked it as bug and fix it in spec.

**BREAKING CHANGE:**

No

Ref: https://github.com/whatwg/html/issues/7922
Ref: https://github.com/html5lib/html5lib-tests/pull/150


alexander-akait:
@kdy1 Ready to merge, the bug was fixed in spec

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser

PR #5780, opened 2022-09-08
feat(es/minifier): Change order of renaming to align with terser

kdy1:
**Description:**

 - Change the order of renaming to match one of terser.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5781, opened 2022-09-08
feat(es/minifier): Reuse variable names

kdy1:
**Description:**

Currently, the name mangler works by processing upper scope and processing children scopes in parallel while avoiding symbols used by the parent scope.
But I want to reuse variable names.

The problem is that I can't know if a variable name is safe to reuse without looking at children nodes.


kdy1:

PR #5790, opened 2022-09-09
refactor(es/renamer): Rename `PARALLEL` to `MANGLE`

kdy1:
**Description:**

Actually, performance issue comes from collision, not parallelism. So I renamed the associated constant.


kdy1 (commented):
swc-bump:

PR #5791, opened 2022-09-09
chore: use `yarn@3.2.3`

magic-akari:
**Description:**

**BREAKING CHANGE:**

- #https://github.com/swc-project/swc/discussions/5614

### DIFF
--- a/.github/workflows/publish-node.yml
+++ b/.github/workflows/publish-node.yml
@@ -249,6 +249,9 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Enable Corepack
+        run: corepack enable
+
       - uses: actions/setup-node@v3
         with:
           node-version: "16"
@@ -363,6 +366,8 @@ jobs:
     runs-on: ${{ matrix.settings.host }}
     steps:
       - uses: actions/checkout@v3
+      - name: Enable Corepack
+        run: corepack enable
       - name: Setup node
         uses: actions/setup-node@v3
         with:
@@ -402,6 +407,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - name: Enable Corepack
+        run: corepack enable
       - name: Setup node
         uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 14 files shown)


Brooooooklyn:
I prefer `nodeLinker: node-modules` config for yarn@3, default `nodeLinker` is pnp

Austaras:
Maybe the simplest solution is revert to yarn 1 and update yarn lock.

kdy1:
Maybe, I changed it to reduce time but if this is problematic I think we can go back to yarn v1

.github/workflows/publish-node.yml
```diff
@@ -228,7 +228,6 @@ jobs:
           image: ${{ matrix.settings.docker }}
           options: -v ${{ env.HOME }}/.cargo/git:/root/.cargo/git -v ${{ env.HOME }}/.cargo/registry:/root/.cargo/registry -v ${{ github.workspace }}:/build -w /build
           run: >-
-            yarn global add wasm-pack &&
```
magic-akari, line 231: https://github.com/swc-project/swc/blob/c746611f262d8b3fbac1c29c23f6c2a4c5f3ce2b/.github/workflows/publish-node.yml#L219-L223

We have installed wasm-pack above these lines.
Is it necessary to install it again in docker step?
@kdy1

kdy1: I don't think so, seems like a mistake

ojkelly:
> I prefer `nodeLinker: node-modules` config for yarn@3, default `nodeLinker` is pnp

Just for what it's worth, you can use which ever linker you prefer. In most cases `pnp` is good choice, if it's not, or if it introduces some breaking changes you can safely use `node-modules`.

magic-akari:
@Brooooooklyn @ojkelly 

https://github.com/swc-project/swc/blob/360767394fcd99703dc351150d3c9aac66807d38/.yarnrc.yml#L1

It's already `node-modules`

magic-akari:
blocked by nested package structure.
Maybe we should use yarn workspace.

PR #5792, opened 2022-09-09
refactor(swc_core): enable bytechecked flag

kwonoj:
**Description:**

Another update to enable bytecheck rkyv feature flag.

Our dep tree is quite tangled with implicit enable (plugin -> rkyv). PR tries to detach some of it while trying to preserve existing behavior as much as it can.

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_common/src/plugin/mod.rs
+++ b/crates/swc_common/src/plugin/mod.rs
@@ -1,7 +1,7 @@
 pub mod diagnostics;
 pub mod metadata;
-#[cfg(feature = "plugin-base")]
-#[cfg_attr(docsrs, doc(cfg(feature = "plugin-base")))]
+#[cfg(feature = "__plugin")]
+#[cfg_attr(docsrs, doc(cfg(feature = "__plugin")))]
 pub mod serialized;
 
 /**
--- a/crates/swc_plugin_runner/src/lib.rs
+++ b/crates/swc_plugin_runner/src/lib.rs
@@ -7,8 +7,11 @@ use transform_executor::TransformExecutor;
 
 pub mod cache;
 mod host_environment;
+#[cfg(feature = "__rkyv")]
 mod imported_fn;
+#[cfg(feature = "__rkyv")]
 mod load_plugin;
+#[cfg(feature = "__rkyv")]
 mod memory_interop;
 mod transform_executor;
 
@@ -20,6 +23,7 @@ mod transform_executor;
  * Note you CANNOT reuse executor once trasform has been executed: executor
  * is stateful.
  */
+#[cfg(feature = "__rkyv")]
 pub fn create_plugin_transform_executor(
     path: &Path,
     cache: &Lazy<cache::PluginModuleCache>,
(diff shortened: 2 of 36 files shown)


kdy1 (commented):
Did you verify rustdoc action, too?

kwonoj:
Do you mean this? https://github.com/swc-project/swc/blob/54b108dc5133e27d8b4c48b35892a7928174fce7/.github/workflows/docs.yml#L25 at least it seems to build without failure.

crates/swc_common/src/plugin/mod.rs
```diff
@@ -1,7 +1,7 @@
 pub mod diagnostics;
 pub mod metadata;
-#[cfg(feature = "plugin-base")]
-#[cfg_attr(docsrs, doc(cfg(feature = "plugin-base")))]
+#[cfg(feature = "__plugin")]
```
kdy1, line 3: I think this name is misleading

kwonoj: I feel overall it's a bit complex with namings, while I choose the short path as I think we may get rid of all of this once verification is done. If we're going to keep this permanent, it'd be probably better to organize bit more.

crates/swc_plugin_runner/src/lib.rs
```diff
@@ -7,8 +7,11 @@ use transform_executor::TransformExecutor;
 
 pub mod cache;
 mod host_environment;
+#[cfg(feature = "__rkyv")]
```
kdy1, line 10: I think it's better to have

```rust
#[cfg(all(not(feature = "rkyv"), not(feature = "rkyv-bytecheck-impl")))
compile_error!("Error message")
```

kdy1: `swc_plugin_*` is useless if there's no rkyv, and we should fail on compile time instead of runtime

kdy1 (changes requested):
Thanks!

crates/swc_plugin_runner/src/lib.rs
kwonoj: There's no fancy message, but this still fails on compile time if any plugin transform misses rkyv-impl or rkyv-bytecheck-impl.

kwonoj: after all, I'm not too much worried about those case as `swc_core` should have right control for those, between `plugin_transform` vs `plugin_transform_bytecheck`.

crates/swc_plugin_runner/src/lib.rs
```diff
@@ -29,3 +33,14 @@ pub fn create_plugin_transform_executor(
...
+
+#[cfg(not(feature = "__rkyv"))]
+pub fn create_plugin_transform_executor(
+    path: &Path,
+    cache: &Lazy<cache::PluginModuleCache>,
+    source_map: &Arc<SourceMap>,
+    metadata_context: &Arc<TransformPluginMetadataContext>,
+    plugin_config: Option<serde_json::Value>,
+) -> Result<TransformExecutor, Error> {
+    unimplemented!("Transform plugin cannot be used without serialization support")
```
kdy1, line 45: I was worried about this

kdy1: Is this really required?
I think it should not compile if there's no rkyv

kwonoj: all of the imports are guarded by flags, this won't occur in normal condition. This was introduced to let plain `cargo check` does not fail due to missing exported signatures.

I'd like to remove it, but then all of CI / and `cargo check` will break right away.

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_atoms
 - swc_common --breaking

PR #5800, opened 2022-09-10
perf(css/minifier): Setup benchmark

kdy1:
**Description:**

I created a simple benchmark for the css minifier. It's full benchmark, which also profiles parser and codegen


kdy1 (commented):
swc-bump:
 - swc_css_minifier

PR #5803, opened 2022-09-10
perf(es/codegen): Skip useless operations if not required

kdy1:
**Description:**

The column is used only for sourcemap, so we can skip `chars().count()` if we are not generating source maps.


kdy1 (commented):
swc-bump:
 - swc_ecma_codegen

PR #5804, opened 2022-09-10
perf(css/parser): Skip whitespaces from lexer

kdy1:
**Description:**

Lexer can handle this much more efficiently.


kdy1 (commented):
swc-bump:
 - swc_css_parser

Issue #5765, opened 2022-09-06
tsxGenericArrowFunctionParsing


magic-akari:
### Describe the bug

found in #5764 

### Input code

```typescript
var x1 = <T>() => {}</T>;

var x2 = <T extends {}>() => {};
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "target": "es5"
  },
  "module": {
    "type": "es6"
  },
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.248&code=H4sIAAAAAAAAAytLLFKoMFSwVbAJsdPQVLC1U6iutdEPsbPm4ioDSRmBpRRSK0pS81KKgZJwVdYAQhjHbTsAAAA%3D&config=H4sIAAAAAAAAAy2LSwqAMAxE7zJrt7roHTxEqUEqfkqSglK8u0lxN%2FNmXsMmCaGhRBZiT%2FKcGm8E6FNIEueiGKBiSLnSaznySmoPkhHWj2upO7nrSueT8yzzv3TxA8O52TtsAAAA

### Expected behavior

error report.
tsc: 
```
Unexpected token. Did you mean `{'>'}` or `&gt;`?
```

### Actual behavior

RuntimeError: unreachable

### Version

1.2.248

PR #5805, opened 2022-09-10
chore(ecma_parser): Fixed jsx scanner error. (issue related to tsxGenericArrowFunctionParsing)

IWANABETHATGUY:
**Description:**
1. Fixed jsx scanner error.
2. Related code you could found here https://github.dev/microsoft/TypeScript/blob/eb40134373cc524bd15432e15490d4e369754c50/src/compiler/scanner.ts#L2312-L2317

**BREAKING CHANGE:**

Fixed https://github.com/swc-project/swc/issues/5765


crates/swc/tests/tsc-references/jsxParsingError3.1.normal.js
```diff
@@ -1,14 +1,44 @@
 //// [jsxParsingError3.tsx]
 //// [file.tsx]
 //// [Error1.tsx]
-var x1 = /*#__PURE__*/ React.createElement("div", null, "}");
+//!
```
IWANABETHATGUY, line 4: I don't know why the error message is wrong, but if you put it under `ecma_parser/tests/err/`, the error message should be right.
截图 2022-09-10 15-14-31

crates/swc_ecma_parser/tests/jsx/errors/13/input.js.stderr
```diff
@@ -0,0 +1,6 @@
+
```
IWANABETHATGUY, line 1: This file change is expected, the error is the same as typescript, https://www.typescriptlang.org/play?#code/DwGQpgZgLgSglgcwBZQAQBtJQLzAIaoD0AfKgE6Iq4BGxAtgPYB2A1mAJ4DORpCDF6dHk7BCtHkA

crates/swc_ecma_parser/tests/typescript/variance-annotations/with_jsx/input.tsx
```diff
@@ -1,5 +1,3 @@
-const validJSX = <in T>() => {}</in>;
```
IWANABETHATGUY, line 1: This jsx snippet is not valid, since it has syntax error, so I removed it.

crates/swc/tests/tsc-references/jsxParsingError3.1.normal.js
kdy1: Maybe it's related to parser mode being typescript

IWANABETHATGUY: So would you mind keeping it as is for now?
Maybe we could change it in a separated pr, this is not the unique place that has this issue. https://github.com/swc-project/swc/blob/52f6ca5fb826b9952b32946379cd696afb58867b/crates/swc/tests/tsc-references/jsxParsingError2.1.normal.js#L14

.vscode/settings.json
```diff
@@ -25,9 +25,9 @@
     "eslint.enable": false,
     "cSpell.allowCompoundWords": true,
     "cSpell.caseSensitive": true,
-    "rust-analyzer.checkOnSave.command": "clippy",
+    // "rust-analyzer.checkOnSave.command": "clippy",
```
IWANABETHATGUY, line 28: Running `clippy`  when onSave is slow on my computer, I will recover it when this pr is finished.

crates/swc/tests/tsc-references/jsxParsingError3.1.normal.js
kdy1: I think it's fine to fix it in another PR

.vscode/settings.json
kdy1: Can you revert this?

crates/swc_ecma_parser/src/error.rs
```diff
@@ -684,6 +687,17 @@ impl SyntaxError {
             )
             .into(),
             SyntaxError::SetterParamRequired => "Setter should have exactly one parameter".into(),
+            SyntaxError::DidYouMean {
```
kdy1, line 690: This name looks like a bad name for an error.

kdy1 (changes requested):
Thanks!

crates/swc_ecma_parser/src/error.rs
IWANABETHATGUY: Any naming suggestions?
What about `UnexpectedTokenWithSuggestion`?

kdy1: I think it's a good name

IWANABETHATGUY: Done

.vscode/settings.json
IWANABETHATGUY: Done

crates/swc_ecma_parser/tests/jsx/errors/13/input.js.stderr
kdy1: But it's js, not ts.

https://babeljs.io/repl#?browsers=node%200.8&build=&builtIns=false&corejs=false&spec=false&loose=false&code_lz=DwGQpgZgLgSglgcwBZQAQBtJQLzAIaoD0AfKgE6Iq4BGxAtgPYB2A1mAJ4DORpCDF6dHk7BCtHkA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=env%2Creact&prettier=false&targets=&version=7.19.0&externalPlugins=&assumptions=%7B%7D

IWANABETHATGUY: Babel will not emit any error even in tsx

IWANABETHATGUY: So do we have the ability to emit a recoverable error?

kdy1: Yes, `emit_err` exists for it

kdy1 (approved):
Thanks!


---

swc-bump:
 - swc_ecma_parser

PR #5807, opened 2022-09-10
perf(css/parser): Use `JsWord` instead of `&str`

kdy1:
**Description:**


kdy1:
This didn't improve performance

PR #5808, opened 2022-09-10
perf(css/parser): Use ascii operations

kdy1:
**Description:**

We don't need unicode handling in these cases.


kdy1 (commented):
swc-bump:
 - swc_css_parser

PR #5810, opened 2022-09-10
perf(css/parser): Skip whiltespace from lexer

kdy1:
**Description:**

Lexer can handle it more efficiently.


kdy1 (commented):
swc-bump:
 - swc_css_parser

PR #5811, opened 2022-09-10
fix(node/types): correct type for jsc.paths

SukkaW:
**Description:**

Fix the type definition of `jsc.paths`.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - dbg-swc

PR #5813, opened 2022-09-10
chore: Publish `v1.3.0`

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - dbg-swc

mischnic:
Were there any breaking changes in the npm packages between 1.3.0 and 1.2.249? Or was this mostly because the patch version number became quite large?

kdy1:
There was no breaking change. That's why I didn't bump the major version.

I just wanted to align with next.js

PR #5814, opened 2022-09-10
perf(es/minifier): Optimize usage of `IndexSet`

kdy1:
**Description:**

This is a really small change, but I'm creating a PR because I want to test compiler options with a separate PR.


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base

PR #5817, opened 2022-09-10
perf(es/minifier): Reduce edges of the graph for DCE

kdy1:
**Description:**

We don't need to render the whole graph.

e.g.
`A -> B -> C` is better than `A -> B -> C && A -> C`


kdy1 (commented):
swc-bump:

PR #5820, opened 2022-09-11
perf(css/minifier): Merge passes

kdy1:
**Description:**

Single visitor is better for performance.


kdy1 (commented):
swc-bump:
 - swc_css_minifier

PR #5821, opened 2022-09-11
perf(css/minifier): Merge more passes into the compressor

kdy1:


kdy1 (commented):
swc-bump:
 - swc_css_minifier

PR #5822, opened 2022-09-11
perf(css/minifier): Merge passes and reduce clone

kdy1:


kdy1 (commented):
swc-bump:
 - swc_css_minifier

Issue #5688, opened 2022-08-31
`css = (rule = css)` should be optimized


kdy1:
### Describe the feature

<img width="1267" alt="image" src="https://user-images.githubusercontent.com/29931815/187594276-6b589896-64b5-4082-be91-3880a968e088.png">

Input:

```js

var sanitize = function sanitize(rule) {
  return rule.replace(/\/style/gi, '\\/style');
};


function computeSelector(id, css) {
  var selectoPlaceholderRegexp = /__jsx-style-dynamic-selector/g; // Sanitize SSR-ed CSS.
  // Client side code doesn't need to be sanitized since we use
  // document.createTextNode (dev) and the CSSOM api sheet.insertRule (prod).

  if (typeof window === 'undefined') {
    css = sanitize(css);
  }

  var idcss = id + css;

  if (!cache[idcss]) {
    cache[idcss] = css.replace(selectoPlaceholderRegexp, id);
  }

  return cache[idcss];
}
```

[...]

PR #5832, opened 2022-09-12
feat(es/minifier): Extend inlining ident to multi usage

Austaras:
**Description:**

And fix an issue where swc inline global idents like
```js
var a = g
console.log(a)
```
due to side effect evalutaion order

 - Closes https://github.com/swc-project/swc/issues/5688


crates/swc/tests/tsc-references/recursiveInitializer.2.minified.js
```diff
@@ -1,2 +1,2 @@
 //// [recursiveInitializer.ts]
-var n1 = n1++, n2 = n2 + n2, n3 = n3 + n3, s2 = s2 + s2, s3 = s3 + s3, b3 = !b3 || b3, b4 = !b4 && b4;
+var n1 = n1++, n2 = n2 + n2, n3 = n3 + n3, s1 = s1 + "", s2 = s2 + s2, s3 = s3 + s3, s4 = "" + s4, b1 = !b1, b2 = !!b2, b3 = !b3 || b3, b4 = !b4 && b4;
```
Austaras, line 2: This is a regressiong, however I don't believe people would write code like this.

crates/swc_ecma_minifier/tests/terser/compress/inline/inline_annotation_2/output.js
```diff
@@ -1 +1,2 @@
-console.log(42, NaN);
+const b = NaN;
```
Austaras, line 1: This is also a regression, I will fix it in another PR.

kdy1:
It failed only in concurrent mode...
Not sure what's the cause

kdy1:
Oh, we only verify benchmarks in concurrent mode.
I think this will happen even if you use single-thread mode

Austaras:
Luckily this can be reduced to a very simple reproduction.

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_minifier

Issue #5084, opened 2022-07-01
Converted CommonJS cannot be imported using native ESM


sounisi5011:
### Describe the bug

In swc 1.2.206 and later, exported fields are assigned to the `exports` variable using the internal function `_export()`. However, the native ESM in Node.js cannot discover dynamically defined export fields.

### Input code

```typescript
import path from 'path';

export const PACKAGE_ROOT = path.resolve(__dirname, '../..');

export function getFixturesPath(subpath: string): string {
    return path.resolve(__dirname, '../fixtures', subpath);
}
```

### Config

```json
// This config was generated by @swc/jest
{
  "jsc": {
    "parser": {
      "syntax": "typescript"
    },
    "target": "es2020",
    "transform": {
      "hidden": {
        "jest": true
      }
    }
  },
  "sourceMaps": "inline",
  "module": {
    "type": "commonjs"
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.206&code=H4sIAAAAAAAAA32NMQvCMBCF9%2FyK29JCibviUEQdHFrEPdR6rQGTlMtFCuJ%2FNy0VN296cO%2F7nrGDJ4ah4Tt05C3IKcqNEDjOn9a7wFCXu1N53OtzVV1gO9cVYfCPJ2Za3wy5xmIBUqmVUjL%2F4V10LRvvoEc%2BmJFjoupEZyFeJ8saApNxff4N8BKQjjBV3d%2BhbtHJAhZZ2n2LD9YBty%2FSAAAA&config=H4sIAAAAAAAAAxWNQQrDMAwE%2F6JzDyHH%2FKGPELbS2sSykWRoCPp75duyO8M%2BUDXB8cBAUZKV9GbDHxxg9yBNUoaBv8BQPmRRk%2B7bvkE0gqxnl7asb8mZeKVKGpjJJPfwtE9J9Mah4Ra%2BClO4red50cLXSyypt9a5Krj%2FAe6AQguUAAAA

### Expected behavior

The converted CommonJS file can be imported using the `import` statement.

### Actual behavior

If you try to import on Node.js, which supports native ESM, the following error is thrown:

```
import { getFixturesPath, PACKAGE_ROOT } from './helpers/index.cjs';
                          ^^^^^^^^^^^^
SyntaxError: Named export 'PACKAGE_ROOT' not found. The requested module './helpers/index.cjs' is a CommonJS module, which may not support all module.exports as named exports.
[...]

PR #5835, opened 2022-09-12
fix(es/module): Improve compatibility with `cjs-module-lexer`

magic-akari:
**Description:**

Support reexport by emitting TS `__export` function call.

**BREAKING CHANGE:**


kdy1:
Why is this required? Do you have any real-world usecase related to this?

magic-akari:
Sometimes you need CommonJS while allowing them to be imported into node esm.
Node supports interoperability features by detecting CJS exports through the `cjs-module-lexer` library.

- #5084  can be considered a use case.

----

There are two issues about compatibility with `cjs-module-lexer` that need to be fixed now.

1. reexport

The following code will output `app page` in native esm.
The `page` is reexport from `page.js` to `app.js`.

```JavaScript
// page.js
export const page = "page";
```

```JavaScript
// app.js
export * from "./page.js";

export const app = "app";
```

```JavaScript
// index.js
import { app, page } from "./app.js";

console.log(app, page);
```

When compiling `app.js` and `page.js` into CommonJS, we expect the same result.
Therefore, we must tell node that we have re-export the `page.js` from `app.js`

We emit `0 && __export(require("foo"))` to hint we reexport "foo".

This is similar to the way used by esbuild. 
- evanw/esbuild/issues/2486#issuecomment-1222680909

---

2. arrow getter

Take the same example as above, we will get the following transformed code.

```JavaScript
Object.defineProperty(exports, "page", {
    enumerable: true,
    get: ()=>page
});
```

Once `cjs-module-lexer` detects the `page`, but not in the format it expects,
`cjs-module-lexer` adds it to unsafe getter export and never exports it.

https://github.com/nodejs/cjs-module-lexer/blob/4641ae5419634deeefa41f943080f2eb7b2c925e/src/lexer.c#L442-L444
https://github.com/nodejs/cjs-module-lexer/blob/4641ae5419634deeefa41f943080f2eb7b2c925e/src/lexer.js#L31-L35

- I tried to add support for arrow getters nodejs/cjs-module-lexer/pull/74

We need to change the output to a method/function getter until `cjs-module-lexer` supports arrow getters.

kdy1 (approved):
It makes sense, thanks!

---

swc-bump:
 - swc_ecma_transforms_module

PR #5841, opened 2022-09-13
perf(es/minifier): Make `expr_simplifier` more parallel

kdy1:
**Description:**

It was not parallel enough.


kdy1 (commented):
swc-bump:

PR #5843, opened 2022-09-13
perf(es/minifier): Reduce type sizes

kdy1:
**Description:**

I reduced type size to make it faster on cpus with 64 byte cache line


kdy1 (commented):
swc-bump:

PR #5848, opened 2022-09-13
fix(swc/core): fix export for testing

kwonoj:
**Description:**

Minor fix to correctly reexport testing utilities.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!
---
swc-bump:
 - swc_core

Issue #5846, opened 2022-09-13
function arguments are unexpectedly replaced with void 0 in recursive functions


chentsulin:
### Describe the bug

When "Minify" and "Compress" are enabled and all of compress options are disabled, function arguments seem be unexpectedly transformed in recursive functions.

![](https://user-images.githubusercontent.com/3382565/189840498-872cb285-d44e-4ad4-a196-d6ba1b4fc124.png)

And the following code is broken in the latest Next v12.3.0 with `swcMinify: true`.

### Input code

```typescript
function processNode(
    node: DataNode,
    index?: number,
    parent?: { node: DataNode; pos: string; level: number },
    pathNodes?: DataNode[],
  ) {
    const children = node ? node[mergeChildrenPropName] : dataNodes;
    const pos = node ? getPosition(parent.pos, index) : '0';
    const connectNodes = node ? [...pathNodes, node] : [];

    // Process node if is not root
    if (node) {
      const key: Key = syntheticGetKey(node, pos);
      const data = {
        node,
        index,
        pos,
        key,
        parentPos: parent.node ? parent.pos : null,
        level: parent.level + 1,
        nodes: connectNodes,
      };

      callback(data);
    }

    // Process children node
    if (children) {
      children.forEach((subNode, subIndex) => {
        processNode(
          subNode,
          subIndex,
          {
            node,
            pos,
            level: parent ? parent.level + 1 : -1,
          },
          connectNodes,
        );
      });
    }
  }

  processNode(null);
```

https://github.com/react-component/tree/blob/63644b4edd0386cd3d3153837a2998c989dbf47c/src/utils/treeUtil.ts#L225-L268

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2022",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": false,
        "booleans": false,
        "booleans_as_integers": false,
        "collapse_vars": false,
        "comparisons": false,
        "computed_props": false,
        "conditionals": false,
        "dead_code": false,
        "directives": false,
        "drop_console": false,
        "drop_debugger": false,
        "evaluate": false,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": false,
        "hoist_vars": false,
        "if_return": false,
        "join_vars": false,
        "keep_classnames": false,
        "keep_fargs": false,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": false,
[...]

Issue #5682, opened 2022-08-31
Minifier should drop function identifiers while inlining


kdy1:
### Describe the feature

<img width="1267" alt="image" src="https://user-images.githubusercontent.com/29931815/187591855-15008fe9-c987-449d-9fae-f62651991884.png">

swc vs terser

Austaras:
Where does this come from? I guess that's because swc currently need an extra pass to strip identifier.

kdy1:
https://github.com/vercel/ms/blob/78ce59eab01c197b9133bf9752ae01fbc11f4976/src/index.ts

I think this is the source.

Yes, it's an issue of `passes` but this affects real-world code

PR #5852, opened 2022-09-14
fix(es/minifier): Don't replace parameters of IIFE if it's recursive

kdy1:
**Description:**

While investigating, I found that problem is not analysis of recursive functions. Instead, it was a problem of IIFE evaluator, which replaces parameters even if it's recursive.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/5846


kdy1 (commented):
reset ci

---

swc-bump:
 - swc_ecma_minifier

Austaras:
I'm wondering if it's worth it to add a field in analyzer like `used_recursively`, as #5682 can also benefit from it.

kdy1:
@Austaras What's the purpose of the field? I think adding a field is fine in general

Austaras:
If an ident is `used_recursively`, it

1. cannot be replaced parameters with
2. cannot be drop when inline function

kdy1:
I think it will worth

PR #5863, opened 2022-09-14
refactor: Refactor code using clippy

kdy1:
**Description:**

I temporarily updated rustc to use newer version of `cargo clippy`.


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #5798, opened 2022-09-09
refactor(swc/ecma_ast): apply rkyv-bytecheck flag

kwonoj:
**Description:**

This PR updates ast* related package's rkyv bytecheck flags.

**BREAKING CHANGE:**

**Related issue (if exists):**

PR #5866, opened 2022-09-14
refactor(swc/ecma_ast): apply rkyv-bytecheck flag

kwonoj:
**Description:**

Reattempting #5798, but need addtional investigation for the build issue.

I couldn't repro it locally yet unfortunately so hard to dig where / what's happening exactly. When I tried, only one lint error (which seems harmless) were appear.

**BREAKING CHANGE:**


kwonoj:
The steps I tried repro was

- Checkout PR
- run patch-section, apply path-based dep to `bindings/node`'s Cargo.tml
- `npm run build`

Something I miss or did incorrectly to repro? @kdy1

kdy1:
I added required patch to reproduce

kwonoj:
> I added required patch to reproduce

Oh wow thanks! 🏃 :digging:

kdy1:
Oh I can't reproduce

kwonoj:
:oh-no:

kdy1 (approved):
swc-bump:
 - ast_node
 - swc_plugin_proxy

PR #5867, opened 2022-09-14
fix(html/minifier): perf and memory

alexander-akait:
**Description:**

Perf and memory

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_parser --breaking

PR #5870, opened 2022-09-15
chore(ci): Reduce CI time

kdy1:


magic-akari:
Is it time to use yarn@3 ?

kdy1:
Not sure. I'm doing this mainly because some crates like `swc` uses way too much time

kdy1 (commented):
swc-bump:
 - dbg-swc

PR #5873, opened 2022-09-15
perf(es/parser): Reduce memory operations

kdy1:
**Description:**

We don't need to store `Atom` in the `Token` for most cases. Because of the `Atom`s, the `Token` type is currently 48 bytes, but it's simply a waste. I drastically reduced the size of the `Token` type.

---

I'll post the performance difference after matching the current behavior.

---

cc @dsherret as this PR will change `Token` or lexer API a lot.


kdy1:
Closing because I found that I can't support some token-related APIs

PR #5878, opened 2022-09-15
feat(swc/core): enable bytecheck for plugin

kwonoj:
**Description:**

Enable internal flag for the plugin transform with bytecheck.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_plugin_proxy

PR #5882, opened 2022-09-16
feat(node/comment): expose `CommentMap`

h-a-n-a:
**Description:**

Expose `CommentMap`

**BREAKING CHANGE:**


kwonoj:
Is there some background detail where it's being used?

h-a-n-a:
> Is there some background detail where it's being used?

Looks like the `Comments` trait is exposed. I will use this. Thank you for quick review

PR #5885, opened 2022-09-16
feat(swc): enable plugin for bytecheck flags

kwonoj:
**Description:**

Fixes oversight to enabling plugin when plugin-bytecheck is enabled.


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc

PR #5892, opened 2022-09-18
test(es/plugin): Add a noop plugin

kdy1:
**Description:**

The plugin is used for benchmark and verification of rkyv.


kdy1 (commented):
swc-bump:
 - dbg-swc

Issue #5889, opened 2022-09-17
[css-minifier] Correctly parsed and compressed `<keyframes-name> = <string>`


yisibl:
## Input

```css
@keyframes "one two" {
  0% {
    left: 100px;
  }
}

@keyframes "none" {
  to {
    right: 100px;
  }
}

.foo {
  animation-name: "none", "unset", "one two";
}
```

## Actual  output

`@keyframes one two` is an invalid name and needs to be converted from a space to an **escape** character.

This WPT PR has more detailed test cases. https://github.com/web-platform-tests/wpt/pull/35903

```css
@keyframes one two{0%{left:100px}}@keyframes"none"{to{right:100px}}.foo{animation-name:"none","unset","one two"}
```

## Expected

```css
@keyframes one\ two{0%{left:100px}}@keyframes"none"{to{right:100px}}.foo{animation-name:"none","unset",one\ two}
```

### Version

0.125.9

yisibl:
I have added more detailed test cases in https://github.com/parcel-bundler/lightningcss/pull/295 for reference.

yisibl:
I've filed an issue in the spec about empty strings(`@keyframes "" {}` or `@keyframes " " {}`), which we can address later once the spec is clarified.

https://github.com/w3c/csswg-drafts/issues/7762

PR #5894, opened 2022-09-18
feat(css-minifier): add keyframes naming minify support

CGQAQ:
**Description:**
This PR adds support for CSS identifier escaping and minifies both `@keyframes` and `animation-name | animation`. Previously, this string type name is only supported on the firefox, but I'm currently working on blink and the webkit one is already merged, so I think it's the time that swc support it.

ref: https://github.com/WebKit/WebKit/pull/4234
https://chromium-review.googlesource.com/c/chromium/src/+/3865188

Fixes: #5889


yisibl:
Nice works!

For generating escaped characters, it seems to lose its meaning if the length of bytes after escaping is greater than before.

crates/swc_css_minifier/src/compressor/declaration.rs
```diff
@@ -422,6 +422,71 @@ impl Compressor {
                         declaration.value.remove(1);
                     }
                 }
+                js_word!("animation") if !declaration.value.is_empty() => {
+                    let first = declaration.value.remove(0);
```
kdy1, line 426: This is a slow operation.

kdy1: Can you match on first element and call remove only if its' required?

crates/swc_css_minifier/src/compressor/declaration.rs
```diff
@@ -422,6 +422,71 @@ impl Compressor {
...
+                    if let ComponentValue::Str(ident) = &first {
+                        match &*ident.value.to_ascii_lowercase() {
+                            it if crate::is_css_wide_keywords(it) => {
+                                declaration.value.insert(0, first);
```
kdy1, line 430: This is also slow

crates/swc_css_minifier/src/compressor/declaration.rs
```diff
@@ -422,6 +422,71 @@ impl Compressor {
...
+                                            span: ident.span,
+                                            value: to_be_identify.into(),
+                                            raw: None,
+                                        })
+                                    },
+                                );
+                            }
+                        }
+                    } else {
+                        declaration.value.insert(0, first);
```
kdy1, line 454: Same here

crates/swc_css_minifier/src/escape.rs
```diff
@@ -0,0 +1,100 @@
...
+        Some(escaped)
+    } else {
+        None
+    }
+}
+
+// https://github.com/servo/rust-cssparser/blob/4c5d065798ea1be649412532bde481dbd404f44a/src/serializer.rs#L193
+fn escape_ident(mut value: &str) -> String {
+    if value.is_empty() {
+        return "".to_string();
```
kdy1, line 29: Please use `String::new()` instead

crates/swc_css_minifier/src/escape.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+    let mut result = "".to_string();
```
kdy1, line 31: Same here

crates/swc_css_minifier/src/escape.rs
```diff
@@ -0,0 +1,100 @@
...
+        }
+        result += &*escape_name(value);
+    }
+
+    result
+}
+
+// https://github.com/servo/rust-cssparser/blob/4c5d065798ea1be649412532bde481dbd404f44a/src/serializer.rs#L220
+fn escape_name(value: &str) -> String {
+    let mut result = "".to_string();
```
kdy1, line 55: Same here

crates/swc_css_minifier/src/escape.rs
```diff
@@ -0,0 +1,100 @@
...
+    for (i, b) in value.bytes().enumerate() {
+        let escaped = match b {
+            b'0'..=b'9' | b'A'..=b'Z' | b'a'..=b'z' | b'_' | b'-' => continue,
+            _ if !b.is_ascii() => continue,
+            b'\0' => Some("\u{FFFD}"),
+            _ => None,
+        };
+        result += &value[chunk_start..i];
+        result += &*(if let Some(escaped) = escaped {
+            escaped.to_string()
```
kdy1, line 66: This is a waste. Please append in each branch.

crates/swc_css_minifier/src/escape.rs
```diff
@@ -0,0 +1,100 @@
...
+    let bytes = if ascii_byte > 0x0f {
+        let high = (ascii_byte >> 4) as usize;
+        let low = (ascii_byte & 0x0f) as usize;
+        b4 = [b'\\', HEX_DIGITS[high], HEX_DIGITS[low], b' '];
+        &b4[..]
+    } else {
+        b3 = [b'\\', HEX_DIGITS[ascii_byte as usize], b' '];
+        &b3[..]
+    };
+    unsafe { str::from_utf8_unchecked(bytes) }.to_string()
```
kdy1, line 93: Please add a comment about safety

crates/swc_css_minifier/src/escape.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// https://github.com/servo/rust-cssparser/blob/4c5d065798ea1be649412532bde481dbd404f44a/src/serializer.rs#L185
+fn char_escape(ascii_byte: u8) -> String {
+    let bytes = [b'\\', ascii_byte];
+    unsafe { str::from_utf8_unchecked(&bytes) }.to_string()
```
kdy1, line 99: Same here.
Like

```rust
// Safety: It's all ascii
```

crates/swc_css_minifier/src/lib.rs
```diff
@@ -10,9 +10,26 @@ use self::{compress::at_rule::compress_at_rule, compressor::compressor, options:
...
 pub mod options;
 
 pub fn minify(stylesheet: &mut Stylesheet, _options: MinifyOptions) {
     stylesheet.visit_mut_with(&mut compressor());
     stylesheet.visit_mut_with(&mut compress_at_rule());
 }
+
+#[inline]
+fn is_css_wide_keywords(ident: &str) -> bool {
+    match &*ident.to_ascii_lowercase() {
```
kdy1, line 23: I think we can use `JsWord` here

kdy1 (changes requested):
Thanks

crates/swc_css_minifier/src/lib.rs
CGQAQ: how? I don't get it

CGQAQ: Oh, I see

kdy1 (dismissed):
Thanks! Rust code change LGTM.

---

swc-bump:
 - swc_atoms

crates/swc_css_minifier/src/lib.rs
```diff
@@ -3,16 +3,34 @@
...
+
+#[inline]
+fn is_css_wide_keywords(ident: &JsWord) -> bool {
+    match ident.to_ascii_lowercase() {
+        js_word!("none")
+        // CSS Values and Units Level 3: https://drafts.csswg.org/css-values-3/#common-keywords
+        |  js_word!("initial")
+        |  js_word!("inherit")
+        |  js_word!("unset")
+        |  js_word!("default")
```
yisibl, line 30: It is recommended to add a separate `is_default_keyword` like in Blink.

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/css/properties/css_parsing_utils.cc;l=3449-3451?q=%20IsCSSWideKeyword&ss=chromium

crates/swc_css_minifier/src/lib.rs
```diff
@@ -3,16 +3,34 @@
...
 
 pub fn minify(stylesheet: &mut Stylesheet, _options: MinifyOptions) {
     stylesheet.visit_mut_with(&mut compressor());
     stylesheet.visit_mut_with(&mut compress_at_rule());
 }
+
+#[inline]
+fn is_css_wide_keywords(ident: &JsWord) -> bool {
+    match ident.to_ascii_lowercase() {
+        js_word!("none")
```
yisibl, line 25: This is not CSS-Wide keywords. Ref: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/css/properties/css_parsing_utils.cc;l=3439-3445?q=%20IsCSSWideKeyword&ss=chromium

```c++
// https://drafts.csswg.org/css-values-4/#css-wide-keywords
bool IsCSSWideKeyword(StringView keyword) {
  return EqualIgnoringASCIICase(keyword, "initial") ||
         EqualIgnoringASCIICase(keyword, "inherit") ||
         EqualIgnoringASCIICase(keyword, "unset") ||
         EqualIgnoringASCIICase(keyword, "revert") ||
         EqualIgnoringASCIICase(keyword, "revert-layer");
  // This function should match the overload before it.
}
```

alexander-akait: Yes, `none` is only for keyframes, so we should not put it here to avoid future problems when we will use `is_css_wide_keyword`

alexander-akait (dismissed):
Let's fix some notes above and we can merge

crates/swc_css_minifier/src/escape.rs
```diff
@@ -0,0 +1,106 @@
...
+
+/// Try to escape a ident to a shorter one if possible.
+/// if so return the escaped ident, otherwise return None.
+#[inline]
+pub fn try_escape_if_shorter(input: &str) -> Option<String> {
+    let escaped = escape_ident(input);
+    // escaped: without double quotes, so need plus 2 here
+    if escaped.len() < input.len() + 2 {
+        Some(escaped)
+    } else {
```
yisibl, line 21: Awesome, this feature will make escaped very useful.

crates/swc_css_minifier/src/lib.rs
CGQAQ: I'll keep it

kdy1 (approved):
Thank you!

PR #5900, opened 2022-09-19
fix(es/minifier): remove more pure calls

Austaras:
**Description:**

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_minifier

PR #5901, opened 2022-09-19
feat(es/minifier): Drop function identifier in a single pass

kdy1:
**Description:**

We only inline function declarations if `ref_count` is 1, so we don't need to preserve the name of the function.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/5682.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

Issue #5398, opened 2022-08-05
Support mixed-unit calc in media queries


yisibl:
### Describe the bug

Cannot resolve the calc() function in `@media`.

See also: https://chromium-review.googlesource.com/c/chromium/src/+/3683420

cc @g-plane 

### Input code

```css
@media ( width > calc(5px + 1rem) ) {}
@media ( width < calc(Infinity * 1px) ) {}
@media ( max-width: calc(5px + 1rem) ) {}
@media ( width: calc(5px + 1rem) ) {}
```

Maybe we should also check `@container`:

```css
@container ( width > calc(5px + 1rem) ) {}
@container ( width < calc(Infinity * 1px) ) {}
```

### Playground link

https://swc-css.netlify.app/?code=eJxzyE1NyUxU0FAoz0wpyVCwU0hOzEnWMC2oUNBWMCxKzdVU0FSoruVyQFVmA1HmmZeWmZdZUqmgpWBYUAFVCgAXkxjz

### Expected behavior

The calc() function can be parsed correctly.

### Version

swc_common = "0.26"
swc_css_ast = "0.104"
swc_css_parser = "0.113"

g-plane:
SWC CSS parser doesn't support `@container` currently. It should be supported in the future, but this is out of this topic.

alexander-akait:
I will fix it tomorrow

yisibl:
@alexander-akait Hi, Do you have time to support this yet?

alexander-akait:
Yes, today, I will do it (it was already in my todo list on this week :smile: )

PR #5904, opened 2022-09-19
fix(css/parser): math function in `@media`

alexander-akait:
**Description:**

Parse `calc` function in media queries, we support more function and not only calc because new spec contain them more (chrome and firefox don't support it currently, only `calc`)

**BREAKING CHANGE:**

No

fixes https://github.com/swc-project/swc/issues/5398


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_ast --breaking

PR #5906, opened 2022-09-20
perf(es/minifier): Make more passes parallel

kdy1:
**Description:**

We only unwrap parenthesis in precompress optimizer, so I refactored it. Also, as it's embarrassingly parallel, I parallelized it.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5908, opened 2022-09-20
perf(es/minifier): Optimize analyzer

kdy1:
**Description:**

 - Use faster hash algorithm.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

Austaras:
What does `used_by_nested_fn` do?

kdy1:
It's used to prevent inlining in some cases. `used_by_nested_fn` is required to fill `is_fn_local`.

I moved the field because it's not accessed after analysis.

Issue #5918, opened 2022-09-21
@import layer function with invalid(not ident) arguments will cause parser infinite loop


CGQAQ:
### Describe the bug

@import layer function with invalid(not ident) arguments will cause parser infinite loop

```css
@import "foo.css" layer("invalid");
```

### Expected behavior

don't infinite loop

### Version

1.3.2

PR #5919, opened 2022-09-21
fix: fix-layer-infinite-loop

CGQAQ:
**Description:**

**BREAKING CHANGE:**

#5918


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_parser

PR #5920, opened 2022-09-21
Add `common_tty` feature to swc_core

shuding:
**Description:**

This PR adds a `common_tty` feature to swc_core which enables `tty-emitter` for swc_common to make colored terminal output available.


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_core

PR #5930, opened 2022-09-22
fix(es/minifier): Prepend vars when dropping vars in a for loop initializer

kdy1:
**Description:**

The minifier appends initializer of for statements instead of prepending it.

**Related issue:**

 - https://github.com/vercel/next.js/discussions/30237#discussioncomment-3704795

---

```js
NodeSelection.prototype.getIndex = function(node) {
    var index;
    var num = 0;
    node = !node.previousSibling && node.tagName === 'BR' ? node : node.previousSibling;
    if (node) {
        for (var type = node.nodeType; node; null) {
            index = node.nodeType;
            num++;
            //eslint-disable-next-line
            type = index;
            node = node.previousSibling;
        }
    }
    return num;
};
```

is mis-optimized as

```js
NodeSelection.prototype.getIndex = function(node) {
    var num = 0;
    if (node = node.previousSibling || 'BR' !== node.tagName ? node.previousSibling : node) {
        for (; node;) node.nodeType, num++, node = node.previousSibling;
        node.nodeType;
    }
    return num;
}

```

and this is wrong because of the order of `node.nodeType`


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5931, opened 2022-09-22
feat(es/minifier): Improve analysis of reassignments

kdy1:
**Description:**

This PR improves the analysis of reassignments. We don't need to mark a variable as reassigned if it's not initialized already.

**Related issue:**


crates/swc_ecma_minifier/tests/fixture/issues/5910/1/output.js
```diff
@@ -1,7 +1,6 @@
 export function fn1() {
     let walkingIndex = 0;
     return function() {
-        const myIndex = walkingIndex;
-        walkingIndex += 1, console.log(myIndex, walkingIndex);
+        walkingIndex += 1, console.log(walkingIndex, walkingIndex);
```
kdy1, line 4: This is a bug

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5932, opened 2022-09-22
feat(es/minifier): Use char frequency data for property manglers

kdy1:


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5937, opened 2022-09-23
feat(es/minifier): Drop unused properties

kdy1:
**Description:**

We can drop some properties


kdy1:

crates/swc_ecma_minifier/src/compress/optimize/unused.rs
```diff
@@ -244,17 +244,116 @@ where
...
                         }
                     }
                 }
             }
 
+            if self.options.side_effects
+                && !v.used_as_arg
+                && !v.used_as_callee
+                && !v.indexed_with_dynamic_key
+                && !v.is_infected()
```
jridgewell, line 264: Does this capture escape analysis with `return statements`, assignments to hoisted variables, etc?

kdy1: No, currently the analyzer doesn't analyze if a variable escapes

PR #5940, opened 2022-09-23
fix(es/minifier): Track aliasing via assignments

kdy1:


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5954, opened 2022-09-26
fix(es/minifier): Fix analysis of parameters

kdy1:
**Description:**

Although the author of the issue talked about `@react-pdf/renderer`, it's a bug related to `brotli`, not react pdf renderer.

After investigation, I found that the bug is caused by not marking parameters as initialized. So I fixed the analyzer.

**Related issue:**

 - https://github.com/vercel/next.js/issues/40803

---

# Investigation

This occurs only if `inline` is `true`.

```rust
                            defaults: false,
                            inline: TerserInlineOption::Bool(true).into(),
                            ..Default::default()
                        }
```

breaks

# More investigation

image

`start_table` should not be replaced with `table` because `table` is modified.

image


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5955, opened 2022-09-26
mod(es/minifier): cache recursive use

Austaras:
**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_minifier/src/analyzer/storage/normal.rs
+++ b/crates/swc_ecma_minifier/src/analyzer/storage/normal.rs
@@ -85,9 +85,10 @@ impl Storage for ProgramData {
 
                     e.get_mut().pure_fn |= var_info.pure_fn;
 
-                    if !var_info.is_fn_local {
-                        e.get_mut().is_fn_local = false;
-                    }
+                    e.get_mut().used_recursively |= var_info.used_recursively;
+
+                    e.get_mut().is_fn_local &= var_info.is_fn_local;
+                    e.get_mut().used_in_non_child_fn |= var_info.used_in_non_child_fn;
 
                     if var_info.var_initialized {
                         if e.get().var_initialized || e.get().ref_count > 0 {
@@ -99,12 +100,14 @@ impl Storage for ProgramData {
                     match kind {
                         ScopeKind::Fn => {
                             e.get_mut().is_fn_local = false;
-                            e.get_mut().used_by_nested_fn = true;
+                            if !var_info.used_recursively {
+                                e.get_mut().used_in_non_child_fn = true
+                            }
                         }
                         ScopeKind::Block => {
-                            if var_info.used_by_nested_fn {
+                            if var_info.used_in_non_child_fn {
                                 e.get_mut().is_fn_local = false;
-                                e.get_mut().used_by_nested_fn = true;
+                                e.get_mut().used_in_non_child_fn = true;
                             }
                         }
                     }
@@ -112,7 +115,9 @@ impl Storage for ProgramData {
                 Entry::Vacant(e) => {
                     match kind {
                         ScopeKind::Fn => {
-                            var_info.used_by_nested_fn = true;
+                            if !var_info.used_recursively {
+                                var_info.used_in_non_child_fn = true
+                            }
                         }
                         ScopeKind::Block => {}
                     }
@@ -149,7 +154,7 @@ impl Storage for ProgramData {
                     v.assign_count += 1;
                 }
 
-                if v.used_by_nested_fn {
+                if v.used_in_non_child_fn {
                     v.is_fn_local = false;
                 }
             })
@@ -337,5 +342,9 @@ impl VarDataLike for VarUsageInfo {
 
     fn mark_used_above_decl(&mut self) {
         self.used_above_decl = true;
+    }
+
+    fn mark_used_recursively(&mut self) {
+        self.used_recursively = true;
     }
 }
--- a/crates/swc_ecma_minifier/src/analyzer/mod.rs
+++ b/crates/swc_ecma_minifier/src/analyzer/mod.rs
@@ -135,9 +135,10 @@ pub(crate) struct VarUsageInfo {
     /// PR. (because it's hard to review)
     infects: Vec<Id>,
 
-    used_by_nested_fn: bool,
-
+    pub used_in_non_child_fn: bool,
     pub accessed_props: Box<AHashMap<JsWord, u32>>,
+
+    pub used_recursively: bool,
 }
 
 impl VarUsageInfo {
@@ -363,6 +364,7 @@ where
                     self.data.report_usage(self.ctx, i, is_assign);
                     self.data.var_or_default(i.to_id()).mark_used_above_decl()
                 }
+                self.data.var_or_default(i.to_id()).mark_used_recursively();
                 return;
             }
         }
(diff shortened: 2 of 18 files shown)


crates/swc_ecma_minifier/src/analyzer/storage/normal.rs
```diff
@@ -98,8 +101,10 @@ impl Storage for ProgramData {
 
                     match kind {
                         ScopeKind::Fn => {
-                            e.get_mut().is_fn_local = false;
-                            e.get_mut().used_by_nested_fn = true;
+                            if !var_info.used_recursively {
```
kdy1, line 104: I'm not sure if this if statement is correct

kdy1 (commented):
I'm just not sure if it's correct

---

swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/src/analyzer/storage/normal.rs
Austaras: The general idea is for following code:
```js
function a() {
  x(a)
}
module.export = a
```
`a` needs to be `fn_local` to be inlined. But I'm also not sure if this is right and I believe we should be extremely cautious for this.

kdy1:
Can you split this PR into two PRs?
One for resolver fix, and one for minifier improvement

Austaras:
Fine. But there would be some regressions.

kdy1:
Ah, I see. It's fine and thank you!

crates/swc_ecma_minifier/src/analyzer/storage/normal.rs
```diff
@@ -98,8 +101,10 @@ impl Storage for ProgramData {
...
+                                e.get_mut().is_fn_local = false;
```
kdy1, line 105: I've been thinking about this for a while

kdy1: I found that this is problematic for some cases.


In the code below, 

```js

var foo;
function init() {
    function bar() {
        foo = bar;
    }

    escape(bar)
}

```

`bar` is not fn-local but it wil be marked as fn-local

kdy1: I expect something similar to this pattern to exist in real-world apps, so I think we should fix it

kdy1: I think tracking the count of non-fn-local usage can be a way

kdy1: But for 

```js
function a() {
  x(a)
}
module.export = a
```

I think it's better to extend sequential inliner.

kdy1 (changes requested):
Thanks!
If you think my opinion is wrong, feel free to tell me

crates/swc_ecma_minifier/src/analyzer/storage/normal.rs
Austaras: Oh, you're right. Let me check how terser deal with this when I have time.

Austaras: Your example is wrong on [playground](https://play.swc.rs/?version=1.3.3&code=H4sIAAAAAAAAA0utKMgvKlEoSyxSSMvPt%2BZKhfDTSvOSSzLz8xQy8zJLNDQVqrngIkmJRWABBSAAalGwBYlYg7nJ%2BXnF%2BTmpejn56RqGhoaaXLVgYbAOIBsARRRWE2wAAAA%3D&config=H4sIAAAAAAAAA21TQZLjIAz8i8972MpxHrC3fQNFQDiaxcgliUxcU%2Fn7iCTGcbI31O6WRav5Hj4lDB%2Ffw%2BxZgNtJlqL%2BMnwMuswggXHW4degYlDyWeBqBfsiiXhqfAYf9HaoRXECU%2FqqNHnFMFyNnYkEHmLTeh7B%2BAPI4ffhYK0nLJiW1iHQNDOItLPR6gRFpSs9M31ZqVytOhJlsDFea%2BfFYVEYgTdpoJz9LODOnrui%2Fc0zCpUdVBWim5nmJ7REVKRizVYsgo8uUIQOIENQPMNGsR5GKWKD9UluYIRjHcdm950JZ5%2Br194LLjcb7I9ddyIUdakWeYF2k96h%2ByUfLEyOQSuXlfNJWHY%2B%2FAOwObMXKX6CTXnDk%2B1hz0z%2FYWFJtkRdOmo736YqMNrlHGLqN2xTAytubjHEGqDdMGzNH%2BDuQoIRHKRkbnexfKGG09asRZdSL801n7Yd3Eu35mmPtgi8gX9sKn1exwO3kJ9eMVmmI%2BW3FhPoieIbbEYovYJs6b3Mr2gtEcxmiE8fqrSyx9SWr%2BRye1%2BrzxYI07gx07HH99qe5frs1pzIX4o1PzZ0%2FQEAk%2FXaGQQAAA%3D%3D)-- let me try to fix it first.

Austaras: Current behavior is aligned with terser.
```js
export var foo;
export function init() {
    function bar() {
        foo = bar;
        console.log(111);
    }
    return bar;
}
```
would inline, while
```
export var foo;

function bar() {
    foo = bar;
    console.log(111);
}
export function init() {
    return bar;
}
```
would not. I believe this is fine.

kdy1: Can you add a test for

```js

export function foo() {
    function bar() {
        function baz() {
            return bar;
        }
        return bar && baz;
    }

    return bar;
}

```

?
I think in this case `bar` should not be `fn-local`, but it will not be marked as non-fn-local because of recursion

Austaras: In this case I see no reason why `bar` should not be inlined.

kdy1: It's fine if fn-local is used only for inlining, but not sure if it's the case

kdy1: I checked the code and `fn-local` is used by much more places. So I think we should analyze it correctly, or use another field like `fn_local_for_inlining` just for inlining

crates/swc_ecma_minifier/src/analyzer/storage/normal.rs
```diff
@@ -85,9 +85,9 @@ impl Storage for ProgramData {
 
                     e.get_mut().pure_fn |= var_info.pure_fn;
 
-                    if !var_info.is_fn_local {
-                        e.get_mut().is_fn_local = false;
-                    }
+                    e.get_mut().used_recursively |= var_info.used_recursively;
+
+                    e.get_mut().is_fn_local &= var_info.is_fn_local;
```
kdy1, line 90: I think we need `e.get_mut().used_in_non_child_fn &= var_info.used_in_non_child_fn;`

kdy1 (changes requested):
Thank you!

kdy1 (dismissed):
Thank you!

---

swc-bump:
 - swc_ecma_minifier

kdy1 (approved):

Issue #5935, opened 2022-09-22
Next.js Chart.js broken when `swcMinify` enabled for non-development builds


grempe:
### Describe the bug

This bug was originally filed here where more details can be found. The issue has been narrowed down to seemingly broken output from swc.

https://github.com/chartjs/Chart.js/issues/10673

https://github.com/reactchartjs/react-chartjs-2/issues/1071

With NextJS, the library works well on development. But an error is thrown after running the app after building it.
`RangeError: minimumFractionDigits is out of range`.

If the next.js config is changed to `swcMinify: false` and a production build is done the issue goes away.

It seems that during production this line:

https://github.com/chartjs/Chart.js/blob/d4e106cc9cf19a273e5485ffaf6158f663e65717/src/helpers/helpers.intl.js#L4

The minimumFractionDigits and maximumFractionDigits in the options become NaN instead of 0.

There is sample code here created by the original filer of the bug:

https://github.com/LouisJackson/react-chart-test/blob/main/next.config.js

### Playground link

https://github.com/LouisJackson/react-chart-test/blob/main/next.config.js

### Expected behavior

React should not die when a page with the Chart compiled with swc is accessed.

### Version

"dependencies": {     "chart.js": "^3.9.1",     "next": "12.3.0",     "react": "18.2.0",     "react-chartjs-2": "^4.3.1",     "react-dom": "18.2.0"   },

siinghd:
Would like to add that this started happening to me from version 12.3.

alexphl:
Can confirm that the last release of Next where this issue doesn't occur is 12.2.5. 
Breaking change must have happened between that and 12.3.

PR #5956, opened 2022-09-26
fix(es/minifier): Fix skipping logic of sequential inliner

kdy1:
**Description:**

This PR implements `is_skippable_for_seq` fully.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/5935

---

# Investigation

`collapse_vars: true` is the condition

```rust
                            defaults: false,
                            collapse_vars: true.into(),
                            ..Default::default()
                        }

```

this breaks

---

`factor` should be inlined to the first usage, but it's instead inlined into second usage.

image


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5959, opened 2022-09-27
fix(es/base): Use seperate mark for FnExpr Ident

Austaras:
**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_minifier/tests/terser/compress/collapse_vars/collapse_rhs_lhs_2/output.js
+++ b/crates/swc_ecma_minifier/tests/terser/compress/collapse_vars/collapse_rhs_lhs_2/output.js
@@ -1,6 +1,5 @@
 var b = 1;
-(function f(f) {
-    f = b;
-    f[b] = 0;
+(function f(f1) {
+    (f1 = b)[b] = 0;
 })();
 console.log("PASS");
--- a/crates/swc_ecma_minifier/tests/fixture/issues/quagga2/1.4.2/1/output.mangleOnly.js
+++ b/crates/swc_ecma_minifier/tests/fixture/issues/quagga2/1.4.2/1/output.mangleOnly.js
@@ -276,7 +276,7 @@
                         a.rad = e / n.length;
                         a.vec = o.clone([
                             Math.cos(a.rad),
-                            Math.sin(a.rad), 
+                            Math.sin(a.rad)
                         ]);
                     }
                     function f() {
@@ -578,7 +578,7 @@
                     var i = [];
                     var u = 0;
                     var c = 0;
-                    function t(t, n) {
+                    function f(t, n) {
                         var a;
                         var o;
                         var i;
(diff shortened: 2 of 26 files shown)


crates/swc_ecma_minifier/tests/terser/compress/collapse_vars/collapse_rhs_lhs_2/output.js
```diff
@@ -1,6 +1,5 @@
 var b = 1;
-(function f(f) {
-    f = b;
-    f[b] = 0;
+(function f(f1) {
```
Austaras, line 2: Don't know why this is renamed -- Could you help @kdy1 ?

kdy1 (dismissed):
Thanks, although I think you need to update more tests.

---

swc-bump:
 - swc_ecma_transforms_base

crates/swc_ecma_minifier/tests/terser/compress/collapse_vars/collapse_rhs_lhs_2/output.js
kdy1: Because function name (`f`) and the parameter (`f1`) is now in a different scope

kdy1: Previously those were treated as identical to each other

kdy1: As there's `(` in front of function, this is fn expr, not fn decl

Austaras: But why do other colliding ident not got renamed?

kdy1: At where? We reuse variable names in the name mangler, and we only rename declared identifiers in normal var renamer

Austaras:
```js
function foo(foo) {}
```
would remain the same after renamer.

kdy1: We are using same analyer for minifier and normal var renamer, and the analyzer knows that you can't refer `foo` (function) from in `foo`

kdy1: Hmm... you are right I'm not sure why is this renamed.

kdy1: I think https://github.com/swc-project/swc/blob/3d393dd709f9a41dc53b15c0fc1edb93997f9d5b/crates/swc_ecma_transforms_base/src/rename/analyzer/mod.rs#L155-L171 is the issue

kdy1 (approved):
I think there are room for improvement, but let's split PR.

PR #5960, opened 2022-09-27
test(es/minifier): Categorize terser tests

kdy1:
**Description:**

I'm going to focus on rules enabled by default. I updated some tests because our one was better, and postponed `unsafe` tests.


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #5961, opened 2022-09-27
feat(es/minifier): Ignore more expressions in sequential inliner

kdy1:
**Description:**

We can skip **member access** to `console` as it does not have any side effect.

Ref: https://developer.mozilla.org/en-US/docs/Web/API/console


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5963, opened 2022-09-27
fix(es/minifier): Fix analysis of nested function-like properties

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/5950.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

Issue #5189, opened 2022-07-12
Decorators for symbol property are compiled incorrectly


dko-slapdash:
### Describe the bug

When using a @Decorator for a symbol property, the compiler generates a wrong output.

### Input code

```typescript
const sym = Symbol("sym");

class Cls {
  @Memoize()
  [sym]() {}
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "decorators": true
    },
    "target": "es2020"
  },
  "module": {
    "type": "commonjs"
  }
}
```

### Playground link

https://play.swc.rs/?version=1.2.212&code=H4sIAAAAAAAAA0vOzysuUSiuzFWwVQiuzE3Kz9FQAvKUNK25uJJzEouLFZxzihWquRQUHHxTc%2FMzq1I1NIGcaKCaWA1NheparloAlz70DEMAAAA%3D&config=H4sIAAAAAAAAAxWMQQrAIAwE%2F7JnD8WjvxENpaU2kkSoiH%2Bv3padnR24NSEM1ChKspP21%2BKHAOuVNMlVDQ6ZEks0FkUwaTQdLMpJtoak%2FvAHVlU4t4f2y5YXSlwKv7dizh8GxEdwawAAAA%3D%3D

### Expected behavior

It should define _key before defining the class.

### Actual behavior

_key is undefined, so the property does not define. Compiled result:

```
...
const sym = Symbol("sym");
var _key;
class Cls {
    [_key]() {}
}
_key = sym; // <--- This must be BEFORE the class definition
__decorate([
    Memoize()
], Cls.prototype, _key, null);
```

### Version

1.2.212

CMCDragonkai:
I think we saw this as well.

PR #5964, opened 2022-09-27
fix(es/compat): Fix the order of initialization for decorators on computed keys

kdy1:
**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/5189.


kdy1 (commented):
swc-bump:

Issue #5951, opened 2022-09-26
asyncToGenerator function unrolls secondary parameter if it is an array


finnboeger:
Calling an async function with the result of two async functions that both return an array leads to the second array being unrolled and each member of it being passed as a seperate parameter to the function.

With `a` and `b` being
```typescript
async function a(): Promise<string[]> {
    return [1,2];
}

async function b(x: string[], y: string[]): Promise<void> {
    console.log(x, y);
    console.log(...x, ...y);
}
```

the following code is being transpiled incorrectly:
```typescript
(
    async () => b(await a(), await a())
)().then();
```

into

```javascript
_asyncToGenerator(function() {
    var _tmp;
    return __generator(this, function(_state) {
        switch(_state.label){
            case 0:
                return [
                    4,
                    a()
                ];
            case 1:
                _tmp = [
                    _state.sent()
                ];
                return [
                    4,
                    a()
                ];
            case 2:
                return [
                    2,
                    b.apply(void 0, _tmp.concat(_state.sent()))
                ];
        }
    });
})().then();
```

whereas the line `b.apply(void 0, _tmp.concat(_state.sent()))` should be instead `b.apply(void 0, _tmp.push(_state.sent()))`

The issue is reproduced here: https://github.com/finnboeger/swc-loader-async-bug

PR #5965, opened 2022-09-27
fix(es/compat): counteract unrolling of awaited async arrays in generator

finnboeger:
**Description:**
This is a workaround addressing the issue #5951 by wrapping the arguments of the `.concat` call in individual arrays to counteract the unrolling done by `concat`. 

A different way of handling the problem would be to `push` the arguments to `_tmp` before the return statement and then use `_tmp` in the `apply` call. However I'm not comfortable enough with the codebase to implement it that way.

**BREAKING CHANGE:**
None afaik, all tests in [crates/swc](https://github.com/swc-project/swc/tree/main/crates/swc) complete successfully and anything that depended on the broken implementation would not have worked with a different transformer.

 - Closes https://github.com/swc-project/swc/issues/5951.


kdy1 (changes requested):
Thanks!

crates/swc_ecma_parser/tests/tsc/awaitCallExpression9_es5.ts
```diff
@@ -0,0 +1,16 @@
+// @target: ES5
```
kdy1, line 1: Why did you add this file? Please remove

finnboeger: To explicitly test the behaviour for arrays, however the current solution doesn't really require it

kdy1 (approved):
Thank you!

---


swc-bump:
 - swc_ecma_transforms_compat

Issue #5860, opened 2022-09-14
[css-minifier]expand the static `calc()`


yisibl:
### Describe the feature

Often the expressions in calc() can be expanded further in order to reduce the number of bytes.

There is a lot of this CSS in [`animate.css`](https://github.com/animate-css/animate.css/blob/3235f27325ebc721fc4f8435d3c2d436642278bc/animate.css#L51-L56):

`animation-delay: calc(1s * 2) = animation-delay: 2s`

```css
.animate__animated.animate__delay-2s {
  -webkit-animation-delay: calc(1s * 2);
  animation-delay: calc(1s * 2);
  -webkit-animation-delay: calc(var(--animate-delay) * 2);
  animation-delay: calc(var(--animate-delay) * 2);
}
```

But if there are percentages in `calc()`, we have to treat them differently and not expand them. see https://github.com/evanw/esbuild/issues/1821

yisibl:
https://cssnano.co/docs/optimisations/calc/

kdy1:
Closing as not planned.

Issue #4946, opened 2022-06-11
Use better string type


kdy1:
Using `JsWord` everywhere was my mistake. We should use `Arc<str>` instead of `JsWord` for some types.
But I'll use a wrapper named `Atom`

# Tasks

 - [x] Add `Atom` and `AtomGenerator`. (https://github.com/swc-project/swc/pull/4945)

 - [ ] Improve `Atom` (https://github.com/swc-project/swc/pull/5326)

Atoms should be (conceptually)

```rust
enum Atom {
  Inlined([u8; 7])
  Static(&'static str),
  Arc(Arc<String>)
}
```

but with pointer tagging.
 
 - [ ] Add thread-local interner

 - [ ] Use it in es ast

 - [ ] Use it in css ast

 - [x] Use it in html ast

overlookmotel:
@kdy1 While working on #2175, I've noticed that serializing strings (`JsWord`s) is one of the slowest parts of the process. I've been working on an alternative serializer for `JsWord` to speed it up, but I've realized that may be wasted effort if plan is to replace `JsWord` with `Atom` anyway.

Can I ask a few questions:

1. Why was using `JsWord` a mistake?
2. You say above that using `Arc<str>` should be used for "some" types. What types should continue to use `JsWord` and not the new `Atom` struct?
3. Is the implementation of `Atom` complete or does it need more work before it can be used?
4. Is what's required to complete the transition simply replacing `JsWord` everywhere with `Atom`?

Please let me know if you have time. I may be able to make a PR.

kdy1:
1. It's useful in only some cases
2. I think we can use `Atom` for all cases.
Problem is `Str`... Not sure about it
3. It's complete
4. Parser should be patched to use atom generator

overlookmotel:
Thanks for swift reply. Sorry for more questions, but:

1. Can you expand on "It's useful in only some cases"? In which cases is it not useful?

2. Will `Atom` be faster than `JsWord` in all cases?

3. Presumably the usage of `Atom::new_bad` should be replaced with `AtomGenerator::gen` in these places?

https://github.com/swc-project/swc/blob/aef6de4647d2815819ba38c234c17694f83845d5/crates/swc_atoms/src/lib.rs#L151
https://github.com/swc-project/swc/blob/aef6de4647d2815819ba38c234c17694f83845d5/crates/swc_atoms/src/lib.rs#L159-L160

[...]

PR #5966, opened 2022-09-27
feat(css/minifier) Static computation for 'calc' expressions

ghostd:
**Description:**

This is a first step to fold the "calc" expressions:
* division operator is not (yet) included: i don't know if we want to transform "1/3" into "0,33333"
* other optimizations could follow (operation with infinity, NaN, etc)

**BREAKING CHANGE:**

None except the CSS output

#5860


ghostd:
I needed to add some "copy_" functions because the AST enums does not have the "Copy" trait ; i also tried to add it, but i would need to change JsWord... i guess this limitation will be gone with a fix for #4946

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,846 @@
...
+                }
+            }
+            _ => None,
+        };
+    }
+
+    None
+}
+
+fn copy_ident(i: &Ident) -> Ident {
```
kdy1, line 50: You can do `i.clone()` instead

kdy1: And same for others, too

kdy1 (changes requested):
Thanks!

crates/swc_css_minifier/src/compressor/calc_sum.rs
ghostd: Fixed

alexander-akait (changes requested):
Please move **ALL** tests from https://github.com/postcss/postcss-calc/blob/master/test/index.js

I think it can be simplifed, because we have a spec how to calculate it https://www.w3.org/TR/css-values-3/#calc-type-checking

Also current solution only for `calc`, but in spec we have a general solution for all math functions, we should adopt it (AST is almost the same for this functions), but we can do it in another PR

alexander-akait:
Anyway let's finish calc firstly (all tests to check everything works fine) and then I will help you to finish other things

ghostd:
@alexander-akait i copied most of the tests from postcss-calc except these ones (which depend on an postcss configuration or an exception):

```js
test(
  'should handle precision correctly (3)',
  testValue('calc(5/1000000)', '0.000005', { precision: 6 })
);

test(
  'should preserve the original declaration when `preserve` option is set to true',
  testCss('foo{bar:calc(1rem * 1.5)}', 'foo{bar:1.5rem;bar:calc(1rem * 1.5)}', {
    preserve: true,
  })
);

test(
  'should not yield warnings when nothing is wrong',
  testValue('calc(500px - 0px)', '500px', { warnWhenCannotResolve: true })
);

test(
  'should warn when calc expression cannot be reduced to a single value',
  testValue('calc(100% + 1px)', 'calc(100% + 1px)', {
    warnWhenCannotResolve: true,
  })
);

test(
  'should not parse variables as calc expressions (#35)',
  testCss(
    'foo:nth-child(2n + $var-calc){}',
    'foo:nth-child(2n + $var-calc){}',
    { selectors: true }
  )
);

test(
  'should return the same and not thrown an exception for attribute selectors without a value',
  testCss('button[disabled]{}', 'button[disabled]{}', { selectors: true })
);

test(
  'should reduce calc in media queries when `mediaQueries` option is set to true',
  testCss('@media (min-width:calc(10px+10px)){}', '@media (min-width:20px){}', {
    mediaQueries: true,
  })
);

test(
  'should reduce calc in selectors when `selectors` option is set to true (1)',
  testCss('div[data-size="calc(3*3)"]{}', 'div[data-size="9"]{}', {
    selectors: true,
  })
);

test(
  'should reduce calc in selectors when `selectors` option is set to true (2)',
  testCss('div:nth-child(2n + calc(3*3)){}', 'div:nth-child(2n + 9){}', {
    selectors: true,
  })
);

test(
  'should throw an exception when attempting to divide by zero',
  testThrows('calc(500px/0)', 'calc(500px/0)', 'Cannot divide by zero')
);

test(
  'should throw an exception when attempting to divide by unit (#1)',
  testThrows(
    'calc(500px/2px)',
    'calc(500px/2px)',
    'Cannot divide by "px", number expected'
  )
);

test(
'error with parsing',
  testThrows(
    'calc(10pc + unknown)',
    'calc(10pc + unknown)',
    'Lexical error on line 1: Unrecognized text.\n\n  Erroneous area:\n1: 10pc + unknown\n^.........^'
  )
);

```

Let me know if it correctly understood your purpose.

alexander-akait:
This should work (no need option, should work by default):
```
test(
  'should reduce calc in media queries when `mediaQueries` option is set to true',
  testCss('@media (min-width:calc(10px+10px)){}', '@media (min-width:20px){}', {
    mediaQueries: true,
  })
);
```

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,749 @@
...
+        }),
+        CalcValue::Sum(_) => {
+            unreachable!("CalcValue::Sum cannot be transformed into a ComponentValue")
+        }
+    }
+}
+
+fn fold_calc_sum(calc_sum: &mut CalcSum) {
+    // TODO we should sort the operands by unit to handle this:
+    // "4px + 3% + 5px + 10%" = "9px + 13%"
```
alexander-akait, line 118: Do we have tests case for this for future improvements?

crates/swc_css_minifier/tests/fixture/compress-calc/ignore-media-query/input.css
```diff
@@ -0,0 +1,2 @@
+@media (min-width:calc(10px+10px)){
+}
```
alexander-akait, line 2: To avoid this please add `color: red` for all tests where simple block is empty (maybe we should add option to disable it for better testing, but let's do it late)

crates/swc_css_minifier/tests/fixture/compress-calc/units/output.min.css
```diff
@@ -0,0 +1 @@
+.class1{width:calc(1px + 1)}.class2{width:calc(100% - 60px)}.class3{width:calc(100% - -10px)}.class4{width:950ms}.class5{width:calc(99.99%*1/1 - 0rem)}.class6{width:calc(50% + (5em + 5%))}.class7{width:calc((100px - 1em) + (-50px + 1em))}
```
alexander-akait, line 1: `width:calc(100% - -10px)` -> `width:calc(100% + 10px)`

crates/swc_css_minifier/tests/fixture/compress-calc/vendor/output.min.css
```diff
@@ -0,0 +1 @@
+.class1{width:-webkit-calc(2px)}.class2{width:-WEBKIT-CALC(2px)}.class3{width:-webkit-calc(100% - -webkit-calc(50% + 25px))}.class4{width:-WEBKIT-CALC(1px + -WEBKIT-CALC(2px/2))}.class5{width:-moz-calc(2px)}.class6{width:-webkit-calc(2px)}
```
alexander-akait, line 1: `width:-webkit-calc(100% - -webkit-calc(50% + 25px))` -> `width:-webkit-calc(50% - 25px)`

crates/swc_css_minifier/tests/fixture/compress-calc/units/output.min.css
alexander-akait: `width:calc((100px - 1em) + (-50px + 1em))` -> `width:50px`, but I see `TODO`, so we can postpone it

alexander-akait (commented):
Let's do some improvement (see comments above)

yisibl:
```js
test(
  'should throw an exception when attempting to divide by zero',
  testThrows('calc(500px/0)', 'calc(500px/0)', 'Cannot divide by zero')
);
```

I tested that this is valid and equivalent to `calc(infinity * 1px)`, but I didn't find the relevant text for the specification. 
@devSDK, can you help confirm?

DevSDK:
Oh sorry, I thought your question is about serialization.

The division of zero is allowed by the css-values-4 specification which is the new version of the spec.

Please check the spec here:

https://www.w3.org/TR/css-values-4/#calc-type-checking

>  Division by zero is possible, which introduces certain complications. [Math functions](https://www.w3.org/TR/css-values-4/#math-function) follow IEEE-754 semantics for these operations:

yisibl:
@DevSDK Thank you for your professional answer. My God, allowing division by `0` is so complicated!

crates/swc_css_minifier/src/compressor/calc_sum.rs
ghostd: Yes we hate this test: `width:calc((100px - 1em) + (-50px + 1em))`

crates/swc_css_minifier/tests/fixture/compress-calc/ignore-media-query/input.css
ghostd: Fixed

ghostd:
I implemented the support of calc with the vendor prefix, and added more unit conversions. I think we can be more clever to not lose some precision and factorize some code with the `length.rs` module.

crates/swc_atoms/words.txt
```diff
@@ -112,8 +114,6 @@ DeviceOrientationEvent
 Document
 DocumentFragment
 DocumentType
-dpi
-dppx
```
alexander-akait: Let's don't remove them

crates/swc_css_minifier/tests/fixture/compress-calc/convert-units/output.min.css
```diff
@@ -1 +1 @@
...
+.class1{width:1.0264583333333333cm}.class2{width:38.79527559055118px}.class3{width:20Q}.class4{width:111.48333333333333q}.class5{width:105.33858267716536px}.class6{width:10.026458333333334cm}.class7{width:1.0264583333333333cm}.class8{width:10.94488188976378px}.class9{width:10.025cm}.class10{width:1.025cm}.class11{width:10.009842519685039in}.class12{width:10.708661417322835pt}.class13{width:10.059055118110237pc}.class14{width:11.583333333333334q}.class15{width:401q}.class16{width:41q}.class17{width:1017q}.class18{width:15.11111111111111q}.class19{width:170.33333333333334q}.class20{width:3unknown}.class21{width:calc(1unknown + 2px)}.class22{width:calc(1px + 2unknown)}
```
alexander-akait, line 1: Sounds like we need option for this, can we implement this?

crates/swc_css_minifier/tests/fixture/compress-calc/media-query/output.min.css
```diff
@@ -0,0 +1 @@
+@media(min-width:calc(10px + 10px)){.class1{color:red}}
```
alexander-akait, line 1: Can you investigate why it was not compressed?

alexander-akait (commented):
Good job, some improvements and we can move ahead and in future improve it more

crates/swc_css_minifier/tests/fixture/compress-calc/media-query/output.min.css
ghostd: Here what i figured out: for now, the "calc folding" work with enums like CalcValue::Dimension, CalcValue::Number, and so on. In the case of the "@media" rule, the AST contains other enums ; the Function value is a vector of PreservedToken.

What should we do:
* change the parser to have a Function with a vector of CalcValue enum?
* enhance the calc computation to work with the PreservedToken enum?

crates/swc_css_minifier/tests/fixture/compress-calc/convert-units/output.min.css
ghostd: Can you precise what do you have in mind? Something like a "enableMixedUnit" option? "allowSumOfMixedUnit" ?

crates/swc_atoms/words.txt
ghostd: Re-added

kdy1:
Can you rebase?

ghostd:
I did a rework to fix some computations and allow more compression.

Do you wish i squash the commits before rebasing ?

crates/swc_css_minifier/tests/fixture/compress-calc/vendor/output.min.css
ghostd: This case is now handled.

crates/swc_css_minifier/tests/fixture/compress-calc/units/output.min.css
ghostd: This case is now handled :)

jridgewell (commented):
Careful with accepting this. esbuild had to fix a bug with imprecise calculations because even small fractions can lead to pixel inaccuracy: https://github.com/evanw/esbuild/issues/1821

ghostd:
@jridgewell for now i did not implement the division because of this potential loss of precision.

jridgewell:
It's not just division: `calc(1cm + 1px)` is computing to `1.0264583333333333cm`, which is imprecise. Matching units could be combined (and even `cm` and `mm` if you convert the output to whole `mm`), but combining others can lead to pixel difference between the browsers interpretation and Rust's.

ghostd:
Yes, i think this why @alexander-akait [suggested](https://github.com/swc-project/swc/pull/5966#discussion_r983838520) an option for this kind of transformation.

I'll look into this tomorrow or the next weekend.

ghostd:
Actually, the loss precision can happen whatever the operation is done, for example: `calc(100 + (0.5px - 0.4px - 0.1px))` is currently transformed into `calc(100 + -.000000000000000027755575615628914px)` (instead of `100`) because of the floating representation. There is no division involved, only substractions with floating numbers.

What limit should we set:
* an option to disable all arithmetic operations (and the folding would just remove the 0, unnecessary parenthesis, etc) ?
* an option to only allow operations when the number is an integer and the result is also an integer (in the case of floating point, we'd skip the operation) ?
* something smarter?

kdy1:
You don't need to squash :)

ghostd:
I added a check after each operation (add, sub, mul) to verify the result accuracy, now `100-75.37`is not reduced anymore to not lose precision.

ghostd:
I removed all inaccurate ratios, now we should not have imprecise computations anymore.

There are still a lot of improvement area (remove zeroes, transform `y /x * x` into `y`, handle infinity/nan, reduce some cos/sin/tan expressions, etc)

alexander-akait (approved):
Okay, let merge, I see these improvements in future:
1. container queries with `calc`
2. `@media`/`@support`, I need investigate why we have list of `PreservedToken` there
3. Some perf improvement, but need look after we fully finished it
4. Better compress in some cases

Anyway I think the current solution is enough to solve popular problems

alexander-akait:
@ghostd Also can you solve yarn conflicts?

ghostd:
@alexander-akait done :)

crates/swc_css_minifier/src/compressor/unit/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+    "dpcm" => &DPCM,
+    "dppw" => &DPPX,
+    "x" => &DPPX,
+};
+
+fn get_ratio<'a>(
+    ratio_maps: &'a phf::Map<&'static str, &'static RatioMap>,
+    unit1: &'a str,
+    unit2: &'a str,
+) -> Option<&'a f64> {
```
kdy1, line 123: Let's change the return type to `Option<f64>`

crates/swc_css_minifier/src/compressor/unit/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+    if unit1 == unit2 {
+        Some(&1.0)
+    } else {
+        ratio_maps
+            .get(unit1)
+            .and_then(|ratio_map| ratio_map.get(unit2))
+    }
+}
+
+pub fn get_duration_ratio<'a>(unit1: &'a str, unit2: &'a str) -> Option<&'a f64> {
```
kdy1, line 133: Same here

crates/swc_css_minifier/src/compressor/unit/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+    get_ratio(&DURATION_RATIO, unit1, unit2)
+}
+
+pub fn get_frequency_ratio<'a>(unit1: &'a str, unit2: &'a str) -> Option<&'a f64> {
```
kdy1, line 137: Same here

crates/swc_css_minifier/src/compressor/unit/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+    get_ratio(&FREQUENCY_RATIO, unit1, unit2)
+}
+
+pub fn get_absolute_length_ratio<'a>(unit1: &'a str, unit2: &'a str) -> Option<&'a f64> {
```
kdy1, line 141: Same here

crates/swc_css_minifier/src/compressor/unit/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+    get_ratio(&ABSOLUTE_LENGTH_RATIO, unit1, unit2)
+}
+
+pub fn get_resolution_ratio<'a>(unit1: &'a str, unit2: &'a str) -> Option<&'a f64> {
```
kdy1, line 145: Same here

kdy1 (changes requested):

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+    duration: Option<IndexedOperatorAndOperand<Time>>,
+    frequency: Option<IndexedOperatorAndOperand<Frequency>>,
+    resolution: Option<IndexedOperatorAndOperand<Resolution>>,
+    flex: Option<IndexedOperatorAndOperand<Flex>>,
+    unknown_dimension: HashMap<String, IndexedOperatorAndOperand<UnknownDimension>>,
+    expressions: Vec<CalcProductOrOperator>,
+}
+
+impl CalcSumContext {
+    pub fn new() -> Self {
```
kdy1, line 170: I think we can derive `Default` instead

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+        }
+    }
+
+    pub fn fold(&mut self, calc_sum: &mut CalcSum) {
+        self.nested_fold(&None, calc_sum);
+        calc_sum.expressions = self.expressions.to_vec();
+        remove_unnecessary_nesting_from_calc_sum(calc_sum);
+    }
+
+    fn nested_fold(&mut self, surrounding_operator: &Option<CalcOperator>, calc_sum: &mut CalcSum) {
```
kdy1, line 192: Please use `Option<&CalcOperator>` instead

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+            };
+
+            if let Some(CalcProductOrOperator::Operator(op)) = expr_it.next() {
+                operator = CalcSumContext::merge_operators(surrounding_operator, &Some(op.clone()));
+            }
+        }
+    }
+
+    fn merge_operators(
+        surrounding_operator: &Option<CalcOperator>,
```
kdy1, line 216: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+        nested_operator: &Option<CalcOperator>,
```
kdy1, line 217: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+                    } else {
+                        CalcOperatorType::Sub
+                    },
+                }),
+                _ => unreachable!("Operator value can only be Add or Sub"),
+            },
+        }
+    }
+
+    fn reduce(&mut self, operator: &Option<CalcOperator>, operand: &CalcProduct) {
```
kdy1, line 243: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+                    self.push(operator, operand);
+                }
+            }
+        } else {
+            // The calc product is not "simple", just push the data
+            self.push(operator, operand);
+        }
+    }
+
+    fn push(&mut self, operator: &Option<CalcOperator>, operand: &CalcProduct) {
```
kdy1, line 290: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+        if let Some(op) = operator {
+            self.expressions
+                .push(CalcProductOrOperator::Operator(op.clone()));
+        }
+
+        self.expressions
+            .push(CalcProductOrOperator::Product(operand.clone()));
+    }
+
+    fn sum_number(&mut self, operator: &Option<CalcOperator>, operand: &CalcProduct, n: &Number) {
```
kdy1, line 300: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+                    self.push(operator, operand);
+                }
+            }
+            None => self.number = Some(self.new_indexed_data(operator, operand, n.clone())),
+        }
+    }
+
+    fn sum_percentage(
+        &mut self,
+        operator: &Option<CalcOperator>,
```
kdy1, line 335: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+                    );
+                } else {
+                    self.push(operator, operand);
+                }
+            }
+            None => self.percentage = Some(self.new_indexed_data(operator, operand, p.clone())),
+        }
+    }
+
+    fn sum_length(&mut self, operator: &Option<CalcOperator>, operand: &CalcProduct, l: &Length) {
```
kdy1, line 371: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+        if is_absolute_length(&unit) {
+            self.sum_absolute_length(operator, operand, l)
+        } else {
+            self.sum_other_length(operator, operand, l)
+        }
+    }
+
+    fn sum_absolute_length(
+        &mut self,
+        operator: &Option<CalcOperator>,
```
kdy1, line 382: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+            }
+            None => {
+                self.absolute_length = Some(self.new_indexed_data(operator, operand, l.clone()))
+            }
+        }
+    }
+
+    fn sum_other_length(
+        &mut self,
+        operator: &Option<CalcOperator>,
```
kdy1, line 430: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+        operand: &CalcProduct,
+        l: &Length,
+    ) {
+        let unit = l.unit.value.to_ascii_lowercase().to_string();
```
kdy1, line 434: This is inefficient. Please remove `to_string()`

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+        }
+    }
+
+    fn sum_unknown_dimension(
+        &mut self,
+        operator: &Option<CalcOperator>,
+        operand: &CalcProduct,
+        u: &UnknownDimension,
+    ) {
+        let unit = u.unit.value.to_ascii_lowercase().to_string();
```
kdy1, line 674: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+    number: Option<IndexedOperatorAndOperand<Number>>,
+    percentage: Option<IndexedOperatorAndOperand<Percentage>>,
+    absolute_length: Option<IndexedOperatorAndOperand<Length>>,
+    other_lengths: HashMap<String, IndexedOperatorAndOperand<Length>>,
+    angle: Option<IndexedOperatorAndOperand<Angle>>,
+    duration: Option<IndexedOperatorAndOperand<Time>>,
+    frequency: Option<IndexedOperatorAndOperand<Frequency>>,
+    resolution: Option<IndexedOperatorAndOperand<Resolution>>,
+    flex: Option<IndexedOperatorAndOperand<Flex>>,
+    unknown_dimension: HashMap<String, IndexedOperatorAndOperand<UnknownDimension>>,
```
kdy1, line 165: Please use `JsWord` as the key.

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+fn get_precision(n: f64) -> usize {
+    let n_as_str = n.to_string();
+    n_as_str.find('.').map_or(0, |sep| n_as_str.len() - sep)
+}
+
+struct CalcSumContext {
+    number: Option<IndexedOperatorAndOperand<Number>>,
+    percentage: Option<IndexedOperatorAndOperand<Percentage>>,
+    absolute_length: Option<IndexedOperatorAndOperand<Length>>,
+    other_lengths: HashMap<String, IndexedOperatorAndOperand<Length>>,
```
kdy1, line 159: Please change type to `AHashMap`.
You can import it from `swc_common`

crates/swc_css_minifier/src/compressor/calc_sum.rs
kdy1: Same here, please use `AHashMap`

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+                }
+            }
+            _ => {}
+        }
+    }
+}
+
+fn try_to_extract_into_calc_value(calc_sum: &CalcSum) -> Option<CalcValue> {
+    if calc_sum.expressions.len() == 1 {
+        return match calc_sum.expressions.get(0).unwrap() {
```
kdy1, line 38: `&calc_sum.expressions[0]` seems better than `calc_sum.expressions.get(0).unwrap()`

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+fn is_calc_function_name(ident: &Ident) -> bool {
+    ident.value.to_ascii_lowercase() == js_word!("calc")
+        || ident.value.to_ascii_lowercase() == js_word!("-webkit-calc")
+        || ident.value.to_ascii_lowercase() == js_word!("-moz-calc")
+}
+
+// transform "(simple calc-value)" into "simple calc-value"
+fn remove_unnecessary_nesting_from_calc_sum(calc_sum: &mut CalcSum) {
+    if calc_sum.expressions.len() == 1 {
+        match calc_sum.expressions.get(0).unwrap() {
```
kdy1, line 17: Same here, please use index syntax

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+            CalcProductOrOperator::Product(CalcProduct {
+                expressions: calc_product_expressions,
+                ..
+            }) if calc_product_expressions.len() == 1 => {
+                if let CalcValueOrOperator::Value(CalcValue::Sum(CalcSum {
+                    expressions: nested_expressions,
+                    span: nested_span,
+                })) = calc_product_expressions.get(0).unwrap()
```
kdy1, line 25: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+}
+
+fn try_to_extract_into_calc_value(calc_sum: &CalcSum) -> Option<CalcValue> {
+    if calc_sum.expressions.len() == 1 {
+        return match calc_sum.expressions.get(0).unwrap() {
+            CalcProductOrOperator::Product(CalcProduct {
+                expressions: calc_product_expressions,
+                ..
+            }) if calc_product_expressions.len() == 1 => {
+                match calc_product_expressions.get(0).unwrap() {
```
kdy1, line 43: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+                    },
+                }),
+                _ => unreachable!("Operator value can only be Add or Sub"),
+            },
+        }
+    }
+
+    fn reduce(&mut self, operator: &Option<CalcOperator>, operand: &CalcProduct) {
+        if operand.expressions.len() == 1 {
+            match operand.expressions.get(0).unwrap() {
```
kdy1, line 245: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+        &mut self,
+        component_value: &mut ComponentValue,
+    ) {
+        match &component_value {
+            // Transform "calc(calc-sum)" into "simple value" when calc-sum is not a complex
+            // expression
+            ComponentValue::Function(Function { name, value, .. })
+                if is_calc_function_name(name) && value.len() == 1 =>
+            {
+                match value.get(0).unwrap() {
```
kdy1, line 1063: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+                    ComponentValue::CalcSum(CalcSum {
+                        expressions: calc_sum_expressions,
+                        ..
+                    }) if calc_sum_expressions.len() == 1 => {
+                        match calc_sum_expressions.get(0).unwrap() {
```
kdy1, line 1068: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+                            CalcProductOrOperator::Product(CalcProduct {
+                                expressions: calc_product_expressions,
+                                ..
+                            }) if calc_product_expressions.len() == 1 => {
+                                match calc_product_expressions.get(0).unwrap() {
```
kdy1, line 1073: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+                    self.push(operator, operand);
+                }
+            }
+            None => self.duration = Some(self.new_indexed_data(operator, operand, d.clone())),
+        }
+    }
+
+    fn sum_frequency(
+        &mut self,
+        operator: &Option<CalcOperator>,
```
kdy1, line 548: Same here, please move `&` into `Option`

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+                    self.push(operator, operand);
+                }
+            }
+            None => self.frequency = Some(self.new_indexed_data(operator, operand, f.clone())),
+        }
+    }
+
+    fn sum_resolution(
+        &mut self,
+        operator: &Option<CalcOperator>,
```
kdy1, line 592: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+                    );
+                } else {
+                    self.push(operator, operand);
+                }
+            }
+            None => self.resolution = Some(self.new_indexed_data(operator, operand, r.clone())),
+        }
+    }
+
+    fn sum_flex(&mut self, operator: &Option<CalcOperator>, operand: &CalcProduct, f: &Flex) {
```
kdy1, line 634: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+                    self.push(operator, operand);
+                }
+            }
+            None => self.flex = Some(self.new_indexed_data(operator, operand, f.clone())),
+        }
+    }
+
+    fn sum_unknown_dimension(
+        &mut self,
+        operator: &Option<CalcOperator>,
```
kdy1, line 670: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+                let indexed_data: IndexedOperatorAndOperand<UnknownDimension> =
+                    self.new_indexed_data(operator, operand, u.clone());
+                self.unknown_dimension.insert(unit, indexed_data);
+            }
+        }
+    }
+
+    fn new_indexed_data<T>(
+        &mut self,
+        operator: &Option<CalcOperator>,
```
kdy1, line 715: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+        index: usize,
+        operator: CalcOperator,
+    ) {
+        if let Some(elem) = expressions.get_mut(index) {
+            *elem = CalcProductOrOperator::Operator(operator);
+        }
+    }
+
+    fn try_to_sum_values(
+        operator1: &Option<IndexedData<CalcOperator>>,
```
kdy1, line 783: Same here

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1102 @@
...
+        operator2: &Option<CalcOperator>,
```
kdy1, line 784: Same here

crates/swc_css_minifier/src/compressor/unit/mod.rs
```diff
@@ -0,0 +1,151 @@
+use phf::phf_map;
+
+// We need to use several declarations (instead of one giant nested map)
+// because of a current limitation of phf (https://github.com/rust-phf/rust-phf/issues/183)
+type RatioMap = phf::Map<&'static str, f64>;
+
+// For length ratio, see https://www.w3.org/TR/css-values-4/#absolute-lengths
+static CM: RatioMap = phf_map! {
```
kdy1, line 8: I expect a match expression of `JsWord` to be faster than `phf`

kdy1 (changes requested):

ghostd:
@kdy1 I fixed all your remarks

crates/swc_css_minifier/src/compressor/calc_sum.rs
kdy1: Also here

crates/swc_css_minifier/src/compressor/calc_sum.rs
ghostd: My bad, i missed that part. Done now :)

crates/swc_css_minifier/src/compressor/calc_sum.rs
kdy1: Seems like you missed this

crates/swc_css_minifier/src/compressor/calc_sum.rs
kdy1: And this

crates/swc_css_minifier/src/compressor/calc_sum.rs
```diff
@@ -0,0 +1,1083 @@
...
+            }
+        }
+    }
+
+    fn merge_operators(
+        surrounding_operator: Option<&CalcOperator>,
+        nested_operator: Option<&CalcOperator>,
+    ) -> Option<CalcOperator> {
+        match nested_operator {
+            None => surrounding_operator.map(|so| (*so).clone()),
```
kdy1, line 203: I think `surrounding_operator.cloned()` is better than `surrounding_operator.map(|so| (*so).clone())`

crates/swc_css_minifier/src/compressor/calc_sum.rs
kdy1: You can change `other_lengths` to `AHashMap<JsWord, IndexedOperatorAndOperand<Length>>`

crates/swc_css_minifier/src/compressor/calc_sum.rs
kdy1: Same here, you can change the key type of `unknown_dimension` to `JsWord`

crates/swc_css_minifier/src/compressor/calc_sum.rs
kdy1: Please change this to `Option<&IndexedData<CalcOperator>>`

kdy1 (changes requested):

crates/swc_css_minifier/src/compressor/calc_sum.rs
ghostd: You're right, done

kdy1 (dismissed):
Thank you!

Merging as @alexander-akait  approved, and I verified rust code styles

---

swc-bump:
 - swc_css_minifier

ghostd:
It seems i forgot to update some "snapshots" for swc_css_parser... but i don't find an easy way to update the files :-/

kdy1:
`UPDATE=1 cargo test` in the crate directory will do the trick

kdy1 (approved):

crates/swc_css_minifier/tests/fixture/compress-calc/media-query/output.min.css
yisibl: It will be done here: https://github.com/swc-project/swc/pull/6140

PR #5975, opened 2022-09-28
perf(es): Don't share `Globals`

kdy1:
**Description:**

We don't need to share one instance of `Globals`.

---

# Investigation of test failure

Added log: 

```
  DEBUG  ===== Before pure =====
"use strict";
function _checkPrivateRedeclaration(obj, privateCollection) {
    if (privateCollection.has(obj)) {
        throw new TypeError("Cannot initialize the same private elements twice on an object");
    }
}
function _classApplyDescriptorGet(receiver, descriptor) {
    if (descriptor.get) {
        return descriptor.get.call(receiver);
    }
    return descriptor.value;
}
function _classExtractFieldDescriptor(receiver, privateMap, action) {
    if (!privateMap.has(receiver)) {
        throw new TypeError("attempted to " + action + " private field on non-instance");
    }
    return privateMap.get(receiver);
}
function _classPrivateFieldGet(receiver, privateMap) {
    var descriptor = _classExtractFieldDescriptor(receiver, privateMap, "get");
    return _classApplyDescriptorGet(receiver, descriptor);
}
function _classPrivateFieldInit(obj, privateMap, value) {
    _checkPrivateRedeclaration(obj, privateMap);
    privateMap.set(obj, value);
}
var foo = "bar";
var _bar = new WeakMap();
class Foo {
    test() {
        return _classPrivateFieldGet(this, _bar);
    }
    constructor(){
        _classPrivateFieldInit(this, _bar, {
            writable: true,
            value: foo
        });
    }
}
const f = new Foo;
expect(f.test()).toBe(foo);
expect("bar" in f).toBe(false);

===== After pure =====
"use strict";
function _checkPrivateRedeclaration(obj, privateCollection) {
    if (privateCollection.has(obj)) {
        throw new TypeError("Cannot initialize the same private elements twice on an object");
    }
}
function _classApplyDescriptorGet(receiver, descriptor) {
    if (descriptor.get) {
        return descriptor.get.call(receiver);
    }
    return descriptor.value;
}
function _classExtractFieldDescriptor(receiver, privateMap, action) {
    if (!privateMap.has(receiver)) {
        throw new TypeError("attempted to " + action + " private field on non-instance");
    }
    return privateMap.get(receiver);
}
function _classPrivateFieldGet(receiver, privateMap) {
    var descriptor = _classExtractFieldDescriptor(receiver, privateMap, "get");
    return _classApplyDescriptorGet(receiver, descriptor);
}
function _classPrivateFieldInit(obj, privateMap, value) {}
var foo = "bar";
var _bar = new WeakMap();
class Foo {
    test() {
        return _classPrivateFieldGet(this, _bar);
    }
    constructor(){
        _classPrivateFieldInit(this, _bar, {
            writable: true,
            value: foo
        });
    }
}
const f = new Foo;
expect(f.test()).toBe(foo);
expect("bar" in f).toBe(false);
```

and I found

```

  DEBUG  ignore_return_value: Dropping a pure call, kind: "change"
    at crates/swc_ecma_minifier/src/compress/pure/misc.rs:662
    in visit_mut_fn_decl with id: _classPrivateFieldInit#5
    in apply pure optimizer
    in optimize with pass: 1
    in compress ast
    in minify
    in process_js_inner
    in process_js_with_custom_pass
    in process_js_file with fm: SourceFile(/Users/kdy1/projects/s/issues/crates/swc/tests/babel-exec/packages/babel-plugin-proposal-class-properties/test/fixtures/private/constructor-collision/exec.js)
    in test_file with idx: 12
```

too

So I think the `BytePos` for pure comment is wrong, due to misusage of `BytePos`

### DIFF
--- a/.github/workflows/cargo.yml
+++ b/.github/workflows/cargo.yml
@@ -569,7 +569,7 @@ jobs:
         env:
           SWC_FORCE_CONCURRENT: "1"
         run: |
-          cargo test -p swc --all-targets --features concurrent
+          cargo test -p swc --features concurrent
 
       - name: Install cargo-hack
         uses: baptiste0928/cargo-install@v1.1.0
--- a/crates/swc/benches/typescript.rs
+++ b/crates/swc/benches/typescript.rs
@@ -7,7 +7,9 @@ use std::{
 
 use criterion::{black_box, criterion_group, criterion_main, Bencher, Criterion};
 use swc::config::{Config, IsModule, JscConfig, Options, SourceMapsConfig};
-use swc_common::{errors::Handler, FileName, FilePathMapping, Mark, SourceFile, SourceMap};
+use swc_common::{
+    errors::Handler, FileName, FilePathMapping, Mark, SourceFile, SourceMap, GLOBALS,
+};
 use swc_ecma_ast::{EsVersion, Program};
 use swc_ecma_parser::{Syntax, TsConfig};
 use swc_ecma_transforms::{fixer, hygiene, resolver, typescript};
@@ -63,7 +65,7 @@ fn base_tr_group(c: &mut Criterion) {
 
 fn base_tr_fixer(b: &mut Bencher) {
     let c = mk();
-    c.run(|| {
+    GLOBALS.set(&Default::default(), || {
         let module = as_es(&c);
 
         b.iter(|| {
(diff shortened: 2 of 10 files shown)


kdy1 (commented):
swc-bump:
 - swc --breaking

PR #5977, opened 2022-09-28
perf(es/minifier): Optimize `MultiReplacer`

kdy1:
**Description:**

This PR splits the `MultiReplacer` into two variants. The first one is the cloning version, and this is parallel. The other one is the non-cloning version, which is not parallel.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5981, opened 2022-09-28
feat(es/minifier): Ignore more return values

kdy1:
**Description:**

I copied logic from the pure optimizer.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5984, opened 2022-09-28
fix(es/minifier): Don't panic on jsx or ts

kdy1:


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5986, opened 2022-09-29
test(es/minifier): Add tests for benchmark files

kdy1:
**Description:**

I added them to the test system so I can check the issues in them.


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #5988, opened 2022-09-29
feat(es/minifier): Handle more expressions while inlining IIFE

kdy1:
**Description:**


kdy1:
Closing as this is wrong due to the evaluation order

PR #5990, opened 2022-09-29
feat(es/minifier): Handle more expressions while inlining IIFE

kdy1:
**Description:**

I was wrong. I'll look for exact condition.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5991, opened 2022-09-29
feat(es/minifier): Invoke sequential inliner for inlined IIFEs

kdy1:
**Description:**

Inlined IIFEs have the exact pattern sequential inliner wants, so it's good to invoke it.


kdy1:
cc @Austaras Do you think this is a correct change?

Austaras:
At least this would do no harm.

kdy1:
Thank you!

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5993, opened 2022-09-29
feat(es/minifier): Drop vars from sequential inliner

kdy1:
**Description:**

This PR renames `CloningMultiReplacer` to `Finalizer` and makes it handle the removal of unused variables. Also, this PR improves the sequential inliner so we can drop variables within a single pass.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #5996, opened 2022-09-29
feat(html/codegen): added option to keep quotes when minified

alexander-akait:
**Description:**

Some old system can't parser HTML attributes without quotes (yes, it is legacy), currently we don't have ability to prevent it, so let's add an option

**BREAKING CHANGE:**

Yes

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_codegen

PR #6001, opened 2022-09-30
build(cargo): use workspace deps for internal pkg

kwonoj:
**Description:**

This PR attempts to apply workspaces dep inheritance over internal packages. If this publishes as expected, we can reduce lot of duplications across pkgs.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you! Let's try...

---

swc-bump:
 - swc_common

PR #6002, opened 2022-09-30
fix(es/minifier): Don't inline into nested scope

kdy1:
**Related issue:**

 - https://github.com/vercel/next.js/discussions/30237#discussioncomment-3768734


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #6003, opened 2022-09-30
fix(es/codegen): Improve performance of debug build

kdy1:


kdy1 (commented):
swc-bump:
 - swc_ecma_codegen

PR #6005, opened 2022-09-30
feat(es/minifier): Clone trivial literals in sequential inliner

kdy1:
**Description:**

As our minifier is two-pass by default, this is enough.

At the first pass, we inline all numeric literals using sequential inliner, and minifier can then inline constants correctly in the second pass.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #6006, opened 2022-09-30
perf: 🤖  remove unused field, better parallel resolve

IWANABETHATGUY:
**Description:**
1. Remove some unused function
2. Reimplement `Mark::fresh` logic, making it faster in parallel scenarios, especially in contention-heavy scenarios.
About 10% faster compare to **main** branch

**BREAKING CHANGE:**
1. `Mark::fresh` now has zero arguments.


kdy1 (changes requested):
If you are trying to fix the perf issue of `resolver`, I expect this to make the situation worse

IWANABETHATGUY:
> If you are trying to fix the perf issue of `resolver`, I expect this to make the situation worse

Would you mind giving me some advice?

kdy1:
You can optimize it by allocating more marks at once and storing it in a thread-local

IWANABETHATGUY:
> You can optimize it by allocating more marks at once and storing it in a thread-local

Thanks, I found another way to optimize `Mark::new()`, give me some time to explain about it.

crates/swc_common/src/syntax_pos.rs
```diff
@@ -505,39 +505,7 @@ impl Span {
 
     #[inline]
     pub fn private(self) -> Span {
-        self.apply_mark(Mark::fresh(Mark::root()))
-    }
-
-    #[inline]
-    pub fn adjust(&mut self, expansion: Mark) -> Option<Mark> {
```
IWANABETHATGUY, line 512: These functions are never used through our codebase, according to the function name, I think it is more related to macro. In our context, we may don't have such a complex scenario, so delete it.

crates/swc_common/src/syntax_pos/hygiene.rs
```diff
@@ -53,8 +53,6 @@ struct SyntaxContextData {
     prev_ctxt: SyntaxContext,
     // This context, but with all transparent and semi-transparent marks filtered away.
     opaque: SyntaxContext,
-    // This context, but with all transparent marks filtered away.
-    opaque_and_semitransparent: SyntaxContext,
```
IWANABETHATGUY, line 57: This field is never used so delete it.

crates/swc_common/src/syntax_pos/hygiene.rs
```diff
@@ -138,85 +133,6 @@ impl Mark {
         Mark(raw)
     }
 
-    #[inline]
-    pub fn parent(self) -> Mark {
```
IWANABETHATGUY, line 142: I search through the codebase, and find this function is never used, and also `MarkData.parent` field is never used too, so delete it.

crates/swc_common/src/syntax_pos/hygiene.rs
```diff
@@ -138,85 +133,6 @@ impl Mark {
...
-        #[cfg(all(feature = "__plugin_mode", target_arch = "wasm32"))]
-        return Mark(unsafe { __mark_parent_proxy(self.0) });
-
-        #[cfg(not(all(feature = "__plugin_mode", target_arch = "wasm32")))]
-        return HygieneData::with(|data| data.marks[self.0 as usize].parent);
-    }
-
-    #[inline]
-    pub fn is_builtin(self) -> bool {
```
IWANABETHATGUY, line 151: Also, I search through the codebase, and find this function is never used, and also  `MarkData.is_builtin` field is never used too, so delete it.

crates/swc_common/src/syntax_pos.rs
kdy1: cc @kwonoj thoughts? Do you think we can delete this?

crates/swc_common/src/syntax_pos/hygiene.rs
kdy1: This should not be deleted

crates/swc_common/src/syntax_pos/hygiene.rs
```diff
@@ -289,19 +177,13 @@ impl Default for HygieneData {
 impl HygieneData {
     pub(crate) fn new() -> Self {
         HygieneData {
```
IWANABETHATGUY, line 179: Because neither `MarkData.parent` or `MarkData.builtin`  is used through the codebase. Also, `data.marks` is only changed in `Mark::fresh` before, let's think about what is `Mark::fresh` did, it get "parent" `Mark` , and put it in `data.marks` arrray and use latest `data.marks.len() - 1` to create global unique `Mark`, if the data of `data.marks` is never used, we could just use a `u32` to do the same thing, every time we just increase one and create a new mark

crates/swc_common/src/syntax_pos/hygiene.rs
```diff
@@ -289,19 +177,13 @@ impl Default for HygieneData {
...
-                is_builtin: true,
-            }],
             syntax_contexts: vec![SyntaxContextData {
                 outer_mark: Mark::root(),
                 prev_ctxt: SyntaxContext(0),
                 opaque: SyntaxContext(0),
-                opaque_and_semitransparent: SyntaxContext(0),
             }],
             markings: HashMap::default(),
+            mark_count: 1,
```
IWANABETHATGUY, line 186: If you search through the codebase `Mark::fresh()` always pass one mark which mostly is `Mark::root` which means it will at least > `Mark (0)`, so we start with 1, it has the same effect, make the `Mark::fresh()` create mark consistently great than `Mark::root()`

crates/swc_common/src/syntax_pos/hygiene.rs
IWANABETHATGUY: I search the whole codebase the `parent param` of `Mark::fresh` actually does not mean any special thing, It always pass `Mark::root()`, It is just a placeholder IMO. Using `u32` instead of `Vec` will reduce a lot of lock time because a register operation is much cheaper than a memory operation.

IWANABETHATGUY:
All the rust test case is passed, the rest test cases are not passed because of the `Mark::refresh` breaking change.

crates/swc_common/src/syntax_pos.rs
kwonoj: What was the original use case? I think it's ok if there's no usage at all - but if there are legit usecase, even though it's not being used internally it may need to be preserved. swc/core is not the only consumer of this module, as this is a public interface.

kdy1: I didn't use this. I simply copied whole module from rustc

IWANABETHATGUY:
Here is the `swc_ecma_transforms_base/bench/parallel/ result: 
# main
```
es/base/parallel/resolver/typescript                                                                          
                        time:   [13.508 s 13.631 s 13.743 s]
Found 1 outliers among 10 measurements (10.00%)
```

# PR
```
es/base/parallel/resolver/typescript                                                                          
    time:   [12.448 s 12.577 s 12.722 s]
    change: [-11.218% -9.6772% -7.9568%] (p = 0.00 < 0.05)
    Performance has improved.
```


About 10%, not bad.

crates/swc_common/src/syntax_pos.rs
```diff
@@ -86,6 +88,7 @@ impl Globals {
         Globals {
             hygiene_data: Mutex::new(hygiene::HygieneData::new()),
             dummy_cnt: AtomicU32::new(DUMMY_RESERVE),
+            mark_count: AtomicU32::new(1),
```
IWANABETHATGUY, line 91: Using `AtomicU32`,  because `mark_count`(from original `HygineData.marks`) is only used in `Mark::fresh`, the `HygineData.marks` is used exclusive with `HygineData.syntax_contexts` and `HygineData.markings`, splitting them could improve concurrency in parallel scenario (`Mark::fresh` will not lock anymore, `Atomic` is faster than Mutex, even `parking_lot`), you could see the benchmark result.

crates/swc_common/src/syntax_pos/hygiene.rs
kdy1: I know it's not used in main swc codebase at now, but it was used and it will be used

crates/swc_common/src/syntax_pos/hygiene.rs
```diff
@@ -96,35 +93,39 @@ extern "C" {
...
 impl Mark {
-    /// Shortcut for `Mark::fresh(Mark::root())`
+    /// Shortcut for `Mark::fresh()`
     #[allow(clippy::new_without_default)]
     pub fn new() -> Self {
-        Mark::fresh(Mark::root())
+        Mark::fresh()
     }
 
-    pub fn fresh(parent: Mark) -> Self {
```
kdy1, line 105: Please restore this

crates/swc_common/src/syntax_pos/hygiene.rs
kdy1: And this

crates/swc_common/src/syntax_pos/hygiene.rs
```diff
@@ -138,85 +139,6 @@ impl Mark {
...
-        {
-            assert_ne!(self, Mark::root());
-
-            HygieneData::with(|data| data.marks[self.0 as usize].is_builtin = is_builtin)
-        }
-    }
-
-    #[allow(unused_assignments)]
-    #[cfg(all(feature = "__plugin_mode", target_arch = "wasm32"))]
-    pub fn is_descendant_of(mut self, ancestor: Mark) -> bool {
```
kdy1, line 179: And this

crates/swc_common/src/syntax_pos/hygiene.rs
```diff
@@ -138,85 +139,6 @@ impl Mark {
...
-            )
-            .expect("Should able to deserialize")
-        };
-        self = Mark::from_u32(context.0);
-
-        return context.2 != 0;
-    }
-
-    #[cfg(not(all(feature = "__plugin_mode", target_arch = "wasm32")))]
-    pub fn is_descendant_of(mut self, ancestor: Mark) -> bool {
```
kdy1, line 208: Same here

crates/swc_common/src/syntax_pos/hygiene.rs
```diff
@@ -242,40 +164,11 @@ impl Mark {
...
-    /// to) the returned mark. That is, the following holds:
-    ///
-    /// ```rust,ignore
-    /// let la = least_ancestor(a, b);
-    /// assert!(a.is_descendant_of(la))
-    /// assert!(b.is_descendant_of(la))
-    /// ```
-    #[allow(unused_mut)]
-    #[cfg(not(all(feature = "__plugin_mode", target_arch = "wasm32")))]
-    pub fn least_ancestor(mut a: Mark, mut b: Mark) -> Mark {
```
kdy1, line 256: Same here

kdy1 (changes requested):
I'll not accept this PR as-is, and please use github for questions

IWANABETHATGUY:
@kdy1 thanks for your reviewing, I close this pull request and make another pr https://github.com/swc-project/swc/pull/6019/files, with no API deletion, only adjusting the data structures.

PR #6008, opened 2022-09-30
fix(es/base): Renamer use seperate scope for FnExpr Ident

Austaras:
**Description:**

A follow up of #5959 

**BREAKING CHANGE:**


crates/swc_ecma_minifier/tests/terser/compress/typeof/duplicate_lambda_arg_name/output.mangleOnly.js
```diff
@@ -1,3 +1,3 @@
-console.log((function o(n) {
-    return typeof n;
+console.log((function o(o) {
```
kdy1, line 1: This seems like a regression. Is this patch really required?

kdy1: Oh nevermind I misread it

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_base

PR #6011, opened 2022-10-01
feat(es/minifier): Make sequential inliner self-repeat

kdy1:
**Description:**

We should process as much as possible within the two-pass limit.


kdy1:
Closing as the gain (~0.001%) is too small compared to the performance difference (1%~)

PR #6019, opened 2022-10-02
perf(swc_common):split-hygine-data

IWANABETHATGUY:
**Description:**
1. Split `HygineData` into two parts. `HygineData.maks` and (`HygineData.syntax_contexts`, `HygineData.markings`), are always used exclusively, using two `Mutex` could reduce some unnecessary locking. e.g. It is not necessary to lock  (`HygineData.syntax_contexts`, `HygineData.markings`) when invoke `Mark::refresh() `, vise versa.
This give us about 10% performance improvement in `transform_base/parallel`
**before**: 
```
es/base/parallel/resolver/typescript                                                                          
                        time:   [13.508 s 13.631 s 13.743 s]
Found 1 outliers among 10 measurements (10.00%)
```

**after**:
```
es/base/parallel/resolver/typescript                                                                          
                        time:   [12.107 s 12.247 s 12.382 s]
```

**BREAKING CHANGE:**


.vscode/settings.json
```diff
@@ -25,7 +25,7 @@
     "eslint.enable": false,
     "cSpell.allowCompoundWords": true,
     "cSpell.caseSensitive": true,
-    "rust-analyzer.checkOnSave.command": "clippy",
+    // "rust-analyzer.checkOnSave.command": "clippy",
```
kdy1, line 28: Please revert this

crates/swc_common/src/syntax_pos/hygiene.rs
```diff
@@ -316,6 +309,16 @@ impl HygieneData {
     }
 }
 
+pub fn with_marks<T, F: FnOnce(&mut Vec<MarkData>) -> T>(f: F) -> T {
```
kdy1, line 312: Please use `pub(crate)`

IWANABETHATGUY: Done

crates/swc_common/src/syntax_pos/hygiene.rs
```diff
@@ -63,9 +63,9 @@ pub struct Mark(u32);
 
 #[allow(unused)]
 #[derive(Clone, Debug)]
-struct MarkData {
-    parent: Mark,
-    is_builtin: bool,
+pub struct MarkData {
```
kdy1, line 66: Please reduce the scope

kdy1: In all places, not only this

IWANABETHATGUY: Done

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_common

PR #6020, opened 2022-10-02
feat(es/minifier): Ignore return values of ignored IIFEs

kdy1:
**Description:**

As we are not using the return value of the function, we can call `ignore_return_value` on the argument of the return statements of IIFE.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #6027, opened 2022-10-03
feat(es/minifier): Inline initialized before declaration vars

Austaras:
**Description:**

For cases like 
```js
export function foo() {
    b = 1;
    console.log(2);
    console.log(b);

    for (var b, c = 0; c < 10; c++) {
        console.log(c);
    }
}
```
Normally people won't write code like this, but sometimes `collapse_vars` would generate such code.
**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_minifier

PR #6033, opened 2022-10-04
fix(es/minifier): Don't evaluate `String.fromCharCode` for non-ascii values

kdy1:
**Related issue:**
 - https://github.com/vercel/next.js/discussions/30237#discussioncomment-3793184


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

Issue #5913, opened 2022-09-20
Incorrect yield and try-finally behaviour


tomas-c:
### Describe the bug

I'm using Next.js with SWC and redux-saga and noticed that updating Next.js from 12.2.1 to 12.3.1 broke redux-saga for me.

Turns out wrapping a yield statement in a try-finally block doesn't behave as it should with SWC.

### Input code

```javascript
(() => {
  function* generator() {
    try {
      while (true) {
        yield "foo";
      }
    } finally {
      yield "bar";
    }
  }

  function test() {
    const gen = generator();
    console.log(gen.next().value);
    console.log(gen.next().value);
    console.log(gen.next().value);
    console.log(gen.next().value);
  }

  test();
})();
```

### Config

Copied from playground defaults:
```
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.3.2&code=H4sIAAAAAAAAA72PwQ2DMAxF70zxxSn0wAIR3SWlhkayYimYtqjK7gQaUSboxbb8v%2B1nYxp0V3wqYJhDr17CBSMFik4lZnFTAI1LqYDXwzPBaJypOZrA4onvqAeR2pZm2nPC4INj%2Fm0o1puLxboZU3VigNKkx%2FVewqQbFbozmz1EYWpZRpPFNtA7T7ZPx5nvr5b9gy%2B4rVKT4wq0x%2FTnXgEAAA%3D%3D&config=H4sIAAAAAAAAA0WNSwrDMAxE76J1ts0id%2BghjKsEF%2F%2FQKBBjfPfawSU7aebpqdIXlrZK2QhYxoQS1Vy0kZbMsOKy0kKKHu3Gg1tfjBysHWG8eudTAs92oeCi28sw2RSyMPBUJh7%2BT7YuCulzjqDez27hSu1xzDuH9wRVTm4%2FOkwQUrUAAAA%3D

### Expected behavior

Running the **input** code from the playground outputs

```
foo
foo
foo
foo
```

### Actual behavior

Running the **output** code from the playground outputs

```
foo
bar
foo
bar
```

### Version

1.3.2

kdy1:
Expected output:

```js

"use strict";

[...]

PR #6042, opened 2022-10-05
fix(es/compat): Fix `yield` in try-finally

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/5913


kdy1:
Postponing

Issue #6039, opened 2022-10-04
Faulty compression leads to arrays with holes / Removes assignments passed by value


icyJoseph:
### Describe the bug

Alright, I've got my case:

```js
function foo() {
  let walker = 0;

  let arr = [];

  function bar(defaultValue) {
    const myIndex = walker;
    walker += 1;

    console.log({ arr });

    if (arr.length < myIndex + 1) {
      arr[myIndex] = defaultValue;
    }
  }

  return bar;
}

const bar = foo();

bar(null);
bar(null);
bar(null);
bar(null);
bar(null);
```

When compiled with `swc` using this `.swcrc` file:

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript"
    },
    "target": "es5",
    "minify": {
      "compress": true,
      "mangle": true
    }
  }
}
```

Outputs code that causes the walker to `+1` before the `if`-statement that assigns the first default value. So for the first item we are setting the index 1, which creates a whole on the array.

```js
var l = (function () {
  var l = 0,
    n = [];
  return function (u) {
    (l += 1),
      console.log({
        arr: n,
      }),
      n.length < l + 1 && (n[l] = u);
  };
})();
l(null), l(null), l(null), l(null), l(null);
```

The final log outputs: `{ arr: [ <1 empty item>, null, null, null, null ] }`, but since the very beginning there's a hole at the beginning of the array.

`package.json` used:

```json
{
  "name": "swc-test",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "devDependencies": {
    "@swc/cli": "^0.1.57",
    "@swc/core": "^1.3.4"
  }
}
```

If we set ` "compress": false,` on the `.swcrc` file, then the bug goes away, aka the last log would be : `{ arr: [ null, null, null, null ] }` - no log ever shows the hole in the array.

What does this mean for Next.js? Well that perhaps a new version with a fixed version of SWC is needed to be able to use the minifier. I'll try to report the bug to SWC as best as I can so that they can fix it.

This was first observed with libraries that use this custom React hook: https://github.com/react-component/overflow/blob/master/src/hooks/useBatchFrameState.tsx - You see the core of the issue is that:

```js
    const myIndex = walkingIndex;
    walkingIndex += 1;

    // Fill value if not exist yet
    if (statesRef.current.length < myIndex + 1) {
      statesRef.current[myIndex] = defaultValue;
    }
```

[...]

PR #6043, opened 2022-10-05
fix(es/minifier): Fix analysis of var declaration after usage

kdy1:
**Description:**

`var_initialized` should be `true` even if the declaration of variable comes after its usage.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6039.


crates/swc_ecma_minifier/src/util/base54.rs
```diff
@@ -186,16 +186,16 @@ impl CharFreq {
...
-        {
-            let considered = s
-                .chars()
-                .filter(|&c| Ident::is_valid_continue(c))
-                .collect::<String>();
-            if !considered.is_empty() {
-                tracing::debug!("Scanning: `{}` with delta {}", considered, delta);
-            }
-        }
+        // #[cfg(feature = "debug")]
```
kdy1, line 189: I commented this out as this disturbs debugging

crates/swc_ecma_minifier/src/pass/precompress.rs
```diff
@@ -34,6 +34,23 @@ impl VisitMut for PrecompressOptimizer {
         }
     }
 
+    fn visit_mut_pat_or_expr(&mut self, n: &mut PatOrExpr) {
```
kdy1, line 37: I added a small normalization to reduce possibility of bugs

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

jridgewell (approved):
I'm confused by you're wording in the OP. Everything appears to be declared before use, when does an item get inserted into vars?

kdy1:
I meant

```
function foo() {
    console.log(a)
}

var a = 1
```

`before` here means the order of visit

kdy1:
Anyway, thank you!

PR #6044, opened 2022-10-05
fix(es/minifier): Don't track usage assign

Austaras:
**Description:**
Don't know if it's right, but it generate correct output.

**BREAKING CHANGE:**


crates/swc_ecma_minifier/src/analyzer/storage/normal.rs
```diff
@@ -257,21 +257,28 @@ impl ProgramData {
...
-                    self.initialized_vars.insert(i);
+                    self.initialized_vars.insert(i.clone());
                     e.assign_count -= 1;
                     e.var_initialized = true;
                 } else {
                     e.reassigned_with_assignment = true
                 }
             }
 
+            print!("{} infects: ", i.0);
```
kdy1, line 276: Please remove this line

crates/swc_ecma_minifier/src/analyzer/storage/normal.rs
```diff
@@ -257,21 +257,28 @@ impl ProgramData {
...
+
+            for other in e.infects.clone() {
+                print!("{}, ", other.0);
```
kdy1, line 279: Same here

crates/swc_ecma_minifier/src/analyzer/storage/normal.rs
```diff
@@ -257,21 +257,28 @@ impl ProgramData {
...
+            }
+            println!();
```
kdy1, line 281: Same here

crates/swc_ecma_minifier/src/analyzer/storage/normal.rs
```diff
@@ -280,6 +287,8 @@ impl ProgramData {
                 e.mutation_by_call_count += 1;
             }
 
+            println!("use {}", i.0);
```
kdy1, line 290: Same here

crates/swc_ecma_minifier/src/analyzer/storage/normal.rs
```diff
@@ -257,21 +257,28 @@ impl ProgramData {
             }
 
             if ctx.is_op_assign {
+                println!("use {}", i.0);
```
kdy1, line 260: Same here

kdy1 (changes requested):
Thank you

Austaras:
Oh, sorry

crates/swc_ecma_minifier/tests/benches-full/lodash.js
```diff
@@ -1504,8 +1504,8 @@
             var func = Math[methodName];
             return function(number, precision) {
                 if (number = toNumber(number), (precision = null == precision ? 0 : nativeMin(toInteger(precision), 292)) && nativeIsFinite(number)) {
-                    var pair = (toString(number) + 'e').split('e'), value = func(pair[0] + 'e' + (+pair[1] + precision));
```
kdy1, line 1507: I think this can be problematic if `pair[0]` modifies `toString`

kdy1: It's the goal of infection analysis (although I'm not sure if this is a correct name)

crates/swc_ecma_minifier/tests/projects/output/angular-1.2.5.js
```diff
@@ -2757,7 +2756,7 @@
                     $watchCollection: function(obj, listener) {
                         var oldValue, newValue, self = this, changeDetected = 0, objGetter = $parse(obj), internalArray = [], internalObject = {}, oldLength = 0;
                         return this.$watch(function() {
-                            if (newValue = objGetter(self), isObject(newValue)) {
```
kdy1, line 2760: Same here. We need to prove that `objGetter(self)` does not modify `isObject`.

kdy1 (commented):
I'm not sure if we are proving these facts.
I added two-way infection, which is clearly not optimal, because it prevents bugs

crates/swc_ecma_minifier/tests/benches-full/lodash.js
Austaras: I've tested this code in terser and it generates identical output.

> I think this can be problematic if pair[0] modifies toString

Do you mean access to pair[0] may modify toString?

kdy1: Yes, we may need more execution tests with getters with side effects. e.g.

```js

let foo = () => 1;

const obj = {
    get 0() {
        foo = () => 2
        return 40
    }
};

```

kdy1: Oh, I have an idea. I'll review this by diffing with terser

Austaras: Seems like this to be the same reason as terser/terser/#1244

kdy1: Oh... I see. Thank you!

Austaras: In this case
```js
let foo = ()=>1;
const obj = {
    get 0 () {
        return foo = ()=>2, 40;
    }
};
var c = obj(0);
console.log(foo(c));
```
this PR would generate
```js
let foo = ()=>1;
const obj = {
    get 0 () {
        return foo = ()=>2, 40;
    }
};
var c = obj(0);
console.log(foo(c));
```
So I guess it's fine.

kdy1: It's not the exact problematic case. We can't know input.
Sometimes there are even already-minified code in the input, which is hard to imagine

Austaras: Oh, wait, this is wrong, let me recheck it.

Austaras: So, for code
```js
let foo = () => 1;

const obj = {
    get 0() {
        foo = () => 2;
        return 40;
    },
};
console.log(obj)

var c = obj[0];

console.log(foo(c));
```
SWC would incorrectly inline c because of #5961. I believe the solution is very simple -- any reassignment to foo should prevent seq inline. However this has nothing to do with this PR.

kdy1: Yeah ignoring side effects of getters is fine, as terser does it too

kdy1: Problem is the case below

Austaras: No, terser wouldn't inline `c` here.

kdy1: Oh, so infection analysis is not catching the related of `obj` and `foo`.
Maybe I need to file an issue...

crates/swc_ecma_minifier/tests/projects/output/angular-1.2.5.js
Austaras: Same here. terser would inline newValue as long ass isObject is declared.

crates/swc_ecma_minifier/tests/benches-full/lodash.js
Austaras: I don't think it has any relation with infection analysis in this case. Any reassignment to `foo` should prevent seq inline.

kdy1: Oh yes it would be better

crates/swc_ecma_minifier/tests/projects/output/angular-1.2.5.js
kdy1: No, the declaration is not the condition. I changed the definition of `isObject` to `var isObject = globalThis.isObject;` and terser preserves 

```js
newValue = objGetter(self);
                        var newLength, key;
                        if (!isObject(newValue)) {
                            if (oldValue !== newValue) {
                                oldValue = newValue;
                                changeDetected++
                            }
```

https://gist.github.com/kdy1/c4247944514115421aaed0c9288b6d38#file-beutified-js-L5291-L5293

https://gist.github.com/kdy1/c4247944514115421aaed0c9288b6d38

kdy1 (changes requested):

Austaras:
Yeah, if `isObject` is a normal function it would work fine. Maybe there's some more complex mechanism.

Austaras:
The difference can be run down to https://github.com/terser/terser/blob/master/lib/compress/tighten-body.js#L887. It checks if it's a AST_SymbolDefun(definition of function) or defined in current scope, which I believe is just a convenient implementation.

Austaras:
> The difference can be run down to https://github.com/terser/terser/blob/master/lib/compress/tighten-body.js#L887. It checks if it's a AST_SymbolDefun(definition of function) or defined in current scope, which I believe is just a convenient implementation.

So basically `function isObject` is fine, but `const isObject = function` is fine only if it's defined in same scope of `newValue`.

jridgewell (approved):

kdy1 (approved):
This PR seems innocent.
I'll merge this after merging https://github.com/swc-project/swc/pull/6053

---


swc-bump:
 - swc_ecma_minifier

kdy1 (approved):
Thank you!

Issue #6047, opened 2022-10-05
minifier should not skip callee in more cases


kdy1:
### Describe the bug

Related to https://github.com/swc-project/swc/pull/6044

If an expression has a side effect on the identifier in the callee position, this is not caught by the infection analysis.

We can fix this by aborting on reassignments to the callee, ~and this seems better.~
(This resulted in a too much regression)

### Input code

```typescript
let foo = () => 1;

const obj = {
    get 0() {
        foo = () => 2;
        return 40;
    },
};
console.log(obj)

var c = obj[0];

console.log(foo(c));
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2022",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": true,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true,
        "const_to_let": true,
        "pristine_globals": true
      },
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

[...]

PR #6048, opened 2022-10-05
fix(es/minifier): Fix sequential inliner

kdy1:
**Description:**

cc @Austaras
I think we should fix this by improving infection analysis, as regression is way too much

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6047.


Austaras:
Seems like many are caused by unresolved idents, maybe we could fix #6049 first?

kdy1:
I'll try

kdy1:
There are still lots of regressions

Austaras:
Oh, I mean we should stop at where **callee** is reassigned

kdy1:
Ah, thank you!
I did something wrong 🤣

kdy1:
It didn't work, and actually we were not skipping callee anyway

Austaras:
Well basically I thought it should be
```rs
// sequences.rs#1530
 Expr::Call(CallExpr {
                callee: Callee::Expr(b_callee),
                args: b_args,
                ..
            }) => {
                let is_this_undefined = b_callee.is_ident();
                trace_op!("seq: Try callee of call");

                if let Expr::Ident(i) = &**b_callee {
                    if let Some(usage) = self.data.vars.get(&i.to_id()) {
                        if usage.reassigned() {
                            return Ok(false);
                        }
                    }
                }
```
But that would cause lots of regressions too.

Issue #6051, opened 2022-10-05
Required leading zeroes are removed from numeric object keys with `compress`


alattalatta:
### Describe the bug

With `compress` turned on, SWC optimizes numeric string keys in objects and get rid of quotes. But this behavior is too aggressive and breaks codes when the said keys start with zeroes.

### Input code

```typescript
const o = { '010': '', '123': 'bar' }
console.log(o)
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": true,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true,
        "const_to_let": true,
        "pristine_globals": true
      },
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

[...]

PR #6052, opened 2022-10-05
fix(es/minifier): Fix optimization of numeric keys

kdy1:
**Description:**

We should not optimize if the key starts with `0`.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6051.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #6054, opened 2022-10-05
fix(swc): exclude option not processed from cli 

### LINKED ISSUE #6009: swc-cli: -C "exclude=.*\.spec\.ts" doesn't work from the cli, while it works ok from .swcrc file
### Describe the bug

```sh
npx swc src --out-dir "some dist" -C "exclude=.*.spec.ts$"
```
Doesn't work.

While `exclude: [".*.spec.ts$"]` works if used in `.swcrc`.

### Expected behavior

Should work just like it does with `.swcrc`

### Version

@swc/cli: 0.1.57, @swc/core: 1.3.1

MohamedLamineAllal:
**Description:**

First time contributing here. And big thanks too for all the great work.

A Draft that fixes the issue https://github.com/swc-project/swc/issues/6009.
Wrote a test too.

I wonder if I didn't miss anything.

From my understanding, it seems the only place where the exclude matching happens is through the `RC.into_config()` method. And in `Compiler.read_config()` method. The excluded value from `opts` is not used. And only the one read from files.

And i guess adding the bellow in `Compiler.parse_js_as_input()` before `Compiler.read_config()`:

```rust
   // adding
   if let FileName::Real(ref path) = name {
       if !opts.config.matches(path)? {
           return Ok(None);
       }
   }
   // before read_config
   let config = self.read_config(opts, name)?;
   // .....
```

does fix the problem.

If cli options have no `exclude` in `config`. or `test`, the file would automatically match. Otherwise, the exclude would be run, And then test. And so the condition work in all cases.

`read_config()` in case `.swcrc` contains and exclude that match, would return to `None`. So for that, I run the match before it. And in case of a no matching, I returned `None`.

I run the tests through `cargo test -p swc` they pass well. I didn't test beyond that. Including building and running the binding against `swc/cli`.

**BREAKING CHANGE:**

None.

 - Closes #6009


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc

PR #6060, opened 2022-10-06
build(cargo): dedupe indexmap with known version

kwonoj:
**Description:**

We are experiencing https://github.com/tkaitchuck/aHash/issues/95 when we try to bring some deps into next-swc. Looks like using indexmap 1.6.1 seems most straightforward way to avoid, I'd like to try this as first step. 

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---
swc-bump:
 - swc_config
 - swc_ecma_utils
 - swc_fast_graph

Issue #6004, opened 2022-09-30
Compress is using deleted properties with `"sourceType": "module"`


vgurkov:
### Describe the bug

Compressing code with `"sourceType": "module"` can result in using properties in conditions **after** they are deleted.

This is a very trimmed-down example of code running in `jQuery` (there's also a playground link). It may be encountered within other libraries as well.
Example from `jQuery`:
There is a variable initialized reading properties at the start of the function that is used in checks later.:
```javascript
function defaultPrefilter( elem, props, opts ) {
	var prop, value, toggle, hooks, oldfire, propTween, restoreDisplay, display,
		isBox = "width" in props || "height" in props, // <-- the important line in this example
	...
```
Those properties are then deleted:
```javascript
for ( prop in props ) {
    value = props[ prop ];
    if ( rfxtypes.test( value ) ) {
        delete props[ prop ]; // <-- the properties are deleted here, so reading them later will not yield anything
...
```
Then the variable is checked:
```javascript
if ( isBox && .... ) { // <-- there result of the property check is stored in `isBox` is it does not matter that the properties are now deleted
...
```

When compress is enabled, the variable at the start of the function is omitted and the property check is used in the condition instead after the property is deleted:
```javascript
if ( ("width" in props || "height" in props) && .... ) { // <-- properties are deleted, so the check will never pass here
...
```

I have attached a playground link. 
If the source type is set as `script` the variable is kept instead, which will result in the correct behavior. 
With source type `module` the variable is omitted and checking deleted properties will result in conditions never passing.

### Input code

```typescript
const props = {'a': 1, 'b': 2};

const isBox = 'a' in props || 'b' in props;

for (const p in props) {
    delete props[p];
}

console.log(isBox);
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript"
    },
    "target": "es2016",
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
[...]

PR #6063, opened 2022-10-06
fix(es/minifier): Mark `delete` as a property mutation

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6004.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/src/analyzer/mod.rs
```diff
@@ -1040,7 +1061,7 @@ where
                 v.mark_indexed_with_dynamic_key();
             }
 
-            if self.ctx.in_assign_lhs {
+            if self.ctx.in_assign_lhs || self.ctx.is_delete_arg {
```
jridgewell, line 1064: I think this is only catching `delete foo.bar`, but fails to capture `delete foo.bar.baz`? Based on the `Expr::Ident(obj) = e.obj` we're inside of. Come to think of it, there are a few others like `foo().bar` that won't be caught?

kdy1: For `delete foo.bar.baz`, `foo` will be marked as property mutated because we don't unset `is_delete_arg` for the `obj` of `MemberExpr`. ( https://github.com/swc-project/swc/blob/114bc97cf5ba430da4b9561613f4657c9b4afeb8/crates/swc_ecma_minifier/src/analyzer/mod.rs#L1032-L1040 )
And currently, we do not inline properties. But nice catch, we should ensure it when we add partial, per-property inlining.

jridgewell (approved):
Ok, we're tracking if a variable reference has any mutations, and this is now detecting property mutations as a mutation of the variable.

Issue #6066, opened 2022-10-06
Compress `unused` slows down with deeply nested conditions


dgreif:
### Describe the bug

I'm working on updating SWC (from `1.2.205` to `1.3.4`) and noticed that the latest version is very slow when trying to compress deeply nested conditions. We are using [peggy.js](https://peggyjs.org/) which can quickly generate some very deep nesting. As the nesting approaches a depth of ~20, compression seems to slow down exponentially, and is on the order of many seconds even for a relatively small file size.

I was able to recreate the issue with some very basic peggy files - see https://github.com/dgreif/peggy-swc

The "large" file is only 600 lines long, but has over 20 levels of nesting [here](https://github.com/dgreif/peggy-swc/blob/main/large.js#L539-L576). This file takes ~20 seconds to compress. If I set `compress.unused = false`, it is almost instant.

I believe this is a duplicate of #5470, but that issue was closed because they were unable to provide a reproduction. I believe they were also using peggy, though they did not explicitly say so.

### Config

```json
{
  "$schema": "https://json.schemastore.org/swcrc",
  "minify": true,
  "jsc": {
    "minify": {
      "compress": {
        "unused": true
      },
      "mangle": true
    }
  }
}
```

### Expected behavior

SWC is blazing fast with all of our other code. I would expect it to handle deeply nested conditional statements with similar speed.

### Version

1.3.4

kdy1:
I think the new DCE is the cause, although bisecting is required to verify it

kdy1:
<img width="957" alt="image" src="https://user-images.githubusercontent.com/29931815/194230855-f70da7a7-7ac6-4ea6-b75d-2fd37d6214b4.png">

Okay it was a correct guess

PR #6069, opened 2022-10-06
perf(es/minifier): Disable DCE for deeply nested inputs

kdy1:
**Description:**

This is an inherent weak point of the visitor pattern, so we just give up.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6066.


kdy1:
Oh I found the cause

PR #6075, opened 2022-10-06
Support aarch64 Linux with page sizes >= 16k

### LINKED ISSUE #5967: Support aarch64 Linux with page sizes >= 16k
### Describe the bug

swc uses jemalloc, which in turn uses the build system's page size as internal page size, and fails if the target's page size is larger [1]. Some Linux distributions patch jemalloc itself (e.g. [2]), but Github actions uses a "standard" jemalloc, which has this problem.

Instead of using a patched jemalloc, it is possible to configure jemalloc in runtime to use largest possible size, thus making it compatible with all `aarch64` systems.

**The solution is simple: add `export JEMALLOC_SYS_WITH_LG_PAGE=16` before running `cargo build` for `aarch64` system.**

This is a very common case with rust projects, as you can imagine 🙂 Here's just one example [3].

Let me know if it makes sense!

1: https://github.com/jemalloc/jemalloc/issues/467
2: https://github.com/archlinuxarm/PKGBUILDs/blob/master/extra/jemalloc/PKGBUILD
3: https://github.com/archlinuxarm/PKGBUILDs/pull/1940

### Input code

_No response_

### Config

_No response_

### Playground link

_No response_

### Expected behavior

No crash on Linux with 16k page size.

### Actual behavior

_No response_

### Version

1.3.3

### Additional context

_No response_

max-baz:
**Description:**

This adds support for aarch64 Linux with page sizes >= 16k, by compiling with jemalloc configured with largest possible page size.

It will still work for all other aarch64 Linux systems. To give you some confidence, I ran your Github actions in my fork, and compiled `1.5.1` release with this patch. You can grab the binaries and test for yourself!

https://github.com/maximbaz/swc/releases/tag/v1.5.1-1

I didn't touch Android as I don't have enough knowledge to confirm that this is needed for that platform or how to verify it.

Fixes https://github.com/swc-project/swc/issues/5967

I originally came to report this, because I wanted to fix Next.js swc plugin, which suffers from exactly the same issue and requires exactly the same fix. I can confirm that Next.js can finally be used on my aarch64 Linux machine with this patch.

### DIFF
--- a/.github/workflows/publish-node.yml
+++ b/.github/workflows/publish-node.yml
@@ -88,6 +88,7 @@ jobs:
             docker: ghcr.io/napi-rs/napi-rs/nodejs-rust:lts-debian-aarch64
             build: >-
               set -e &&
+              export JEMALLOC_SYS_WITH_LG_PAGE=16 &&
               rustup target add aarch64-unknown-linux-gnu &&
               RUSTFLAGS='' cargo build --manifest-path ./bindings/swc_cli/Cargo.toml --release --features plugin --target aarch64-unknown-linux-gnu &&
               cp ./bindings/target/aarch64-unknown-linux-gnu/release/swc . && chmod +x ./swc &&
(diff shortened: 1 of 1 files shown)


maximbaz:
I'm sorry, but I agree with this post:
https://drewdevault.com/2018/10/05/Dont-sign-a-CLA.html

My bad, I should have checked if you require CLA before opening a PR.
Feel free to take over getting the fix committed, if you want to support this CPU architecture 👍

kwonoj:
Thanks for the contribution and sorry that you're stuck with our CLA policy. Unfortunately, we require CLA for the contributors for the various reasons, something we can't move it away. Let me close this PR since you'd like to avoid sign it. Still appreciate for taking time to contribute.

PR #6085, opened 2022-10-08
doc(es/minifier): Document assumptions

kdy1:
**Description:**

I only added two assumptions. Those assumptions are same for almost all other tools, so it's fine to assume.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #6086, opened 2022-10-08
fix(es/codegen): Set `sourceRoot` of sourcemaps

### LINKED ISSUE #2362: Source maps aren't loaded correctly in Chrome Node DevTools
I ran `node --inspect-brk -r @swc/register some_script.ts` and then opened `chrome://inspect` - DevTools for Node. Source maps weren't loaded properly (was showing a transpiled JS content under a TS filename). I checked with babel and it works fine. There seems to be a difference between babel and swc that makes it break -  I used debugger to dynamically modify swc source maps and managed to see it working.

It looks like embedding full filename path in sources doesn't work in Chrome DevTools for Node. What babel does, and seems to make it work - is encodes directory path in sourceRoot and just filename in sources.

This works (babel):
```
{
  "version": 3,
  "sourceRoot": "/Users/pbadenski/workspace/project/",
  "sources": [
    "test.ts"
  ],
  ...
}
```

This doesn't (swc):
```
{
  "version": 3,
  "sources": [
    "/Users/pbadenski/workspace/project/test.ts"
  ],
  ...
}
```

kdy1:
**Description:**

**Related issue:**

 - Closes #2362


kdy1 (commented):
swc-bump:
 - swc_common

crates/swc/tests/fixture/deno/deno-10014/case1/output/index.map
```diff
@@ -6,6 +6,7 @@
     "x",
     "Error"
   ],
+  "sourceRoot": "$DIR/tests/fixture/deno/deno-10014/case1/input",
   "sources": [
     "../../input/index.ts"
```
jridgewell, line 11: This looks invalid. To resolve a source file, you take the concatenation of the `sourceRoot` (treating it always as a directory) and the individual `source[i]`. In this case, we'd get the result `"$DIR/tests/fixture/deno/deno-10014/case1/input" + "/" + "../../input/index.ts"` which resolves to `"$DIR/tests/fixture/deno/deno-10014/input/index.ts".

crates/swc_common/src/source_map.rs
```diff
@@ -1222,6 +1223,16 @@ impl SourceMap {
                 continue;
             }
 
+            if source_root.is_none() {
+                if let FileName::Real(f) = &f.name {
+                    if let Some(parent) = f.parent() {
+                        let s = parent.display().to_string();
+                        builder.set_source_root(Some(&s));
```
jridgewell, line 1230: This looks like you're pulling `sourceRoot` from the the first file that's iterated in mappings, but I think that's going to give a lot of bugs.

Do injected code helper have a synthetic source file in the mappings? Or, can SWC generate a single output from multiple inputs (Babel has a `{ 'file1.js': '...', 'file2.js': '...' }` transformer).

Further, `sourceRoot` is generally configured via options. Eg, [Babel](https://babeljs.io/docs/en/options#sourceroot), [Terser](https://github.com/terser/terser#:~:text=You%20can%20also%20specify%20sourceRoot%20property%20to%20be%20included%20in%20source%20map%3A), [Closure](https://github.com/google/closure-compiler/wiki/Flags-and-Options#:~:text=source_map_location_mapping), [esbuild](https://esbuild.github.io/api/#source-root)

jridgewell (changes requested):

crates/swc_common/src/source_map.rs
kdy1: Oh, thank you!
I'll check them

kdy1: Injected helpers are skipped, but this can be problematic for bundler

kdy1 (commented):
swc-bump:
 - swc_compiler_base --breaking

PR #6088, opened 2022-10-08
feat(es/minifier): Mark ref to fn as non-call in alias analyzer

kdy1:
**Description:**

This is follow-up PR of https://github.com/swc-project/swc/pull/6080


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

jridgewell (approved):

PR #5737, opened 2022-09-04
feat(es/minifier): Inline lazy inited vars

Austaras:
**Description:**

**BREAKING CHANGE:**

**Related issue (if exists):**
related #5559

PR #6089, opened 2022-10-08
feat(es/minifier): Inline lazy inited vars in condition

Austaras:
**Description:**
At the time of writing #5737, I thought such case would be rare. That is wrong.

**BREAKING CHANGE:**


crates/swc_ecma_minifier/src/analyzer/storage/normal.rs
```diff
@@ -44,12 +44,22 @@ impl Storage for ProgramData {
...
                     e.get_mut().inline_prevented |= var_info.inline_prevented;
 
-                    e.get_mut().ref_count += var_info.ref_count;
                     e.get_mut().cond_init |= if !inited && e.get().var_initialized {
                         true
                     } else {
                         var_info.cond_init
                     };
+                    if var_info.var_initialized {
+                        if e.get().var_initialized || e.get().ref_count > 0 {
```
Austaras, line 53: This line needs to happen before `ref_count` merge and after `cond_init` merge

kdy1:
Sorry, I rebased this as I made a mistake while trying codeball

kdy1 (dismissed):
Thank you!

---

swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/tests/benches-full/d3.js
```diff
@@ -1969,19 +1969,19 @@
         var t0, i0;
         function tween() {
             var fullname1, i, i1 = value.apply(this, arguments);
```
jridgewell, line 1971: It looks like `fullname1` is unused?

jridgewell (changes requested):
We've eliminated some assignments, but there are still the declarations.

crates/swc_ecma_minifier/tests/benches-full/d3.js
kdy1: Typically we did it in another PR because dropping variables require more analysis, but in this case, I think we can do it within this PR.

We can leverage `Finalizer` or drop variables directly when we inline I think.

Austaras: For lazily inited vars, we need some extra works to remove it in one pass, because at the time of inline we don't have the mutable reference to var declaration. I believe the most optimal way is to run DCE pass at the end but that seems to be bad for performance somehow.

Austaras: For lazily inited vars, we need some extra works to remove it in one pass, because at the time of inline we don't have the mutable reference to var declaration. I believe the most optimal way is to run DCE pass at the end but that seems to be bad for performance somehow.

kdy1: Can't we add it to the removal list of `Finalizer`?

https://github.com/swc-project/swc/blob/4f92225db78ed796fdba75328821a4914997de9e/crates/swc_ecma_minifier/src/compress/optimize/util.rs#L168

Sequential pass removes variables in a single pass by using `Finalizer`.

Austaras: Fine. But I'd prefer to do it in a seperate PR as it involves around 800 lines of change.

kdy1: Then can you rebase?

kdy1 (approved):

jridgewell (approved):

PR #6093, opened 2022-10-09
fix(es/minifier): Inline and remove vars in one pass

Austaras:
**Description:**

**BREAKING CHANGE:**


crates/swc_ecma_minifier/src/compress/optimize/inline.rs
```diff
@@ -651,31 +652,8 @@ where
...
-                    if ids.contains(&i.to_id()) {
-                        None
-                    } else {
-                        Some(v)
-                    }
-                })
                 .cloned()
             {
-                match &*value {
-                    Expr::Lit(Lit::Num(..)) => {
```
kdy1, line 666: Is this really useless?

crates/swc_ecma_minifier/src/compress/optimize/inline.rs
```diff
@@ -651,31 +652,8 @@ where
...
-                        if self.ctx.is_lhs_of_assign {
-                            return;
-                        }
-                    }
-                    Expr::Member(..) => {
-                        if self.ctx.executed_multiple_time {
```
kdy1, line 672: I remember the reason I added this code, but not sure because situation is changed quite a lot

kdy1 (commented):
I'm simply not sure

Austaras:
I'm also not sure, so I'm letting it running through CI first.

crates/swc_ecma_minifier/src/compress/optimize/inline.rs
Austaras: If it's inlined in assign LHS, then it should be reassigned so it won't be inlined at first place.

Also if it is necessary, then why do string literal not need this check?

crates/swc_ecma_minifier/src/compress/optimize/inline.rs
Austaras: We don't inline member expression at all.

crates/swc_ecma_minifier/src/compress/optimize/inline.rs
kdy1: Not sure, I think this code is wrong anyway

crates/swc_ecma_minifier/src/compress/optimize/util.rs
```diff
@@ -59,6 +59,16 @@ where
             orig_ctx,
         }
     }
+
+    pub(super) fn try_remove(&mut self, id: Id) {
```
Austaras, line 63: I don't like this solution around `ref_count` -- but there's some wired case preventing add everything inlined to removed
```js
var x,y;
x = y, y = x;
console.log(x, y, x, y)
```

crates/swc_ecma_minifier/tests/terser/compress/functions/issue_2531_2/output.js
```diff
@@ -1,5 +1,4 @@
 function outer() {
-    var value;
```
kdy1, line 2: I think the change to these tests should be considered carefully

crates/swc_ecma_minifier/src/compress/optimize/util.rs
```diff
@@ -59,6 +59,16 @@ where
...
+        if let Some(usage) = self.data.vars.get_mut(&id) {
+            usage.ref_count -= 1;
+
+            if usage.ref_count == 0 {
+                self.vars.removed.insert(id.clone());
```
kdy1, line 68: Can you add check for `self.options.unused` for this line?

kdy1 (changes requested):
I think we should respect options.


---

swc-bump:
 - swc_ecma_minifier

Austaras:
I need to rethink about this.

crates/swc_ecma_minifier/tests/terser/compress/functions/issue_2531_2/output.js
Austaras: This one is fine, issue_1606 is wrong

crates/swc_ecma_minifier/src/compress/optimize/inline.rs
```diff
@@ -149,7 +149,7 @@ where
                         ..
                     }) => false,
 
-                    Expr::Ident(id) => self
+                    Expr::Ident(id) if !id.eq_ignore_span(ident) => self
```
Austaras, line 152: This is also a dirty hack -- but at least it's good for performance.

crates/swc_ecma_ast/src/ident.rs
```diff
@@ -254,6 +254,11 @@ impl Ident {
 
         Err(buf)
     }
+
+    #[inline]
+    pub fn is_dummy(&self) -> bool {
+        self.sym == *"" && self.span.is_dummy()
```
kdy1, line 260: Can you change the symbol comparison to `self.sym == js_word!("")`?

kdy1: Comparison of `JsWord` with `js_word!("")` is a comparison of `u32` to a constant

kdy1 (changes requested):

crates/swc_ecma_minifier/src/compress/optimize/inline.rs
```diff
@@ -709,6 +687,14 @@ where
 
                 *e = *value;
 
+                if let Some(usage) = self.data.vars.get_mut(&id) {
+                    usage.ref_count -= 1;
+
+                    if usage.ref_count == 0 {
+                        self.vars.removed.insert(id.clone());
```
kdy1, line 694: Please add a `report_change!`

Austaras: This line should be removed...

kdy1 (approved):
Thanks!

crates/swc_ecma_minifier/tests/terser/compress/hoist_props/issue_2377_2/output.js
```diff
@@ -1,2 +1 @@
-var x;
```
jridgewell, line 1: This is only safe if `toplevel` is off. Otherwise, this declare a global variable that other scripts can then read without throwing.

crates/swc_ecma_minifier/src/compress/optimize/inline.rs
```diff
@@ -368,9 +368,10 @@ where
                     ident
                 );
                 self.changed = true;
+
                 self.vars
                     .vars_for_inlining
-                    .insert(ident.to_id(), init.take().into());
+                    .insert(ident.take().to_id(), init.take().into());
```
jridgewell, line 374: If I'm following correctly, the core change is here. We've identified that the variable is useless, by `take`ing the identifier, later code will recognize it as a dummy?

Awesome.

jridgewell (approved):

PR #6094, opened 2022-10-09
chore: change file extension in TypeScript example

amorriscode:
**Description:**

I was browsing the examples and notices the file extension for the TypeScript example was `*.js` so I updated them to `*.ts`.


kdy1 (approved):
Thanks!

---

swc-bump:
 - dbg-swc

PR #6097, opened 2022-10-10
feat(es/minifier): Enable `reduce_funcs` by default

kdy1:
**Description:**

This is to align with terser. There's no test change, though.
I'll merge this so we can see diff of real-world test cases when working on `reduce_fucs`.

**Related issue:**


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #6099, opened 2022-10-10
feat(es/minifier): Inline pure array literal partially

kdy1:
**Description:**

```js

function f() {
    var a = [94, 173, 190, 239];
    var b = 0;
    b |= a[0];
    b <<= 8;
    b |= a[1];
    b <<= 8;
    b |= a[2];
    b <<= 8;
    b |= a[3];
    return b;
}
console.log(f().toString(16));

```

This can be precomputed.

**Related issue:**


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

jridgewell (approved):
Can we expand this to operate on regular objects as well?

crates/swc_ecma_minifier/src/compress/optimize/inline.rs
```diff
@@ -89,8 +89,42 @@ where
...
+                && is_inline_enabled
+            {
+                if let Expr::Array(arr) = init {
+                    if arr.elems.len() < 32
+                        && arr.elems.iter().all(|e| match e {
+                            Some(ExprOrSpread { spread: None, expr }) => match &**expr {
+                                Expr::Lit(..) => true,
+                                _ => false,
+                            },
+                            _ => false,
```
jridgewell, line 118: Couldn't we still optimize `array[0]` even if `array[1]` is an expression?

jridgewell (approved):

crates/swc_ecma_minifier/src/compress/optimize/inline.rs
kdy1: We currently don't analyze accessed indexes.
And I doubt if such analysis will be useful for real-world codes.

PR #6106, opened 2022-10-10
fix(es/preset-env): Update data

kdy1:
**Description:**

**Related issue:**


kdy1 (commented):
swc-bump:
 - string_enum

PR #6112, opened 2022-10-10
feat(css/parser): improve selector AST

alexander-akait:
**Description:**

Resolve TODO + tests + improve AST

**BREAKING CHANGE:**

Yes

No


crates/swc_css_codegen/tests/fixture/values/ident/output.min.css
```diff
@@ -1 +1 @@
-:root{--fg-color:blue;--fg-\color-other:blue}.box{background-color:rebeccapurple;background-color:var(--fg-color);border-radius:10px;width:100px;height:100px}.box:hover{animation-name:rotate\ rotate;animation-duration:.7s}@keyframes rotate\ rotate{0%{transform:rotate(0)}100%{transform:rotate(360deg)}}@keyframes rotate{0%{transform:rotate(0)}100%{transform:rotate(360deg)}}@keyframes rotat\e{0%{transform:rotate(0)}100%{transform:rotate(360deg)}}*|*{color:blue}*|*{color:blue}@media screen and (min-width:0\0){body{color:red}}body{margin:-\1px}
+:root{--fg-color:blue;--fg-\color-other:blue}.box{background-color:rebeccapurple;background-color:var(--fg-color);border-radius:10px;width:100px;height:100px}.box:hover{animation-name:rotate\ rotate;animation-duration:.7s}@keyframes rotate\ rotate{0%{transform:rotate(0)}100%{transform:rotate(360deg)}}@keyframes rotate{0%{transform:rotate(0)}100%{transform:rotate(360deg)}}@keyframes rotat\e{0%{transform:rotate(0)}100%{transform:rotate(360deg)}}*|*{color:blue}\*|*{color:blue}@media screen and (min-width:0\0){body{color:red}}body{margin:-\1px}
```
alexander-akait, line 1: Original code:
```css
\*|*{color:blue}
```

According spec it is `\*` namespace (so we print it as ident). 

Chrome has bug here `\*|*{color:blue}` parsed as `*{color:blue}`, firefox doesn't affected this bug and just can't parse it.

And both wrong here, chrome should not parse this as any namespace, firefox should parse this with `\*` namespace :smile: I will report this in both browsers

kdy1 (changes requested):
There's a merge conflict

alexander-akait:
@kdy1 Yeah, I will fix it soon

alexander-akait:
@kdy1 Rebased :+1:

kdy1:
Is this necessary? I think this may make working with AST harder

alexander-akait:
@kdy1 Yep, `*` is not the `Ident`, it is the special thing (any namespace), anyway it should simplify using AST for manipulation, because now you can easy to catch all `Any` namespace (minify purposes and svg minification), but namespace is not popular in current CSS, also here a small perf improvement too

kdy1 (approved):
swc-bump:
 - swc_css_ast --breaking

PR #6113, opened 2022-10-11
feat(css/codegen): lowercase case insensivity

alexander-akait:
**Description:**

Lowercased value are better for gzip (and other compression).

We should add more then we start to parse declaration according spec (not generic parsing as now we do)

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_css_codegen

PR #6114, opened 2022-10-11
fix(swc_core): Reexport `swc_css_compat` correctly

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_core

PR #6115, opened 2022-10-11
feat(es/minifier): Ignore return value if a param of IIFE is not used

kdy1:
**Description:**

**Related issue:**


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
```diff
@@ -322,6 +296,96 @@ where
...
+            }
+        }
+
+        let mut removed = vec![];
+        let params = find_params(callee);
+        if let Some(mut params) = params {
+            // We check for parameter and argument
+            for (idx, param) in params.iter_mut().enumerate() {
+                if let Pat::Ident(param) = &mut **param {
+                    if let Some(usage) = self.data.vars.get(&param.to_id()) {
```
jridgewell, line 353: Does `data.vars` track all scopes? I'm assuming `to_id` generates a globally unique value for every binding?

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
```diff
@@ -322,6 +296,96 @@ where
         clean_params(callee);
     }
 
+    /// If a parameter is not used, we can ignore return value of the
+    /// corresponding argument.
+    pub(super) fn ignore_unused_args_of_iife(&mut self, e: &mut CallExpr) {
+        let has_spread_arg = e.args.iter().any(|v| v.spread.is_some());
+        if has_spread_arg {
+            return;
+        }
```
jridgewell, line 305: Is this necessary? I'm assuming this is because we can't appropriately track the index position of the call arg and the function param after a spread, but we could eliminate args before the spread.

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
```diff
@@ -322,6 +296,96 @@ where
...
+                    // Use `0` if it's removed.
+                    arg.expr = Number {
+                        span: arg.expr.span(),
+                        value: 0.0,
+                        raw: None,
+                    }
+                    .into();
+                }
+            } else {
+                break;
```
jridgewell, line 384: Is this actually reachable?

jridgewell (approved):
Is it possible to start adding test cases for the explicit feature under test? It seems we're testing indirectly, by updating the outputs to (extremely large) built projects. Particularly for negative cases (eg, I wanted to see that `arguments[0]` usage inside an arrow would deoptimize), but because a test outputs are already committed, I would need to scan through every file in the project to see the case.

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
```diff
@@ -322,6 +296,96 @@ where
...
+                .filter(|usage| usage.used_recursively)
+                .is_some()
+            {
+                return;
+            }
+        }
+
+        let mut removed = vec![];
+        let params = find_params(callee);
+        if let Some(mut params) = params {
```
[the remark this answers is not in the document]
kdy1: Good idea, thanks.

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
kdy1: Yes, `data.vars` is whole data of the Program and all `Id` is unique in swc.

You can read more about it at https://rustdoc.swc.rs/swc_ecma_transforms_base/fn.resolver.html and https://rustdoc.swc.rs/swc_ecma_ast/struct.Ident.html

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
kdy1: I think it would be better, will do it tomorrow.

kdy1:
I'll add small tests starting from the next PR

jridgewell (approved):

PR #6116, opened 2022-10-11
test(es/minifier): Organize terser tests

kdy1:
**Description:**

The minifier of swc is better than terser in some parts, so we don't need to match the output of terser exactly.

**Related issue:**


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #6118, opened 2022-10-11
feat(bindings): Apply `resolver` to the output of `parse` apis

kdy1:
**Description:**

This PR make `parse*()` api return AST with `resolver` applied to help plugin authors


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #6119, opened 2022-10-11
feat(dbg-swc): Remove comments from the output of `terser`

kdy1:
**Description:**

This PR removes comments by parsing and printing using swc APIs because the comments disturb looking at diffs.


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #6121, opened 2022-10-11
test(css/prefixer): verify

alexander-akait:
**Description:**

Port tests from https://github.com/postcss/autoprefixer/commit/2d286dd8b7dda067c4d5f3445f6bd1f17094202f

Currently we don't support very old `linear-gradient` (we have TODO for it, but it is really old syntax, so it is not high priority right now)

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - dbg-swc

PR #6127, opened 2022-10-12
feat(es/minifier): Inline more with sequential inliner

kdy1:
**Description:**

I wrote descriptions in the code with comments.

**Related issue:**


crates/swc_ecma_minifier/tests/single-pass/pr-6127/1/output.js
```diff
@@ -0,0 +1,9 @@
+console.log((function() {
```
kdy1, line 1: Previous input:

```js
console.log((function() {
    var b = 0;
    return b |= 94, b <<= 8, b |= 173, b <<= 8, b |= 190, b <<= 8, b |= 239;
})().toString(16));
```

Austaras: What about update expr? Something like
```js
var a = 10;
a++,++a,a--,--a
console.log(10)
```

kdy1: It should be handled in future IMO

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/tests/projects/output/react-dom-17.0.2.js
```diff
@@ -7847,7 +7847,7 @@
     }
     function unbatchedUpdates(fn, a) {
         var prevExecutionContext = executionContext;
-        executionContext &= -2, executionContext |= 8;
+        executionContext = -2;
```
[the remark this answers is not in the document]
kdy1: I think you are right

crates/swc_ecma_minifier/tests/benches-full/terser.js
```diff
@@ -1924,8 +1924,8 @@
         base && base.PROPS && (props = props.concat(base.PROPS));
         for(var code = "return function AST_" + type + "(props){ if (props) { ", i = props.length; --i >= 0;)code += "this." + props[i] + " = props." + props[i] + ";";
         const proto = base && Object.create(base.prototype);
-        (proto && proto.initialize || methods && methods.initialize) && (code += "this.initialize();"), code += "}", code += "this.flags = 0;";
-        var ctor = Function(code += "}")();
+        (proto && proto.initialize || methods && methods.initialize) && (code += "this.initialize();");
+        var ctor = Function(code = (code += "}") + "this.flags = 0;}")();
```
kdy1, line 1928: I think we can optimize this further

kdy1: I tried and concluded that it would be better to do this with another PR

kdy1 (commented):
This pr is now correct, but this increases the size in general.
I'll try handling only literals

crates/swc_ecma_minifier/tests/pass-default/pr-6127/1/output.js
```diff
@@ -0,0 +1,4 @@
+console.log((function() {
+    var b = 0;
+    return b |= 94, b <<= 8, b |= 173, b <<= 8, b |= 190, b <<= 8, b |= 239;
```
kdy1, line 3: This is not ideal

kdy1:
This was a wrong approach

PR #6128, opened 2022-10-12
fix(css/modules): Fix parsing and interference

kdy1:
**Description:**

This PR fixes parsing of css modules and fixes interference with nesting css transform.

**Related issue:**


kdy1 (commented):
swc-bump:
 - swc_atoms
 - swc_css_parser --breaking

crates/swc_css_parser/src/parser/base.rs
```diff
@@ -107,6 +107,69 @@ where
...
+                            value: CombinatorValue::Descendant,
+                        }),
+                    );
+                }
+            }
+        }
+
+        Some(nested)
+    }
 }
```
alexander-akait, line 173: It is not valid solution...

crates/swc_css_parser/src/parser/selector/mod.rs
```diff
@@ -408,6 +443,40 @@ where
...
+                    nesting_selector: None,
+                    type_selector: None,
+                    subclass_selectors: vec![SubclassSelector::PseudoClass(PseudoClassSelector {
+                        span: span!(self, span.lo),
+                        name: kind,
+                        children: None,
+                    })],
+                });
+            }
+        }
```
alexander-akait, line 478: We can't use it here, otherwise another pseudo will not work

crates/swc_css_parser/src/parser/selector/mod.rs
```diff
@@ -408,6 +443,40 @@ where
 
         let mut nesting_selector = None;
 
+        if self.config.css_modules && is!(self, ":") && peeked_is_one_of!(self, "local", "global") {
+            bump!(self);
+
+            let kind: Ident = self.parse()?;
+
+            if eat!(self, "(") {
+                let selectors = self.parse_as::<SelectorList>()?;
```
alexander-akait, line 452: Inside `local` and `global` is not `SelectorList`, it is not valid to use `a, b` inside `:local`, because every local contain only compound selector

alexander-akait (changes requested):
Give me some time to investigate it, because I really don't want to vilolate spec and have a lot of workarounds by it

kdy1:
Closing as it's fixed by https://github.com/swc-project/swc/pull/6132

PR #6131, opened 2022-10-12
fix(css/parser): AST for simple block

alexander-akait:
**Description:**

Fix TODO for future improvements

**BREAKING CHANGE:**

Yes

NO


alexander-akait:
@kdy1 Rebased, there are a lot of changes, but they are because simple block hav `TokenAndSpan` now, no big changes in rust code

kdy1:
Why is this required? I think it'll make the code slower due to bigger types, so I'm curious about the gains

kdy1 (approved):
swc-bump:
 - swc_css_ast --breaking

PR #6139, opened 2022-10-13
feat(es/minifier): Inline uninitialized vars with sequential inliner

kdy1:
**Description:**

An uninitialized variable is inlinable.

**Related issue:**


crates/swc_ecma_minifier/tests/benches-full/victory.js
```diff
@@ -25305,15 +25305,15 @@
                 if (n = (values = Float64Array.from(Object(_number_js__WEBPACK_IMPORTED_MODULE_3__.numbers)(values, valueof))).length) {
                     if ((p = +p) <= 0 || n < 2) return Object(_min_js__WEBPACK_IMPORTED_MODULE_1__.default)(values);
                     if (p >= 1) return Object(_max_js__WEBPACK_IMPORTED_MODULE_0__.default)(values);
-                    var n, i = (n - 1) * p, i0 = Math.floor(i), value0 = Object(_max_js__WEBPACK_IMPORTED_MODULE_0__.default)(Object(_quickselect_js__WEBPACK_IMPORTED_MODULE_2__.default)(values, i0).subarray(0, i0 + 1));
```
kdy1, line 25308: This seems wrong, so `reassigned()` is wrong I guess

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

kdy1 (commented):
This PR is for feature parity with terser and future optimizations

Austaras:
I don't believe seq inliner is the proper place for this. Unintialized vars should be inlined anywhere as accessing them are always side effect free. However this doesn't really matter as it's very rare to have unintialized vars in real code.

kdy1:
You are right, nice catch!
I'll change it to use inliner instead

kdy1:
I have thought about this, and I concluded this is useless

PR #6140, opened 2022-10-13
fix(css/parser): math functions in at-rules

alexander-akait:
**Description:**

Fix handle math function in `@media` and `@container` at-rules + better minificatio them.

There are potential case to improve `width: calc(calc(calc(100px + 100px)))`, I will improve it late, because no one write code in this style + fix keep `calc` inside `@supports`, beucase it can be used to check `calc` exists or not

**BREAKING CHANGE:**

No

No


crates/swc_css_parser/src/parser/value/mod.rs
```diff
@@ -2328,6 +2328,9 @@ where
             value: vec![],
         };
 
+        println!("FUCK {:?}", function);
+        println!("FUCK {:?}", self.ctx.block_contents_grammar);
```
kdy1, line 2332: Please remove this

crates/swc_css_minifier/tests/fixture/compress-calc/container/output.min.css
```diff
@@ -0,0 +1 @@
+@container(inline-size>=0px){h2{font-size:2cqw}}@container(inline-size>=0px){h2{font-size:2cqh}}@container(inline-size>=0px){h2{font-size:2cqi}}@container(inline-size>=0px){h2{font-size:2cqb}}@container(inline-size>=0px){h2{font-size:2cqmin}}@container(inline-size>=0px){h2{font-size:2cqmax}}@container(inline-size>=200px){h2{font-size:2cqw}}@container(inline-size>=calc(200px)){h2{font-size:2cqw}}
```
yisibl, line 1: It looks like `inline-size>=0px` can be further reduced to `inline-size>=0`

crates/swc_css_parser/src/parser/value/mod.rs
alexander-akait: Oh, sorry :smile:

alexander-akait:
@kdy1 Fixed

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_parser
 - swc_css_minifier

PR #6144, opened 2022-10-14
feat(css/modules): Add a lifetime to config parameter

kdy1:


kdy1 (commented):
swc-bump:
 - swc_css_modules

PR #6145, opened 2022-10-14
doc(repo): Document maintenance

kdy1:
This is to allow the other team members manage crates when I'm sleeping or busy


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #6149, opened 2022-10-14
feat(es/utils): Add indexed API to the parallel helper

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_ecma_utils

PR #6150, opened 2022-10-14
fix(es/minifier): Fix ordering issue of analyzer

kdy1:
**Description:**

This is the groundwork for a parallel analyzer. This PR fixes the ordering issue of the analyzer so that the analyzer works identically regardless of visit order. This patch contains some improvements because previously, we mixed `.and_modify()` and `.or_default()`.

**Related issue:**


kdy1:
cc @Austaras Do you think this analyzer patch is correct?

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/src/analyzer/storage/normal.rs
```diff
@@ -153,32 +153,27 @@ impl Storage for ProgramData {
...
 
-                if v.used_in_non_child_fn {
-                    v.is_fn_local = false;
-                }
+        if has_init && (v.declared || v.var_initialized) {
+            trace_op!("declare_decl(`{}`): Already declared", i);
 
-                v.var_initialized |= has_init;
-            })
-            .or_insert_with(|| VarUsageInfo {
```
kdy1, line 174: This is problematic because this closure is not called if the analyzer called `.var_or_default()` for the infection analysis

Austaras:
Looks fine to me, but why tests changed?

kdy1:
Because `or_insert_with` may not be called due to `var_or_default()` calls.
If the alias analyzer calls `var_or_default().add_infects()`, there's an entry in the variable hash map so `or_insert_with` will not be called.

See https://github.com/swc-project/swc/pull/6150#discussion_r995619126

crates/swc/tests/tsc-references/classDeclarationLoop.2.minified.js
```diff
@@ -1,9 +1,8 @@
 //// [classDeclarationLoop.ts]
 import _class_call_check from "@swc/helpers/src/_class_call_check.mjs";
 for(var _loop = function(i) {
-    var C = function C() {
+    arr.push(function C() {
         "use strict";
         _class_call_check(this, C), this.prop = i;
-    };
-    arr.push(C);
+    });
```
Austaras, line 7: I see. In this case `arr` used to be marked as `var_kind` `None`.

kdy1: Yeah and that's why it was not inlined I guess

crates/swc_ecma_minifier/tests/projects/output/react-dom-17.0.2.js
```diff
@@ -8072,8 +8072,7 @@
         0 === workInProgressRootExitStatus && (workInProgressRootExitStatus = 5);
     }
     function commitRoot(root) {
-        var renderPriorityLevel = getCurrentPriorityLevel();
-        return runWithPriority$1(99, commitRootImpl.bind(null, root, renderPriorityLevel)), null;
+        return runWithPriority$1(99, commitRootImpl.bind(null, root, getCurrentPriorityLevel())), null;
```
jridgewell, line 8075: This isn't strictly safe, but probably close enough.  We can agree that the access of `runWithPriority$1` variable should be fine, but the `commitRootImpl.bind` might have a side-effect getter. I think that's ok?

jridgewell (approved):

crates/swc_ecma_minifier/tests/projects/output/react-dom-17.0.2.js
kdy1: This is safe, because this is done by the sequential inliner.
We prove that `getCurrentPriorityLevel()` cannot modify `commitRootImpl.bind` by proving that `getCurrentPriorityLevel` does not have any direct/indirect reference to `commitRootImpl`

PR #6152, opened 2022-10-14
feat(css/parser): parse custom media

alexander-akait:
**Description:**

Parsing `@custom-media` https://www.w3.org/TR/mediaqueries-5/#custom-mq

**BREAKING CHANGE:**

Yes

No


crates/swc_css_ast/src/at_rule.rs
```diff
@@ -801,3 +804,36 @@ pub enum SizeFeatureName {
...
 }
+
+#[ast_node("ExtensionName")]
+#[derive(Eq, Hash)]
+pub struct ExtensionName {
+    pub span: Span,
+    #[cfg_attr(feature = "rkyv", with(swc_atoms::EncodeJsWord))]
+    pub value: JsWord,
+    #[cfg_attr(feature = "rkyv", with(swc_atoms::EncodeJsWord))]
+    pub raw: Option<JsWord>,
```
kdy1, line 815: Let's use `Option<Atom>` instead

kdy1 (changes requested):

crates/swc_css_ast/src/at_rule.rs
alexander-akait: @kdy1 Let's do it in another PR, we need to use it in more places, so I want to do it in one PR. What do you think?

kdy1 (approved):
I'll rebase/merge this after publishing a new version, thanks!


---

swc-bump:
 - swc_css_ast --breaking

PR #6155, opened 2022-10-14
fix(css/minifier): handle zero length

alexander-akait:
**Description:**

Compress `0px` -> `0` in at-rules

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_css_minifier

PR #6157, opened 2022-10-14
fix(quote_macros/expr): fix quote macro with template string struct

JSerFeng:
Just find out that there is a missing field `span` in Tpl when I'm using quote macro, when I use quote!("\`Hello\`" as Box<Expr>) macro expansion failed because missing field span here.

It seems that there are no tests for this macro right ? If there are, please let me know so I can add test about it.


kdy1:
Can you sign the CLA?

kdy1 (approved):
swc-bump:
 - swc_ecma_quote_macros

JSerFeng:
> Can you sign the CLA?

Done

PR #6163, opened 2022-10-15
fix(es/plugin): Disable SIMD

kdy1:
**Description:**

This PR disables SIMD for all hosts because wasmer cache is not portable.

**Related issue:**


kdy1 (commented):
swc-bump:
 - swc_plugin_runner

Issue #6154, opened 2022-10-14
Minifier removes parens around optional chain followed by member access


jridgewell:
### Describe the bug

The minifier is removing the parens in `(a?.b).c`.

The parens around `(a?.b).c` have a semantic meaning that prevents the short-circuiting behavior of later chaining. Eg, `(null?.b).c` will throw a `TypeError`, because the `.c` access is not optional.

### Input code

```typescript
(a?.b).c
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2020",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": true,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true,
        "const_to_let": true,
        "pristine_globals": true
      },
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

[...]

PR #6164, opened 2022-10-15
fix(es/fixer): Preserve parens of optional chaining expressions

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6154.


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base

crates/swc_ecma_transforms_base/src/fixer.rs
```diff
@@ -577,6 +577,18 @@ impl VisitMut for Fixer<'_> {
 
             _ => {}
         }
+
+        dbg!(&*n.obj);
```
kdy1, line 581: cc @Austaras I think AST definition is wrong.
We can't store `ParenExpr` in `OptChainBase`. Any idea?
Should we add a variant to it?

Austaras: Maybe it should be `Paren (OptChain (OptChainMember))`?

crates/swc_bundler/tests/fixture/deno-9591/output/entry.inlined.ts
```diff
@@ -242,7 +242,7 @@ let NATIVE_OS = "linux";
 const navigator = globalThis.navigator;
 if (globalThis.Deno != null) {
     NATIVE_OS = Deno.build.os;
-} else if (navigator?.appVersion?.includes?.("Win") ?? false) {
+} else if ((navigator?.appVersion)?.includes?.("Win") ?? false) {
```
jridgewell, line 245: In this very specific case `(foo?.bar)?.baz`, you can't actually observe the difference, because the chain is immediately made optional again.

crates/swc_ecma_transforms_base/src/fixer.rs
kdy1: It does not work.

```js
global.a = null;
let foo = "PASS";
a?.b[(foo = "FAIL")];
console.log(foo);
```

this is a terser test case, and it should **not** be


```js
global.a = null;
let foo = "PASS";
(a?.b)[(foo = "FAIL")];
console.log(foo);
```

Austaras: If `a?.b[(foo = "FAIL")];` is the raw input, it should be `OptChain(OptChainMember (Member))` while `(a?.b)[(foo = "FAIL")]` should be `Member(OptChain(OptChainMember))`, at least this is what babel is doing

kdy1: Thanks! I'm not sure if it's ideal, but we can avoid a breaking change for wasm plugin authors with that way 👍

Issue #6146, opened 2022-10-14
Wrong simplified expression in tagged template


iheyunfei:
### Describe the bug

I'm using `crates/swc_ecma_transforms_optimization/src/simplify/expr/mod.rs#expr_simplifier` and found this bug. 

Using  `Compress` option will reproduce the bug in the playground, but I'm not sure what `expr_simplifier` is used in the playground. 

__(true && o.f)\`\`;__    is simplify to   __o.f\`\`__   , which causes `this` pointing to a wrong object.

### Input code

```typescript
let o = {
	f() {
		assert.ok(this !== o);
	}
};
(1, o.f)``;
```

### Playground link

https://play.swc.rs/?version=1.3.7&code=H4sIAAAAAAAAA8tJLVHIV7BVqObiTNPQBFGcicXFqUUlevnZGiUZmcUKira2Cvma1lyctVy11lwahjoK%2BXppmgkJ1lwABzEzUjwAAAA%3D&config=H4sIAAAAAAAAA22TQZLjIAxF7%2BJ1FjO97APMrs9AERAOaUAuJNJxpXL3EY4NcdI76%2FlLiC9xG85khs%2FbMOlMkOsXzYn1dfgcwERNJvuJh4PIBDkdCO6HgXUegauEPv58%2FJXfAZFgFRyG6JN3cy1mME4ZiOq3JJUIianpdM74IyHnItERMYBOb7HSpHxiGCH3VIMh6IlAXXRuGfU0nT1h2qHCYNWUcXqiyXr2mKTYxixoqwxaaMBnMOwv0CVSQySJpLHWyQItHMs4VgMfSrjoUDS3WnBdbJATW94JPbFyJdEL2nX6QI9LrirvVAYuOW2aM%2Fq08%2BEbQPoMmijpCD1z4U7msFe6X1Q%2BORkiz43KhHtXCUa5nPLetRvWriGz725lsMVAvaHpxVe4uxB5CwqcE7dbMv14NqdejOcJ0LVQXNOuz%2BARqm2f9rSuwBv8J13x8zhWHjWfXhnN8YjhrUQEPqF9w2IE4yvMsr3X6ZWWZEFsBvv0o1AN25rK8BlVqK9t81kWQnLUGPDY1lceZdRpbGt5rwBtqeC2mCevVYyImM403PsT3XaKvlb1Uu0%2FcmhO6BcEAAA%3D

### Expected behavior

[...]

PR #6165, opened 2022-10-15
fix(es/minifier): Preserve `this` of tagged template literals

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6146


kdy1 (commented):
swc-bump:

iheyunfei:
@kdy1 The fix seems not complete. After the fix, 
```js
let o = {
    f() {
        assert.ok(this !== o);
    }
};
(1, o.f)``;
(true ? o.f : false)``;
(true && o.f)``;

```
is still transformed to

```js
let o = {
    f () {
        assert.ok(this !== o);
    }
};
(0, o.f)``, o.f``, o.f``;
```

The last two tagged templates were still transformed wrong.

PR #6166, opened 2022-10-15
fix(es/minifier): Don't create a numeric literal with `NaN` value

kdy1:
**Description:**


kdy1 (commented):
swc-bump:

kdy1:
It should be `Ident`, not `Number`. It's what the parser do

Austaras:
Maybe it's better to let NaN be a `Number` since we do lots of complex work around `Ident`. But that's a very minor problem.

PR #6167, opened 2022-10-15
fix(es/minifier): Don't drop op-assignments

kdy1:
**Description:**

Op-assignments should not be dropped, and we add `report_change` to those modifications.

**Related issue:**


kdy1 (commented):
swc-bump: 
 - swc_ecma_minifier

jridgewell (approved):
Can we add a test case that demonstrates what's changing? I'm assuming things like `+=` were getting improperly dropped, then we lost the assignment when merging later?

kdy1:
Closing in favor of https://github.com/swc-project/swc/pull/6169

PR #6173, opened 2022-10-16
chore(ci): Lock old issues automatically

kdy1:
**Description:**

Lefting a comment on an old issue does not help at all, so I'm adding an action to lock them.


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #6178, opened 2022-10-17
fix(es/plugin): Fix plugin on non `x86_64`

kdy1:
**Description:**

**Related issue:**


kdy1 (commented):
swc-bump:
 - swc_plugin_runner

PR #6180, opened 2022-10-17
build(es/plugin): Fix Wasm build, really

kdy1:


kdy1 (commented):
swc-bump:
 - swc_plugin_runner

PR #6183, opened 2022-10-17
feat(html/minifier): merge the same metadata elements

alexander-akait:
**Description:**

Merge the same `style` elements into one:

```css
<style>a { color: red; }</style><style>b { color: red; }</style>
```

->

```css
<style>a { color: red; }b { color: red; }</style>
```

The same possible with `script` tags, but need more complex logic and will be implemented in future

+ extra tests

Ref: https://github.com/kangax/html-minifier/issues/1126

**BREAKING CHANGE:**

Yes

No


ghostd:
We should not merge tags ("style" or other) having some attributes like "id"; but maybe this is a very rare use case.

alexander-akait:
We don't merge them when attributes are different

kdy1 (approved):
swc-bump:
 - swc_html_minifier

kdy1 (changes requested):
There's a merge conflict

alexander-akait:
@kdy1 Fixed

PR #6184, opened 2022-10-17
Add type for version export

ArnaudBarre:
`require` breaks TS inference, so we need to add an explicit type for version (currently typed to any)
Screenshot 2022-10-17 at 22 07 58


kdy1 (approved):
swc-bump:
 - dbg-swc

ArnaudBarre:
Wow hours of CI check for a single JS line change.

I'm really thankful for all the great work on SWC and that's great that Microsoft gives you free CI time, but be careful to not waste too much of it. In some parts of Europe we're not sure to have enough electricity this winter.

kdy1:
I agree that it's a waste, but auto-merge is not working for some reason

PR #6186, opened 2022-10-17
fix(css/codegen): printing pseudo class selectors

alexander-akait:
**Description:**

Fix codegen for normal and minified output

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_css_codegen

PR #6188, opened 2022-10-17
fix(html/minifier): avoid `;` at the end of js

alexander-akait:
**Description:**

Avoid `;` at the end, it is unecessary

**BREAKING CHANGE:**

No


kdy1 (approved):
swc-bump:
 - swc_html_minifier

PR #6191, opened 2022-10-18
refactor(es/minifier): Don't create invalid nodes

kdy1:
**Description:**

 - A sequential expression should have at leat two elements.

**Related issue:**


crates/swc_html_parser/tests/html5lib_tests.rs
```diff
@@ -666,11 +669,9 @@ fn html5lib_test_tree_construction(input: PathBuf) {
             let need_skip_fragment = relative_path_to_test.contains("template_dat")
                 && matches!(counter, 109 | 110 | 111);
 
-            if !need_skip_fragment {
-                if !document_fragment.is_empty() {
-                    file_stem += ".fragment_";
-                    file_stem += &document_fragment.join("").replace(' ', "_");
-                }
+            if !need_skip_fragment && !document_fragment.is_empty() {
```
kdy1, line 672: This change is for clippy warnings

kdy1 (commented):
swc-bump:

crates/swc/tests/tsc-references/ES5For-of16.2.minified.js
```diff
@@ -1,5 +1,6 @@
 //// [ES5For-of16.ts]
 for(var _i = 0, _iter = []; _i < _iter.length; _i++){
     _iter[_i];
-    for(var _i1 = 0, _iter1 = []; _i1 < _iter1.length; _i1++)_iter1[_i1];
+    for(var _i1 = 0, _iter1 = []; _i1 < _iter1.length; _i1++);
```
kdy1, line 4: These tests changes are bugfix

kdy1: Previously we generated foo = `Expr::Seq(seq)` so logic expecting identifier in `foo` didn't work correctly.

kdy1: Now we generate `foo = seq.0` so it's analyzed/processed correctly

kdy1 (commented):
swc-bump:

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
```diff
@@ -986,7 +982,9 @@ where
                     };
                     self.merge_sequences_in_seq_expr(&mut e);
 
-                    return Some(Expr::Seq(e));
+                    let mut e = Expr::Seq(e);
```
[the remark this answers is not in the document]
kdy1: We normalize instead of verifying, at the line below.
I think we can improve this in the future

crates/swc_ecma_minifier/src/compress/optimize/mod.rs
```diff
@@ -2364,21 +2454,46 @@ where
...
+
+            debug_assert_valid(s);
         }
 
         self.remove_duplicate_var_decls(s);
 
+        match s {
+            Stmt::Decl(Decl::Var(v)) if v.decls.is_empty() => {
+                s.take();
+                if self.prepend_stmts.is_empty() && self.append_stmts.is_empty() {
```
kdy1, line 2466: This is the fix for wrong appending

crates/swc_ecma_minifier/src/compress/optimize/util.rs
```diff
@@ -17,11 +17,45 @@ where
...
-                *e = *seq.exprs.take().into_iter().next().unwrap();
+        match e {
+            Expr::Seq(seq) => {
+                for e in &mut seq.exprs {
+                    self.normalize_expr(e);
+                }
+
+                if seq.exprs.len() == 1 {
+                    *e = *seq.exprs.take().into_iter().next().unwrap();
+                    self.normalize_expr(e);
```
jridgewell, line 28: Nit: isn't this already normalized from the above `for e in seq.exprs`? Can extracting it out of the sequence expression change the normalization?

jridgewell (approved):

crates/swc_ecma_minifier/src/compress/optimize/util.rs
kdy1: Oh I think you are right

kdy1: Actually it had an effect beause `for e` does not handle `len() == 1`

PR #6197, opened 2022-10-18
feat(html/minifier): improve the `remove_redundant_attributes` option

alexander-akait:
**Description:**

Due https://github.com/kangax/html-minifier/issues/1075

A lot of CSS frameworks use (and not only):
```
input[type=text] {
  /* styles */
}
```

So it is not safe to remove them by default, anyway I implemented the `safe` mode - it will remove:
- deprecated attributes (they doesn't used in CSS almost always, expect very old frameworks, which already doesn't supported)
- SVG attributes, it  is common ptractide don't use them in CSS
- attributes from metadata elements, because they will not be rendered

**BREAKING CHANGE:**

Yes

No


crates/swc_html_minifier/src/option.rs
```diff
@@ -44,6 +44,20 @@ pub enum CollapseWhitespaces {
     OnlyMetadata,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "kebab-case")]
+pub enum RemoveRedundantAttributes {
```
kdy1, line 50: Let's implement `Default` for this type instead of using `#[serde(default = "")]`.

kdy1: You can then change `#[serde(default = "...")]` to `#[serde(default)]`

kdy1 (changes requested):

alexander-akait:
@kdy1 Done, for another enum too :+1:

kdy1 (changes requested):
CI faileda

alexander-akait:
@kdy1 Fixed again :smile:

kdy1 (approved):

kdy1 (approved):
swc-bump:

PR #6198, opened 2022-10-18
refactor(css/parser): dashed ident

alexander-akait:
**Description:**

Ref: https://www.w3.org/TR/css-values-4/#typedef-dashed-ident

> The [<dashed-ident>](https://www.w3.org/TR/css-values-4/#typedef-dashed-ident) production is a [<custom-ident>](https://www.w3.org/TR/css-values-4/#identifier-value), with all the case-sensitivity that implies, with the additional restriction that it must start with two dashes (U+002D HYPHEN-MINUS).

Also don't store `--` because it should be the first two characters + small perf improvement

**BREAKING CHANGE:**

Yes

No


crates/swc_css_codegen/src/lib.rs
```diff
@@ -1140,7 +1140,7 @@ where
                         })) => false,
                         _ => !self.config.minify,
                     },
-                    ComponentValue::Ident(_) => match next {
+                    ComponentValue::Ident(_) | ComponentValue::DashedIdent(_) => match next {
```
alexander-akait, line 1143: Fix

kdy1 (approved):
swc-bump:
 - swc_atoms
 - swc_css_ast --breaking

PR #6199, opened 2022-10-18
fix(css/minifier): don't touch `@supports`

alexander-akait:
**Description:**

Don't tocuh `@supports`, because it can be used for check a support of `calc`, so

```css
// Will not work in old browser
@supports (width: calc(100px + 100px)) {
    div {
        background: red;
    }
}
```

will be

```css
// Will work in odl browser
@supports (width: 200px) {
   div {
        background: red;
    }
}
```

But some styles can be used for old browser (and verca vise), this is definitely not the expected behavior

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_css_minifier

PR #6200, opened 2022-10-19
fix(html/minifier): bug with merging and removing metadata elements

alexander-akait:
**Description:**

Small bugfixes + tests

**BREAKING CHANGE:**

No

No


crates/swc_html_minifier/src/lib.rs
```diff
@@ -11,6 +11,7 @@ use swc_common::{
     collections::AHashMap, comments::SingleThreadedComments, sync::Lrc, EqIgnoreSpan, FileName,
     FilePathMapping, Mark, SourceMap, DUMMY_SP,
 };
+use swc_ecma_transforms_base::perf::Items;
```
kdy1, line 14: Why is the imported?

alexander-akait: auto-insertion by IDE :smile: Not sure how it happens, Fixed

PR #6204, opened 2022-10-19
feat(es/minifier): Mark more expressions as pure

kdy1:
**Description:**

**Related issue:**

 - https://github.com/swc-project/swc/issues/5559


crates/swc_ecma_minifier/src/compress/optimize/iife.rs
```diff
@@ -301,7 +301,7 @@ where
     /// If a parameter is not used, we can ignore return value of the
     /// corresponding argument.
     pub(super) fn ignore_unused_args_of_iife(&mut self, e: &mut CallExpr) {
-        if !self.options.unused {
+        if !self.options.unused && !self.options.reduce_vars {
```
kdy1, line 304: I changed this because of https://github.com/swc-project/swc/blob/a0b0fd3e0d8a1084a0399e8b0345d66cfe835b0f/crates/swc_ecma_minifier/tests/terser/compress/drop_unused/chained_3/input.js#L1-L8

crates/swc/tests/tsc-references/templateStringWithEmbeddedNewOperator.2.minified.js
```diff
@@ -1,2 +1,2 @@
 //// [templateStringWithEmbeddedNewOperator.ts]
-"abc".concat(new String("Hi"), "def");
+new String("Hi");
```
jridgewell, line 2: Probably isn't important with `String`, but it's common for `Map`/`WeakMap` to be used. Are we eliminating built in constructors?

crates/swc_ecma_utils/src/lib.rs
```diff
@@ -1263,6 +1267,17 @@ pub trait ExprExt {
                             sym: js_word!("Math"),
                             ..
                         }) => span.ctxt == SyntaxContext::empty(),
+
+                        // Some methods of string are pure
```
jridgewell, line 1271: I think we could simplify this entirely, all methods should be pure. Same for methods on the other primitives?

jridgewell (approved):

crates/swc/tests/tsc-references/templateStringWithEmbeddedNewOperator.2.minified.js
kdy1: Yes, but we are not dropping `String`. I think we should, though.

https://github.com/swc-project/swc/blob/7d5b5444580ab334d822bad2f78a741760f0dd61/crates/swc_ecma_minifier/src/compress/pure/misc.rs#L1096-L1100

crates/swc_ecma_utils/src/lib.rs
kdy1: Oh, you are right. NIce catch and thanks!

kdy1: Ah, I missed one point.
`String.replace(/asd/, ()=>{})` can have side effects

PR #5576, opened 2022-08-21
fix(es/compat): Destructuring assignments and updates of super prop

magic-akari:
This PR fix the following code.

```JavaScript 
class B {
    static a = 1;
}

class C extends B {
    static z8 = [super.a] = [0];
    static z14 = ++super.a;
    static z15 = --super.a;
    static z16 = ++super["a"];
}

console.log(B.a, C.a, C.z14, C.z15, C.z16);

```

PR #6207, opened 2022-10-19
fix(es/compat): Super method call in loose mode

### LINKED ISSUE #5936: super.method() in extended class in loose mode generates code with infinite recursive loop
### Describe the bug

The playground output, when executed, logs "hola" until the call stack size is exceeded. This does not happen if you turn loose mode off. The bug seems to happen with many browser targets, so it's hard to pin down exactly which transform is problematic here.

### Input code

```typescript
class Superclass {
  doStuff() {
  }
}

class Subclass extends Superclass {
  doStuff() {
    console.log('hola');
    super.doStuff();
  }
}

new Subclass().doStuff()
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true,
      "importMeta": true
    },
    "transform": {
      "react": {
        "refresh": true
      }
    },
    "target": "es2022",
    "loose": true,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true,
  "env": {
    "targets": ""
  }
}
```

### Playground link

https://play.swc.rs/?version=1.3.2&code=H4sIAAAAAAAAA0vOSSwuVgguLUgtSgYzq7kUFFLyg0tK09I0NMG8Wq5aLq5kqLokCCO1oiQ1L4WARgWF5Py84vycVL2c%2FHQN9Yz8nER1TWuwRDFInx5cuTXMmrzUcrglGpoIBQC1OJKqpwAAAA%3D%3D&config=H4sIAAAAAAAAA0WOOw7DMAiG78KcocrQIXfIIZCLU1d%2BCUjVKPLdazuuusH%2F%2BOCElxhYTsjIQtwmOaLiBxbQI5MYdllhApUqKe80gQs5sa6keCmluoxRbOLQAExo9BoskzxHqsWQN6oWkMy3ea5Yn5LQDxxcdPZoTZNCrlWBxaKXZmHcPI21oUJ67E04%2B5sdeYfyZ4yek3UErxMU373TH6l4gFK%2B2VlBQgQBAAA%3D

### Expected behavior

I would expect the output code to log `hola` once and then exit. `super.doStuff()` should not cause recursion.

### Actual behavior

`super.doStuff()` seems to call itself.

### Version

1.3.2

### Additional context

_No response_

Austaras:
**Description:**

Seems like to be caused by #5576 cc @magic-akari 

**BREAKING CHANGE:**

closes #5936

### DIFF
--- a/crates/swc_ecma_transforms_classes/src/super_field.rs
+++ b/crates/swc_ecma_transforms_classes/src/super_field.rs
@@ -353,18 +353,39 @@ impl<'a> SuperFieldAccessFolder<'a> {
     }
 
     fn super_to_get_call(&mut self, super_token: Span, prop: SuperProp) -> Expr {
-        let proto_arg = self.proto_arg();
+        if self.constant_super {
+            Expr::Member(MemberExpr {
+                span: super_token,
+                obj: Box::new({
+                    let name = self.super_class.clone().unwrap_or_else(|| {
+                        quote_ident!(if self.is_static { "Function" } else { "Object" })
+                    });
+                    // in static default super class is Function.prototype
+                    if self.is_static && self.super_class.is_some() {
+                        Expr::Ident(name)
+                    } else {
+                        name.make_member(quote_ident!("prototype"))
+                    }
+                }),
+                prop: match prop {
+                    SuperProp::Ident(i) => MemberProp::Ident(i),
+                    SuperProp::Computed(c) => MemberProp::Computed(c),
+                },
+            })
+        } else {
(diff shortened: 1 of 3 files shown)


kdy1:
Can you add a fixture test for this case?

kdy1 (approved):
swc-bump:
 - swc_node_base
 - swc_ecma_transforms_compat

PR #6211, opened 2022-10-19
fix(html/ast): Fix compilation

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_html_ast

PR #6213, opened 2022-10-20
perf(html/parser): small improvement

alexander-akait:
**Description:**

Missing this in the previously PR

Avoid cloning when we are working with active formatting elements

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_html_parser

Issue #5068, opened 2022-06-29
top-level `eval` breaks the variable renamer


chentsulin:
### Describe the bug

When using `eval()` with styled-components, it generates duplicated `_templateObject` functions.

### Input code

```typescript
import styled from 'styled-components';

const C1 = styled.div`
  position: absolute;
`;

const C2 = styled.div`
  position: absolute;
`;

const debug = eval('');
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.2.207&code=H4sIAAAAAAAAA5XLuwqAMAyF4T1Pka06KOhocfJFvDRKoTZiYsG3V1Bwdjz85%2FPrxrui6BnI4bzziuYZxcR3ixRVjAWYOIpiV2H7nkvnUw%2BIG4tXz7HBYRQOh5KF%2FgP1T%2BBoPJbbUBpCZkxu4QJigxVtowAAAA%3D%3D&config=H4sIAAAAAAAAA0WOSw7DIAxE7%2BJ1tu2CO%2FQQiDoRFT95HKkIcfcCSpXdaGb8PI0%2BcGQaFStgmQo1qf2SIa2F4cQXpY0Uw1I5uQ9t5WAdDcZjRCFnMJndBvBG0Se%2F1wlyORZh4I5sOsK%2F2Qco5vc5jbZ%2BLeCT%2Bs247jxeV3EN%2BAFJDkBftAAAAA%3D%3D

### Expected behavior

It should not output duplicated function name: `_templateObject`.

### Actual behavior

It generates duplicated `_templateObject`:

```js
function _taggedTemplateLiteral(strings, raw) {
    if (!raw) {
        raw = strings.slice(0);
    }
    return Object.freeze(Object.defineProperties(strings, {
        raw: {
            value: Object.freeze(raw)
        }
    }));
}
function _templateObject() {
    var data = _taggedTemplateLiteral([
        "\n  position: absolute;\n"
    ]);
    _templateObject = function _templateObject() {
        return data;
    };
    return data;
}
function _templateObject() {
    var data = _taggedTemplateLiteral([
        "\n  position: absolute;\n"
    ]);
[...]

PR #6215, opened 2022-10-20
fix(es/minifier): Fix detection of direct `eval`

kdy1:
**Description:**

The existence of `eval` make the name mangler abort, but partial try logic causes minifier to generate an invalid code.

This is what I get if I disable eval detection.

image

With eval detection, I get 

```js

                     function r(r) {
                        var o = e[r];
                        if (void 0 !== o) return o.exports;
                        var n = e[r] = {
                            exports: {}
                        };
                        return t[r](n, n.exports, r), n.exports;
                    }
```

which is wrong

**Related issue:**

 - https://github.com/vercel/next.js/discussions/30237#discussioncomment-3920682

---

# Investigation

image

## Conclusion

`eval` breaks the name mangler.


crates/swc_ecma_minifier/tests/terser/compress/issue_973/this_binding_collapse_vars/output.mangleOnly.js
```diff
@@ -1,6 +1,6 @@
-var c = a;
-c();
-var d = a.b;
-d();
-var e = eval;
```
kdy1, line 5: This is an indirect eval, so it cannot use `c` nor `b`

crates/swc_ecma_minifier/tests/terser/compress/issue_973/this_binding_side_effects/output.js
```diff
@@ -1,11 +1,11 @@
...
     (0, foo.bar)();
     (0, eval)("console.log(foo);");
 })();
-(function (foo) {
-    var eval = console;
+(function(foo) {
+    var eval1 = console;
     foo();
     (0, foo.bar)();
-    (0, eval)("console.log(foo);");
```
kdy1, line 10: This is also an indirect eval, so we can mangle names

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/tests/fixture/next/feedback-3/579-dcac359116b2707c/output.js
```diff
@@ -0,0 +1,2415 @@
...
+                            t.exports = r(9720);
+                        }
+                    }, e = {};
+                    function o(r) {
+                        var n = e[r];
+                        if (void 0 !== n) return n.exports;
+                        var i = e[r] = {
+                            exports: {}
+                        };
+                        return t[r](i, i.exports, o), i.exports;
```
kdy1, line 1253: This is the fix

Smrtnyk:
Does this also fix https://github.com/swc-project/swc/issues/5068?

kdy1:
Actually this is a workaround and we should fix #5068 ideally

crates/swc_ecma_transforms_base/src/rename/eval.rs
```diff
@@ -31,9 +31,17 @@ impl Visit for EvalFinder {
...
 
-    fn visit_ident(&mut self, i: &Ident) {
-        if i.sym == js_word!("eval") {
-            self.found = true;
+    fn visit_callee(&mut self, c: &Callee) {
+        c.visit_children_with(self);
+
+        if let Callee::Expr(e) = c {
+            if let Expr::Ident(Ident {
+                sym: js_word!("eval"),
```
jridgewell, line 39: Not sure how you represent optional chains, but `eval?.()` is not a direct eval either.

crates/swc_ecma_minifier/tests/terser/compress/issue_976/eval_unused/output.mangleOnly.js
```diff
@@ -1,5 +1,5 @@
-function f1(a, eval, c, d, e) {
-    return a("c") + eval;
+function f1(n, eval, c, r, t) {
+    return n("c") + eval;
```
jridgewell, line 2: Nit: this `eval` can be renamed.

jridgewell (approved):

crates/swc_ecma_transforms_base/src/rename/eval.rs
kdy1: image

Nice catch, thank you!

kdy1: Oh it was `OptCall` which does not contain `Callee`

PR #6222, opened 2022-10-20
test(css/minifier): catch error in minifier

alexander-akait:
**Description:**

Catch errors to avoid crate broken tests cases

**BREAKING CHANGE:**

No

No


crates/swc_css_minifier/tests/fixture/compress-calc/ignore-selectors/input.css
```diff
@@ -1,5 +1,3 @@
 div[data-size="calc(3*3)"] {
-}
-
-div:nth-child(2n + calc(3*3)) {
```
alexander-akait, line 4: `calc` can't be used here, it is not allowed by spec here, so I remove this tests, because it is list of component values and we even parser this as `Calc`

crates/swc_css_minifier/tests/fixture/compress-color/input.css
```diff
@@ -194,9 +194,9 @@
     color: hwb(120 20% 30%);
     color: hwb(90deg 50% 50%);
     color: hwb(90deg 50% 50%);
-    color: hwb(90deg 50% 50%, 0.2);
+    color: hwb(90deg 50% 50% / 0.2);
     color: hwb(90 50% 50%);
-    color: hwb(90 50% 50% 0.2);
+    color: hwb(90 50% 50% / 0.2);
```
alexander-akait, line 199: Invalid color syntax, I fix them

kdy1 (approved):
swc-bump:
 - dbg-swc

PR #6224, opened 2022-10-21
refactor(css/parser): new approach

alexander-akait:
**Description:**

New CSS parser approach

**BREAKING CHANGE:**

No, but AST was improved in some cases

By spec we should get `ComponentValue` and parse list of components values, but it is more slow than using tokens, so under the hood our parser uses only token, but if we can't apply required grammar we convert tokens to lost of component values to be align with spec


crates/swc_css_parser/tests/recovery/at-rule/extra-semi/output.json
```diff
@@ -2,7 +2,7 @@
   "type": "Stylesheet",
```
alexander-akait, line 2: Changes expected, because now we output list of component values when we can't parser (as in spec)

crates/swc_css_parser/tests/recovery/at-rule/media/wrong-stylesheet/output.json
```diff
@@ -72,135 +72,118 @@
...
-                  "end": 25,
-                  "ctxt": 0
-                },
-                "token": {
-                  "Ident": {
-                    "value": "color",
-                    "raw": "color"
-                  }
-                }
+            "prelude": {
```
alexander-akait, line 81: Expected too due the reason above

alexander-akait:
Tests passed +  fix many cases with invalid AST + more tests (no regressions), WIP on perf

kdy1 (approved):
swc-bump: 
 - swc_css_parser

PR #6236, opened 2022-10-23
fix: change process_js_with_custom_pass api

JSerFeng:
## Fix

Options missing unresolved_mark field

## Breaking Change

It's a breaking change I know but the usage of current implementation is hard to use, or maybe I did it wrong.
It's very easily to encounter lifetime limitation.

```rust
compiler.process_js_with_custom_pass(..., |_, comments| { build_pass(comments) });
-----------------------------------------------^^^^^^^ compile error
```

I believe this is a common pattern, `build_pass` should hold the comments' lifetime, but borrow checker would not allow this.

So I think maybe pass `comments` to it should be a better idea.

If anything wrong with my point of view, please tell me

### DIFF
--- a/crates/swc/src/config/mod.rs
+++ b/crates/swc/src/config/mod.rs
@@ -179,6 +179,9 @@ pub struct Options {
     #[serde(skip_deserializing, default)]
     pub top_level_mark: Option<Mark>,
 
+    #[serde(skip_deserializing, default)]
+    pub unresolved_mark: Option<Mark>,
+
     #[cfg(not(target_arch = "wasm32"))]
     #[serde(default = "default_cwd")]
     pub cwd: PathBuf,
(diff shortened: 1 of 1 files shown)


kdy1 (changes requested):
Comments are cheap to clone, so you should clone it instead

JSerFeng:
Yes, but
If I have cloned one called C_A, and the one used by process_js_with_custom_pass called C_B
I made some changes in C_A, but the actual used one is C_B, and codegen will use C_B instead of C_A right?

kdy1:
`Clone` of Comments does not clone data

JSerFeng:
> `Clone` of Comments does not clone data

Oh, thanks, I will revert my changes on that, but keep change on unresolved_mark

kdy1 (approved):
---

swc-bump:
 - swc

PR #6240, opened 2022-10-24
fix(css/parser): Add more error recovery

kdy1:
**Description:**

cc @alexander-akait I did this because this is very important


kdy1 (commented):
swc-bump:
 - swc_css_parser

PR #6248, opened 2022-10-25
feat(es/minifier): Support more statements in seqential inliner

kdy1:
**Description:**

We now inline into the discriminant of a switch statement and into the initializer of for/for-in/for-of statements.

**Related issue:**


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

jridgewell (approved):

PR #6257, opened 2022-10-26
fix(css/parser): Fix parsing of math expression in `hsl`

kdy1:
**Description:**


kdy1:
cc @alexander-akait Can you take a look?

alexander-akait:
@kdy1 Yeah, we have bugs with `var(...)`/`env()` parsing, looking at this

PR #6262, opened 2022-10-26
test(css/parser): many tests after refactor

alexander-akait:
**Description:**

Now we parse without crash non standard at-rules and rules

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - dbg-swc

PR #6270, opened 2022-10-27
feat(es/plugin): Add AST query API

kdy1:
**Description:**

Typically, finding an ast node involves 3 ~ 8 levels of nested pattern matching, but I think it can be reduced with this.

 - Basically, it consists of a pattern matcher and an operator
 - Pattern matcher can be written using query! macro
 - $n is one capture group.
 - $n can capture an expression, pattern, or a statement
 - The operator is a closure (need more thinking)
 - Operators are called once for each target node
 - Operators are called with the target AST node of the query by default and called with captured node if there are one or more capturing ($n).

**Related issue:**


kdy1:
Closing as stale

PR #6276, opened 2022-10-27
Remove mut from build_source_map_from

jridgewell:
**Description:**

The `mappings` vector was required to be mutable, but the value is never mutated.

**BREAKING CHANGE:**

I don't believe so. Any code that passed in a mutable reference will still work without any changes.

### DIFF
--- a/crates/swc/src/lib.rs
+++ b/crates/swc/src/lib.rs
@@ -498,7 +498,7 @@ impl Compiler {
 
                         self.cm
                             .build_source_map_with_config(
-                                &mut src_map_buf,
+                                &src_map_buf,
                                 orig,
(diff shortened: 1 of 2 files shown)


kdy1:
Can you sign the CLA? Honestly, I don't care how people use swc, but I configured it to avoid processes like https://github.com/swc-project/plugins/pull/99 or https://github.com/swc-project/plugins/pull/98

kdy1:
Oh it was bug

kdy1 (dismissed):
swc-bump:
 - swc_common

kdy1:
clippy failed. Seems like it want `&[Foo]`

kdy1 (approved):

PR #6281, opened 2022-10-28
refactor(css/parser): many fixes

alexander-akait:
**Description:**

Big refactor:
1. Refactor logic according spec https://www.w3.org/TR/css-syntax-3/#parser-algorithms
2. Don't try to parse value in declaration which previously has error, in this case we will output only one error (on the porblem places) and store the list of component values (that is why we reduced number of useless errors)
3. Better AST for custom properties
4. Fix bug with span when it contains spaces after `!important`
5. Fix parsing `!important`
6. A lot of small fixes
7. Better recovery mode

**BREAKING CHANGE:**

No

No


crates/swc_css_parser/tests/recovery/bad-url-token/double-quotes/output.json
```diff
@@ -88,49 +88,23 @@
         },
         "value": [
           {
-            "type": "ListOfComponentValues",
+            "type": "Declaration",
```
alexander-akait, line 91: Example of the better recovery mode - now we can parse it as declaration (due logic in spec)

kdy1 (approved):
swc-bump:
 - swc_css_parser

kdy1 (changes requested):
clippy failed

alexander-akait:
Yeah WIP on small regressions :+1:

kdy1 (approved):

PR #6272, opened 2022-10-27
fix(es/minifier): Don't ignore nullish coalescing

Austaras:
**Description:**

**BREAKING CHANGE:**

**Related issue (if exists):**

PR #6282, opened 2022-10-28
test(es/minifier): Add a test for `??` operator

kdy1:
**Description:**

I'm still investigating.

**Related issue:**

 - https://github.com/vercel/next.js/issues/41992.

---

# Investigation

```rust
                        &JsMinifyOptions {
                            compress: BoolOrDataConfig::from_bool(false),
                            mangle: BoolOrDataConfig::from_bool(false),
                            ..self.opts.clone()
                        }
```

works, while

```rust
                        &JsMinifyOptions {
                                defaults: false,
                                ..Default::default()
                            }
                            .into(),
                            mangle: BoolOrDataConfig::from_bool(false),
                            ..self.opts.clone()
                        },

```

does not

----

I found the cause.

image

is compiled as 

image

so we should preserve `??`


crates/swc_ecma_minifier/tests/fixture/next/41992/output.js
```diff
@@ -0,0 +1,184 @@
...
+                        callbackVisible: !1,
+                        parametersVisible: !0
+                    };
+                }
+                render() {
+                    var e;
+                    let { onTryoutClick: t , parameters: r , allowTryItOut: a , tryItOutEnabled: n , specPath: l , fn: s , getComponent: o , getConfigs: i , specSelectors: u , specActions: d , pathMethod: g , oas3Actions: v , oas3Selectors: E , operation: b  } = this.props;
+                    const S = o("parameterRow"), _ = o("TryItOutButton"), w = o("contentType"), C = o("Callbacks", !0), x = o("RequestBody", !0), A = n && a, I = u.isOAS3(), R = b.get("requestBody"), N = (0, p.default)(e = (0, ft.default)((0, p.default)(r).call(r, (e, t)=>{
+                        const r = t.get("in");
+                        return e[r] ?? (e[r] = []), e[r].push(t), e;
```
kdy1, line 44: This was previously `e[r], e[r] = []`

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

jridgewell (dismissed):

jridgewell:
Is it possible to add a reduced test case?

crates/swc_ecma_minifier/tests/fixture/next/41992/1/output.js
```diff
@@ -0,0 +1,184 @@
...
+                    }))));
+                }
+            }
+            (0, W.default)(ht, "defaultProps", {
+                onTryoutClick: Function.prototype,
+                onCancelClick: Function.prototype,
+                tryItOutEnabled: !1,
+                allowTryItOut: !0,
+                onChangeKey: [],
+                specPath: []
```
kdy1, line 180: @jridgewell Do you think it's better to remove this file?

jridgewell: Yes. If this file changes in the future, there's very little chance I'll spot the previously buggy code again.

kdy1: Good idea, I'll do it for other patches too

Austaras:
Has this already been fixed by #6272?

kdy1:
@Austaras No, as we have removing logic in more places

Austaras:
No, there's alreay a `may_short_circuit` in L668, and I've testes the input on main branch which generates identical output.

Austaras:
The `ignore_return_value` is really too long to parse for human...

kdy1:
Oh, if then it's possible that I was using a old base branch.

kdy1:
You are right, this was fixed already and I was using an old base

PR #6283, opened 2022-10-28
fix(es/minifier): Don't inline regex for IIFEs

### LINKED ISSUE #6279: Avoid inlining regex in loop conditional
### Describe the bug

SWC compression will currently inline a regex, even if doing so might change the behavior for global (/g) and sticky(/y) regular expressions. When these flags are included, the regex will [keep an internal state](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#description) of the `lastIndex` matched. By inlining these regex in certain scenarios, such as a while loop condition, you can cause the internal state to be reset each time the code is run

### Input code

```typescript
function run(str, r) {
  let m
  while(m = r.exec(str)) {
    console.log(m)
  }
}
run('abcda', /a/g)
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "decorators": true
    },
    "keepClassNames": true,
    "target": "es2020",
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": true,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true,
        "const_to_let": true,
        "pristine_globals": true
      },
      "mangle": false
    },
    "loose": false
  },
  "module": {
    "type": "es6"
  },
  "minify": false
}
```

### Playground link

https://play.swc.rs/?version=1.3.11&code=H4sIAAAAAAAAAyWMSwqAMBDF9j3F29kB0RN4mDqOtdAP9IOC9O4qrgIhZG%2BRq0sRuUVdah6RCbcCvFSEl%2BfhvOiABXmSS%2FiL6E8ATrEkL5NPVgd6VVddfafBrLyZYcRsZksPt5hAeGUAAAA%3D&config=H4sIAAAAAAAAA22TsZLjIAyG34U6xc4WV6S9mS3vFRgCwmHPRh5JZOPJ5N1P2Am2k%2BvQhySkX%2BJmvtmb482MjhionnjK4q7maGQagT2lUczBCCuKrmc4mAAeyQkSm6NQgfvB%2FAUYf%2FeO%2BY8b4IE1yFEHoqmAPz8%2BPzTNkHKKU33G4zASMNezupUBsnB7whHhT8tzQuzB5TfbOrYpC3RAa6jHvncjg704ahH1NUeJMe9QEQh2JBw3NIckCbMme7IALliPARpIBF7SZe00aA51yayFrTpVGOBUuq5Ku3jCxfXFScsF11kGfbHFnTGx2Fgyv6BdpQtamnx4pWgJpFB%2B%2Bnxjyjsd6pysr4PKy6AekTOPOoe9Z%2FyPV8pRhyhToz1uqsrQaXM2pdg6rFUDSVrVIgjFQ%2B3Qr8kfcNcQpwAWYlS1WzD%2FJPHnzZLpmmJspqrm4jqDxbTPfdrTugJv8Eurku04Hnxwcn5lPA0n7N9SDCBnDG9YhRB8haTbex1fackBVGYIm4vC1WxrqsMXtH39X0%2BddSE0xnY9ntr66uccXO7aWt7ngfHGHDCUen2bpZx%2F6y9zX%2F%2Fq4vcPjvcwryoEAAA%3D

### Expected behavior

The regex should not be inlined

```
!function(str, r) {
    let m;
    for(; m = r.exec(str);)console.log(m);
}('abcda', /a/g);
```

### Actual behavior

The regex is inlined, which can lead to an infinite loop

```
!function(str, r) {
    let m;
    for(; m = /a/g.exec(str);)console.log(m);
}('abcda', 0);
```

### Version

1.3.11

### Additional context

_No response_

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6279.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
```diff
@@ -200,6 +200,7 @@ where
 
                         if let Some(arg) = arg {
                             match &**arg {
+                                Expr::Lit(Lit::Regex(..)) => continue,
```
jridgewell, line 203: This likely means that any non-literal expression is unsafe. Eg,

```js
const r = new Regex('a', 'g');
function run(str, r) {
    let m
    while (m = r.exec(str)) {
        console.log(m)
    }
}
run('abcda', r)

**Review by jridgewell — CHANGES_REQUESTED** [review]

File: crates/swc_ecma_minifier/src/compress/optimize/iife.rs (hunk 1 above)
Comment (kdy1) [reply crates/swc_ecma_minifier/src/compress/optimize/iife.rs:203]: `_ => continue,` below will catch it and do same thing (`continue`

Comment (kdy1) [reply crates/swc_ecma_minifier/src/compress/optimize/iife.rs:203]: I'll add a test anyway

Comment (jridgewell) [reply crates/swc_ecma_minifier/src/compress/optimize/iife.rs:203]: Perfect, I didn't notice it.

Comment (kdy1) [reply crates/swc_ecma_minifier/src/compress/optimize/iife.rs:203]: I added one

File: crates/swc_ecma_minifier/src/compress/optimize/iife.rs [hunk 2]
```diff
@@ -241,6 +242,7 @@ where
                                     }
 
                                     match &*arg.expr {
+                                        Expr::Lit(Lit::Regex(..)) => false,
```
jridgewell, line 245: Can you add a test case for this? I initially thought that it should return true, but the regex literal should be pure so it can be dropped.

jridgewell (dismissed):

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
kdy1: Oh you are right, I'll revert this change

kdy1: Thank you so much

crates/swc_ecma_minifier/tests/fixture/issues/6279/2/input.js
```diff
@@ -0,0 +1,8 @@
+const r = new Regex('a', 'g');
```
jridgewell, line 1:
```suggestion
const r = new RegExp('a', 'g');
```

crates/swc_ecma_minifier/tests/fixture/issues/6279/2/output.js
```diff
@@ -0,0 +1,5 @@
+const r = new Regex('a', 'g');
```
jridgewell, line 1:
```suggestion
const r = new RegExp('a', 'g');
```

jridgewell (dismissed):

PR #6289, opened 2022-10-28
Support trailing commas in swcrc

johanholmerin:
**Description:**

As request in https://github.com/vercel/turbo/pull/2429, switches from json_comments to jsonc-parser to align dependencies with Turbopack. Used the same options as in Turbopack, which adds support for trailing commas but still requires properties to be quoted.

**BREAKING CHANGE:**

No

No

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1370,10 +1370,13 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


kdy1 (approved):
swc-bump:
 - swc

PR #6290, opened 2022-10-29
refactor: process_js_with_custom_pass should accept comments as paramter

JSerFeng:
**Description:**
[`process_js_with_custom_pass()`](https://github.com/swc-project/swc/blob/main/crates/swc/src/lib.rs#L851) should accept `comments` as parameter instead of using it internally, as user may pass an already parsed AST, and parse also need `comments` as parameter, so those `comments` should be the same.

**Breaking Change**
It's a breaking change, and I remember that next.js is using this API.
Also bindings/binding_core_wasm is using too.
Do I need to edit CHANGELOG, or if it can be done by `swc-bot`

### DIFF
--- a/crates/swc/src/lib.rs
+++ b/crates/swc/src/lib.rs
@@ -851,15 +851,15 @@ impl Compiler {
         program: Option<Program>,
         handler: &Handler,
         opts: &Options,
-        custom_before_pass: impl FnOnce(&Program, &SingleThreadedComments) -> P1,
-        custom_after_pass: impl FnOnce(&Program, &SingleThreadedComments) -> P2,
+        comments: SingleThreadedComments,
+        custom_before_pass: impl FnOnce(&Program) -> P1,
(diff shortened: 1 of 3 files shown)


crates/swc/src/lib.rs
```diff
@@ -851,24 +851,24 @@ impl Compiler {
         program: Option<Program>,
         handler: &Handler,
         opts: &Options,
-        custom_before_pass: impl FnOnce(&Program, &SingleThreadedComments) -> P1,
-        custom_after_pass: impl FnOnce(&Program, &SingleThreadedComments) -> P2,
+        comments: Option<&SingleThreadedComments>,
```
kdy1, line 854: Please make this non-optional, and take ownership

JSerFeng: Ok, modified

kdy1 (approved):
swc-bump:
 - swc

PR #6337, opened 2022-11-02
feat(css/parser): new nesting spec

alexander-akait:
**Description:**

Ref: https://drafts.csswg.org/css-nesting/#nesting

**BREAKING CHANGE:**

Yes, qualified rule accept relative selector list now when it is nested

No


alexander-akait:
WIP on refactor our compat logic

alexander-akait:
@kdy1 Ready for review :+1:

crates/swc_css_minifier/src/compressor/rules.rs
```diff
@@ -197,26 +222,50 @@ impl Compressor {
...
+                (
+                    QualifiedRulePrelude::SelectorList(prev_selector_list),
+                    QualifiedRulePrelude::SelectorList(current_selector_list),
+                ) => {
+                    let selector_list =
+                        self.merge_selector_list(prev_selector_list, current_selector_list);
+                    let mut qualified_rule = QualifiedRule {
+                        span: Span::new(
+                            left.span.span_lo(),
+                            right.span.span_lo(),
```
kdy1, line 235: I think this should be `span_hi()`

crates/swc_css_minifier/src/compressor/rules.rs
```diff
@@ -155,6 +155,21 @@ impl Compressor {
         }
     }
 
+    fn merge_relative_selector_list(
+        &self,
+        left: &RelativeSelectorList,
```
kdy1, line 160: Can we take ownership here?
We are unconditionally cloning.

alexander-akait: @kdy1 I think no because we have `captured outer variable`, but I have ideas how we can rewrite it, can we postpone it for future PRs

kdy1: Yes, I think so

alexander-akait:
@kdy1 Ready to review again, span fixed

kdy1 (approved):
swc-bump:
 - swc_css_ast

PR #6338, opened 2022-11-03
chore(bindings): Remove optional dependencies

kdy1:
**Description:**

I'm not sure if these packages are the reason for napi injecting optional dependencies, but these are not required anyway.


kdy1 (commented):
swc-bump:
 - dbg-swc

Issue #6209, opened 2022-10-19
Package imports are rewritten incorrectly if they have 3 or more segments separated by dots


gzzhanghao:
### Describe the bug

Issue https://github.com/swc-project/swc/issues/5214 still exists in 1.3.9

Reproduction repository:
https://github.com/gzzhanghao/swc-paths-extensions

### Input code

```typescript
import { modulo } from '@modules/style.module.less'
import { sum } from './sum.js'
```

### Config

```json
{
  "jsc": {
    "baseUrl": "./src",
    "paths": {
      "@modules/*": ["./modules/*"]
    },
    "target": "es2020",
    "parser": {
      "syntax": "typescript"
    }
  },
  "module": {
    "type": "es6"
  }
}
```

### Expected behavior

```ts
import { modulo } from "./modules/style.module.less";
import { sum } from "./sum";
```

### Actual behavior

```ts
import { modulo } from "./modules/style.module";
import { sum } from "./sum.js";
```

### Version

1.3.9

PR #6339, opened 2022-11-03
fix(es/modules): Preserve extensions

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6209.

### DIFF
--- a/crates/swc_ecma_transforms_module/src/path.rs
+++ b/crates/swc_ecma_transforms_module/src/path.rs
@@ -112,49 +112,11 @@ where
     R: Resolve,
 {
     fn resolve_import(&self, base: &FileName, module_specifier: &str) -> Result<JsWord, Error> {
-        fn to_specifier(
-            target_path: &str,
-            is_file: Option<bool>,
-            orig_ext: Option<&str>,
-        ) -> JsWord {
-            let mut p = PathBuf::from(target_path);
+        fn to_specifier(target_path: &str, orig_ext: Option<&str>) -> JsWord {
+            let p = PathBuf::from(target_path);
 
             if cfg!(debug_assertions) {
                 trace!("to_specifier: orig_ext={:?}", orig_ext);
-            }
-
-            let dot_count = p
-                .file_name()
-                .map(|s| s.to_string_lossy())
-                .map(|v| v.as_bytes().iter().filter(|&&c| c == b'.').count())
-                .unwrap_or(0);
-
-            match orig_ext {
-                Some(orig_ext) => {
-                    if is_file.unwrap_or_else(|| p.is_file()) {
-                        if let Some(..) = p.extension() {
-                            if orig_ext == "ts"
-                                || orig_ext == "tsx"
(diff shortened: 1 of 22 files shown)


kdy1 (commented):
swc-bump:
 - dbg-swc

crates/swc_ecma_transforms_module/src/path.rs
```diff
@@ -112,51 +112,13 @@ where
...
-                _ => {
-                    if is_file.unwrap_or_else(|| p.is_file()) {
-                        if let Some(v) = p.extension() {
-                            if v == "ts" || v == "tsx" || v == "js" || v == "jsx" {
-                                p.set_extension("");
-                            }
-                        }
-                    }
-                }
-            }
```
alexander-akait, line 158: I think it was workaround for ts, i.e.
```
import foo from "./test.ts"
```

to

```
import foo from "./test"
```

So `test.js` can be resolved, not sure why it was implemented in this way :confused: 

babel keeps exetensions too

kdy1: 🤣 I made lots of mistakes

Nickersoft:
Hey @kdy1! I just had to downgrade because of this change, for the exact reason I think you mentioned in your comment 😅  I have a NestJS project where none of my imports use extensions (as per NestJS convention) – however, now my TypeScript compile from this:

```typescript
import foo from './bar.module';
``` 

to

```typescript
import foo from './bar.module.ts';
```

even though SWC is generating a directory full of JavaScript files. My code crashes because it can't find `bar.module.ts` (because it's compiled to `bar.module.js`). Even changing my code to import from `bar.module.js` (the weird ESM convention for TS), SWC still generates the import with the `.ts` file extension lol.

Is there any workaround for this? Or am I stuck on 1.3.14 for now?

kdy1:
Use `.js` for omports instead.  swc is not a fixer. Your code is wrong. swc should not **fix** it.

kdy1:
You can **fix your code**. It's not a something to workaround. Your code is simply wrong.

Nickersoft:
As I mentioned, **using .js imports get rewritten back to \*.ts anyway**. 

I spent some time looking into this, and realized that as long as you have a `paths` option in `.swcrc`, it will break imports. 

**[I set up this example repo](https://github.com/Nickersoft/swc-paths)**. 

The project is ESM, and with a `main.ts` file that imports a TypeScript file (`import.ts`) using a `.js` extension. There is a bogus `paths` entry in `.swcrc` (doesn't matter if it's valid or not – I checked). Run `npm install` and then `npm run build`. It'll compile to a `dist` directory, but the code will be un-runnable because the `.js` import was rewritten _back_ to a `.ts` import.

This _is_ a legitimate problem, regardless of how you're importing your modules.

kdy1:
My bad, I misread your comment, sorry. Can you file an issue?

Nickersoft:
Yep 👍🏻

PR #6380, opened 2022-11-08
fix(css/parser): `<general-enclosed>` parsing + fix todo

alexander-akait:
**Description:**

Fix `<general-enclosed>` parsing with whitespaces + fix TODO

**BREAKING CHANGE:**

No

No


crates/swc_css_codegen/tests/fixture/values/url/1/output.css
```diff
@@ -45,7 +45,7 @@ div {
   background: url('\"foo\"');
 }
 .bar {
-  background: url(http://example.com/image.svg param(--color var(--primary-color);
+  background: url(http://example.com/image.svg param(--color var(--primary-color)));
```
alexander-akait, line 48: Now we don't lose `)` in declarations

crates/swc_css_parser/tests/recovery/at-rule/import/invalid-supports/output.swc-stderr
```diff
@@ -1,5 +1,5 @@
 
-  x Expected function or '(' token
```
alexander-akait, line 2: I think we should avoid using `token` in error messages, because developers can don't know what is `token`

kdy1 (approved):
swc-bump:
 - swc_css_parser

kdy1 (approved):

PR #6383, opened 2022-11-08
refactor(css/parser): avoid unnecessary checks

alexander-akait:
**Description:**

After refactor we don't need it anymore

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_css_parser

PR #6385, opened 2022-11-09
refactor(css/parser): union inputs

alexander-akait:
**Description:**

Reduce size of binary

**BREAKING CHANGE:**

Yes

No


crates/swc_css_parser/src/lib.rs
```diff
@@ -75,26 +53,30 @@ where
...
     res
 }
 
 /// Parse a given file as `T`.
 ///
 /// If there are syntax errors but if it was recoverable, it will be appended
 /// to `errors`.
-pub fn parse_tokens<'a, T>(
-    tokens: &'a Tokens,
+pub fn parse_input<'a, T>(
```
alexander-akait, line 66: Now we can accept and Tokens and ListOfComponentValues as input (as in spec)

crates/swc_css_parser/src/lib.rs
```diff
@@ -37,28 +37,6 @@ where
...
-    src: &'a str,
-    start_pos: BytePos,
-    end_pos: BytePos,
-    config: ParserConfig,
-    errors: &mut Vec<Error>,
-) -> PResult<T>
-where
-    Parser<Lexer<StringInput<'a>>>: Parse<T>,
-{
-    let lexer = Lexer::new(StringInput::new(src, start_pos, end_pos), config);
```
alexander-akait, line 54: I think we don't need it:
- `ecma` and `html` doesn't support it in favor `FileName::Anon` for `SourceFile`

kdy1 (approved):
swc-bump:
 - swc_css_parser

PR #6393, opened 2022-11-09
fix(html/minifier): merge bugs + merge modules

alexander-akait:
**Description:**

Fix bugs with `<script>` merging + improve merging for `<script type="module">`

**BREAKING CHANGE:**

No

No


crates/swc_html_minifier/tests/fixture/element/script/output.min.html
```diff
@@ -19,11 +19,7 @@
 
         </script><script id=data type=application/json>{"foo":"bar"}</script><script>
             broken    broken
-        ;
-            var a = 1 + 2;
-        ;;
-            var a = 3 + 6;
-        ;</script><script type=module>import t from"foo.js";let o=58800,e="testtest"</script><script>let e=new Date;alert("Today's date is "+e)</script><script src=javascript.js>
+        </script><script>var a=3,a=9</script><script type=module>import t from"foo.js";let o=58800,e="testtest"</script><script>let e=new Date;alert("Today's date is "+e)</script><script src=javascript.js>
```
alexander-akait, line 22: Bug fix, now we don't merge `script` with broken content

crates/swc_html_minifier/src/lib.rs
```diff
@@ -1915,6 +1948,168 @@ impl Minifier<'_> {
...
+        }
+
+        if is_modules {
+            swc_ecma_visit::VisitMutWith::visit_mut_with(
+                &mut left_program,
+                &mut swc_ecma_transforms_base::hygiene::hygiene(),
+            );
+        }
+
+        let left_program = swc_ecma_visit::FoldWith::fold_with(
```
kdy1, line 2075: Please use `left_program.fold_with(&mut ...)` instead

crates/swc_html_minifier/src/lib.rs
```diff
@@ -1915,6 +1948,168 @@ impl Minifier<'_> {
...
+                _ => return None,
+            }
+        };
+
+        // Avoid compress potential invalid JS
+        if !right_errors.is_empty() {
+            return None;
+        }
+
+        let right_unresolved_mark = Mark::new();
```
kdy1, line 2036: You should share `unresolved_mark` for all files

kdy1 (changes requested):

crates/swc_html_minifier/src/lib.rs
alexander-akait: I don't use this because I will need to use `use swc_ecma_visit::FoldWith;`, but there are ecma/css/html crates here, so I tried to avoid renaming using `as` and use it directly

alexander-akait:
@kdy1 Answered and fixed

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_html_minifier

Issue #6232, opened 2022-10-21
I can't compile swc_plugin_runner tests


Bruce-Hopkins:
### Describe the bug

I'm attempting to contribute to the project, so I tried to run the tests first. For some reason, I get this error when I try to compile:

```bash
error[E0061]: this function takes 0 arguments but 1 argument was supplied
   --> crates/swc_plugin_runner/tests/rkyv.rs:24:5
    |
24  |     init_plugin_module_cache_once(&None);
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ----- argument of type `&Option<_>` unexpected
    |
```

This causes my terminal to force exit. 

### Input code

Run the bash command:

```bash
cargo test --all --no-default-features --features swc_v1 --features memory_cache
```

### Expected behavior

The tests run smoothly

### Actual behavior

The tests fail

### Version

The main branch

### Additional context

I'm using Fedora 35.

kdy1:
You need `__rkyv` feature

Bruce-Hopkins:
> You need `__rkyv` feature

How do I implement this? I can't seem to run the test, or most of the cargo commands because they keep failing.

kdy1:
`--features rkyv-impl`

PR #6396, opened 2022-11-10
doc(contributing): Change cache feature for running all tests

orionmiz:
**Description:**

`swc_plugin_runner` cannot be compiled with `memory_cache` feature.

Also, there are some feature issues about `memory_cache` with `rkyv-impl`.

So let new comers use the `filesystem_cache` not affected by above things.

**BREAKING CHANGE: X** 

- https://github.com/swc-project/swc/issues/6232

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,13 +115,13 @@ For running all tests, take the following steps:
 7. Run tests
 
     ```bash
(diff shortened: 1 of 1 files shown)


kdy1 (approved):
Thanks!


---

swc-bump:
 - dbg-swc

PR #6399, opened 2022-11-10
chore(es/utils): More uniqure name for alias

Austaras:
**Description:**

I'm debugging some issue with optional chaining, and find the redundant aliases are quite hard to read.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!


---

swc-bump:
 - swc_ecma_utils

Issue #6404, opened 2022-11-11
[Bug]: Memory leak when invoking functionality from a WASM plugin


stahlbauer:
### Describe the bug

I have written a plugin that uses the `metadata.source_map` to guide some decisions. Calling methods of this object, in particular `span_to_snippet`, leads to a memory leak.

Here is an example of a plugin that reproduces the issue:

```
#[plugin_transform]
pub fn process_transform(
    mut program: Program,
    metadata: TransformPluginProgramMetadata,
) -> Program {

    for i in 1..50000 {        
        let j: u32 = i;
        println!("i {} j {}", i, j);
        metadata.source_map.span_to_snippet(Span::new(BytePos(j), BytePos(j+10), SyntaxContext::empty()));
    }

    program
}
```
I have used this plugin in my test environment which invokes it two times. This results in the following memory graph:

![with_sourcemap_access_leak](https://user-images.githubusercontent.com/15855927/201302182-ff6826fb-cc96-407c-b2f4-5b67ecf4791d.png)

Removing the loop also eliminates the memory leak.

![without_sourcemap_access](https://user-images.githubusercontent.com/15855927/201302242-04d55ab9-f9d0-4163-83d4-a5e1eb63bab0.png)

Note that I am using a version of SWC that allows for using the full 4GB of WASM memory---see https://github.com/swc-project/swc/pull/6378 .

### Expected behavior

The memory should be freed after use.

### Version

1.3.14

kwonoj:
I'm bit unsure if this is expected amount of pressure or if there's an actual leak. If leak is true, it'll be probably allocated buffer in the wasm memory space is not explicitly freed somewhere.

stahlbauer:
@kwonoj Are you planing to address this proactively? Which means are currently in place to free memory allocated in context of sharing data between the native process and the WASM world?

kwonoj:
Need confirmation on https://github.com/swc-project/swc/issues/6404#issuecomment-1312305783 .

stahlbauer:
See, for example, https://www.oreilly.com/library/view/what-is-webassembly/9781492076902/ch04.html#:~:text=WebAssembly%20doesn't%20have%20a,for%20managing%20allocation%20of%20memory.

kwonoj:
I meant _someone_. If I have some time I could, but I can't promise anything. 

[...]

PR #6426, opened 2022-11-13
refactor(swc/plugin_proxy): simplify read_result_* control flow

kwonoj:
**Description:**

Trying to apply some changes for https://github.com/swc-project/swc/issues/6404, but before doing it attempts to simplify control flows mostly for the readers (human). `read_returned_result_from_host_*` is supposed to be called from plugin binary only, changes conditional to be applied into whole fn itself instead of inner partial logic to easily read through whole flows.

**BREAKING CHANGE:**


kdy1 (approved):
swc-bump:
 - swc_plugin_proxy

Issue #6407, opened 2022-11-11
Value is lost because of compression with comma operator


krasnoperov:
### Describe the bug

Value is lost because of compression with comma operator.

### Input code

```typescript
export default class Demo {
  static encode(value) {
    const ranges = [];
    let range = [];
    let retrString = A.encode(value);
    let bitField = '';
    value.forEach((curValue, i) => {
        bitField += B.encode(curValue);
        range.push(i);
        ranges.push(range);
    });
    retrString += '.';
    retrString += C.encode(ranges);
    return retrString;
  }
}
```

### Config

```json
{
  "minify": false,
  "jsc": {
    "parser": {
      "syntax": "ecmascript"
    },
    "transform": {
      "legacyDecorator": true,
      "react": {
        "runtime": "automatic"
      }
    },
    "target": "es2022",
    "loose": false,
    "externalHelpers": true,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": true,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true,
        "const_to_let": true,
        "pristine_globals": true
      },
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  }
}
```

### Playground link

[...]

PR #6428, opened 2022-11-14
fix(es/minifier): Preserve op of the op-assignments in seq inliner

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6407.

---

# Investigation

## Output of Single-pass

```js
export default class Demo {
    static encode(value) {
        const ranges = [];
        let range = [], retrString = A.encode(value), bitField = '';
        return value.forEach((curValue, i)=>{
            bitField += B.encode(curValue);
            range.push(i);
            ranges.push(range);
        }), retrString = '.' + C.encode(ranges);
    }
}
console.log(Deno.encode());

```


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

jridgewell (approved):

Issue #6405, opened 2022-11-11
Whole `if` block is omitted in case throw after the `return undefined`


flex-mookeun:
### Describe the bug

Hi there, 

In if block, throwing an error after `return undefined` makes whole `if` block is omitted.

Please see the input code and playground for detail.

I've been experincing this issue since I updated my nextjs version.
I checked the issue is reproducible @1.3.14 **when compress is on**.

Thanks in advance.

### Input code

```typescript
const fn = () => {
  let val;

  if (!val) {
    return undefined;
    // works as expected if comment out below line
    throw new Error('first');
  }

  if (val.a?.b !== true) { // Uncaught TypeError: Cannot read properties of undefined (reading 'a')
    throw new Error('second');
  }
  return val;
}
```

### Playground link

[...]

PR #6429, opened 2022-11-14
fix(es/minifier): Fix code for dropping unreachable statements

kdy1:
**Description:**

We should preserve return in this case.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6405.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/src/compress/pure/dead_code.rs
```diff
@@ -269,15 +269,7 @@ impl Pure<'_> {
...
-            match stmts[idx].as_stmt() {
-                Some(Stmt::Return(ReturnStmt { arg: None, .. })) => {
-                    // Exclude return
-                    new_stmts.extend(stmts.drain(..idx));
-                }
-                _ => {
-                    new_stmts.extend(stmts.drain(..=idx));
-                }
-            }
+            new_stmts.extend(stmts.drain(..=idx));
```
jridgewell, line 272: We have a before and after diff of:

```diff
+    if (!!val) {
         if (val.a?.b !== !0) throw Error('second');
         return val;
+    }
```

What I think this change is doing is keeping the return statement in the new output, which prevents us from optimizing `if (!!val) if (val.a?.b !== !0) …` into `if (val.a?.b !== !0) …`. But that optimization is still broken, right? That's not a valid compression, and I think we've missed the real bug.

jridgewell (changes requested):

crates/swc_ecma_minifier/src/compress/pure/dead_code.rs
kdy1: `drop_unreachable_stmts ` is called for all kinds of statements, including `stmts` of `BlockStmt`.
And works only if there's an unreachable statement.


```js
if (foo) {
   return;
   throw foo
}
```

will trigger this function, but

```js
if (foo) {
   return;
}
```

will not.

It seems like I think it's a function called only for functions while writing this function initially.
But actually this is called for all kind of statements, and if so we should not drop `return foo` in

```js
if (foo) {
   return foo;
   throw foo
}

return bar
```

jridgewell: Ok, so that's one bug that's fixed. But I'm still confused how, even if we were to drop the return statement, we end up also dropping the `if (!!val)` wrapper around the throw?

kdy1: # Logs

I removed unrelated logs

Main:

```
  DEBUG  Dropped `undefined` from `return undefined`, kind: "change"
    at crates/swc_ecma_minifier/src/compress/pure/misc.rs:233
    in apply pure optimizer
    in optimize with pass: 1
    in compress ast
    in minify

  DEBUG  Dropping statements after a control keyword, kind: "change"
    at crates/swc_ecma_minifier/src/compress/pure/dead_code.rs:229
    in apply pure optimizer
    in optimize with pass: 1
    in compress ast
    in minify


  DEBUG  Converting empty block to empty statement, kind: "change"
    at crates/swc_ecma_minifier/src/compress/optimize/mod.rs:1228
    in visit_mut_stmt
    in visit_mut_if_stmt
    in visit_mut_stmt
    in handle_stmt_likes
    in visit_mut_block_stmt
    in visit_mut_arrow_expr
    in visit_mut_var_declarator
    in visit_mut_var_declarators
    in visit_mut_var_decl
    in visit_mut_decl
    in visit_mut_export_decl
    in handle_stmt_likes
    in apply full optimizer
    in optimize with pass: 1
    in compress ast
    in minify

  DEBUG  conditionals: `if (foo);` => `foo` , kind: "change"
    at crates/swc_ecma_minifier/src/compress/optimize/conditionals.rs:221
    in visit_mut_stmt
    in handle_stmt_likes
    in visit_mut_block_stmt
    in visit_mut_arrow_expr
    in visit_mut_var_declarator
    in visit_mut_var_declarators
    in visit_mut_var_decl
    in visit_mut_decl
    in visit_mut_export_decl
    in handle_stmt_likes
    in apply full optimizer
    in optimize with pass: 1
    in compress ast
    in minify

  DEBUG  Compressing boolean literal, kind: "change"
    at crates/swc_ecma_minifier/src/compress/optimize/mod.rs:578
    in visit_mut_bin_expr
    in visit_mut_if_stmt
    in visit_mut_stmt
    in handle_stmt_likes
    in visit_mut_block_stmt
    in visit_mut_arrow_expr
    in visit_mut_var_declarator
    in visit_mut_var_declarators
    in visit_mut_var_decl
    in visit_mut_decl
    in visit_mut_export_decl
    in handle_stmt_likes
    in apply full optimizer
    in optimize with pass: 1
    in compress ast
    in minify

  DEBUG  sequences: Compressing statements as a sequences, kind: "change"
    at crates/swc_ecma_minifier/src/compress/optimize/sequences.rs:157
    in handle_stmt_likes
    in visit_mut_block_stmt
    in visit_mut_arrow_expr
    in visit_mut_var_declarator
    in visit_mut_var_declarators
    in visit_mut_var_decl
    in visit_mut_decl
    in visit_mut_export_decl
    in handle_stmt_likes
    in apply full optimizer
    in optimize with pass: 1
    in compress ast
    in minify
```


Fixed:

```
DEBUG  Dropped `undefined` from `return undefined`, kind: "change"
    at crates/swc_ecma_minifier/src/compress/pure/misc.rs:233
    in apply pure optimizer
    in optimize with pass: 1
    in compress ast
    in minify

  DEBUG  Dropping statements after a control keyword, kind: "change"
    at crates/swc_ecma_minifier/src/compress/pure/dead_code.rs:229
    in apply pure optimizer
    in optimize with pass: 1
    in compress ast
    in minify

   DEBUG  if_return: Negating `foo` in `if (foo) return; bar()` to make it `if (!foo) bar()`, kind: "change"
    at crates/swc_ecma_minifier/src/compress/pure/if_return.rs:89
    in postcompress
    in minify

```

kdy1: So...

previously

```js
export const fn = () => {
    let val;

    if (!val) {
        return undefined;
        // works as expected if comment out below line
        throw new Error('first');
    }

    if (val.a?.b !== true) { // Uncaught TypeError: Cannot read properties of undefined (reading 'a')
        throw Error('second');
    }
    return val;
}
```

=>

```js
export const fn = () => {
    let val;

    if (!val) {
    }

    if (val.a?.b !== true) { // Uncaught TypeError: Cannot read properties of undefined (reading 'a')
        throw Error('second');
    }
    return val;
}
```

=>

```js
export const fn = () => {
    let val;

    if (!val) ;

    if (val.a?.b !== true) { // Uncaught TypeError: Cannot read properties of undefined (reading 'a')
        throw Error('second');
    }
    return val;
}
```

=>

```js
export const fn = () => {
    let val;

    if (val.a?.b !== true) { // Uncaught TypeError: Cannot read properties of undefined (reading 'a')
        throw Error('second');
    }
    return val;
}
```

jridgewell: Got it, it's because it's the first if statement that has the return dropped.

jridgewell (approved):

crates/swc_ecma_minifier/src/compress/pure/dead_code.rs
kdy1: New version:

```js
export const fn = () => {
    let val;

    if (!val) {
        return;
    }

    if (val.a?.b !== true) { // Uncaught TypeError: Cannot read properties of undefined (reading 'a')
        throw Error('second');
    }
    return val;
}
```

=> (negation from if_return)


```js
export const fn = () => {
    let val;

    if (val) {
    // condition is negated to remove return token

        if (val.a?.b !== true) { // Uncaught TypeError: Cannot read properties of undefined (reading 'a')
            throw Error('second');
        }
        return val;
     
    }

}
```

PR #6436, opened 2022-11-14
refactor(bindings/wasm): deprecate jsvalue::*_serde

kwonoj:
**Description:**

Refactor wasm binding to not to use `JsValue::*_serde`. It is effectively deprecated (https://github.com/rustwasm/wasm-bindgen/pull/3031) , and also one of the reason to create circular dependencies around indexmap.

**BREAKING CHANGE:**


kdy1 (approved):
swc-bump:
 - binding_macros

PR #6439, opened 2022-11-15
fix(es/minifier): Change the default of `mangle.toplevel` to false

kdy1:


kdy1 (commented):
swc-bump:
  - swc_ecma_minifier

PR #6441, opened 2022-11-15
feat(css/parser): Add `parse_string_input`

kdy1:


kdy1 (commented):
swc-bump:
 - dbs-swc

alexander-akait:
@kdy1 Can you describe why we need it? I think I know potential place where is bug with spans

PR #6442, opened 2022-11-15
chore: Publish `v1.3.17`

kdy1:


kdy1 (commented):
swc-bump:
 - dbg-swc

kdy1:
cc @kwonoj I reverted https://github.com/swc-project/swc/commit/1dd8b3d9d5393fa382345dce3d6977f7d4b5aa63 as it broke wasm

kwonoj:
@kdy1 can I see Ci logs to fix it forward?

kdy1:
https://github.com/swc-project/swc/actions/runs/3467260819/jobs/5791963901

kdy1:
Oh revert was not enough

kwonoj:
Let bindings use published version of swc_core serves well, but we definitely need some way to test latest swc_core directly before publishing bindings :/ I'll file an issue and think about some way.

kdy1:
It does not work. With the latest version wasm CI fails

kdy1:
See https://github.com/swc-project/swc/actions/runs/3467460473/jobs/5792360658
It's using `binding_macros@ v0.20.94`

kwonoj:
> It does not work. With the latest version wasm CI fails

Do you mean previous version of swc_core, or one more bump of swc_core with revert? if it's reverted + swc_core has been bumped up and still same compilation error occurs?

kdy1:
No, it's not reverted yet and the broken version (with #6436) is the latest version at the moment

kwonoj:
I see.

What I meant

> Let bindings use published version of swc_core serves well, 

was not for this specific case, it's for more general cases (aka `when it works, it works well`). As said above, seems we're hitting bumps bit more than we expected - I'll try to see if I can come up with some plan.

PR #6455, opened 2022-11-16
fix(html/minifier)/(ecma/minifier): bug with mangling

alexander-akait:
**Description:**

Here two things:
- **FIXED**: bug when we don't apply `compressor` for `script` node
- **FIXED**: bug with `topLevel` for `Program::Script` and keep top level stuff by default not mangled
- **FIXED**: better default options for JS optimization

I have

```html
<script>
 var topLevel = function test() {
            let foo = "bar";

            if (foo) {
                foo += "baz";
            }

            console.log(foo);
        };
</script>
```

For options I use: `swc_ecma_minifier::option::MinifyOptions::default()` 

After minification I have:

```html
<script>var topLevel=function test(){let foo="bar";if(foo){foo+="baz"}console.log(foo)}</script>
```

Yes, we keep `topLevel`, but `foo` is not mangling

Also if you use
```rust
swc_ecma_minifier::option::MinifyOptions {
                    compress: Some(swc_ecma_minifier::option::CompressOptions::default()),
                    mangle: Some(swc_ecma_minifier::option::MangleOptions {
                        top_level: false,
                        ..Default::default()
                    }),
                    ..Default::default()
                }
```

You will get (**only for `swc_ecma_ast::Program::Script`**, but if I changed it on `swc_ecma_ast::Program::Module` everything is fine):

```html
<script>var o=function o(){let o="bar";if(o){o+="baz"}console.log(o)}</script>
```

i.e. `top_level` is not respected

**BREAKING CHANGE:**

No

No


crates/swc_html_minifier/tests/fixture/element/script-options-2/output.min.html
```diff
@@ -0,0 +1,5 @@
+<!doctype html><html lang=en><title>Document</title><div>Script:</div>
+<script>/* Should mangle top level stuff */ var o=function(){let o="bar";o&&(o+="baz"),console.log(o)}</script>
```
alexander-akait, line 2: Here another bug with printing comments - space between `/* Should mangle top level stuff */` and `var o`, we can should avoid it, because it increase size of bundle (I think this bug happens everywhere with comments), I will fix it in another PR

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_minifier

Issue #6323, opened 2022-11-02
[Bug]: Division of an object after `<`


p51lee:
### Describe the bug

Dividing an object in a right-hand-side of `<` or `>` causes a transpilation fail.
This issue is similar with the previous issue https://github.com/swc-project/swc/issues/6322, in that they are both related to division of an object,
but I am submitting these issues separately because I think those two might have separated root cause in that the previous one requires the division to be in for-loop condition and this one requires the division to be at RHS of `<` or `>`.

### Input code

```typescript
let x = 0 < { } / 0 ;
```

### Playground link

https://play.swc.rs/?version=1.3.11&code=H4sIAAAAAAAAA8tJLVGoULBVMFCwUahWqFXQB7KsAb6dryUVAAAA&config=H4sIAAAAAAAAA0WMTQrEIAxG75K1286id5hDBCctFv9IUhgR714tlu7C915ehUMsrBUyshCPS0pU%2FMMKZAOKZZcVTNf6tKEXagYUeScdiiyd%2BZSEJjUQXHRbGSWbQmYSeRHG3T9m66GQfucYKmjJdAc%2F0N7G%2FHPynaLySe0Cj9ke9LUAAAA%3D

### Expected behavior

`input.js` is a valid JavaScript code:
```console
$ node input.js
```

### Actual behavior

SWC fails to transpile `input.js`; it thinks `/ 0 ;` is an unterminated regexp literal:
<img width="572" alt="image" src="https://user-images.githubusercontent.com/68288688/199450000-e9ae396e-2efb-4195-9d15-b07543e0c741.png">

### Version

1.3.11 (default SWC playground)

Issue #6322, opened 2022-11-02
[Bug]: Division of an object in for-loop condition


p51lee:
### Describe the bug

`input.js` is a valid JS code but SWC transpilation fails.

### Input code

```typescript
for ( ; { } / 1 ; ) ;
```

### Playground link

https://play.swc.rs/?version=1.3.11&code=H4sIAAAAAAAAA0vLL1LQUChLLFKoULBWqFaoVdBX0AazNRWsAWEswikdAAAA&config=H4sIAAAAAAAAA0WMTQrEIAxG75K1286id5hDBCctFv9IUhgR714tlu7C915ehUMsrBUyshCPS0pU%2FMMKZAOKZZcVTNf6tKEXagYUeScdiiyd%2BZSEJjUQXHRbGSWbQmYSeRHG3T9m66GQfucYKmjJdAc%2F0N7G%2FHPynaLySe0Cj9ke9LUAAAA%3D

### Expected behavior

The input code is a valid JavaScript code.
```console
$ node input.js
// terminates without error
```

### Actual behavior

SWC fails to transpile `input.js`:
<img width="576" alt="image" src="https://user-images.githubusercontent.com/68288688/199492179-4ccc89a4-8e31-41b6-aa20-8e06bedaeb8b.png">

### Version

1.3.11 (default SWC playground)

PR #6469, opened 2022-11-18
fix(ecma/parser): regexp parsing

alexander-akait:
**Description:**

Fix regexp parsing

**BREAKING CHANGE:**

Potential yes due new method

fixes https://github.com/swc-project/swc/issues/6322
fixes https://github.com/swc-project/swc/issues/6323


crates/swc_ecma_parser/src/lexer/state.rs
```diff
@@ -22,6 +22,7 @@ use crate::{
 #[derive(Clone)]
 pub(super) struct State {
     pub is_expr_allowed: bool,
```
alexander-akait, line 24: Not sure why we need it, because I looked at babel and esbuild logic and don't found the same semantic, so I think it can be simplified

crates/swc_ecma_parser/src/parser/expr.rs
```diff
@@ -1880,7 +1899,10 @@ impl<I: Tokens> Parser<I> {
         }
 
         if is!(self, ';')
-            || (!is!(self, '*') && !cur!(self, false).map(Token::starts_expr).unwrap_or(true))
+            || (!is!(self, '*')
+                && !is!(self, '/')
+                && !is!(self, "/=")
+                && !cur!(self, false).map(Token::starts_expr).unwrap_or(true))
```
alexander-akait, line 1905: Babel and esbuild has a little different logic here, I think we should refactor it, I will put in TODO too

crates/swc_ecma_parser/tests/test262-error-references/fail/095bea002b10b8e1.js.stderr
```diff
@@ -1,6 +1,6 @@
 
-  x Unexpected eof
+  x Unterminated regexp literal
    ,-[$DIR/tests/test262-parser/fail/095bea002b10b8e1.js:1:1]
  1 | foo[/42
```
alexander-akait, line 4: Better error

alexander-akait:
@kdy1 Ready for review

kdy1 (approved):
swc-bump:
 - swc_ecma_parser

Issue #6463, opened 2022-11-17
`compress: true` breaks the code with hoisted function, assigned before the declaration


nd0ut:
### Describe the bug

SWC replaces the first function call with its body and removes the function itself from the code. And does nothing with the second function call.

### Input code

```typescript
var foo_1 = foo;

function foo() {
  console.log("foo");
}

foo_1();
foo_1();
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2022",
    "loose": false,
    "minify": {
      "compress": true,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.3.18&code=H4sIAAAAAAAAAytLLFJIy8%2BPN1SwBdHWXFxppXnJJZn5eSCuhqZCNZeCQnJ%2BXnF%2BTqpeTn66hhJQWEnTmqsWqBKkTwPIhjEAB2p%2B1U0AAAA%3D&config=H4sIAAAAAAAAA0XMTQoDIQyG4btk7aK46MI79BDBZgaLfyQOVMS71xRLd%2BHLyzPgJR7cgIosxHpJzw3f4IB8QvEcagOzsjUdGIWmgYZ8UtNE7M3a9Y6lCO3AQAo5HF0xX1JlEgHX%2BNIP5jP%2BwrmoVJ6XDgNar%2FQl7zD%2FxBaDPHaozvwAc2ge8LcAAAA%3D

### Expected behavior

```js
var foo_1 = function() {
    console.log("foo");
};
foo_1(), foo_1();
```

### Actual behavior

```js
console.log("foo"), foo();
```

### Version

1.3.18

PR #6473, opened 2022-11-18
fix(es/minifier): Abort inliner on fn declarations used multiple time

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6463.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/tests/fixture/issues/6463/output.js
```diff
@@ -0,0 +1,4 @@
+var foo_1 = function() {
```
jridgewell, line 1: Is it not possible to avoid the reassignment?

jridgewell (approved):

crates/swc_ecma_minifier/tests/fixture/issues/6463/output.js
kdy1: I think it's possible. I'll check, thank you!

kdy1: I tried, but we can't detect it while deciding to inline a function declaration. We need a costly reverse map to access `foo` from `bar = foo`.

I'll revisit this in the future

PR #6475, opened 2022-11-18
test(ts/compat): Migrate inline tests to fixture testing

kdy1:
**Description:**

I'm trying to test source map of each transforms, and I found that I should migrate tests to fixture test first.


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #6478, opened 2022-11-18
fix(es/minifier): Abort IIFE invoker on `eval`

kdy1:
**Description:**

**Related issue:**

 - https://github.com/vercel/next.js/issues/43052

---

# Investigation

This is an issue of `invoke_iife`. I used patch script to check each function.

image

There are two variables named `exports`


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/tests/fixture/next/43052/input.js
```diff
@@ -0,0 +1,224 @@
+(self["webpackChunk_N_E"] = self["webpackChunk_N_E"] || []).push([
```
jridgewell, line 1: Can we include a more minimal example that only has a an eval, and assert the vars aren't mangled?

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
```diff
@@ -728,6 +728,13 @@ where
             }
         }
 
+        // If we has an eval, we cannot remap variables correctly.
+        let has_eval = contains_eval(body, false);
+        if !param_ids.is_empty() && has_eval {
```
jridgewell, line 733: I'm kinda surprised we allow any further modifications in the presence of eval.

crates/swc_ecma_minifier/tests/fixture/next/43052/input.js
kdy1: I thought this was the minimal repro case, but it seems like I accidentally deleted `eval` while trying to reduce.

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
kdy1: `eval` is used by lots of libraries, so I think we should only give up if it's going to be problematic. (e.g. name mangler aborts)

crates/swc_ecma_minifier/tests/fixture/next/43052/input.js
kdy1: I added `mangle.json` to enable mangler and added some test code to input to ensure that name mangler is applied to the test case.

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
kdy1: But I'm not sure about this. Do you think it's better to abort completely on `eval`?

jridgewell: Personally, yah, I would. Looking at terser, `eval` deopts:
- name mangling
- unused var removal
- joining of expressions
- **inlining functions**
- conversion from functions/methods to arrows

kdy1: Did it and checked other passes, too. Thank you!

jridgewell (approved):

PR #6462, opened 2022-11-16
refactor(bindings/wasm): deprecate jsvalue::*_serde

kwonoj:
**Description:**

Second attempts to https://github.com/swc-project/swc/pull/6436, with various fixes to ensure things are not breaking.

PR also adds naive testing to swc_core to test with latest workspace root to build. This allows to catch problems upfront before publish swc_core (not @swc/core). It actually caught various failures while I'm working on this PR.

Also, PR updates `bindings_macro` to reflect latest changes.

**BREAKING CHANGE:**

**Related issue (if exists):**

PR #6485, opened 2022-11-20
refactor(preset_env_base): remove serde-serialize from bindgen

kwonoj:
**Description:**

Same as https://github.com/swc-project/swc/pull/6462, but for the transitive dep browserslist-rs. There is an open PR targeting the latest version of browserslist (0.12), but we are currently using 0.11 so use a commit https://github.com/kwonoj/browserslist-rs/commit/bd47b8e17c937d15ed75fe9c3522ef0ac94ffcd9#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R58 to safely remove features for now, and then later try to bump up version separately to avoid regressions.

**BREAKING CHANGE:**

### DIFF
--- a/crates/preset_env_base/Cargo.toml
+++ b/crates/preset_env_base/Cargo.toml
@@ -1,23 +1,24 @@
 [package]
-authors = ["강동윤 <kdy1997.dev@gmail.com>"]
-description = "Common logic for targetting vairous browsers"
+authors       = ["강동윤 <kdy1997.dev@gmail.com>"]
+description   = "Common logic for targetting vairous browsers"
 documentation = "https://rustdoc.swc.rs/preset_env_base/"
-edition = "2021"
-license = "Apache-2.0"
-name = "preset_env_base"
-version = "0.3.2"
+edition       = "2021"
+license       = "Apache-2.0"
+name          = "preset_env_base"
+version       = "0.3.2"
 
 [lib]
 bench = false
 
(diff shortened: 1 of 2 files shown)


crates/preset_env_base/Cargo.toml
```diff
@@ -1,23 +1,24 @@
...
-browserslist-rs = "=0.11.0"
-dashmap = "5.1.0"
-from_variant = { version = "0.1.3", path = "../from_variant" }
-once_cell = "1.12.0"
-semver = { version = "1.0.4", features = ["serde"] }
-serde = { version = "1", features = ["derive"] }
-st-map = "0.1.2"
-tracing = { version = "0.1.32" }
+# https://github.com/browserslist/browserslist-rs/pull/13
+browserslist-rs = { git = "https://github.com/kwonoj/browserslist-rs", rev = "bd47b8e17c937d15ed75fe9c3522ef0ac94ffcd9" }
```
kdy1, line 17: This will block publishing

kwonoj: ah, I keep forgetting it 🤦 . Guess this is no go and I have to wait https://github.com/browserslist/browserslist-rs/pull/13?

kwonoj: Closing for now. as we can't ship it this anyway.

PR #6486, opened 2022-11-20
ci(actions): scope pr trigger type

kwonoj:
**Description:**

Curious if we're interested in having reduced scope to trigger pr workflow? currently it is wide, runs all CI workflow for the commit / pr both when new commit's being updated in pull request. `synchronize` supposed to do the work for the pr, while we keep main branch push behavior as is.

ref: https://github.com/vercel/next.js/blob/canary/.github/workflows/build_test_deploy.yml#L3-L5

**BREAKING CHANGE:**

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,8 +1,11 @@
 name: CI
 
 on:
-  - push
-  - pull_request
+  pull_request:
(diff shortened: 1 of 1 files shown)


kdy1 (approved):
Thanks!

---
swc-bump:
 - dbg-swc

PR #6487, opened 2022-11-21
refactor(binding/wasm): use binding_macro

kwonoj:
**Description:**

I think I did this before 🤔 somehow wasm binding doesn't use macros. Trying to reapply.

**BREAKING CHANGE:**


kdy1 (approved):
swc-bump:
 - dbg-swc

PR #6488, opened 2022-11-21
Upgrade `browserslist-rs`

jaydenseric:
**Description:**

Upgrades the `browserslist-rs` dependency.

Importantly, `browserslist-rs` v0.12 marked Internet Explorer as dead:

https://github.com/browserslist/browserslist-rs/releases/tag/v0.12.0

**BREAKING CHANGE:**

Not sure if this is a breaking change, because browserslist queries are supposed to have different results over time. Most likely this PR could be considered a bug fix, because people expect the browserslist query to match the current spec and behavior.

N/A.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -216,9 +216,9 @@ dependencies = [
 
 [[package]]
 name = "browserslist-rs"
-version = "0.11.0"
+version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


kwonoj (changes requested):
We need https://github.com/browserslist/browserslist-rs/pull/13 change before upgrading brwoserslists-rs.

kdy1:
Closes as this is fixed by #6547

PR #6489, opened 2022-11-21
ci(actions): skip benchmark on fork

kwonoj:
**Description:**
Minor optimization for ci workflow: people can run ci on fork for various reasons (I do it for testing before PR) but running benchmark doesn't make sense lot as it is uncontrolled machine doesn't give reliable number anyway. 

PR changes flow to shortcurcuit those to not waste ci turnaround time for the fork.

**BREAKING CHANGE:**

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,8 +1,11 @@
 name: Benchmark
 
 on:
-  - push
-  - pull_request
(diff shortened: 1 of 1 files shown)


kdy1 (approved):
Thank you!

---

swc-bump:
 - dbg-swc

PR #6502, opened 2022-11-24
perf(es/transform_base): return early if any helper is used

iheyunfei:
**Description:**

Probably not a noticeable improvement.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_transforms_base

Issue #6491, opened 2022-11-21
swc minifier breaks react-joyride


kdy1:
https://github.com/vercel/next.js/discussions/30237#discussioncomment-4195233

Repro: https://github.com/JClackett/next-react-joyride

I'll work on this issue hopefully on this weekend

kdy1:
Investigation:

This seems like an issue of inliner.

<img width="681" alt="image" src="https://user-images.githubusercontent.com/29931815/203891117-45cad254-857c-4064-83a1-d03692dc8e2e.png">

works, while

<img width="681" alt="image" src="https://user-images.githubusercontent.com/29931815/203891176-acded091-9428-4224-a4a1-3d254ad3bffa.png">


is broken

PR #6509, opened 2022-11-25
fix(es/minifier): Make sequential inliner respect resolution order

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6491.

---

# Investigation

This requires `inline`, `sequences` **and** `conditional`

## `obj` of `MemberExpr` is **resolved** first

So the code below is invalid, because `fragments` of fragments[...]` is undefined, even if it's reassigned

image


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/src/compress/optimize/sequences.rs
```diff
@@ -1597,6 +1597,33 @@ where
                     return Ok(false);
                 }
 
+                // We can't merge into `[]` in some cases because `obj` is **resolved** before
+                // evaluating `[]`.
+                //
+                // See https://github.com/swc-project/swc/pull/6509
+
+                // We check only for an identifier because resolution is the problematic part.
+                // Evalutation order is not a problem.
```
jridgewell, line 1606: Is this true? Wouldn't `(1, obj)[obj = {}, 'key']` hit the same problem?

kdy1: Oh, I was wrong. Nice catch, and thank you!

kdy1: Nevermind, 
```js
var obj;
console.log((1, obj)[obj = {}, 'key'])
```

is expected to fail even without minification.

jridgewell (approved):

PR #6511, opened 2022-11-25
fix(es/visit): Visit TypeScript satisfies expressions

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base

PR #6517, opened 2022-11-26
ci(publish): trying to fix upload cli binaries

kwonoj:
**Description:**

Seems like previous CI adjustment was not working correctly, trying to fix forward.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---
swc-bump:
 - dbg-swc

Issue #6504, opened 2022-11-24
swc minifier breaks hls.js


kdy1:
- https://github.com/video-dev/hls.js/issues/5015

PR #6521, opened 2022-11-28
fix(es/minifier): Use unsafe option for arrow => method

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6504.
 - https://github.com/vercel/next.js/issues/43208.

---

# Investigation

This is caused by `arrows`, and seems like a same issue as https://github.com/swc-project/swc/pull/6507

```js
({foo(){}}).foo.toString()
```

prints `foo(){}`.


kdy1 (commented):
swc-bump:
 - dbg-swc

jridgewell (commented):
Does this really fix either bug? It seems like we're just avoiding it because we're not converting to method shorthand anymore, but the real bug is that the compresser is omitting the `{}` enclosing object?

jridgewell (approved):
Ok, crawling through the bugs it's being caused because they're doing string concatenation of functions, which is inherently buggy if we're minifying it.

Issue #6411, opened 2022-11-12
SourceMap performance issue


JSerFeng:
### Describe the bug

Input: A large bundle generated by webpack.
minify: 2s without `sourcemap`, and 20s with `sourcemap`

When minify and transform very large file, enable source map will have huge impact on performance, very huge impact that sometimes even slower than `terser`.

[Reprod repo here](https://github.com/JSerFeng/swc_sourcemap_perf_issue/tree/main)

### Playground link

As the file is too large, I think we better download the reprod repo locally

### Expected behavior

SourceMap should be slow to generate, but shouldn't be this slow

### Actual behavior

2s without `sourcemap`
20s with `sourcemap`

### Version

1.3.14

JSerFeng:
Additional information.
MacOs Monterey 12.1
CPU: 2.6 GHz Intel Core i7
32 GB mem

willstott101:
I can repro this with `@swc/core` but not a locally built swc...
```
will@will-ubuntu:~/repos/others/swc_sourcemap_perf_issue$ npm run bench

> perf_issue@1.0.0 bench
> esno bench.ts

swc: 1.861s
terser: 23.946s
swc sourcemap: 24.095s
terser sourcemap: 24.888s
```
```
will@will-ubuntu:~/repos/others/swc_sourcemap_perf_issue$ npm i ../swc/swc-core-1.3.19.tgz 
...
```
```
will@will-ubuntu:~/repos/others/swc_sourcemap_perf_issue$ npm run bench

> perf_issue@1.0.0 bench
> esno bench.ts

swc: 8.311s
terser: 24.047s
swc sourcemap: 8.592s
terser sourcemap: 24.509s
```
Not sure I can dig more without some additions to CONTRIBUTING.md about building the node package. As far as I got was seeing that serializing sourcemaps isn't benched in then `minify` bench.

https://github.com/swc-project/swc/blob/48bbebf5e9cc8d3336fc48ca8b07f51486f8da04/crates/swc/benches/minify.rs#L49 should be `BoolOrDataConfig::from_bool(true)` I reckon.

Once it is... then the benchmark is massively slowed down (like, x10 slower or worse) by: https://github.com/swc-project/swc/blob/48bbebf5e9cc8d3336fc48ca8b07f51486f8da04/crates/swc_common/src/source_map.rs#L974

[...]

PR #6523, opened 2022-11-28
perf(common): Fix perf bug of sourcemap for inputs with multi-byte chars

kdy1:
**Description:**

This PR makes the source map generator cache the previous position instead of searching it from 0 every time.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6411.


kdy1 (commented):
swc-bump:
 - swc_common

PR #6524, opened 2022-11-28
Perf/source map mutibyte

IWANABETHATGUY:
**Description:**

**BREAKING CHANGE:**


kdy1:
I think https://github.com/swc-project/swc/pull/6523 is a better fix, but not sure if it's correct fix

IWANABETHATGUY:
I would like to help to testing the performance.

kdy1:
You can add `large.js` to `swc_ecma_minifier/benches/full` and uncomment https://github.com/swc-project/swc/blob/1425475f4af7c50db4fc415feb4e3f49eea6b101/crates/swc/benches/minify.rs#L87

IWANABETHATGUY:
using the issue https://github.com/swc-project/swc/issues/6411 `large.js` 
## before: 
image
## this pr: 
image

PR #6532, opened 2022-11-28
refactor(css/parser): refactor logic

alexander-akait:
**Description:**

To be align with CSS spec and https://www.w3.org/TR/css-syntax-3/#declaration-rule-list

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_css_parser

PR #6555, opened 2022-11-30
fix(xml/parser): parse legacy document type

alexander-akait:
**Description:**

Fix legacy XML document type, it was removed from spec - https://developer.mozilla.org/en-US/docs/Web/API/DocumentType, but still generated some tools, inner parsing is not implemented, because it requires more time (I don't know should we really support this legacu stuff, so just put them in `raw` and avoid throwing errors) 

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_xml_parser

PR #6556, opened 2022-11-30
ci(actions): trying to debug cli binaries copy

kwonoj:
**Description:**

I really don't know why uploading cli binaries doesn't work 😭 PR adds few more output to see what's happening.

**BREAKING CHANGE:**


kdy1 (approved):
swc-bump:
 - dbg-swc

PR #6560, opened 2022-12-01
fix(css/parser): recovery more for non valid component values in declaration value

alexander-akait:
**Description:**

Bug fix - we lose `;` +  tests + better error recovery

**BREAKING CHANGE:**

No

No


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_parser

PR #6569, opened 2022-12-02
perf(css/ast): Reduce token size

alexander-akait:
**Description:**

This is one of the performance improvements PR

```
Gnuplot not found, using plotters backend
css/lexer/bootstrap_5_1_3                                                                             
                        time:   [5.0974 ms 5.1213 ms 5.1480 ms]
                        change: [-11.146% -10.572% -10.024%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) high mild
  6 (6.00%) high severe

css/lexer/foundation_6_7_4                                                                             
                        time:   [4.2814 ms 4.2999 ms 4.3219 ms]
                        change: [-6.3111% -4.5331% -2.8664%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) high mild
  7 (7.00%) high severe

css/lexer/tailwind_3_1_1                                                                            
                        time:   [802.97 us 807.05 us 812.42 us]
                        change: [-10.144% -7.5341% -4.9534%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  7 (7.00%) high severe

```

Parser was improved too

**BREAKING CHANGE:**

Yes

No


alexander-akait:
I think we can `boxed` more stuff and reduce it to `24 bytes` :thinking:

crates/swc_css_ast/src/token.rs
```diff
@@ -35,135 +35,202 @@ pub enum NumberType {
...
+)]
+#[cfg_attr(
+    feature = "rkyv",
+    archive(bound(
+        serialize = "__S: rkyv::ser::Serializer + rkyv::ser::ScratchSpace + \
+                     rkyv::ser::SharedSerializeRegistry",
+        deserialize = "__D: rkyv::de::SharedDeserializeRegistry"
+    ))
+)]
+pub enum Token {
```
kdy1, line 196: I think this is too much boxing.
Are you trying to optimive `memmove`?

alexander-akait: I tried different things, lexer is faster when we box tokens, because we use them in AST, i.e. `ComponentValues::PreservedTokens`, so memory usage was improved good, parser performance unfortunately was not very improved

kdy1: Also `Dimension`

crates/swc_css_lints/src/rules/font_family_no_duplicate_names.rs
```diff
@@ -56,18 +56,19 @@ impl FontFamilyNoDuplicateNames {
                         (fonts, Some((value.to_string(), *span)))
                     }
                 }
-                ComponentValue::Str(Str {
-                    raw: Some(raw),
-                    span,
-                    ..
-                }) => {
-                    fonts.push((FontNameKind::from(raw), *span));
+                ComponentValue::Str(value) if value.raw.is_some() => {
```
kdy1, line 59: Please use patter matching instead of an if guard

alexander-akait: fixed

kdy1 (approved):
Thanks!

---
swc-bump:
 - swc_css_ast --breaking

PR #6574, opened 2022-12-03
Fix BytePos -> CharPos calculations

### LINKED ISSUE #6552: SWC generate wrong sourcemap when have mutibyte character after 1.3.20
## Minimal reproduce repo
https://github.com/IWANABETHATGUY/swc-sourcemap-issue
## How to reproduce
1. git clone 
2. pnpm i 
3. node index
###  @swc/core 1.3.20
image
### @swc/core 1.3.21
image

### Input code

```typescript
const xxx = ', something'
console.error(`❌ ${message}`);

const bbb = ''
```

### Config

```json
{
    sourceMaps: true,
    jsc: {
      parser: {
        syntax: "ecmascript",
        jsx: false,
      },
      target: "es5",
      loose: false,
      minify: {
        compress: false,
        mangle: false,
      },
    },
    module: {
      type: "commonjs",
    },
    minify: true,
    isModule: true,
}
```

### Playground link

_No response_

### Expected behavior

Should generate right source map

### Actual behavior

_No response_

### Version

@swc/core 1.3.21

### Additional context

_No response_

jridgewell:
**Description:**

This fixes the BytePos -> CharPos calculation necessary for source maps. There were a few issues in the old code:

1. UTF-8 maps 1-3 bytes into 1 UTF-16 char, but 4 bytes into 2 UTF-16 chars
2. The starting offset was not recorded when we reach the end of the `multibyte_chars` iteration
3. The `mappings` can be unordered, meaning we need to restart UTF-16 offset calculation

**BREAKING CHANGE:**

Fixes #6552

### DIFF
--- a/crates/swc_common/src/source_map.rs
+++ b/crates/swc_common/src/source_map.rs
@@ -1714,7 +1714,6 @@ mod tests {
         for c in input.chars() {
             let actual = bpos.to_u32() - sm.calc_utf16_offset(&file, bpos, &mut state);
 
-            dbg!(&bpos, &cpos, &state);
             assert_eq!(actual, cpos.to_u32());
 
             bpos = bpos + BytePos(c.len_utf8() as u32);
--- a/crates/swc_common/src/source_map.rs
+++ b/crates/swc_common/src/source_map.rs
@@ -17,9 +17,7 @@
 //! within the SourceMap, which upon request can be converted to line and column
 //! information, source code snippets, etc.
 use std::{
-    cmp,
-    cmp::{max, min},
-    env, fs,
+    cmp, env, fs,
     hash::Hash,
     io,
     path::{Path, PathBuf},
@@ -295,8 +293,7 @@ impl SourceMap {
                 );
 
                 let linechpos = self.bytepos_to_file_charpos_with(&f, linebpos);
-
-                let col = max(chpos, linechpos) - min(chpos, linechpos);
+                let col = chpos - linechpos;
 
                 let col_display = {
                     let start_width_idx = f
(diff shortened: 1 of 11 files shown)


kdy1 (approved):
Thank you so much!

---

swc-bump:
 - swc_common

jridgewell:
Sorry, I just pushed a reverse conversion for when the mapping isn't ordered. This hopefully makes it a bit faster.

crates/swc_common/src/source_map.rs
```diff
@@ -1694,19 +1708,27 @@ mod tests {
...
-        let mut start = 0;
+        let mut state = ByteToCharPosState::default();
         let mut bpos = file.start_pos;
         let mut cpos = CharPos(bpos.to_usize());
         for c in input.chars() {
-            let actual = bpos.to_u32()
-                - sm.calc_utf16_offset(&file, &mut prev_extra_bytes, &mut start, bpos);
+            let actual = bpos.to_u32() - sm.calc_utf16_offset(&file, bpos, &mut state);
 
+            dbg!(&bpos, &cpos, &state);
```
kdy1, line 1717: Can you remove this line?

kdy1: Nevermind, it was a test

jridgewell: Done.

kdy1 (dismissed):
Thank you!

jridgewell:
Ah, one of my recent changes is causing a failure. Working to fix it now.

jridgewell:
Should be good now.

kdy1 (approved):

IWANABETHATGUY:
It seems that this pr regressed performance again
## PR 
image
## Main
image

IWANABETHATGUY:
large.js https://gist.github.com/IWANABETHATGUY/8cbfa7246af46a26cd8480ce7e9a5f47

jridgewell:
Can you explain how you're doing the benchmark? I'm actually seeing an ~2% improvement with that file running `time ./target/release/swc compile --source-maps true large.js > /dev/null`

kdy1:
Let's fix performance regression with another PR..

IWANABETHATGUY:
Sorry, you are right, I just using the wrong branch. No performance regression.

PR #6575, opened 2022-12-04
chore: 🤖 remove unnecessary code

IWANABETHATGUY:
**Description:**
1. remove unnecessary code, this function is invoked but did nothing if it return true,
also https://github.com/swc-project/swc/blob/9a20ddff71d31049a6b03d324ba3f20b04979645/crates/swc_ecma_minifier/src/metadata/mod.rs#L350-L444, Although this visitor visit `&mut ` Ast node, but we could see that it does't modified ast node, over all, remove 
those code is safe.
2. This pr would improve a little performance.

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_minifier/src/lib.rs
+++ b/crates/swc_ecma_minifier/src/lib.rs
@@ -171,7 +171,7 @@ pub fn optimize(
             options.compress.as_ref(),
             comments,
             marks,
-            extra.unresolved_mark,
+            // extra.unresolved_mark,
         ));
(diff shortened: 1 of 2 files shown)


IWANABETHATGUY:
@kdy1 , the benchmark CI is broken due to network issue,  but I don't have permission to retry it, would you mind helping with rerunning it?

kdy1 (approved):
swc-bump:
 - swc_ecma_minifier

Issue #6522, opened 2022-11-28
[Parser Bug] Failed to Parse JSX with ASI


g-plane:
### Describe the bug

Failed to parse JSX code with ASI. See the code below.

### Input code

```typescript
let x

<Comp></Comp>
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": true
    }
  }
}
```

### Playground link

https://play.swc.rs/?version=1.3.20&code=H4sIAAAAAAAAA8tJLVGo4OKycc7PLbCz0QdTAMVkXkUUAAAA&config=H4sIAAAAAAAAA0WOSw6DMAxEr4K8hkUX7YI79ARVF1ZwaKp8kG1oEcrdmyAqdtb4zdNs8BYD%2FQYTshDXS9ao%2BIUeyAQUw25SaAtWIuWZcguKPJJWQq7l5VMSgt6iF2ohuOjsWkUmhYlJ5HxhHP2fzEUU0jDXYANdJ9qFN8in4%2Bg5uR9gHdACxWXv7DOKHjyKNpfGvDgFahZicSmWZS4aPw%2Bl9wBljGIThw6Z06ezczRaKIFnzj%2BmMFVPBgEAAA%3D%3D

### Expected behavior

No parsing errors.

### Actual behavior

Failed to parse.

### Version

1.3.20

### Additional context

- [TypeScript playground](https://www.typescriptlang.org/play?#code/DYUwLgBAHgUDA8BhA9gWwA4D54HoUcyA)
- [Babel REPL](https://babel.dev/repl#?browsers=&build=&builtIns=false&corejs=3.6&spec=false&loose=false&code_lz=DYUwLgBAHgUDA8BhA9gWwA4D54HoUcyA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=react&prettier=false&targets=&version=7.20.4&externalPlugins=&assumptions=%7B%7D)

PR #6577, opened 2022-12-05
fix(ecma/parser): parse JSX with ASI

alexander-akait:
**Description:**

fix ASI with JSX

**BREAKING CHANGE:**

No

fixes https://github.com/swc-project/swc/issues/6522


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_parser

PR #6585, opened 2022-12-06
perf(css/parser): Reduce clones and allocations

kdy1:
**Description:**

 - We use `(BytePos, BytePos)` instead of `Box<Span>`.
 - We use `Cow` for `Input`.

---

image


kdy1 (commented):
swc-bump: 
 - swc_css_parser

alexander-akait (approved):

PR #6586, opened 2022-12-06
Extract the ES usage analyzer from the ES minifier crate

alexkirsz:
**Description:**

This PR extracts the EcmaScript usage analyzer from the ES minifier crate into its own crate, to make it reusable by third parties. Notably, we plan on using this in Turbopack to implement tree shaking.

**BREAKING CHANGE:**

The `swc_ecma_minifier::marks::Marks` type is now exported from `swc_ecma_usage_analyzer::marks::Marks` instead. It might make sense to also re-export it from the `swc_ecma_minifier` crate.

Fixes WEB-269


wbinnssmith (dismissed):

crates/swc_ecma_usage_analyzer/src/analyzer/mod.rs
```diff
@@ -234,7 +234,7 @@ enum RecursiveUsage {
 
 /// Analyzed info of a whole program we are working on.
 #[derive(Debug, Default)]
-pub(crate) struct ProgramData {
+pub struct ProgramData {
```
kdy1, line 237: Are you going to use this type directly?

kdy1: I prefer storing this type in `swc_ecma_minifier`, to reduce breaking changes.

kdy1 (commented):
I'm bit worried about the number of breaking changes

crates/swc_ecma_usage_analyzer/src/analyzer/mod.rs
alexkirsz: The current definition of `UsageAnalyzer` depends on it: `UsageAnalyzer<S = ProgramData>`. But I agree this makes more sense to live in `swc_ecma_minifier` for now so I'll move it and remove the default type.

From what I can tell the only breaking change is the `marks::Marks` type changing crates, and we can always re-export it from `swc_ecma_minifier`.

kdy1: Yeah, I think it's fine. Thank you!

crates/swc_ecma_usage_analyzer/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+pub mod alias;
+pub mod analyzer;
+pub mod marks;
+pub mod util;
+
+#[cfg(test)]
```
kdy1, line 6: Can you remove this?

crates/swc_ecma_usage_analyzer/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+mod tests {
```
kdy1, line 7: Can you remove this module?

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_minifier

PR #6592, opened 2022-12-07
fix(css/parser): avoid skipping extra whitespaces

alexander-akait:
**Description:**

Avoid skipping whitespaces in `[`/`(`/`{` because:
- it is invalid, whitespaces can allowed and can be disallowed based on context, we can't apply this logic for any CSS context
- we already skip whitespaces where it is necessary (no failed tests)
- avoid extra actions

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_css_parser

PR #6599, opened 2022-12-08
perf(css/codegen): Reduce allocations

kdy1:
image


kdy1 (commented):
swc-bump:
 - swc_css_utils

PR #6603, opened 2022-12-08
fix(node-swc): sync available settings for optimizer.globals

meskill:
**Description:**

Some fixes of typescript types for `@swc/core` package


kdy1 (approved):
swc-bump:
 - dbg-swc

Issue #6589, opened 2022-12-06
SWC produces invalid code if types have a trailing coma


skutyepov:
### Describe the bug

If types contain a trailing coma, swc adds to the actual code, which is a syntax error. 

For reference, TSC handles it correctly. [TSC playground ref.](https://www.typescriptlang.org/play?target=7#code/KYDwDg9gTgLgBAMwK4DsDGMCWEWNWgCgG8AoOOAQwBoy4AjG8gOhamAGcYSBfALjlLkK-TlEwoA5o3oiYYyTW4BKAWTQ52EADbAmWiBILUGcNpyU8gA)  

### Input code

```typescript
export function func({
  a,
  b,
  ...rest
}: {
  a: string,
  b: string,
}) {
 console.log(a,b, rest)
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2022"
  }
}
```

### Playground link

https://play.swc.rs/?version=1.3.21&code=H4sIAAAAAAAAA0utKMgvKlFIK81LLsnMzwMzNKq5FBQSdYBEEojQ09MrSi0u4aq1UgBLWCkUlxRl5qWDFSA4tZog6eT8vOL8nFS9nPx0jUSdJB0FkFZNrloAL7Tel2gAAAA%3D&config=H4sIAAAAAAAAAxWKQQqAMAwE%2F7JnD5JjfxNKFEWkZHNoKfm79TYzzMTNijLR1Gn%2BE8cb2lEQoxmrXy2wIbjSoQ8tl6ifFmsxyi6CzA%2Br99KJSAAAAA%3D%3D

### Expected behavior

SWC should not add trailing coma after `...rest`

### Actual behavior

SWC adds trailing coma: `{ a, b, ...rest,}`, which isn't valid syntax

### Version

1.3.21

### Additional context

The issue goes away if trailing coma from types is removed.
```
export function func({
  a,
  b,
  ...rest
}: {
  a: string,
  b: string // (no coma here)
}) {
 console.log(a,b, rest)
}
```

alexander-akait:
[...]

PR #6610, opened 2022-12-09
fix(ecma/codegen): don't print trailing coma for rest argument

alexander-akait:
**Description:**

https://github.com/swc-project/swc/issues/6589

**BREAKING CHANGE:**

No

Fixed https://github.com/swc-project/swc/issues/6589

### DIFF
--- a/crates/swc_ecma_codegen/src/lib.rs
+++ b/crates/swc_ecma_codegen/src/lib.rs
@@ -2481,23 +2481,12 @@ where
         self.emit_leading_comments_of_span(node.span(), false)?;
 
         srcmap!(node, true);
-
-        let is_last_rest = match node.props.last() {
-            Some(ObjectPatProp::Rest(..)) => true,
-            _ => false,
-        };
-        let format = if is_last_rest {
-            ListFormat::ObjectBindingPatternElements ^ ListFormat::AllowTrailingComma
-        } else {
-            ListFormat::ObjectBindingPatternElements
-        };
-
         punct!("{");
 
         self.emit_list(
             node.span(),
             Some(&node.props),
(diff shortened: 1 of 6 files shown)


crates/swc_ecma_codegen/tests/fixture/issue-6589/js/output.js
```diff
@@ -0,0 +1,14 @@
...
+const other = {
+    unknow: "foo"
+};
+let foo = {
+    ...other,
+    bar: "baz"
+};
+let bar = {
+    bar: "baz",
+    ...other
```
alexander-akait, line 13: ES2022 allows trailing comma here, but we don't have option to force enable/disable it (or we can print it when it supported, but currently we don't use this logic), so I think we should solve it in the separated PR

crates/swc_ecma_codegen/src/lib.rs
```diff
@@ -2481,23 +2481,12 @@ where
...
-
-        let is_last_rest = match node.props.last() {
-            Some(ObjectPatProp::Rest(..)) => true,
-            _ => false,
-        };
-        let format = if is_last_rest {
-            ListFormat::ObjectBindingPatternElements ^ ListFormat::AllowTrailingComma
-        } else {
-            ListFormat::ObjectBindingPatternElements
-        };
```
alexander-akait, line 2493: Ideally this logic should be in `emit_list` or in parent, not in `ObjectPat` itself, because `ObjectPat` doesn't know parent context

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_codegen

Issue #6601, opened 2022-12-08
Syntax Error thrown on valid typescript input


FelixSchuSi:
### Describe the bug

SWC seems to get confused when using type annotations within a ternary operation.

The type annotation in line 4 of the reproduction example causes swc to wrongly throw a syntax error.

### Input code

```typescript
function exampleFunction() {
  return Math.random() > 0.5
    ? `<button               
        @click="${(): void => console.log('this line causes a syntax error')}"
      ></button>`
    : `<button               
        @click="${(): void => console.log('this line does NOT causes a syntax error')}"
      ></button>`;
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2022"
  },
  "module": {
    "type": "es6"
  },
  "isModule": false
}
```

### Playground link

https://play.swc.rs/?version=1.3.21&code=H4sIAAAAAAAAA62OvQ6CMBSFd57ixJgAC7q4IFQnN3XxAailSmNpTX8MhvDuNoKDm4N3uveenPOdi1fMCa3AO9reJd9Nd5KijwDDnTcKe%2BqazFBV6zYIBMtsFURgg6o4e%2BeC%2F3uiz7JlUrBbOZv3SZrjoUWNkoBpZbXkmdTXJHaNsJBCcTDqLbegsE%2FlaAdujDZxOsymOFIsRhqp3p%2F8v%2FhaB%2FjheBp7%2FFRjHQ0vUdsGlUIBAAA%3D&config=H4sIAAAAAAAAAy2LsQ2AMAwEd3FNgVJQZAeGsIJBQQEi20hEEbvjAN3%2F%2FV%2BFVQL4ChlZiFuSsite4EFLJgkcs0IHKoZmTEK3FeSF1C4krncODG3HdCZqftPeaWg8yvgvn%2FwA6RS7OnEAAAA%3D

### Expected behavior

No syntax error because the input is valid typescript.

### Actual behavior

A syntax error is thrown.

### Version

1.3.21

PR #6611, opened 2022-12-09
fix(ecma/parser): parse types in `CallExpression` inside templates

alexander-akait:
**Description:**

https://github.com/swc-project/swc/issues/6601

**BREAKING CHANGE:**

No

Fixed https://github.com/swc-project/swc/issues/6601

### DIFF
--- a/crates/swc_ecma_parser/src/parser/expr.rs
+++ b/crates/swc_ecma_parser/src/parser/expr.rs
@@ -378,8 +378,13 @@ impl<I: Tokens> Parser<I> {
                 }
 
                 tok!('`') => {
+                    let ctx = Context {
+                        will_expect_colon_for_cond: false,
+                        ..self.ctx()
+                    };
+
                     // parse template literal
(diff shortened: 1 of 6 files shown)


crates/swc_ecma_parser/src/parser/expr.rs
```diff
@@ -841,7 +844,6 @@ impl<I: Tokens> Parser<I> {
         let return_type = if !self.ctx().will_expect_colon_for_cond
             && self.input.syntax().typescript()
             && is!(self, ':')
-            && !self.ctx().dont_parse_colon_as_type_ann
```
alexander-akait, line 844: I think we don't need `will_expect_colon_for_cond`, because `will_expect_colon_for_cond` is used to check should we parse `:` as type or not, also I found we duplicate logic, i.e. set `will_expect_colon_for_cond` and `dont_parse_colon_as_type_ann` to the same value

kdy1 (approved):
Thank you!

---
swc-bump:
 - swc_ecma_parser

PR #6615, opened 2022-12-11
fix(node-swc): Add missing `const_to_let` to the type

SukkaW:
**Description:**

https://github.com/swc-project/swc/blob/a5efa8af503c3e570b36248b7a05bfa393c82f0b/crates/swc_ecma_minifier/src/option/terser.rs#L243-L244

SWC has `const_to_let`, but it is missing in the type.

**BREAKING CHANGE:**


kdy1 (approved):
swc-bump:
 - dbg-swc

PR #6617, opened 2022-12-11
fix(css/minifier): transform function compressions

alexander-akait:
**Description:**

Fix regression, looks I missed that change before and so we have regression

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_css_minifier

PR #6621, opened 2022-12-11
fix(es/codegen): fix issue with typescript's mapped types output

### LINKED ISSUE #6616: swc_ecma_codegen emits `/` in some mapped types
### Describe the bug

I'm using swc_ecma_codegen directly to emit a .d.ts file.  It should look like this:
```
declare type Custom<A> = {
    -readonly [P in keyof A]+?: A[P];
}
```
but the actual output was:
```
declare type Custom<A> = {
    -readonly [P in keyof A]+/: A[P];
}
```

### Input code

_No response_

### Config

_No response_

### Playground link

_No response_

### Expected behavior

The following should be output 

### Actual behavior

_No response_

### Version

swc_ecma_codegen = 0.127.16

### Additional context

Updating https://github.com/swc-project/swc/blob/main/crates/swc_ecma_codegen/src/typescript.rs#L508 to output a `?` instead of `/` should fix the issue.  I can work on a PR for this.

kevinbarabash:
**Description:**

In some situations the code generated for mapped types would include a `/` instead of a `?`.

**BREAKING CHANGE:**

This is a small bug fix and shouldn't be breaking change.

fixes #6616

### DIFF
--- a/crates/swc_ecma_codegen/src/typescript.rs
+++ b/crates/swc_ecma_codegen/src/typescript.rs
@@ -505,7 +505,7 @@ where
                 }
                 TruePlusMinus::Plus => {
                     punct!("+");
-                    punct!("/");
+                    punct!("?");
                 }
                 TruePlusMinus::Minus => {
                     punct!("-");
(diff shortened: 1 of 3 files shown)


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_codegen

PR #6626, opened 2022-12-12
feat(css/compat): Add `Compiler`

kdy1:
**Description:**

Visitor patterns are inherently slow, so I decided to merge all compatibility passes into a single pass.


kdy1 (commented):
swc-bump:
 - swc_css_compat

PR #6632, opened 2022-12-13
fix(css/minifier): small bug

alexander-akait:
**Description:**

Small bug fix - don't touch nothing inside `@supports` (because it can be used to detect features supported by browser)

**BREAKING CHANGE:**

No

No


crates/swc_css_minifier/src/compressor/mod.rs
```diff
@@ -36,7 +36,7 @@ pub fn compressor() -> impl VisitMut {
 struct Compressor {
     ctx: Ctx,
     need_utf8_at_rule: bool,
-    in_supports_conidition: bool,
+    in_supports_condition: bool,
```
alexander-akait, line 39: typo

kdy1 (approved):
swc-bump:
 - swc_css_minifier

PR #6633, opened 2022-12-13
chore(es/minifier): expose `globals_defs` pass

iheyunfei:
**Description:**

I'm trying to do some global expression replacement and found I could reuse this part. Would it be ok to expose this function?

**BREAKING CHANGE:**

**Related issue (if exists):**

PR #6634, opened 2022-12-13
chore: export minifier global_defs

underfin:
**Description:**

The `global_defs` is a common visitor, export it to let another library can reference it.

**BREAKING CHANGE:**


kdy1:
Are you aware of https://github.com/swc-project/swc/pull/6633 ?

underfin:
Oh. thanks.

PR #6645, opened 2022-12-14
feat(css/prefixer): nested calc

alexander-akait:
**Description:**

Port - https://preset-env.cssdb.org/features/#nested-calc

Also bug fix - always use `(` and `)` for prefixed

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_css_prefixer

PR #6657, opened 2022-12-14
feat(css/parser): normalize selector AST

alexander-akait:
**Description:**

Normalize selector AST where it is allowed by spec

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_css_parser

PR #6663, opened 2022-12-15
feat(css/parser): normalize declaration and important

alexander-akait:
**Description:**

Normalize declartion and important, also found a bug in `swc_css_prefixer` (because we don't check case) and fixed

**BREAKING CHANGE:**

No

No

### DIFF
--- a/crates/swc_css_codegen/src/lib.rs
+++ b/crates/swc_css_codegen/src/lib.rs
@@ -2763,9 +2763,9 @@ fn minify_hex_color(value: &str) -> String {
             if length == 6 || chars[6] == b'f' && chars[7] == b'f' {
                 let mut minified = String::with_capacity(3);
 
-                minified.push((chars[0] as char).to_ascii_lowercase());
-                minified.push((chars[2] as char).to_ascii_lowercase());
-                minified.push((chars[4] as char).to_ascii_lowercase());
+                minified.push(chars[0] as char);
+                minified.push(chars[2] as char);
+                minified.push(chars[4] as char);
 
                 return minified;
             }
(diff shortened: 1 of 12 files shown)


crates/swc_css_codegen/src/lib.rs
```diff
@@ -2763,27 +2763,27 @@ fn minify_hex_color(value: &str) -> String {
...
                 return minified;
             }
             // 8 -> 4
             else if length == 8 && chars[6] == chars[7] {
                 let mut minified = String::with_capacity(4);
 
-                minified.push((chars[0] as char).to_ascii_lowercase());
-                minified.push((chars[2] as char).to_ascii_lowercase());
-                minified.push((chars[4] as char).to_ascii_lowercase());
-                minified.push((chars[6] as char).to_ascii_lowercase());
```
alexander-akait, line 2779: Forgot to remove it too, we don't need it, because we already normalize hex colors

crates/swc_css_codegen/src/lib.rs
```diff
@@ -2763,27 +2763,27 @@ fn minify_hex_color(value: &str) -> String {
...
+                minified.push(chars[4] as char);
+                minified.push(chars[6] as char);
 
                 return minified;
             }
         }
     }
 
-    value.to_ascii_lowercase()
+    value.to_string()
```
alexander-akait, line 2786: Ideally rewrite it on `Cow`, but let's do it in the separate PR

kdy1 (approved):
swc-bump:
 - swc_css_codegen
 - swc_css_parser
 - swc_css_minifier
 - swc_css_prefixer

PR #6666, opened 2022-12-15
test(css/parser): check we don't have a regression

alexander-akait:
**Description:**

I have an internal report (on twitter) that we've already had issues with this, just check we don't have an issue anymore

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - dbg-swc

PR #6667, opened 2022-12-15
feat(css/parser): normalize and improve function name

alexander-akait:
**Description:**

Improve function name:
1. Normalize name (lowercase)
2. Use dashed ident (we use the same for at-rules/declarations)

From spec (https://www.w3.org/TR/css-values-4/#dashed-idents):
> CSS will use [<dashed-ident>](https://www.w3.org/TR/css-values-4/#typedef-dashed-ident) more in the future, as more author-controlled syntax is added. CSS authoring tools, such as preprocessors that turn custom syntax into standard CSS, should use <dashed-ident> as well, to avoid clashing with future CSS design.
For example, if a CSS preprocessor added a new "custom" at-rule, it shouldn’t spell it @custom, as this would clash with a future official @custom rule added by CSS. Instead, it should use @--custom, which is guaranteed to never clash with anything defined by CSS.

3. use `js_word!()`

**BREAKING CHANGE:**

Yes

No


alexander-akait:
Rebased

crates/swc_css_ast/src/base.rs
```diff
@@ -100,12 +101,39 @@ impl Take for SimpleBlock {
     }
 }
 
+#[ast_node]
+#[derive(Eq, Hash, Is, EqIgnoreSpan)]
+pub enum FunctionName {
```
kdy1, line 106: Why is this required?

alexander-akait: @kdy1 Yeah, it reduce code, we have the same for `AtRule` and `Declaration`, so we will have less code

alexander-akait:
@kdy1 Rebased :+1:

crates/swc_css_ast/src/lib.rs
```diff
@@ -32,3 +32,26 @@ macro_rules! matches_eq_ignore_ascii_case {
...
+/// ```
+/// use swc_atoms::JsWord;
+/// use swc_atoms::js_word;
+/// use swc_css_ast::*;
+///
+/// assert!(matches_eq!(JsWord::from("a"), js_word!("a")));
+/// assert!(matches_eq!("a", "a"));
+/// ```
+#[macro_export]
+macro_rules! matches_eq {
```
kdy1, line 51: I think this is not required. We can use `matfches!` instead

crates/swc_css_minifier/src/compressor/mod.rs
```diff
@@ -341,8 +341,8 @@ impl VisitMut for Compressor {
     }
 
     fn visit_mut_function(&mut self, n: &mut Function) {
-        if matches_eq_ignore_ascii_case!(
-            n.name.value,
+        if matches_eq!(
```
kdy1, line 344: Please use `matches!` instead

crates/swc_css_compat/src/compiler/color_space_separated_parameters.rs
```diff
@@ -12,11 +12,12 @@ impl Compiler {
         n: &mut AbsoluteColorBase,
     ) {
         if let AbsoluteColorBase::Function(function) = n {
-            let name = function.name.value.to_ascii_lowercase();
-
-            if !matches!(
-                name,
-                js_word!("rgb") | js_word!("rgba") | js_word!("hsl") | js_word!("hsla")
+            if !matches_eq!(
```
kdy1, line 15: Ditto

kdy1 (changes requested):

kdy1 (approved):
swc-bump:
 - swc_css_ast --breaking

PR #6674, opened 2022-12-18
refactor(css/codegen): do not use `raw`

alexander-akait:
**Description:**

Refactor codegen and do not use `raw`, also:
1. fix bug with printing `e\32px` dimension (we need to escape as in spec)
2. do not parser wrong hash token as color
3. fix especaping in `"` in `Str`
4. fix serialization in raw URL
5. support `element()` function (before we parse it as hash, now as id selector)

**BREAKING CHANGE:**

Yes, we don't use `raw` in codegen, except printing tokens, so if you want to inject something non standard, you should use `ListOfComponentValues` with `PreservedToken`

No


crates/swc_css_ast/src/base.rs
```diff
@@ -190,6 +190,8 @@ pub enum ComponentValue {
     SupportsCondition(Box<SupportsCondition>),
     #[tag("Declaration")]
     Declaration(Box<Declaration>),
+    #[tag("IdSelector")]
+    IdSelector(Box<IdSelector>),
```
kdy1, line 194: Is this change really required?

alexander-akait:
@kdy1 Rebased too

kdy1 (dismissed):
swc-bump:
 - swc_css_ast --breaking

alexander-akait:
@kdy1 Rebased too

kdy1 (changes requested):
CI failed

alexander-akait:
@kdy1 Fixed too :+1:

kdy1 (approved):

PR #6676, opened 2022-12-18
feat(css/parser): normalize at-rule name

alexander-akait:
**Description:**

Normalize at-rule name

**BREAKING CHANGE:**

No

No


kdy1 (approved):
swc-bump:
 - swc_css_parser

PR #6686, opened 2022-12-19
refactor(css/compat): union nesting and compiler

alexander-akait:
**Description:**

Union nesting and compiler, it will be faster and we have the option for nesting (before it doesn't work).

We still have many TODOs for nesting, I want to start fix them after merge

**BREAKING CHANGE:**

Yes

No


crates/swc_css_compat/src/utils.rs
```diff
@@ -0,0 +1,5 @@
+use swc_css_ast::{ComponentValue, Rule};
+
+pub(crate) fn rule_to_component_value(rule: Rule) -> ComponentValue {
```
kdy1, line 3: Is this required? If you need to use it with `.map`, you can do `.map(From::from)` intead.

alexander-akait: Just copy/paste from old code, yeah, let's fix it

alexander-akait:
@kdy1 Fixed and rebased

PR #6708, opened 2022-12-21
fix(cli): Use the compile --source-maps option

realtimetodie:
Fixes an issue where the source maps are not generated when using the cli compile command with the `--source-maps` option and without providing a configuration file.

The validation handling is passed down

https://github.com/swc-project/swc/blob/ef1d966b5746d97bfb5b27a7330915b8d86806dc/crates/swc/src/config/mod.rs#L255-L258

Example

```bash
$ swc compile --source-maps=true --source-map-target=output.js.map --out-file=output.js input.js
```

Actual behavior

```
output.js
```

Expected behavior

```
output.js
output.js.map
```

Related

* https://github.com/aspect-build/rules_swc/issues/88


bindings/swc_cli/src/commands/compile.rs
```diff
@@ -266,28 +266,32 @@ struct InputContext {
...
         if let Some(env_name) = &self.env_name {
-            ret.env_name = env_name.to_string();
+            options.env_name = env_name.to_string();
+        }
+
+        if let Some(source_maps) = &self.source_maps {
+            options.source_maps = Some(match source_maps.as_str() {
+                "false" => SourceMapsConfig::Bool(false),
+                "true" => SourceMapsConfig::Bool(true),
+                value => SourceMapsConfig::Str(value.to_string()),
```
kwonoj, line 290: I'm curious if we need to validate if value is acceptable and warn?

realtimetodie: As mentioned above, the validation handling is passed down

https://github.com/swc-project/swc/blob/ef1d966b5746d97bfb5b27a7330915b8d86806dc/crates/swc/src/config/mod.rs#L255-L258

kwonoj: that's assert: doesn't it panic instead of gracefully error out?

realtimetodie: Yes, it does. But it would also panic now, if an invalid value is set in the configuration file.

kwonoj: I still think it should be handled and emit proper warning instead of error, but probably not a blocker.

kdy1 (approved):
swc-bump:
 - dbg-swc

kdy1:
@realtimetodie Can you check `allow edit by maintainers`? I think it's required for the bot to auto-rebase

realtimetodie:
Yes

PR #6714, opened 2022-12-24
fix(cli): Skip empty stdin in non tty environments

realtimetodie:
Fixes an issue where an error is thrown when using the compile command with a file input and a stdin of length zero.

The [`atty::is`](https://docs.rs/atty/0.2.14/atty/fn.is.html) function only validates whether a stream source is connected to a tty. If no tty is connected, the function returns true and the stdin gets processed.

https://github.com/swc-project/swc/blob/d8d9965aa474e3c57d8712d99935673bb28fb59d/bindings/swc_cli/src/commands/compile.rs#L244-L246

Example (in a non tty environment)

```bash
$ swc compile input.js --out-file=output.js
```

Actual behavior

```
Error: Cannot specify inputs from stdin and files at the same time
```

Related

* https://github.com/aspect-build/rules_swc/issues/88


realtimetodie:
Alternative implementation. Please clarify which one you prefer.

```rust
fn collect_stdin_input() -> Option<String> {
    if atty::is(atty::Stream::Stdin) {
        return None;
    }

    let mut lines = io::stdin()
        .lock()
        .lines()
        .map(|line| line.expect("Not able to read stdin"));

    if lines.next().is_some() {
        Some(lines.collect::<Vec<String>>().join("\n"))
    } else {
        None
    }
}
```

kwonoj:
I'm actually thinking if we should be explicit for tty like `swcx --stdin`, as current behavior is mostly legacy from current cli.

realtimetodie:
Yes, I can take of that in a subsequent PR. Please let me know how to proceed.

realtimetodie:
@kwonoj This is currently blocking https://github.com/aspect-build/rules_swc/issues/88. Please let me know how to proceed.

realtimetodie:
I would like to offer my own opinion. I find the way the cli works at the moment more intuitive and user-friendly. Reading from stdin is the default mode and user does not need to set an option to enable stdin. This is the case for most unix command line programs. Why complicate things. My bugfix simply verifies, that the stdin buffer is not empty.

@kdy1

kdy1:
@kwonoj Thoughts?

kwonoj:
I'm still intrigued to have an explicit option, however it doesn't have to be in this PR and probably think later. PR as-is probably ok to go for now.

kdy1 (approved):
Thank you!

---

swc-bump:
 - dbg-swc

PR #6743, opened 2023-01-03
fix(swc): example

RiESAEX:
**Description:**
fix example of swc
I think it was broken by https://github.com/swc-project/swc/pull/5975

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

---

swc-bump:
 - dbg-swc

Issue #6729, opened 2022-12-29
Unknown minify bug, causes infinite run


JSerFeng:
### Describe the bug

Simple code that makes SWC hang for ever.

## How to reprod ?

Go to the playground link below, then enable `compress`, you will find browser blocks

### Input code

```typescript
export async function foo() {
  if (undefined_var_1) {
    let replace;
  
    if (undefined_var_2) {
      replace = 1;
    } else {
      replace = 2;
    }
  
    await a({ replace })
  }
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "decorators": true,
      "tsx": false
    },
    "target": "es2020",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": true,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.3.24&code=H4sIAAAAAAAAA22OSwqEMBAF9znFW8albsWzSBM7EAgdiXFUJHf3m9nMbKuKx%2BN1DDGBpk0M7CwmuSCwIegKuwKchZ5lYOuEh%2F5Dsa8fAXhOiDx6Mtye4Ga%2FeVNylBgd6vZGGewn%2FuOb15dZWsidJ%2FX%2BbXKlLp%2FVATGG0ebBAAAA&config=H4sIAAAAAAAAA0WOPQ7DIAxG7%2BKZIcrQgTv0EBZxIioCyHakIsTdC0mqbP5537MrfMSBrZCRhXhUUqLiFyxoySSOfVYwsJBLjJpYwCofZEClQysGodYb5I20h0jmaZ56IKQkdAMGdh%2F9WobepT0ziTwrjFv4k6279rQcY1DPD07nC9rjuM57ed%2Fc6NsPsPYZ%2F8kAAAA%3D

### Expected behavior

Normally get result.

### Actual behavior

SWC running without error and don't stop.

### Version

1.3.24

JSerFeng:
I might have found reason, Enable `unused` and `reduce_vars` both will reproduce the problem

PR #6744, opened 2023-01-03
fix(es/minifier): Fix usage counter to fix infinite loop

kdy1:
**Description:**

We skip non-computed property names while checking if we can inline an expression.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6729.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

jridgewell (approved):

Issue #6694, opened 2022-12-20
Windows line endings break source maps


mischnic:
### Describe the bug

A block comment before a function produces invalid sourcemaps

![Bildschirm­foto 2022-12-20 um 18 24 14](https://user-images.githubusercontent.com/4586894/208728187-2c0d6475-f971-411b-a160-f07e472ab299.png)

### Input code

(See https://github.com/swc-project/swc/issues/6694#issuecomment-1360240667, this is due to Windows line endings)

```typescript
/**
 * foo
 * @param data foo
 * @returns foo
 */
export const fixupRiskConfigData = (data: any): types.RiskConfigType => {
  if (x) {
    return 123;
  } else {
    return 456;
  }
};
```

### Config

```json
{
	"jsc": {
		"parser": {
			"syntax": "typescript"
		},
		"target": "es2022"
	},
	"sourceMaps": true,
	"inlineSourcesContent": true
}
```

### Expected behavior

The sourcemap should be correct

### Version

1.3.23

### Additional context

Found in https://github.com/parcel-bundler/parcel/issues/8691

jridgewell:
[...]

PR #6752, opened 2023-01-05
fix(common): Mark `\r` as a line ending

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6694.


kdy1 (commented):
swc-bump:
 - swc_common

crates/swc_common/src/syntax_pos/analyze_source_file.rs
```diff
@@ -47,167 +48,6 @@ pub fn analyze_source_file(
     (lines, multi_byte_chars, non_narrow_chars)
 }
 
-cfg_if::cfg_if! {
```
kdy1, line 50: This is not ideal, but I'm not sure if this optimization is really required.

crates/swc_common/src/syntax_pos/analyze_source_file.rs
```diff
@@ -240,7 +80,7 @@ fn analyze_source_file_generic(
             let pos = BytePos::from_usize(i) + output_offset;
 
             match byte {
-                b'\n' => {
+                b'\n' | b'\r' => {
```
jridgewell, line 83: I think we need to handle `\r\n` as a single line ending and not two line endings (unless that's been normalized as some prior point I'm not seeing).

jridgewell (changes requested):

crates/swc_common/src/syntax_pos/analyze_source_file.rs
kdy1: Did it, and I'll add some tests for it.

crates/swc_common/src/syntax_pos/analyze_source_file.rs
```diff
@@ -240,6 +80,14 @@ fn analyze_source_file_generic(
             let pos = BytePos::from_usize(i) + output_offset;
 
             match byte {
+                b'\r' => {
+                    lines.push(pos + BytePos(1));
```
jridgewell, line 84: Won't this need to be `+ BytePos(2)` if we're hitting a `\r\n`?

crates/swc/tests/source_map.rs
```diff
@@ -424,3 +426,77 @@ fn issue_4578() {
...
+            Ok(result) => {
+                println!("{}", result.code);
+
+                assert!(result.map.is_some());
+                let map = result.map.unwrap();
+
+                let source_map = sourcemap::SourceMap::from_slice(map.as_bytes())
+                    .expect("failed to deserialize sourcemap");
+
+                dbg!(&source_map);
```
jridgewell, line 485: Are the debug and prints supposed to still be here?

jridgewell (changes requested):

jridgewell (dismissed):

Issue #6767, opened 2023-01-08
Inserting node with DUMMY_SP reuses (next) existing source mapping


mischnic:
### Describe the bug

When a node with `DUMMY_SP` is inserted, it should create no source mapping, but it does.

```rust
                module.body.splice(
                    0..0,
                    vec![ModuleItem::ModuleDecl(ModuleDecl::Import(ImportDecl {
                        specifiers: vec![],
                        asserts: None,
                        span: DUMMY_SP,
                        src: Box::new("new".into()),
                        type_only: false,
                    }))],
                );
```

### Input code

```typescript
/*------------------------------------
  ------------------------------------*/
var _a;
import z from "./z.js";
console.log("xyz", z);
```

### Config

```rust
use std::collections::HashSet;

use swc::{self};
use swc_atoms::JsWord;
use swc_common::{
    chain,
    comments::SingleThreadedComments,
    errors::{DiagnosticBuilder, Emitter, Handler},
    source_map::SourceMapGenConfig,
    sync::Lrc,
    FileName, Globals, Mark, SourceMap, SyntaxContext, DUMMY_SP,
};
use swc_ecma_ast::{
    CatchClause, Decl, Id, Ident, ImportDecl, Module, ModuleDecl, ModuleItem, Pat, Stmt,
    VarDeclKind,
};
use swc_ecma_codegen::{text_writer::JsWriter, Config};
use swc_ecma_parser::{lexer::Lexer, EsConfig, PResult, Parser, StringInput, Syntax, TsConfig};
use swc_ecma_preset_env::{preset_env, BrowserData, Mode, Targets, Version};
use swc_ecma_transforms::{
    compat::reserved_words::reserved_words,
    fixer,
    fixer::paren_remover,
    helpers, hygiene,
    optimization::simplify::{dead_branch_remover, expr_simplifier},
    react::{self, Runtime},
    resolver,
};
use swc_ecma_visit::{Fold, FoldWith, Visit, VisitWith};

#[derive(Debug, Clone, Default)]
pub struct ErrorBuffer(std::sync::Arc<std::sync::Mutex<Vec<swc_common::errors::Diagnostic>>>);

type SourceMapBuffer = Vec<(swc_common::BytePos, swc_common::LineCol)>;

impl Emitter for ErrorBuffer {
    fn emit(&mut self, db: &DiagnosticBuilder) {
        self.0.lock().unwrap().push((**db).clone());
    }
}

fn main() {
[...]

PR #6773, opened 2023-01-10
fix(es/codegen): Skip `lo` of a program if body is not empty

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6767.


kdy1 (commented):
reset ci


---

swc-bump:
 - swc_ecma_codegen

kdy1:
@jridgewell Do you think this is correct?
The actual problem of #6767 is that the `span.lo` of `Module` does not match `span.lo` of the first item.

jridgewell (dismissed):

jridgewell (commented):
I think this solves this particular case, but the larger solution would be to to mark `DUMMY_SPAN` locations with a source-less mapping. We'd need to remove the dummy checks from https://github.com/swc-project/swc/blob/b102504eb9e0afc5a79989c9cd8a3952ab1e2689/crates/swc_html_codegen/src/writer/basic.rs.

kdy1:
I tried it, but it resulted in extremely huge source maps.
So I reverted it.

The problem was that so many nodes have `DUMMY_SP`

jridgewell:
We can optimize source maps output, but we still need all locations:
- https://github.com/jridgewell/gen-mapping/blob/9ff6f25fa9bd12eeb9d265a17dd7631133669fdc/src/gen-mapping.ts#L420-L458
- https://github.com/jridgewell/babel/blob/518e1a6de4e05613ebf3aaf2c35e598e1dbcdb63/packages/babel-generator/src/source-map.ts#L82-L101

crates/swc_ecma_codegen/src/lib.rs
```diff
@@ -94,7 +94,10 @@ where
     #[emitter]
     pub fn emit_module(&mut self, node: &Module) -> Result {
         self.emit_leading_comments_of_span(node.span(), false)?;
-        srcmap!(node, true);
+
+        if node.body.is_empty() {
```
kdy1, line 98: Note: This is still required to fix #6767

crates/swc/tests/fixture/sourcemap/issue-3854/1-true/output/index.map
```diff
@@ -1,5 +1,5 @@
...
...qBAAA,CAiQe,IAAI,EAAEe,AAjQrB,YAAA,EAiQgCf,AAjQhC,KAAA,AAiQqC;IACjC;IAZAiB,YAAY,GAAG5B,AAtPnB,KAAA,AAsPkD,CAAE;QAC5C,KAAK;QAHT,AApPJ,sBAAA,CAAA,IAAA,EAoPI,AApPJ,YAAA;YAAA,QAAA,EAAA,IAAA;YAAA,KAAA,EAoP2C,IAAII,AApP/C,uBAAA;;QAwPQ,IAAI,CAACJ,AAxPb,KAAA,AAwPkB,IAAIA,AAxPtB,KAAA;IAyPI;AAUJ;IAGI,AAtQJ,OAAA,qBAAA,OAAA,IAuQI,AAvQJ,QAAA,qBAAA,OAAA;AAqQA,MAAME,AArQN,qBAAA,SAqQoCG,AArQpC,iBAAA,CAAA,eAAA,AAqQmD;IAI/CL,AAzQJ,KAAA,CAyQU,GAAGA,AAzQb,KAAA,AAyQ4C,EAAQ;QAzQpD,qBAAA,CAAA,IAAA,EA0QcA,AA1Qd,OAAA,EA0QsBA,AA1QtB,KAAA,CA0Q4BO,AA1Q5B,GAAA,AA0Q+B,CAACC,AA1QhC,aAAA,CAAA,WAAA,AA0Q2C;QACnC,OAAO,IAAI;IACf;IAEAK,AA9QJ,MAAA,CA8QW,GAAGA,AA9Qd,MAAA,AA8Q8C,EAAQ;QA9QtD,qBAAA,CAAA,IAAA,EA+QcA,AA/Qd,QAAA,EA+QuBA,AA/QvB,MAAA,CA+Q8BN,AA/Q9B,GAAA,AA+QiC,CAACC,AA/QlC,aAAA,CAAA,WAAA,AA+Q6C;QACrC,OAAO,IAAI;IACf;IAEAG,AAnRJ,KAAA,GAmR4B;QACpB,OAAO;YACHX,AArRZ,KAAA,AAqRiB,EArRjB,qBAAA,CAqRmB,IAAI,EAAEA,AArRzB,OAAA;YAsRYa,AAtRZ,MAAA,AAsRkB,EAtRlB,qBAAA,CAsRoB,IAAI,EAAEA,AAtR1B,QAAA;QAuRQ;IACJ;gBAxRJ,GAAA,IAAA;QAAA,KAAA,IAAA,IAAA;QAsQI,AAtQJ,sBAAA,CAAA,IAAA,EAsQI,AAtQJ,OAAA;YAAA,QAAA,EAAA,IAAA;YAAA,KAAA,EAsQ8B,EAAE;;QAC5B,AAvQJ,sBAAA,CAAA,IAAA,EAuQI,AAvQJ,QAAA;YAAA,QAAA,EAAA,IAAA;YAAA,KAAA,EAuQ+B,EAAE;;;AAkBjC",
```
kdy1, line 2: @jridgewell Is this better than the previous one?

kdy1:
I'll use a separate PR.

jridgewell (approved):

PR #6795, opened 2023-01-12
refactor: unnecessary `GLOBALS.set` and `GLOBALS.with`

IWANABETHATGUY:
**Description:**
1. `GLOBALS.with` will panic if inner is not set, the `GLOBALS.with` and reset with `GLOBALS.set` seems a little redundant. 
2. Also it would cause panic when you only use `dce` visitor with feature `concurrent_minifier` enable.
image

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_transforms_base/src/perf.rs
+++ b/crates/swc_ecma_transforms_base/src/perf.rs
@@ -1,13 +1,9 @@
 use rayon::prelude::{IntoParallelIterator, ParallelIterator};
 use swc_common::util::move_map::MoveMap;
 #[cfg(feature = "concurrent")]
-use swc_common::{errors::HANDLER, GLOBALS};
 use swc_ecma_ast::*;
 pub use swc_ecma_utils::parallel::*;
 use swc_ecma_visit::{Fold, FoldWith, Visit, VisitMut, VisitMutWith, VisitWith};
-
-#[cfg(feature = "concurrent")]
-use crate::helpers::HELPERS;
 
 pub trait Check: Visit + Default {
     fn should_handle(&self) -> bool;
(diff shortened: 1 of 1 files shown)


crates/swc_ecma_transforms_base/src/perf.rs
```diff
@@ -50,38 +47,26 @@ where
         N: Send + Sync + VisitWith<Self>,
     {
         if nodes.len() >= threshold || option_env!("SWC_FORCE_CONCURRENT") == Some("1") {
-            GLOBALS.with(|globals| {
```
IWANABETHATGUY, line 53: process will panic if `GLOBALS` is not `set` before, if it is set we reset the `GLOBALS` looks a little redundant.

crates/swc_ecma_transforms_base/src/perf.rs
```diff
@@ -50,38 +47,26 @@ where
...
-                HELPERS.with(|helpers| {
-                    HANDLER.with(|handler| {
-                        use rayon::prelude::*;
-
-                        let visitor = nodes
-                            .into_par_iter()
-                            .map(|node| {
-                                GLOBALS.set(globals, || {
```
IWANABETHATGUY, line 61: the `globals`,  from `GLOBALS.with` they are just the same thing, reset the `GLOBALS` here is also a little redundant

kdy1 (changes requested):
I'll not accept this PR.
This is a documented API.

kdy1:
If this is a problem, you should provide `GLOBALS` and `HELPERS`

IWANABETHATGUY:
Actually I provide GLOBALS and HELPERS, lol

IWANABETHATGUY:
Alright, I would try to provide a repo

kdy1:
The closure passed to `.map` of a parallel iterator does not have `GLOBALS` and `HELPERS` so it's not redundant at all.

IWANABETHATGUY:
Ah, the weird thing is I remove the nested `GLOBALS.set` and `GLOBALS.with` all tests passed and the panic just gone.

PR #6800, opened 2023-01-12
fix(es): Resolve plugin paths using config path as base

titanous:
**Description:**

This makes relative paths work for plugin references, they will be relative to the location of the config that imports the plugin.

Previously relative paths did not work, as the resolver attempted to resolve using the plugin path as the base and the target, resulting in mangled paths.

### DIFF
--- a/crates/swc/src/config/mod.rs
+++ b/crates/swc/src/config/mod.rs
@@ -554,6 +554,7 @@ impl Options {
                 experimental.plugins,
                 transform_metadata_context,
                 Some(plugin_resolver),
+                cfg.config_filename,
                 comments,
                 source_map,
                 unresolved_mark,
@@ -587,6 +588,7 @@ impl Options {
             crate::plugin::plugins(
                 experimental.plugins,
                 transform_metadata_context,
+                None,
                 None,
                 comments,
                 source_map,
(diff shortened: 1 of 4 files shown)


crates/swc/src/config/mod.rs
```diff
@@ -204,10 +204,6 @@ pub struct Options {
     #[serde(default = "default_swcrc")]
     pub swcrc: bool,
 
-    #[cfg(not(target_arch = "wasm32"))]
```
kdy1, line 207: Please restore this

kdy1 (changes requested):

crates/swc/src/config/mod.rs
titanous: Done.

kdy1:
@kwonoj Thoughts?

kwonoj:
I am not sure if we want this. The reason I left the current experimental resolver to absolute only is, it is unclear which point we should use for the relative resolve. Having swcrc / config as implicit root may work, but also some cases may not work as expected. 

Having base to resolve in plugin context seems ok to do, but public-facing interface seems need some thought.

titanous:
In the case of `rules_swc`, Bazel does not provide a way for us to provide absolute paths (and there's no `node_modules` to climb to), so we need some sort of relative path mechanism for plugins. Alternatively it could be relative to the current working directory that swc is called from. Given that relative paths don't work at all currently, and this entire config section is marked experimental, would you be open to merging this and seeing how it works out?

titanous:
I'd also be fine with a CLI flag or config setting explicitly specifying where relative paths should be resolved from. Implicitly using the config location seemed like good UX so I went with it, but any of these options are a couple lines of code for me to get working in my build at this point so I don't have a preference :)

titanous:
Closing in favor of adding a `--plugin` flag to `swc_cli`, which will work even better for `rules_swc`.

PR #6811, opened 2023-01-14
feat(bindings/cli): Add --plugin and --cache-root

titanous:
**Description:**

- Add `--plugin` CLI option that takes a `plugin/path={"config":"object"}` formatted argument and adds the plugin to the `jsc.experimental.plugins` list (can be used multiple times). If the plugin path is explicitly relative (starts with `./` or `../`), then it is converted to an absolute path using the current working directory as the base.
- Add a `--cache-root` CLI option that sets `jsc.experimental.cacheRoot`

This is an alternative to https://github.com/swc-project/swc/pull/6800 that works better for `rules_swc` and is more straightforward.

### DIFF
--- a/bindings/swc_cli/src/commands/compile.rs
+++ b/bindings/swc_cli/src/commands/compile.rs
@@ -1,7 +1,7 @@
 use std::{
     fs::{self, File},
     io::{self, Read, Write},
-    path::{Path, PathBuf},
+    path::{Component, Path, PathBuf},
     sync::Arc,
 };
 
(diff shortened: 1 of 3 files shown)


kwonoj:
Techinically, if swcx supports --config, specifying these can be done via --config, is it correct?

titanous:
There's a field for `--config` but I don't see any code implementing it.

kwonoj:
Yes, it is not implemented. My question is should we implement --config instead, by partially supporting plugin first instead of having new top level args.

titanous:
@kdy1 What do you think about implementing `--config` vs the approach in this PR?

kwonoj:
We'll need / have --config anyway, matter of time though. If we add top-level --plugin then once we have --config, what should happen? allow duplicate? or breaking changes to remove duplicated one?

titanous:
It's easy enough to just append if both flags are used. I can implement `--config` if that's preferred, I just want to get approval for the path forward before doing more work.

kdy1:
In this regard, I would like to follow @kwonoj 's opinion.

titanous:
Superceded by #6835.

PR #6812, opened 2023-01-14
feat(atoms): supports static atoms

hyfdev:
**Description:**

This PR supports allowing `Atom` to store `&'static str` in it.

**BREAKING CHANGE:**

#4946

### DIFF
--- a/crates/swc_atoms/src/lib.rs
+++ b/crates/swc_atoms/src/lib.rs
@@ -2,6 +2,8 @@
 
 #![allow(clippy::unreadable_literal)]
 
+mod tests;
+
 /// Not a publish API.
 #[doc(hidden)]
 pub extern crate once_cell;
@@ -284,10 +286,7 @@ impl<'de> serde::de::Deserialize<'de> for Atom {
 #[macro_export]
 macro_rules! atom {
     ($s:literal) => {{
-        static CACHE: $crate::once_cell::sync::Lazy<$crate::Atom> =
-            $crate::once_cell::sync::Lazy::new(|| $crate::Atom::new($s));
-
-        $crate::Atom::clone(&*CACHE)
+        $crate::Atom::new_static($s)
     }};
 }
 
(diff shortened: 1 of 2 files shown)


crates/swc_atoms/src/lib.rs
```diff
@@ -190,9 +219,9 @@ impl Ord for Atom {
 impl PartialEq for Atom {
     fn eq(&self, other: &Self) -> bool {
         // Fast path
```
iheyunfei, line 221: @kdy1 After supporting `Atom::static`, I'm not sure the fast path will work as before. To keep the logic the same as before, we need to call `match` twice. I'm going to remove the code, in case I miss something, what are your opinions?

crates/swc_atoms/src/lib.rs
```diff
@@ -44,12 +44,37 @@ include!(concat!(env!("OUT_DIR"), "/js_word.rs"));
...
+        ))
+    }
+
+    fn new_static(s: &'static str) -> Self {
+        Self::Static(s)
+    }
+}
+
+// fn _assert_size() {
+//     let _static_assert_size_eq = std::mem::transmute::<Atom, usize>;
```
iheyunfei, line 75: @kdy1 The `transmute` seems not to work, after making `Atom(ThinArc<HeaderWithLength<()>, u8>)` to `Atom(Inner)`.

Is this expected?
image

crates/swc_atoms/src/lib.rs
iheyunfei: Also, should `Atom` promise `assert(Atom::Arc("foo") == Atom::Static("foo"))`?

crates/swc_atoms/src/lib.rs
kdy1: Yes, but it will degrade performance greatly.

I didn't implement static support because it's hard

crates/swc_atoms/src/lib.rs
kdy1: I think it's fine, but `assert!(Atom::Arc("foo") == Atom::Static("foo"))` should work

crates/swc_atoms/src/lib.rs
iheyunfei: I get the gist. We need something like `ThinStaticStr`. It's indeed hard after looking into [thin_tr](https://github.com/thomcc/thin_str). I will try it, but it will take a while.

If anyone wants to try this, feel free to do it.

Issue #6814, opened 2023-01-14
Plugin CLI template generates wrong package.json main


alex-dixon:
### Describe the bug

From https://swc.rs/docs/plugin/ecmascript/getting-started:

```
swc plugin new --target-type wasm32-wasi my-first-plugin
```

Generates a project with this package.json:
```json
{
    "name": "my-first-plugin",
    "version": "0.1.0",
    "description": "",
    "author": "",
    "license": "ISC",
    "keywords": ["swc-plugin"],
    "main": "target/wasm32-wasi/release/my-first-plugin.wasm",
    "scripts": {
        "prepublishOnly": "cargo build-wasi --release"
    },
    "files": []
}
```

Note the hypens in `my-first-plugin` in the `main` field above.

`npm run prepublishOnly` outputs `target/wasm32-wasi/release/my_first_plugin.wasm` (note `_` instead of `-`).

This produces the following error when running the plugin with swc:
```
thread '<unnamed>' panicked at 'failed to invoke plugin: failed to invoke plugin on 'Some("input.js")'

Caused by:
    Impossible to get the node_modules path', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/swc-0.241.5/src/plugin.rs:228:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
node:internal/process/promises:279
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[Error: failed to handle: failed to invoke plugin: failed to invoke plugin on 'Some("input.js")'

Caused by:
    Impossible to get the node_modules path] {
  code: 'GenericFailure'
}
```

### Expected behavior

The generated plugin should have the correct main field.

### Actual behavior

Error described above.

### Version

SWC 0.91.12, @swc/core@1.3.26

PR #6815, opened 2023-01-14
fix(es/plugin): Fix starter template main field (#6814)

alex-dixon:
**Description:**
Replace `-` with `_` in generated package.json `main` field to match what cargo names the output `.wasm` file. Allows swc-core to load the plugin at runtime.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6814.


kdy1 (approved):
Thanks!

---

swc-bump:
 - dbg-swc

Issue #1388, opened 2021-02-09
cwd, source_root, and source_file_name options do not have any effect


kherock:
**Describe the bug**
I was trying to figure out why certain options in the CLI weren't working but it looks like the transform functions simply don't do anything with the options listed in the PR title.

**Input code**
```console
> swc = require('@swc/core');
> swc.transformSync('var foo = 1;\nexport { foo };\n', { sourceMaps: true, sourceRoot: '..', sourceFileName: 'foo'  })
{
  code: 'var foo = 1;\nexport { foo };\n',
  map: '{"version":3,"sources":["<anon>"],"sourcesContent":["var foo = 1;\\nexport { foo };\\n"],"names":[],"mappings":"IAAA,GAAA,GAAA,CAAA;SACA,GAAA"}'
}

> swc.transformFileSync('fixtures/amd.js', { sourceMaps: true, cwd: 'out', sourceRoot: '..' })
{
  code: 'var foo = 1;\nexport { foo };\n',
  map: '{"version":3,"sources":["fixtures/amd.js"],"sourcesContent":["\\nconst foo = 1;\\nexport { foo };"],"names":[],"mappings":"IACA,GAAA,GAAA,CAAA;SACA,GAAA"}'
}
> swc.transformFileSync('../fixtures/amd.js', { sourceMaps: true, cwd: 'out' })
thread '<unnamed>' panicked at 'failed to load file: Os { code: 2, kind: NotFound, message: "No such file or directory" }', native/src/transform.rs:158:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**Config**
```json
{
    "jsc": {
        "parser": {
            "syntax": "typescript"
        }
    },
    "module": {
        "type": "commonjs"
    }
}
```

**Expected behavior**
Here, I'd expect `"sources"` to not have an anonymous filename, and `"sourceRoot"` should be set to `".."`. When a `cwd` is passed to `transformFileSync`, the source file in the sibling directory should resolve with a path relative to `cwd`.

**Version**
The version of @swc/core: 1.2.47

**Additional context**
I'm currently fighting this when trying to use `@swc/cli` as a drop-in replacement for `tsc`, TypeScript generates its maps using paths relative to the output directory, but currently the only way to achieve the same effect with swc is by starting the process from the output directory itself.

realtimetodie:
I am working on this

Related
- https://github.com/swc-project/swc/pull/6817
[...]

PR #6817, opened 2023-01-15
feat(swc): Add a new source maps option "linked"

realtimetodie:
Adds a new source maps option "linked" to generate a source map file and append a sourceMappingURL comment to the end of the output, containing an URL to the generated source map file.

This is only implemented in `@swc/cli`

https://github.com/swc-project/cli/blob/933589ae907637afb7f0b0a43b7a24a7a920595e/src/swc/compile.ts#L31

Possible values for the --source-maps option

- **false** (default): Disables the source map generation.
- **true**: Generates a source map file, but without to append a sourceMappingURL comment to the end of the output.
- **inline**: Generates a source map file and appends a sourceMappingURL comment to the end of the output, containing the base64 encoded source map.

    ```
    //# sourceMappingURL=data:application/json;base64,<encoded_source_map_data>
    ```

New

- **linked**: Generates a source map file and appends a sourceMappingURL comment to the end of the output, containing an URL to the generated source map file.

    ```
    //# sourceMappingURL=<source_map_url>
    ```

Example

```bash
$ swc compile input.js --out-file=output.js --source-maps=linked
```

Output

```
output.js
output.js.map
```

Show the last line of the generated output.js file

```
$ tail -n=1 output.js
//# sourceMappingURL=output.js.map
```

Summary

- The source map path is now known to the compiler.
- Replaced the enum ~~`SourceMapsConfig::Str`~~ with the enum `SourceMapsConfig::Inline` and the enum `SourceMapsConfig::Linked`.
- Replaced the enum ~~`InputSourceMap::Str`~~ with the enum `InputSourceMap::Inline`.
- `SourceMapsConfig::default()` is now set to `SourceMapsConfig::Bool(false)` to match the compiler default (source map generation is disabled by default).
- Added the `FromStr` trait to `SourceMapsConfig`.
- Added the `serde::Deserializer` trait to `SourceMapsConfig`.
- Added the `serde::Deserializer` trait to `InputSourceMap`.
- Added tests
- Improved the error message when the `SourceMapsConfig::from_str` method is used

    Example

    ```bash
    $ swc compile input.js --out-file=output.js --source-maps=foo
    ```

    Error message

    ```
    invalid source maps value: got "foo" but expected either one of false, true, inline, linked
    ```

Related

- https://github.com/swc-project/swc/discussions/3859
- https://github.com/swc-project/swc/discussions/6479
- https://github.com/aspect-build/rules_swc/issues/146


jridgewell:
This breaks with both [Babel](https://babeljs.io/docs/en/options#sourcemaps)'s and [esbuild](https://esbuild.github.io/api/#sourcemap)'s definition of `"both"`, which is to add an a `//# sourceMappingURL=data:application/json;base64,…` and have an external sourcemap file. Babel doesn't have an option like what's proposed here (probably because it doesn't' bundle), but esbuild calls this `"linked"`

realtimetodie:
Thank you @jridgewell. This is a good idea, I like it.

I renamed the enum ~~`SourceMapsConfig::Both`~~ ("both") -> `SourceMapsConfig::Linked` ("linked") to not break with [Babel](https://babeljs.io/docs/en/options#sourcemaps)'s and [esbuild](https://esbuild.github.io/api/#sourcemap)'s definition of "both" as suggested.

Example

```bash
$ swc compile input.js --out-file=output.js --source-maps=linked
```

realtimetodie:
I updated the title and the initial comment.

kdy1:
> **true**: Generates a source map file, but without to append a sourceMappingURL comment to the end of the output

Is this true? I think this is a bug..

kdy1 (changes requested):
Please use `true` for this instead

kdy1:
Nevermind. I concluded `linked` is fine.

bindings/binding_core_node/Cargo.toml
```diff
@@ -56,7 +56,7 @@ tracing-chrome = "0.5.0"
 tracing-futures = "0.2.5"
 tracing-subscriber = {version = "0.3.9", features = ["env-filter"]}
 
-swc_core = {version = "0.55.5", features = [
+swc_core = {version = "0.56.1", path = "../../crates/swc_core", features = [
```
kdy1, line 59: Please revert this

kdy1: I think we should have only `verison = "0.56.2"` (latest)

realtimetodie:
@kdy1 I'm not sure why the bindings are in an isolated workspace. It complicates the development.

jridgewell (dismissed):

alexeagle:
@realtimetodie do you think that build failure for the node binding is caused by this change?
```
error[E0061]: this function takes 12 arguments but 13 arguments were supplied
   --> binding_core_node/src/bundle.rs:127:51
    |
127 | ...                   let output = self.swc.print(
    |                                             ^^^^^
    |
note: associated function defined here
   --> /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/swc-0.242.2/src/lib.rs:438:12
    |
438 |     pub fn print<T>(
    |            ^^^^^
help: did you mean
    |
127 |                             let output = self.swc.print(&m, None, None, true, codegen_target, SourceMapsConfig::Bool(true), &Default::default(), None, minify, None, false, true)?;
```

realtimetodie:
@alexeagle Yes, e4d80362e12c8c718c65957efad4a448c16e44d8 broke the build. Cargo pulls the swc_core crate from the remote crate registry now, which does not contain the new changes. Before that, the build passed. This is what I meant in https://github.com/swc-project/swc/pull/6817#issuecomment-1383645651. I think @kdy1 will bump the version manually, once this is merged.

alexeagle:
Ping @kdy1 - we're waiting for this one so we can release the Bazel swc integration to 1.0.0

kdy1:
There's a merge conflict

realtimetodie:
I will resolve the merge conflict in a couple of hours, I've a night shift today in the warehouse

realtimetodie:
I resolved the merge conflict. The version of the swc_core crate is now set to 0.58.0.

crates/swc/src/config/mod.rs
```diff
@@ -240,37 +244,106 @@ impl Options {
...
-/// Configuration related to source map generated by swc.
-#[derive(Clone, Serialize, Deserialize, Debug)]
-#[serde(untagged)]
+/// Configuration related to source maps generation.
+#[derive(Debug, Clone, Serialize, Eq, PartialEq)]
 pub enum SourceMapsConfig {
     Bool(bool),
-    Str(String),
+    Inline,
+    Linked,
```
kdy1, line 252: I think this should be `Linked(PathBuf)` instead

kdy1: `Linked` + `None` for `source_map_path` is an invalid state

kdy1 (changes requested):

crates/swc/src/config/mod.rs
realtimetodie: Yes, I also thought about this. I want to emphasize that this will not solve the dependent-type problem.

For example, the instantiation creates an invalid Path to a nonexistent source maps file already

```json
{
    "sourceMaps": "linked"
}
```

```rust
SourceMapsConfig::Linked(PathBuf::new())
```

The value of the `SourceMapsConfig::Linked` enum can not be universal. Its inner `PathBuf` value (which is the path to the source maps file) is dynamic and changes with every input.

realtimetodie: I tried this out and the code is even more ugly now.

realtimetodie: To get this right, the `SourceMapsConfig::Bool(bool)` needs to be replaced with `SourceMapsConfig::Enabled(PathBuf)` and `SourceMapsConfig::Disabled` respectively. Please let me know if I this is alright.

kdy1: Why do we need pathbuf as an argument? `true` simply generates `.map` file with same file name, so I think we should do same thing for `linked`, too

realtimetodie:
I had to delete the fixture test for the linked option due to the way the test runner works. It is not possible to construct a source map path from a .swcrc configuration alone, as the initial value is `PathBuf::new()`.

crates/swc/src/config/mod.rs
realtimetodie: The reason is that I would prefer to move entire generation of the .map path into the swc core crate. If we would do that, then we could also return a map path from there.

realtimetodie: I implemented it as you suggested, but I had to delete the fixture test for the linked option.

realtimetodie: I implemented it as you suggested, but I had to delete the fixture test for the linked option. This is the down-side of your idea.

crates/swc/src/config/mod.rs
```diff
@@ -240,37 +241,106 @@ impl Options {
...
-/// Configuration related to source map generated by swc.
-#[derive(Clone, Serialize, Deserialize, Debug)]
-#[serde(untagged)]
+/// Configuration related to source maps generation.
+#[derive(Debug, Clone, Serialize, Eq, PartialEq)]
 pub enum SourceMapsConfig {
     Bool(bool),
-    Str(String),
+    Inline,
+    Linked(PathBuf),
```
kdy1, line 249: I think this should be `Linked`, not `Linked(PathBuf)`

kdy1 (changes requested):
I was confused while reviewing for the first time, my bad.

Please see https://github.com/swc-project/swc/pull/6817#discussion_r1083376046

crates/swc/src/config/mod.rs
realtimetodie: No problem. I will revert my previous commit.

crates/swc/tests/exec.rs
```diff
@@ -144,10 +144,20 @@ fn create_matrix(entry: &Path) -> Vec<Options> {
...
+            SourceMapsConfig::Linked(entry.with_extension(format!(
+                "{}.map",
+                entry.extension().unwrap().to_string_lossy(),
+            ))),
+        ]
+    })
     .into_iter()
     .map(
-        |((((target, syntax), minify), external_helpers), source_map)| {
+        |((((target, syntax), minify), external_helpers), source_maps)| {
```
kdy1, line 160: Please revert this. The execution test suite is very slow and adding a matrix will prolong it

bindings/swc_cli/src/commands/compile.rs
```diff
@@ -380,77 +346,133 @@ impl CompileOptions {
...
+                                    fs::create_dir_all(out_dir)?;
+                                }
+
+                                fs::write(&out_file, &transform_output.code)?;
+
+                                if let Some(map) = &transform_output.map {
+                                    fs::write(map_file.unwrap(), map)?;
+                                }
+                            } else {
+                                println!(
```
kdy1, line 461: Seems like a mistake?

realtimetodie: Previously, the name of the source map was defined in the aforementioned function. But this does not work anymore. It is now required to know the name of the source map beforehand. Therefore, I moved this function and made it part of the executor.

realtimetodie:
I dropped 233b6746303b24b4add6aeb81a14340c2f02149e and reverted the changes.

bindings/swc_cli/src/commands/compile.rs
realtimetodie: The function still exists, it is just inside of the executor.

realtimetodie:
@kdy I want to ask you politely if anything is missing, please let me know. Thank you.

crates/swc/src/config/mod.rs
```diff
@@ -1136,6 +1274,8 @@ pub struct BuiltInput<P: swc_ecma_visit::Fold> {
     /// `pass`.
     pub minify: bool,
     pub external_helpers: bool,
+    /// Path to the source map file.
+    pub source_map_path: Option<PathBuf>,
```
kdy1, line 1278: I'm not merging this PR because of this

realtimetodie: I know that this is a large PR and I understand that you are doing this voluntarily. I apologize for the inconvenience.

kdy1: No, I'm talking about adding this field. I meant removing this field by `implicit` path.

https://github.com/swc-project/swc/pull/6817#discussion_r1083376046

realtimetodie: I can't follow. It would be nice if you could provide me a more detailed example. Thank you.

kdy1, line 1278:
```suggestion

```

crates/swc/src/config/mod.rs
```diff
@@ -1136,6 +1274,8 @@ pub struct BuiltInput<P: swc_ecma_visit::Fold> {
     /// `pass`.
     pub minify: bool,
     pub external_helpers: bool,
+    /// Path to the source map file.
```
kdy1, line 1277:
```suggestion

```

realtimetodie: I did exactly as you said. Thank you for your explanation.

realtimetodie:
I invested a lot of hours to help https://github.com/swc-project/swc/discussions/3859. I do this voluntarily and no one is paying me for doing this. The build is now broken.

realtimetodie:
@paullewis @alexeagle Please help me I think @kdy1 is playing games with me here.

realtimetodie:
I was nice and accommodating. I don't know @kdy1. But the way he reviewed my work and communicated with me is rude and disrespectful.

kdy1:
@realtimetodie  I told we should preserve the previous behavior (implicitly determining the path of the `.map` file) you several times. It's you ignoring my comment. I attached the link to my previous comment each time I request a change.

Of course, it's possible that someone fails to understand my change request. I reviewed an enormous amount of PRs until now and I have lots of experience like it. In my experience, they understood if I showed code to them. So I show you the code change for my request using the GitHub feature. Game? Rude? For which point?

realtimetodie:
I think what you mean with "implicitly determined" is to generate the source maps path within the compiler from.the filename, instead of passing it down as an argument? 

The reason why we can't "implicitly determine" the path to the source map file is because of the cli --source-map-target option. 

Currently, the user can define a custom output path for the generated source maps file when using the SourceMapsConfig::Bool(true) enum. I simply adopted this feature for the new SourceMapsConfig::Linked enum too. This is the only reason why the source_maps_path is passed down to the compiler.

If you say that the cli --source-map-target option should be ignored let me know. I thought you saw this part in the code.

kdy1:
> I think what you mean with "implicitly determined" is to generate the source maps path within the compiler from.the filename, instead of passing it down as an argument?

Yep. I mean filename + `.map`, which is the previous logic.

But seems like it's my bad, sorry. I didn't know that `--source-map-target` exist. I'm not sure how did previous code work

kdy1:
Ah, I got it. It simply returned the `map` object. But with `linked` it's impossible because `linked` requires writing the output path to the source file, right?

I'm sorry, I didn't think about this carefully. (I stopped thinking to work on stc while thinking about the source map path option)

realtimetodie:
It is also my fault, I did not point this out in the initial comment when i listed the changes and features.

kdy1 (dismissed):
Thank you, and sorry.

---

swc-bump:
 - swc --breaking

realtimetodie:
I deleted my previous comments. I was too emotional.

kdy1:
Oh CI failed

realtimetodie:
I need to revert https://github.com/swc-project/swc/pull/6817/commits/4b303a27853c7434716dc2a5c1d804c30bbcf1f1.

kdy1 (approved):

kdy1 (approved):

kdy1:
cc @kwonoj I think this can also be implemented by adding logic like https://github.com/swc-project/cli/blob/933589ae907637afb7f0b0a43b7a24a7a920595e/src/swc/compile.ts#L30-L33 in `swc_cli`.
Which way do you prefer?

kdy1:
I thought more about this, and I think sourceMap=true should generate `sourceMappingUrl`, and `linked` seems a bit redundant. But this can't be done by the `swc` crate.

realtimetodie:
I apologize for not writing my thoughts about the implementationin the initial post.

The main reason why I put the generation of the `sourceMappingURL` string into the swc crate is because I favor centralization. The code to generate the `sourceMappingURL` would be in one place, in the `print` function and would not be scattered around.

Secondly, I thought that third party consumers of the swc crate would greatly benefit from such a new feature SourceMapsConfig::Linked. Otherwise the feature would be hidden in the cli bindings crate, which is not meant to be consumed.

I understand the concerns about the new `source_map_path` variable. It seems unfortunate at first, but the path to the generated output is passed down using the `output_path` variable already.

https://github.com/swc-project/swc/blob/33dab568531fa3fc77697b75bce2df7514938d0b/crates/swc/src/lib.rs#L438-L442

I also want to emphasize on @jridgewell https://github.com/swc-project/swc/pull/6817#issuecomment-1383230637. The introduction of a new "linked" option helps users, because the configuration and naming convention will be aligned with other tools.

kdy1:
I thought more about this. I think an ideal solution is use `SourceMapsConfig::Bool(true)` for this behavior and patch  `@swc/cli` not to add `sourceMappingUrl`.

The problem is that, `@swc/cli` 

 - generates `sourceMappingUrl` for `SourceMapsConfig::Bool(true)`
 - generates two `sourceMappingUrl`s for `SourceMapsConfig::Linked`

and `swc_cli` 

 - generates no `sourceMappingUrl` for `SourceMapsConfig::Bool(true)`
 - generates one `sourceMappingUrl` for `SourceMapsConfig::Linked`

realtimetodie:
@kdy1 Yes, the @swc/cli needs to be patched.

However, the Node.js cli can handle multiple inputs at once and concats the outputs. If we combine the SourceMapsConfig::Bool(true) option as suggested, this will create a regression because every generated output will contain an URL to a non-existent source map file

https://github.com/swc-project/cli/blob/7217f7ad6a68f5bc788094affd8c5530231911c5/src/swc/file.ts#L33

For this reason, we will always need an option to generate an output and a source map without a sourceMappingURL string. 

Ideally, we could try to move this part from the Node.js cli into Rust. But this would require a lot of effort.
The new option SourceMapsConfig::Linked requires minimal changes to the Node.js cli and does not break existing behaviour.

realtimetodie:
@kdy1 The @swc/cli can be patched in such a way, that users will not require to update their configuration (from true -> "linked") and the option "linked" would be invalid in Node.js.

realtimetodie:
I linked to https://github.com/swc-project/swc/issues/1388 which is something I will be working on after this PR, as it is a regression of the Rust cli.

kdy1:
I'm against adding an option which is invalid for `@swc/cli`

PR #6822, opened 2023-01-16
fix(es/codegen): Don't print comments before directives

kdy1:
**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/5675.


kdy1 (commented):
swc-bump:
 - swc_ecma_codegen

jridgewell (approved):
Note: Comments don't effect directive parsing, so the prior behavior was valid.

kdy1:
I think this is not required

PR #6835, opened 2023-01-19
feat(bindings/cli): Implement --config

titanous:
**Description:**

Implements `--config`. Plugin paths that are explicitly relative (start with `./` or `../`) are converted to an absolute path using the current working directory as the base.


bindings/swc_cli/src/commands/compile.rs
```diff
@@ -271,10 +275,32 @@ impl CompileOptions {
         });
 
         let mut options = Options {
+            config: self.config.to_owned().unwrap_or_default(),
             config_file,
             ..Options::default()
         };
 
+        options.config.jsc.experimental.plugins =
```
kdy1, line 283: I'm not sure if this is correct.

kdy1: Personally I'm against this change unless it's unevitable

titanous: Can you elaborate on your concerns? The code is correct in that it works. Currently relative paths don’t work at all, so mapping them to absolute is the only way to get them working.

Given that there was no interest in making relative paths work in the plugin resolver (#6800), I’m not sure what other solutions there are. Bazel doesn’t use absolute paths, so we can’t find plugins without some form of relative path handling.

titanous:
Hi, any thoughts on this? It's blocking implementing plugins in `rules_swc`.

bindings/swc_cli/src/commands/compile.rs
kwonoj: My personal take: (not kdy's opinion, so this should't approve full PR)

1. I _think_ it's sort of acceptable to try path resolve in here, but it should be clearly commented this is not a finalized behavior we'll support and should be subject to plugin resolver itself's behavior change in the future.
2. Do we need to check if path stars with either relative or absolute? `absolutize` against non-relative path should noop and return as-is.

titanous: That check is there so that npm-style imports (like `@swc/plugin-foo`) still work, if we don't have this check, those imports will end up absolutized relative to the CWD, which will not work.

kwonoj: and reason we can't use `path::is_relative` then? not sure if we verbosely need to check each path component.

titanous: `is_relative` will return true for all non-absolute paths, including strings that are NPM packages.

kwonoj: hmm :thinking: I'm not super in favor of current implementation, curious if there's way to reduce it but sounds like no easy way to do.

titanous: I'm not aware of any way of simplifying this. The only other viable approach is supporting relative paths in the plugin resolver (#6800).

kdy1: Did you test this function on Windows?
I ordered a Windows machine, but I can't test this PR on windows at the moment.

kwonoj: Maybe path.is_relative && path.exists?

titanous: I didn't, I don't have a Windows setup currently, though I'm not relying on any platform-specific behavior. `CurDir` and `ParentDir` are the same everywhere.

titanous: > Maybe path.is_relative && path.exists?

I'm fine with that, let me know if you want me to change it. The only downside is that if a directory is named for an npm package it would take precendence, but that seems fine?

kwonoj: > The only downside is that if a directory is named for an npm package it would take precendence, but that seems fine

Given conversion only triggers if input is relative, that'll happen if things like `./${somenpmpkg}` and there is npm pkg also, isn't it? I'd guess that's sort of expected behavior since input is path, not a pkg to resolve.

titanous: Right, you'd never write an npm package reference like that.

kdy1: Can you add some docs to the `config` field to describe that this is used only for experimental plugins and also experimental?

titanous: Docs added.

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc

Issue #6836, opened 2023-01-19
"\x000" falsely gets turned into "\00"


dferber90:
### Describe the bug

The string `"\x000"` gets turned into `\00`, but should stay as `"\x000"` afaik. 

Producing `\00` leads to errors as it is interpreted as an octal escape character. For example this throws an error
```
// node test.mjs
console.log("\00")
```

throws

```
console.log("\00")
              ^

SyntaxError: Octal escape sequences are not allowed in strict mode.
    at ESMLoader.moduleStrategy (node:internal/modules/esm/translators:115:18)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:289:14)
    at async link (node:internal/modules/esm/module_job:70:21)
```

### Input code

```typescript
"\x000"
```

### Playground link

https://play.swc.rs/?version=1.3.27&code=H4sIAAAAAAAAA1OKqTAwMFACAG%2F%2BxqkHAAAA&config=H4sIAAAAAAAAA0WMTQrEIAxG75K1286id5hDBCctFv9IUhgR714tlu7C915ehUMsrBUyshCPS0pU%2FMMKZAOKZZcVTNf6tKEXagYUeScdiiyd%2BZSEJjUQXHRbGSWbQmYSeRHG3T9m66GQfucYKmjJdAc%2F0N7G%2FHPynaLySe0Cj9ke9LUAAAA%3D

### Expected behavior

It should keep the string as-is afaik.

### Actual behavior

It turns the string into `"\00"`

### Version

1.3.27

### Additional context

I ran into this through Next.js: https://github.com/vercel/next.js/issues/45002

PR #6838, opened 2023-01-20
fix(es/codegen): Fix codegen of string literals with `\x000`

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6836.


kdy1 (commented):
swc-bump:
 - swc_ecma_codegen

crates/swc_ecma_codegen/src/lib.rs
```diff
@@ -3481,7 +3481,7 @@ fn get_quoted_utf16(v: &str, ascii_only: bool, target: EsVersion) -> String {
 
                 match next {
                     Some('1'..='9') => buf.push_str("\\x00"),
-                    _ => buf.push_str("\\0"),
+                    _ => buf.push_str("\\x00"),
```
jridgewell, line 3484: I think the only problem was the range here:
```suggestion
                    Some('0'..='7') => buf.push_str("\\x00"),
                    _ => buf.push_str("\\0"),
```

kdy1: Thank you!

jridgewell (approved):

crates/swc_ecma_codegen/src/lib.rs
```diff
@@ -3480,7 +3480,7 @@ fn get_quoted_utf16(v: &str, ascii_only: bool, target: EsVersion) -> String {
                 let next = iter.peek();
 
                 match next {
-                    Some('1'..='9') => buf.push_str("\\x00"),
+                    Some('0'..='7') => buf.push_str("\\x00"),
```
jridgewell, line 3483: Sorry, no, it still fails!
```js
() => { "use strict"; 1; "\09"; }
```

jridgewell (changes requested):

crates/swc_ecma_codegen/src/lib.rs
jridgewell: Yup: https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#prod-EscapeSequence

Sorry, it's gotta be `0..=9` to properly pass. `"\09"` is a [LegacyOctalEscapeSequence](https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#prod-LegacyOctalEscapeSequence), which is only valid in loose mode.

jridgewell (dismissed):

PR #6840, opened 2023-01-20
perf(es/minifier): check whether a function is pure with O(1) cost

hyf0:
**Description:**

We might need to pass a giant pure function list to the SWC minifier as esbuild did [here](https://github.com/evanw/esbuild/blob/5fe21253ee75fb4c5ea395a0877b2a5ab51c3575/internal/config/globals.go).

The current implementation could slow the minifier.

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_minifier/src/metadata/mod.rs
+++ b/crates/swc_ecma_minifier/src/metadata/mod.rs
@@ -1,3 +1,4 @@
+use rustc_hash::FxHashSet;
 use swc_common::{
     comments::{Comment, CommentKind, Comments},
     EqIgnoreSpan, Span, SyntaxContext,
@@ -11,6 +12,15 @@ use swc_ecma_visit::{
 
 use crate::option::CompressOptions;
 
+#[derive(Debug, Eq, Hash)]
+struct EqIgnoreSpanExprRef<'a>(&'a Expr);
+
+impl<'a> PartialEq for EqIgnoreSpanExprRef<'a> {
+    fn eq(&self, other: &Self) -> bool {
+        Ident::within_ignored_ctxt(|| self.0.eq_ignore_span(&other.0))
+    }
+}
+
 #[cfg(test)]
 mod tests;
 
@@ -21,10 +31,18 @@ pub(crate) fn info_marker<'a>(
     marks: Marks,
     // unresolved_mark: Mark,
 ) -> impl 'a + VisitMut {
+    let pure_fns = options.map(|options| {
+        options
+            .pure_funcs
+            .iter()
+            .map(|f| EqIgnoreSpanExprRef(f))
+            .collect()
+    });
     InfoMarker {
         options,
         comments,
         marks,
+        pure_fns,
         // unresolved_mark,
         state: Default::default(),
     }
@@ -38,7 +56,7 @@ struct State {
 
 struct InfoMarker<'a> {
     options: Option<&'a CompressOptions>,
-
+    pure_fns: Option<FxHashSet<EqIgnoreSpanExprRef<'a>>>,
     comments: Option<&'a dyn Comments>,
     marks: Marks,
     // unresolved_mark: Mark,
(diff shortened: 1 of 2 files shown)


hyf0:
> Can we have those list also in swc minifier? (itself)

Absolute yes. I would do it in another PR. 

We might need to provide an option(enable by default) to disable it. Because users might do some hacky things to the environment.

kdy1:
We have `pristine_globals`

hyf0:
> We have `pristine_globals`

Get it.

kdy1:
I think we can use a hash set of `(JsWord, JsWord)`.
Is there more complex expression other than those globals?

hyf0:
> I think we can use a hash set of `(JsWord, JsWord)`. Is there more complex expression other than those globals?

Yes. Dealing `Expr::Member` and `Expr::Ident` should be enough base on the input [here](https://github.com/evanw/esbuild/blob/5fe21253ee75fb4c5ea395a0877b2a5ab51c3575/internal/config/globals.go).

crates/swc_ecma_minifier/src/metadata/mod.rs
```diff
@@ -11,6 +14,37 @@ use swc_ecma_visit::{
 
 use crate::option::CompressOptions;
 
+#[derive(Debug, Eq)]
+struct HashEqIgnoreSpanExprRef<'a>(&'a Expr);
```
kdy1, line 18: I'm not sure if this will work.

kdy1: Can you add some tests?

crates/swc_ecma_minifier/src/metadata/mod.rs
```diff
@@ -124,17 +167,18 @@ impl VisitMut for InfoMarker<'_> {
 
         if self.has_pure(n.span) {
             n.span = n.span.apply_mark(self.marks.pure);
-        } else if let Some(options) = self.options {
+        } else if let Some(pure_fns) = &self.pure_funcs {
             if let Callee::Expr(e) = &n.callee {
+                println!("--pure_fns: {pure_fns:#?}");
```
kdy1, line 172: Also, please remove this

kdy1 (changes requested):

crates/swc_ecma_minifier/src/metadata/mod.rs
hyf0: Sure. I guess we only need to test the `Eq` and `Hash` behaviors of `HashEqIgnoreSpanExprRef<'a>`.

kdy1: I think inserting and getting various values from `HashSet<HashEqIgnoreSpanExprRef>` is enough

crates/swc_ecma_minifier/src/metadata/tests.rs
```diff
@@ -158,3 +158,63 @@
...
+    use swc_ecma_ast::*;
+
+    fn expr_ref(expr_ref: &Expr) -> HashEqIgnoreSpanExprRef {
+        HashEqIgnoreSpanExprRef(expr_ref)
+    }
+
+    testing::run_test(false, |_cm, _handler| {
+        let dummy_sp = DUMMY_SP;
+        let meaningful_sp = dummy_sp.apply_mark(Mark::new());
+
```
kdy1, line 179: Can we have some check for expressions which is not an ident nor a member expression?

hyf0: Done

crates/swc_ecma_minifier/src/metadata/mod.rs
```diff
@@ -11,6 +14,37 @@ use swc_ecma_visit::{
...
+        // In pratice, most of cases/input we are dealing with are Expr::Member or
+        // Expr::Ident.
+        match self.0 {
+            Expr::Ident(i) => {
+                i.sym.hash(state);
+            }
+            Expr::Member(i) => {
+                Self(&i.obj).hash(state);
+                if let MemberProp::Ident(prop) = &i.prop {
+                    prop.sym.hash(state);
```
jridgewell, line 37: I think there's a bug here that `console.log` will hash the same as the identifier `consolelog`. We'd need to hash `"."` or similar sigil.

kdy1: Does it matter? `Eq` will be used if the hash value matches

jridgewell: Ohhhhhhh.

jridgewell (approved):

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_minifier

kdy1 (approved):

PR #6842, opened 2023-01-21
feat(es/minifier): mark some builtin functions pure.

hyf0:
**Description:**

              Can we have those list also in swc minifier? (itself)

_Originally posted by @kdy1 in https://github.com/swc-project/swc/issues/6840#issuecomment-1398415208_
            

**BREAKING CHANGE:**

The PR relies on code of #6840.


crates/swc_ecma_minifier/src/metadata/pure_exprs.rs
```diff
@@ -0,0 +1,822 @@
...
+        member_expr!(DUMMY_SP, console.countReset),
+        member_expr!(DUMMY_SP, console.debug),
+        member_expr!(DUMMY_SP, console.dir),
+        member_expr!(DUMMY_SP, console.dirxml),
+        member_expr!(DUMMY_SP, console.error),
+        member_expr!(DUMMY_SP, console.group),
+        member_expr!(DUMMY_SP, console.groupCollapsed),
+        member_expr!(DUMMY_SP, console.groupEnd),
+        member_expr!(DUMMY_SP, console.info),
+        member_expr!(DUMMY_SP, console.log),
```
hyf0, line 222: I'm not sure about `console`. I often use `console.log` to debug or capture some variables to make the minifier/mangler not optimize it.

kdy1: Same here 🤣

kdy1: I used it while debugging the minifier

crates/swc_ecma_minifier/src/metadata/mod.rs
```diff
@@ -167,10 +180,12 @@ impl VisitMut for InfoMarker<'_> {
...
         if self.has_pure(n.span) {
             n.span = n.span.apply_mark(self.marks.pure);
-        } else if let Some(pure_fns) = &self.pure_funcs {
+        }
+
+        if !self.pure_funcs.is_empty() {
             if let Callee::Expr(e) = &n.callee {
                 // Check for pure_funcs
-                if pure_fns.contains(&HashEqIgnoreSpanExprRef(e)) {
+                if self.pure_funcs.contains(&HashEqIgnoreSpanExprRef(e)) {
```
kdy1, line 188: Can we use two hashset instead?
I think it will avoid allocations

kdy1: One hashset with type `Lazy<FxHashSet<HashEqIgnoreSpanExprRef<'static>>>`

hyf0: Get it.

hyf0:
I think is misunderstand the usage of the esbuild's [list](https://github.com/evanw/esbuild/blob/5fe21253ee75fb4c5ea395a0877b2a5ab51c3575/internal/config/globals.go).

> Note that membership in this list says nothing about whether calling any of
> these functions has any side effects. It only says something about
> referencing these function without calling them (doesn't have side effects).

hyf0:
But still, there are indeed some expressions in the list that are calling without side effects. I need to tweak the list.

PR #6844, opened 2023-01-23
refact(es/preset-env): use const fn instead of macro

Austaras:
**Description:**

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_preset_env

kdy1 (changes requested):
CI failed

kdy1 (approved):
Thanks!

PR #6867, opened 2023-01-28
feat(es/typescript): Support `export type * from "mod"`

magic-akari:
**Description:**

Adds support for `export type * from "mod"` and `export type * as ns from "mod"`.

**BREAKING CHANGE:**

- https://github.com/microsoft/TypeScript/pull/52217


kdy1:
I'll review/merge this once 5.0 is published, as this will break Wasm plugins

PR #6884, opened 2023-01-31
build(cargo): Drop dependencies on `retain_mut`

kdy1:


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #6893, opened 2023-02-01
perf: Using better `string cache`

IWANABETHATGUY:
**Description:**
Using better string-cache
The latest `Mutex` implementation is faster than `parking_lot`, especially when locking contention is heavy, 
https://github.com/rust-lang/rust/pull/95035#issuecomment-1073966631, 
This could have a huge performance improvement when using swc in some parallel scenarios, like bundler.

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_atoms/Cargo.toml
+++ b/crates/swc_atoms/Cargo.toml
@@ -28,7 +28,7 @@ rkyv      = { package = "rkyv", version = "=0.7.37", optional = true }
 rkyv-latest  = { package = "rkyv-test", version = "=0.7.38-test.2", optional = true }
 rustc-hash   = "1.1.0"
 serde        = "1"
-string_cache = "0.8.4"
+string_cache = { git = "https://github.com/swc-project/string-cache", branch = "swc" }
 triomphe     = "0.1.8"
 
 [build-dependencies]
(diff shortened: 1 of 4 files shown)


crates/swc_atoms/Cargo.toml
```diff
@@ -28,7 +28,7 @@ rkyv      = { package = "rkyv", version = "=0.7.37", optional = true }
 rkyv-latest  = { package = "rkyv-test", version = "=0.7.38-test.2", optional = true }
 rustc-hash   = "1.1.0"
 serde        = "1"
-string_cache = "0.8.4"
+string_cache = { git = "https://github.com/swc-project/string-cache", branch = "swc" }
```
kdy1, line 31: This prevents publishing crates

kdy1 (changes requested):

crates/swc_atoms/Cargo.toml
hyf0: Could we make reuse the [repo](https://github.com/swc-project/string-cache) and publish a string-cache of SWC version? Something like `string-cache-swc`.

For some reason, I guess they(string-cache) just don't want to change anything. https://github.com/servo/string-cache/pull/265

kdy1: Yes, but I think you can use `patch` section of cargo to use it, and changing `string_cache` with something different is a breaking change.

Issue #6896, opened 2023-02-02
Optimisation breaks boolean statement


fweinhold:
### Describe the bug

We have next.js project, which recently switched to SWC as the preferred optimiser lib. The components we use in that project originate from an internal component lib which still uses terser for the same purpose. 

Now we discovered some very strange behaviour with one of our components, but only when used in the next.js project. 
 
After digging into it a bit, we discovered that the problem seems to be SWC wrongly resolving a boolean construct, when it was beforehand optimised by terser.

In addition to the example I put in below, I made the example executable in a Fiddle: https://jsfiddle.net/nk1x74th/1/ I hope this helps...

### Input code

```typescript
// this the original code, just for comparison
function Limit(min, max) {
  var length = Math.abs(min - max);

  function reachedMin(n) {
    return n < min;
  }

  function reachedMax(n) {
    return n > max;
  }

  function reachedAny(n) {
    return reachedMin(n) || reachedMax(n);
  }

  function constrain(n) {
    if (!reachedAny(n)) return n;
    return reachedMin(n) ? min : max;
  }

  function removeOffset(n) {
    if (!length) return n;
    return n - length * Math.ceil((n - max) / length);
  }

  var self = {
    length: length,
    max: max,
    min: min,
    constrain: constrain,
    reachedAny: reachedAny,
    reachedMax: reachedMax,
    reachedMin: reachedMin,
    removeOffset: removeOffset
  };
  return self;
}
console.log("The result should be 0. And it is:", Limit(0,3).constrain(-1))

// this is the same code after being optimised by terser (with default settings)
console.log("The result should be 0. And it is:",function(n,e){var r=Math.abs(n-e);function t(e){return e<n}function u(n){return n>e}function c(n){return t(n)||u(n)}return{length:r,max:e,min:n,constrain:function(r){return c(r)?t(r)?n:e:r},reachedAny:c,reachedMax:u,reachedMin:t,removeOffset:function(n){return r?n-r*Math.ceil((n-e)/r):n}}}(0,3).constrain(-1));
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": true
    },
    "target": "es5",
    "loose": true,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
[...]

PR #6899, opened 2023-02-05
fix(es/minifier): Respect span hygiene from expr simplifier

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6896.

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/optimize/mod.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/mod.rs
@@ -2642,9 +2642,6 @@ where
         debug_assert_eq!(self.prepend_stmts.len(), prepend_len);
         debug_assert_eq!(self.append_stmts.len(), append_len);
         debug_assert_valid(s);
-
-        #[cfg(feature = "debug")]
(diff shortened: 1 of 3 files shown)


kdy1:
# Investigation

This looks like an issue of span hygiene.

After the first pass, the input becomes

```js
console.log("The result should be 0. And it is:", (function(n, e) {
    var r = Math.abs(n - e);
    function t(e) {
        return e < n;
    }
    function u(n) {
        return n > e;
    }
    function c(n1) {
        ;
        ;
        return n1 < n || n1 > e;
    }
    return {
        length: r,
        max: e,
        min: n,
        constrain: function(r) {
            ;
            return c(r) ? r < n ? n : e : r;
        },
        reachedAny: c,
        reachedMax: u,
        reachedMin: t,
        removeOffset: function(n) {
            return r ? n - r * Math.ceil((n - e) / r) : n;
        }
    };
})(0, 3).constrain(-1));
```

but if I pass this to the minifier it passes.

kdy1 (commented):
swc-bump:

kdy1:
Skipping the review as this is a very trivial change

Issue #6897, opened 2023-02-03
css_module use &:global() will output incorrect css


jerrykingxyz:
### Describe the bug

Using `&:global` in `css/module` will output incorrect css, can add these cases to crates/swc_css_modules tests to reproduce them

### Input code

```typescript
.a:global(.b) {}
.a {
  &:global(.b) {}
}
```

### Expected behavior

.__local__a.b {}
.__local__a.b {}

### Actual behavior

.b {}
.b {}

### Version

0.21.2

PR #6900, opened 2023-02-05
fix css_modules with .class:global(.class2)

ahabhgk:
**Description:**

**BREAKING CHANGE:**

fix part of #6897


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_css_modules

Issue #6901, opened 2023-02-05
`@swc/core-darwin-universal` probably does not exist.


AsPulse:
### Describe the bug

https://github.com/swc-project/swc/blob/c223113338e47979d37129abcaae560c0d8d7580/node-swc/src/binding.js#L105-L113
This part requires `@swc/core-darwin-universal`, but there is no package with this name on npm.
(https://www.npmjs.com/package/@swc/core-darwin-universal)

### Expected behavior

I think `@swc/core-darwin-universal` should be removed from the code or binaries should be published.

### Actual behavior

This appears as a problem when esbulid and others try to process the contents of require on the source.

For example, esm.sh:
- https://github.com/ije/esm.sh/issues/514

### Version

1.3.32

PR #6902, opened 2023-02-05
fix(swc/core): Stop doing require unnecessary packages (`@swc/core-darwin-universal`).

AsPulse:
**Description:**
### Removed the part that requires a package that does not exist.
(Previously, this was enclosed in a try-catch and did not manifest itself as a clear error, but was problematic in the situation described in the issue below.)

<br>

**BREAKING CHANGE:**

If `process.platform === 'darwin' && (process.arch !== 'x64' && process.arch !== 'arm64')`,
then Error "Unsupported architecture on macOS: (process.arch)" will be thrown.

Previously, **however**, the Error "Failed to load native binding" would have been thrown because `@swc/core-darwin-arm64` does not exist. In this point of view, Probably I can argue that this changes make error easy-to-understand.

<br>

- ref #6901

### DIFF
--- a/node-swc/src/binding.js
+++ b/node-swc/src/binding.js
@@ -102,15 +102,6 @@ switch (platform) {
     }
     break
   case 'darwin':
-    localFileExisted = existsSync(join(__dirname, 'swc.darwin-universal.node'))
-    try {
-      if (localFileExisted) {
-        nativeBinding = require('./swc.darwin-universal.node')
-      } else {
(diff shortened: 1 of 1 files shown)


kwonoj (approved):
🤔 Honestly I don't remember if I added this or not,

kdy1 (approved):
Same here 🤣 I wasn't aware of this code

---

swc-bump:
 - dbg-swc

PR #6907, opened 2023-02-06
refactor(plugin): Fix typo

eltociear:
**Description:**
arbitary -> arbitrary


kdy1:

PR #6909, opened 2023-02-06
feat(cli): Support for *.mts/*.cts files

### LINKED ISSUE #4679: Add support for *.mts/*.cts files
### Describe the feature

The current default supportet file extensions by the swc-cli are:
```["js", "jsx", "es6", "es", "mjs", "ts", "tsx"]```
I would love having support for `.mts` and `.cts` files which work like `.ts` or `.tsx` files
`.mts` => `.mjs`
`.cts` => `.cjs`
just like with `tsx` and `jsx`
`.tsx` => `.jsx`

it would work just like `ts` files but with an other extension (or optional even change the module settings)

_No response_

### Additional context

https://www.typescriptlang.org/docs/handbook/esm-node.html#new-file-extensions

### PR DESCRIPTION
**Pull request by nissy-dev**
**Description:** Fix #4679

**BREAKING CHANGE:**

**Related issue (if exists):** Fix #4679 I also referred to https://github.com/babel/babel/pull/13838

### CODE REVIEW

File: crates/swc_ecma_parser/src/lib.rs [hunk 1]
```diff
@@ -281,6 +288,14 @@ pub struct TsConfig {
...
     #[serde(skip, default)]
     pub no_early_errors: bool,
+
+    /// babel: `disallowAmbiguousJSXLike`
+    /// Even when JSX parsing is not enabled, this option disallows using syntax
+    /// that would be ambiguous with JSX (`<X> y` type assertions and
+    /// `<X>()=>{}` type arguments)
+    /// see: https://babeljs.io/docs/en/babel-plugin-transform-typescript#disallowambiguousjsxlike
+    #[serde(skip, default)]
+    pub disallow_ambiguous_jsx_like: bool,
```
kdy1, line 298: Can you extract this to a separate PR?

nissy-dev: I made #6913 🚀

webstrand:
Should this be available in v1.3.35?
```js
await (await import("@swc/core")).transform("let x: number = 5", { filename: "foo.mts" })
```
does not work in v1.3.35, it throws an error on encountering the `: number` annotation

kdy1:
@webstrand Can you file an issue?

PR #6912, opened 2023-02-07
chore(cli): Enable `preferUnplugged` by default for the plugins

orionmiz:
**Description:**

SWC plugin interface is not allowed when binary in .zip archive. (pnp environment)
Therefore, in order to easily support the yarn pnp environment, it is set to be installed as unplug by default.

- https://github.com/blitz-js/next-superjson-plugin/pull/61


kdy1 (approved):
swc-bump:
 - dbg-swc

PR #6916, opened 2023-02-08
fix(es/parser): Don't use `stacker` for armv7

kdy1:


kdy1 (commented):
swc-bump:
 - swc_ecma_parser

Issue #6807, opened 2023-01-13
Plugin crashes with rkyv LayoutError.


Drevoed:
### Describe the bug

SWC crashes when trying to invoke a plugin.

versions:
* `@swc/core` - 1.3.26
* `swc_core` - 0.56.1

backtrace:
```
❯ SWC_DEBUG=1 RUST_BACKTRACE=full swc --sync demo/src/index.ts
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: LayoutError', /Users/k.mironov/.cargo/registry/src/github.com-1ecc6299db9ec823/rkyv-0.7.37/src/impls/core/mod.rs:265:67
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Panic: PanicInfo { payload: Any { .. }, message: Some(failed to invoke plugin: failed to invoke plugin on 'Some("demo/src/index.ts")'

Caused by:
    0: failed to invoke `@effector/swc-plugin` as js transform plugin at node_modules/@effector/swc-plugin/target/wasm32-wasi/debug/effector_swc_plugin.wasm
    1: RuntimeError: unreachable
           at __rust_start_panic (<module>[11103]:0x438b25)
           at rust_panic (<module>[11093]:0x438815)
           at std::panicking::rust_panic_with_hook::had8ec4151dd4c568 (<module>[11092]:0x4387e4)
           at std::panicking::begin_panic_handler::{{closure}}::h2ac729f2457e9805 (<module>[11080]:0x437aa2)
           at std::sys_common::backtrace::__rust_end_short_backtrace::hfe2c5b74343c0cd6 (<module>[11079]:0x4379cc)
           at rust_begin_unwind (<module>[11087]:0x438060)
           at core::panicking::panic_fmt::hd7053c514eb59581 (<module>[11226]:0x4461cd)
           at core::result::unwrap_failed::h7e5c994e75ab37e8 (<module>[11275]:0x44ca56)
           at core::result::Result<T,E>::unwrap::h2356290abca8ee50 (<module>[10820]:0x4287c8)
           at rkyv::impls::core::<impl rkyv::DeserializeUnsized<[U],D> for [T]>::deserialize_unsized::h315b3a8965376629 (<module>[5344]:0x2b5bfa)
[...]

PR #6920, opened 2023-02-08
refactor: Fix lints using clippy from `nightly-2023-03-13`

kdy1:
**Description:**

We may need to bump `rkyv`.
(https://github.com/swc-project/swc/issues/6807#issuecomment-1401556054)


kdy1 (commented):
swc-bump:
 - dbg-swc

kdy1:
cc @kwonoj I think we find the condition for the UB

kwonoj:
HUH, so old version of nightly was the reason?

kdy1:
Not sure, but seems like one of two nightly versions are problem.
I think the newer version breaks `rkyv`, mainly because that's _correct_ for UBs. i.e. A correct compiler optimization is added, but code with UB is broken by the optimization.

kwonoj:
> seems like one of two nightly versions are problem.
I think the newer version breaks rkyv, mainly because that's correct for UBs. i.e. A correct compiler optimization is added, but code with UB is broken by the optimization.

sounds quite scary, few random thinking

- if a plugin uses certain version of compiler will it cause trouble even if swc/core have correct working compiler?
- is there a way to check regressions like this?
- in a long run, should we try to use stable compiler instead?

kdy1:
> if a plugin uses certain version of compiler will it cause trouble even if swc/core have correct working compiler?

Yes, single use is enough to break the plugin.

> is there a way to check regressions like this?

Yes, that's why I added tests which invoke noop plugin with all AST we can parse, although it's not caught by it at this time.

https://github.com/swc-project/swc/blob/29fee382249c9b91155d0d3c39d0df82bbd531c1/crates/swc_plugin_runner/tests/ecma_rkyv.rs#L65-L66

https://github.com/swc-project/swc/blob/29fee382249c9b91155d0d3c39d0df82bbd531c1/crates/swc_plugin_runner/tests/css_rkyv.rs#L63

> in a long run, should we try to use stable compiler instead?

It's a bug of `rkyv`. For a compiler, UB is perfectly fine to break, in the name of optimization.
But I think we should consider changing serialization/deserialization method to something with less unsafe...

kwonoj:
> we should consider changing serialization/deserialization method to something with less unsafe.

Given wasmer itself uses rkyv as underlying serialization mechanism, I guess this'll require to take out wasmer entirely. Not impossible to do, but may need some considerations.

kwonoj:
Forgot to mention, it looks like this occurred with serialization / deserialization of `Program` afaik. Something special occurs in those struct causes this?

kdy1:
We can use other serialization/deserialization for `Program`, and let wasmer use `rkyv`.
I'm not sure which part is problematic.

```
           at rkyv::impls::core::<impl rkyv::DeserializeUnsized<[U],D> for [T]>::deserialize_unsized::h315b3a8965376629 (<module>[5344]:0x2b5bfa)
           at rkyv::impls::alloc::vec::<impl rkyv::Deserialize<alloc::vec::Vec<T>,D> for rkyv::vec::ArchivedVec<<T as rkyv::Archive>::Archived>>::deserialize::h04d1ec3ea3e970d8 (<module>[222]:0x13be0)
           at swc_ecma_ast::module_decl::_::<impl rkyv::Deserialize<swc_ecma_ast::module_decl::ImportDecl,__D> for <swc_ecma_ast::module_decl::ImportDecl as rkyv::Archive>::Archived>::deserialize::h185c61ae65c5c0b5 (<module>[5304]:0x2a51dd)
           at swc_ecma_ast::module_decl::_::<impl rkyv::Deserialize<swc_ecma_ast::module_decl::ModuleDecl,__D> for <swc_ecma_ast::module_decl::ModuleDecl as rkyv::Archive>::Archived>::deserialize::h952789b151c9a970 (<module>[5305]:0x2a597c)
           at swc_ecma_ast::module::_::<impl rkyv::Deserialize<swc_ecma_ast::module::ModuleItem,__D> for <swc_ecma_ast::module::ModuleItem as rkyv::Archive>::Archived>::deserialize::h88d8e5b435c0c70e (<module>[5724]:0x2cf7ff)
           at rkyv::impls::core::<impl rkyv::DeserializeUnsized<[U],D> for [T]>::deserialize_unsized::h006528daf2248e8c (<module>[5738]:0x2d413c)
           at rkyv::impls::alloc::vec::<impl rkyv::Deserialize<alloc::vec::Vec<T>,D> for rkyv::vec::ArchivedVec<<T as rkyv::Archive>::Archived>>::deserialize::h380c70590928e990 (<module>[227]:0x147f2)
           at swc_ecma_ast::module::_::<impl rkyv::Deserialize<swc_ecma_ast::module::Module,__D> for <swc_ecma_ast::module::Module as rkyv::Archive>::Archived>::deserialize::h86f80a54809dc175 (<module>[5721]:0x2ce741)
           at swc_ecma_ast::module::_::<impl rkyv::Deserialize<swc_ecma_ast::module::Program,__D> for <swc_ecma_ast::module::Program as rkyv::Archive>::Archived>::deserialize::he0c1e07561c3b81a (<module>[5723]:0x2cefcf)
           at swc_common::plugin::serialized::PluginSerializedBytes::deserialize::hd942b90ece579bfb (<module>[4013]:0x1ad338)
           at swc_common::plugin::serialized::deserialize_from_ptr::h3a43eb3fbcea6d36 (<module>[4012]:0x1ac686)
           at __transform_plugin_process_impl (<module>[6596]:0x31c0a3)
           at __transform_plugin_process_impl.command_export (<module>[11393]:0x45418b)

```

This is the clue, though.
So some array in ``ImportDecl` is problematic, and it should be [`specifiers`](https://rustdoc.swc.rs/swc_core/ecma/ast/struct.ImportDecl.html#structfield.specifiers) but not sure why it fails

kwonoj:
I'd like to consider replacing serialization as a last resort, attempt to try fix to see if it works or not.

kdy1:
Same here 👍

kdy1:
I think `rkyv` is serializing data as `ImportDecl`, so I left a comment.

https://github.com/rkyv/rkyv/issues/338#issuecomment-1423737421

Issue #6888, opened 2023-01-31
Core-js polyfill for `.replaceAll()` stopped working in 1.3.21


gu-stav:
### Describe the bug

The core-js config to polyfill `.replaceAll()` stopped working in 1.3.21.

1.3.20: https://play.swc.rs/?version=1.3.20&code=H4sIAAAAAAAAA1MvSS0uUdcrSi3ISUxOdczJ0VBX11FQV9e05lI31DPSM1bXKy7IySzRUNdT19RLLNHQNQRKRRsaGceCeCAOAG324sZDAAAA&config=H4sIAAAAAAAAA0WOSwrDMAxE76J1CMGFFnKEQg8hXCW4xB8kuzQE371ycclOmnkz0gEvsTAfkJCFuE2yh4wfmIGsR7HsUoZBMZUW3ITqABl5pdwQMZOZ1N5iFOrAAN4Ft%2BytzEafmEROC8O6%2FcmqXT4%2BSxMOyHuiX%2BcV6tnRc04eHcxcdKfwbhlNt0wRXEnfsJHprsfgMprbaKDWL8W4QhrhAAAA

1.3.21: https://play.swc.rs/?version=1.3.21&code=H4sIAAAAAAAAA1MvSS0uUdcrSi3ISUxOdczJ0VBX11FQV9e05lI31DPSM1bXKy7IySzRUNdT19RLLNHQNQRKRRsaGceCeCAOAG324sZDAAAA&config=H4sIAAAAAAAAA0WOSwrDMAxE76J1CMGFFnKEQg8hXCW4xB8kuzQE371ycclOmnkz0gEvsTAfkJCFuE2yh4wfmIGsR7HsUoZBMZUW3ITqABl5pdwQMZOZ1N5iFOrAAN4Ft%2BytzEafmEROC8O6%2FcmqXT4%2BSxMOyHuiX%2BcV6tnRc04eHcxcdKfwbhlNt0wRXEnfsJHprsfgMprbaKDWL8W4QhrhAAAA

### Input code

```typescript
'test'.replaceAll('', '');
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2020",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true,
  "env": {
    "mode": "usage",
[...]

PR #6921, opened 2023-02-09
fix(es/preset-env): Fix default target

kdy1:
**Description:**

Previously the default value of `targets` was `""`, which is not deterministic.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6888.


kdy1 (commented):
swc-bump:
 - swc_ecma_preset_env

PR #6922, opened 2023-02-09
Use rkyv in strict mode

djkoloski:
*Note:* This PR turns on an rkyv feature that requires support from `wasmer`. I haven't done the work to push `strict` through `wasmer-types` yet.

**Description:**

By default, rkyv data may change format across Rust compiler versions. Enabling the `strict` feature guarantees that the data format will not change over time.

**Motivation:**

The end-to-end `plugins.compat.test.js` is failing on nightly.

**BREAKING CHANGE:**

Any rkyv-serialized data may change format. I'm not sure if SWC has an ABI or not, so this may not actually be considered a breaking change.

https://github.com/rkyv/rkyv/issues/338

### DIFF
--- a/crates/swc_atoms/Cargo.toml
+++ b/crates/swc_atoms/Cargo.toml
@@ -22,10 +22,10 @@ rkyv-bytecheck-impl = ["__rkyv", "rkyv-latest"]
 [dependencies]
 bytecheck = { version = "0.6.9", optional = true }
 once_cell = "1"
-rkyv      = { package = "rkyv", version = "=0.7.37", optional = true }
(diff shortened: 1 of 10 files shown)


kwonoj:
Oh this looks nice. Any elaborations may able to share what strict mode supposed to do & why it is not default behavior?

kwonoj:
Tracking: https://github.com/wasmerio/wasmer/issues/3695

kdy1:
Closing as https://github.com/swc-project/swc/pull/7197 merged this

kdy1:
Thank you!

PR #6932, opened 2023-02-10
fix(bindings/cli): Rename --config to --config-json

titanous:
**Description:**

`@swc/cli` uses a different format for `--config`, so use a different name.

**BREAKING CHANGE:**

Experimental `--config` option has been renamed to `--config-json`.

**Related issue (if exists):** https://github.com/swc-project/swc/issues/4017#issuecomment-1425273484

### DIFF
--- a/bindings/swc_cli/src/commands/compile.rs
+++ b/bindings/swc_cli/src/commands/compile.rs
@@ -28,11 +28,11 @@ use crate::util::trace::init_trace;
 /// Configuration option for transform files.
 #[derive(Parser)]
 pub struct CompileOptions {
-    /// Experimental: provide additional configuration to override the .swcrc.
-    /// Can be used to provide experimental plugin configuration,
+    /// Experimental: provide an additional JSON config object to override the
+    /// .swcrc. Can be used to provide experimental plugin configuration,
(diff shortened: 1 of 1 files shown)


alexeagle:
I wish it would just be `--config` and match semantics, but I agree this is an improvement over what's on HEAD right now, and you could add `--config` back to `swcx` later such that it matches.

titanous:
Unfortunately for the purposes of `rules_swc`, the old semantics are unworkable: we need to be able to pass arbitrary plugin configuration, which would require a really annoying flattening routine to spread every part of the config object  into separate CLI args and even then it would be lossy without type information.

Aghassi:
@titanous can you make this non breaking by leaving `--config` in and emitting a warning for using it and say it will be breaking in the future? Just makes the adoption right now slightly easier.

titanous:
I'd prefer to just rename it quickly given that it was already marked as experimental and this release has only been out for a few days. Otherwise the breakage in the future will be bigger, not everyone reads warnings.

Aghassi:
Sounds good to me thanks!

kdy1 (approved):
Thanks!

---

swc-bump:
 - dbg-swc

PR #6934, opened 2023-02-12
[Draft/WIP]: Brings back the wasmer upgrade that got reverted

NyanHelsing:
**Description:**

**BREAKING CHANGE:**


kwonoj:
I don't believe I'll try this until 

1. https://github.com/swc-project/swc/pull/6922 
2. With 1, upgrade rkyv separately verify layout works

ExProbitasFiducia:
> I don't believe I'll try this until
> 
>     1. [Use rkyv in strict mode #6922](https://github.com/swc-project/swc/pull/6922)
> 
>     2. With 1, upgrade rkyv separately verify layout works

Understand that; i dont expect this to get merged in right away.

I do want to do this with my own project though and Am happy to take direction on what next steps are to implement this given you mention there's a lot of other priorities for yourself.

kwonoj:
> I do want to do this with my own project though

Why swc's internal wasmer version is relavant to your project?

ExProbitasFiducia:
> > I do want to do this with my own project though
> 
> Why swc's internal wasmer version is relavant to your project?

yep because i need to run a plugin where the host is wasm and 2.x gives buffer undefined.

ExProbitasFiducia:
I guess i can run swc binary for now, but problem is that it's not sandboxed, where if i run wasm version of swc then it's sanboxed; this prevents some carefully crafted program from finding a way onto the host machine by some zero-day in swc; would need a zeroday in swc _&&_ zero-day in wasm runtime.

kwonoj:
> because i need to run a plugin where the host is wasm 

I'm not sure if I understood this correctly. Are you referring running swc/wasm and trying to run plugin? Then that's totally irrelevant as we do not support plugins in wasm host yet.

ExProbitasFiducia:
> > because i need to run a plugin where the host is wasm
> 
> I'm not sure if I understood this correctly. Are you referring running swc/wasm and trying to run plugin? Then that's totally irrelevant as we do not support plugins in wasm host yet.

Something like:

```
import { transform } from "../swc/bindings/binding_core_wasm/dist-deno/binding_core_wasm.js";

console.log(
    await transform(
        "console.log(\"Hello World\");",
        {
            jsc: {
                parser: {
                    syntax: "typescript"
                },
                target: "es2022",
                experimental: {
                    plugins: [
                       ["transform-extension", {}]
                    ]
                }
            },
            swcrc: false
        },
        {
            "transform-extension": await Deno.readFile("./transform-extension/transform_extension.wasm")
        }
    )
);
```

ExProbitasFiducia:
> > because i need to run a plugin where the host is wasm
> 
> I'm not sure if I understood this correctly. Are you referring running swc/wasm and trying to run plugin? Then that's totally irrelevant as we do not support plugins in wasm host yet.

Yeahh... its not supported because wasmer version prevents it from working. :thinking:

kwonoj:
1. Yes, `binding_core_wasm`. You can't run plugins in wasm host. It is expected.
3. What is dist-deno even? We don't have deno bindings.


Overall I'm starting to feeling this PR attempts to change core behavior unexpected way, at least the way you want to solve.

kwonoj:
> because wasmer version prevents it from working

No, *we* do not have fully working support for it yet. You can't make it work by upgrading wasmer.

ExProbitasFiducia:
> 1. Yes, `binding_core_wasm`. You can't run plugins in wasm host. It is expected.
> 
>     2. What is dist-deno even? We don't have deno bindings.
> 
> 
> Overall I'm starting to feeling this PR attempts to change core behavior unexpected way, at least the way you want to solve.

https://github.com/swc-project/swc/pull/6934/files#diff-cd6d875e1f0df1038c6c14faada55a1da1f5fd8c5ca6c60c82f8cc464db4e3deR1-R2

ExProbitasFiducia:
> > because wasmer version prevents it from working
> 
> No, _we_ do not have fully working support for it yet. You can't make it work by upgrading wasmer.

The error that is thrown when It tries to run the plugin is a bug in wasmer that was fixed in the 3.0 version so ya if swc upgrades wasmer the wasm-host plugins should work unless theres _another_ bug in adittion to that.

kwonoj:
1. No, the reason we intentionally carefully trying to proceed plugin support in wasm and it doesn't work is related to resolvers to load pluigins & other known failures in the road. Wasmer update is not the only reason. Especially the resolver syntax you used in the example is not supposed to be used currently.

2. I strongly do not believe we'll going to have `dist-deno` and these kind of sudden new bindings for the specific platforms. 

Overall, I would like to suggest to close this PR. Things we can promise are

1. we'll going to update rkyv, and validate
2. we'll going to upgrade wasmer one day
3. we'll going to evaluate & propose way to execute plugins in the wasm host

But we definitely won't mix 2&3 in a single PR in here, moreover as already mentioned 1 should be proceed first and it'll take reasonably long time. Lastly, regardless of 1-3 I'd strongly against to include `build_deno` or other platform support without deeper consideration such as feature parity like native bindings & wasm, as well as other logistics.

kdy1:
I agree with @kwonoj
We have lots of dependencies for this job, and once they are all resolved, we will do it one day.

PR #6936, opened 2023-02-13
feat(es/minifier): Drop expressions using sequential inliner

kdy1:
**Description:**

I'm trying to optimize

```js
import { upper } from "module";
let foobar = "foo";
export const foo = foobar;
const bar = "bar";
foobar += bar;
let foobarCopy = foobar;
foobar += "foo";
console.log(foobarCopy);
foobarCopy += "Unused";
function internal() {
    return upper(foobar);
}
```

Theoretically, a control flow graph is required for this optimization. But our sequential inliner shares some characteristics with it, so I'm investigating to check if this optimization is possible without a full control flow graph.


kdy1:
This is now ready for review, but I'll wait as much as possible

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

jridgewell (approved):

PR #6937, opened 2023-02-13
perf: patch string-cache with dashmap version

Boshen:
**Description:**

This is a preliminary testing for replacing the global `Mutex<HashMap>` with a `Dashmap`.

**BREAKING CHANGE:**

Maybe?

https://github.com/Boshen/string-cache/pull/1


Boshen:
superseded by upstream PR https://github.com/servo/string-cache/pull/268

PR #6938, opened 2023-02-13
doc(contributing): bold text

easonshimmer:
Signed-off-by: YangYi <yangyi.makeryi@gmail.com>

**Description:** Use bold syntax for the 'language' word.

**BREAKING CHANGE:** null

**Related issue (if exists):** null


kdy1:

Issue #6939, opened 2023-02-13
No diagnostic for JSX in automatic runtime with empty key which will cause a panic


dsherret:
### Describe the bug

When having an automatic runtime, code with a `key` attribute without a value will panic.

### Input code

```
const test = <div key></div>
```

```
thread 'transpiling::tests::test_transpile_jsx_import_source_empty_key' panicked at 'assertion failed: `(left != right)`
  left: `None`,
 right: `None`: value of property 'key' should not be empty', swc_ecma_transforms_react-0.160.9\src\jsx\mod.rs:536:41
```

https://github.com/swc-project/swc/blob/f1f5a19e617ba5f31a680f4a9007717fbcc93331/crates/swc_ecma_transforms_react/src/jsx/mod.rs#L536

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true,
      "decorators": true
    },
    "target": "es2019",
    "transform": {
      "useDefineForClassFields": true,
      "react": {
        "importSource": "https://esm.sh/preact",
        "runtime": "automatic"
      }
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.3.35&code=H4sIAAAAAAAAA0vOzysuUShJBRK2CjYpmWUK2amVdjb6QJadNQDu0vsJHQAAAA%3D%3D&config=H4sIAAAAAAAAAzWOy27CQAxF%2F8VrRAqLSmRbxK6rfoE1ccBV5iFfj9QI5d%2BZAbqzro%2Bvz51%2BEWi8U2GDWJ%2BwJuc%2FGsnXIgimxWlHjha5VdnRJCEbeza8kq1t2a7i7UZw%2FDicOm%2BcMGeLvbJCzjJrkku2r4WBi8oy4b%2FQhIN3TmPJ5j%2B5WpBWdnMvGIdBEPe4DeXFNb4m19gJrp4juwbatqYR81QX6U3d%2FanzST3XpPNK48wL2j%2FF9xt86j8Ajq3HkQQBAAA%3D

### Expected behavior

Diagnostic maybe?

### Actual behavior

Panics when transforming.

### Version

swc_ecma_parser = "0.124.5", swc_ecma_transforms_react = "0.160.9"

PR #6945, opened 2023-02-15
fix(es/react): Don't panic on `key` without a value

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6939.


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_react

Issue #6958, opened 2023-02-18
`.mts` and `.cts` should be treated as a typescript file by default


condemil:
### Describe the bug

```
/app # swc --version

@swc/cli: 0.1.62
@swc/core: 1.3.35

/app # cat main.ts
console.log('test')
/app # cat main.mts
console.log('test')
/app # cat main.cts
console.log('test')
/app # swc main.ts
Successfully compiled 1 file with swc.
console.log("test");


/app # swc main.mts

/app # swc main.cts

/app #
```

### Config

```json
no .swcrc file
```

### Expected behavior

Successfully compile *.mts/*.cts files

### Version

1.3.35

kdy1:
You need `.swcrc` currently

condemil:
```
/app # touch .swcrc
/app # swc main.mts

/app #
```

kdy1:
You need to configure parser, not just touch it

condemil:
Went though documentation once again and found out about `test`, tried to add the following, still no luck:

```json
[
  {
    "test": ".*\\.mts$",
    "jsc": {
      "parser": {
        "syntax": "typescript"
      }
    }
  }
]
```

PR #6959, opened 2023-02-18
fix(bindings/node): Treat `.cts` and `.mts` as input by default

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6958.


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #6962, opened 2023-02-19
feat: add preserve_imports_with_side_effects option for simplify optimization

### LINKED ISSUE #6771: Dropping unused imports by simplifier()
### Describe the feature

While i understand why it's not an always safe optimization, we rely on it at Qwik. This bug fix is preventing us to update to latest SWC.
https://github.com/swc-project/swc/issues/6503

It would be great to have an option to enable the previous behaviour where unused imports are dropped!

_No response_

### Additional context

_No response_

nissy-dev:
**Description:** Fix #6771

This implementation is still rough, and there are still some implementation decisions that need to be decided on that I mentioned in my comments.

**BREAKING CHANGE:** Yes


crates/swc/src/config/mod.rs
```diff
@@ -1481,6 +1485,9 @@ pub struct OptimizerConfig {
 
     #[serde(default)]
     pub jsonify: Option<JsonifyOption>,
+
+    #[serde(default)]
+    pub drop_unused_imports: BoolConfig<false>,
```
nissy-dev, line 1490: There was a way to modify the simplify option directly, but I added it in this way to avoid breaking change. What do you think?

crates/swc_bundler/src/bundler/optimize.rs
```diff
@@ -28,6 +28,7 @@ where
                         top_retain: Default::default(),
                     },
                     self.unresolved_mark,
+                    false,
```
nissy-dev, line 31: `swc_bundler` and `swc_ecma_minifier` use `dce` function, so I set false in this time. Should the user be able to set this value to arbitray value? I think this probably should, but I don't understand how to implement it in details so I would appreciate any advice you could give me.

kdy1: I think it does not matter, because `swc_bundler` is not actively maintained in favor of turbopack

crates/swc_ecma_transforms_optimization/src/simplify/dce/mod.rs
```diff
@@ -30,6 +30,7 @@ use crate::debug_assert_valid;
 pub fn dce(
     config: Config,
     unresolved_mark: Mark,
+    drop_unused_imports: bool,
```
kdy1, line 33: This is a breaking change anyway

nissy-dev: This is exactly... Sorry for your confusion 🙇‍♂️ 
I updated the PR comment.

nissy-dev:
#6967 is merged, so I closed.

kdy1:
I'll merge PRs with breaking changes at once

crates/swc/src/config/mod.rs
kdy1: Can you move this the `simplify` option?

kdy1 (changes requested):
Thank you!
As it's a breaking change anyway, we can use correct way.

manucorporat:
I am thinking that it would be soo good, to have a callback to resolve if a indentifier has side effects or not:

So instead of being a `bool` it could be a function, that is being passed the import source and return a bool

kdy1:
Rust-side api is fine, but js api is not fine

nissy-dev:
@manucorporat Which API do you use, JS API or Rust API?
It is possible to change the `simplifier` function (in `crates/swc_ecma_transforms_optimization/src/simplify/mod.rs`) to accept callbacks. If you want to do so, it would be great if you could give me more details about the I/F of the callback you want to add.

crates/swc/tests/fixture/globals/env-map-computed/input/.swcrc
```diff
@@ -2,7 +2,6 @@
     "jsc": {
         "transform": {
             "optimizer": {
-                "simplify": false,
```
kdy1, line 5: Why did you touch this? Please revert

nissy-dev: @kdy1 Thank you for your reviews. You mean I need to modify `simplify` option which receive boolean value and json? Like this.

```rust
pub struct OptimizerConfig {
    #[serde(default)]
    pub simplify: SimplifyOption,
}

#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(untagged)]
pub enum SimplifyOption {
    Bool(bool),
    Json(SimplifyJsonOption),
}

impl Default for SimplifyOption {
    fn default() -> Self {
        SimplifyOption::Bool(true)
    }
}

#[derive(Debug, Default, Clone, Copy, Serialize, Deserialize)]
#[serde(deny_unknown_fields, rename_all = "camelCase")]
pub struct SimplifyJsonOption {
    pub preserve_imports_with_side_effects: bool,
}
```

If you think so, I will refactor in this way

kdy1: Yes, thank you!

nissy-dev: I updated!

crates/swc_ecma_transforms_optimization/src/simplify/dce/mod.rs
```diff
@@ -30,6 +30,7 @@ use crate::debug_assert_valid;
 pub fn dce(
     config: Config,
     unresolved_mark: Mark,
+    preserve_imports_with_side_effects: bool,
```
kdy1, line 33: I think this should go into Config type

kdy1 (changes requested):

crates/swc_ecma_transforms_optimization/src/simplify/dce/mod.rs
nissy-dev: I moved this option to dce config.

kdy1 (approved):
Thank you!

---

swc-bump:

PR #6963, opened 2023-02-19
fix(es/resolver): Use type context for `export =` if it's unresolved

kdy1:
**Description:**

**Related issue:**

 - https://github.com/dudykr/stc/issues/717.


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base

PR #6458, opened 2022-11-16
fix(es/minifier): preserve unused imported specifiers

iheyunfei:
**Description:**

**BREAKING CHANGE:**

**Related issue (if exists):**

Fix #6192

PR #6967, opened 2023-02-20
fix(es/minifier): Drop unused import bindings

magic-akari:
**Description:**
It is always safe to remove unused import bindings and keep import statements to preserve side effects.
This transformation does not require additional options like `drop_unused_imports`.

**BREAKING CHANGE:**

- #6458
- #6962


kdy1 (approved):
Thank you!

---

swc-bump:

PR #6969, opened 2023-02-21
fix: properly handle CSS animation on CSS Modules

3c1u:
**Description:**

In the current implementation of `swc_css_modules`, any first identifier in `animation` properties are transformed as the local identifier.

Current behaviour:
```css
/* this */
.a {
  animation: 3s ease animationName;
}

/* will be transformed like this: */
.__local__a {
  animation: 3s __local__ease animationName;
}
```

Expected behaviour:
```css
/* this */
.a {
  animation: 3s ease animationName;
}

/* should be transformed like this: */
.__local__a {
  animation: 3s ease __local__animationName;
}
```

This pull request will fix this.

FYI: https://developer.mozilla.org/en-US/docs/Web/CSS/animation

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_atoms/words.txt
+++ b/crates/swc_atoms/words.txt
@@ -984,6 +984,7 @@ background-position-x
 background-position-y
 background-repeat
 background-size
+backwards
 base
 baseFrequency
 baseProfile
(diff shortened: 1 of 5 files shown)


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_css_modules

PR #6972, opened 2023-02-21
fix(es/decorator): Preserve class prop order

### LINKED ISSUE #6951: Decorated and regular class properties are translated in a wrong order
### Describe the bug

According to Playground demo, swc breaks the definition order of class fields: all the undecorated fields go first, and the decorated ones go last. This behavior leads to some problems if the undecorated field depends on the decorated field value in its class constructor.

TSC compiler doesn't change the order of fields in a constructor.

### Input code

_No response_

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "decorators": true
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    },
    "transform": {
      "legacyDecorator": true,
      "decoratorMetadata": true
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.3.35&code=H4sIAAAAAAAAA0vOSSwuVnBUqOZSUHDITypOLSpLTMpJVUhUsFUoKSpNteYCSiQBOWmJOcUQXnJ%2BXjFQKrkkv0hDE6yxlquWCwBWMAoXSQAAAA%3D%3D&config=H4sIAAAAAAAAA02Ouw6DMAxF%2F8Uzazswd%2BUjrGBQqiSObCM1Qvx7Ex6lm%2BV77mOFtzroV8goStIuLcnwAz1YyaROfDbowLS%2BJgxKHYzkWNBYFHqThbYqo8xk1UT6qHRgVvrx0Sc%2FlZbtOGYh1VvCNIeLbDmCSSeW2OhAM7ryutqOsr%2F6gQxHNDxXVHvkcWlx6z5%2Bn%2FOE7V5wtnodTnA3fgHD3DNVBQEAAA%3D%3D

### Expected behavior

Fields are translated in an order they were declared.

### Actual behavior

Order of fields is changed.

### Version

1.3.35

### Additional context

_No response_

magic-akari:
**Description:**

**BREAKING CHANGE:**

- fix #6951

### DIFF
--- a/crates/swc_ecma_transforms_proposal/src/decorators/legacy/mod.rs
+++ b/crates/swc_ecma_transforms_proposal/src/decorators/legacy/mod.rs
@@ -20,10 +20,9 @@ enum EnumKind {
     Num,
 }
 
-pub(super) fn new(metadata: bool, use_define_for_class_fields: bool) -> TscDecorator {
+pub(super) fn new(metadata: bool) -> TscDecorator {
     TscDecorator {
         metadata,
-        use_define_for_class_fields,
         enums: Default::default(),
         vars: Default::default(),
         appended_exprs: Default::default(),
@@ -36,9 +35,6 @@ pub(super) fn new(metadata: bool, use_define_for_class_fields: bool) -> TscDecor
 
 pub(super) struct TscDecorator {
     metadata: bool,
-
-    #[allow(dead_code)]
-    use_define_for_class_fields: bool,
 
     enums: AHashMap<JsWord, EnumKind>,
 
--- a/crates/swc_ecma_transforms_proposal/src/decorators/legacy/mod.rs
+++ b/crates/swc_ecma_transforms_proposal/src/decorators/legacy/mod.rs
@@ -36,6 +36,8 @@ pub(super) fn new(metadata: bool, use_define_for_class_fields: bool) -> TscDecor
 
 pub(super) struct TscDecorator {
     metadata: bool,
+
+    #[allow(dead_code)]
     use_define_for_class_fields: bool,
 
     enums: AHashMap<JsWord, EnumKind>,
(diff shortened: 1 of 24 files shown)


magic-akari:
https://github.com/swc-project/swc/blob/89a29ad3ec6b5ee5a87434957fb47514d5f917ce/crates/swc/tests/exec/issues-2xxx/2655/1/exec.ts#L36-L38

This test case failed on `target: ES2022`, since we do not transform class field, the `useDefineForClassFields: false` does not work.
Any idea?

kdy1:
We need sort of directive, but I think we can disable it or have a dirty if-else hack in the test matrix for now, and either is fine

kdy1:
Nevermind. You can add .swcrc to force transform

magic-akari:
https://user-images.githubusercontent.com/7829098/220529716-d7c99de0-b378-4f8e-aa9a-f58152064619.mov

With `target: ES2020`, sWc does not transpire class field.
And ECMA native class field always mean  `[[Define]]`.

kdy1:
I think we should lower the jsc.target in this case

crates/swc_ecma_transforms_proposal/src/decorators/legacy/mod.rs
```diff
@@ -36,6 +36,8 @@ pub(super) fn new(metadata: bool, use_define_for_class_fields: bool) -> TscDecor
 
 pub(super) struct TscDecorator {
     metadata: bool,
+
+    #[allow(dead_code)]
     use_define_for_class_fields: bool,
```
magic-akari, line 41: We should leave `class fields` to be handled in `swc_ecma_transforms_compat::es2022::class_properties` stage, not `decorators`.

This `use_define_for_class_fields` option is not used in this transformation.
Is it safe to remove it wihout breaking changes?

kdy1: I think so

kdy1: It's passed, not deserialized from json

kdy1 (approved):
Thank you!

---

swc-bump:

PR #6973, opened 2023-02-21
fix(cli): Use the source-file-name and source-root options

realtimetodie:
Adds the `--source-file-name` and `--source-root` options to allow the user to set the source file URL to the original source file and the URL root from which all sources are relative respectively.

The implementation follows the Node.js @swc/cli

https://github.com/swc-project/cli/blob/461231917181133fd2e98b954fce07b599711502/src/swc/compile.ts#L20

https://github.com/swc-project/cli/blob/461231917181133fd2e98b954fce07b599711502/src/swc/util.ts#L58-L59

Example

```
$ swc compile --source-maps true --source-file-name demo.ts --source-root "./foo" input.ts
```

```
...
{"version":3,"sources":["demo.ts"],"sourceRoot":"./foo","sourcesContent":["export const a: string = \"a\";\n\n"],"names":["a"],"mappings":"AAAA,OAAO,IAAMA,IAAY,IAAI"}
```

The `source_root` feature was added to the [sourcemap](https://crates.io/crates/sourcemap) crate in [v6.2.0](https://github.com/getsentry/rust-sourcemap/releases/tag/6.2.0) (kudos @kamilogorek).

Related

- https://github.com/swc-project/swc/discussions/3859
- https://github.com/swc-project/swc/discussions/6479
- https://github.com/swc-project/swc/issues/4017
- https://github.com/swc-project/swc/issues/1388
- https://github.com/aspect-build/rules_swc/issues/166


realtimetodie:
CI fails due to an unrelated error

```
Error: Couldn't find package "@babel/code-frame@^7.18.6" required by "@babel/template@^7.20.7" on the "npm" registry.
```

bindings/swc_cli/src/commands/compile.rs
```diff
@@ -239,16 +251,13 @@ fn emit_output(
...
-            file_path.display(),
-            output.code,
-            output
-                .map
-                .as_ref()
-                .map(|m| m.to_string())
-                .unwrap_or_default()
-        );
+        let source_map = if let Some(ref source_map) = source_map {
+            String::from_utf8_lossy(source_map).to_string()
```
kdy1, line 255: I think we should use cow instead?

bindings/swc_cli/src/commands/compile.rs
```diff
@@ -239,16 +251,13 @@ fn emit_output(
...
+        } else {
+            String::new()
```
kdy1, line 257: I think `Cow::Borrowed("")` will work

kdy1 (changes requested):
Thank you!

---

swc-bump:
 - dbg-swc

bindings/swc_cli/Cargo.toml
```diff
@@ -26,6 +26,7 @@ rayon = "1"
 relative-path = "1.6.1"
 serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1", features = ["unbounded_depth"] }
+sourcemap = "6"
```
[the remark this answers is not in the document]
realtimetodie: Yes that is very good. The version should be pinned to 6.2.

realtimetodie:
I pinned the version of the sourcemap crate to 6.2 globally.

kwonoj (dismissed):

realtimetodie:
After updating the sourcemap crate for `swc_common`, a fixture test fails. The string of base64 VLQs which contain the actual mappings of the generated source map are different.

```
crates/swc/tests/fixture/sourcemap/004
```

```diff
- "mappings": "AAACA,CAAAA,KAAKC,gBAAmB,GAAGD,KAAKC,gBAAmB,IAAI,EAAC,AAAD,EAAIC,IAAI,CAAC;IAC7D;QACI;KACH;IACD;QACU,MAAY,SAAUC,CAAuB,EAAEC,CAAmB,EAAEC,CAAmB,EAAE;YAC3F;YACA,IAAIC,IAAa,SAAoBC,CAAI,EAAE;gBACvC,IAAIC,IAAOD,EAAKC,IAAI;gBACpB,OAAsB,CAAA,GAAGC,EAA+CC,GAAE,AAAFA,EAAK,OAAO;oBAChFC,UAAUH,EAAKI,GAAG;gBACtB;YACJ;YACAP,EAAoBQ,CAAC,CAACT,IACDC,EAAoBS,CAAC,CAACV,GAAqB;gBACvCW,SAAS,WAAoB;oBAC9C,OAAqBA;gBACzB;gBACqBC,SAAS,WAAoB;oBAC9C,OAAqBV;gBACzB;YACJ,EAAA;YACqB,IAAIG,IAAiDJ,EAAoB,OAC1FU,IAAU,CAAA;QAE1B;QACc,MAAY,SAAUZ,CAAuB,EAAEc,CAAwB,EAAEZ,CAAmB,EAAE;YAC/Fa,CAAAA,OAAOC,QAAQ,GAAGD,OAAOC,QAAQ,IAAI,EAAC,AAAD,EAAIjB,IAAI,CAAC;gBAC3C;gBACA,WAAY;oBACR,OAAOG,EAAoB;gBAC/B;aACH;QAGb;IACI;IACS,SAAUA,CAAmB,EAAE;QAK3BA,EAAoBe,CAAC,CAAC,GAAG;YAC9B;YACA;YACA;SACH,EAAE,WAAY;YACX,OAPOf,EAAoBA,EAAoBgB,CAAC,GAOxB;QAC5B,IAESC,OAD0BjB,EAAoBe,CAAC,EAAA;IAGhE;CACC",
+ "mappings": "AAACA,CAAAA,KAAKC,gBAAmB,GAAGD,KAAKC,gBAAmB,IAAI,EAAC,AAAD,EAAIC,IAAI,CAAC;IAC7D;QACI;KACH;IACD;QACU,MAAY,SAAUC,CAAuB,EAAEC,CAAmB,EAAEC,CAAmB,EAAE;YAC3F;YACA,IAAIC,IAAa,SAAoBC,CAAI,EAAE;gBACvC,IAAIC,IAAOD,EAAKC,IAAI;gBACpB,OAAqB,AAAC,CAAA,GAAGC,EAA+CC,GAAE,AAAFA,EAAK,OAAO;oBAChFC,UAAUH,EAAKI,GAAG;gBACtB;YACJ;YACAP,EAAoBQ,CAAC,CAACT,IACDC,EAAoBS,CAAC,CAACV,GAAqB;gBACvCW,SAAS,WAAoB;oBAC9C,OAAqBA;gBACzB;gBACqBC,SAAS,WAAoB;oBAC9C,OAAqBV;gBACzB;YACJ,EAAA;YACqB,IAAIG,IAAiDJ,EAAoB,OAC1FU,IAAU,CAAA;QAE1B;QACc,MAAY,SAAUZ,CAAuB,EAAEc,CAAwB,EAAEZ,CAAmB,EAAE;YAC/Fa,CAAAA,OAAOC,QAAQ,GAAGD,OAAOC,QAAQ,IAAI,EAAC,AAAD,EAAIjB,IAAI,CAAC;gBAC3C;gBACA,WAAY;oBACR,OAAOG,EAAoB;gBAC/B;aACH;QAGb;IACI;IACS,SAAUA,CAAmB,EAAE;QAK3BA,EAAoBe,CAAC,CAAC,GAAG;YAC9B;YACA;YACA;SACH,EAAE,WAAY;YACX,OAPOf,EAAoBA,EAAoBgB,CAAC,GAOxB;QAC5B,IAESC,OAD0BjB,EAAoBe,CAAC,EAAA;IAGhE;CACC",
```

Should I revert the update of the sourcemap crate for `swc_common` or should I update the fixture test instead?

kamilogorek:
The diff is adding an additional `AAAC`, which in mappings terms translates to "add `1` to a column number".
This is a valid behavior due to the fix that we had to apply here https://github.com/getsentry/rust-sourcemap/pull/54 and which is based on discussion here https://github.com/getsentry/rust-sourcemap/pull/53

realtimetodie:
https://github.com/browserslist/browserslist-rs/issues/15

realtimetodie:
Some tests fail. GitHub actions is unable to run the git checkout step: HTTP status 429. I will retry tomorrow.

```
unable to access 'https://github.com/swc-project/swc/': The requested URL returned error: 429
````

**Review by kdy1 — APPROVED** [review]
Thank you!

**Review by kdy1 — CHANGES_REQUESTED** [review]
CI failed

**Comment by realtimetodie** [comment]
The tests basically compare the generated source maps from the sourcemap crate with the generated source maps from the Node.js [terser](https://www.npmjs.com/package/terser) package

```
test result: FAILED. 1929 passed; 54 failed;
```

https://github.com/swc-project/swc/blob/447c1dad933dd2058b1f18b2c41137b6a908570b/crates/swc_ecma_codegen/tests/sourcemap.rs#L455

@kdy1 Is this a requirement to have parity with terser?

The reason why the tests fail seems to be related to https://github.com/getsentry/rust-sourcemap/pull/53 as mentioned by @kamilogorek in https://github.com/swc-project/swc/pull/6973#issuecomment-1440671331. Terser uses the Node.js [@jridgewell/gen-mapping](https://www.npmjs.com/package/@jridgewell/gen-mapping) package to generate source maps. Was this issue in https://github.com/getsentry/rust-sourcemap/pull/53 fixed in the Node.js @jridgewell/gen-mapping package too?

https://github.com/swc-project/swc/blob/447c1dad933dd2058b1f18b2c41137b6a908570b/yarn.lock#L7266

Ping @jridgewell short summary: this PR updates the [sourcemap](https://crates.io/crates/sourcemap) crate to a version that includes your PR https://github.com/getsentry/rust-sourcemap/pull/53. However, some tests are failing (see above).

**Comment by kdy1** [comment]
It does not have to be same as `terser`. I just looked for a way to ensure that source maps are correct _enough_.
But not sure how to verify it without relying on some other tool or reviewing an enormous amount of source map changes.

@jridgewell Do you have any idea?

**Comment by jridgewell** [comment]
So the sourcemap itself hasn't changed:
```
[crates/swc_ecma_codegen/tests/sourcemap.rs:456] actual.tokens().collect::<Vec<_>>() = [
    <Token 017a45a1919f4006.js:0:0 (0:0)>,
    <Token 017a45a1919f4006.js:0:3 (0:2)>,
    <Token 017a45a1919f4006.js:0:10 (0:8)>,
    <Token 017a45a1919f4006.js:0:14 (0:12)>,
    <Token 017a45a1919f4006.js:0:16 (0:13)>,
    <Token 017a45a1919f4006.js:0:18 (0:14)>,
    <Token 017a45a1919f4006.js:0:27 (0:23)>,
    <Token 017a45a1919f4006.js:0:28 (0:24)>,
    <Token 017a45a1919f4006.js:0:29 (0:24)>,
]
```

Notice those last two tokens. They have the same generated location (`0:24`), but difference source location (`0:28` vs `0:29`). The bug we're seeing is the lookup token returning one or the other. Before, we'd see the last one, and now, we're seeing the first one.

In https://github.com/getsentry/rust-sourcemap/pull/54, we switched from a [manual](https://github.com/getsentry/rust-sourcemap/blob/6.2.0/src/types.rs#L617-L634) greatest-lower-bound binary search to one [using](https://github.com/getsentry/rust-sourcemap/blob/6.2.1/src/utils.rs#L122-L125) Rust's `binary_search_by_key`. The big difference that I can see between the impls is the exact match condition: https://github.com/rust-lang/rust/blob/07c993eba8b76eae497e98433ae075b00f01be10/library/core/src/slice/mod.rs#L2529-L2532. Before, we'd continue searching even after finding an exact match, guaranteeing we'd get the lowest index exact-match. Now, we'll stop on any exact-match.

I think this is a bug in sourcemap, caused because I didn't really look up the binary search behavior when it was suggested we switch to the stdlib impl. For reference, my `trace-mapping` library [guarantees](https://github.com/jridgewell/trace-mapping/blob/23ab4a29244444d825044ff8fb9a8d57f41837fa/src/trace-mapping.ts#LL434) the first/last item when tracing, Mozilla's v0.6.1 (the JS one) `source-map` [does](https://github.com/mozilla/source-map/blob/0.6.1/lib/binary-search.js#L100-L108), but Mozilla's v0.7 (the Rust/WASM one) [doesn't](https://github.com/fitzgen/source-map-mappings/blob/197660faf3dd1167c643797a5ed62cb687482c54/src/lib.rs#L383).

**Comment by kamilogorek** [comment]
`6.2.2` has been released, you can bump it here now

**Comment by jridgewell** [comment]
Ah god, I think I read it in the wrong direction. The test was failing because we received `28` where before it received `29` (the opposite of what I said in https://github.com/swc-project/swc/pull/6973#issuecomment-1442361534). From that, I assumed the old `sourcemap` code was returning the lowest match index, when in fact it seems like it was returning the highest match index. I assumed that because that's what the JS sourcemap packages do.

@kamilogorek: We could revert `6.2.2` and implement the highest-index behavior to match the old `6.2.0` behavior. Or we could align with the JS ecosystem, in which case SWC would need to skip these failing tests (or stop inserting multiple tokens for the same generated position).

**Comment by kdy1** [comment]
I think it's better to align with js ecosystem, so we may inline source maps for those tests. But I think it should be done in a separate PR

**Comment by kamilogorek** [comment]
@swatinem what do you think?

**Comment by Swatinem** [comment]
I’m not sure I fully understand the problem at hand.

> stop inserting multiple tokens for the same generated position

I believe this might be the solution. We shouldn’t even get to the situation where there is more than one token matching (and picking the lowest or higher match). What would that even mean in sourcemaps that you have more than one token for one minified source position?

**Comment by realtimetodie** [comment]
@kdy I created a follow-up PR in https://github.com/swc-project/swc/pull/6989 and reverted the update for the swc component crates.

File: bindings/swc_cli/Cargo.toml [hunk 4]
```diff
@@ -26,6 +26,7 @@ rayon = "1"
 relative-path = "1.6.1"
 serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1", features = ["unbounded_depth"] }
+sourcemap = "6.2.2"
```
kdy1, line 29: Please revert this dep upgrade

kdy1 (changes requested):

bindings/swc_cli/Cargo.toml
realtimetodie: As mentioned in https://github.com/swc-project/swc/pull/6973#issue-1594003697, we need the source_root feature that was added to the [sourcemap](https://crates.io/crates/sourcemap) crate in [v6.2.0](https://github.com/getsentry/rust-sourcemap/releases/tag/6.2.0).

realtimetodie: The sourcemap crate is only used to serialize/deserialize the source map. It is not used to generate a new one.

kdy1: They are used. That's how the cargo work

kdy1: But well, I think this would be fine

kdy1 (approved):
swc-bump:
 - dbg-swc

kdy1 (changes requested):
Please revert the top level `Cargo.lock`

realtimetodie:
There were no changes in the main Cargo.lock file, the git history differed.

kdy1 (approved):

PR #6980, opened 2023-02-22
(perf) bump "string-cache" crate for per-bucket mutex

YoniFeng:
**Description:**

Bump the "string-cache" crate to 0.8.5, which includes [this PR](https://github.com/servo/string-cache/commit/b473a4ad3be989166031f56976f7ce54ae79ac05).
Internally, it now locks on each individual bucket within the DYNAMIC_SET struct, rather than locking on the whole data structure.
Parallel performance should improve.

**BREAKING CHANGE:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3041,9 +3041,9 @@ checksum = "213701ba3370744dcd1a12960caa4843b3d68b4d1c0a5d575e0d65b2ee9d16c0"
 
 [[package]]
 name = "string_cache"
-version = "0.8.4"
+version = "0.8.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "213494b7a2b503146286049378ce02b482200519accc31872ee8be91fa820a08"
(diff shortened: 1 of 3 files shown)


YoniFeng:
Rust newbie.

1. I didn't bump `swc_common`
2. Bumping breaks browserslist-rs v0.12.2.
Locally I altered back to `0.8.4` but still got the build break, so I convinced myself it it's a fake issue.
Needed to modify the lockfile with `cargo update`.

Looking at the issue in `browserslist-rs` (makes no sense on the surface level - string_cache's APIs didn't change).
Will need to PR there, wait for publish and bump dependency version here.

YoniFeng:
Waiting on the breaking change getting reverted and a new `string_cache` version published (with the Mutex PR)

YoniFeng:
@kdy1
is there need/want to benchmark before merging, or should we just look at the regular CI benchmark result?

kdy1:
I think post-merge benchmark is fine

YoniFeng:
Is something missing for merge?

kdy1 (approved):
swc-bump:
 - swc_atoms

PR #6993, opened 2023-02-28
refactor(es/parser): Make lexer not generic over `Input`

kdy1:
**Description:**

Currently, it uses `StringInput` as the input type, but I'll refactor it to use `String` or `Vec<u8>` directly to optimize it further.

**Related issue:**

 - https://github.com/swc-project/swc/discussions/6991


kdy1 (commented):
swc-bump:
 - swc_ecma_parser --breaking

PR #7000, opened 2023-03-02
fix: visitBindingIdentifier should go through type annotations

cyjake:
**Description:**

`Visitor#visitBindingIdentifier()` currently doesn't go through type annotations, which makes customized visitor easy to overlook.

**BREAKING CHANGE:**

Changing the function type shouldn't break existing code because BindingIdentifier can be coerced to Identifier implicitly. Adding some specific typescript tests can better justify that, any suggestions about adding those?

n/a


kdy1 (approved):
swc-bump:
 - dbg-swc

PR #7002, opened 2023-03-02
fix: Fix the problem that the inline sourcemap file reports an error …

limerickgds:
…when inputSourceMap is set to true

**Description:** When I configure InputSourcemap to true in the project, swc can only handle sourcemapurl of file type, and if I set it to inline, swc can only handle inline, otherwise an error will be reported. I hope that swc can recognize the content of sourcemapurl instead of requiring me to decide what configuration should be set based on the content in source content.Similar to babel's processing method

**BREAKING CHANGE:** false


kdy1:
I'm not going to accept this PR, as there's `InputSourceMap::Str("inline")`. Use correct option.

limerickgds:
@kdy1  when the sourcemapUrl of the source file is file, setting InputSourceMap::Str("inline") will report an error:
But when the sourcemapUrl of the source file is file, an error will be reported:
Caused by:
    relative URL without a base

At the same time, the setting of babel does not need to distinguish between true and inline, can we be more automated

kdy1:
> when the sourcemapUrl of the source file is file, setting InputSourceMap::Str("inline") will report an error

It's expected

kdy1:
Why do you expect external source map to work when you give `"inline"` as the option?

limerickgds:
> Why do you expect external source map to work when you give "inline" as the option?
I just want to be able to automatically process `"file"` or `"inline"`, instead of requiring me to analyze the file and choose a type to process. When I set` InputSourceMap` to true, it doesn't work, I followed what you said, using `InputSourceMap::Str("inline")` as well.
But I re-read the implementation, maybe I can resubmit a pr, you can review some

jridgewell:
I agree with @limerickgds here. It's cumbersome to know beforehand whether the file has an inline or external source map without inspecting beforehand. We shouldn't force the user to know that, it should just be a boolean "allow input maps" or "ignore input maps" and let SWC figure out whether it's inline or external. This is how [Babel](https://babeljs.io/docs/options#inputsourcemap) does it (technically, they also allow you to manually pass a `SourceMap` instance, too, but that a separate feature).

limerickgds:
@kdy1 
I'm writing a build tool to compile modules from babel to swc. There will be two types of sourcemap files in the project code, one inline sourcemap and some .map files. When I use swc transform to run, I need to paser the file to analyze the sourcemap type of the file, and then set the swc option to compile, the performance will be relatively low. It turns out that when using babel, I don't need to make such a setting, just set it to true, and babel will automatically recognize sourcemapurl to deal with this kind of problem.

@jridgewell thank you for your understanding

crates/swc/src/lib.rs
```diff
@@ -269,99 +269,119 @@ impl Compiler {
...
 
-                            let path = match src_mapping_url {
-                                Some(src_mapping_url) => {
-                                    dir.join(src_mapping_url).display().to_string()
-                                }
-                                None => {
-                                    format!("{}.map", dir.join(filename).display())
-                                }
-                            };
+                                    let file = File::open(&path)
```
kdy1, line 340: Please use `read_to_string` instead.

crates/swc/src/lib.rs
```diff
@@ -269,99 +269,119 @@ impl Compiler {
...
+                                        .or_else(|err| {
+                                            // Old behavior. This check would prevent
+                                            // regressions.
+                                            let f = format!("{}.map", filename.display());
+
+                                            match File::open(&f) {
```
kdy1, line 346: Same here

crates/swc/src/lib.rs
```diff
@@ -269,99 +269,119 @@ impl Compiler {
...
-                                })
-                                .context("failed to open input source map file");
-
-                            let file = if !is_default {
-                                file?
-                            } else {
-                                match file {
-                                    Ok(v) => v,
-                                    Err(_) => return Ok(None),
+                                    Ok(Some(sourcemap::SourceMap::from_reader(file).with_context(
```
kdy1, line 362: `from_slice` should be used

crates/swc/src/lib.rs
```diff
@@ -269,99 +269,119 @@ impl Compiler {
...
+                                            format!(
+                                                "failed to read input source map from file at {}",
+                                                path
+                                            )
+                                        },
+                                    )?))
                                 }
-                            };
-
-                            Ok(Some(sourcemap::SourceMap::from_reader(file).with_context(
```
kdy1, line 320: This is my mistake

kdy1:
Sorry, I didn't know that babel supports both on `true`.
@jridgewell Thank you!

kdy1 (commented):
swc-bump:
 - swc

limerickgds:
> Sorry, I didn't know that babel supports both on `true`. @jridgewell Thank you!

@kdy1 https://babeljs.io/docs/options#inputsourcemap This is the babel configuration. There is no need to support `inline`, I did not completely cancel the swc configuration option. Because this is a `break change`, this is left to you to decide, okay?

kdy1:
Yes, I know and I don't want to make a breaking change at the moment, and even if it's the case, it should be another PR.

crates/swc/src/lib.rs
limerickgds: @kdy1 sourcemap::SourceMap::from_reader(file)

kdy1: Oh I looked at source code and it uses BufReader internally.

kdy1 (approved):

jridgewell:
I think for the time being, just treating the two options as equivalent and supporting both inline and external for both is the right decision.

kdy1:
It sounds good

kdy1 (changes requested):
Can you address https://github.com/swc-project/swc/pull/7002#issuecomment-1455384200 too?

limerickgds:
> Can you address [#7002 (comment)](https://github.com/swc-project/swc/pull/7002#issuecomment-1455384200) too?

My understanding is that inputSourcemap setting `true` and `inline` should be the same processing flow, right?
If this is the meaning, I agree, so that the branch will not increase the complexity.

kdy1:
Yes, exactly

crates/swc_ecma_preset_env/data/core-js-compat/data.json
```diff
@@ -1082,6 +1082,42 @@
     "safari": "10.0",
     "samsung": "5.0"
   },
+  "es.array.to-reversed": {
```
kdy1, line 1085: Is this change intentional?
Seems like you invoked `yarn`

yarn.lock
```diff
@@ -1,7797 +1,5394 @@
-# This file is generated by running "yarn install" inside your project.
```
kdy1, line 1: Please revert this

kdy1 (changes requested):
Please revert the lockfile and preset-env data

crates/swc_ecma_preset_env/data/core-js-compat/data.json
limerickgds: done

kdy1 (approved):
Thank you!

---

PR #7006, opened 2023-03-03
feat(es/minifier): Optimize calls to `Boolean`/`Number`/`String`/`Symbol`

Austaras:
**Description:**

**BREAKING CHANGE:**


crates/swc_ecma_minifier/tests/terser/compress/issue_269/regexp/output.js
```diff
@@ -2,4 +2,4 @@
 /bar/gi;
 RegExp(foo);
 RegExp("bar", ig);
-RegExp("should", "fail");
+/should/afil;
```
Austaras, line 5: This should be fine as both of them would evaluate to immediate error

kdy1 (approved):
Thank you!

----

swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/tests/terser/compress/issue_269/regexp/output.js
magic-akari, line 5: This is not safe. Polyfill is possible for the RegExp constructor, but it is not possible for RegExp literals.

kdy1 (approved):

jridgewell (approved):

PR #7013, opened 2023-03-04
feat(es/transform): support for importing namespaces from `const_modules`

magic-akari:
**Description:**

**BREAKING CHANGE:**


magic-akari:
The current approach to defining const_modules involves using the Record<ident, expression> format, as exemplified below:

```JSON
{
  "globals": {
    "module_foo": {
      "DEBUG": "true"
    }
  }
}
```

I feel that this method to defining const_modules is a bit cumbersome. Instead, I prefer to use the Record<ident, value> format, as shown below:

```JSON
{
  "globals": {
    "module_foo": {
      "DEBUG": true
    }
  }
}
```

What is your opinion?

Perhaps I am looking for something like const_json_modules.

kdy1:
I agree with your opinion.

kdy1:
Storing an expression in a json file is quite cumbersome

kdy1 (approved):
Thanks!

---

swc-bump:

PR #7029, opened 2023-03-07
feat(common): Add an API to create a `SourceFile` without allocation

kdy1:
**Description:**

Actually allocation is not required in this case.


kdy1 (commented):
swc-bump:
 - swc_common

PR #7030, opened 2023-03-07
feat(es/transform): introduce `NodeIgnoringSpan`

hyf0:
**Description:**

- Introduce `NodeIgnoringSpan`, which will ignore `Span` while doing `eq` or `hash`.
- Improve the performance of `InlineGlobals`.

**BREAKING CHANGE:**


crates/swc_ecma_utils/src/expr_ignore_span.rs
```diff
@@ -0,0 +1,132 @@
...
+impl<'a, Node: EqIgnoreSpan + ToOwned + Debug> PartialEq for NodeIgnoringSpan<'a, Node> {
+    fn eq(&self, other: &Self) -> bool {
+        Ident::within_ignored_ctxt(|| self.0.eq_ignore_span(&other.0))
+    }
+}
+
+impl<'a, Node: EqIgnoreSpan + ToOwned + Debug> Eq for NodeIgnoringSpan<'a, Node> {}
+
+// TODO: This only a workaround for Expr. we need something like
+// `hash_ignore_span` for each node in the end.
```
hyf0: I think we need `hash_ignore_span` in the end if we want to make this implementation energetic.

And the code could be rewritten like
```rust
impl<'a, Node: HashIgnoreSpan> Hash for NodeIgnoringSpan<'a, Node> {
    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
          self.hash_ignore_span(state)
        }
    }
}
```

However, I think this could be done in another PR.

crates/swc_ecma_transforms_optimization/src/inline_globals.rs
```diff
@@ -39,7 +38,12 @@ pub fn inline_globals2(
     as_folder(InlineGlobals {
         envs,
         globals,
-        global_exprs,
+        global_exprs: Lrc::new(
```
hyf0, line 41: That's the hard part I need guidance. I don't find a good way to avoid `clone` without changing the function signature. The current implementation is just a POC. @kdy1 cc

kdy1: Changing function signature is fine

crates/swc_ecma_utils/src/node_ignore_span.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    #[inline]
+    pub fn owned(expr: <Node as ToOwned>::Owned) -> Self {
+        Self(Cow::Owned(expr))
+    }
+}
+
+impl<'a, Node: EqIgnoreSpan + ToOwned + Debug> PartialEq for NodeIgnoringSpan<'a, Node> {
+    fn eq(&self, other: &Self) -> bool {
+        Ident::within_ignored_ctxt(|| self.0.eq_ignore_span(&other.0))
```
kdy1, line 29: I think we can move `Ident::within_ignored_ctxt` to the caller side.

kdy1: Sometimes we need to match on syntax context

kdy1 (changes requested):
Thank you!

---

swc-bump:
 - swc_ecma_utils

crates/swc_ecma_utils/src/node_ignore_span.rs
kdy1: And actually, I think it's better to touch TLS less time. It's quite slow if you don't use LTO

hyf0: Done

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_utils

PR #7031, opened 2023-03-08
perf(es/parser): Use Cow for `_` in numeric literals

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_ecma_parser

Boshen:
Excuse me for the lack of knowledge here, but I don't see how this removes the allocation from `raw_val.replace('_', "")`? 

What I did for my lexer (copied from jsparagus) was to use the `Cow` technique, where we only own the string if there is a `_`.

kdy1:
This PR does not remove allocation of `replace` call

kdy1:
I'll remove it too

Issue #7019, opened 2023-03-06
perf(ast): shrink large struct sizes


Boshen:
## Background

https://nnethercote.github.io/perf-book/type-sizes.html

> Shrinking oft-instantiated types can help performance.

## Problem

I have gathered a list of large structs (> 64 bytes) inside `swc_ecma_ast`, by using `USTFLAGS=-Zprint-type-sizes ` with the help of [top-type-sizes](https://github.com/loyd/top-type-sizes):

<details>
  <summary>Top Type Sizes</summary>
  
  ```
216 jsx::JSXElement align=8
    104 opening
     72 closing
     24 children
     12 span

208 stmt::TryStmt align=8
    112 handler
     40 block
     40 finalizer
     12 span

160 alloc::sync::ArcInner<std::thread::Inner> align=8
    144 data
      8 strong
      8 weak

144 std::thread::Inner align=8
    120 parker
     16 name
      8 id

136 jsx::JSXAttrOrSpread align=8
    136 variant JSXAttr
     24 variant SpreadElement

136 jsx::JSXAttr align=8
     72 value
     48 name
     12 span

120 std::sys::unix::thread_parker::pthread::Parker align=8
     64 lock
     48 cvar
      8 state
      0 _pinned

120 function::Function align=8
     40 body
     24 params
     24 decorators
     12 span
      8 type_params
      8 return_type
      1 is_generator
      1 is_async

120 class::Constructor align=8
     40 key
     40 body
     24 params
     12 span
      1 accessibility
      1 is_optional

120 class::ClassMember align=8
    120 variant Constructor
    112 variant ClassProp, PrivateProp
     80 variant Method, PrivateMethod
     64 variant StaticBlock
     56 variant TsIndexSignature
     20 variant Empty

112 typescript::TsType align=8
      4 <discriminant>
    108 variant TsTypeQuery
     92 variant TsMappedType, TsLitType
     84 variant TsImportType
     76 variant TsInferType
     68 variant TsFnOrConstructorType
     52 variant TsTypeRef, TsUnionOrIntersectionType, TsConditionalType, TsTypePredicate
     44 variant TsTypeLit, TsTupleType
     36 variant TsIndexedAccessType
     28 variant TsArrayType, TsOptionalType, TsRestType, TsParenthesizedType, TsTypeOperator
     16 variant TsKeywordType
     12 variant TsThisType

112 typescript::TsInterfaceDecl align=8
     40 body
     24 id
     24 extends
     12 span
[...]

PR #7034, opened 2023-03-08
perf(es/ast): Box `Stmt` and `ModuleDecl`

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/7019.

### DIFF
--- a/crates/swc_bundler/src/bundler/chunk/cjs.rs
+++ b/crates/swc_bundler/src/bundler/chunk/cjs.rs
@@ -96,7 +96,7 @@ fn wrap_module(
     local_ctxt: SyntaxContext,
     load_var: Ident,
     mut dep: Module,
-) -> Stmt {
+) -> Box<Stmt> {
     {
         // Remap syntax context of `module` and `exports`
         // Those are unresolved, but it's actually an injected variable.
@@ -154,7 +154,7 @@ fn wrap_module(
 
     // var load = __swcpack_require__.bind(void 0, moduleDecl)
 
-    Stmt::Decl(Decl::Var(Box::new(VarDecl {
+    Box::new(Stmt::Decl(Decl::Var(Box::new(VarDecl {
         span: DUMMY_SP,
         kind: VarDeclKind::Var,
         declare: false,
@@ -174,7 +174,7 @@ fn wrap_module(
             }))),
             definite: false,
         }],
-    })))
+    }))))
 }
 
 struct RequireReplacer<'a, 'b, L, R>
@@ -229,7 +229,7 @@ where
             return;
         }
 
-        if let ModuleItem::ModuleDecl(ModuleDecl::Import(i)) = node {
+        if let ModuleItem::ModuleDecl(box ModuleDecl::Import(i)) = node {
             let dep_module_id = self
                 .base
                 .imports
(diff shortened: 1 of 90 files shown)


crates/swc_bundler/src/bundler/chunk/cjs.rs
```diff
@@ -229,7 +229,7 @@ where
             return;
         }
 
-        if let ModuleItem::ModuleDecl(ModuleDecl::Import(i)) = node {
+        if let ModuleItem::ModuleDecl(box ModuleDecl::Import(i)) = node {
```
dsherret, line 232: Box patterns are going to be deprecated https://github.com/rust-lang/rust/issues/29641#issuecomment-1399430625

kdy1: I know, but refactoring will take much more time than this PR itself.

I filed https://github.com/swc-project/swc/issues/7039 instead

kdy1 (commented):
swc-bump:
 - swc_ecma_ast --breaking

crates/swc_bundler/src/bundler/chunk/cjs.rs
dsherret: Can I help on this one so that you don't go nightly?

kdy1: I think a follow-up PR would be better, because there's too many changes for me to review

kdy1: Anyway, thank you!
I was worried about the box_patterns work

dsherret: It's not very good timing because of the TypeScript 5.0 release upcoming.

dsherret: Going nightly for a perf change doesn't seem worthwhile.

kdy1: I did similar work several times, and always, the problem was that the previous branch should match everything from the if guard. Otherwise, it becomes a logic change.

crates/swc_ecma_ast/src/module.rs
```diff
@@ -76,7 +76,7 @@ impl Take for Module {
 pub struct Script {
     pub span: Span,
 
-    pub body: Vec<Stmt>,
+    pub body: Vec<Box<Stmt>>,
```
dsherret, line 79: I think this shouldn’t be boxed become it’s already in a vec?

crates/swc_ecma_ast/src/module.rs
```diff
@@ -119,11 +119,14 @@ pub enum ModuleItem {
     #[tag("TsImportEqualsDeclaration")]
     #[tag("TsExportAssignment")]
     #[tag("TsNamespaceExportDeclaration")]
-    ModuleDecl(ModuleDecl),
+    ModuleDecl(Box<ModuleDecl>),
     #[tag("*")]
-    Stmt(Stmt),
+    Stmt(Box<Stmt>),
```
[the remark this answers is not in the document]
kdy1: Yeah I profiled it using the minifier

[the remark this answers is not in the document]
kdy1: The performance gain was about 10% (8% ~ 15%), on total time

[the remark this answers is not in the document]
kdy1: I'll run it one more time

crates/swc_ecma_ast/src/module.rs
dsherret: Sorry. I was not thinking and it's late here :)

crates/swc_ecma_ast/src/module.rs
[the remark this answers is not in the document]
kdy1: Hmm... I can't reproduce it.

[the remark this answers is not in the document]
dsherret: If there are gains, I bet most of it is from the change on `BlockStmt` because that will be more prevalent than top level module items (generally, most files will have a low number of module items). It seems most of the nightly feature is used to support the boxing here though.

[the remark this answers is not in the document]
kdy1: Sounds reasonable. I'll wait for the profiling result as I requested one to someone

kdy1:
I profiled this again, and seems like this also regresses performance

kdy1:
There are lots of codes that replace `Vec<Stmt>` with another vector and *moves* elements from the old one to the new one. `memmove` is much more costly than you think.

kdy1:
If you clone it, this patch does not help. It's a new allocation anyway. Tihs PR is about removing needless memory operations, not required operations.

kdy1:
Closing in favor of https://github.com/swc-project/swc/pull/7041

kdy1:
Well, I think I may need this/

magic-akari:
> There are lots of codes that replace `Vec<Stmt>` with another vector and _moves_ elements from the old one to the new one. `memmove` is much more costly than you think.

There are many prepend_stmt operations. Should we allocate a separate field for the directive "use strict"?

Alternatively, is there any other data structure that would facilitate our insertion operations?

kdy1:
`Box<Stmt>` is also important for `prepend_stmt`, but I'm not sure if there's another clean way to optimize `memmove` of `Stmt`

kdy1:
I'll create another PR in the future

PR #6797, opened 2023-01-12
feat(es/quote): Allow using `Str` as a var

kdy1:
**Description:**

It's tested by a doc comment.

PR #7038, opened 2023-03-09
Fix doctests for swc_ecma_quote

andersk:
**Description:**

Fix errors from `cargo test -p swc_ecma_quote --doc` (part of `cargo test`).

I suspect this was broken by

- #5713
- #6797

although I wonder why this hasn’t been noticed—am I the first to try running `cargo test`?


kdy1:
I reverted this as this broke the CI

PR #7043, opened 2023-03-09
perf(es/lexer): Use string searcher for lexing line comments

kdy1:
**Description:**

I'm trying string searcher


kdy1 (commented):
swc-bump:
 - swc_common

Issue #6990, opened 2023-02-25
transform and minify functions return different results


devongovett:
### Describe the bug

When using the `transform` function with `minify` options, SWC correctly optimizes unused pure values away. Using the same code and options with the `minify` function does not.

### Input code

```typescript
const {transformSync, minifySync} = require('@swc/core');

const code = `(() => {
function $parcel$interopDefault(a) {
  return a && a.__esModule ? a.default : a;
}
var $074842700de94141$exports = {};
$074842700de94141$exports = "some unused data";


let $f61716f0e972bb72$var$value = (0, (/*@__PURE__*/$parcel$interopDefault($074842700de94141$exports)));

})();`;

const transformResult = transformSync(code, {
  jsc: {
    minify: {
      mangle: true,
      compress: true
    }
  }
});

// => { code: '' }

const minifyResult = minifySync(code, {
  mangle: true,
  compress: true
});

// => { code: '(()=>{var a={};a="some unused data"})();' }
```

### Expected behavior

The result should be the same between them.

### Version

1.3.36

kdy1:
This cannot be reproduced anymore with the input you gave.
I think _moving_ of `PURE` was the cause.

PR #7051, opened 2023-03-10
fix(es/minifier): Produce same result as `transform()`

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6990.


kdy1:
It's not reproducible anymore

https://github.com/swc-project/swc/issues/6990#issuecomment-1510673271

Issue #7044, opened 2023-03-09
Object spread operator is unnecessarily transformed when used with React


spmonahan:
### Describe the bug

When spreading props onto a React component [object literal spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals) is converted to `Object.assign` despite all target browsers natively supporting object literal spread [according to caniuse](https://caniuse.com/mdn-javascript_operators_spread_spread_in_object_literals).

### Input code

```typescript
const Component = (props) => {
    return <div {...props}/>
}

const obj = { a: 1, b: 2};
// This is transformed to Object.assign
<Component {...obj}/>

// This is not transformed
const clone = { ...obj };
```

### Config

```json
{
  "$schema": "https://json.schemastore.org/swcrc",
  "env": {
    "targets": {
      "chrome": "84",
      "edge": "84",
      "firefox": "75",
      "opera": "73",
      "safari": "14.1"
    },
    "bugfixes": true
  },
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true,
      "decorators": false,
      "dynamicImport": false
    },
    "externalHelpers": true,
    "transform": {
      "react": {
        "useBuiltins": true
      }
    },
    "target": "es2019"
  },
  "minify": false,
  "sourceMaps": true
}
```

### Playground link

[...]

PR #7054, opened 2023-03-10
fix: Object spread operator is unnecessarily transformed when used with React

suxin2017:
…th React

**Description:**
fix Object spread operator is unnecessarily transformed when used with React

**BREAKING CHANGE:**

https://github.com/swc-project/swc/issues/7044

### DIFF
--- a/crates/swc/tests/vercel/loader-only/next-33291/2-classic/output/index.js
+++ b/crates/swc/tests/vercel/loader-only/next-33291/2-classic/output/index.js
@@ -1,3 +1,5 @@
+import _object_spread from "@swc/helpers/src/_object_spread.mjs";
+import _object_spread_props from "@swc/helpers/src/_object_spread_props.mjs";
 import _object_without_properties from "@swc/helpers/src/_object_without_properties.mjs";
 import Head from "next/head";
 import Image from "next/image";
@@ -50,9 +52,9 @@ export default function Home() {
         var href = _param.href, linkProps = _object_without_properties(_param, [
             "href"
         ]);
-        return React.createElement("link", Object.assign({
+        return React.createElement("link", _object_spread_props(_object_spread({
             href: href
-        }, linkProps, {
+        }, linkProps), {
             rel: "icon",
             key: href,
             __source: {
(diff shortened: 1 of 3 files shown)


kdy1 (approved):
Thanks!
You can do `UPDATE=1 cargo test` from `./crates/swc` to update test references.

---


swc-bump:
 - swc_ecma_transforms_react

suxin2017:
> 谢谢！您可以从中更新测试引用。`UPDATE=1 cargo test``./crates/swc`
> 
> SWC凸起：
> 
> * swc_ecma_transforms_react

done

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -872,13 +870,7 @@ where
...
             Box::new(Expr::Call(CallExpr {
                 span: DUMMY_SP,
-                callee: {
-                    if self.use_builtins {
-                        member_expr!(DUMMY_SP, Object.assign).as_callee()
-                    } else {
-                        helper!(extends, "extends")
-                    }
-                },
+                callee: helper!(extends, "extends"),
```
kdy1, line 873: I think this should be reverted

crates/swc/tests/vercel/loader-only/next-33291/2-classic/output/index.js
```diff
@@ -50,9 +52,9 @@ export default function Home() {
         var href = _param.href, linkProps = _object_without_properties(_param, [
             "href"
         ]);
-        return React.createElement("link", Object.assign({
+        return React.createElement("link", _object_spread_props(_object_spread({
```
kdy1, line 55: This looks like a regression.

kdy1 (changes requested):
Thanks!

crates/swc_ecma_transforms_react/src/jsx/mod.rs
suxin2017: I don't quite understand the difference between configuring use_buildin and use_spread. It looks like the public is the same, only one is {... } One is Object.assign

suxin2017: Maybe we can support issue without fixing it, just configure use_spread

PR #7065, opened 2023-03-12
fix(es/config): respect the value of assumptions

magic-akari:
**Description:**

**BREAKING CHANGE:**


magic-akari:
The test files have been updated as they have enabled the 'loose' mode.\
With the 'loose' mode, the 'set_public_class_fields' default value is set to true.

https://github.com/swc-project/swc/blob/449e2f0f1a3130247ee7c6b26d1fe460d4b6277f/crates/swc_ecma_transforms_base/src/assumptions.rs#L103-L132

kdy1 (approved):
swc-bump:
 - swc

PR #7067, opened 2023-03-12
fix(es/compat): Insert the variable declaration nearest to the available statements

### LINKED ISSUE #7064: ??= operator in a method retains the lvalue from being GCed in the module global variable
### Describe the bug

When compiling TS to es2020 target, ??= operator "temporarily" saves the target object in a global module variable. The problem is that it is not "temporary": it retains this object in memory forever, it will never be garbage collected.

I stepped on it accidentally, while learning some anomalies in Node memory dump using Chrome memory snapshot profiler.

### Input code

```typescript
function method() {
  const obj = {};
  obj.prop ??= "x".repeat(10000000);
}

method();
// the large string should be eventually GCed along with the large string, but it will never happen in SWC+es2020
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "decorators": true
    },
    "target": "es2020"
  },
  "module": {
    "type": "commonjs"
  }
}
```

### Playground link

https://play.swc.rs/?version=1.3.39&code=H4sIAAAAAAAAA2WMvQ6CMBSF9z7FiRNEA8hKCAODD%2BDgXOBKa2rbtBd%2FQnh3mxgnz3RyvnznutiRtbO4Eys3ZTlWAYzORoYbbmixbk1aUi98cB5d12L32hWBPEnOjtU3eSM2IX4njShLsCIYGWZC5KDtjKjcYiYMBHqQ5UUa88appwnSuMSfmtWfdcCwMDQnagxsMgOU9J4stMX50u8p1lVdfQAVjZi5yQAAAA%3D%3D&config=H4sIAAAAAAAAAxWMQQrAIAwE%2F7JnD8WjvxENpaU2kkSoiH%2Bv3padnR24NSEM1ChKspP21%2BKHAOuVNMlVDQ6ZEks0FkUwaTQdLMpJtoak%2FvAHVlU4t4f2y5YXSlwKv7dizh8GxEdwawAAAA%3D%3D

### Expected behavior

The compiled code should have looked like:

```
"use strict";
function method() {
    var _obj;
    const obj = {};
    (_obj = obj).prop ?? (_obj.prop = "x".repeat(10000000));
}
method();
```

I.e. the temporary _obj should've been created WITHIN the function, not in the global scope.

### Actual behavior

Notice that _obj is created at top level. Which means that the large string allocated (the one created during the very last call to method()) will never be garbage collected, even after `out` passed to method() loses all other references.

```
"use strict";
var _obj; // <-- WRONG
function method() {
    const obj = {};
    (_obj = obj).prop ?? (_obj.prop = "x".repeat(10000000));
}
method();
```

### Version

1.3.39

### Additional context

Obviously, ES2021 target fixes this, because ??= operator is implemented natively there.

It applies not only to functions, but to class methods too (and I suspect, to closures as well): https://play.swc.rs/?version=1.3.39&code=H4sIAAAAAAAAAy2MTQ7CIBSE95xi0hXdEN22aVh4AM9A8VnbVCBAoknD3X1gZzd%2F3%2BoyxaexhPu84RBAiD7oASnH1S2jKELY3aSE255a%2F6b88g%2Fp522on76FAHtVr9B6QvftVKRAJsvr5a9%2B5FlpOO9ShmXcBEefCpbccqBO9lHY%2FwBM7y%2B5mgAAAA%3D%3D&config=H4sIAAAAAAAAAxWMQQrAIAwE%2F7JnD8WjvxENpaU2kkSoiH%2Bv3padnR24NSEM1ChKspP21%2BKHAOuVNMlVDQ6ZEks0FkUwaTQdLMpJtoak%2FvAHVlU4t4f2y5YXSlwKv7dizh8GxEdwawAAAA%3D%3D

magic-akari:
**Description:**

**BREAKING CHANGE:**

- fix #7064

### DIFF
--- a/crates/swc/tests/fixture/issues-7xxx/7064/case1/input/.swcrc
+++ b/crates/swc/tests/fixture/issues-7xxx/7064/case1/input/.swcrc
@@ -1,0 +1,12 @@
+{
+    "jsc": {
+        "parser": {
+            "syntax": "typescript",
+            "decorators": true
+        },
+        "target": "es2020"
+    },
+    "module": {
+        "type": "commonjs"
+    }
+}
--- a/crates/swc/tests/fixture/issues-7xxx/7064/case1/input/index.js
+++ b/crates/swc/tests/fixture/issues-7xxx/7064/case1/input/index.js
@@ -1,0 +1,7 @@
+function method() {
+    const obj = {};
+    obj.prop ??= "x".repeat(10000000);
+  }
+  
+  method();
+  // the large string should be eventually GCed along with the large string, but it will never happen in SWC+es2020
--- a/crates/swc/tests/fixture/issues-7xxx/7064/case1/output/index.js
+++ b/crates/swc/tests/fixture/issues-7xxx/7064/case1/output/index.js
@@ -1,0 +1,7 @@
+"use strict";
+function method() {
+    var _obj;
+    const obj = {};
+    (_obj = obj).prop ?? (_obj.prop = "x".repeat(10000000));
+}
+method(); // the large string should be eventually GCed along with the large string, but it will never happen in SWC+es2020
--- a/crates/swc_ecma_transforms_compat/src/es2021/logical_assignments.rs
+++ b/crates/swc_ecma_transforms_compat/src/es2021/logical_assignments.rs
@@ -174,12 +174,12 @@ impl VisitMut for Operators {
         }
     }
 
-    fn visit_mut_module(&mut self, n: &mut Module) {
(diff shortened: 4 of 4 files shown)


magic-akari:
Furthermore, I believe we should utilize the pure getter of `assumption` to generate more concise code.
The `alias_if_required` function can be adapted to accept this parameter.
What are your thoughts on this? @kdy1

kdy1:
I think it's a good idea

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_compat

Issue #7045, opened 2023-03-09
SWC erases `delete` of global variable


andersk:
### Describe the bug

SWC incorrectly erases a statement that `delete`s a global variable, changing the meaning of the code.

### Input code

```typescript
a = 1;
delete a;
console.log(typeof a);
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript"
    },
    "minify": {
      "compress": true
    },
    "target": "es2015"
  },
  "isModule": false
}
```

### Playground link

https://play.swc.rs/?version=1.3.38&code=H4sIAAAAAAAAA0tUsFUwtOZKSc1JLUlVSLTmSs7PK87PSdXLyU%2FXKKksSM1PU0jUtAYADe1dLScAAAA%3D&config=H4sIAAAAAAAAAxWLQQqAMAwE%2F7JnDyp46R98RKhRKlZLNoIi%2Fl17G5iZBysjwoMiRrVKvHeXCwEaszBaKo63QU57mu8axCMXUxLB7dRfudiiXg%2F2bTfUOnE8pnNThFk26vsBSkMid2gAAAA%3D

### Expected behavior

SWC should keep the `delete` statement, since it has an observable side effect.

### Actual behavior

SWC erases the `delete` statement and outputs `console.log(typeof (a = 1));`. This output program incorrectly logs `number`, while the input program logged `undefined`.

### Version

1.3.38

suxin2017:
![play swc rs_](https://user-images.githubusercontent.com/28481035/224193109-d831b736-f698-482b-b75a-d77537273b33.png)
The compilation error is displayed in AST mode

andersk:
@suxin2017 I think that’s just an issue with the playground. This file only parses with `isModule: false`, which *is* in my config as you can see—but perhaps the playground fails to pass that option when AST mode is selected?

PR #7072, opened 2023-03-13
fix(es/minifier): Preserve `delete` of unresolved variables

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/7045.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

andersk:
Should we mark a `delete` expression as having side effects? Currently we only mark it as having side effects if its argument does, which is wrong:

https://github.com/swc-project/swc/blob/769ebaf7c3f0453d34024eddfa0080fd211b711d/crates/swc_ecma_utils/src/lib.rs#L1350

Fixing that might make this change unnecessary (not tested though).

jridgewell (approved):

PR #7073, opened 2023-03-13
perf(es/lexer): Use jump table for `skip_space`

kdy1:
**Description:**

I'll compare performance with simd


kdy1:
This makes parser about 3% faster

kdy1 (commented):
swc-bump:
 - swc_ecma_parser

PR #7077, opened 2023-03-13
fix(cli): Windows stdin handle waits in an infinite loop

realtimetodie:
Prevents the stdin handle from waiting in an infinite loop for the read stream to finish, when running in a process with a detached console on Windows. 

This is a [known issue](https://doc.rust-lang.org/std/io/fn.stdin.html#note-windows-portability-considerations) on Windows. To prevent this from happening, this change simply flips the order how inputs are parsed. Now, the files option is used first before using the stdin input as a fallback.

As a workaround in Bazel rules_swc, we forward the output of the null device to force the read stream to immediately finish on Windows. However, this workaround appears to cause a difference in argv quoting.

https://github.com/aspect-build/rules_swc/blob/e149dd5fc9e512d5b87b3d52a5aa8422c1a1a309/swc/private/swc.bzl#L194

Related 

- https://github.com/aspect-build/rules_swc/pull/187

### DIFF
--- a/bindings/swc_cli/src/commands/compile.rs
+++ b/bindings/swc_cli/src/commands/compile.rs
@@ -347,34 +347,7 @@ impl CompileOptions {
     fn collect_inputs(&self) -> anyhow::Result<Vec<InputContext>> {
         let compiler = COMPILER.clone();
 
-        let stdin_input = collect_stdin_input();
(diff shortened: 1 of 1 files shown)


kdy1 (approved):
swc-bump:
 - dbg-swc

PR #7081, opened 2023-03-15
fix(es/react): align to `Babel@8` behavior

### LINKED ISSUE #7044: Object spread operator is unnecessarily transformed when used with React

magic-akari:
**Description:**

**BREAKING CHANGE:**

- fix #7044


crates/swc/tests/tsc-references/correctlyMarkAliasAsReferences4.1.normal.js
```diff
@@ -2,11 +2,12 @@
...
 //// [0.tsx]
 ///<reference path="declaration.d.ts" />
-import _extends from "@swc/helpers/src/_extends.mjs";
+import _object_spread from "@swc/helpers/src/_object_spread.mjs";
+import _object_spread_props from "@swc/helpers/src/_object_spread_props.mjs";
 import * as cx from 'classnames';
 import * as React from "react";
 let buttonProps;
-let k = /*#__PURE__*/ React.createElement("button", _extends({}, buttonProps, {
+let k = /*#__PURE__*/ React.createElement("button", _object_spread_props(_object_spread({}, buttonProps), {
```
magic-akari, line 10: https://babeljs.io/docs/assumptions#setspreadproperties

It seems to be a regression.
Does this imply that it is always safe to  use `_extends` helper to transpile JSX spread?

magic-akari: > ## setSpreadProperties
> When using object spread, assume that spreaded properties don't trigger getters on the target object and thus it's safe to assign them rather than defining them using Object.defineProperty.

magic-akari: However, you can indeed write it as following.
I hope no one writes something like this.

```jsx
const obj = {
    get className() {
        console.log("get className");
        return "text";
    },
};

<div {...obj} />;
```

kdy1: I think we can assume that `_extends` is always safe. It looks like an implicit, undocumented assumption.

magic-akari: I have comprehended its usage. I shall proceed to optimize it.

magic-akari: https://github.com/babel/babel/blob/edcd681d915c77bf25a32a2aaec0860e75888d5a/packages/babel-helper-builder-react-jsx/src/index.ts#L201-L205

https://github.com/babel/babel/blob/edcd681d915c77bf25a32a2aaec0860e75888d5a/packages/babel-helper-builder-react-jsx/src/index.ts#LL132-L134

With `Babel@8`, JSXSpreadAttribute will be transpiled into SpreadElement. The downlevel task is handled by other module.
We should follow it.

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_transforms_react

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -74,10 +71,20 @@ pub struct Options {
     #[serde(default)]
     pub development: Option<bool>,
 
-    /// TODO: Remove this field.
+    // @babel/plugin-transform-react-jsx: Since "useBuiltIns" is removed in Babel 8, you can remove
+    // it from the config.
+    #[deprecated(
+        note = r#"This is always true. Use swc_ecma_transforms_compat to downlevel JavaScript syntax."#
```
magic-akari, line 77: Wait a minute, I should update the comments/

magic-akari: @kdy1 done.

kdy1: Cool, thanks!

kdy1 (approved):

PR #7082, opened 2023-03-15
fix(css/parser): Fix parsing of `:global(> *)`

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_css_parser

crates/swc_css_modules/tests/with-compat/vercel/010/output.css
```diff
@@ -0,0 +1,3 @@
+.__local__quickstart .__local__buttons  > * {
```
alexander-akait, line 1: There is extra space and looks like we can cut them more, but it is not critical, I will refactor it so we will fix it

alexander-akait (approved):

crates/swc_css_modules/tests/with-compat/vercel/010/output.css
kdy1: Cool, thank you!

PR #7084, opened 2023-03-15
add anther case in ExportNamedDeclaration comment

stormslowly:
when there is no source

**Description:**

add code example in ExportNamedDeclaration remind there is a case like below
```
export  { foo }
```


kdy1:

stormslowly:
that's OK
the Option<> and example maked me confused for a little while.

PR #7086, opened 2023-03-15
fix(es/compat): prefer `extends` for object spread

magic-akari:
**Description:**

**BREAKING CHANGE:**

- #7081

### DIFF
--- a/crates/swc/tests/tsc-references/destructuringSpread.1.normal.js
+++ b/crates/swc/tests/tsc-references/destructuringSpread.1.normal.js
@@ -1,19 +1,19 @@
 //// [destructuringSpread.ts]
-import _object_spread from "@swc/helpers/src/_object_spread.mjs";
+import _extends from "@swc/helpers/src/_extends.mjs";
 import _object_spread_props from "@swc/helpers/src/_object_spread_props.mjs";
-var x = _object_spread_props(_object_spread({}, {}), {
+var x = _object_spread_props(_extends({}, {}), {
     x: 0
 }).x;
-var y = _object_spread({
+var y = _extends({
     y: 0
 }, {}).y;
-var _$_object_spread = _object_spread({
+var _$_extends = _extends({
     z: 0
 }, {
     a: 0,
     b: 0
-}), z = _$_object_spread.z, a = _$_object_spread.a, b = _$_object_spread.b;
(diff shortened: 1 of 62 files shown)


magic-akari:
https://github.com/swc-project/swc/blob/12fea00b1676435e887ff94133a4a1a82a93da60/crates/swc_ecma_transforms_compat/tests/es2018_object_rest_spread.rs#L2344-L2383

I run into this issue.
When transpiling the object spread syntax(`...`) to `Object.assign` (or `_extends` polyfill), the assigned target is an object.
Thus, I believe that in most cases, this operation is safe and reliable.

However, this particular test case presented a modification to the `object.prototype`, thereby breaking my assumption.

Adhering to the spec, we should never use `Object.assign` as we cannot be sure whether the `Object.prototype` has been modified by someone. (Unless passed `assumptions` or `loose`).

@kdy1  What do you think?

kdy1:
The test is copied from https://github.com/babel/babel/blob/d38f415c9f7704a201cbfb1d73e6884738931bb2/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-spread/no-object-assign-exec/exec.js , and if babel transpiles correctly, I think we should do either **by default** and we should rely on `assumption` for optimization.

kdy1:
Honestly, I don't expect someone to do it so I think it's almost safe, though

magic-akari:
The `pristine_globals` option is present in the `compress` phase, which is not accessible to us during the `compat` phase. Therefore, I believe I should stop here and comply with the spec behavior.

kdy1:
I agree. Thank you for your hard work anyway!

magic-akari:
Note: Babel will use `_objectSpread` in React `Atomatic` mode.

https://babeljs.io/repl#?browsers=chrome%2059&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=DwEwlgbgBAhgvAbwIwF8oIHRYPYCMBWaA9AHxA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact&prettier=false&targets=&version=7.21.3&externalPlugins=&assumptions=%7B%7D

PR #7095, opened 2023-03-18
feat(es/base): Make scope context same to the declared variable

Austaras:
**Description:**
Mark scope with the same `Mark` as the declared variables inside them.

This PR may cause performance regression in non compress use cases, but the effect should be negligible.

**BREAKING CHANGE:**


Austaras:
And, as a side note, the whole `Mark` thing is not necessary for SWC. `rustc` uses `Mark` to mark which macro code come from, and in SWC `Mark` is used in two cases: variable shadowing as showed in this PR, and provide some additional information as in mark pure in compress. If the latter usage can be eliminated, we could use a single `AtomicUSize` to provide for all `SyntaxContext`.

crates/swc_ecma_transforms_base/src/resolver/mod.rs
```diff
@@ -586,18 +594,11 @@ impl<'a> VisitMut for Resolver<'a> {
...
     }
 
-    /// Handle body of the arrow functions
-    fn visit_mut_block_stmt_or_expr(&mut self, node: &mut BlockStmtOrExpr) {
-        match node {
-            BlockStmtOrExpr::BlockStmt(block) => block.visit_mut_children_with(self),
-            BlockStmtOrExpr::Expr(e) => e.visit_mut_with(self),
-        }
-    }
-
```
Austaras, line 600: `visit_mut_arrow_expr` never calls this.

kdy1 (approved):
swc-bump:
 - swc_ecma_transforms_base

kdy1:
I'm unsure if it's fine to simplify' SyntaxContext' to single atomic usize at the moment.

Austaras:
No worry, even if it is, I currently don't have the capacity to apply such huge change.

PR #7096, opened 2023-03-19
mod(es/base): Stable context for unresolved and global

Austaras:
**Description:**

Now unresolved context and global context will always be 1 and 2 respectively in single transform. I believe it's better for debug.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_transforms_base

PR #7097, opened 2023-03-19
mod(es/base): Move comments following the order of insertion in `fixer`

Austaras:
**Description:**

Or when encountering nested paren expr with comments, comments would be lost. Found this problem when testing #7096 

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_transforms_base/src/fixer.rs
+++ b/crates/swc_ecma_transforms_base/src/fixer.rs
@@ -1,4 +1,7 @@
-use rustc_hash::FxHashMap;
+use std::{hash::BuildHasherDefault, ops::RangeFull};
+
+use indexmap::IndexMap;
+use rustc_hash::FxHasher;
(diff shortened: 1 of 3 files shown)


crates/swc_ecma_transforms_base/src/fixer.rs
```diff
@@ -37,7 +39,7 @@ struct Fixer<'a> {
     ///
     /// Key is span of inner expression, and value is span of the paren
     /// expression.
-    span_map: FxHashMap<Span, Span>,
+    span_map: IndexMap<Span, Span>,
```
kdy1, line 42: Please change this to `IndexMap<Span, Span, DefaultBuildHasher<FxBuildHasher>>`

kdy1 (changes requested):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_base

kdy1 (approved):
Thank you!

Issue #3345, opened 2022-01-24
Support spack.config.cjs


creepinson:
### Describe the bug

I put  `"type": "module"` in my package.json, and tried to use spack but it gives the first error message.

If I specify the `--config` option with spack it gives the second error message.

### Error messages

1st error:

```
spack.config.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename spack.config.js to end in .cjs, 
change the requiring code to use dynamic import() which is available in all CommonJS modules, 
or change "type": "module" to "type": "commonjs" in
 /mnt/data/projects/elemental-loader/pkg/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

2nd error:
```
/mnt/data/projects/elemental-loader/pkg/node_modules/.pnpm/@swc+core@1.2.133/node_modules/@swc/core/spack.js:55
                throw new Error(`Error occurred while loading config file at ${config}: ${e}`);
                      ^

Error: Error occurred while loading config file at spack.config.cjs: Error: Cannot find module 'spack.config.cjs'
Require stack:
- /mnt/data/projects/elemental-loader/pkg/node_modules/.pnpm/@swc+core@1.2.133/node_modules/@swc/core/spack.js
```

### Input code

```typescript
Multiple typescript files
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "dynamicImport": true,
      "functionBind": false,
      "decorators": true,
      "topLevelAwait": true,
      "importMeta": true
    },
    "target": "es5",
    "loose": false
  },
  "module": {
    "type": "es6",
    "strict": false,
    "strictMode": true,
    "lazy": false,
    "noInterop": false
  }
}
```

### Expected behavior

The configuration file to be parsed and a bundled js file to be produced.

### Actual behavior

Gives errors about spack config file having the **.cjs** extension even with the `--config` flag.

### Version

1.2.133

kwonoj:
This is due to we don't have full native esm support yet. https://github.com/swc-project/swc/discussions/3097#discussioncomment-1855849

russelldavis:
The second error is because swc requires an absolute path when specifying a file with `--config`. As a workaround, you can do something like `spack foo.ts --config "$PWD/spack.config.cjs"`

PR #7105, opened 2023-03-20
fix(bindings/node): Fix loading of `spack.config.js`

3axap4eHko:
**Description:**
`spack` config file provided from CLI may not be properly imported with `require()`. Without a leading '/', './', or '../' to indicate a file, the module must either be a core module or be loaded from a node_modules folder, but spack.config.js is a local file.

**BREAKING CHANGE:**
N/A

ISSUE #3345


kdy1 (approved):
swc-bump:
 - dbg-swc

PR #7109, opened 2023-03-21
feat(css/parser): Add `legacy_ie` to the `ParserConfig`

kdy1:
**Description:**

**Related issue:**

 - https://github.com/vercel/turbo/issues/4092


kdy1 (commented):
swc-bump:
 - swc_css_parser --breaking

crates/swc_css_parser/tests/fixture/turbo/4092/input.css
```diff
@@ -0,0 +1,4 @@
+.carousel .control-dots .dot.selected, .carousel .control-dots .dot:hover {
+     opacity: 1;
+     filter: alpha(opacity=100);
```
alexander-akait, line 3: That is also deprecated syntax, I think we should move it under the option like https://github.com/swc-project/swc/pull/7110

alexander-akait: But I am fine with it because we have `PreservedToken`, we can handle all non standard cases late and move it under the option

kdy1: Can you recommend a name for the option?
Maybe `legacy_ie`?

PR #7120, opened 2023-03-22
fix(es/visit): Fix handling of `Program` of `AndThen`

JSerFeng:
**Description:**

Visitor used in AndThen or Repeat cannot invoke `fold_program`

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_visit/src/lib.rs
+++ b/crates/swc_ecma_visit/src/lib.rs
@@ -21,6 +21,12 @@ where
     B: Fold,
 {
     #[inline(always)]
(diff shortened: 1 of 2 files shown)


kdy1:
The default impl is enough

JSerFeng:
> The default impl is enough

But imagine I have a pass rely on `fold_program`, others uses my pass as `chain!(pass, pass2)`, then my pass is not working correctly

kdy1:
Can you add a test case for such case?

JSerFeng:
> Can you add a test case for such case?

Done

kdy1 (approved):

Issue #7119, opened 2023-03-22
Should require jsx runtime in Script


JSerFeng:
### Describe the bug

Missing require for `jsx runtime` if source type is script

### Input code

```typescript
<div></div>
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true,
      "decorators": true
    },
    "transform": {
      "react": {
        "runtime": "automatic"
      }
    },
    "externalHelpers": true,
    "target": "es2022"
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": false
}
```

### Playground link

https://play.swc.rs/?version=1.3.41&code=H4sIAAAAAAAAA7NJySyzs9EHkQD5kTeICwAAAA%3D%3D&config=H4sIAAAAAAAAA1WPPQ6DMAxGd06BPHeoGHuCLj2EFUwVlD%2FZRgIh7t4kBUq3%2BD3n%2B%2BS1aVsYxcCjXfMzDwlZiM85E1mC4pwJ6JJIDNukcDusSlHKE52oJxMZNbLsport60EZgwyR%2FbWDCY1eQEFTUOup9OKk0aNaA7ve%2FhJpVuKA7kku0Vl61CG%2FqUQDSXfvuhJRP4KP%2FeTod3m5ruyZ6H0Mo1w2bbDDkt2ATmowWHkd3ytstg9IVTefSwEAAA%3D%3D

### Expected behavior

const _jsxRuntime = require("react/jsx-runtime");
_jsxRuntime.jsx("div", {});

### Actual behavior

/*#__PURE__*/ _jsx("div", {});

### Version

1.3.41

JSerFeng:
Found the reason, you can leave this to me

kdy1:
Thanks!

PR #7126, opened 2023-03-22
feat(es/react): Support script in automatic runtime

JSerFeng:
**Description:**

Inject `const { jsx: _jsx} = require("react/jsx-runtime")` when sourceType is `script`

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/7119.


crates/swc_ecma_transforms_react/src/jsx/tests.rs
```diff
@@ -1483,3 +1489,74 @@ fn integration(input: PathBuf) {
         },
     );
 }
+
+#[testing::fixture("tests/script/**/input.js")]
+fn script(input: PathBuf) {
+    let mut output = input.with_file_name("output.js");
+    if !output.exists() {
+        output = input.with_file_name("output.mjs");
```
kdy1, line 1497: I think this approach is too error-prone

JSerFeng: Do you mean that the inject approach is too error-prone

kdy1: This line. Determining test mode based on thr existence of output

JSerFeng: I deleted .mjs output test for script test

JSerFeng: After lunch, I'll fix conflict

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -975,113 +1033,115 @@ where
...
+                    })
+                    .collect(),
+                optional: false,
+                type_ann: None,
+            }),
+            // require('react')
+            init: Some(Box::new(Expr::Call(CallExpr {
+                span: DUMMY_SP,
+                callee: Callee::Expr(Box::new(Expr::Ident(Ident {
+                    span: DUMMY_SP,
```
kdy1, line 1126: This should have `SyntaxContext` with `unresolve_mark` as the only `Mark`

kdy1 (changes requested):

crates/swc_ecma_transforms_react/src/jsx/mod.rs
JSerFeng: Thanks, done

kdy1 (approved):
Thanks! I'll merge this along with other breaking changes.

---

swc-bump:
 - swc_ecma_transforms_react --breaking

PR #7127, opened 2023-03-23
refactor(es/codegen): Extract code from generic functions to reduce compile time

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_ecma_ast

Boshen:
Nice work, thank you for working on this. Removing generics will definitely help out.

It would be great if you post your `--timings` so we can keep a record of improvements from these.

PR #7131, opened 2023-03-23
fix(css/codegen): Preserve `raw` of numbers

kdy1:
**Description:**

x-ref: https://vercel.slack.com/archives/C03EWR7LGEN/p1679513958649249


kdy1 (commented):
swc-bump:
 - swc_css_codegen

PR #7133, opened 2023-03-23
fix(fast-graph): Use fxhash instead of ahash to make iteration order consistent

kdy1:
**Description:**

Some operations of `petgraph` assumes the same iteration order.


kdy1 (commented):
swc-bump:
 - swc_fast_graph

Issue #6630, opened 2022-12-12
swc_css_parser 0.138 and higher cannot be compiled on the stable channel


nicoburns:
### Describe the bug

Trying to compile `swc_parser_css` on latest stable fails for version >= `0.138.0`:

```
   Compiling swc_css_parser v0.138.7
error[E0554]: `#![feature]` may not be used on the stable release channel
 --> /Users/name/.cargo/registry/src/github.com-1ecc6299db9ec823/swc_css_parser-0.138.7/src/lib.rs:1:1
  |
1 | #![feature(box_patterns)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0554`.
error: could not compile `swc_css_parser` due to previous error
```

### Expected behavior

Code compiles on stable Rust.

### Actual behavior

Code does not compile

### Version

0.138, 0.139, 0.140

### Additional context

Previous Github issues lead me to believe you are targeting the the stable compiler (if you're not then you can ignore this). I believe this issue was introduced by this PR https://github.com/swc-project/swc/pull/6593

alexander-akait:
Because `swc_css` and `swc_html` are still experimental we decided to use new feature to reduce code and simplify logic, `swc_ecma` is only `stable`

kdy1:
MSRV is nightly

PR #7136, opened 2023-03-23
Make `swc_css_parser` compile on stable rust

nicoburns:
**Description:**

This PR removes usage of `#![feature(box_patterns)]` from `swc_css_parser`. It also removes `#![cfg_attr(docsrs, feature(doc_cfg))]`, but as far as I can see this feature was not actually being used in this crate. This is of course subjective, but IMO these changes actually improve the readability of the affected code. It certainly makes it shorter in most cases.

This allows `swc_css_parser` to be compiled using a stable rust compiler.

https://github.com/swc-project/swc/issues/6630

---

This PR only applies changes to the `swc_css_parser` crate as this the only crate (which does not already compile on stable) that I am personally interested in using, but I would be willing to go through the other swc crates and make similar changes to the other crates if you would be open to accepting such changes. It looks like swc's usage of unstable features are all pretty trivial and could easily be replaced.

### DIFF
--- a/crates/swc_css_parser/src/lib.rs
+++ b/crates/swc_css_parser/src/lib.rs
@@ -1,5 +1,4 @@
 #![feature(box_patterns)]
-#![cfg_attr(docsrs, feature(doc_cfg))]
 #![deny(unused_must_use)]
 #![deny(clippy::all)]
 #![allow(clippy::needless_return)]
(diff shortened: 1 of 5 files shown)


kdy1:
I don't think this improves readability, and is there a reason you can't use nightly rust?
Honestly I don't want to accept this PR

nicoburns:
> Is there a reason you can't use nightly rust?

Yes, I'm writing a widely used library that is depended on by multiple projects (such as Bevy) which use stable Rust.

> Honestly I don't want to accept this PR

Fair enough. It's your project. It was worth a try!

kdy1:
I'll not accept this PR because this is about adding a huge API

PR #7138, opened 2023-03-24
refactor(ast): Make serde optional

kdy1:
**Description:**

Turbopack does not use `serde` of swc AST types.


kdy1 (commented):
swc-bump:
 - swc_common
 - string_enum --breaking
 - ast_macro --breaking

PR #7141, opened 2023-03-24
fix: Resolve the error of missing map file.

limerickgds:
**Description:** When there is no sourcemap file, an error occurs because the file path with .map added automatically cannot be found.

**BREAKING CHANGE:** no


limerickgds:
@kdy1 Please help review this.

kdy1 (changes requested):
Can you add a test for this?

limerickgds:
> Can you add a test for this?

@kdy1 done

crates/swc/src/lib.rs
```diff
@@ -324,43 +324,65 @@ impl Compiler {
...
-                                            "failed to read input source map from file at {}",
-                                            path
-                                        )
-                                    },
-                                )?))
+                                match map_path {
+                                    Some(map_path) => {
+                                        let path = map_path.display().to_string();
+                                        let file = File::open(&path)
+                                            .context("failed to open input source map file");
```
kdy1, line 360: This `context` is redundant

crates/swc/src/lib.rs
```diff
@@ -324,43 +324,65 @@ impl Compiler {
...
-                                    None => {
-                                        format!("{}.map", dir.join(filename).display())
+                                // Old behavior. This check would prevent
+                                // regressions.
+                                // Perhaps it shouldn't be supported. Sometimes
+                                // developers don't want to expose their source code.
+                                // Map files are for internal troubleshooting
+                                // convenience.
+                                let fallback_map_path =
+                                    PathBuf::from(format!("{}.map", filename.display()));
```
kdy1, line 334: Can we be lazy here? This allocation is not necessary.

kdy1 (changes requested):

kdy1 (approved):
swc-bump:
 - swc

PR #7147, opened 2023-03-26
fix: @swc/helpers

### LINKED ISSUE #7118: `ts-decorate` and `ts-metadata` helpers not found in Script source type
### Describe the bug

source type: module

```ts
...
var _tsDecorate = require("@swc/helpers/lib/_ts_decorate.js").default;
```
source type: script

```ts
...
var _ts_decorate = require("@swc/helpers").ts_decorate;
```

require("@swc/helpers").ts_decorate is `undefined`

### Input code

```typescript
@aaa
class a {

}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true,
      "decorators": true
    },
    "externalHelpers": true,
    "target": "es2022"
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": false
}
```

### Playground link

https://play.swc.rs/?version=1.3.41&code=H4sIAAAAAAAAA3NITEzkSs5JLC5WSFSo5uKqBQAgLzirEQAAAA%3D%3D&config=H4sIAAAAAAAAA02OMQ6DMAxFd06BPHeoMvYEXXoIK5gqKCSRbSQixN2b0EC7%2Bf%2F3v%2B2t63uYxMKj38pYREIW4ksXR3JQXIsDmhOJZZcUbidVqUh5ocsayEZGjSyNHGD%2FcqBViQP6J%2FlEV6RBRX6T1lsk5m4MdK0IcxwWT78%2F6y81Z%2BM8xzDJX9IFN%2BbCRvRyLAYnr7N%2BmN3%2BAfGuH6b5AAAA

### Expected behavior

ts_decorate is defined

### Actual behavior

ts_decorate is undefined

### Version

1.3.41

### Additional context

_No response_

### LINKED ISSUE #7144: Emit deep imports for helpers when using Common JS
### Describe the feature

Currently when using external helpers SWC will output something like this for ESM:
```js
import _create_class from "@swc/helpers/src/_create_class.mjs";
```

And something like this for CJS:
```js
var _create_class = require("@swc/helpers").create_class;
```

I discovered this when analyzing this issue: https://github.com/webpack/webpack/issues/16861

While I do believe that issue to be a webpack issue and not a SWC issue it would most likely not have happened if SWC emitted deep require statements instead of requireing from the entry point which exports all helpers. Eg, if the require statement would have looked like this:
```js
var _create_class = require("@swc/helpers/lib/_create_class.js").default;
```

This would also have the added benefit of getting smaller bundles when using CJS since you'd only import the helpers actually used. To my knowledge webpack won't be able to correctly tree shake unused requires from lib/index.js.

_No response_

### Additional context

_No response_

magic-akari:
- use snake_case
- ~use named export~

**Description:**

generated by Babel plugin

```JavaScript
const node_path = require("node:path");
const { declare } = require("@babel/helper-plugin-utils");

module.exports = declare((api, opt) => {
    return {
        visitor: {
            ExportDefaultDeclaration(path, state) {
                const { filename } = state;
                const newName = node_path.basename(filename, ".mjs");
                if (path.node.declaration.type !== "FunctionDeclaration") {
                    return;
                }
                const oldName = path.node.declaration.id.name;

                path.replaceWith(api.types.exportNamedDeclaration(path.node.declaration));
                path.node.declaration.id.name = newName;

                path.scope.rename(oldName, newName);
            },

            ImportDefaultSpecifier(path) {
                const source = path.parentPath.node.source.value;
                const newName = node_path.basename(source, ".mjs");
                const oldName = path.node.local.name;
                path.scope.rename(oldName, newName);
                path.replaceWith(api.types.importSpecifier(path.node.local, path.node.local));
            }
        }
    };
});
```

I will doublecheck the result.

**BREAKING CHANGE:**

- fix #7144 
- fix #7118


packages/swc-helpers/src/_extends.mjs
```diff
@@ -1,21 +1,14 @@
-function extends_() {
-  extends_ = Object.assign || function (target) {
-    for (var i = 1; i < arguments.length; i++) {
-      var source = arguments[i];
-
-      for (var key in source) {
-        if (Object.prototype.hasOwnProperty.call(source, key)) {
-          target[key] = source[key];
+export function _extends() {
+    _extends = Object.assign || function(target) {
```
magic-akari, line 2: I cannot rewrite the exported function.
Due to an incorrect translation during the transpilation to CJS, what is your perspective? @kdy1 

https://github.com/swc-project/swc/blob/3f2ba68d4b4b560a2a1e13ba92919fe54a0d3413/crates/swc_ecma_transforms_module/src/common_js.rs#L271-L282

kdy1: I think we can remove this 'optimization' if required

magic-akari: Do you lean towards breaking changes?
Do you prefer named imports?

If it is necessary to introduce a breaking changes, there exists an opportunity for refinement.
This means 0.5.x for `@swc/helpers`

kdy1: I don't _want_ breaking changes, but if it's inevitable I'm fine with it.

magic-akari: The breaking change refers to renaming the function name of exporting the index to `snake_case`.
However, this entry has never been used in our transpilation except for importing the helper in scripts environment (issue #7144). However, it is already broken.

kdy1: Then I think it would be okay to have a breaking change of `@swc/helpers`.

magic-akari:
Test node bindings with `_defineProperty` caused the CI to fail.
Is this indicative of a breaking change for NAPI and WASM? @kdy1

kdy1:
I don't think it's a breaking change of `@swc/core`, but I'm not sure about the exact cause

kdy1 (commented):
Just to avoid mistake

---

swc-bump:
 - swc_atoms

crates/swc_ecma_minifier/benches/full/typescript.js
```diff
@@ -8102,13 +8102,13 @@ var ts;
         A_get_accessor_cannot_have_parameters: diag(1054, ts.DiagnosticCategory.Error, "A_get_accessor_cannot_have_parameters_1054", "A 'get' accessor cannot have parameters."),
         Type_0_is_not_a_valid_async_function_return_type_in_ES5_SlashES3_because_it_does_not_refer_to_a_Promise_compatible_constructor_value: diag(1055, ts.DiagnosticCategory.Error, "Type_0_is_not_a_valid_async_function_return_type_in_ES5_SlashES3_because_it_does_not_refer_to_a_Prom_1055", "Type '{0}' is not a valid async function return type in ES5/ES3 because it does not refer to a Promise-compatible constructor value."),
```
kdy1, line 8103: Please revert changes in this file

crates/swc/tests/fixture/shopify/004-jest/output/index.js
```diff
@@ -8,11 +8,11 @@ Object.defineProperty(exports, "App", {
...
     }
 });
-var _interopRequireDefault = require("@swc/helpers/lib/_interop_require_default.js").default;
-var _slicedToArray = require("@swc/helpers/lib/_sliced_to_array.js").default;
-var _react = /*#__PURE__*/ _interopRequireDefault(require("react"));
-var _reactI18N = require("@shopify/react-i18n");
+var _interop_require_default = require("@swc/helpers/lib/_interop_require_default.js").default;
+var _sliced_to_array = require("@swc/helpers/lib/_sliced_to_array.js").default;
+var _react = /*#__PURE__*/ _interop_require_default(require("react"));
+var _react_i_1_8n = require("@shopify/react-i18n");
```
kdy1, line 14: Is this intended?

magic-akari: Yes, the optional choice here is using snake_case only for swc helpers or using snake_case for all.
I have opted for the latter.
Which option do you prefer?

magic-akari: I find the `_react_i_1_8n` is really strange.

kdy1: Yeah, I think the latter is fine, but I think `_i_1_8_n` would disturb debugging

kdy1: It seems like there's no good alternative.
`convert_case` produces `react_i_18_n`

crates/swc_ecma_minifier/tests/fixture/issues/2257/full/input.js
```diff
@@ -2624,7 +2624,7 @@
                     return /* reexport */ _toPropertyKey;
                 },
                 typeOf: function () {
-                    return /* reexport */ _type_of_typeof;
+                    return /* reexport */ _typeof_typeof;
```
kdy1, line 2627: Can you revert input files of tests, please?

package.json
```diff
@@ -133,6 +133,9 @@
         "ts-node": "^10.5.0",
         "typescript": "^4.5.2"
     },
+    "peerDependencies": {
+        "@swc/helpers": ">0.4.14 <0.5"
```
kdy1, line 137: I think we should publish `@swc/helpers@0.5.0`, but this should done by another commit, right?

crates/swc_ecma_minifier/tests/fixture/issues/2257/full/input.js
magic-akari: When I execute `cargo test` with the flag `UPDATE=1`, the input file is automatically updated.
I am currently investigating the cause.

package.json
magic-akari: Indeed, I believe this Pull Request does not introduce any breaking changes but rather is a bugfix.

In my opinion, the current helpers are insufficient, and I think it would be worthwhile to introduce breaking changes, including the use of named exports in another Pull Request. I am currently preparing an RFC.

crates/swc_ecma_minifier/tests/fixture/issues/2257/full/input.js
magic-akari: never mind. The the It's `_type_of_typeof`. what a confusing name.

kdy1 (approved):

kdy1 (changes requested):
I don't know the reason, but CI failed

magic-akari:
> I don't know the reason, but CI failed

https://github.com/swc-project/swc/blob/c7d57c4b738f497a8b25173ccde713b5348cd769/.github/workflows/CI.yml#L170-L173

It caused by patch-section.
I will run it locally and update the test results.

kdy1 (approved):

PR #7153, opened 2023-03-28
fix(css/compat): Remove usage of `box_syntax`

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_css_compat

PR #7159, opened 2023-03-29
feat: add PorterJs transform module type

cyjake:
**Description:**

This pr adds a new transform module type called PorterJs, which is the bundler I've been working on and off for years. The reason to add this particular module type in SWC, rather than relying purely on SWC plugins, is that the default ones aren't quite the fit for  porter, specifically:

- both import and require dependencies need to be collected for porter,
- condition requires need to be processed early on, no matter minify is activated or not

more details are discussed in this porter documentation <https://porterhq.github.io/porter/swc>, with the new module type porter can transform JavaScript and TypeScript modules in one pass, and concatenating them into bundles accordingly. I've done a few benchmarks on compiling the source of <https://chuangyi.taobao.com>, which builds roughly 50% faster for now.

**BREAKING CHANGE:**

This pr doesn't break existing features, just adds the transform module type

n/a


cyjake:
understood, I'll post an RFC first. The bundler uses several plugins already, such as [glob import](https://github.com/porterhq/porter/tree/master/packages/swc-plugin-glob-import), which is quite handy. But I do need the new module type to properly handle the transform and minify process.

PR #7163, opened 2023-03-29
fix(plugin): Use `rmp_serde` instead of `rkyv`

kdy1:
**Description:**

This PR changes serialization/deserialization for plugins from `rkyv` to `rmp-serde`. The performance is a bit regressed, but I think it's fine. For the input file of size 9.6 MB (typescript), a single invocation of a Wasm plugin with `rkyv` took 0.4 seconds, while `rmp-serde` took 1 second. It's a clear regression, but we can now have backward compatibility to some extent, and we will be able to update `rustc` too.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/6807.


kdy1 (commented):
swc-bump:
 - swc_atoms --breaking
 - ast_node --breaking
 - swc_plugin_proxy --breaking
 - swc_plugin_macro --breaking

kdy1:
Closing in favor of https://github.com/swc-project/swc/pull/7170

PR #7168, opened 2023-03-29
chore(ci): Verify publish action

kdy1:
**Description:**

I found that CI failure may not be related to `rkyv`, so lets' try.


kdy1:
Closing as the cause was `rkyv`

PR #7171, opened 2023-03-30
fix(es/compat): Handle temporary variable injection position

### LINKED ISSUE #7169: Logical Operator Assignment Generates Invalid Javascript
### Describe the bug

When a function contains an operator assignment (`||=`, etc) followed by a closure, SWC will generate invalid Javascript that triggers either an "Uncaught ReferenceError" (in strict mode) or, when not in strict mode, adds a global variable to window.

SWC creates a temporary variable when transpiling operator assignment. For some reason it pushes that variable's declaration into the closure, which means it's not in the outer function's scope.

### Input code

```typescript
function myFunc(options) {
  options.context ||= {}
  const closure = function() {}
}

myFunc({})

console.log({
  _$_options: window._$_options, // <= ES5
  _options: window._options // >= ES2015
})
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript"
    },
    "target": "es5"
  },
  "minify": false
}
```

### Playground link

https://play.swc.rs/?version=1.3.43&code=H4sIAAAAAAAAA2WOwQrCMBBE7%2FsVQ%2FHQgLQq9CLWm%2F6AH9BDTKVQs9KkVGnz72ZpwIO33Zm3M9uOVvuOLZ6faxxzfsnmFGYCyhLZ6AycHzrts6gku9BsvXl7LEuNOUQjCs5D9%2BzGwaBGm3JzJX4gmjp756lYayKwDpRq56CIJIN7U%2FT8yKW%2B2TSp74h0%2FpO28t2pxuVWCfoHJkGos1CH3b6ioL4Ua1OT8AAAAA%3D%3D&config=H4sIAAAAAAAAA0WNSwqAMAxE956iZO2iuHDhbUJNpeKnNBEs4t2NLeguM28euRpjYGYHg7n01BAxMaUva8N5Ezy1AXIrskshChR4t9URTBNJWXBnO%2FvSwmANW%2FBZiceFqVb7eCz0P5Qcqap98Zr7ASqyi6aVAAAA

### Expected behavior

The above input code should not set any properties on window, and should not raise errors in strict mode.

It should put the variable declaration outside the closure, like this:
```javascript
function myFunc(options) {
    var _$_options;
    (_$_options = options).context || (_$_options.context = {});
    var closure = function() {};
}
```

### Actual behavior

It puts the variable declaration inside the closure.

```javascript
function myFunc(options) {
    (_$_options = options).context || (_$_options.context = {});
    var closure = function() {
        var _$_options;
    };
}
```

### Version

1.13.43

### Additional context

Target is es5 in example but the same problem exists for any target that doesn't natively support operator assignment (though the temp variable name is different.) Play with the target, module settings, and "use strict" in the playground to see how it behaves under different circumstances.

[The same code in babel](https://babeljs.io/repl#?browsers=since%202010%0A%0A&build=&builtIns=false&corejs=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABAWwJ4DFwQBRwA6wIDOAlIgN4BQiiA9LYgEQhECmiRUATjNI9YnyEwRAHQQEUVgA8oiAD7yAvBQC-AiSLkQANnCIgu7FaEjDsZcuvWUA7jDAATOLdFpMkRCvdZKPyNhWJJSUmkRwOqyiegDmgQIA-gAkCULwIgBciPZOLqLJqQTpRAA0dAwAPCoAogDKAKyJacRZOc6uhcJE5YgAfDW1AEwADACMjaokQA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=script&lineWrap=true&presets=env&prettier=false&targets=&version=7.21.3&externalPlugins=&assumptions=%7B%7D) doesn't seem to use a temporary variable at all.

magic-akari:
**Description:**

**BREAKING CHANGE:**

- fix #7169


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_compat

dleavitt:
🏎️ !

PR #7176, opened 2023-03-30
fix(es/compat): Handle shorthand object prop when rename

### LINKED ISSUE #7174: props key of window is transformed
### Describe the bug

Object has a key of `window` is been transformed to _$window. when window is used outside of a function.

### Input code

```typescript
const isBrowser = typeof window !== "undefined";

function test({
  window
}) {
  create({ window })
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true,
  "env": {
    "targets": "chrome 44"
  }
}
```

### Playground link

https://play.swc.rs/?version=1.3.44&code=H4sIAAAAAAAAAzXMQQqAIBBG4f2c4s%2BVnSHcdJPQEdzMhI5IhHePoJYPPl5UaYbS9qqjcUWAXSdrxiiSdGAJAa5L4lyEk9uIcpdoRQXGzfxN%2BCjNFW%2FFyoexv%2F%2FDXGnSA6wxs91nAAAA&config=H4sIAAAAAAAAA1WOSw6DMAxE9zlF5HWXqIveoYeIUkNT5Sc7VEWIu5NAQmFnz4yfZxZSwoc1POScx7xERYx07FnhySf1ywqkKSJrMjHBrbmJi9Ury7hJy%2B6ADYGxOVVzxpt%2BOrN1cJGQ%2BRosUeUHi1eyqHRw4TVuZu1cepV%2ByHf4h9qzAwyGn%2B0y0bhr6L8nkKIBU2kD%2Bk3Boey6jSiWFRxKM0YqAQAA

### Expected behavior

The property key should not be transformed

### Actual behavior

_No response_

### Version

1.3.44

### Additional context

_No response_

magic-akari:
**Description:**

**BREAKING CHANGE:**

- fix #7174

### DIFF
--- a/crates/swc/tests/tsc-references/intraExpressionInferences.1.normal.js
+++ b/crates/swc/tests/tsc-references/intraExpressionInferences.1.normal.js
@@ -89,7 +89,7 @@ test({
         return "a";
     },
     c: function(b) {
-        var _$x = b;
+        var x1 = b;
     }
 });
 test({
@@ -100,7 +100,7 @@ test({
         return a;
     },
     c: function(b) {
-        var _$x = b;
+        var x1 = b;
     }
 });
 // Repro from #41712
--- a/crates/swc/tests/tsc-references/intraExpressionInferences.1.normal.js
+++ b/crates/swc/tests/tsc-references/intraExpressionInferences.1.normal.js
@@ -89,7 +89,7 @@ test({
         return "a";
     },
     c: function(b) {
-        var x1 = b;
+        var _$x = b;
     }
 });
 test({
@@ -100,7 +100,7 @@ test({
         return a;
     },
     c: function(b) {
-        var x1 = b;
+        var _$x = b;
     }
 });
 // Repro from #41712
--- a/crates/swc_ecma_transforms_compat/src/es2015/block_scoping/operator.rs
+++ b/crates/swc_ecma_transforms_compat/src/es2015/block_scoping/operator.rs
@@ -1,4 +1,4 @@
-use swc_common::{collections::AHashMap, Spanned, SyntaxContext};
+use swc_common::{collections::AHashMap, SyntaxContext};
 use swc_ecma_ast::*;
 use swc_ecma_visit::{noop_visit_mut_type, VisitMut, VisitMutWith};
 use swc_trace_macro::swc_trace;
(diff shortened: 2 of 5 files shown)


crates/swc/tests/tsc-references/intraExpressionInferences.1.normal.js
```diff
@@ -89,7 +89,7 @@ test({
         return "a";
     },
     c: function(b) {
-        var _$x = b;
+        var x1 = b;
```
magic-akari, line 92: I'm not sure what caused these changes. 
But I think it's OK.

crates/swc_ecma_transforms_compat/src/es2015/block_scoping/operator.rs
```diff
@@ -26,6 +26,26 @@ impl VisitMut for Rename {
         }
     }
 
+    fn visit_mut_prop(&mut self, n: &mut Prop) {
+        let span = n.span();
+        if span.ctxt == SyntaxContext::empty() {
```
magic-akari, line 31: Should I omit this check logic?

kdy1: I think this logic can cause a problem

kdy1: Why did you add this logic?

kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_compat

crates/swc_ecma_transforms_compat/src/es2015/block_scoping/operator.rs
magic-akari: To skip prop which generated by ourselves.
I noticed that it's the reason caused the test result change.
I will correct it.

https://github.com/swc-project/swc/blob/6a05c20a19d048031cc25e20dedbc517a31cc346/crates/swc/tests/tsc-references/intraExpressionInferences.1.normal.js#L91-L93

https://github.com/swc-project/swc/blob/6a05c20a19d048031cc25e20dedbc517a31cc346/crates/swc_ecma_parser/tests/tsc/intraExpressionInferences.ts#L98-L100

kdy1 (approved):
Thank you!

PR #7182, opened 2023-03-31
Implement new `@swc/helpers`

### LINKED ISSUE #7157: [RFC] Next Step for @swc/helpers
## Motivation

Current issues with @swc/helpers:

1. `default export` used in helpers
2. Different import paths used in CJS and ESM

---

The use of `default export` incurs the `interopRequireDefault` overhead when transpiling to CJS.
We should use named export to avoid it.

Additionally, keeping the exported function name consistent with the filename can reduce conversion complexity and aid in debugging.

CJS and ESM should have different module formats, but this should not be explicitly done in `@swc/core`.
We can use a unified import path and rely on the exports field to automatically select the suitable module format for runtime/bundler.

### 1. use named export

This is a simple changes and also the reason for breaking changes.

### 2. unified import path

To ensure semantic clarity and ease of debugging, we have opted to use `esm` and `cjs` path instead of `src` and `lib`.
Taking `_array_like_to_array` as an example, after relocating the module, the directory structure is as follows.

```
.
├── _/_array_like_to_array
│   └── package.json
├── cjs
│   └── _array_like_to_array.cjs
├── esm
│   └── _array_like_to_array.js
└── package.json
```

There are a few details to note, such as utilizing `.js` as the file extension for esm
and declaring `type: module` in the `package.json`.
This prioritizes the use of esm modules and hints the bundler to use them as much as possible.

```json
{
  "type": "module",
  "exports": {
    "./_/_array_like_to_array": {
      "import": "./esm/_array_like_to_array.js",
      "default": "./cjs/_array_like_to_array.cjs"
    }
  }
}
```

By adding the `exports` field to the `package.json` file, we can benefit from [conditional exports] and use a uniform import path during translation.

```JavaScript
// esm
import { _array_like_to_array } from "@swc/helpers/_/_array_like_to_array";

_array_like_to_array(foo);
```

```JavaScript
// cjs
var _array_like_to_array = require("@swc/helpers/_/_array_like_to_array");

_array_like_to_array._array_like_to_array(foo);
```

note: I have chosen the path `_` here to ensure the shortest possible pathway.

To ensure compatibility with bundlers/runtimes that do not support conditional exports,
we add the file `./_/_array_like_to_array/package.json` with the following content:

```json
{
  "main": "../../cjs/_array_like_to_array.cjs",
  "module": "../../esm/_array_like_to_array.js"
}
```

[conditional exports]: https://nodejs.org/api/packages.html#conditional-exports

_No response_

### Additional context

These changes would require introducing breaking changes, meaning that @swc/helpers would need to be updated to version 0.5.0.

---

I have published a [@magic-akari/swc-helpers-experiments](https://www.npmjs.com/package/@magic-akari/swc-helpers-experiments?activeTab=code) package. Compatibility testing is currently in progress and I will update the test results here.

magic-akari:
**Description:**

- named export
- unified import path

**BREAKING CHANGE:**

Breaking changes for `@swc/helpers`.
new major version `0.5.0` is required.

- close #7157


crates/swc/tests/fixture/deno/paths/issue-2724/output/src/index.ts
```diff
@@ -2,8 +2,8 @@
 Object.defineProperty(exports, "__esModule", {
     value: true
 });
-const _interop_require_wildcard = require("@swc/helpers/lib/_interop_require_wildcard.js").default;
+const _interop_require_wildcard = require("@swc/helpers/_/_interop_require_wildcard");
 (async function() {
-    const { displayA  } = await Promise.resolve().then(()=>/*#__PURE__*/ _interop_require_wildcard(require("../packages/a/src/index")));
+    const { displayA  } = await Promise.resolve().then(()=>/*#__PURE__*/ (0, _interop_require_wildcard._interop_require_wildcard)(require("../packages/a/src/index")));
```
kdy1, line 7: I think this will regress the bundle size by margin.

magic-akari: Do you mean the indirect call?

We known that there is no `this` issue in our helpers.
Is it safe to call function directlly?

kdy1: I meant that `_interop_xxx` cannot be optimized by the minifier.

I think it's better to invoke it directly

magic-akari: > I meant that `_interop_xxx` cannot be optimized by the minifier.
> 
> I think it's better to invoke it directly

Done. fixed by keep using the pattern `fn.apply(this, arguments)` .

magic-akari: I still need to handle it in CJS/AMD/UMD

magic-akari: > I meant that `_interop_xxx` cannot be optimized by the minifier.
> 
> I think it's better to invoke it directly

The first solution is to disable rewriting exported functions in esm so that we can safely use `var foo = require("foo").foo`.
But if other bundler handle our helpers, it does not know that our export is constant and may not apply such optimization.

The second solution is to add another constant named export with shorter name such as `_`, and then use it in transpiled file as follows: 

```JavaScript
var foo = require("foo");
foo._(bar);.
```

Which is better?

kdy1: Wow, the second one sounds cool. What a clever trick!

magic-akari: is there a simple way to get helpers SyntaxContext?
I am trying to skip the indrect call `(0, foo)._` for helpers.

magic-akari: I got it by using
```
HELPERS.with(|helper| helper.mark());
```

crates/swc/tests/fixture/config/regenerator/import-path/1/output/index.js
```diff
@@ -1,4 +1,4 @@
-import _ts_generator from "@swc/helpers/src/_ts_generator.mjs";
+import { _ts_generator } from "@swc/helpers/_/_ts_generator";
```
kdy1, line 1: I think this will regress the bundle size, too.

```js
import { _ts_generator as e } from "@swc/helpers/_/_ts_generator";

e()
```

This is how it's optimized, but this looks problematic

magic-akari: The test cases will be updated.
It will be changed to the following:

```JavaScript 
import { _ as _ts_generator } from "@swc/helpers/_/_ts_generator";
```

and with optimized phase:

```JavaScript 
import { _ as e } from "@swc/helpers/_/_ts_generator";

e()
```

kdy1: I see, thank you so much!

kdy1:
I don't know why, but GitHub shows a wrong diff. I'm not sure how should I review this 🤣

magic-akari:
> I don't know why, but GitHub shows a wrong diff. I'm not sure how should I review this 🤣

image

I wish the builtin GitHub plugin from VSCode will solve this issue.

kdy1:
I already tried it, but it didn't work

magic-akari:
> I already tried it, but it didn't work

Try checkout localy?
Or should I rebase and squash some commits?

kdy1:
I checked it out, but `git diff upstream/main...pr/magic-akari/7182` fails with `fatal: upstream/main...pr/magic-akari/7182: no merge base`.
Can you try rebasing?

magic-akari:
> I checked it out, but `git diff upstream/main...pr/magic-akari/7182` fails with `fatal: upstream/main...pr/magic-akari/7182: no merge base`. Can you try rebasing?

Done.
I have split the test cases commits.

kdy1:
It didn't work, but I think I can review it using cli tools now

packages/swc-helpers/.dprint.json
```diff
@@ -0,0 +1,29 @@
+{
```
kdy1: Can you revert this?

kdy1: Did you commit this intentionally?

packages/swc-helpers/esm/_update.js
```diff
@@ -1,6 +1,7 @@
-import _get from "./_get.mjs";
-import _set from "./_set.mjs";
-export default function _update(target, property, receiver, isStrict) {
+import { _get } from "./_get.js";
```
kdy1: Will this work? Isn't the export named `_`?

packages/swc-helpers/.dprint.json
magic-akari: I use it to fomat code.
I should add it to the `.npmignore`

packages/swc-helpers/esm/_update.js
magic-akari: Yes there are two named export.

kdy1 (dismissed):
Thank you! I'll update swc/fixture and swc/tsc soon

---


swc-bump:
 - swc_ecma_utils

kdy1 (approved):

PR #7184, opened 2023-04-01
fix(es/modules): Hint nodejs for multiple `export *`

### LINKED ISSUE #7180: Can not import named-export from transpiled cjs
### Describe the bug

The transpiled cjs code can not be imported by named. Only workaroud is: `import lib from 'lib'; const fn = lib.fn`

```
//test.mjs
import {fn} from 'lib';
```

The codes to be transpiled:

### Input code

```js
// fn.js
export function fn(){}
export default fn
```
```js
// fn2.js
export function fn2(){}
export default fn2
```

```typescript
//index.js

export * from './fn.js'
export * from './fn2.js'
```

### Config

```json
{
  "jsc": {
    "target": "es2015",
    "keepClassNames": true,
    "parser": {
      "syntax": "typescript",
      "decorators": true,
      "dynamicImport": true,
    },
    "minify": {
      "compress": false,
      "mangle": false,
    },
    "transform": {
      "legacyDecorator": true,
    }
  },
  "module": {
    "type": "commonjs",
    "importInterop": "node",
  },
}
```

### Playground link

_No response_

### Expected behavior

Can be imported by name.

### Actual behavior

raise error:

```
SyntaxError: Named export 'fn' not found. The requested module 'lib' is a CommonJS module, which may not support all module.exports as named exports.
```

### Version

1.3.42

### Additional context

```
@swc/cli: 0.1.62
@swc/core: 1.3.42
node: v18.15.0
```

magic-akari:
**Description:**

`cjs-module-lexer` detection is fragile. Redundant parentheses can cause detection failure.
Therefore, I replaced the sequence expression with a binary expression.

before:

```JavaScript 
0 && (__export(require("foo")), __export(require("bar")));
```

after this pull request

```JavaScript 
0 && __export(require("foo")) &&  __export(require("bar"));
```

**BREAKING CHANGE:**

- fix #7180

### DIFF
--- a/crates/swc_ecma_transforms_module/src/common_js.rs
+++ b/crates/swc_ecma_transforms_module/src/common_js.rs
@@ -530,12 +530,13 @@ where
 
     /// emit [cjs-module-lexer](https://github.com/nodejs/cjs-module-lexer) friendly exports list
     /// ```javascript
-    /// 0 && (__export(require("foo")));
+    /// 0 && __export(require("foo")) && __export(require("bar"));
     /// ```
     fn emit_lexer_ts_reexport(&self, link: &Link) -> Option<Stmt> {
-        let mut seq_list = vec![];
-        link.iter().for_each(|(src, LinkItem(_, _, link_flag))| {
-            if link_flag.export_star() {
+        let expr = link
+            .iter()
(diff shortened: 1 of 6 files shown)


kdy1 (approved):
Thank you!

---

swc-bump:
 -  swc_ecma_transforms_module

PR #7188, opened 2023-04-01
feat(plugin_runner): improve cache path

kwonoj:
**Description:**

This PR is minor improvement, non breaking changes to cache path for the plugin. In short, it tries to combine rustc version if available to expand. `option_env!` supposed to embed build time rustc version. It'd be ideal if we could match between plugin's as well, but I doubt if we could before actually load plugin.


kdy1 (approved):
swc-bump:
 - swc_plugin_runner

PR #7191, opened 2023-04-02
fix(es/parser): fix infinite loop on jsx in js

### LINKED ISSUE #7189: Infinite loop when JSX code is in `.js` file
### Describe the bug

`swc` runs into an infinite loop when trying to parse a js file that contains jsx syntax, but isn't told to parse it as jsx.

Steps to reproduce:

1. Create a `file.js` file and paste the following contents:
  ```jsx
  const a = (
	<section>
		<a href="/foo">foo</a>
	</section>
  );
  ```
2. Run `npx swc ./file.js` -> Infinite loop

### Input code

```jsx
const a = (
	<section>
		<a href="/foo">foo</a>
	</section>
);
```

### Config

```json
none, whatever `npx swc` uses by default.
```

### Playground link

Not possible because the infinite loop leads to the tab being killed.

### Expected behavior

SWC doesn't run into an infinite loop. It should error instead if the syntax is deemed invalid.

### Actual behavior

SWC consumes 100% CPU and is stuck in an infinite loop.

### Version

@swc/cli: 0.1.62 @swc/core: 1.3.44

### Additional context

Related https://github.com/denoland/deno/issues/18565

marvinhagemeister:
**Description:**

This fixes an infinite loop in the parser when it encountered an invalid token. This was easiest to reproduce by trying to parse JSX as JS.

I'm new to the swc codebase and I couldn't find places which a have a similar logic. So I opted to break out of the loop instead of modifying the `bump!` macro. Let me know if that is not the proper way to do it. 

**BREAKING CHANGE:**
no

Fixes #7189

### DIFF
--- a/crates/swc_ecma_parser/src/parser/stmt.rs
+++ b/crates/swc_ecma_parser/src/parser/stmt.rs
@@ -845,6 +845,10 @@ impl<'a, I: Tokens> Parser<I> {
 
             while !eat!(self, ';') {
                 bump!(self);
+
+                if let Some(Token::Error(_)) = self.input.cur() {
+                    break;
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_parser

PR #7196, opened 2023-04-03
fix(es/transform_typescript): should consider `A` as a type in `export { type A }`

### LINKED ISSUE #7076: `export { type <ident> }` does not result in `<ident>` being marked as a type
### Describe the bug

`export { type <ident> }` does not result in `<ident>` being marked as a type, and as such does not get stripped out unless declared with `type` in the import.

Additionally `export {}` is omitted with `export { type <ident> }` even though it should not be. This is the distinguishing factor between `export { type <ident> }` and `export type { <ident> }`

### Input code

```typescript
import { A } from "./a.ts";
export { type A };
```

### Config

_No response_

### Playground link

https://play.swc.rs/?version=1.3.39&code=H4sIAAAAAAAAA8vMLcgvKlGoVnBUqFVIK8rPVVDS00%2FUKylWsuZKrYDKlVQWpIIUWAMAXodgqi4AAAA%3D&config=H4sIAAAAAAAAA0WNSwrDMAxE76J1ts0id%2BghjKsEF%2F%2FQKBBjfPfawSU7aebpqdIXlrZK2QhYxoQS1Vy0kZbMsOKy0kKKHu3Gg1tfjBysHWG8eudTAs92oeCi28sw2RSyMPBUJh7%2BT7YuCulzjqDez27hSu1xzDuH9wRVTm4%2FOkwQUrUAAAA%3D

### Expected behavior

```js
export {};
```

### Actual behavior

```js
import { A } from "./a.ts";
```

### Version

swc_ecma_transforms_typescript@0.164.10

### Additional context

TSC behaviour with isolated modules: https://www.typescriptlang.org/play?isolatedModules=true#code/PTAECMDoBcGcCgCWBbADgewE7VAb1AIKgC+oAZpusqAESTACGkAVrDQNzwCmAHhtnlDQAnqi6ES7IA

Babel bug: https://github.com/babel/babel/issues/15493

hyf0:
**Description:**

Fixes #7076.

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_transforms_typescript/src/strip.rs
+++ b/crates/swc_ecma_transforms_typescript/src/strip.rs
@@ -165,7 +165,8 @@ pub fn strip_with_config(config: Config, top_level_mark: Mark) -> impl Fold + Vi
             is_type_only_export: Default::default(),
             decl_names: Default::default(),
             in_var_pat: Default::default(),
-            keys: Default::default()
+            keys: Default::default(),
+            exported_variable_to_is_type: Default::default(),
         }),
         inline_enum(ts_enum_lit, ts_enum_config)
     )
@@ -236,6 +237,7 @@ where
             decl_names: Default::default(),
             in_var_pat: Default::default(),
             keys: Default::default(),
+            exported_variable_to_is_type: Default::default(),
         }),
         inline_enum(ts_enum_lit, ts_enum_config)
     )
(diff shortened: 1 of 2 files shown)


kdy1 (approved):
Thanks!

---

swc-bump:

magic-akari:
I believe this may not be a correct fix. You should add the following test case.

```TypeScript
import { A } from "./a.ts";
new A();
export { type A };

```

hyf0:
> I believe this may not be a correct fix. You should add the following test case.
> 
> ```ts
> import { A } from "./a.ts";
> new A();
> export { type A };
> ```

Good catch! There is still a lot to learn to me. I'll continue the fix weekend and if you want to take this, be my guest.

PR #5456, opened 2022-08-12
feat(swc/plugin): wasmer@3 

kwonoj:
**Description:**

require https://github.com/swc-project/swc/pull/5507 to rebase.

**BREAKING CHANGE:**

**Related issue (if exists):**
 - Closes https://github.com/swc-project/swc/issues/5480

PR #7197, opened 2023-04-03
feat(plugin/runner): Update `wasmer` to `v3`

kwonoj:
**Description:**

This PR reattempts https://github.com/swc-project/swc/pull/5456.

Most of changes are for the breaking changes of wasmer@3, as well as enabling rkyv's strict mode (https://github.com/swc-project/swc/pull/6922). This could not be seperated since wasmer@3 enables strict mode by default without way to turn it off.

There are couple of changes worth to note:

- Disabling in-memory module lookup: https://github.com/swc-project/swc/pull/7197/files#diff-3bda5def6ce2b7553c3b3a5ad241c0bdb7021e67b7de1e594df4cd5a54d403b3R154-R159 
- Disabling plugin_runner in bindings_wasm: https://github.com/swc-project/swc/pull/7197/files#diff-dc3ded556a1fd709a129acd588e5eda651b842c6acc3f5340d40088a1f927facR310-R312
- Skipping plugin compat test: https://github.com/swc-project/swc/pull/7197/files#diff-531197dfcefba05faca53f0cf442ecc2dc6b59d5ead01979f5ffb912aa36249aR64-R66

### DIFF
--- a/crates/swc_plugin/src/allocation.rs
+++ b/crates/swc_plugin/src/allocation.rs
@@ -1,3 +1,5 @@
+use std::alloc::{alloc as global_alloc, dealloc as global_dealloc, Layout};
+
 /// Allocate bytes that won't be dropped by the allocator.
 /// Return the pointer to the leaked allocation so the host can write to it.
 ///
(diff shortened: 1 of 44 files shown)


kwonoj:
Unsure what to do with

- https://github.com/swc-project/swc/actions/runs/4649648894/jobs/8228139357
- CLA: https://github.com/swc-project/swc/pull/7197#issuecomment-1500950948

kdy1:
About cargo deny, you can add a crate to allowlist

About CLA, we can ignore it.

kwonoj:
It seems we can't make an exception for the unlicensed pkgs, (https://github.com/swc-project/swc/actions/runs/4649770971/jobs/8228348524?pr=7197#step:5:15) only global config would work. https://github.com/EmbarkStudios/cargo-deny/issues/451

crates/swc_plugin/src/allocation.rs
```diff
@@ -31,5 +33,46 @@ pub extern "C" fn __alloc(size: i32) -> *mut u8 {
 pub extern "C" fn __free(ptr: *mut u8, size: i32) -> i32 {
```
kdy1, line 33: Did you leave this function intentionally?

kwonoj: These are actual function being used currently. The new implementation is guarded via unused flag (https://github.com/swc-project/swc/pull/7197/files/8ab88722c78c828a2bfa3990cf2b23a185830e1e#diff-ae1999f472ec8ffb61b34fb95390e186e01553af65a00ab8b02f75433ea5f182R44).

I would like to replace, but having it in this PR would not be a good idea since it'll make bisect hard if something starts to breaking.

kdy1: Thank you!

kdy1 (approved):
Thank you so much!

---

swc-bump:
 - swc_atoms --breaking
 - swc_common --breaking

PR #7198, opened 2023-04-03
fix(core) ecma_loader_parking_lot reexport name

ajjahn:
**Description:**

Corrects what appears to be a type in `swc_core`'s reexport of `ecma_loader_parking_lot`.

**BREAKING CHANGE:**


kdy1 (approved):
swc-bump:
 - swc_core

Issue #7194, opened 2023-04-03
SWC creates code resulting in an `Uncaught ReferenceError` with default minify/compress settings


danieldiekmeier:
### Describe the bug

Hello! We're trying to use SWC as an alternative to Terser, but in a specific scenario, it produces invalid code.

The input code below is the result of compiling this TypeScript code, which defines an enum, with Babel:

```ts
enum MyEnum {
  First = 'first',
  Second = 'second',
}
```

[Babel Playground →](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=KYOwrgtgBAsgngUXNA3gKClAYgSwE4DOALlALxQDkAZvsRQDQZQDKwAxgPYgAmZlB7LtwZoAvkA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Cstage-2%2Ctypescript&prettier=false&targets=&version=7.21.4&externalPlugins=&assumptions=%7B%7D)

This seems to be due to the combination of

- The `__PURE__` magic comment
- The `collapse_vars` and `unused` options for `compress`. (When they both are `true`, the error happens.)

### Input code

```typescript
function example() {
  var MyEnum = /*#__PURE__*/function (MyEnum) {
    MyEnum["First"] = "first";
    MyEnum["Second"] = "second";
    return MyEnum;
  }(MyEnum || {});

  return MyEnum
}

example()

// Output Code:

(function() {
    var MyEnum1;
    return (MyEnum1 = MyEnum || {})["First"] = "first", MyEnum1["Second"] = "second", MyEnum1;
    //                ^^^^^^ This variable is never defined, 
    //                       which leads to an Uncaught ReferenceError
})();
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": {
        "collapse_vars": true,
        "unused": true,

        "arguments": false,
        "arrows": false,
        "booleans": false,
        "booleans_as_integers": false,
        "comparisons": false,
        "computed_props": false,
        "conditionals": false,
        "dead_code": false,
        "directives": false,
        "drop_console": false,
        "drop_debugger": false,
[...]

PR #7200, opened 2023-04-04
fix(es/minifier): Don't remove used var decl

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/7194.

---

# Invesitgation:

The problematic code is

https://github.com/swc-project/swc/blob/610e1bb581a04fe9b871dd7afa0cf81ab9ee7fa8/crates/swc_ecma_usage_analyzer/src/analyzer/mod.rs#L1217-L1218


kdy1:
@Austaras Can you take a look?

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

Austaras:
I believe a more fine grained solution would be most optimal -- the pure mark only marks the function call, but not the call arguments, but here we ignore usage in arguments as well.

crates/swc_ecma_usage_analyzer/src/analyzer/mod.rs
```diff
@@ -1214,13 +1214,35 @@ where
...
-                    let can_ignore =
-                        matches!(&**init, Expr::Call(call) if call.span.has_mark(marks.pure));
+
                     let id = id.to_id();
-                    self.used_recursively
-                        .insert(id.clone(), RecursiveUsage::Var { can_ignore });
-                    e.init.visit_with(&mut *self.with_ctx(ctx));
-                    self.used_recursively.remove(&id);
+
+                    match init {
```
kdy1, line 1220: @Austaras Did you mean something like this?

Austaras: I can't see changes, seems like they're force pushed.

jridgewell (approved):

PR #7210, opened 2023-04-04
doc(ecma-macros): fix `*_children_with_path` typos

ajjahn:
**Description:**

resotred -> restored

**BREAKING CHANGE:**
None. Changes to doc string only.


kdy1:

PR #7213, opened 2023-04-04
fix(ecma-transforms-optimization) correct `arguments` typos

ajjahn:
**Description:**

arguemnts -> arguments
arguements -> arguments

**BREAKING CHANGE:**


kdy1:

PR #7214, opened 2023-04-05
refactor(macros/ast-node): Drop `darling` to reduce compile time

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - ast_node

PR #7215, opened 2023-04-05
fix(es/compat): Fix `finally` handling of `generator`

kdy1:
**Description:**

Update generator implementation to match https://github.com/microsoft/TypeScript/blob/e83d61398ea0e4231e882121dd6c6bcfe4fdc9e4/src/compiler/transformers/generators.ts

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/5913.


kdy1:
Expected output:

```js

"use strict";

function generator() {
    return __generator(this, function (_a) {
        switch (_a.label) {
            case 0:
                _a.trys.push([0, , 4, 6]);
                _a.label = 1;
            case 1:
                if (!true) return [3 /*break*/, 3];
                return [4 /*yield*/, "foo"];
            case 2:
                _a.sent();
                return [3 /*break*/, 1];
            case 3: return [3 /*break*/, 6];
            case 4: return [4 /*yield*/, "bar"];
            case 5:
                _a.sent();
                return [7 /*endfinally*/];
            case 6: return [2 /*return*/];
        }
    });
}
function test() {
    var gen = generator();
    console.log(gen.next().value);
    console.log(gen.next().value);
    console.log(gen.next().value);
    console.log(gen.next().value);
}
test();

```

kdy1:
# Investigation:

`label_numbers` is different.
 - Length of tsc: 7
 - Length of swc: 6

kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_compat

PR #7202, opened 2023-04-04
refactor(es/typescript): Reimplement TypeScript pass

magic-akari:
**Description:**

## Bugfixes
- Exported `let`/`var` declarations in TypeScript namespaces should be mutable.
- Fix missing declaration of complex exported patterns in TypeScript namespaces.
- Preserve concrete TS namespaces.

## New Features
- Introducing [Verbatim Module Syntax](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax).
- Enum value will now be inlined whenever possible within a single module, optimizing runtime performance.
- Constant enums will be automatically eliminated when feasible, This can reduce bundle size.
- Added support for cross-referencing enum values.

## Deprecated
`TsEnumConfig` is deprecated
  - The `treat_const_enum_as_enum` transform option is deprecated.
  - The `ts_enum_is_readonly` assumption option is deprecated.

**BREAKING CHANGE:**

TypeScript Config is changed.

**Related issue:**

 - Closes #5197
 - Closes #5259 
 - Closes #7177
 - Closes #7453
 - Closes #7676 
 - Closes #7681
 - Closes #7791 
 - Closes #7961

PR #7221, opened 2023-04-06
fix(testing): Use program as entry

magic-akari:
**Description:**

**BREAKING CHANGE:**

this issue block #7202

### DIFF
--- a/crates/swc_ecma_transforms_testing/src/lib.rs
+++ b/crates/swc_ecma_transforms_testing/src/lib.rs
@@ -154,12 +154,12 @@ impl<'a> Tester<'a> {
         name: &str,
         syntax: Syntax,
         src: &str,
-    ) -> Result<Module, ()> {
+    ) -> Result<Program, ()> {
         let fm = self
             .cm
             .new_source_file(FileName::Real(name.into()), src.into());
 
-        let module = {
+        let program = {
             let mut p = Parser::new_from(Lexer::new(
                 syntax,
                 EsVersion::latest(),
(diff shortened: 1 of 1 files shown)


crates/swc_ecma_transforms_testing/src/lib.rs
```diff
@@ -154,20 +154,20 @@ impl<'a> Tester<'a> {
         name: &str,
         syntax: Syntax,
         src: &str,
-    ) -> Result<Module, ()> {
+    ) -> Result<Program, ()> {
```
kdy1, line 157: Is this change necessary? I think we can alternatively change only the argument passed to `fold_**`

magic-akari: I need to ensure that what `fold_with` is a program, so the return value has been changed to `Program`.

https://github.com/swc-project/swc/blob/9fce494b10b19ffc21ec99f16a246d71be8a3383/crates/swc_ecma_transforms_testing/src/lib.rs#L180-L182

convert back to Module might be difficult, as it is possible for it to be parsed as a script.
do you have some good ways to improve it?

kdy1: You mean you want to use `parse_program`?
If so, your code is the correct direction

magic-akari: Yes, I want to use `visit_mut_program`.
If I get `Module` instead of `Program`, the `visit_mut_program` will be skipped.

See https://github.com/swc-project/swc/pull/7202/files#r1159484231

magic-akari:
There are too many errors about `script` and `module`.
I don't know what is the correct behaviour.

Could you fix it? @kdy1

kdy1:
Of course, I'll do it tomorrow

PR #7222, opened 2023-04-07
fix(es/testing): Ensure that we call `fold_program`

kdy1:
**Description:**

cc @magic-akari for visibility.


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_testing

PR #7232, opened 2023-04-10
fix(es/parser): Reset ctx for cond while parsing a stmt

kdy1:
**Description:**

I found this while working on https://github.com/vercel/next.js/pull/48098

**Related issue:**


kdy1 (commented):
swc-bump:
 - swc_ecma_parser

PR #7242, opened 2023-04-10
fix(parser): parse `const` type parameters in arrow function expressions

dsherret:
Ensures code like the following gets parsed as a `const` type parameter:

```ts
export const fn = <const Data extends ReadonlyArray<unknown>>(payload: Data) => payload;
```

Also if a type parameter is `in` / `out` / `const` incorrectly, the AST should still reflect that while emitting a diagnostic.

### DIFF
--- a/crates/swc_ecma_parser/src/parser/typescript.rs
+++ b/crates/swc_ecma_parser/src/parser/typescript.rs
@@ -398,16 +398,16 @@ impl<I: Tokens> Parser<I> {
         )? {
             match modifer {
                 "const" => {
+                    is_const = true;
                     if !permit_const {
                         self.emit_err(
                             self.input.prev_span(),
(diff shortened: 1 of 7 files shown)


crates/swc_ecma_parser/src/parser/typescript.rs
```diff
@@ -398,16 +398,16 @@ impl<I: Tokens> Parser<I> {
         )? {
             match modifer {
                 "const" => {
+                    is_const = true;
```
dsherret, line 401: It's better to reflect the state of the code in the AST even when a diagnostic is being emitted.

kdy1: I agree, thank you.

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_parser

PR #7243, opened 2023-04-10
fix(es/loader): fix browser overwrites not applying correctly

icewind1991:
- es/loader: add test for browser overwrite
- es/loader: fix browser overwrites

**Description:**

Fix browser `main` overwrites from `package.json` not working when specified like

```json
{
  "name": "jquery",
  "main": "./index.js",
  "browser": {
    "./index.js": "./browser.js"
  }
}
```

Found in the wild in [solidjs](https://github.com/solidjs/solid/blob/main/packages/solid/web/package.json#L5)

This was mainly broken due to a mix of absolute and relative paths being compared.


crates/swc_ecma_loader/src/resolvers/node.rs
```diff
@@ -295,12 +295,13 @@ impl NodeModulesResolver {
 
                                 // Relative file paths are sources for this package
                                 let source = if let Some(Component::CurDir) = components.next() {
+                                    dbg!(pkg_dir, k);
```
kdy1, line 298: Please remove this line

crates/swc_ecma_loader/src/resolvers/node.rs
```diff
@@ -482,13 +487,16 @@ impl Resolve for NodeModulesResolver {
             // Handle path references for the `browser` package config
             if let TargetEnv::Browser = self.target_env {
                 if let FileName::Real(path) = &v {
-                    if let Some(pkg_base) = find_package_root(base) {
+                    if let Some(pkg_base) = find_package_root(path) {
+                        let pkg_base = to_absolute_path(&pkg_base).unwrap();
                         if let Some(item) = BROWSER_CACHE.get(&pkg_base) {
                             let value = item.value();
-                            if value.ignores.contains(path) {
+                            dbg!(path);
```
kdy1, line 494: Ditto

kdy1 (changes requested):
Thanks!

---

swc-bump: 
 - swc_ecma_loader

crates/swc_ecma_loader/src/resolvers/node.rs
icewind1991: oops :see_no_evil:

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_loader

PR #7249, opened 2023-04-11
fix: need more error info

limerickgds:
**Description:**

In a project, there are multiple files. If the error is not located in the specified file, it is difficult for the developer to troubleshoot.

**BREAKING CHANGE:**
no


kdy1 (approved):
swc-bump:
 - swc

PR #7256, opened 2023-04-13
refactor(swc): remove bytecheck flag

kwonoj:
**Description:**

We'll going to use mainline deps to enable bytecheck, removes old experimental flags.


kdy1 (approved):
swc-bump:
 - swc_atoms
 - ast_node
 - swc_plugin_proxy

PR #7257, opened 2023-04-13
build(cargo): bump up rkyv

kwonoj:
**Description:**

Changelogs: https://github.com/rkyv/rkyv/releases/tag/v0.7.41

There seems notable bugfixes, also reexports bytecheck so we can enable bytecheck without adding deps explicitly.


kdy1 (approved):
swc-bump:
 - swc_atoms
 - swc_plugin_proxy

PR #7259, opened 2023-04-13
chore: Publish `v1.3.50` with `swc_core` `v0.75.15`

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #7262, opened 2023-04-14
fix(es): Resolve the inline sourcemap decode error

limerickgds:
**Description:**
The decode error caused by other content after resolving sourceMappingURL. Babel doesn't have this issue because it judges based on the comments, which is more accurate.
https://github.com/babel/babel/blob/master/packages/babel-core/src/transformation/normalize-file.js#L63

**BREAKING CHANGE:**

no


limerickgds:
Whether it is necessary to align with Babel and process based on the last map `comment` in process_js_inner?

kdy1:
I think we can follow the behavior of babel here, although we don't _need_ to follow it.

limerickgds:
> I think we can follow the behavior of babel here, although we don't _need_ to follow it.

So, would this be a solution for now? It seems to cover most use cases. After all, Babel doesn't handle inline sourcemaps perfectly either, and it doesn't solve the problem of multiple inline sourcemaps.

kdy1:
Yeah, I will review it soon (today or tomorrow)

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc

Issue #7261, opened 2023-04-13
Mangling redeclares variable in destructuring when `eval()` is present


nolanlawson:
### Describe the bug

When `eval()` is present, the SWC minifier may redeclare a variable that has already been declared, leading to a syntax error when the JavaScript is parsed by the browser.

This only occurs when `eval()` is present, and it only occurs with `{ mangle: true }`.

### Input code

```typescript
export function func(a, b) {
  const { c: d } = a;
}
eval('');
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2015",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": {
        "toplevel": false,
        "keep_classnames": false,
        "keep_fnames": false,
        "keep_private_props": false,
        "ie8": false,
        "safari10": false
      }
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.3.50&code=H4sIAAAAAAAAA0utKMgvKlFIK81LLsnMzwMzNBJ1FJI0Faq5FBSS8%2FOKSxSqFZKtFFIUahVsFRKtuWq5UssSczTU1TWtAUyFWmI%2BAAAA&config=H4sIAAAAAAAAA32QSw7CMAxE9z1F5TULQAIh7sAZkBWcKiVprDhUVIi7k34C5SM2%2Bcy8iZW5FWUJtSjYl7d0TBfGIBSe96RI10S8JgVIORQVDEdYZLeW3tJohQbpPjoQMVQUh5Ssl6vNlADrvVBOTJozjdHdfKbyjgOJvIM9ik1laYb2ozxbasl%2Bwsk6E%2FFRWRRp0NHXc5nQf10OpsVIaff8CzK0%2B6EKagxmtXxrJ%2FVT5HVoCpw%2FXWYfgtgxja1t4QXlgp5DwMghJ2O4UHF%2FAE8y38fLAQAA

### Expected behavior

The variable should not be redeclared, and there should be no syntax error.

### Actual behavior

The output is:

```js
export function func(c, n) {
    const { c: n  } = c;
}
eval('');
```

Note that the variable `n` is redeclared. If you run this in a browser you will get the error:

    SyntaxError: Identifier 'n' has already been declared

[...]

PR #7263, opened 2023-04-14
fix(es/minifier): Fix `eval` handling of name mangler

kdy1:
**Description:**

Name mangler aborts in the existence of `eval`, and this can cause a problem for block-scoped variables because previously, name mangler assumes that those are renamed and we can use a new name freely.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/7261.


kdy1:
Closing in favor of https://github.com/swc-project/swc/pull/7271

PR #7266, opened 2023-04-14
fix(es/helpers): Do not duplicate property names

### LINKED ISSUE #7264: `@swc/helpers` does not pass `npx es-check@7.0.1 es5`
### Describe the bug

Due to the way we generate the common js index file, it's not a valid ES5, and it prevents me from updating `@swc/helpers` in `next.js` I guess.

```js
    /* @Annotate start: the CommonJS named exports for ESM import in node */
    _apply_decorated_descriptor: null,
    _array_like_to_array: null,
    _array_with_holes: null,
    _array_without_holes: null,
    _assert_this_initialized: null,


```

Note: This is a guess and I'm not sure if this is the actual cause. I need to investigate by creating a next.js app and look at the built files.

### Input code

_No response_

### Config

_No response_

### Playground link

_No response_

### Expected behavior

The test at https://github.com/vercel/next.js/blob/43519cf5868c6781f203c07c0b4794a9b6bd4c7e/test/production/escheck-output/index.test.ts#L27-L50 should pass

### Actual behavior

_No response_

### Version

@swc/helpers@v0.5.0

### Additional context

_No response_

magic-akari:
**Description:**

This pattern works in cjs-module-lexer as well. And it's valid synatx in es5 strict mode.

**BREAKING CHANGE:**

- fix #7264

### DIFF
--- a/packages/swc-helpers/scripts/build.js
+++ b/packages/swc-helpers/scripts/build.js
@@ -68,9 +68,11 @@ modules.forEach((p) => {
 
 indexCJS.push(
     `module.exports = {`,
-    "/* @Annotate start: the CommonJS named exports for ESM import in node */",
+    "/* @Annotate_start: the CommonJS named exports for ESM import in node */",
     ...cjs_module_lexer,
-    "/* @Annotate end */",
+    "/* @Annotate_end */",
+    `};`,
+    `module.exports = {`,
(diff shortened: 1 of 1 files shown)


kdy1 (approved):
Thank you so much!

---

swc-bump: 
 - dbg-swc

PR #7270, opened 2023-04-14
chore: bump bitflags

IWANABETHATGUY:
**Description:**
1. `napi-rs` has been updated `bitflags` to `2.1.x` , https://github.com/napi-rs/napi-rs/blob/main/crates/napi/Cargo.toml#L58, for now, swc still use `1.x`, when a downstream app dependends on both `swc` and `napi-rs` it may be leads to multiple version `bitflags` exists in `Cargo.lock`
2. Also, reducing multiple versions of `bitflags` would reduce compile time.

**BREAKING CHANGE:**


Cargo.lock
```diff
@@ -240,6 +240,12 @@ version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
+[[package]]
+name = "bitflags"
```
kdy1, line 244: Seems like `bitflags` is now included two times

kdy1 (approved):
swc-bump:
 - swc_common

Cargo.lock
IWANABETHATGUY: I will try to investigate that issue.

PR #7273, opened 2023-04-15
feat(es/modules): Use function instead of arrow function unconditionally

kdy1:
**Description:**

I think the minifier will optimize this as an arrow function.
This is causing a problem in next.js, so I'm checking if this will fix it.

But basically, **this is not our _problem_**. Client files of next.js are compiled to `es2016` but one of unit test in next.js expects the final app to be es5.

@magic-akari I'm sorry to ask this, but do we really need this optimization in module transform?


magic-akari:
IMO, it's OK to remove this optimization.

kdy1:
Thank you!

kdy1 (commented):
swc-bump: 
 -swc_ecma_transforms_module
 -swc_ecma_preset_env

Issue #7238, opened 2023-04-10
SourceMapProxy::lookup_char_pos causes out of bounds memory access


kwonoj:
### Describe the bug

```
Caused by:
    0: failed to invoke `/home/git_repo/swc-coverage-instrument/target/wasm32-wasi/debug/swc_plugin_coverage.wasm` as js transform plugin at /home/git_repo/swc-coverage-instrument/target/wasm32-wasi/debug/swc_plugin_coverage.wasm
    1: RuntimeError: out of bounds memory access
           at rkyv::impls::core::primitive::_::<impl rkyv::Deserialize<u32,D> for u32>::deserialize::hf987dcd4d5a23f0f (<module>[17174]:0x5db044)
           at swc_common::syntax_pos::_::<impl rkyv::Deserialize<swc_common::syntax_pos::BytePos,__D> for <swc_common::syntax_pos::BytePos as rkyv::Archive>::Archived>::deserialize::h5163a6ad2e68099f (<module>[15351]:0x54ac2b)
           at rkyv::impls::core::<impl rkyv::DeserializeUnsized<[U],D> for [T]>::deserialize_unsized::h76e0cc48824035e5 (<module>[15382]:0x54edcc)
           at rkyv::impls::alloc::vec::<impl rkyv::Deserialize<alloc::vec::Vec<T>,D> for rkyv::vec::ArchivedVec<<T as rkyv::Archive>::Archived>>::deserialize::h0fa332b260d140d0 (<module>[15121]:0x5378db)
           at swc_common::syntax_pos::_::<impl rkyv::Deserialize<swc_common::syntax_pos::SourceFile,__D> for <swc_common::syntax_pos::SourceFile as rkyv::Archive>::Archived>::deserialize::hec4b05966557a119 (<module>[15354]:0x54c143)
           at rkyv::impls::core::<impl rkyv::DeserializeUnsized<T,D> for <T as rkyv::Archive>::Archived>::deserialize_unsized::h2d776c7271413762 (<module>[15376]:0x54e631)
           at rkyv::de::SharedDeserializeRegistry::deserialize_shared::h300466f196606cc5 (<module>[15505]:0x55ef10)
[...]

PR #7280, opened 2023-04-17
feat(swc): enable bytecheck

kwonoj:
**Description:**

Second attempt to enable bytecheck. This PR does not have cersioned struct yet, just enabling bytecheck whereever possible. Also it is for the ast only yet, so transformmetadata and other might need this later.

PR seems to passing all the ci, but as we've experienced before there might be some unexpected outcome with release. Maybe better to hold this until clear https://github.com/swc-project/swc/issues/7238, then land as separate release.


kdy1 (approved):
Thank you! Let's postpone merging this..

kwonoj:
@kdy1 Resolved conflict, I think we can land this now (which.. requires another bump for swc_core and @swc/core probably though)

kwonoj:
@kdy1 thoughts on this? I think this is now ready to go in.

kdy1:
I agree, but I'm curious if this is a breaking change of `swc_core`

kwonoj:
I may wrong, but I guess not? This PR enables `repr` and bytecheck, however does not enable versioned struct we attempted previously (https://github.com/swc-project/swc/pull/5062/commits/1f3eba7f240b5d08e4688be450d5c00e6e38c2b1#diff-ef22f81d236ee08fd2bed74657c8eddc2dbef9f02f454a25941daa453b0ab2e0R173) so unless anyone explicitly relies on some of repr-based struct layout, it would work (again, I _guess_ though)

kdy1 (approved):
swc-bump:
 - swc_plugin_proxy
 - swc_atoms
 - ast_node

PR #7284, opened 2023-04-18
perf(es/react): Don't use regex

kdy1:
**Description:**

It's causing a performance problem in concurrent scenarios.

x-ref: https://linear.app/vercel/issue/WEB-921
x-ref: https://vercel.slack.com/archives/C02HY34AKME/p1681757237489689

**Related issue:**


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_react

PR #7296, opened 2023-04-19
fix(es/base): Remove unnecessary parameters in helper_expr macro

magic-akari:
**Description:**

We have move to snake_case, so the last parameter of helper_expr is unnecessary.

next step:
We have a copy of the swc/helpers directory in the swc_ecma_transforms_base directory, which has a subtle difference.
It is used for inline helpers. 

Currently, it is manually maintained. Can it be generated automatically?

**BREAKING CHANGE:**


kdy1:
> Currently, it is manually maintained. Can it be generated automatically?


I was thinking something similar. Given that we now process it via a script, I think we can update inline helpers by running a script, too.

crates/swc_ecma_transforms_base/src/helpers/_ts_decorate.js
```diff
@@ -1,4 +1,4 @@
-var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {
+var _ts_decorate = (this && this._ts_decorate) || function (decorators, target, key, desc) {
```
magic-akari, line 1: I found that I had to change export name in order to make the ts inline helpers work. 
From my personal opinion, I don't particularly like this change. It makes our code look different from the ts helpers.

kdy1: I agree witg you, but the problem was the SyntaxContext

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_react

Issue #2753, opened 2021-11-15
`jsc.paths` should handle index


kdy1:
Hey everybody, I saw a PR that was dealing with this was merged, so I went ahead and tried the latest `1.2.108` SWC version.

Unfortunately, I have to report that, at least on my end, the paths are still not being transformed the way they should be.

The steps I did were the following...

1. Created a brand new project and installed SWC.
2. Created the following files:

```
src
├── index.ts
└── utils
    └── index.ts
```

The `src/index.ts` looks like the following:

```ts
import test from "~/utils";

test();
```

3. I created the same `.swcrc` config I used above:

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript"
    },
    "target": "es2020",
    "paths": {
      "~/*": ["./src/*"]
    },
    "baseUrl": "."
  }
}
```

4. Ran `swc src -d dist --source-maps --copy-files --extensions '.ts'`

Results?

Well, ultimately, the results were the same as reported initially. 😔

The compiled code was:

```js
import test from "~/utils";
test();

//# sourceMappingURL=index.js.map
```

So, "~" wasn't replaced.

And, as stated above, once I switched to `commonjs` via ...

```json 
"module": {
    "type": "commonjs"
}
```

... the path transformation was successful:

```js
"use strict";
var _utils = _interopRequireDefault(require("./utils"));
function _interopRequireDefault(obj) {
    return obj && obj.__esModule ? obj : {
        default: obj
    };
}
(0, _utils).default();

//# sourceMappingURL=index.js.map
```

Hope I'm not missing something. If so, I'm sorry in advance for the trouble. 🙂

_Originally posted by @doitadrian in https://github.com/swc-project/swc/issues/2523#issuecomment-968595223_

doitadrian:
Thanks @kdy1 for reopening.

peterp:
I'm in the same boat, it appears that it's just replacing the imports, but it's not resolving them. As an example:

```js
{
  jsc: {
     target: 'es2020',
     baseUrl: '.',
     paths: {
       '@src/*': ['./src/*]
     }
  }
}
```

Input, files path is `./src/command/restore.ts`
```js
import { module } from `@src/config`
```

output is:

```js
var _module = require("./src/config");
```

It should be:

```js
var _module = require("../config");
```

PR #7298, opened 2023-04-19
fix: reserved function name should be private

### LINKED ISSUE #7253: Invalid variable assignment name change in ES5 with eval
### Describe the bug

The compiled code assigns a variable that doesn't exist `bmRnd1`. The variable declaration isn't renamed to match the assignment. I ran into this with `lottie-web@5.7.12` (build/player/lottie.js file). Upgrading that dependency fixed the problem (bmRnd is just a function definition in the updated code).

### Input code

```typescript
var bmRnd;
function a() {
  bmRnd = function () {};
}
eval('');
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
  }
}
```

### Playground link

https://play.swc.rs/?version=1.3.49&code=H4sIAAAAAAAAAz2KTQqAIBBG93OKb6eeYegSLdpbagQ2gqmb8O4JQsv302zGfq%2FimChUOcqVBDlVcZuN1T86RHsavIS5YcG%2F6WbjTED2pWbBMDy4M3UiP0grZfgDK1lJ%2FmUAAAA%3D&config=H4sIAAAAAAAAA6vmUlBQyipOVrJSqAYygZyCxKLi1CI4HyhSXJlXklgBFFFKTc5NLE4uyiwoUdKByWYVg6TSEnOKU8FCtRAZpZLEovTUErCuYlOw8lquWgDsxhJabwAAAA%3D%3D

### Expected behavior

swc should rename (or not rename?) both the variable declaration and the variable assignment to `bmRnd`.

### Actual behavior

swc renames the variable assignment to a different name from the variable declaration. This results in a runtime error for the built code.

### Version

1.3.46

### Additional context

I'm not 100% sure if this is expected in ES5 with eval (maybe it is?).

hyf0:
**Description:**

Fixes #7253

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_transforms/tests/es2015_function_name.rs
+++ b/crates/swc_ecma_transforms/tests/es2015_function_name.rs
@@ -57,7 +57,7 @@ test!(
     r#"number = function (x) {
   return x;
 };"#,
-    r#"number = function number(x) {
+    r#"number = function number1(x) {
   return x;
 };"#
 );
--- a/crates/swc/tests/fixture/issues-1xxx/1681/es3/output/input.js
+++ b/crates/swc/tests/fixture/issues-1xxx/1681/es3/output/input.js
@@ -14,7 +14,7 @@ function _templateObject1() {
     ], [
         "a\\nb\\nc\\n"
     ]);
-    _templateObject1 = function _templateObject1() {
+    _templateObject1 = function _templateObject() {
         return data;
     };
     return data;
--- a/crates/swc/tests/fixture/issues-1xxx/1681/es5/output/input.js
+++ b/crates/swc/tests/fixture/issues-1xxx/1681/es5/output/input.js
@@ -14,7 +14,7 @@ function _templateObject1() {
     ], [
         "a\\nb\\nc\\n"
     ]);
-    _templateObject1 = function _templateObject1() {
+    _templateObject1 = function _templateObject() {
         return data;
     };
     return data;
--- a/crates/swc/tests/fixture/issues-1xxx/1742/case1/output/index.js
+++ b/crates/swc/tests/fixture/issues-1xxx/1742/case1/output/index.js
@@ -17,7 +17,7 @@ function _templateObject1() {
     var data = _tagged_template_literal([
         "tagged template"
     ]);
-    _templateObject1 = function _templateObject1() {
+    _templateObject1 = function _templateObject() {
         return data;
     };
     return data;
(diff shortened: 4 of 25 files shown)


hyf0:
I wonder why anonymous class expression doesn't have this problem.🤔

https://play.swc.rs/?version=1.3.49&code=H4sIAAAAAAAAAytLLFJIyg3KS7Hm4korzUsuyczPUyjKL81LCUvMKU0t1kjLSUzXVKjmUoAoU7BVSM5JLC5WqK7lquXiSi1LzNFQV9e0BgA%2BVasUSAAAAA%3D%3D&config=H4sIAAAAAAAAA6vmUlBQyipOVrJSqAYygZyCxKLi1CI4HyhSXJlXklgBFFFKTc5NLE4uyiwoUdKByWYVg6TSEnOKU8FCtRAZpZLEovTUErCuYlOw8lquWgDsxhJabwAAAA%3D%3D

hyf0:
sync up: I'm going to keep investigating this Sunday.

crates/swc_ecma_transforms/tests/es2015_function_name.rs
```diff
@@ -57,7 +57,7 @@ test!(
     r#"number = function (x) {
   return x;
 };"#,
-    r#"number = function number(x) {
+    r#"number = function number1(x) {
```
hyf0, line 60: @kdy1 

After I used `private_ident!`, the problem of #2753 is solved. But the change in this test case is unexpected.

Hygiene Renamer would rename `number#private` to `number1` for these is an unresolved `number` in the global scope.

I wonder if Hygiene Renamer needs to apply special behaviors on top-level and non-top-level variables.

kdy1: I think the ideal behavior is to check if there's a reference for `number#global` in the scope of `number#private` and rename only if there are usages for both, but I'm not sure how we can implement it.

hyf0: Do you consider these test changes unacceptable? If not, we could merge this PR first after the CI is passed.

kdy1: I think it's fine

kdy1 (approved):
swc-bump:
 - swc_ecma_transforms_compat

PR #7299, opened 2023-04-19
fix path normalization on windows for testing diffs

ForsakenHarmony:
**Description:**

This makes sure that even if there's a windows path with `/` it still gets picked up

Example: `"C:/User/test/Code/project/src/f.js"` -> `"$DIR/src/f.js"`


kdy1 (approved):
swc-bump:
 - testing

PR #7300, opened 2023-04-20
test(es/compat): Add a test for a wrong issue

### LINKED ISSUE #7290: Double Nullish coalescing assignment outputs an error
### Describe the bug

When I do two nullish coalescing assignments on object second one been marked as incorrect with strange error
If I swap calls, the error will remain on exactly second call, doesn't matter which one of them it's now. 
When I remove second call, or rewrite it by handy check and assign (using if) - error gets away. 
With babel there are no such an error

Interesting fact. The error appears only on first page load. After HMR error goes away

### Input code

```typescript
filter = clone(initialFilter)

  filter.start_point ??= {
    location: null,
    radius: null
  }
  
  filter.end_point ??= {
    location: null,
    radius: null
  }
```

### Config

```json
{
  "env": {
    "targets": "> 0.25%, not dead"
  },
   "jsc": {
      "parser": {
        "syntax": "ecmascript",
        "jsx": false,
        "dynamicImport": false,
        "privateMethod": false,
        "functionBind": false,
        "exportDefaultFrom": false,
        "exportNamespaceFrom": false,
        "decorators": false,
        "topLevelAwait": false,
        "importMeta": false
      },
      "transform": null,
      "target": "es5",
      "loose": false,
      "externalHelpers": true,
      "keepClassNames": false
    },
}
```

### Playground link

_No response_

### Expected behavior

Double use of the operator is possible

### Actual behavior

image
image

### Version

0.4.14

### Additional context

The code is placed in package of monorepo. Package been loaded and transpiled by nextjs 13.3 (using built in transpilePackages option).

kdy1:
**Description:**

**Related issue:**

 - Closes #7290.

### DIFF
--- a/crates/swc_ecma_transforms_compat/tests/es2020_nullish_coalescing.rs
+++ b/crates/swc_ecma_transforms_compat/tests/es2020_nullish_coalescing.rs
@@ -1,6 +1,6 @@
 use swc_ecma_parser::Syntax;
 use swc_ecma_transforms_compat::es2020::nullish_coalescing::{nullish_coalescing, Config};
-use swc_ecma_transforms_testing::{test, test_exec};
+use swc_ecma_transforms_testing::{compare_stdout, test, test_exec};
 use swc_ecma_visit::Fold;
 
 fn tr(c: Config) -> impl Fold {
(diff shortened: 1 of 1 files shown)


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #7319, opened 2023-04-23
fix(es/codegen): Fix codegen of `TsModuleDecl`

dsherret:
This improves/fixes the emit of TsModuleDecl when it hasn't been transformed to JavaScript. I'm working on a tool that emits typescript code as-is in some scenarios.

### DIFF
--- a/crates/swc_ecma_codegen/tests/fixture/typescript/module/input.js
+++ b/crates/swc_ecma_codegen/tests/fixture/typescript/module/input.js
@@ -1,0 +1,23 @@
+module Test {
+  1;
+}
+
+export module Test.Test {
+  1;
+}
+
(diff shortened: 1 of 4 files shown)


crates/swc_ecma_codegen/tests/fixture/typescript/module/input.js
```diff
@@ -0,0 +1,23 @@
+module Test {
+  1;
+}
+
+export module Test.Test {
```
dsherret, line 5: Previously, this would panic and now it doesn't.

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_codegen

PR #7321, opened 2023-04-24
fix(syntax_pos): apply arc workaround for the rkyv archive

kwonoj:
**Description:**

Attempt to workaround #7238.

Originally I thought this is related with oom or memory layout, then realized actual deserialization error occurs in weird position - it tries to deserialize BytePos (which is obvious type) and it fails. It made me think maybe a property next to the struct layout have some unexpected behavior, notably Arc<String>.

PR applies same workaround as Atom does, and it seems to at least pass swc-coverage-instrument's usecases. May need bit more verification with other plugins to see if we can call this out as reliable workaround.


kdy1 (approved):
swc-bump:
 - swc_common

PR #7328, opened 2023-04-25
fix(swc-helpers): Add `src/*.mjs` entry back

magic-akari:
**Description:**

**Related issue:**

 - https://github.com/vercel/next.js/issues/48593

### DIFF
--- a/packages/swc-helpers/.dprint.json
+++ b/packages/swc-helpers/.dprint.json
@@ -15,7 +15,7 @@
         "indentWidth": 4
     },
     "includes": [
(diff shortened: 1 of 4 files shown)


magic-akari:
Another question: Should we add the old cjs entry `lib` as well?

kdy1:
I don't think we need `lib`, and thank you so much!

kdy1:
I'll close the issue with update PR (for `@swc/helpers`) towards next.js

kdy1 (approved):
swc-bump:
 - dbg-swc

PR #7335, opened 2023-04-25
fix(es/codegen): Emit type arguments for call expressions

sgeisenh:
**Description:**

The ECMA code generation implementation was not emitting type arguments for call expressions. This PR fixes that.


kdy1 (approved):
swc-bump:
 - swc_ecma_codegen

PR #7339, opened 2023-04-27
fix(es/react): Fix location for `$RefreshSig$` injection

kdy1:
**Description:**

This PR fixes HMR.

**Related issue:**


kdy1:
@Austaras @ForsakenHarmony Can you review this PR? I'm not 100% sure if this change is correct.

kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_react

Austaras:
I don't know if this change is wrong or right but the original output is identical to official react-refresh babel plugin.

ForsakenHarmony:
yes seems the babel plugin does it the same way, maybe defining/aliasing the function in the root, but calling inside the wrapped function would work, I think swc is handling this correctly

kdy1:
Thank you!

PR #7341, opened 2023-04-29
feat(plugin/runner): Improve caching

kwonoj:
**Description:**

Previously disabled cache (storing compiled module in memory) still doesn't work, so falling back to storing raw bytes and create new module each time. This is the same behavior to the wasm build's approach.

Also minor improved caching root to include more information. Previous rustc didn't seem to work unfortunately, so removed but instead using plugin_runner's version.


kdy1 (approved):
swc-bump:
 - swc_plugin_runner

PR #7346, opened 2023-04-30
refactor(plugin_runner): refine cache location

kwonoj:
**Description:**
I don't see how original change could affect ci in a negative way (theoretically, compile a module from IO / or in memory shouldn't be that different even in the worst case) but anyway, let's split changes to confirm. 

PR takes part of previous pr for fine grained cache path. _if this causes same regression_, that means accessing env is somehow excessively slow (which also shouldn't, but we'll see).


kdy1:
Seems like CI is fast enough

kdy1 (commented):
swc-bump:
 - swc_plugin_runner

Gabriel-Gotti (commented):
self runner deny cache for placement of another person

PR #7351, opened 2023-05-04
fix(xml/codegen): Escape `<` and `>` in child

SyMind:
**Description:**

**BREAKING CHANGE:**


kdy1 (approved):
swc-bump:
 - swc_xml_codegen

PR #7362, opened 2023-05-08
fix(es/minifier): Fix remapping of vars upon inlining

### LINKED ISSUE #7331: Re-declaration of function argument results in incorrect code
The following valid JavaScript is not handled correctly. The `var arg` is hoisted and should become a no-op, but instead the function argument seems to be optimized away as dead code (or removed for a different reason).

```js
export default function () {
	function foo (arg) {
		var arg = arg.slice()
		return arg
	}
	foo([])
}
```

Which becomes

```js
export default function() {
    var arg;
    arg = arg.slice();
}
```

[Playground](https://play.swc.rs/?version=1.3.53&code=H4sIAAAAAAAAA0utKMgvKlFISU1LLM0pUUgrzUsuyczPU9DQVKjm4oRz0%2FLzFTQSi9LBopxliUUKQI6CLYjUK87JTE7V0ASKF6WWlBblgQS5OGuBuvPzNaJjNblqAebzyNRlAAAA&config=H4sIAAAAAAAAA32VwXLiMAyG730KJmeu3cM%2BQG%2F7DB5jy8E0tjKWTGE6vPsqTqBQFE6APumXLcni%2B22z6Q7kur%2Bbb%2FkqP7jYTAFLupnEWMA6vjM0UyhA%2B8l4Wazz52U7C422EJR7FTpntiexdOCSJVfiyN32Sg80IS4VHmTYlh64BdH74t0NiASL92JKMcdwvk%2FnMI1yRHo8t8jVBJknc7ADwfaeFfyiB91m3yEOIGVZJ8aSiZmhh6IJOxwGOxKYoy2KynRSWyKhlmKClcGbseCo8uwjR8yS85l6sN449KCgWMBxPIIWJrkkLJNcT7lPwx52te9bi39Fw9EO1bKSE06tJXJaRXWPkdiEmrUSznClBjNcivs7MgZTgGvJz3EHjHmlJ58AUoHBEmWbQNNtHkHmaS06vIyMOcjI8lnhMt7aLTP0UlQTY1AqO1UGCketmwV8dTBV1mnHWfBK%2BSh6MBCCzIoiTV%2BR3V5LyucRMChA%2BmuDNlUzMLdXuMKnB%2FECf8gtWR%2BwxSNZ3q9TOqcdDi8SJOA9%2BhcO0grGdVxkS5zGdV6zBxkN8KpLpQael4A8AEYztF35NBvyPETR9APuftbEdW3fVnCyuf957%2FMaflscuoS%2BNnj9o5D%2Bzmv5T3dzivTv6tYyXP4DY%2BLMJWEGAAA%3D)

This seems to be happening in the compress step, and (unless the playground is broken) happens even if all the sub-items in the compress options are unchecked and passes is set to 0.

kdy1:
**Description:**

**Related issue:**

 - Closes #7331.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/tests/fixture/issues/7331/1/output.js
```diff
@@ -0,0 +1,5 @@
+export default function() {
+    var arg;
+    var arg;
```
jridgewell, line 3: What's up with the double `var arg`? Is it because of the param and the function body's var?

jridgewell (approved):

crates/swc_ecma_minifier/tests/fixture/issues/7331/1/output.js
kdy1: Yes. It can be dropped I guess, but it should be another PR IMO.

PR #7380, opened 2023-05-11
refactor(swc_common): bytecheck arcstring

kwonoj:
**Description:**

Simple update to missing flags.


kdy1:
Is this the correct place?
Any idea to verify this PR?

kwonoj:
We have same thing for `EncodeJsWord` as well.

I don't have way to verify yet (after all we don't use versionedstruct yet). However, this should be noop even in worst case?

kdy1:
Make sense. Thank you!

kdy1 (approved):
swc-bump:
 - swc_common

PR #7382, opened 2023-05-12
feat(plugin): Add versioned wrapper struct

kwonoj:
**Description:**

Second attempt to https://github.com/swc-project/swc/issues/5060, enabling versioned wrapper struct against serializable values for the plugin.

**BREAKING CHANGE:**

This PR is clear breaking changes for the plugin, requires bump up swc_core, @swc/core and relavant plugins (and next-swc).


kdy1 (approved):
I'll not mark this as a breaking change to avoid updating `swc_core` again needlessly, where it's not necessary because we didn't published `@swc/core` with `swc_core@v0.76.x`

---

swc-bump:
 - swc_common
 - swc_plugin_proxy
 - swc_plugin_runner

PR #7388, opened 2023-05-13
fix(es/parser): Fix `>=` with type instantiate

### LINKED ISSUE #7387: TypeScript: `x < y >= z` is incorrectly interpreted as a type argument list
### Describe the bug

SWC's interpretation of type arguments in TypeScript code is different than the TypeScript compiler, which can cause SWC to generate incorrect JavaScript.

### Input code

```typescript
fn(x < y, x >= y)
```

### Config

```json
(I only ran SWC via the playground)
```

### Playground link

https://play.swc.rs/?version=1.3.57&code=H4sIAAAAAAAAA0vL06hQsFGo1FGoULCzVajU5AIAIT%2B1gxIAAAA%3D&config=H4sIAAAAAAAAA1VPOw7DIAzdOQXy3CHK0KF36CEQdSIqfsKOVBTl7oUE0maz38fveRVSwps0PORaxrJElQjTuReEsmf1KQhwjkg6mchw6yxTpSZlCXdoOxhglWbk6kIah3FoDrAhEHZHw5zxZsr%2FmTq4mJDoKqxS5WeL10TRUsGF17KT7Zfa92hwh5%2Boh52HwdCzOzktKLYvIO%2B2AhcBAAA%3D

### Expected behavior

SWC should produce this to match the TypeScript compiler:

```js
fn(x < y, x >= y);
```

### Actual behavior

SWC produces this instead, which is incorrect:

```js
fn(x = y);
```

### Version

1.3.57

### Additional context

Specifically TypeScript's `parseTypeArgumentsInExpression` function backtracks unless the trailing `>` comes from a `>` real token. This is subtle because TypeScript's type argument parser normally strips off the leading `>` from a token, which is why `>>` works in `a<b<c>>(d)`. But it turns out that should only be done when you're in a type context. When you're in an expression context, the trailing `>` must not come from a token that has anything else after it. So for example `a<b<c>>=(d)` should not be considered a type argument list because `parseTypeArgumentsInExpression` encounters a `>=` token.

I just fixed this bug in esbuild: https://github.com/evanw/esbuild/issues/3111. I'm reporting it here because SWC appears to also have this issue. The bug does not introduce a syntax error, which means the problematic output might be missed before releasing it, so I considered this bug to be higher severity.

Austaras:
**Description:**

**BREAKING CHANGE:**

closes #7387


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_parser

PR #7390, opened 2023-05-14
feat(swc_core): allow native env plugin to use memory cache

kwonoj:
**Description:**

I just realized there could be some case who'll want plugin in native env, but without filesystem cache. If there's a custom cache implementation, it'll make a conflict to swc's caching mechanism since swc does not expose any interface to the its cache. 

PR takes simple approach to expose another feature to opt in native env with memory cache for those case. Ideally we should make `plugin_transform_host_native` to not to opt-in any cache, and then make `swc_plugin_runner` to run without any cache implementation - but that'll be a breaking changes with few more involved changes.


kdy1 (approved):
swc-bump:
 - swc_core

PR #7391, opened 2023-05-14
feat(swc_core): expose plugin proxy to the host env

kwonoj:
**Description:**

There are some host-specific interfaces in plugin proxy to setup runners.


kdy1 (approved):
swc-bump:
 - swc_core
 - swc_plugin_macro

PR #7401, opened 2023-05-16
fix(common): Disable `tracing/release_max_level_info`

kdy1:


kdy1 (commented):
swc-bump:
 - swc_nodejs_common

PR #7419, opened 2023-05-19
refactor(plugin/runner): Add attributes to the module bytes

kwonoj:
**Description:**

Seems Turbopack's Vc requires some attributes to wrap into, attempt to support it.


kdy1 (approved):
swc-bump:
 - swc_plugin_runner

PR #7420, opened 2023-05-19
feat(plugin/runner): Enable in-memory precompiled module cache

kwonoj:
**Description:**

Finally we can enable this again. Locally tested against one of my plugin (swc-coverage-instrument) and it seems working, and shows noticeable performance boost when processing multiple files with same plugin in a single process.


kdy1 (approved):
swc-bump:
 - swc_plugin_runner

PR #7422, opened 2023-05-20
refactor(swc_core): Make `common_plugin_transform` agnostic to mode

kwonoj:
**Description:**

PR moves out mode-specific features (`plugin-mode`) so `common-plugin-transform` can be used in either plugin host or plugin transforms.


kdy1 (approved):
swc-bump:
 - swc_core

PR #7427, opened 2023-05-22
fix(core): correctly expose plugin with host

kwonoj:
**Description:**

Following up https://github.com/swc-project/swc/pull/7422, `swc_core::plugin` itself is not exposed to `__plugin_transform_host` so still not able to access inner plugin proxy.


kdy1 (approved):
swc-bump:
 - swc_core

PR #7436, opened 2023-05-23
fix(es): Init filesystem cache only if plugin exists

kwonoj:
**Description:**
Minor fix to regressions for filesystem cache init timing.


kdy1 (approved):
swc-bump:
 - swc

kdy1 (approved):
swc-bump:
 - swc

PR #7439, opened 2023-05-24
fix(es/parser): Re-lex `<<` as two `<`-s if required

### LINKED ISSUE #7187: Parser error for type parameters containing a generic function signature
### Describe the bug

This code appears in a d.ts file on DefinitelyTyped, but fails to parse.

### Input code

```typescript
assertType<<K>(key: K) => K>(foo);
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "target": "es2022",
    "transform": {
      "useDefineForClassFields": true,
      "react": {
        "importSource": "https://esm.sh/preact",
        "runtime": "automatic"
      }
    },
    "loose": false
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.3.44&code=H4sIAAAAAAAAA0ssLk4tKgmpLEi1sfG208hOrbRS8NZUsLVTAPLS8vM1rQGKcDutIgAAAA%3D%3D&config=H4sIAAAAAAAAA1WQvW7DMAyE9z6FoTmoCw8ZsrbIlilPQCh0w0J%2FICkgRuB3jyTHcbqJ3%2FF04N0%2Fus78iTWH7l6eZUjAgvyaC5EpKNwKMTolFMuU1OxWVaVKyhkbmRfBKPAvajWhDF%2FDYFbOEGSM7N8TsuAPjhTwGPnbgciR0F3k%2Be8rihGsvvkKIp8i6zlmtljDrqpJDn2P4j%2Fl2qfFstsMnIOSb7uQNXpQsuYpz%2F8ucDFK3RvBSb2tcePjJTvc6qqVLFfuzbZEgcZp9TZEclqdrav5AeQIxLB6AQAA

### Expected behavior

Parses.

### Actual behavior

Doesn't parse!

### Version

1.3.44

### Additional context

_No response_

kdy1:
**Description:**

**Related issue:**

 - Closes #7187.


sandersn:
I looked at this and found the source of the bug, but I'm not sure how to fix it. It's in expr.rs:691, near the end of `parse_member_expr_new_expr_inner`. For the example, `assertType<<K ...`, `<<` gets lexed as `LShift` not `<`,`<` because the context has `should_not_lex_lt_or_gt_as_type` set (and maybe `!in_type`, I'm not sure). 

Problem is: this context is correct at the time. The lexing actually happens in `parse_primary_expr`, which *should* not be in a type context. But by the time `parse_primary_expr` returns, the lexer has already read `LShift`, so on line 691, `let type_args = if self.syntax().typescript() && is!(self, '<') {` will fail because it's `LShift` not `<`.

The Typescript parser relies on contexts less heavily than swc, and its lexer has a `reScanLessThanToken` method that it calls at this point; the lexer backs up 1 character and changes its token to `<` if the current token is `LShift`. However, I don't know if swc has this capability. It's possible that you can change the context in order to cause rescanning of the current token, but my initial reading of the code says No.

@kdy1 could you advise on whether rescanning is a possibility in the swc lexer?

A third option would be massive lookahead, in which `parse_primary_expr` is parsed in a context with `should_not_lex_lt_or_gt_as_type` set to `false`, but I expect this would be expensive.

kdy1 (commented):
swc-bump:
 - swc_ecma_parser

PR #7441, opened 2023-05-24
refactor(es/ast): Reimplement optional chaining

### LINKED ISSUE #7003: Invalid emitted code when using optional chaining twice.
### Describe the bug

I found that code using [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) twice don't work as expected.

Given the following snippet:
```typescript
const a = ["hello", "world"];

a.filter?.((a) => true).map?.((a) => console.log(a));
```
I expect after running `scw --no-swcrc reproducer.ts -o reproducer.js && node reproducer.js` to have:
```
hello
world
```
As an output.
But the code generated don't work as expected.
```javascript
var _a_filter, _a_filter_map, _a_filter1;
var a = [
    "hello",
    "world"
];
(_a_filter_map = _a_filter = (_a_filter1 = a.filter) === null || _a_filter1 === void 0 ? void 0 : _a_filter1.call(a, function(a) {
    return true;
}).map) === null || _a_filter_map === void 0 ? void 0 : _a_filter_map.call(_a_filter, function(a) {
    return console.log(a);
});
```
If I use `tsc` I get:

```js
"use strict";
var _a, _b, _c;
const a = ["hello", "world"];
(_c = (_a = a.filter) === null || _a === void 0 ? void 0 : (_b = _a.call(a, (a) => true)).map) === null || _c === void 0 ? void 0 : _c.call(_b, (a) => console.log(a));
```
That works as expected.

### Input code

```typescript
const a = ["hello", "world"];

a.filter?.((a) => true).map?.((a) => console.log(a));
```

### Config

_No response_

### Playground link

https://play.swc.rs/?version=1.3.37&code=H4sIAAAAAAAAA0vOzysuUUhUsFWIVspIzcnJV9JRUCrPL8pJUYq15uJK1EvLzClJLbLX09BI1FSwtVMoKSpN1dTLTSxACCUDzcjPSdXLyU8HimhaAwCkjiqEVAAAAA%3D%3D&config=H4sIAAAAAAAAA0WNSwrDMAxE76J1tt3kDj2ESZXg4B8aBWqM7165OGQnzedNoxMbrY2KE7CMCzWp%2B9JKWgtjE1%2BUFlKYtLsA7vY4OVgtwniZF3IGT3eh6JPf6yBtORZh4LFcOsKd7AaK%2BXMNof3HDGiVmNMJ6g9olj3eM61ycf8BnwbyVboAAAA%3D

### Expected behavior

I expect after running `scw --no-swcrc reproducer.ts -o reproducer.js && node reproducer.js` to have:
```
hello
world
```

### Actual behavior

No output

### Version

1.3.37

### Additional context

_No response_

### LINKED ISSUE #7156: Optional chaining with nested property accessors produces invalid code
### Describe the bug

When compressing optional chains on values that won't be present at runtime, the identifier is replaced by `(void 0)` but this fails at runtime when there are nested accessors. 

### Input code

```typescript
let missing = null;

console.log(missing?.thing.name ?? 'unknown')
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2022",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": true,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true,
        "const_to_let": true,
        "pristine_globals": true
      },
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link

https://play.swc.rs/?version=1.3.42&code=H4sIAAAAAAAAA8tJLVHIzSwuzsxLV7BVyCvNybHm4krOzyvOz0nVy8lP14BK2uuVZAApvbzE3FQFe3sF9dK87Lz88jx1TQCsW9wjQgAAAA%3D%3D&config=H4sIAAAAAAAAA32UO3LjMAyG%2B5zCozrFjostcoDtcgYOTYIyvXxoCNCxJuO7L0TJj40hdRI%2B%2FAAJgPh%2B2%2B26E5ruY%2FfNn%2Fwz6IJQ7v9swTGRvrClAxM1muIH6t5v9IQTcjogNNN1Jh3p0gM1Fe5%2F7feLogs5I9wUiy365N34nNPkOBRAfLKxlUPWCInwf%2F3CSv6aAJX6bD%2FkHECnDaI0Kp8IeihSYJND0AOCOusiRJlOqovHLKWYYCWwaih5EHmynnxOnPOVWtBWmWxBQL6AIX8GSca5WJaQryfcp2ELh9r3rc8%2F1HDWoWoScsKltYRPK0Q9Zo%2BkXE1SCWe4UoMZLsX9qfROFaBa0qvulH1a6clfAK5A0IhJR5DiNg%2FH87SmdptKnxyPLI0C5%2FmWbpmg56Iq751Q2akyUMhL3Sxgq4GpskY6zoJXyofeggLneFaE0PjlyRylpDQOkJ0AuL%2FaSVM1A3V%2FhSt8ehAb%2BA%2FfkuQBWzyipuM6xTEecthIEIGO2W44cCsor%2BPCW%2BIyrPOaLPBogBVdKjbwugT4AVBWoe3Ll9ng58ERVR%2Fy4bEmFofrfQ9HnfrHe59X8dvi0MVsa4PLkp%2F6O6%2Fm393D6baF7wfvPH7elC3p9R%2BoYzQdMAYAAA%3D%3D

### Expected behavior

```js
console.log((void 0) ?? 'unknown')

// or potentially

console.log('unknown')
```

### Actual behavior

The nested property accessor is retained in the final code, where it errors trying to access a missing value on `(void 0)`.

```js
console.log((void 0).name ?? 'unknown');
```

### Version

1.3.42

### Additional context

_No response_

kdy1:
**Description:**

**Related issue:**

 - Closes #7003.
 - Closes #7156.


kdy1:
@magic-akari I think I can't debug the current implementation, so I decided to reimplement it. Do you think it's the correct way?

magic-akari:
> @magic-akari I think I can't debug the current implementation, so I decided to reimplement it. Do you think it's the correct way?

I agree.

Austaras:
Well, here's my previous investigation.

The `optional chain` pass is fundamentally broken and there's no easy way out.

Imagine writing a recursive descendant AST evaluator and use that to eval `a?.b.c` which in swc AST would be `Member(OptionalChain(a, b), c)`. We should run `eval_member_expr` first, which can be sketched as
```rust
let obj = obj.eval()
let key = key.eval()

index_obj_with_key()
```
However, if `a` evaluates to `null`, we cannot go through such process -- the evaluation of object must short circuit, it must not get through normal member expression eval. So unless there's some non local control flow, the recursive part won't hold.

So as a conclusion, for `a?.b.c`, we must know at the top level that this is in an optional chain. Babel, TSC and other compilers do the same thing, basically they have a flag for all member expression to show if it's in an optional chain. And I have found another way -- parse `a?.b.c` as `Optional(a, [Member b, Member c])`. Anyway, a AST change would be needed.

jridgewell:
Babel represents optional chains as a different type of `MemberExpression` node, an `OptionalMemberExpression`. The `a?.b` in `a?.b.c` means that we have 2 `OptionalMemberExpression`s:

```js
{
  type: "OptionalMemberExpression",
  object: {
    type: "OptionalMemberExpression",
    object: {
      type: "Identifier",
      name: "a"
    },
    computed: false,
    property: {
      type: "Identifier",
      name: "b"
    },
    optional: true
  },
  computed: false,
  property: {
    type: "Identifier",
    name: "c"
  },
  optional: false
}
```

`OptionalMemberExpression`s have an `optional` flag, which means that the access itself uses `?.`. So the `a?.b` is an `OptionaMemberExpression` with `optional: true`, and the `.c` is an `OptionalMemberExpression` with `optional: false`.

2 bits are required to signal the optional and short-circuit behavior. The `optional` flag signals that `a?.b` property access is optional if the `a` is nullish. But you need the second bit to tell that the chained `.c` will be skipped entirely, and that's why the `.c` is also an `OptionalMemberExpression`, the node type is the second bit. And it think this is the flaw with SWC's current AST, you don't have a second bit.

- - -

ESTree's AST is similar to yours, except that the `ChainExpression` is always hoisted to the root node of the chain:

```js
{
  type: "ChainExpression",
  expression: {
    type: "MemberExpression",
    object: {
      type: "MemberExpression",
      object: {
        type: "Identifier",
        name: "a"
      },
      computed: false,
      property: {
        type: "Identifier",
        name: "b"
      },
      optional: true
    },
    computed: false,
    property: {
      type: "Identifier",
      name: "c"
    },
    optional: false
  }
}
```

This represents the 2nd bit with the single `ChainExpression` node at the root of the chain. Notice that `a?.b.c` has a chain around both the `a?.b` and the `.c`, even though the `.c` is above the `a?.b`. The 1st bit is represented with an `optional` flag on all `MemberExpression` and `CallExpression` nodes.

I personally find this AST the be subpar, because now every member carries an optional boolean, and you can't tell in isolation whether this node actually descends from a chain expression.

- - -

A third proposal I've seen is to represent each individual link in the chain as a brand new type. You descend from the root `OptionalChain` into an `object`, which has an array of additional expressions to perform if the base isn't nullish:

```js
{
  type: "OptionalChain",
  object: {
    type: "Identifier",
    name: "a"
  },
  chain: [
    {
      type: "Property",
      computed: false,
      property: "b"
    },
    {
      type: "Property",
      computed: false,
      property: "c"
    }
  ]
}
```

This one's ok, it takes the root node from ESTree but doesn't suffer from mixing `optional` into all members/call expressions. The downside is that it's means you can't easily share code to operate on both `MemberExpression` and `OptionalMemberExpression`, everything has to be rewritten to handle this novel AST shape.

kdy1:
I think the downside of the third approach would be too much for most of the non-core-contributors. But I got the points, and I'll modify AST. Thank you all!

Austaras:
Personally I prefer solution 3, but I agree that the core of the problem lies in we must distinguish normal member expression and member expression in optional chain. And thank you for the detailed explanations!

kdy1 (commented):
swc-bump:
 - swc_ecma_ast --breaking

PR #7445, opened 2023-05-24
feat(es/minifier): Inline for loop variable

Austaras:
**Description:**

**BREAKING CHANGE:**


crates/swc_ecma_minifier/src/compress/optimize/collapse_vars.rs
```diff
@@ -1,73 +1 @@
-use swc_ecma_ast::*;
 
-use super::Optimizer;
```
Austaras, line 3: This pass doesn't have any effect. I believe it's completely shadowed by `inline` pass.

jridgewell (approved):

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_usage_analyzer

PR #7446, opened 2023-05-24
fix: ignore input sourcemap error

limerickgds:
**Description:**

User's npm package is missing a .map file. If there is a compilation error, the user will not be able to compile the project.

babel :
https://github.com/babel/babel/blob/master/packages/babel-core/src/transformation/normalize-file.js#L63

**BREAKING CHANGE:**

no


kdy1 (approved):
---


swc-bump:
 - swc

PR #7452, opened 2023-05-26
Fix for minor typo in cache.rs error meesage

EvHaus:
**Description:**

Fixes a minor typo in cache.rs error message which I noticed today.

**BREAKING CHANGE:**

N/A

**Related issue (if exists):** No


kdy1:
We don't accept PRs like this

PR #7458, opened 2023-05-27
fix(plugin/runner): Disable `wasi_env` cleanup

kwonoj:
**Description:**

Running cleanup for wasi_env in @swc/core was redundant and harmless since we don't have multi threaded plugin and do not need to teardown files in the host. However in some environment (turbopack, notably) it looks like it causes collision with turbopack's runtime setup vs. wasmer's runtime setup using __asyncify.

For now simply disables it and try to revisit once we have multithreaded plugin support.


kwonoj:
```
thread 'env_entry_chrome_71' panicked at 'failed to parse json: Error("unknown variant `transform-unicode-sets-regex`, there are no variants", line: 2, column: 32)'
```

doesn't seem to be caused by this PR? 🤔

kdy1 (approved):
swc-bump:
 - swc_plugin_runner

PR #7460, opened 2023-05-29
feat(serialized): deserialize with bytechecked

kwonoj:
**Description:**

Note: this PR currently **fails**, description's to explain those and discuss how to move forward.

This PR replaces current `deserialize` from unsafe, non-bytechecked to bytechecked version. This truly enables bytechecked operation since now we actually deserializes with bytechecked one. Enabling this revelas few place doesn't mark bytecheck attributes, addresses it as well.

However, this PR fails to build with this error message: 

```
error[E0275]: overflow evaluating the requirement `swc_ecma_ast::stmt::ArchivedBlockStmt: bytecheck::CheckBytes<rkyv::validation::validators::DefaultValidator<'_>>`

error[E0275]: overflow evaluating the requirement `ArchivedComponentValue: bytecheck::CheckBytes<rkyv::validation::validators::DefaultValidator<'_>>`
```

There is an upstream issue explains this already: https://github.com/rkyv/rkyv/issues/214 , and it _looks_ like those 2 struct s(`BlockStmt` and `ComponentValue`) may include circular references in its structs somehow. for example `BlockStmt -> Vec<Stmt> -> Stmt::Block(BlockStmt)`.

 If this is true, this may explain some of esoteric failures when it gets to deserialization fail (or unexpected behaviors).

Now the question is how to resolve this. Quick peeking, it doesn't seem to obvious I can amend existing AST to conform the requirement.

### DIFF
--- a/crates/swc_common/src/plugin/serialized.rs
+++ b/crates/swc_common/src/plugin/serialized.rs
@@ -1,8 +1,4 @@
-use std::any::type_name;
-
 use anyhow::Error;
-#[cfg(feature = "__rkyv")]
-use rkyv::Deserialize;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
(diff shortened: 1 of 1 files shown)


kdy1:
Closing as already implemented

PR #7463, opened 2023-05-29
Fix comment typo (Serailize -> Serialize)

rivertam:
**Description:**

I noticed a typo in the documentation, and this fixes it!


kdy1:

rivertam:
Ok... why? Is that documented somewhere e.g. the contributing readme?

PR #7478, opened 2023-06-01
feat(es/minifier): Remove unused label

Austaras:
**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc/tests/tsc-references/doWhileBreakStatements.2.minified.js
+++ b/crates/swc/tests/tsc-references/doWhileBreakStatements.2.minified.js
@@ -1,4 +1,2 @@
 //// [doWhileBreakStatements.ts]
-FOUR: for(;;)FIVE: for(;;)break FOUR;
 for(;;);
-for(;;)for(;;);
--- a/crates/swc_ecma_minifier/src/compress/optimize/util.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/util.rs
@@ -4,7 +4,7 @@ use std::{
 };
 
 use rustc_hash::{FxHashMap, FxHashSet};
-use swc_atoms::js_word;
+use swc_atoms::{js_word, JsWord};
 use swc_common::{util::take::Take, Span, DUMMY_SP};
 use swc_ecma_ast::*;
 use swc_ecma_transforms_base::perf::{Parallel, ParallelExt};
@@ -100,6 +100,39 @@ impl<'b> Optimizer<'b> {
         WithCtx {
             reducer: self,
             orig_ctx,
+        }
+    }
+
+    pub(super) fn try_remove_label(&mut self, s: &mut LabeledStmt) {
+        if !self.options.dead_code {
+            return;
+        }
+
+        let mut analyer = LabelAnalyzer {
+            label: s.label.sym.clone(),
+            ..Default::default()
+        };
+
+        match &mut *s.body {
+            Stmt::For(ForStmt { body, .. })
+            | Stmt::ForIn(ForInStmt { body, .. })
+            | Stmt::ForOf(ForOfStmt { body, .. })
+            | Stmt::While(WhileStmt { body, .. })
+            | Stmt::DoWhile(DoWhileStmt { body, .. }) => {
+                analyer.top_breakable = true;
+                body.visit_mut_with(&mut analyer)
+            }
+            Stmt::Switch(SwitchStmt { cases, .. }) => {
+                analyer.top_breakable = true;
+                cases.visit_mut_with(&mut analyer)
+            }
+            _ => s.body.visit_mut_with(&mut analyer),
+        };
+
+        if analyer.count == 0 {
+            let _label = s.label.take();
+            self.changed = true;
+            report_change!("Removing label `{}`", _label);
         }
     }
 }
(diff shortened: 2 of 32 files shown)


crates/swc/tests/tsc-references/doWhileBreakStatements.2.minified.js
```diff
@@ -1,4 +1,2 @@
 //// [doWhileBreakStatements.ts]
-FOUR: for(;;)FIVE: for(;;)break FOUR;
 for(;;);
-for(;;)for(;;);
```
Austaras, line 4: Original implementation would transform
```js
SEVEN: do do do break SEVEN;
while (true);
while (true);
while (true);
```
into
```js
for(;;)for(;;);
```
but it's actually a loop that immediately break.

kdy1 (commented):
Thank you!
Test changes look good, but I think I need to run benchmark before merging this, as visitors sometimes explode.

---

swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/src/compress/optimize/util.rs
```diff
@@ -567,3 +600,86 @@ impl Drop for SynthesizedStmts {
...
+
+        n.visit_mut_children_with(self);
+
+        self.break_layer -= 1;
+    }
+
+    fn visit_mut_break_stmt(&mut self, n: &mut BreakStmt) {
+        if let Some(lb) = &n.label {
+            if lb.sym == self.label {
+                if self.break_layer > 0 || !self.top_can_break {
```
jridgewell, line 665: What is the use of `top_can_break`? I can't think of a situation where it will be necessary.

jridgewell (dismissed):

crates/swc_ecma_minifier/src/compress/optimize/util.rs
Austaras: If top is a noraml break, we must preserve that labelled break because we don't have another control flow statement in JS; if it is, that label can be simply removed. So
```js
a: for (let i = 0; i < 10; i++) {
  if (i % 5 === 0) break a
  console.log(i)
}
```
can be turned into
```js
for (let i = 0; i < 10; i++) {
  if (i % 5 === 0) break
  console.log(i)
}
```

jridgewell: While replying to this, I realized that you can add labels to non-loop/switch:

```js
FOO: {
  break FOO;
}
```

Here, removing `FOO` leaves `{ break }`, which is invalid.

Can we add a comment?

crates/swc_ecma_minifier/src/compress/optimize/util.rs
```diff
@@ -102,6 +102,39 @@ impl<'b> Optimizer<'b> {
...
+            | Stmt::While(WhileStmt { body, .. })
+            | Stmt::DoWhile(DoWhileStmt { body, .. }) => {
+                analyer.top_breakable = true;
+                body.visit_mut_with(&mut analyer)
+            }
+            Stmt::Switch(SwitchStmt { cases, .. }) => {
+                analyer.top_breakable = true;
+                cases.visit_mut_with(&mut analyer)
+            }
+            _ => s.body.visit_mut_with(&mut analyer),
```
jridgewell, line 129: Actually thinking about it some more, do we even need to run this if we're not a loop/switch? We can't remove `break` statements, and we know a `continue` statement can't exist. Can we just early return here and remove `top_breakable` entirely?

crates/swc_ecma_minifier/src/compress/optimize/util.rs
```diff
@@ -567,3 +600,87 @@ impl Drop for SynthesizedStmts {
...
+    break_layer: usize,
+    continue_layer: usize,
+}
+
+impl LabelAnalyzer {
+    fn visit_mut_loop(&mut self, n: &mut impl VisitMutWith<LabelAnalyzer>) {
+        self.break_layer += 1;
+        self.continue_layer += 1;
+
+        n.visit_mut_children_with(self);
```
jridgewell, line 619: Similarly, I think we can skip this? Any `break FOO` we find under here can't be removed (same for continue).

crates/swc_ecma_minifier/src/compress/optimize/util.rs
```diff
@@ -567,3 +600,87 @@ impl Drop for SynthesizedStmts {
...
+    }
+
+    fn visit_mut_do_while_stmt(&mut self, n: &mut DoWhileStmt) {
+        self.visit_mut_loop(n)
+    }
+
+    fn visit_mut_switch_stmt(&mut self, n: &mut SwitchStmt) {
+        self.break_layer += 1;
+
+        n.visit_mut_children_with(self);
```
jridgewell, line 658: This is still needed, since a `continue FOO` could be removed.

crates/swc_ecma_minifier/src/compress/optimize/util.rs
Austaras: We still need to find out if this label is ever used.

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_minifier

PR #7481, opened 2023-06-02
chore: publish npm packages with provenance option

Brooooooklyn:


kdy1 (commented):
Thank you! Let's wait for CI...

---

swc-bump:
 - dbg-swc

kdy1:
@Brooooooklyn It seems like this PR makes CI not create GitHub releases. Can you take a look?

PR #7482, opened 2023-06-02
perf(bindings): Enable `share-generics` to reduce binary size

kdy1:
**Description:**

**Related issue:**

 - https://github.com/vercel/next.js/pull/50673


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #7490, opened 2023-06-05
perf(es): Use `&dyn Comments` for jsx and TypeScript

kdy1:


kdy1 (commented):
swc-bump:
 - swc

PR #7504, opened 2023-06-07
feat(plugin_runner): support shared wasix runtime

kwonoj:
**Description:**
ref: https://github.com/wasmerio/wasmer/discussions/3966#discussioncomment-6107851

If the host of the plugin runner have an initialized runtime, wasix creates a conflict as default runtime attempt to create a new one. This PR exposes a new compile time flag to indicate to use shared one if available. This could be a runtime flag, but did not want to bring tokio as direct dep when it's unused..


kdy1 (approved):
swc-bump:
 - swc_plugin_runner

PR #7514, opened 2023-06-10
fix(es/react): Default to empty string when emitting refresh signature

### LINKED ISSUE #7471: react refresh transform causes panic when used with a plugin which produces nodes with dummy spans
### Describe the bug

Here is reproduction repo. https://github.com/behnamazimi/lingui-swc-plugin-issue-58

`yarn && yarn dev` then open a browser, or change the file. 

It uses `@lingui/swc-plugin` which is transforming nodes and some nodes may have a Dummy Spans (because there are nothing to map to). 
It seems that `/swc_ecma_transforms_react-0.172.14/src/refresh/mod.rs` is trying to process these nodes in some circumstances and  doesn't expect to receive node without a source mapping. 

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: DummyBytePos', /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/swc_ecma_transforms_react-0.172.14/src/refresh/mod.rs:159:74
```

### Input code

_No response_

### Config

_No response_

### Playground link

_No response_

### Expected behavior

should not panic

### Actual behavior

_No response_

### Version

next@13.3.2

### Additional context

_No response_

Austaras:
**Description:**

**BREAKING CHANGE:**

Closes #7471


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_transforms_react

PR #7515, opened 2023-06-11
fix(es/parser): Disallow tagged tpl in optional chaining

g-plane:
**Description:**

Reject the code below:

```js
a?.b``
```

- [Babel REPL](https://babel.dev/repl#?browsers=&build=&builtIns=usage&corejs=3.21&spec=false&loose=false&code_lz=IYfgdARgBlQ&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=script&lineWrap=true&presets=env%2Creact%2Ctypescript&prettier=false&targets=&version=7.22.5&externalPlugins=&assumptions=%7B%7D)
- [TypeScript playground](https://www.typescriptlang.org/play?#code/IYfgdARgBlQ)


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_parser

Issue #7500, opened 2023-06-07
Minifier outputs invalid `1=1=1=0` assignment


mischnic:
### Describe the bug

The minified output is invalid JS.

### Input code

```typescript
var globalArray = [1, 1, 1];
module.exports = function () {
    var localArray = globalArray;
    localArray[0] = localArray[1] = localArray[2] = 0;
    return localArray;
};
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": true,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true,
        "const_to_let": true,
        "pristine_globals": true
      },
      "mangle": {
        "toplevel": false,
        "keep_classnames": false,
        "keep_fnames": false,
        "keep_private_props": false,
        "ie8": false,
        "safari10": false
      }
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": true,
  "isModule": true
}
```

### Playground link

[...]

PR #7517, opened 2023-06-12
fix(es/minifier): Fix a bug breaking `plotly.js`

### LINKED ISSUE #7512: Chained-Array Index Assignment does not minify correctly
### Describe the bug

[Plotly.JS](https://github.com/plotly/plotly.js) has a function that performs chained array index assignment seen [here](https://raw.githubusercontent.com/plotly/plotly.js/master/dist/plotly-gl2d.js) at line 66282
image

When building Next.JS with the [Plotly React component](https://github.com/plotly/react-plotly.js/) (which uses Plot.JS (this has been narrowed down)) locally (minimization, but no minification), the output is a proper result (chained assignment)
image

However, when building the same Next.JS code in a production environment (we're using [Render](https://render.com/)) which minimizes AND minifies (removing whitespace), SWC skips a variable declaration, causing a syntax error in assignment
image

### Input code

_No response_

### Config

_No response_

### Playground link

_No response_

### Expected behavior

The code should minify properly, going to shortvariable `j` for the array instead of skipping and omitting it and using `j` for the next one.

### Actual behavior

In the eyes of the compiler, the array variable doesn't exist and all references to it get replaced with void, leaving the `j` shortvariable to be used for the next variable

### Version

13.4.4

### Additional context

This project is a Next.JS project and was generated using [T3 Stack](https://create.t3.gg/) and utilizes the [React Plotly.JS component](https://github.com/plotly/react-plotly.js/)

This issue has been narrowed down to SWC as when in the [Next.JS config](https://nextjs.org/docs/pages/api-reference/next-config-js/webpack), Webpack's [optimization: minimize](https://webpack.js.org/configuration/optimization/#optimizationminimize) config option is set to `false`, this does not occur.

This issue also does not occur when built on a local machine (I'm not 100% sure why, but it seems to minimize without minifying while in the Render environment, it minimizes AND minifies)

kdy1:
**Description:**

**Related issue:**

 - Closes #7512.


crates/swc_ecma_minifier/src/compress/optimize/inline.rs
```diff
@@ -209,6 +209,8 @@ impl Optimizer<'_> {
                                 )
                             }
 
+                            should_inline &= !usage.has_property_mutation;
```
kdy1, line 212: @Austaras We need a way to prevent the inline of variables initialized with arrays, I guess, but this seems to cause too much regression. Any ideas?

Austaras: Wait, wasn't this effectively the same as #7500?

kdy1: Oh... I'll check.

PR #7522, opened 2023-06-13
fix(es/codegen): Emit type arguments of jsx element names

liaoyinglong:
**Description:**

The ECMA code generation implementation was not emitting type arguments for jsx. This PR fixes that.


kdy1 (approved):
Thank you!


---

swc-bump:
 - swc_ecma_codegen

PR #7524, opened 2023-06-13
chore(ci): Verify serde_json

kdy1:
**Description:**

It seems like `serde_json` has a bug in some platforms. Let's report it with a repro/log.


kdy1:
Closing as obsolete

PR #7528, opened 2023-06-14
feat(es/module): Preserve custom `use` directives

### LINKED ISSUE #7315: Directive not preserved for commonjs
### Describe the bug

```
// input.ts
"use client"

export function ReactClientComponent() {
    return "Hello world"
}
```

`swc input.ts -o output.js -C module.type=commonjs`

```
// output.js
"use strict";
    value: true
});
Object.defineProperty(exports, "ReactClientComponent", {
    enumerable: true,
    get: function() {
        return ReactClientComponent;
    }
});
"use client";
function ReactClientComponent() {
    return "Hello world";
}
```

Notice how `"use client"` is underneath the helpers generated by swc, which doesn't work with Next.js, which expects `"use client"` to come before any expressions.

Related to https://github.com/swc-project/swc/issues/6508

### Input code

_No response_

### Config

_No response_

### Playground link

_No response_

### Expected behavior

`"use client"` should be either 1 line above or below `"use strict"`. I believe either is acceptable for Next.js.

```
// output.js
"use client";
"use strict";
    value: true
});
Object.defineProperty(exports, "ReactClientComponent", {
    enumerable: true,
    get: function() {
        return ReactClientComponent;
    }
});
function ReactClientComponent() {
    return "Hello world";
}
```

### Actual behavior

_No response_

### Version

@swc/cli: 0.1.62 @swc/core: 1.3.53

### Additional context

_No response_

kdy1:
**Description:**

**Related issue:**

 - Closes #7315.


kdy1 (commented):
swc-bump:
 - swc_ecma_utils

crates/swc/tests/fixture/issues-7xxx/7315/1/output/index.js
```diff
@@ -0,0 +1,14 @@
+// input.ts
+"use client";
```
magic-akari, line 2: I believe that `"use strict"` is also necessary.

kdy1: Does it have any effect? Hmm

kdy1: AFAIK, The only first directive has an effect.

magic-akari: continuous directives will be recognized in Babel.

https://github.com/babel/babel/blob/980b63c829de369ace21461d3268d04bb80fc380/packages/babel-parser/test/fixtures/estree/directives/program-order/input.js

magic-akari: And all valid directive will be parsed as directive before first `non directive`.

https://github.com/babel/babel/blob/980b63c829de369ace21461d3268d04bb80fc380/packages/babel-parser/src/parser/statement.ts#L1403-L1430

crates/swc_ecma_transforms_module/src/amd.rs
```diff
@@ -145,7 +145,7 @@ where
 
         // "use strict";
         if self.config.strict_mode {
-            stmts.push(clone_first_use_strict(n).unwrap_or_else(use_strict));
+            stmts.push(clone_first_use_directive(n).unwrap_or_else(use_strict));
         }
```
ammmze, line 149: Maybe I'm not understanding what this is doing, but at first glance, does this mean that if an input file has a directive, it would only get hoisted to the top of the file if `strictMode` is enabled? I'm using strict mode, so it doesn't really affect me, but I imagine this would be a bug that may present itself in the future.

ammmze: I guess I must be mis-understanding. I just finished upgrading and tried it out setting strictMode to false and the directive still makes its way to the top. So you can ignore me...it is working as it should.

PR #7543, opened 2023-06-19
fix(es): Accept `parse` option for `minify()`

### LINKED ISSUE #7455: unknown field `parse`, expected one of `compress`, ...
https://swc.rs/docs/configuration/minification#replacing-terser implies that the swc minifier tries to by fully API-compatible with Terser.

Terser supports a `parse` options property: https://github.com/terser/terser#parse-options

But swc doesn't, leading to
```
unknown field `parse`, expected one of `compress`, `mangle`, `format`, `ecma`, `keep_classnames`, `keepClassnames`, 
`keep_fnames`, `keepFnames`, `module`, `safari10`, `toplevel`, `sourceMap`, `outputPath`, `inlineSourcesContent`, `emitSourceMapColumns`
```

Is this supposed to be allowed or should Parcel filter that out?

### Context

Parcel invokes the swc minifier here: https://github.com/parcel-bundler/parcel/blob/dd9435be8afed35c5ffc161cf4b586fd6c78fc1f/packages/optimizers/swc/src/SwcOptimizer.js#L41

(This was reported to Parcel in https://github.com/parcel-bundler/parcel/issues/9041)

kdy1:
**Description:**

We don't support them. but this is to make API more compatible with `terser`.

**Related issue:**

 - Closes #7455.


kdy1 (commented):
swc-bump:
 - swc

PR #7548, opened 2023-06-20
fix(es/preset-env): Don't log `Yield` to the console

cjwilsontech:
**Description:**

When a `yield` token is encountered by the parser, the word "Yield" is unnecessarily logged to the console. This PR removes the extra logging.


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_preset_env

PR #7551, opened 2023-06-20
feat(es/codegen): Support `preamble`

### LINKED ISSUE #7475: Support the format preamble format option in swc.minify
### Describe the feature

See https://swc.rs/docs/configuration/minification#jscminifyformat

> `preamble`, Currently noop

It should be possible to pass `preamble`. SWC should be able to insert it to the minified file without breaking source map associations.

_No response_

### Additional context

FWIW, jQuery is currently switching from UglifyJS to Terser in https://github.com/jquery/jquery/pull/5258. We considered switching to SWC instead as it's way faster and optimizes better than Terser but the lack of the `preamble` option is a blocker for us at the moment.

kdy1:
**Description:**

**Related issue:**

 - Closes #7475.


kdy1 (commented):
swc-bump:
 - swc_ecma_codegen

PR #7552, opened 2023-06-20
feat(testing): Add `CARGO_TARGET_DIR` and use it from other crates

kdy1:
**Description:**

This is required to avoid creating lots of `target` directory where the user is testing with execution tester.


kdy1 (commented):
swc-bump:
 - testing

kdy1 (commented):
reset ci

PR #7585, opened 2023-06-26
fix(es/fixer): Don't change default decls to default expr exports

### LINKED ISSUE #7584: SimplifyExpr & Fixer should skip ParenExpr at ExportDefaultExpr
### Describe the bug

SimplifyExpr & Fixer should skip ParenExpr at ExportDefaultExpr

### Input code

```typescript
import someFunction from './b';

export default (function someFunction () {
});
```

### Expected behavior

Transform by SimplifyExpr & Fixer should be the origin code.

``` js
import someFunction from './b';

export default (function someFunction () {
});
```

### Actual behavior

But print the below code, the semantics has changed.
``` js
import someFunction from './b';

export default function someFunction () {
};
```

### Version

master code

underfin:
**Description:**

**Related issue:**

 - Closes #7584.

### DIFF
--- a/crates/swc_ecma_transforms_base/src/fixer.rs
+++ b/crates/swc_ecma_transforms_base/src/fixer.rs
@@ -438,6 +438,10 @@ impl VisitMut for Fixer<'_> {
     }
 
     fn visit_mut_export_default_expr(&mut self, node: &mut ExportDefaultExpr) {
+        if let Expr::Paren(p) = &mut *node.expr {
+            p.visit_mut_children_with(self);
+            return;
+        }
         let old = self.ctx;
         self.ctx = Context::Default;
         node.visit_mut_children_with(self);
--- a/crates/swc_ecma_transforms_base/src/fixer.rs
+++ b/crates/swc_ecma_transforms_base/src/fixer.rs
@@ -439,9 +439,12 @@ impl VisitMut for Fixer<'_> {
 
     fn visit_mut_export_default_expr(&mut self, node: &mut ExportDefaultExpr) {
         if let Expr::Paren(p) = &mut *node.expr {
-            if let Expr::Fn(f) = &mut *p.expr {
-                f.visit_mut_children_with(self);
-                return;
+            match &mut *p.expr {
+                Expr::Fn(..) | Expr::Class(..) => {
+                    p.expr.visit_mut_children_with(self);
+                    return;
+                }
+                _ => {}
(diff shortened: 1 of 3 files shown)


crates/swc_ecma_transforms_optimization/src/simplify/expr/tests.rs
```diff
@@ -1544,3 +1544,8 @@ fn test_es6_features() {
         "function foo() {return `${false}`}",
     );
 }
+
+#[test]
+fn test_export_default_paren_expr() {
+    fold_same("import fn from './b'; export default (function fn () {});");
```
magic-akari, line 1550: Another case should be added

```JavaScript
export default ((foo));
```

and expected output:

```JavaScript
export default foo;
```

magic-akari (changes requested):

crates/swc_ecma_transforms_optimization/src/simplify/expr/tests.rs
underfin: @magic-akari i resolved it, please review it again.

crates/swc_ecma_transforms_base/src/fixer.rs
```diff
@@ -438,6 +438,12 @@ impl VisitMut for Fixer<'_> {
     }
 
     fn visit_mut_export_default_expr(&mut self, node: &mut ExportDefaultExpr) {
+        if let Expr::Paren(p) = &mut *node.expr {
+            if let Expr::Fn(f) = &mut *p.expr {
```
kdy1, line 442: I thnk we need to check for `Expr::Class`, too

kdy1 (changes requested):
Thanks!

---

swc-bump:
 - swc_ecma_transforms_base

crates/swc_ecma_transforms_base/src/fixer.rs
underfin: Thanks a lot !

kdy1 (approved):
Thanks!

underfin:
@kdy1 I don't know update all snapshot with test, can you point me out? Thanks !

kdy1:
`UPDATE=1 cargo test` will do it, but it does not modify rust files.

underfin:
Thanks! I'm so sad because i want to update the snapshot at rust files test ...

PR #7587, opened 2023-06-27
feat(es/minifier): Common sub expression in seq in condition

Austaras:
**Description:**

**BREAKING CHANGE:**


crates/swc_ecma_minifier/src/compress/optimize/conditionals.rs
```diff
@@ -616,6 +616,81 @@ impl Optimizer<'_> {
...
             }
 
+            (Expr::Seq(left), Expr::Seq(right)) => {
+                let last = left.exprs.len().min(right.exprs.len()) - 1;
+
+                let mut idx = 0;
+
+                while idx <= last
+                    && left.exprs[last - idx].eq_ignore_span(&right.exprs[last - idx])
+                    && !matches!(&*left.exprs[last - idx], Expr::Yield(..) | Expr::Fn(..))
```
kdy1, line 626: I think we need `Expr::Arrow`, too.

Austaras: Why? I just copy this from L356, and I don't believe the original is necessary

kdy1: Ah, I see. I just thought `Expr::Arrow `is required because there's `Expr::Fn`

Austaras: Do you remember why the original restriction is written?

crates/swc_ecma_minifier/tests/benches-full/victory.js
```diff
@@ -6526,7 +6526,7 @@
                 remove: function(node) {
                     node.N && (node.N.P = node.P), node.P && (node.P.N = node.N), node.N = node.P = null;
                     var sibling, next, red, parent = node.U, left = node.L, right = node.R;
-                    if (next = left ? right ? RedBlackFirst(right) : left : right, parent ? parent.L === node ? parent.L = next : parent.R = next : this._ = next, left && right ? (red = next.C, next.C = node.C, next.L = left, left.U = next, next !== right ? (parent = next.U, next.U = node.U, node = next.R, parent.L = node, next.R = right, right.U = next) : (next.U = parent, parent = next, node = next.R)) : (red = node.C, node = next), node && (node.U = parent), !red) {
```
kdy1, line 6529: I can't find `right.U = next` from other code

kdy1 (changes requested):
Thank you!

---


swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/tests/benches-full/victory.js
Austaras: Oh, I wrote something completely wrong...

kdy1 (approved):
Thank you!

---


swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/src/compress/optimize/conditionals.rs
```diff
@@ -616,6 +616,87 @@ impl Optimizer<'_> {
...
+
+                let mut idx = 0;
+
+                while idx < min_len
+                    && left.exprs[left_len - idx - 1]
+                        .eq_ignore_span(&right.exprs[right_len - idx - 1])
+                    && !matches!(
+                        &*left.exprs[left_len - idx - 1],
+                        Expr::Yield(..) | Expr::Fn(..)
+                    )
```
jridgewell, line 632: Nit: we could trim both the start and the end of the sequences.

jridgewell (approved):

PR #7595, opened 2023-06-29
fix(xml/codegen): Fix wrong minification of spaces in a self-closing tag

SyMind:
**Description:**

xml codegen with option `minify: true`.

input:
```xml
<foo attributeName="OK" />
```

expected:
```xml
<foo attributeName="OK"/>
```

actual:

```xml
<foo attributeName="OK" />
```

---

We already escape the attribute value in `escape_string` function, so we do not need to check whether it contains a specific string.

### DIFF
--- a/crates/swc_xml_codegen/src/lib.rs
+++ b/crates/swc_xml_codegen/src/lib.rs
@@ -166,22 +166,7 @@ where
         }
 
         if is_void_element {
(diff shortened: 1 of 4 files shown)


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_xml_codegen

PR #7604, opened 2023-06-29
feat(es/optimization): Support `default` imports for const modules

### LINKED ISSUE #7601: Const module default imports
### Describe the feature

Default imports from const modules shouldn't panic. Instead, they should be substituted by the `"default"` key of the const module's definition
```typescript
import something from 'my-mod';
console.log(something);
```
given the following config:
```json
{
  "jsc": {
    "transform": {
      "constModules": {
        "my-mod": { "default": "\"substitution\"" }
      }
    }
  }
}
```
should be transpiled to
```typescript
console.log("substitution");
```
The panic can be found here:
https://github.com/swc-project/swc/blob/be33efaecdd00491ab68d31e65365c447d42d646/crates/swc_ecma_transforms_optimization/src/const_modules.rs#L124

More than happy to open a PR if nobody is working on this

_No response_

### Additional context

[The spec treats default exports as exports with an `[[ExportName]]` of `"default"`](https://tc39.es/ecma262/multipage/ecmascript-language-scripts-and-modules.html#sec-exports)

liaoyinglong:
Fixes #7601

**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_transforms_optimization/src/const_modules.rs
+++ b/crates/swc_ecma_transforms_optimization/src/const_modules.rs
@@ -120,8 +120,18 @@ impl VisitMut for ConstModules {
                             ImportSpecifier::Namespace(ref s) => {
                                 self.scope.namespace.insert(s.local.to_id());
                             }
-                            ImportSpecifier::Default(..) => {
-                                panic!("const_module does not support default import")
+                            ImportSpecifier::Default(ref s) => {
+                                let imported = &s.local.sym;
+                                let default_import_key = JsWord::from("default");
+                                let value =
+                                    entry.get(&default_import_key).cloned().unwrap_or_else(|| {
+                                        panic!(
+                                            "The requested const_module `{}` does not provide \
+                                             default export",
+                                            import.src.value
+                                        )
+                                    });
+                                self.scope.imported.insert(imported.clone(), value);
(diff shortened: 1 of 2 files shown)


crates/swc_ecma_transforms_optimization/src/const_modules.rs
```diff
@@ -120,8 +120,18 @@ impl VisitMut for ConstModules {
                             ImportSpecifier::Namespace(ref s) => {
                                 self.scope.namespace.insert(s.local.to_id());
                             }
-                            ImportSpecifier::Default(..) => {
-                                panic!("const_module does not support default import")
+                            ImportSpecifier::Default(ref s) => {
+                                let imported = &s.local.sym;
+                                let default_import_key = JsWord::from("default");
```
kdy1, line 125:
```suggestion
                                let default_import_key = js_word!("default");
```

kdy1 (changes requested):

crates/swc_ecma_transforms_optimization/src/const_modules.rs
liaoyinglong: updated

kdy1 (approved):
Thanks!

---


swc-bump:

PR #7607, opened 2023-06-29
fix(plugin/runner): Disable wasi env cleanup

kwonoj:
**Description:**

Even after setting shared runtime for the wasi env, `cleanup` still panics. I suspect something internal to cleanup doesn't honor those handle correctly, for now disable and try to see if this works.

As commented, this is not a huge issue as plugin in most cases doesn't have resources to be cleaned up anyway.


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_plugin_runner

Issue #7559, opened 2023-06-21
Private property not transformed inside optional chain when using loose mode


jridgewell:
### Describe the bug

I imagine this broke with the new optional chaining implementation. The `#x` in `this?.y.#x` isn't transformed into the appropriate `this?.y[_x]` access

### Input code

```typescript
class Foo {
  #x;

  test() {
    this?.y.#x
  }
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2021",
    "loose": true,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": false
}
```

### Playground link

https://play.swc.rs/?version=1.3.66&code=H4sIAAAAAAAAA0vOSSwuVnDLz1eo5lJQUK6w5gJSJanFJRqaYBEgJyOz2F6vUk%2B5Asit5aoFAHwm0I8yAAAA&config=H4sIAAAAAAAAA1WPOw7DIBBEe58CbZ0icek75BCIrC0QP7FYCrJ89wAGknTMDDNPe0yMgSIBCzvyMwvPA2EYOjuUbOTv7AAKw0kE6SPceqqoRCvXhNU6rwQiDxvG2qL5Pj9aA7RzhNmOYcdmGWnlmn6RwhkfkKgvD5rhdtP4D5waFIx77TVsp8TkiyprxllF8P3ZiWMdJD17%2Fdo%2BPw6eMFYcAQAA

### Expected behavior

The `y.#x` is transformed into `y[_x]`

```js
var id = 0;
function _class_private_field_loose_key(name) {
    return "__private_" + id++ + "_" + name;
}
var _x = _class_private_field_loose_key("_x");
class Foo {
    test() {
        this?.y[_x];
    }
    constructor(){
        Object.defineProperty(this, _x, {
            writable: true,
            value: void 0
        });
    }
}
```

### Actual behavior

The `#x` remains in the output.

```js
var id = 0;
function _class_private_field_loose_key(name) {
    return "__private_" + id++ + "_" + name;
}
var _x = _class_private_field_loose_key("_x");
class Foo {
    test() {
        this?.y.#x;
    }
    constructor(){
        Object.defineProperty(this, _x, {
            writable: true,
            value: void 0
        });
    }
}
```

### Version

1.3.66

komyg:
[...]

PR #7610, opened 2023-06-30
fix(es/compat): Fix handling of private members in optional chaining pass

### LINKED ISSUE #7561: Private property after an optional chain transforms into nonsense code
### Describe the bug

I think this also broke when we implemented the new transform. Currently, if you have a private property access after an optional chain, the chain will run. The `_memo == null` conditional check happens _before_ the assignment of `_memo`.

### Input code

```typescript
class Foo {
  #x;

  test() {
    this?.y.#x
  }
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2021",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": false
}
```

### Playground link

https://play.swc.rs/?version=1.3.66&code=H4sIAAAAAAAAA0vOSSwuVnDLz1eo5lJQUK6w5gJSJanFJRqaYBEgJyOz2F6vUk%2B5Asit5aoFAHwm0I8yAAAA&config=H4sIAAAAAAAAA1VPOw7DIAzdcwrkuUObMXfoISzqRET8hKlUFOXuBQJps%2Fn95W0QAlaWMIktnxl4DEzhxJnhZCN%2BMgMkDbIMyke4dXXlIs2omSq1HwpEDAvFmuLxPj5aArRzTD3ROKOsmtP%2FpnTGB2K%2BGosV7aLpuji0VTDu9a5i%2ByUmX1BpM86uDD9nXzzbQfGzx4%2Fu%2FQutaQ0NHQEAAA%3D%3D

### Expected behavior

The memo is assigned before the conditional, and the private access is scoped to the conditional's alternate branch:

```js
// …
var _x = new WeakMap();
class Foo {
    test() {
        (_this = this) === null || _this === void 0
          ? void 0
          : _class_private_field_get(_this.y , _x);
    }
    // …
}
```

### Actual behavior

The chain is nonsense:

```js
// …
var _x = new WeakMap();
class Foo {
    test() {
        var _this_y, _this_y1;
        _this_y === null || _this_y === void 0
            ? void 0
            : _class_private_field_get(_this_y1 = this?.y, _x);
    }
    // …
}
```

### Version

1.3.66

### Additional context

_No response_

komyg:
**Description:**

Fix: https://github.com/swc-project/swc/issues/7561

https://github.com/swc-project/swc/issues/7561


crates/swc_ecma_transforms_compat/tests/optional-chaining-private-var/issue-7561/output.js
```diff
@@ -0,0 +1,10 @@
+// …
+var _x = new WeakMap();
+class Foo {
+    test() {
+        (_this = this) === null || _this === void 0
+          ? void 0
+          : _class_private_field_get(_this.y , _x);
```
kdy1, line 7: This is wrong.

```js
class Foo {
  #x;
  test() {
    this === null || this === void 0 ? void 0 : this.y.#x;
  }
}
```

Please try locally using 

```json
{
  "plugins": [
    "@babel/plugin-transform-optional-chaining"
  ]
}
```

This is how I got the expected output

kdy1 (changes requested):

crates/swc_ecma_transforms_compat/tests/optional-chaining-private-var/issue-7561/output.js
komyg: Hum, but on the issue: https://github.com/swc-project/swc/issues/7561, @jridgewell says that this is the expected output:

```ts
// …
var _x = new WeakMap();
class Foo {
    test() {
        (_this = this) === null || _this === void 0
          ? void 0
          : _class_private_field_get(_this.y , _x);
    }
    // …
}
```

Also, your output is not transforming the private field. Shouldn't it be transformed as well?

komyg: Hey @kdy1, 

On the original issue: https://github.com/swc-project/swc/issues/7561, @jridgewell says that this is the expected output:

```ts
// …
var _x = new WeakMap();
class Foo {
    test() {
        (_this = this) === null || _this === void 0
          ? void 0
          : _class_private_field_get(_this.y , _x);
    }
    // …
}
```

Also, isn't your output the same as mine, but without the transformation for the private field?

komyg: PS: I thought that the main problem was that we were not transforming the private field.

kdy1: No. Expected output of `optional-chaining` is mine. What you uploaded is output of full transforms.

komyg: I'm sorry, @kdy1, I don't think I understand your comment.

Here is the problem that I want to solve with this PR (please note that I haven't been able to actually solve it yet, this is why you only saw a test): whenever we have an optional chaining expression with a private field, the private field is not transformed (we don't replace it with: `_class_private_field_get`).

This is what I understood from these issues: https://github.com/swc-project/swc/issues/7561 and https://github.com/swc-project/swc/issues/7559.

Regarding your comment:

> No. Expected output of optional-chaining is mine. What you uploaded is output of full transforms.

Do you mean that I've put the test in the wrong place? Since it is inside the optional chaining module, it shouldn't be testing private fields? If so, where would be the most appropriate place to put it?

Or do you mean that I am solving the wrong problem?

kdy1: > whenever we have an optional chaining expression with a private field, the private field is not transformed

This is wrong. Private fields are transformed by another pass.

kdy1: The optional chaining pass should produce

```js
class Foo {
  #x;
  test() {
    this === null || this === void 0 ? void 0 : this.y.#x;
  }
}
```

and another pass is responsible for the rest.

komyg: Ok, there are two problems?

One for the optional chaining and another for tge private field?

kdy1: No, private field transforms are working correctly

komyg: I've updated the test as you requested.

komyg:
Hey @kdy1, I believe I have solved the problem. Could you review it please?

crates/swc_ecma_transforms_compat/tests/es2020_optional_chaining.rs
```diff
@@ -303,3 +303,16 @@ fn fixture_loose(input: PathBuf) {
         Default::default(),
     );
 }
+
+#[testing::fixture("tests/optional-chaining-private-var/issue-7561/input.js")]
```
kdy1, line 307: Please remove this

kdy1: You don't need to add a new function with `testing::fixture`

kdy1: Just add a test file to the correct directory.

kdy1 (changes requested):

crates/swc_ecma_transforms_compat/tests/es2020_optional_chaining.rs
komyg: I think I've added it to the correct folder.

kdy1 (approved):
Thank you, nice work!

---

swc-bump:
 - swc_ecma_transforms_compat

PR #7638, opened 2023-07-07
fix(es/minifier): Don't drop unused properties of top-level vars

### LINKED ISSUE #7635: Object contents are incorrectly removed on minify
### Describe the bug

When running `minify` / `minifySync` contents of objects declared at the global scope are being removed unless you reference that variable directly elsewhere in the source. If you only use part of the object (reference only some properties), it inlines those properties and drops the entire top-level variable declaration which seems broken.

### Input code

```typescript
const {minifySync} = require('@swc/core');

const options = {compress: true, mangle: true, module: false};
const source = `const index = {foo: 5, bar: 10};`;
const useIndex = `console.log(index);`
const usePartOfIndex = `console.log(index.foo);`

console.log('test1:', minifySync(source, options).code);
console.log('test2:', minifySync(`${source}\n${useIndex}`, options).code);
console.log('test3:', minifySync(`${source}\n${usePartOfIndex}`, options).code);
```

### Config

```json
{
  "compress": true,
  "mangle": true,
}
```

### Playground link

_No response_

### Expected behavior

The output Terser produces in this case matches my expectations. Since we have `module: false`, top-level variables should not be mangled, or removed since they are in the global scope. As a result, each of the 3 test cases has the full object definition for index, and references that variable for the `console.log` statements:

```js
test1: const index={foo:5,bar:10};
test2: const index={foo:5,bar:10};console.log(index);
test3: const index={foo:5,bar:10};console.log(index.foo);
```

### Actual behavior

The code currently produced by SWC varies based on how/if the `index` variable is used ranging from the variable not being declared at all, being fully declared, or being an empty object.

```js
test1: const index={};
test2: const index={foo:5,bar:10};console.log(index);
test3: console.log(5);
```

### Version

1.3.68

### Additional context

The behavior when setting `module: true` is a lot more consistent because in that context the top-level variables are only in module scope and aren't externally visible outside of the module. 

What we get in that case is the following result which matches expectations for when setting `module: true`:
```js
test1: 
test2: console.log({foo:5,bar:10});
test3: console.log(5);
```
or if you `export const index`:
```js
test1: export const index={foo:5,bar:10};
test2: export const index={foo:5,bar:10};console.log(index);
test3: export const index={foo:5,bar:10};console.log(index.foo);
```

kdy1:
**Description:**

 - Closes #7635.

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/optimize/unused.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/unused.rs
@@ -835,6 +835,10 @@ impl Optimizer<'_> {
             return None;
         }
 
+        if self.ctx.top_level && !self.options.top_level() {
+            return None;
+        }
+
         let name = v.name.as_ident()?;
         let obj = v.init.as_mut()?.as_mut_object()?;
 
--- a/crates/swc_ecma_minifier/tests/fixture/projects/wmr/archive-1/chunks/index.bf24abaa/output.js
+++ b/crates/swc_ecma_minifier/tests/fixture/projects/wmr/archive-1/chunks/index.bf24abaa/output.js
@@ -1,5 +1,6 @@
 import { s as style, m } from "../index.f66dda46.js";
 const process = {
+    browser: !0,
     env: {
         FOO: "bar",
         OVERRIDE: "11",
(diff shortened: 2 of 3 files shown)


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

Issue #7641, opened 2023-07-07
Should remove unused recursive function and minify objects 


nkitku:
### Describe the bug

this is not bug but may be feature request, the given code minify to 2407char using swc but terser make it 1851char on https://try.terser.org/

and if we remove word "export" on line 143 terser will only keep import statements but swc give 1674char

### Input code

```typescript
import { readFileSync, createReadStream } from "node:fs";
import { createInterface } from "node:readline";
import { WEB_IFC_SCHEMA_INDEX } from "./schema/ifc-schema.ts";
const IfcTokenType = {
    UNKNOWN: 0,
    STRING: 1,
    LABEL: 2,
    ENUM: 3,
    REAL: 4,
    REF: 5,
    EMPTY: 6,
    SET_BEGIN: 7,
    SET_END: 8,
    LINE_END: 9,
    LIST: 10,
    ENTITY: 11,
};
const parseArgs = (data, i) => {
    if (data.ifc[0] === "(")
        i++;
    const args = [];
    for (; i < data.ifc.length;) {
        const c = data.ifc[i];
        if (c === "'") {
            i++;
            const start = i;
            while (true) {
                if (data.ifc[i] === "'") {
                    i++;
                    if (data.ifc[i] !== "'") {
                        i--;
                        break;
                    }
                }
                i++;
            }
            const end = i;
            args.push(IfcTokenType.STRING, data.ifc.slice(start, end));
            // args.push([IfcTokenType.STRING, data.ifc.slice(start, end)]);
        }
        else if (c === "#") {
            i++;
            const start = i;
            while (data.ifc[i] >= "0" && data.ifc[i] <= "9")
                i++;
            const end = i;
            args.push(IfcTokenType.REF, Number(data.ifc.slice(start, end)));
            // args.push([IfcTokenType.REF, Number(data.ifc.slice(start, end))]);
            continue;
        }
        else if (c === "$")
            args.push(IfcTokenType.EMPTY, 0);
        else if (c === "*")
            args.push(IfcTokenType.UNKNOWN, 0);
        else if (c === "-" || (c >= "0" && c <= "9")) {
            const start = i;
            while ((data.ifc[i] >= "0" && data.ifc[i] <= "9") ||
                data.ifc[i] === "." ||
                data.ifc[i] === "e" ||
                data.ifc[i] === "E" ||
                data.ifc[i] === "-" ||
                data.ifc[i] === "+")
                i++;
            const end = i;
[...]

PR #7649, opened 2023-07-12
feat(es/minifier): Drop recursively used var declaration

Austaras:
**Description:**

This is only possible if we completely ignore TDZ error, but since this behavior matches terser, I believe it's fine.

**BREAKING CHANGE:**

#7641


kdy1 (dismissed):
Thank you!

---

swc-bump:
 - swc_ecma_usage_analyzer

jridgewell (approved):

PR #7677, opened 2023-07-19
feat(css/parser): Support `@starting-style`

CGQAQ:
This at rule allow authors to start CSS transitions on first style update. That is, you can now make the transition take effect when the display property changes from none to block.

- Spec: https://drafts.csswg.org/css-transitions-2/#defining-before-change-style-the-starting-style-rule
- Shipped in Chrome 117: https://chromestatus.com/feature/4515377717968896

Based on https://github.com/evanw/esbuild/pull/3249


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_atoms

PR #7671, opened 2023-07-18
refactor(es/minifier): Respect `toplevel` and `module` options

Austaras:
**Description:**

**BREAKING CHANGE:**

**Related issue (if exists):**

PR #7690, opened 2023-07-21
refactor(es/minifier): Respect top-level when invoking IIFE 

Austaras:
**Description:**

Follow up of #7671 

**BREAKING CHANGE:**


crates/swc/tests/fixture/issues-2xxx/2679/output/index.js
```diff
@@ -1,4 +1,2 @@
-!function() {
-    var a = {};
-    a.b = 1, a = null, window.test = a;
-}();
+var a;
```
jridgewell, line 1: This is leaking `a` into the global variable list. Is that ok?

jridgewell (approved):

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_minifier

crates/swc/tests/fixture/issues-2xxx/2679/output/index.js
Austaras: Yes, it's in es module.

kdy1 (commented):
reset ci

Issue #7621, opened 2023-07-03
Left shift in binary expression op is lexed as two `<` instead of as `<<`


dsherret:
### Describe the bug

Probably caused by https://github.com/swc-project/swc/pull/7439/files

```
Tokens for `a << 1`: [
    TokenAndSpan {
        token: a,
        had_line_break: false,
        span: Span {
            lo: BytePos(
                101,
            ),
            hi: BytePos(
                102,
            ),
            ctxt: #0,
        },
    },
    TokenAndSpan {
        token: <,
        had_line_break: false,
        span: Span {
            lo: BytePos(
                103,
            ),
            hi: BytePos(
                104,
            ),
            ctxt: #0,
        },
    },
    TokenAndSpan {
        token: <,
        had_line_break: false,
        span: Span {
            lo: BytePos(
                104,
            ),
            hi: BytePos(
                105,
            ),
            ctxt: #0,
        },
    },
    TokenAndSpan {
        token: numeric literal (1, 1),
        had_line_break: false,
        span: Span {
            lo: BytePos(
                106,
            ),
            hi: BytePos(
                107,
            ),
            ctxt: #0,
        },
    },
]
```

### Input code

```typescript
const test = a << 1;
```

### Expected behavior

Lexed as `a`, `<<`, `1`

### Actual behavior

Lexed as `a`, `<`, `<`, `1`

### Version

swc_ecma_parser 0.137.1

dsherret:
One interesting point is that the op in the AST is correctly parsed as `<<`, so maybe this is not considered a bug.

kdy1:
Can you post the rust code?

kdy1:
Ah nevermind I know the cause

PR #7692, opened 2023-07-21
Fix capturing tokens when backtracking

dsherret:
**Description:**

Working on fixing capturing tokens when backtracking. This is still broken and not exactly correct (not sure why atm).

**Related issue (if exists):** #7621

### DIFF
--- a/crates/swc_ecma_parser/src/parser/input.rs
+++ b/crates/swc_ecma_parser/src/parser/input.rs
@@ -483,15 +483,4 @@ impl<I: Tokens> Buffer<I> {
     pub(crate) fn set_token_context(&mut self, c: lexer::TokenContexts) {
         self.iter.set_token_context(c)
     }
-
-    #[inline]
-    pub(crate) fn state(&self) -> I {
-        self.iter.clone()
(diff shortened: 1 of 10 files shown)


crates/swc_ecma_parser/src/parser/input.rs
```diff
@@ -143,15 +141,14 @@ impl Tokens for TokensInput {
...
-    fn reset_to(&mut self, _: BytePos) {}
 }
 
 /// Note: Lexer need access to parser's context to lex correctly.
 #[derive(Debug)]
 pub struct Capturing<I: Tokens> {
     inner: I,
-    captured: Rc<RefCell<Vec<TokenAndSpan>>>,
+    // use a buffered singly linked list because clones occur infrequently
+    captured: BufferedSinglyLinkedList<TokenAndSpan>,
```
dsherret, line 151: I have yet to profile whether `BufferedSinglyLinkedList`, `SinglyLinkedList`, or something else would be more appropriate.

dsherret:
Going to close this one because it's out of date.

PR #7716, opened 2023-07-27
fix: ending block comment symbols did not start a new line.

### LINKED ISSUE #7492: Ability to emit jsdocs ending on previous line instead of same line?
### Describe the feature

Currently code like the following:

```ts
/** Test */
function test() {}

/** Test */
export class Test {
  /**
   * Multi-line
   */
  method() {}
}
```

Will emit with comments like so:

```ts
/** Test */ function test() {}
/** Test */ export class Test {
    /**
   * Multi-line
   */ method() {}
}
```

https://play.swc.rs/?version=1.3.62&code=H4sIAAAAAAAAA9PX0lIISS0uUdDS5%2BVKK81LLsnMz1MoAYpoaCpU1%2FJy8XLpIytJrSjILypRSM5JLC6GiFbzcikoANWAKAUtBd%2FSnJJM3ZzMvFSIgD6Iyk0tychPgZpYCwCN0JyldAAAAA%3D%3D&config=H4sIAAAAAAAAA11Ru27DMAzc8xWG5gxthg7dihTe%2BhGEQjsq9DBIOqgR%2BN8rKpLrdJHEu%2BPxofuh68w3W%2FPe3fMzBxMQI21xRniJAj8ZMbJMyJbcJObYWGGlhGYsyHqsNoTZ5oYf3p9TCBiFq6wKBGhEUVfk08vp1TScIPKQKOxbmBk%2FcXAR%2B0RnD8y9Q395dswyQrCyyyt9wBhAy1y3nje4z6dSemuLewXNUVxApa1WdNZUcn0a1KfEqhrAc5stuOiGZT9AgDj6f7IM2xR0UdyIQ%2FMv7iaky1yy6t%2Fo%2Fh8bezN%2FolZsszaOv1pm%2BZj1F5W1tcznAQAA

I'm emitting some typescript declarations using swc where only the jsdoc comments are preserved and the output doesn't look great because of this (I'd rather avoid doing post-processing on it, but it's not a big deal if I have to do that)

What are your thoughts on making swc do this? For a fast implementation, perhaps it should only do it for block comments that start with `*` and preceed declarations?

Both Babel and TypeScript keep the jsdoc on the previous line, but will also keep it on the same line if the person has the code that way.

_No response_

### Additional context

_No response_

liaoyinglong:
**Description:**

**BREAKING CHANGE:**

Closes #7492


liaoyinglong:
need more work

liaoyinglong:
I updated the code and test snapshots. I'll check the CI results later.

liaoyinglong:
@kdy1  Can you help me figure out why the test case didn't pass? It passed on my local machine.

It looks like the old code was used for compilation. I have already handled the potential panic that may be caused by calling `lookup_char_pos`.

```
let result = panic::catch_unwind(panic::AssertUnwindSafe(|| {
    $cm.lookup_char_pos(cmt.span.hi())
}));
```

kdy1:
Why did you use SourceMap after all?

liaoyinglong:
Retrieving the character immediately following `**/` to determine if a line break is needed.

Are there any other alternative solutions? Please let me know. Thank you.

kdy1:
You should check for the emitted character, not the input source code. I think you need a field to store the state for newline and use it from the comment emitter method.

liaoyinglong:
I will try. Thank you.

liaoyinglong:
Should i add state to `Comment` struct by parser stage ? 

Also, I have another imperfect solution: determine if a new line should be started by checking if cmt.text.starts_with('\n').

PR #7720, opened 2023-07-27
refactor(es/parser): Make `stacker` an optional dependency

### LINKED ISSUE #7719: Make `stacker` an optional dependency of `swc_ecma_parser`
### Describe the feature

Growing the stack via the `stacker` library might be undesired in some circumstances:
- Cross-compilation with [cargo-zigbuild](https://github.com/rust-cross/cargo-zigbuild), where the library doesn't compile successfully
- Memory-constrained environments, where the stack expansion might be considered too intrusive

Since `stacker` is only enabled on some platforms, it doesn't seem like the dependency is strictly necessary (and there are other ways to increase the stack size if required). Thus, it should be fine to make `stacker` an optional dependency that is enabled by default.

_No response_

### Additional context

_No response_

SkaillZ:
**Description:**

Makes the `stacker` dependency optional (enabled by default). This allows disabling the dependency e.g. due to one of the reasons outlined in https://github.com/swc-project/swc/issues/7719.

Closes https://github.com/swc-project/swc/issues/7719

### DIFF
--- a/crates/swc_ecma_parser/Cargo.toml
+++ b/crates/swc_ecma_parser/Cargo.toml
@@ -19,7 +19,7 @@ bench = false
 [features]
 # Used for debugging
 debug      = []
-default    = ["typescript"]
+default    = ["typescript", "stacker"]
 typescript = []
 verify     = ["swc_ecma_visit"]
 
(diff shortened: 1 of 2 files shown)


kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_parser

PR #7740, opened 2023-08-02
fix(es): Fix typo in a warning

BiscuiTech:
**Description:**

With the recent versions, there's a new error that tells us that `jsc.env` should not be used with `jsc.target`. But `jsc.env` does not exist. It's a typo with `env` at the root of the config.

**BREAKING CHANGE:**

No breaking changes.

No linked issue.


KthulupwnsAsura (approved):

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc

PR #7748, opened 2023-08-03
fix(plugin/runner): Use fs cache properly

JSerFeng:
**Description**

## compile the same WASM module multiple times

I tried using swc-loader in webpack, and found out that if I have multiple entries, and using WASM plugins, it increased build time significantly. That's because Webpack run swc-loader multiple times, and since swc-loader is async, and then invoke process_js multiple times at nearly the same time, while the first WASM module is being compiled but not finished, the second time process_js get invoked, it sees that the WASM is not compiled, so it begins compiling too.

## File system cache is not used

Currently SWC does not read file system cache at all, and write new cache to FS every time.
As I'm very likely not understand the design of swc WASM cache that much, and cache is hard, so this fix needs carefully review from you team.

After the modification, my little app's build time has been from 12s to 7s

### DIFF
--- a/crates/swc/src/config/mod.rs
+++ b/crates/swc/src/config/mod.rs
@@ -696,17 +696,17 @@ impl Options {
                     // target.
                     init_plugin_module_cache_once(true, &experimental.cache_root);
 
+                    let mut inner_cache = PLUGIN_MODULE_CACHE
+                        .inner
+                        .get()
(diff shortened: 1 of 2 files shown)


kwonoj (approved):

kdy1 (approved):
swc-bump:
 - swc_plugin_runner

PR #7757, opened 2023-08-04
Make usage of lexical from swc_ecma_parser optional

gmcsorley-work:
This PR makes lexical part of the default features for the swc_ecma_parser crate but allows it to be turned off to use BigInt for equivalent parsing of large numbers.

**Description:** As discussed in https://github.com/swc-project/swc/issues/7752, lexical contains a number of soundness issues but doesn't appear to be actively supported. Given the relatively low integration surface it seems reasonable to replace the usage of lexical with another package to avoid this issue. This does so in an optional fashion so that the decision can be made by the user which mechanism to use.

Related issue (if exists): https://github.com/swc-project/swc/issues/7752

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4117,6 +4117,7 @@ dependencies = [
  "either",
  "lexical",
  "num-bigint",
+ "num-traits",
  "pretty_assertions",
  "serde",
  "serde_json",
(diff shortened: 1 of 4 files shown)


gmcsorley-work:
Note that either this PR or https://github.com/swc-project/swc/pull/7758 should be accepted, but not both.

kdy1:
Thank you! I'll close this in favor of https://github.com/swc-project/swc/pull/7758

PR #7758, opened 2023-08-04
refactor(es/parser): Do not use `lexical`

### LINKED ISSUE #7752: Add ability to disable `lexical` dep

gmcsorley-work:
This PR replaces the current usage of lexical within the swc_ecma_parser crate with equivalent parsing of large numbers using BigInt.

**Description:** As discussed in https://github.com/swc-project/swc/issues/7752, lexical contains a number of soundness issues but doesn't appear to be actively supported. Given the relatively low integration surface it seems reasonable to replace the usage of lexical with another package to avoid this issue.

**Related issue:**

- Closes #7752

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4115,8 +4115,8 @@ version = "0.137.5"
 dependencies = [
  "criterion",
  "either",
- "lexical",
  "num-bigint",
+ "num-traits",
  "pretty_assertions",
  "serde",
  "serde_json",
--- a/crates/swc_ecma_parser/Cargo.toml
+++ b/crates/swc_ecma_parser/Cargo.toml
@@ -25,8 +25,8 @@ verify     = ["swc_ecma_visit"]
 
 [dependencies]
 either      = { version = "1.4" }
-lexical     = { version = "6.1.0", features = ["power-of-two", "parse-integers", "parse-floats"], default-features = false }
 num-bigint  = "0.4"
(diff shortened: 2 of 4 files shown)


KthulupwnsAsura (approved):

gmcsorley-work:
Note that either this PR or https://github.com/swc-project/swc/pull/7757 should be accepted, but not both.

kdy1 (approved):
Thank you so much and sorry for the late review 😄 
I was very busy the last weekend.

---

swc-bump:
 - swc_ecma_parser

PR #7766, opened 2023-08-07
refactor(es/parser): Do not validate top-level await with target

JSerFeng:
**Description:**

Now that we cannot use jsc.target and env.targets at the same time, top-level await will be always failed to parse if we use env.targets.

How about adding extra browserslist validate for it ?

Or is parser not supposed to take care target ?

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_parser/src/error.rs
+++ b/crates/swc_ecma_parser/src/error.rs
@@ -67,7 +67,6 @@ pub enum SyntaxError {
     GetterParam,
     SetterParam,
 
-    TopLevelAwait,
     TopLevelAwaitInScript,
 
     LegacyDecimal,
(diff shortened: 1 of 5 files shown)


kdy1:
Hmm.. I think it's fine to refactor parser to not care about top-level awaits.

kdy1 (approved):
I think CI will fail, but the current changes look good.

---

swc-bump:
 - swc_ecma_parser

kdy1:
Can you fix CI?

kdy1 (dismissed):
Thank you!

JSerFeng:
Sry, still some errors. It's too late here, I fix that tomorrow

PR #7773, opened 2023-08-07
fix(es/minifier): Preserve `cooked` while compressing template literals

### LINKED ISSUE #7770: Some parts of template literals are copied with `minify()`
### Describe the bug

- https://github.com/swc-project/swc/issues/7769

I think this is related to `cm: Lrc<SourceMap>` in the ES codegen

### Input code

```typescript
(self["webpackChunkabc_orchestra"] = self["webpackChunkabc_orchestra"] || []).push([[118],{

    /***/ "./node_modules/path/to/abc.js":
    /***/ (function(__unused_webpack_module, exports, __webpack_require__) {
    
    "use strict";
    
    Object.defineProperty(exports, "__esModule", ({ value: true }));
    exports.MainCSS = exports.sHeight = void 0;
    exports.sHeight = "asdasd";
    const sWidth = "asdasd";
    const sAsideWidth = "weqweqw";
    const sHeight = "asdsads";
    function absolute(abeetew = {}) {
        return `
        abc: cde;
        def: ${abeetew.def || 0};
        qwert: ${abeetew.qwert || 0};
        asda: ${abeetew.asda || 0};
        left: ${abeetew.abc || 0};
    `;
    }
    function flex() {
        return `
        abc: ab;
        cdef: de;
        ghn: fg;
        sadasn: hi;
    `;
    }
    exports.MainCSS = `
    .ThisshouldOnlyBeonTop {
        ${absolute()}
        asda-asdasda: asdad-asdsa;
    }
    .abcBlablaOneQSR {
        line-height: ${exports.sHeight};
        text-shadow: none;
    }
    .abcBlablaOne .asdsad {
        ${flex()}
        ${absolute()}
        
        asdasd-adsdas: asdasd-asdas;
        xzczxc: zxczxcc;
    }
    .asdasd {
        ${absolute({ abc: sAsideWidth + "!important", def: exports.sHeight + "!important", qwert: "0 !important" })}
        sdsd: auto !important;
        xcxcx: auto !important;
    }
    .aasdasdasd .asdsada {
        we: ${sWidth} !important;
    }
    .asdasdasd:not(.adsadsad) .asdasdasd {
        ds: ${parseFloat(exports.sHeight) + parseFloat(sHeight)}rem !important;
    }
    `;
    //# sourceMappingURL=abc.js.map
    
    /***/ })
    
    
    }]);
```

### Config

_No response_

### Playground link

_No response_

### Expected behavior

It should not be copied

### Actual behavior

_No response_

### Version

1.3.74

### Additional context

_No response_

kdy1:
**Description:**

**Related issue:**

 - Closes #7770


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

crates/swc_ecma_minifier/tests/fixture/issues/7770/input.js
```diff
@@ -0,0 +1,73 @@
+(self["webpackChunkabc_orchestra"] = self["webpackChunkabc_orchestra"] || []).push([
```
jridgewell, line 1: I was able to reduce this down to 
```
const sWidth = "asdasd";
function absolute() {
    return `
    `;
}
function flex() {
    return `
    `;
}
exports.MainCSS = `
.ThisshouldOnlyBeonTop {
    ${absolute()}
}
.abcBlablaOne .asdsad {
    ${flex()}
}
.aasdasdasd .asdsada {
    we: ${sWidth} !important;
}
`;
```

jridgewell (changes requested):
I think this algorithm is still faulty. We never reinitialize `cur_cooked` to a `Some(_)`, so we'll only push cooked values for the first quasi? What about all the other quasis?

I think the better algorithm here is to just peek the last item in `quasis`, and mutate the value directly when we find a string literal.

kdy1:
I added logic for setting `cur_cooked` to `None` to ensure that we only set `cooked` of the new template literal strings if all elements have `cooked` set.

crates/swc_ecma_minifier/src/compress/pure/strings.rs
```diff
@@ -258,13 +264,15 @@ impl Pure<'_> {
...
+                        }
                     }
                     _ => {
                         quasis.push(TplElement {
                             span: DUMMY_SP,
                             tail: true,
-                            cooked: Some(Atom::from(&*cur_cooked)),
+                            cooked: cur_cooked.take().map(From::from),
                             raw: take(&mut cur_raw).into(),
                         });
```
jridgewell, line 277: We need to reinitialize `cooked = Some(String::new())` here.

jridgewell (changes requested):
My point is that that the second quasi in the following situations will not have a `cooked` value, when it should:

```js
foo`
  \x<-invalid, so no cooked
  ${'this breaks splits the quasis'}
  this should have a cooked
`;

foo`
  \x<-invalid, so no cooked
  ${'this breaks splits the quasis'}
  this should have a cooked
  ${'this breaks splits the quasis'}
  this should also have a cooked
`;
```

crates/swc_ecma_minifier/src/compress/pure/strings.rs
kdy1: Sorry! I missed the point. And thanks!

jridgewell (approved):

PR #7774, opened 2023-08-07
refactor(es/parser): Do not validate top-level await with target 

JSerFeng:
Description:

Original: #7766 

I accidentally force pushed after bump comment, so create this new pr

BREAKING CHANGE:

Related issue (if exists):


kdy1 (approved):
Thank you! The order of force pushes and the bump commit does not matter, but well...

---

swc-bump:
 - swc_ecma_parser

PR #7775, opened 2023-08-07
fix(es/module): Use `jsc.baseUrl` while resolving absolute paths

### LINKED ISSUE #7761: Wrong resolved import paths when using 'paths'
### Describe the bug

I just ran into #4057 which is open and stated as fixed by #5222. Nevertheless it was reopened as requested by @alexeagle (the thread is locked so I created a new issue) in combination with bazel and swc (also where I ran into the paths problem). I diged into it and looked into the `swc_ecma_tranforms_module`. I think it is not a problem with symlinks (at least they were no problem in my small test setup) like referenced in #4057 but with how the absolute path of the resolved file path is generated [(swc_ecma_transforms_module/src/path.rs#L262-L271).](https://github.com/swc-project/swc/blob/c9a55603ae7369e43b8a4dea4e3abdfdd43389ca/crates/swc_ecma_transforms_module/src/path.rs#L262-L271)

My test setup: https://github.com/joelMuehlena/swc-paths-bug

Output from running ./bash in the root of my test setup with a modified SWC version for more output.

```bash
# Output of a custom script
Current dir is /home/joel/swc-paths-bug-repro
Build from 'root (/home/joel/swc-paths-bug-repro)'

# Executed command
  ./swc compile tmp/execroot/test/base/src/index.ts tmp/execroot/test/base/lib/test/a.ts  --out-dir swc_out --config-file base/.swcrc

# Output of previous cmd (swc modified for some output)
resolve_import: base=Real("/home/joel/swc-paths-bug-repro/base/src/index.ts"), module_specifier=@test/a
target before transform=Real("./lib/test/a")
file with import to resolve="/home/joel/swc-paths-bug-repro/base/src/index.ts" # This is correct
import target base="./lib/test/a" # This seems also correct (created by the Resolver itself)

# Output produced by the method referenced earlier
absolute_path="/home/joel/swc-paths-bug-repro/base/src/index.ts" # Already absolute and correct -> nothing altered

path="./lib/test/a", current_dir="/home/joel/swc-paths-bug-repro" # Here the issue begin with CWD is the where the swc binary was called but which is not the ts project root.
absolute_path="/home/joel/swc-paths-bug-repro/lib/test/a"

# Output after creating the absolute path
import target base2="/home/joel/swc-paths-bug-repro/lib/test/a"
resolved path="../../lib/test/a"
```

### Input code

_No response_

### Config

```json
{
  "exclude": ["node_modules"],
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "decorators": true,
      "dynamicImport": true
    },
    "transform": null,
    "target": "es2021",
    "loose": false,
    "externalHelpers": false,
    "keepClassNames": false,
    "baseUrl": ".",
    "paths": {
      "@test/*": ["lib/test/*"]
    }
  },
  "module": { "type": "commonjs" },
  "sourceMaps": true
}
```

### Playground link

_No response_

### Expected behavior

Correct resolving of the paths.

### Actual behavior

Resolving with the dir from where swc ran instead of the project root.

### Version

main branch

### Additional context

_No response_

kdy1:
**Description:**

**Related issue:**

 - Closes #7761


kdy1 (commented):
swc-bump: 
 - swc_ecma_transforms_module

PR #7772, opened 2023-08-07
fix(es/minifier): Abort seq inliner if a same var is defined in outer scope

Austaras:
**Description:**

The algorithm here is directly copied from terser, I don't if it's correct or can be improved, but it does fixes the issue.

**Related issue:**

 - Closes #7749

PR #7804, opened 2023-08-14
fix(es/minifier): Abort seq inliner if var is not fn_local or reassigned

### LINKED ISSUE #7784: Seq inliner would change result in op assign
### Describe the bug

Similar to #7749 but won't be fixed by #7772

### Input code

```typescript
let a = 1

function foo(g) {
  var t = g()
  a += t
}

function g() {
  a = 2
  return 1
}

foo(g)

console.log(a)
```

### Config

_No response_

### Playground link

_No response_

### Expected behavior

evals to 3

### Actual behavior
Output is
```js
let a = 1;
console.log(a += (a = 2, 1));
```

evals to 2

### Version

latest

### Additional context

_No response_

Austaras:
**Description:**

It turns out that the original implementation of `infect`is incomplete because it cannot cover function param, and the new implementation introduced in #7772 is redundant because what [terser](https://github.com/terser/terser/blob/master/lib/compress/tighten-body.js#L909C18-L909C28) do is basically checking `fn_local`

**Related issue:**

 - Closes #7784


kdy1 (approved):
Thank you so much!

---

swc-bump:
 - swc_ecma_usage_analyzer

jridgewell (dismissed):
How does @kdy1 feel about all the comments?

crates/swc_ecma_minifier/src/program_data.rs
```diff
@@ -540,69 +533,65 @@ impl VarDataLike for VarUsageInfo {
...
-                    }
-                    if ranges.is_empty() {
-                        break;
-                    }
-                }
-                Ok(res)
-            })
-            .ok()
-    }
+    //     #[allow(clippy::single_range_in_vec_init)]
```
kdy1, line 536: @Austaras Is there a reason you commented this out instead of removing?

Austaras: No. It's just a large chunk of code so I thought maybe it would be useful one day. But we could also retrieve it from git history.

kdy1: If so, can you remove them?

Austaras: Done.

kdy1 (approved):

PR #7811, opened 2023-08-15
feat(es/ast): Expose `Archived` types

### LINKED ISSUE #7810: Export Archived types from `rkyv::Archive` implementations
### Describe the feature

Looking to make the `Archived<T>` versions of the AST nodes public.

Related PR: https://github.com/swc-project/swc/pull/7811

I'm trying to implement a data structure that's based on rkyv's Archive, Serialize, and Deserialize traits. The data structure seeks to allow accessing fields on a value that is either Archived or not Archived, but only completely deserializes when necessary. It's working, but I'd primarily wanted to use it with the types exported from `swc_ecma_ast`, but the Archived types for the nodes are not public.

This means that when trying to access the Archived type of a `Module` for example, even when doing `rkyv::Archived<Module>`, the experience is poor because there won't be any type completion. It _does_ compile if you look at the source and assign the appropriate types to references of fields, but it's really difficult to work with.

I have basically two options (and of course, open to suggestions):

- Create transparent wrappers for each of the AST nodes which add getter methods for the underlying Archived types and can easily cast/transmute to/from them. This won't update as the source changes though. Also considering a hacky approach which would reach into the source to use in a macro which expands to the same thing.
- Create the wrapper types and then a separate implementation of `rkyv::{Archive, Serialize, Deserialize}`.

I understand that in the event that swc decides to remove the `rkyv` implementation, these types won't exist anymore which leaves only the second route anyways. But I thought it was worth asking what folks think and saving some time now, since the types are already there.

_No response_

### Additional context

Originally asked [on discord](https://discord.com/channels/889779439272075314/1140696751200800850/1140696751200800850)

JQuezada0:
**Description:**

For crates using both `rkyv` and `swc`, where ast nodes from `swc_ecma_ast` are included in data that's serialized with `rkyv`, it's difficult to access the archived data from the archived ast nodes since the types are private.

More context in the issue.

**Related issue:**

 - Closes #7810.

### DIFF
--- a/crates/swc_common/src/lib.rs
+++ b/crates/swc_common/src/lib.rs
@@ -78,3 +78,15 @@ pub mod util;
 
 #[cfg(all(not(debug_assertions), feature = "plugin-rt", feature = "plugin-mode"))]
 compile_error!("You can't enable `plugin-rt` and `plugin-mode` at the same time");
+
+/// Warning: The particular implementation of serialization and deserialization
+/// of the ast nodes may change in the future, and so these types would be
+/// removed. It's safe to say they will be serializable in some form or another,
+/// but not necessarily with these specific types underlying the implementation.
+/// As such, *use these types at your own risk*.
+#[cfg(feature = "rkyv-impl")]
+#[doc(hidden)]
+pub use self::syntax_pos::{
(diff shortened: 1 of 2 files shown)


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_common

PR #7824, opened 2023-08-17
perf(es/visit): Use `debug_unreachable` to reduce binary size

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_ecma_visit

PR #7839, opened 2023-08-19
fix(es/minifier): Don't inline properties if the var is not `fn-local`

kdy1:
**Description:**

**Related issue:**

 - https://github.com/vercel/next.js/issues/54192.


kdy1:
@Austaras @jridgewell 

I think we need to improve logic for `is_fn_local`, but I think I should focus on fixing https://github.com/vercel/next.js/issues/54192 for now

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

Austaras:
Should we abort in the else block of [this](https://github.com/swc-project/swc/blob/main/crates/swc_ecma_minifier/src/compress/optimize/props.rs#L54)? It's causing many item being wrongfully inlined. For example
```js
export function createCaseFirst(g, methodName) {
  var chr = g.get(0)
  console.log(123)
  return chr[methodName]() + trailing;
}
```
In this case chr cannot be inlined because of side effects.

kdy1:
I reflected it 👍

jridgewell (approved):

hexiaokang (approved):

PR #7856, opened 2023-08-24
fix(es/minifier): Don't remove exports

lewisl9029:
**Description:**

Hi there! I recently upgraded SWC from 1.3.60 to the latest 1.3.78 and discovered that some of my builds were broken because SWC's minifier was removing exports.

I've added some minimal repros of the cases I encountered as failing test cases, and a small change to the minifier to fix them.

Happy to iterate on this as you see fit. Cheers!

PR #7853, opened 2023-08-24
fix(es): Fix default value of `jsc.minify.format.comments`

kdy1:
**Description:**

**Related issue:**

 - Closes #7842

PR #7858, opened 2023-08-25
fix(es/codegen): Fix placing of comments of yield arguments

lewisl9029:
**Description:**

Looks like the bug I ran into had nothing to do with the changes in #7856, since it's reproducible without it. Looks like it might have only surfaced now because https://github.com/swc-project/swc/pull/7853 changed the default value of `jsc.minify.format.comments`? Added a minimal test case here with the expected result.

Here's the actual output:

```js
 export var padding = '';
 export function exec2({ commands }) {
     return __awaiter(this, void 0, void 0, function*() {
         for(let i2 = 0; i2 < commands.length; i2++){
             let command = commands[i2];
             yield // some-comment
             function({ command }) {
                 command();
             }({
                 command,
                 handleError
             });
         }
     });
 }
```

The comment ends up getting added after the yield, which makes the output invalid.

Going to see if I can figure out a fix tomorrow, but let me know if you have any ideas on where to start looking in the meantime!

### DIFF
--- a/crates/swc/tests/fixture/minify/comments/false-02/input/.swcrc
+++ b/crates/swc/tests/fixture/minify/comments/false-02/input/.swcrc
@@ -1,0 +1,19 @@
+{
+    "isModule": true,
+    "jsc": {
+        "target": "es2022",
+        "transform": {
+            "optimizer": {
+                "simplify": true,
+            }
+        },
+        "minify": {
+            "format": {
+                "comments": true
+            },
+            "compress": true,
+            "mangle": false
+        },
+    },
+    "minify": false,
+}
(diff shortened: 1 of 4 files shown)


kdy1:
We may need `need_paren` like in https://github.com/swc-project/swc/blob/5ae2e810d8d8a2d23c9e81eae128fb72c5c1f09c/crates/swc_ecma_codegen/src/lib.rs#L2936-L2940 for `emit_yield_expr` too.

https://github.com/swc-project/swc/blob/5ae2e810d8d8a2d23c9e81eae128fb72c5c1f09c/crates/swc_ecma_codegen/src/lib.rs#L1882-L1886

lewisl9029:
@kdy1 thank you so much for the suggestion, it worked great!
Added the fix and updated the expected output accordingly in https://github.com/swc-project/swc/pull/7858/commits/7d1bc9f0f56c02c57d92bbfeed45a4c3114b5a49

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_codegen

PR #7873, opened 2023-08-26
feat(es): Add an option to disable builtin transforms

kdy1:
**Description:**

I'm experimenting with a new API. If this option is enabled, all builtin transforms are disabled and only Wasm plugins are applied.
 


kdy1 (commented):
swc-bump:
 - swc

PR #7883, opened 2023-08-28
fix(es/dep-graph): Analyze import type children

dsherret:
**Description:**

It was not analyzing children of a dynamic imports.

### DIFF
--- a/crates/swc_ecma_dep_graph/src/lib.rs
+++ b/crates/swc_ecma_dep_graph/src/lib.rs
@@ -177,6 +177,7 @@ impl<'a> Visit for DependencyCollector<'a> {
             specifier_span: node.arg.span,
             import_attributes: Default::default(),
         });
+        node.visit_children_with(self);
     }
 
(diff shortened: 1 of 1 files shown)


crates/swc_ecma_dep_graph/src/lib.rs
```diff
@@ -810,4 +811,47 @@ const d10 = await import("./d10.json", { assert: { type: "json", ["type"]: "bad"
...
+                DependencyDescriptor {
+                    kind: DependencyKind::ImportType,
+                    is_dynamic: false,
+                    leading_comments: Vec::new(),
+                    span: Span::new(BytePos(48), BytePos(176), Default::default()),
+                    specifier: JsWord::from("./a.d.ts"),
+                    specifier_span: Span::new(BytePos(55), BytePos(65), Default::default()),
+                    import_attributes: ImportAttributes::None,
+                },
+                DependencyDescriptor {
```
dsherret, line 836: These bottom two were missing before.

PR #7890, opened 2023-08-29
refactor(es/minifier): Remove `mutated` and `mutation_by_call_count`

Austaras:
**Description:**

`mutated` can be seen as a combination of `reassigned` and `has_property_mutation`, and `mutation_by_call_count` is simply useless.

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/optimize/inline.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/inline.rs
@@ -98,9 +98,7 @@ impl Optimizer<'_> {
             //
             // TODO: Allow `length` in usage.accessed_props
             if usage.declared
-                && !usage.reassigned
-                && !usage.mutated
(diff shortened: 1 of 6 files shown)


crates/swc_ecma_minifier/src/compress/optimize/inline.rs
```diff
@@ -323,7 +322,7 @@ impl Optimizer<'_> {
                 && usage.declared
                 && may_remove
                 && !usage.reassigned
-                && (usage.can_inline_var() || usage.is_mutated_only_by_one_call())
+                && usage.assign_count == 0
```
Austaras, line 325: There's no need to check if a used only once var is mutated -- we inline it anyway.

kdy1 (approved):
Thank you!

---

swc-bump:
 - swc_ecma_usage_analyzer

jridgewell (approved):

PR #7943, opened 2023-09-13
fix(es/minifier): Parse a file as a module only if `opts.module` is `true`

### LINKED ISSUE #6130: swc.minify() always assumes strict mode
### Describe the bug

`swc.minify()` always behaves as if the passed-in code is in strict mode, regardless of whether `JsMinifyOptions["module"]` is `true` or `false`. You can successfully configure `swc.transform()` in this way with `Options["isModule"]`, but there doesn't seem to be a way to pass this configuration directly to `minify` _or_ to have `minify` look for and use a `.swcrc` file.

### Input code

```typescript
swc.minify('const f = () => "\240";', {
    compress: {
        defaults: false,

        arguments: false,
        arrows: false,
        booleans: false,
        booleans_as_integers: false,
        collapse_vars: false,
        comparisons: false,
        computed_props: false,
        conditionals: true,
        dead_code: true,
        directives: false,
        drop_console: false,
        drop_debugger: false,
        ecma: 5,
        evaluate: false,
        expression: false,
        hoist_funs: false,
        hoist_props: false,
        hoist_vars: false,
        ie8: false,
        if_return: false,
        inline: 0,
        join_vars: false,
        keep_classnames: false,
        keep_fargs: false,
        keep_fnames: false,
        keep_infinity: false,
        loops: false,
        module: false,
        negate_iife: false,
        passes: 1,
        properties: false,
        reduce_vars: false,
        sequences: false,
        side_effects: false,
        switches: false,
        toplevel: false,
        typeofs: false,
        unused: false,
    },
    format: { comments: false, semicolons: true },
    mangle: { toplevel: false },
    module: false,
    toplevel: false,
    ecma: 5,
});
```

### Config

```json
// see object passed to `minify` above
```

### Playground link

_No response_

### Expected behavior

Because `JsMinifyOptions["module"]` is set to `false` in the configuration object passed to `swc.minify()`, I expect the following output:
```js
const f=()=>"\xa0";
```

### Actual behavior

Instead, I receive the following error:
```
× Legacy octal escape is not permitted in strict mode

    ╭────
01 │ const f = () => "\240";
    ·                  ──
    ╰────
Caused by:
    0: failed to parse input file
    1: error was recoverable, but proceeding would result in wrong codegen
    2: Syntax Error
```

### Version

1.3.6

### Additional context

I can get this to work in the playground, but only because the playground allows a full `.swcrc` where I can set `isModule` to `true`. I'm assuming the playground uses `transform` instead of `minify`.

andersk:
**Description:**

Previously it was impossible to minify non-strict code (e.g. `delete identifier;`) because `minify()` unconditionally parsed the input as a module regardless of the `module` setting.

**BREAKING CHANGE:**

`minify()` now respects the `module` parameter instead of ignoring it, so users who had been processing modules with `minify()` without passing `{ module: true }` will now need to pass it.

**Related issue:**

 - Closes #6130.


kdy1 (approved):
swc-bump:
 - swc

kdy1 (approved):

andersk (commented):
CI failed the first time with what looked like an unrelated filesystem error, but passed the second time.

andersk (commented):
(empty comment review to re-trigger bot)

PR #7958, opened 2023-09-15
fix(es/compat): Handle `PrivateName` in `logical_assignments`

### LINKED ISSUE #7956: "private should be removed by class_properties" error when use preset_env
### Describe the bug

When use preset_env to fold ast of using `this.#` and `||=`, for  example`this.#a ||= 1`, swc will panic at `private should be removed by class_properties`.

Discovery these after read the logic of swc:
- `preset_env` will use `folder` according to inner `plugin.json` data and `target` config.
- In Chrome 74, `class_property` have been natively implemented , see [plugin.json](https://github.com/swc-project/swc/blob/e67bf05f9d4c558e1601b1d11ba860c93980e006/crates/swc_ecma_preset_env/data/%40babel/compat-data/data/plugins.json#L59) . Therefore, it will not be used
- but `logical_assignments` depends on `class_property`. So it will run into the `unreachable!("private should be removed by class_properties")`

If change the target to `chrome 73` or include `class_property` when use `preset_env`, the code will run as normal

### Input code

```rs
fn main() {
    let source = r#"
    class A {
        b() {
            this.#a ||= 1;
        }
    }
"#;

    let cm: Lrc<SourceMap> = Default::default();
    let fm = cm.new_source_file(FileName::Real(Path::new("").to_owned()), source.to_owned());

    let lexer = Lexer::new(
        Syntax::Es(EsConfig {
            ..Default::default()
        }),
        swc_ecma_ast::EsVersion::default(),
        StringInput::from(&*fm),
        None,
    );

    let ast = parser.parse_module().unwrap();

    let globals = Globals::default();
    let targets: Targets = serde_json::from_str(r#"{"chrome":74}"#).unwrap();
    GLOBALS.set(&globals, || {
        let _ = try_with_handler(cm.clone(), Default::default(), |handler| {
            HELPERS.set(&Helpers::new(true), || {
                HANDLER.set(handler, || {
                    let mut preset_env = swc_ecma_preset_env::preset_env(
                        Mark::new(),
                        Some(NoopComments),
                        swc_ecma_preset_env::Config {
                            targets: Some(targets),
                            mode: Some(swc_ecma_preset_env::Mode::Entry),
                            ..Default::default()
                        },
                        Assumptions::default(),
                        &mut FeatureFlag::default(),
                    );
                    preset_env.fold_module(ast.clone());
                });
            });

            Ok(())
        });
    });
}
```

### Config

_No response_

### Playground link

https://github.com/zhangpanweb/swc-private-assign-error

### SWC Info output

_No response_

### Expected behavior

`preset_env` will work as normal, and not panic

### Actual behavior

_No response_

### Version

swc_common: 0.31.17

### Additional context

_No response_

magic-akari:
**Related issue:**

 - Closes #7956

### DIFF
--- a/crates/swc/tests/fixture/issues-7xxx/7956/input/1/.swcrc
+++ b/crates/swc/tests/fixture/issues-7xxx/7956/input/1/.swcrc
@@ -1,0 +1,10 @@
+{
+    "jsc": {
+        "parser": {
+            "syntax": "ecmascript"
+        }
+    },
+    "env": {
+        "targets": "chrome 74"
+    }
+}
--- a/crates/swc/tests/fixture/issues-7xxx/7956/input/1/index.js
+++ b/crates/swc/tests/fixture/issues-7xxx/7956/input/1/index.js
@@ -1,0 +1,20 @@
+class A {
+    #a;
+    b() {
+        this.#a ||= 1;
+    }
+    c(x) {
+        x.#a ||= 1;
+    }
+    log() {
+        console.log(this.#a);
+    }
+}
+
+const x = new A();
+x.b();
+x.log();
+
+const y = new A();
+y.c(y);
+y.log();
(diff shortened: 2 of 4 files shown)


kdy1 (approved):
swc-bump:
 - swc_ecma_transforms_compat

Issue #7977, opened 2023-09-20
bug: nullish coalescing with function invokation is not properly compiled


layershifter:
### Describe the bug

The problem appears on `ArrowFunctionExpression` that don't have `BlockStatement`:

```js
const foo = () => baz() ?? qux;
```

### SWC result 🟥 

```js
var _baz; // <<<< variable is in a module scope 💣 
const foo = ()=>(_baz = baz()) !== null && _baz !== void 0 ? _baz : qux;
```

### Babel 🟢 

```js
const foo = () => {
  var _baz;
  return (_baz = baz()) !== null && _baz !== void 0 ? _baz : qux;
};
```

### TS 🟢 

```js
const foo = () => {
    var _baz;
    return (_baz = baz()) !== null && _baz !== void 0 ? _baz : qux;
};
```

### Input code

```typescript
const foo = () => baz() ?? qux;
```

### Config

```json
{
  "$schema": "https://json.schemastore.org/swcrc",
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "dynamicImport": true
    },
    "transform": {
      "react": {
        "runtime": "automatic"
      }
    },
    "baseUrl": "./",
    "target": "es2019",
    "keepClassNames": true,
    "minify": {
      "compress": false
    }
  },
  "isModule": true,
  "module": {
    "type": "es6"
  },
  "minify": false
}
```

### Playground link

[...]

PR #7980, opened 2023-09-20
fix(es/compat): Handle nullish in fn expr scope

magic-akari:
**Description:**

**Related issue:**

  - Closes: #7977

### DIFF
--- a/crates/swc_ecma_transforms_compat/src/es2020/nullish_coalescing.rs
+++ b/crates/swc_ecma_transforms_compat/src/es2020/nullish_coalescing.rs
@@ -1,7 +1,7 @@
 use std::mem::take;
 
 use serde::Deserialize;
-use swc_common::{util::take::Take, Span, DUMMY_SP};
+use swc_common::{util::take::Take, Span, Spanned, DUMMY_SP};
 use swc_ecma_ast::*;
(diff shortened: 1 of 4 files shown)


crates/swc_ecma_transforms_compat/src/es2020/nullish_coalescing.rs
```diff
@@ -192,6 +192,27 @@ impl VisitMut for NullishCoalescing {
...
+        if !self.vars.is_empty() {
+            if let BlockStmtOrExpr::Expr(expr) = n {
+                // () => expr
+                // () => { return expr; }
+                let span = expr.span();
+                let mut stmts = vec![Stmt::Return(ReturnStmt {
+                    span: DUMMY_SP,
+                    arg: Some(expr.take()),
+                })];
+                stmts.visit_mut_with(self);
```
jridgewell, line 209: Do we need to revisit? Can't we just inject the variables?

magic-akari: Yes, we should avoid to revisit.

Hotell:
many thanks for quick fix ! any ETA when new release will land on npm ? cheers

Issue #8016, opened 2023-09-27
Incorrect error `Import declarations in a namespace cannot reference a module`?


mischnic:
### Describe the bug

Since the Typescript pass rewrite, I get a `Import declarations in a namespace cannot reference a module` for code which works fine in TSC and old swc

(This is inside of Parcel which uses the swc crate)

### Input code

```typescript
import commonPathPrefix = require('common-path-prefix');
```

The whole file: https://github.com/parcel-bundler/parcel/blob/c182930c631e3bf06034d5b350064152b57206ce/packages/utils/parcel-lsp/src/LspServer.ts#L27

### Config

You can see the options used for invoking the Typescript swc pass here: https://github.com/parcel-bundler/parcel/pull/9271/files#diff-871dd9618b173ec16a893d692b991948e079f1f85c6db41db35eded117ba58f2

### Playground link

Not yet published to the playground

### SWC Info output

```
swc_core = { version = "0.83.26"
```

### Expected behavior

Should work without an error

### Actual behavior

```
@parcel/transformer-js: Import declarations in a namespace cannot reference a 
module.

  /home/runner/work/parcel/parcel/packages/utils/parcel-lsp/src/LspServer.ts:27:1
    26 | import * as url from 'url';
  > 27 | import commonPathPrefix = require('common-path-prefix');
  >    | ^
    28 | 
    29 | // import {TextDocument} from 'vscode-languageserver-textdocument';
```

### Version

swc_core "0.83.26"

### Additional context

https://github.com/parcel-bundler/parcel/pull/9271

magic-akari:
It looks like the parameter names for TypeScript pass are incorrect.
It should be changed to `top_level_mark`.

mischnic:
Thanks!

PR #8018, opened 2023-09-27
fix(es/typescript): Rename wrong `unresolved_mark`

magic-akari:
**Description:**

This is just a typo fix. No functional changes are expected. 
However, correcting the wrong parameter name will avoid confusing our Rust-side users.

**BREAKING CHANGE:**

- Closes: #8016

### DIFF
--- a/crates/swc_ecma_transforms_typescript/src/transform.rs
+++ b/crates/swc_ecma_transforms_typescript/src/transform.rs
@@ -934,9 +934,13 @@ impl Transform {
         let mut should_inject = false;
         let create_require = private_ident!("_createRequire");
         let require = private_ident!("__require");
-        let global_span = DUMMY_SP.apply_mark(self.top_level_mark);
-        let cjs_require = quote_ident!(global_span, "require");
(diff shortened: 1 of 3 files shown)


crates/swc_ecma_transforms_typescript/src/transform.rs
```diff
@@ -1101,7 +1101,7 @@ impl Transform {
                                 expr.make_assign_to(
                                     op!("="),
                                     member_expr!(
-                                        DUMMY_SP.apply_mark(self.unresolved_mark),
+                                        DUMMY_SP.apply_mark(self.top_level_mark),
                                         module.exports
```
magic-akari, line 1105: It does look like `unresolved_mark` needs to be used here instead of `top_level_mark`, I'll confirm that later.

magic-akari: However, `TsImportExportAssignConfig::Classic` is not used in any part of the code, so it may be unnecessary to handle it here.

Issue #8021, opened 2023-09-28
Static decorators are executed at the wrong time


jrandolf:
### Describe the bug

Decorators get hoisted before the class declaration which causes problems when a static function in a class is used to decorate a method.

### Input code

```typescript
class A {
  static test(a: any, b: any) {}

  @A.test()
  hi() {}
}
```

### Config

```json
{
  "$schema": "https://json.schemastore.org/swcrc",
  "module": {
    "type": "commonjs"
  },
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "decorators": true
    },
    "target": "es2022",
    "transform": {
      "decoratorVersion": "2022-03"
    }
  }
}
```

### Expected behavior

It should work.

### Version

1.3.90

jrandolf:
I suspect that would require a lot of change with the private methods and what not. I don't have time to work on this more, so I'll leave a summary.

This doesn't work. I suspect this needs to be done outside the class, after it has been declared, similar to how TypeScript and the proposal encourage it.

Something like

```
let a = class A {}
(() => {
 ...do static stuff here
})
return a;
```

PR #8022, opened 2023-09-28
fix(es/compat): wrap transpiled decorators in arrow functions

jrandolf:
**Description:**

This PR wraps transpiled decorators with arrow functions to prevent https://github.com/swc-project/swc/issues/8021. This allows lazy execution of the decorator until it is executed in the initial static block, which exists _within_ the class declaration.

**Related issue:**

 - #8021.


kdy1:
Does this fix #8021? I can work on it if required.

jrandolf:
This doesn't work. Sorry, I commented on the issue instead of here on accident. This PR is what doesn't work, not the issue.

Issue #8025, opened 2023-09-29
Default value not preserved in TypeScript exported binding


magic-akari:
### Describe the bug

Default values are lost when converting TS namespace export bindings.

### Input code

```typescript
namespace Foo {
  export let { a = 1 } = {};
  export const { b = 2 } = {};
}

console.log(Foo.a);
console.log(Foo.b);
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript"
    },
    "target": "es2022"
  },
  "module": {
    "type": "commonjs"
  },
  "isModule": true,
  "minify": false,
}
```

### Playground link

[playground](https://play.swc.rs/?version=1.3.90&code=H4sIAAAAAAAAA8tLzE0tLkhMTlVwy89XqOZSUEitKMgvKlHISS1RqFZIVLBVMFSoBZLVtdYIyeT8vGKQdBJQwgguXcvFBZLIz0nVy8lP1wAaqJeoaY0hlqRpDQBDcZdedgAAAA%3D%3D&config=H4sIAAAAAAAAA0WMOw6EMAxEe04RuaZYpeQOe4goGBREPrKNtBHi7jggls7z5o33zhhY2MNgdj01FEeM9M9KuCZxPyUgtSB7CkXgKo%2F%2B3oijGaUZyPZjbWuvDmIetxXf7%2B1D83yOMaeFXzPw93GFNrzXIYWpKpncyoqOE4bJAnGvAAAA)

### SWC Info output

N/A

### Expected behavior

log 1 and 2

### Actual behavior

```JavaScript 
"use strict";
var Foo;
(function(Foo) {
    ({ a: Foo.a } = {});
    const { b = 2 } = { b: Foo.b } = {};
})(Foo || (Foo = {}));
console.log(Foo.a);
console.log(Foo.b);
```
log undefined

### Version

1.3.90

PR #8029, opened 2023-09-29
fix(es/typescript): Preserve default value of an exported binding in a namespace

magic-akari:
**Related issue:**

 - Closes: #8025


kdy1 (approved):
swc-bump:

PR #8042, opened 2023-10-02
fix(es/minifier): Fix `node-fetch`

kdy1:
**Description:**

The variable mangler is renaming `class AbortSignal` to `class s`. The problem goes away if I pass `mangle: false`.

**Related issue:**

 - https://github.com/vercel/next.js/issues/55682


kdy1:
It seems like this is not a bug of the minifier. I'll try patching `node_modules a bit.

kdy1:
This was not a bug of SWC minfiier

PR #8044, opened 2023-10-02
fix(es/compat): Fix scoping of `explicit-resource-management`

### LINKED ISSUE #8020: incorrect scope of declarations surrounding `using` declarations
### Describe the bug

SWC incorrectly handles the scope of declarations surrounding `using` declarations.

### Input code

```typescript
using foo = null

const bar = 1

console.log(baz())

function baz() {
  return bar
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    },
    "target": "es2022",
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true,
}
```

### Playground link

https://play.swc.rs/?version=1.3.90&code=H4sIAAAAAAAAAyXIPQrAIAwG0D2n%2BEZdCj1AD6OiUggJRLO09O79G9%2FzsUtHU8UGcWaiojImcrJn1p%2FKdWHtIacjxEjUXMrcVfAFTgKsTrc3jK4b1lYZUlQAAAA%3D&config=H4sIAAAAAAAAA22PPQ7CMAyFd04ReWZAGRi4A4eIglsF5U%2B2K1FVvTtJ2hSQ2Oz3PdvPy0kpeLKFm1pKWZpsiJGOvig8RzGvooDMGdmSywLnToUrGoxnbNK6EfApMXaya8FFN8zfu20KmZD511itJo4e%2F24WQyNKzYOsL1q3LA1CSI%2BpTe3P1MCb8QofU09xXATH9z4pNBVtfQNyL9zVGQEAAA%3D%3D

### SWC Info output

(N/A; reproduced using Deno / SWC playground)

### Expected behavior

The transpiled program prints `1`.

### Actual behavior

The `bar` constant is defined inside the `try` block, rendering it inaccessible to the function `baz`, which is defined outside of the `try` block. This results in a `ReferenceError` at runtime.

### Version

1.3.90

### Additional context

When `target` is `ES5`, this actually works correctly, due to the `const` being transformed to a `var`.

As a point of comparison, TSC transpiles this with a `var bar;` statement outside of the `try`, even when `target` is set above `ES5`.

kdy1:
**Description:**

**Related issue:**

 - Closes #8020

### DIFF
--- a/crates/swc/tests/fixture/issues-8xxx/8020/input/.swcrc
+++ b/crates/swc/tests/fixture/issues-8xxx/8020/input/.swcrc
@@ -1,0 +1,19 @@
+{
+    "jsc": {
+        "parser": {
+            "syntax": "typescript",
+            "tsx": false
+        },
+        "loose": false,
+        "minify": {
+            "compress": false,
+            "mangle": false
+        },
+        "target": "es2022",
+    },
+    "module": {
+        "type": "es6"
+    },
+    "minify": false,
+    "isModule": true,
+}
--- a/crates/swc/tests/fixture/issues-8xxx/8020/input/1.js
+++ b/crates/swc/tests/fixture/issues-8xxx/8020/input/1.js
@@ -1,0 +1,9 @@
+using foo = null
+
+const bar = 1
+
+console.log(baz())
+
+function baz() {
+    return bar
+}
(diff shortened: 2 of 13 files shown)


kdy1 (commented):
swc-bump:

PR #8045, opened 2023-10-02
refactor(cli): Make CLI testable/managable with `swc-bump`

kdy1:
**Description:**

It was hard to maintain CLI because actual logic was in `./bindings`.


kdy1 (commented):
swc-bump:
 - swc_cli_impl

Issue #8047, opened 2023-10-02
Literal string starts with `use` are droped unexpectedly


magic-akari:
### Describe the bug

All valid directive should be recognized as directive before first non directive.

```JavaScript
"foo";
"bar";

"use strict";
// ^ This is valid strict directive
function foo(a, a) {
  //         ^^^^ SyntaxError: Duplicate parameter name not allowed in strict context
}
```

### Input code

```typescript
"use client";
"foo";
"use bar"; // this is directive and should be preserved
"use strict";
// All above are directives


function foo() {
    
}

"use hello"; // This is not directive and should be preserved
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript"
    },
    "target": "es2022",
    "keepClassNames": true
  },
  "minify": false,
  "isModule": true,
  "module": {
    "type": "commonjs"
  }
}
```

### Playground link

[playground](https://play.swc.rs/?version=1.3.91&code=H4sIAAAAAAAAA4WOTQoDMQhG955CsmpXc4BZ9Q69QH4cEpBYNJlN6d2bkIEuK4IifO%2FpuhFGLlSb28EdInPMY%2FDqdtw2bLkYjk5FKbZyEvqa0LJ0ThgIX0pGelJaOWta4oSN6IMZfZAZUfoBDACOXscuFYfydsc34Cj4wIJkYpalf176Ku3%2FC187heQjzwAAAA%3D%3D&config=H4sIAAAAAAAAAz2OvQ7CMBCD9zxFdDMDysjKDO9wCleUkj%2FlrhJRlXdv2gY2%2B7MteVVaw8wWbnrtspuMhan8fSdco%2BC3EyAbkG1xWeAI2%2BXcCJY3ydFgczUGBv8Q5btH5icG4p5LWUiNHQQX3VQ7ndAzHcjxI70WT6N61n5kHJSadwc2hZDizPuVptoGaftx9soAAAA%3D)

### Expected behavior

All literal string should be preserve

### Actual behavior

Literal string starts with `use` are droped unexpectedly.

```JavaScript
"use client";
"use strict";
"foo";
// All above are directives
function foo() {}
 // This is not directive and should be preserved

```

### Version

1.3.91

PR #8048, opened 2023-10-02
fix(es/module): Handle directives

magic-akari:
**Description:**

**BREAKING CHANGE:**

- Closes: #8047

### DIFF
--- a/crates/swc_ecma_utils/src/lib.rs
+++ b/crates/swc_ecma_utils/src/lib.rs
@@ -2262,12 +2262,10 @@ pub fn prepend_stmts<T: StmtLike>(
 
 pub trait IsDirective {
     fn as_ref(&self) -> Option<&Stmt>;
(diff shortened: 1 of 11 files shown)


crates/swc_ecma_utils/src/lib.rs
```diff
@@ -2262,12 +2262,10 @@ pub fn prepend_stmts<T: StmtLike>(
 
 pub trait IsDirective {
     fn as_ref(&self) -> Option<&Stmt>;
-    fn is_directive(&self) -> bool {
```
kdy1, line 2265: Can you restore this and deprecate it instead?

kdy1: I don't want to make a breaking change for this.

kdy1 (approved):
swc-bump:
 - swc_ecma_utils

PR #8070, opened 2023-10-05
fix(es/codegen): Fix codegen of type-only imports

ef4:
Type-only imports like:

```ts
import type { A } from 'b';
import { type C } from 'd';
```

currently emit as regular imports:

```ts
import { A } from 'b';
import { C } from 'd';
```

This means that if you're using swc to apply transformations to TS that *don't* include the typescript-to-js transform, you accidentally change the semantics of these import statements.

This PR adds tests for these cases and fixes the emitters.


kdy1 (approved):
swc-bump:
 - swc_ecma_codegen

PR #8081, opened 2023-10-08
perf(es/typescript): Avoid redundant `retain_mu` by mark `type_only`

magic-akari:
**Description:**

Before this pull request: 
There were two calls to `retian_mut` on the module items.

After this pull request: 
The calls of `retian_mut` on `strip_import_export` pass has been eliminated and replaced with `type_only` to mark the module items that need to be removed.
In the subsequent `strip_type` pass, invoke `retian_mut` for batch removal.

A slight improvement in performance is expected.

**BREAKING CHANGE:**


magic-akari:
Closed as performance regression.

PR #8088, opened 2023-10-10
chore(xtask): Add `git reduce core-ver`

kdy1:
**Description:**

This command should be useful for reducing regressions to commits.


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #8091, opened 2023-10-10
feat(css/parser): Store lexed comments

kdy1:
**Description:**

For now, we leave every comment as a leading comment. As logic for determining leading vs trailing is quite complex, I'll do it with another PR.


kdy1 (commented):
swc-bump:
 - swc_css_parser --breaking

PR #8093, opened 2023-10-10
fix(es/minifier): Don't inline functions if `keep_fnames` is enabled

kdy1:
**Description:**

**Related issue:**

 - https://github.com/vercel/next.js/issues/56550


kdy1:
image


`this` and `U` is `undefined`

kdy1:
It seems like it's constructed twice, and those values are `undefined` only on the second construction.
Second call looks like a reflection

kdy1:
image

`u(u)` in the output (broken) vs `u(t)` in the input (below)

image

kdy1:
image

I think the mangling variable renamer is skipping renaming `u1`.

kdy1:
Ah, https://github.com/vercel/next.js/pull/56281 was the cause. I'll update `mangle.json`

kdy1:
The name mangler was correct; The problem was that inliner happily inlined `t`  as `u` in `a(u)`

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

jridgewell (dismissed):

PR #8094, opened 2023-10-10
chore(xtask): Add `--features` to `xtask bench`

kdy1:
**Description:**

 - Add `--features` to `xtask bench`


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #8112, opened 2023-10-13
chore: Publish `v1.3.93` with `swc_core` `v0.85.7`

kdy1:


kdy1 (commented):
swc-bump:
 - dbg-swc

JReinhold:
It looks like this version didn't include releasing `@swc/wasm@1.3.93`, is that on purpose and I'm misunderstanding the setup, or was there an error here?

I think this might be causing some errors with trying to use `@swc/core@1.3.93` in StackBlitz.

https://github.com/swc-project/swc/actions/runs/6503658050/job/17664453856

valentinpalkovic:
@kdy1 Sorry for pinging you directly, but this seems to be a pressing issue, since stackblitz reproductions doesn't work anymore for Storybook, see e.g. https://github.com/storybookjs/storybook/issues/24567

PR #8113, opened 2023-10-13
fix(codegen): remove non_exhaustive on Config

dsherret:
It is incredibly useful to create `Config` using a struct expression in order to be notified about any config changes and to set the config exactly how the consumer would like.

```rs
let config = crate::swc::codegen::Config {
    minify: false,
    ascii_only: false,
    omit_last_semi: false,
    target: ES_VERSION,
    // with a struct expression, I get a compiler error when upgrading swc and
    // can decide how I want this functionality to behave. Additionally, I'm forced
    // to explicitly document the config behaviour I want.
    emit_assert_for_import_attributes: ,
}
```

### DIFF
--- a/crates/swc_ecma_codegen/src/config.rs
+++ b/crates/swc_ecma_codegen/src/config.rs
@@ -5,7 +5,6 @@ use swc_ecma_ast::EsVersion;
 #[derive(Debug, Clone, Copy)]
 #[cfg_attr(feature = "serde-impl", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde-impl", serde(rename_all = "camelCase"))]
-#[non_exhaustive]
 pub struct Config {
     /// The target runtime environment.
     ///


kdy1:
@kwonoj Any thoughts about this? It was a  noticable maintenance burden

crates/swc_ecma_codegen/src/config.rs
```diff
@@ -5,7 +5,6 @@ use swc_ecma_ast::EsVersion;
 #[derive(Debug, Clone, Copy)]
 #[cfg_attr(feature = "serde-impl", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde-impl", serde(rename_all = "camelCase"))]
-#[non_exhaustive]
```
dsherret, line 8: @kdy1 I'm a little confused about the maintenance burden of not having this. It seems like this struct rarely ever changes?

image

Of those, it seems like it's changed one time per year, so it just requires a minor version bump on update, especially if all downstream swc crates do not use the struct expression.

Overall, it's not a big deal to have this, I just personally think it's a much better API for consumers to use the struct expression, which is why I opened this PR.

kdy1: I made fewer changes to avoid breaking changes. https://github.com/swc-project/swc/blob/06a2b444993f760fa23d4b4ebc25201c64cde0fa/crates/swc_ecma_codegen/src/text_writer/basic_impl.rs#L49

I think this should go into the `Config` struct, but I didn't want a breaking change.

kdy1: There was many similar situations IIRC

dsherret: Ok, I thought there's a script in swc that automatically updates all the versions on a breaking change. I'll close this PR.

kdy1: Versions in `Cargo.toml` are automatic, but actually the problem is that I don't have control over `mdxjs-rs` and I have to wait for a long time.

dsherret: Oh, yeah I just re-read your other comment.

PR #8115, opened 2023-10-13
feat(es/typescript): Add `ts_enum_is_mutable` to disable enum inlining

### LINKED ISSUE #8114: Ability to disable TS enum value inlining?
### Describe the feature

Given the following code:

```ts
enum D {
  A,
  B,
}

(D as any).A = 5;
console.log(D.A);
```

swc 1.3.88 outputs 5 and 1.3.89 outputs 0. This is because swc started inlining enums within a module.

```js
var D;
(function(D) {
    D[D["A"] = 0] = "A";
    D[D["B"] = 1] = "B";
})(D || (D = {}));
D.A = 5;
console.log(0);
```

Is there a way to disable this optimization? Although unlikely, I'm concerned it might cause bugs in people's Deno code since it would be a breaking change.

_No response_

### Additional context

_No response_

dsherret:
**Description:**

This adds the ability to disable the enum inlining optimization that was recently introduced as the default.

This only updates the Rust API and nothing else.

**Related issue:**

 - Closes #8114.


magic-akari:
https://github.com/swc-project/swc/blob/06a2b444993f760fa23d4b4ebc25201c64cde0fa/crates/swc_ecma_transforms_base/src/assumptions.rs#L99-L101

Something related.

crates/swc_ecma_transforms_typescript/src/config.rs
dsherret, line 1: This only adds this option to the Rust API for backwards compatibility reasons with Deno. It's not exposed to non-Rust API users of swc.

crates/swc_ecma_transforms_typescript/src/config.rs
```diff
@@ -18,6 +18,26 @@ pub struct Config {
...
     #[serde(default)]
     pub import_export_assign_config: TsImportExportAssignConfig,
+
+    /// An optimization that inlines TS enum member values within
+    /// the same module. This assumes the enum member values are
+    /// never modified.
+    ///
+    /// Defaults to true.
+    #[serde(default = "true_by_default")]
+    pub ts_enum_is_readonly: bool,
```
dsherret, line 28: @magic-akari oh, I've never noticed https://github.com/swc-project/swc/blob/06a2b444993f760fa23d4b4ebc25201c64cde0fa/crates/swc_ecma_transforms_base/src/assumptions.rs#L99-L101

Should it use it instead? I'm not sure what the API would look like here though.

kdy1 (dismissed):
Thanks! Someone told me that it's used by some `dsl2ts` solutions, which sounds like a reasonable usecase.

---


swc-bump: 
 - swc_ecma_transforms_typescript --breaking

kdy1:
Btw, is this urgent? I don't like updating `swc_core` when there's a breaking change 🤣

crates/swc_ecma_transforms_typescript/src/config.rs
magic-akari: It's a JavaScript API for end users.
Since we only change the rust side API, this option is irrelevant.

dsherret:
Nope, not urgent. I can wait to upgrade. Maybe #[non_exhaustive] should be added to this struct as well? (I'm going to add some upgrading instructions that we inspect all the changed configuration in swc on each upgrade)

crates/swc_ecma_transforms_typescript/src/config.rs
```diff
@@ -70,3 +90,7 @@ pub enum ImportsNotUsedAsValues {
 
 #[deprecated = "ImportNotUsedAsValues is renamed to ImportsNotUsedAsValues"]
 pub type ImportNotUsedAsValues = ImportsNotUsedAsValues;
+
+const fn true_by_default() -> bool {
```
magic-akari, line 94: We have to use create a `true_by_default` for `ts_enum_is_readonly`.
Is setting `ts_enum_is_mutable: false` preferable?

kdy1:
I requested permission of `mdxjs-rs`, and the situation must be far better if I have control

crates/swc_ecma_transforms_typescript/src/config.rs
kdy1: I think we need `derive(SerdeDefault)` 🤣 
I'll create it soon.

dsherret: I think so. It would allow me to revert a decent amount of this code. @kdy1 does that sound good?

kdy1: Yes, let's do it

kdy1 (approved):

dsherret:
Thanks @kdy1!

PR #8127, opened 2023-10-16
feat(es/minifier): Inline into the arguments of `new` using seq inliner

Austaras:
**Description:**

**BREAKING CHANGE:**


crates/swc_ecma_minifier/src/compress/optimize/sequences.rs
```diff
@@ -1855,13 +1855,33 @@ impl Optimizer<'_> {
...
+                        if self.merge_sequential_expr(a, &mut arg.expr)? {
+                            return Ok(true);
+                        }
+
+                        if !self.is_skippable_for_seq(Some(a), &arg.expr) {
+                            return Ok(false);
+                        }
+                    }
+                }
+
```
Austaras, line 1884: `flatten` can be used here, but I prefer a less smart solution.

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_minifier

jridgewell (approved):

PR #8133, opened 2023-10-17
feat(es/parser): Implement `AsRef<str>` for `IdentLike`

dsherret:
**Description:** This is a new type and it's helpful to be able to easily get a string from it.


kdy1 (approved):
swc-bump:
 - swc_ecma_parser

PR #8145, opened 2023-10-18
fix(es/minifier): Abort function inliner if `keep_fnames` is `true`

kdy1:
**Description:**

**Related issue:**

 - https://github.com/vercel/next.js/issues/56408


kdy1:
# Investigation

This is a bug of name mangler. `ie` points to different `ie` when the mangler is enalbed.

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

kdy1:
Marking as draft because it fixes the bug but the size regression looks too much.

```
default: 89197 bytes
new: 99631 bytes

```

jridgewell (approved):

Issue #8074, opened 2023-10-07
Roadmap (huge refactorings)


kdy1:
```[tasklist]
### Tasks
- [ ] https://github.com/swc-project/swc/issues/8078
- [ ] Make parser faster by using `TokenKind`
- [ ] https://github.com/swc-project/swc/issues/8026
- [x] Split `swc_ecma_transforms_compat` into multiple crates
```

kdy1:
Closing as done. I'll try rewriting parser from scratch.

PR #8153, opened 2023-10-19
fix(es/utils): Fix regression of `RefRewriter`

kdy1:
**Description:**

**Related issue:**

 - Found while invetigating https://github.com/swc-project/swc-ecosystem-ci/actions/runs/6581045465/job/17880208108


kdy1:
cc @magic-akari Can you take a look?

kdy1 (commented):
swc-bump:
 - swc_ecma_utils

kdy1:
cc @magic-akari Is this the correct fix?

magic-akari (approved):

crates/swc_ecma_utils/src/lib.rs
```diff
@@ -3063,6 +3063,10 @@ where
     }
 
     fn visit_mut_var_declarator(&mut self, n: &mut VarDeclarator) {
+        if !n.name.is_ident() {
```
magic-akari, line 3066: We should remove the whole `visit_mut_var_declarator` function.
We should only skip the ident declartion.

magic-akari:
https://github.com/swc-project/swc/blob/cea140eb5ade3d2abfb4c2a1fc9a57b6418db18d/crates/swc_ecma_utils/src/lib.rs#L3070-L3079

The key issue here is that we hope that when we access Pat::Ident, we get the Ident ref, not the ident decl

```JavaScript
var a = "value-a", b = "value-b";
//  ^^^ decl       ^^^ decl

var { [a]: d, c } = { "value-a": 1, b };
//     |   |  ^^^ decl              ^^^ ref
//     |   |- decl
//     |
//     |-- ref

var [e = a] =     [a];
//   |   ^^^ ref   ^^^ ref
//   |- decl
```

I think that the Pat in a declaration name and the Pat in an assignment expression are different.
In an assignment expression, `Pat::Ident` can be safely replaced with `Pat::Expr`,
but you can't do that in a declaration statement.

Maybe we should wrap all ident ref cases in an IdentRef struct  #8074.

PR #8164, opened 2023-10-21
refactor(es/minifier): Simplify analyzer context

Austaras:
**Description:**

**BREAKING CHANGE:**


Austaras:
This PR eagerly mark idents as assigned, this may add some performance cost, but generally speaking LHS is very small.

crates/swc/tests/tsc-references/controlFlowAssignmentPatternOrder.2.minified.js
```diff
@@ -37,19 +37,19 @@
 }// same as above but on left of a binary expression
 {
     let b, a = 0;
-    [{ 1: b } = [
+    [{ [a = 1]: b } = [
```
kdy1, line 40: Oh this is a bugfix

crates/swc_ecma_minifier/src/compress/optimize/inline.rs
```diff
@@ -254,7 +255,7 @@ impl Optimizer<'_> {
                     indexed_with_dynamic_key,
                     usage_count,
                     has_property_access,
-                    has_property_mutation,
+                    property_mutation_count: has_property_mutation,
```
kdy1, line 258: Please rename this one, too.

crates/swc/tests/tsc-references/controlFlowAssignmentPatternOrder.2.minified.js
Austaras: Yeah. This PR arises from the problem that assign happens after the evaluation of LHS and RHS, so it's impossible to report reference and assign at the same time.

jridgewell (approved):

kdy1 (approved):
swc-bump:
 - swc_ecma_minifier

PR #8171, opened 2023-10-23
fix(es/typescript): Support `ts_enum_is_mutable` with const enums

dsherret:
**Description:** Bug I introduced in https://github.com/swc-project/swc/pull/8115/files (it was not implemented correctly)

### DIFF
--- a/crates/swc_ecma_transforms_typescript/tests/strip.rs
+++ b/crates/swc_ecma_transforms_typescript/tests/strip.rs
@@ -848,7 +848,7 @@ test!(
     _ts_decorate([
         DefineAction()
     ], Child.prototype, "action", void 0);
-
+    
     "#,
     ok_if_code_eq
 );
@@ -4469,7 +4469,7 @@ test!(
     import { createRequire as _createRequire } from "module";
     const __require = _createRequire(import.meta.url);
     const foo = __require("foo");
-
+    
     foo();
     "#
 );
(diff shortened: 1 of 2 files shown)


crates/swc_ecma_transforms_typescript/tests/strip.rs
```diff
@@ -4581,5 +4586,7 @@ test!(
     })(D || (D = {}));
     D.A = 5;
     console.log(D.A);
+    var E;
```
dsherret, line 4589: Improvement for swc in the future might be to not emit this for const enums.

kdy1 (approved):
swc-bump:

crates/swc_ecma_transforms_typescript/tests/strip.rs
```diff
@@ -4567,19 +4567,57 @@ test!(
...
+    console.log(G.A);
+    var H;
+    (function(H) {
+        H[H["A"] = 2] = "A";
+    })(H || (H = {}));
+    var I;
+    (function(I) {
+        I[I["A"] = H.A] = "A";
+    })(I || (I = {}));
+    console.log(I.A);
```
dsherret, line 4621: This is not ideal, but it's ok output. It's a rare edge case for someone to use an enum in a const enum.

kdy1 (approved):

PR #8187, opened 2023-10-27
feat(es/base): Add `TransformConext` to reduce breaking changes

### LINKED ISSUE #8186: `swc_ecma_transforms_base::TransformContext`
All transform functions for compatibility and bugfixes should accept it.

This can be a bit cumbersome, but this will reduce breaking changes due to more assumption required.

```rust

pub struct TransformContext<C> {
    pub comments: C,
    pub assumptions: Assumptions,
    pub unresolved_mark: Mark.
}


```

I'm not sure about the extra fields

kdy1:
**Description:**

This will force users to create needlessly many input, but this will reduce breaking change.
I want to create an `swc-rs-ecosystem-ci` and this is groundwork for it.

**Related issue:**

 - Closes #8186


crates/swc_ecma_compat_es2015/src/block_scoping/mod.rs
```diff
@@ -37,7 +38,14 @@ mod vars;
 ///    });
 /// }
 /// ```
-pub fn block_scoping(unresolved_mark: Mark) -> impl VisitMut + Fold {
+pub fn block_scoping<C>(
```
kdy1, line 41: This became much verbose

crates/swc_ecma_compat_es2015/src/classes/mod.rs
```diff
@@ -31,14 +30,25 @@ use self::{
 mod constructor;
 mod prop_name;
 
-pub fn classes<C>(comments: Option<C>, config: Config) -> impl Fold + VisitMut
+pub fn classes<C>(
```
kdy1, line 33: This is actually cleaner, and it also allows future optimization

crates/swc_ecma_compat_es2015/src/classes/mod.rs
```diff
@@ -31,14 +30,25 @@ use self::{
...
+) -> impl Fold + VisitMut
 where
-    C: Comments,
+    C: Comments + Clone,
 {
     as_folder(Classes {
         in_strict: false,
         comments,
-        config,
+        config: Config {
```
kdy1, line 46: The constructor drops `Assumptions` early to avoid making difference of memory access

kdy1 (commented):
I want to hear opinions from others.

The problem I'm trying to solve is `swc-rs-ecosystem-ci`, which is about slowing the release process of `swc_core` down and preventing regression of it.

But swc exposes an enormous amount of API, especially for transforms. This means preventing a breaking change is hard. To reduce the direct API surface, I'm experimenting with a type `TransformContext`, which should be passed to almost all transforms. Because each transform reads relevant **assumptions** from the context, you don't need to pass fields of assumptions like at the moment.

This will prevent breaking changes due to added optimizations, where a new assumption is required.

Of course, this is not enough to prevent all breaking changes, and I'll make other PRs to reduce the API surface, and breaking chanages.

# Alternatives

## `#[non_exhaustive]` on every config stfuct.


---

Pros:
 - It's not a breaking change if we implement a new optimization using a new assumption.
 - Users do not need to create read API docs to figure out the types.

Cons:
 - APIs become compilcated.
 - A user has to create `TransformContext` even if it's not going to be used at all.

kdy1:
cc @kwonoj @magic-akari @Austaras

crates/swc_ecma_compat_es2015/src/arrow.rs
```diff
@@ -57,7 +58,14 @@ use swc_trace_macro::swc_trace;
 /// };
 /// console.log(bob.printFriends());
 /// ```
-pub fn arrow(unresolved_mark: Mark) -> impl Fold + VisitMut + InjectVars {
+pub fn arrow<C>(
```
magic-akari, line 61: Does this mean that every visitor have to add a generic parameter of type C?

magic-akari (dismissed):
We should do it

crates/swc_ecma_compat_es2015/src/arrow.rs
kdy1: Yes, at least all visitors accept `TransformContext`

Austaras:
Do people really relies on separate compat passes? Maybe only change esxxx and preset-env would be enough?

kdy1:
Some do, but I think it makes sense

kdy1:
But then the problem is that it's a _technically_ breaking change.

PR #8192, opened 2023-10-29
chore(ci): Remove benchmark action

kdy1:
It's not used at the moment, while creating noticeable cost


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #8205, opened 2023-11-01
feat(es/minifier): Respect inline level and preserve native names

kdy1:
**Description:**

Note: Preserving native names is a hack, but it's used by `terser` and it's the only way to preserve the name of `class AbortSignal` while mangling without `keep_classnames: true`. We can special case `AbortSignal`, but let's just follow `terser`.

**Related issue:**

 - https://github.com/vercel/next.js/pull/57904


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base

Issue #8204, opened 2023-11-01
const enum exports are not consistently compiled


aweary:
### Describe the bug

This appears to have regressed in the `1.3.89` release. If you have a `const enum` that is exported like `export const enum` then `swc` does not attempt to remove the enum definition, but if you have a `const enum` that is exported in a separate statement (like an `export default` or even just a separate `export` statement) then it DOES remove the enum and that breaks imports.

### Input code

```typescript
// This does NOT get removed and is exported correctly
export const enum Foo { VALUE = 1 }


// This DOES get removed and breaks the export
const enum Bar { VALUE = 1 }
export default Bar
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "transform": {
      "react": {
        "refresh": {},
        "runtime": "automatic"
      }
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.3.96-nightly-20231031.1&code=H4sIAAAAAAAAA2WNMQ7CMBAE%2B3vF%2FsDiARQgQoVIQaA38UEQiQ%2BdLwiE%2BDtGuEBQ7uxq1jk03SkhCCes6wZHNigPcuUAHwNyx7eLqOXciiq31t%2FpgzKIycBxHLAUwQO72WpbYYoJnkTkinxRV5s%2F8V7ZnxOs43JAX7a51x9beQx88GNv78ELC5n%2BCbsAAAA%3D&config=H4sIAAAAAAAAA1WQMQ6DMAxFd06BPLO2Q%2B%2FQQ1ipoalIgmxHKkLcvUkgFLbkfdvP8tK0LXzEwKNd0jN9JmQhPv6JyOwVv4mAzhOJYTspdDVVyZFypELWLQBl9NIHdudJTGj0BArqmeSd4dqdcPRqHWUpRg0O1RrY4%2FUqQh4oDwWS274WjCFIbu5xFNqZs97283kdE9yU5HItzKXoh%2FHob6q0KMGFVyzhfrB8lE1%2Fh39RlR2DwcqzdpZrrT8SUsOrfAEAAA%3D%3D

### Expected behavior

Any `const enum` that is exported should behave the same, whether its a default export or exported in a separate statement.

### Version

[...]

PR #8208, opened 2023-11-02
fix(es/typescript): Preserve const enum for named export

magic-akari:
**Related issue:**

- Closes: #8204


kdy1 (approved):
Thanks!

---

swc-bump:

PR #8213, opened 2023-11-03
perf(es/typescript): Visit ts import/export only once

magic-akari:
**Description:**

Move `visit_mut_for_ts_import_export` into `visit_mut_module` to make sure we only visit the top-level items.


kdy1 (approved):
Thank you!

---

swc-bump: 

PR #8222, opened 2023-11-06
perf(es/minifier): Remove needless operations for char freq analysis

kdy1:
**Description:**

This skips a string search operation, which is quite costly, without affecting the output.

**Related issue:**


kdy1 (commented):
swc-bump: 
 - swc_ecma_codegen

PR #8245, opened 2023-11-08
fix(es/decorators): Minimize class declarations transformation

magic-akari:
**Description:**

**BREAKING CHANGE:**


kdy1 (commented):
I think we should check if the mutability of the variable (let/class) is identical, and if not we should add an assumption entry.

I'll check it using chrome

kdy1:
```js

class Foo {
    static {
         Foo = 1;
    }

}
```


```js

let Foo = class Foo {
    static {
         Foo = 1;
    }

}
```


```js

class Foo {
    static field = Foo = 1;

}
```


```js

let Foo = class Foo {
    static field = Foo = 1;

}
```

kdy1 (approved):
Thank you!

---

swc-bump:

kdy1:
Oops there's a merge conflict

kdy1 (approved):

magic-akari:
There are some broken snapshots introduced in #8257.

kdy1:
Can you update them too?

kdy1:
`Ok` action of CI is bit broken, but not sure about the reason

kdy1:
🤣 CI is so broken so it passed again

PR #8247, opened 2023-11-09
perf(css/parser): Use `AtomStoreCell`

kdy1:


kdy1 (commented):
swc-bump:
 - swc_css_parser

PR #8255, opened 2023-11-09
fix(es/minifier): Enable seq inliner for const declarations

Austaras:
**Description:**

**BREAKING CHANGE:**


crates/swc/tests/fixture/issues-5xxx/5112/output/index.js
```diff
@@ -1,7 +1,6 @@
 function c(r, n) {
-    let e = new Uint8Array(4 * r * r);
-    return {
-        e,
+    return new Uint8Array(4 * r * r), {
```
kdy1, line 2: This seems wrong

Austaras: This is wrong. It's caused by `unused: false`. However after fix, it's still wrong -- under `unused: false` var declaration should not be drop, but that's another (low proority) issue.

crates/swc/tests/tsc-references/exportAssignmentOfExportNamespaceWithDefault.2.minified.js
```diff
@@ -2,7 +2,5 @@
 // https://github.com/microsoft/TypeScript/issues/39149
 Object.defineProperty(exports, "__esModule", {
     value: !0
-});
-const _interop_require_default = require("@swc/helpers/_/_interop_require_default"), _a = /*#__PURE__*/ _interop_require_default._(require("a"));
-(0, _a.default)();
+}), (0, /*#__PURE__*/ require("@swc/helpers/_/_interop_require_default")._(require("a")).default)();
```
kdy1, line 5: This is different from the original code, but it looks like a bugfix.

crates/swc_ecma_minifier/tests/benches-full/d3.js
```diff
@@ -1381,7 +1381,7 @@
             return (1 === (a = isNaN(a) ? 1 : Math.max(0, Math.min(1, a))) ? "hsl(" : "hsla(") + (this.h || 0) + ", " + 100 * (this.s || 0) + "%, " + 100 * (this.l || 0) + "%" + (1 === a ? ")" : ", " + a + ")");
         }
     }));
-    const radians = Math.PI / 180, degrees = 180 / Math.PI, t0 = 4 / 29, t1 = 6 / 29, t2 = 3 * t1 * t1, t3 = t1 * t1 * t1;
+    const radians = Math.PI / 180, degrees = 180 / Math.PI, t0 = 4 / 29, t1 = 6 / 29, t2 = 3 * (6 / 29) * (6 / 29), t3 = 6 / 29 * (6 / 29) * (6 / 29);
```
kdy1, line 1384: This will degrade performance (although it's very minor)

Austaras: Yes, but this is well beyond the scope of this PR.

crates/swc/tests/tsc-references/exportAssignmentOfExportNamespaceWithDefault.2.minified.js
Austaras: This is the same as old code. both `_interop_require_default` and `_a` are inlined.

crates/swc_ecma_minifier/tests/benches-full/d3.js
kdy1: Make sense. Thanks!

kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_minifier

kdy1:
Can you update the test references? (I think those are unrelated to this PR, though)

kdy1 (approved):

PR #8276, opened 2023-11-12
fix(es/module): support resolving exports from package.json

eandre:
This adds support in `swc_ecma_resolver::node::NodeModulesResolver` to resolve import paths from the `exports` field in `package.json`. This is needed to resolve the correct file path for packages that use `exports`.

By default it uses the export conditions based on the `TargetEnv`. It also adds a new constructor `NodeModulesResolver::with_export_conditions` that allows adding additional conditions.


kdy1 (changes requested):
Please add some tests

PR #8277, opened 2023-11-13
feat(es/minifier): Swap bin expr to save paren

Austaras:
**Description:**

**BREAKING CHANGE:**


magic-akari (approved):

kdy1:
I think we need a way to verify this without publishing...
It would be nice if there were some vendors helping us testing swc

Austaras:
I believe this is generally fine. Both terser and gcc has similar feature.

kdy1 (approved):

kdy1 (approved):
swc-bump:
 - swc_ecma_minifier

PR #8291, opened 2023-11-15
refactor(common): Remove string_cache dependency

dsherret:
**Description:**

I believe we can remove this now because of hstr?

(I don't need this. I just noticed it was not used in the swc repo anymore. Maybe it's used elsewhere though?)

**BREAKING CHANGE:** Removes `EqIgnoreSpan` and `TypeEq` implementations for `string_cache::Atom`


kdy1 (approved):
swc-bump:
 - swc_common

PR #8293, opened 2023-11-15
fix(es/minifier): Remove hack for built-in class names

kdy1:
**Description:**

Regarding https://github.com/vercel/next.js/issues/55682, I made a mistake while investigating. I assumed it's `swcMinify`-only but it seems like `node-fetch` is fundamentally incompatible with our minification options.

I thought `node-fetch` works with terser, but it did not, and it **should not**. But as I thought `terser` works, I assumed that `terser` has a hack for `AbortSignal`. And that's how I fixed `swcMinify: true`. The correct fix is `keep_classnames: true` but it will result in +~10% bundle size bloat.
At first, I used the correct way (`keep_classnames: true`), but it resulted in the bundle size bloat, so I mimicked the hack of terser with https://github.com/vercel/next.js/pull/57904.

While working on other minification issues, I found that there's no such hack in `terser` and I need to remove the hack in the SWC minifier.
The repro in https://github.com/vercel/next.js/issues/55682 does not work with `swcMinify: false`, even with next@14. The problem of minifier and node-fetch was not exploited before `serverMinification: true`.

**Related issue:**

 - https://github.com/brix/crypto-js/issues/415


kdy1:
Cause is https://github.com/swc-project/swc/issues/8205

kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base

PR #8298, opened 2023-11-16
feat(es/parser): Add `peek_ahead` to the input buffer

kdy1:
**Description:**

cc @magic-akari 


magic-akari:
- #8309 

With a few tricks, I found that there isn't a strong imperative for utilizing `peek_ahead`.
Let's wait before we make a decision.

Issue #8300, opened 2023-11-17
SWC output causes accessing `this` before `super()`


JSerFeng:
### Describe the bug

If I have a class Foo extends Bar, access `this` inside a fn decl, the output code declare a _this variable to store this, however it should not access this before super();

### Input code

```typescript
class Foo extends Bar {
    constructor() {
        super();
        for (const i of [1]) {
            test(() => {
                this.i = i
            })
        }
    }
}

function test(f){
    f()
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "env": {
    "targets": [
        "chrome >= 51",
        "edge >= 15",
        "firefox >= 54",
        "safari >= 10",
        "ios_saf >= 10"
      ],
  },
  "minify": false,
  "isModule": true
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.3.98-nightly-20231117.1&code=H4sIAAAAAAAAA12MsQoCMRBE%2B3zFlEkjWMtZWPgTYnHkshiQrGQ3IEj%2B%2FXKX41BfsbDzhvHPUQRXZoS3hjQJLmPGx6DhOYnm4pWzdVu2IOUVWnLaA%2BIMu7YRwYTb8f7dX9AgatvKcP4Tq3xEOUQMiD%2Bquv2tpt9qDJXkNXLqk%2BT6HFln6gy%2Bp95ezQAAAA%3D%3D&config=H4sIAAAAAAAAA1WQTQ6CMBCF95yi6ZqFJOLCRG%2FgCYwxDUyxhFIyUwyEcHfbWrDuOu978%2FO6ZIzxlip%2BZot7umIQSIB77RSaeysmp3CotKAK1WB5vtGWPJKiIwjS%2BiW8M4ZgI1HTqldyTmdXRg8IRP9GbxV908H%2F5CxO59rUY4DxZjsPEO6jE99N0L8Th8AGrF9zjxv88hcaDex6YWWxB%2FKddRPUokxVqRCkmYL9mAISUqAKDYdUV4aejkUQ9Uf%2BS7H9xp6cK7pt0SyOkK0fVhicpaABAAA%3D

### Expected behavior

this should be accessed after super() call

### Actual behavior

```javascript
class Foo extends Bar {
[...]

PR #8311, opened 2023-11-19
fix(es/compat): Insert loop Var into closest place

magic-akari:
**Description:**

**BREAKING CHANGE:**

- Closes: #8300

### DIFF
--- a/crates/swc/tests/fixture/issues-8xxx/8300/output/3.js
+++ b/crates/swc/tests/fixture/issues-8xxx/8300/output/3.js
@@ -1,0 +1,13 @@
+class Foo extends Bar {
+    constructor(){
+        var _this = this, _loop = function(i) {
+            setTimeout(()=>{
+                _this.i = i;
+            });
+        };
+        for (var i of [
+            1
+        ])_loop(i);
+        super();
+    }
+}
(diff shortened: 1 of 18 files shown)


Austaras:
This doesn't seem to be the best solution. Consider following example
```js
class Foo extends Bar {
    constructor() {
        for (const i of [1]) {
            setTimeout(() => {
                console.log(this)
            });
        }
        super();
    }
}
```

magic-akari:
> This doesn't seem to be the best solution. Consider following example
> 
> ```js
> class Foo extends Bar {
>     constructor() {
>         for (const i of [1]) {
>             setTimeout(() => {
>                 console.log(this)
>             });
>         }
>         super();
>     }
> }
> ```

@Austaras Test cases updated. IMO, it's OK

Austaras:
It's not. The problem arises in access this before call super.

crates/swc/tests/fixture/issues-8xxx/8300/output/3.js
```diff
@@ -0,0 +1,13 @@
+class Foo extends Bar {
+    constructor(){
+        var _this = this, _loop = function(i) {
```
magic-akari, line 3: 🤔 This is a bug.

Austaras: > Uncaught ReferenceError: must call super constructor before using 'this' in derived class constructor

PR #8314, opened 2023-11-20
fix(es/ast): Fix definition of `SetterProp`

### LINKED ISSUE #8157: AST of `SetterProp` is wrong
But let's fix this after the conf. This breaks Wasm plugins

Related: https://github.com/swc-project/swc/issues/8156

### LINKED ISSUE #8377: Unexpected enum variant in AssignExpr
### Describe the bug

Expressions in `left` of `AssignExpr` are `PatOrExpr::Pat(Pat::Expr(_))` instead of `PatOrExpr::Expr(_)`, but this looks inconsistent with documentation.

https://github.com/swc-project/swc/blob/7327f257d2133ba2507990a73a11a2f58dfd826c/crates/swc_ecma_ast/src/pat.rs#L34-L36

I'm not sure if it's the documentation or the implementation that might be incorrect, or if I might have misunderstood something.

### Input code

```typescript
o.n = 42;
```

### Config

_No response_

https://github.com/shqld/repro/tree/swc/8376

### SWC Info output

_No response_

### Expected behavior

Either documentation or implementation is corrected.

### Actual behavior

_No response_

### Version

swc_ecma_ast = "0.110.12", swc_ecma_parser = "0.141.30"

### Additional context

_No response_

kdy1:
**Description:**

**Related issue:**

 - Closes #8157
 - Closes #8377


kdy1 (commented):
swc-bump:
  - swc_ecma_ast --breaking

crates/swc_ecma_parser/src/parser/object.rs
```diff
@@ -351,28 +351,15 @@ impl<I: Tokens> ParseObject<Box<Expr>> for Parser<I> {
...
                                 )
                                 .map(|v| *v)
-                                .map(
-                                    |Function {
-                                         mut params, body, ..
-                                     }| {
-                                        params.retain(|p| match &p.pat {
-                                            Pat::Ident(p) => p.sym != "this",
-                                            _ => true,
-                                        });
```
kdy1, line 361: Removal of this code is the `fix` of this Pr.

kdy1 (commented):
---

crates/swc_ecma_ast/src/prop.rs
```diff
@@ -76,7 +75,7 @@ pub struct GetterProp {
 pub struct SetterProp {
     pub span: Span,
     pub key: PropName,
-    pub param: Box<Pat>,
+    pub params: Vec<Param>,
```
michaelangeloio, line 78: Curious, just for my own learning, how are these breaking changes documented?

Austaras, line 78: What about
```rs
pub this_param: Option<Box<Pat>>,
pub param: Box<Pat>,
```

kdy1: It looks better. @magic-akari Thoughts?

magic-akari: Yes, It looks better.
Should we apply the same tricks to other `this` parameters as well?

magic-akari: However, its AST should be 

```rust
pub this_param: Option<Span>,
pub param: Box<Pat>,
```
 right?

kdy1: I think it should be `Option<Box<Pat>>` because of `this: ClassName`

magic-akari: > I think it should be `Option<Box<Pat>>` because of `this: ClassName`

That makes sense. We should generate raw TypeScript code during the code generation phase.

magic-akari: Or maybe:

```rust
pub struct ThisParam {
    pub span: Span,
    pub type_ann: Option<Box<TsTypeAnn>>,
}
```

dsherret: Personally, I feel like this parameters shouldn't be treated any differently from regular parameters. It creates more complication in the AST to have additional nodes and properties. With a single vector it's easy to check for a this parameter.

magic-akari: I extracted the paramter `this` from the parameter list in `oxc`.  This simplifies subsequent linters and transformers.
https://github.com/oxc-project/oxc/pull/1728

Using arrays consistently to describe function parameters can simplify handling and improve error tolerance.
It is feasible to extract `this` while ensuring that all function parameters are arrays. However, these are two separate matters.

Issue #2992, opened 2021-12-07
Non-leading `@jsxImportSource` pragma is emitted as import


nayeemrmn:
Using Deno with `--no-check`.

### Input code

```typescript
console.log(1);
/** @jsxImportSource foo */
console.log(<div></div>);
```

### Expected

In tsc, the pragma comment is ignored because there is code before it, and it must be a leading comment.
```ts
console.log(1);
console.log(/*#__PURE__*/ React.createElement("div", null));
```

### Actual

```ts
import { jsx as _jsx } from "https://esm.sh/preact/jsx-runtime";
console.log(1);
console.log(/*#__PURE__*/ _jsx("div", {}));
```

### Version

swc_ecma_transforms_react 0.53.0

kdy1:
This is intended, as mdx-loader emits non-leading `/** @jsxImportSource foo */`

nayeemrmn:
> This is intended, as mdx-loader emits non-leading `/** @jsxImportSource foo */`

I think this should be an extended compatibility option, since it doesn't comply with tsc. Or it's just an mdx-loader bug. In Deno this case can lead to something like a panic because these aren't picked up at the static "prepare module load" step but the import is emitted.

kdy1:
But babel supports this by default.

nayeemrmn:
@kdy1 It doesn't look like it: https://www.typescriptlang.org/play?#code/MYewdgziA2CmB00QHMAUBGAlAbgFAHoAqQgAgAEArCADwEkBbABxACcAXAZRAFcXhYSAMxAgShfLlCQYCJGgA8AEwCWANwB88-Co04gA whereas [when the pragma is on top](https://www.typescriptlang.org/play?#code/PQKhAIAECsGcA8CSBbADgewE4BcDK6BXTAYwFNwAzddcEYAKGPQDtZ0AbUgOnfQHMAFAEYAlAG5GLNpx78BAHgAmASwBuAPnnAVG8UA), `React.createElement()` is replaced with `_jsx()` showing the pragma is respected only when leading.

One thing I couldn't figure out is that swc _does_ seem to be scanning leading comments only: https://github.com/swc-project/swc/blob/017392aa5b681e4a8240259e7e26709b7160d623/crates/swc_ecma_transforms_react/src/jsx/mod.rs#L929-L934 so I don't know how this is the current behaviour.

kdy1:
Why are you linking to the typescript playground?

kdy1:
[...]

PR #8318, opened 2023-11-20
fix(es/typescript): Handle shebang with jsx pragma

dsherret:
**Description:**

This handles a shebang with a jsx prag. Previously `h` would be removed as an import in this case.

**Related issue:**

 - https://github.com/denoland/deno/issues/21258


crates/swc_ecma_transforms_typescript/src/typescript.rs
```diff
@@ -183,7 +183,12 @@ where
         }
 
         if !self.config.verbatim_module_syntax {
-            let span = n.span;
+            let span = match n.shebang.is_some() {
+                true => n
+                    .span
+                    .with_lo(n.body.first().map(|s| s.span_lo()).unwrap_or(n.span.lo)),
+                false => n.span,
+            };
```
dsherret, line 191: FWIW, I think probalby the jsx prag detection code should be consolidated, but I believe the jsx prag detection code in swc_ecma_transforms_react is doing too much work as it checks every module item instead of just the top of the file: https://github.com/swc-project/swc/blob/5fa2ed4bfdb49f77d4cb77004e1f61c8e2c36b64/crates/swc_ecma_transforms_react/src/jsx/mod.rs#L1033-L1040

dsherret:
A potential fix to CI issues: https://github.com/actions/setup-node/issues/899#issuecomment-1819151595

crates/swc_ecma_transforms_typescript/src/typescript.rs
```diff
@@ -183,7 +183,12 @@ where
         }
 
         if !self.config.verbatim_module_syntax {
-            let span = n.span;
+            let span = match n.shebang.is_some() {
```
kdy1, line 186: Please change this to `if + else`

crates/swc_ecma_transforms_typescript/src/typescript.rs
kdy1: https://github.com/swc-project/swc/issues/2992#issuecomment-990739450

It's due to mdxjs

magic-akari (approved):

kdy1 (approved):
Thanks!
---

swc-bump:

Issue #1584, opened 2021-04-17
[Tracker] TypeScript: Multi-file mode


kdy1:
This issue is a tracker for the decorator metadata and const enum support.

Initially, the decorator metadata was requested with #1160 and I implemented a very simple dependency analyzer with #1500.
But it has lots of issues. This issue tracks it.
This issue also tracks const enum, because multi-file mode also allows const enums.

# Decorator metadata

 - [ ] Decl + named export.
 - [ ] `export`s from other module.
 - [ ] Circular dependency.

# Const enum

 - [ ] Basic implementation

But I'm postponing these tasks because I'm not sure what's the correct way to implement them.
I'm not sure about the way to avoid degrading performance while reusing codes of bundler.
(It would be dead simple to implement both of them if we could use the system of the bundler)

matthew-dean:
Will this also fix the issue like "[name] is not exported by [module]" when the export is a type?

kdy1:
Closing as we are not going to implement `project mode` of `tsc`

PR #8320, opened 2023-11-20
fix(es/decorators): Resolve enum for `design:returntype`

ZakisM:
**Description:**

Resolve enum types in `design:returntype`.


kdy1:
cc @magic-akari What's your thought about this? We cannot support enums in other files due to some fundamental reasons.

magic-akari:
> cc @magic-akari What's your thought about this? We cannot support enums in other files due to some fundamental reasons.

I have no idea. I suppose that tsc will emit error for this case.

ZakisM:
I don't think `isolatedModules: true` will highlight this as an issue. This change was meant to match the behaviour that we are [currently](https://github.com/swc-project/swc/blob/97d57f0e4fcaca2e03241ddb4fbe19f449b33ecd/crates/swc_ecma_transforms_proposal/src/decorators/legacy/metadata.rs#L202) doing for `design:type` where if the enum is declared in the same file then the type is resolved correctly. For enums imported from other files we will need to re-look at https://github.com/swc-project/swc/issues/1584.

kdy1 (approved):
Thanks!

---


swc-bump:

PR #8325, opened 2023-11-21
build(cargo): Update `vergen` to `v8`

Redfire75369:
**Description:**
Updates `vergen` build dependency in `swc_core` to v8. This reduces the amount of dependency duplication. In my case in particular, it allows me to remove `syn` v1 from my [repository](https://github.com/redfire75369/spiderfire).

**BREAKING CHANGE:** None


kdy1 (approved):
swc-bump:
 - swc_core

PR #8343, opened 2023-11-25
fix(es/parser): Fix conditional compilation

arHSM:
**Description:**

`swc_ecma_parser` crate has a `typescript` feature which enables TS parsing, the `Syntax::Typescript` variant is behind the `typescript` feature, compiling without this feature fails as `Syntax::Typescript` is referenced at many places without `#[cfg(feature = "typescript")]`

None

### DIFF
--- a/crates/swc_ecma_parser/src/lib.rs
+++ b/crates/swc_ecma_parser/src/lib.rs
@@ -171,6 +171,7 @@ impl Syntax {
                 auto_accessors: true,
                 ..
             }) => true,
+            #[cfg(feature = "typescript")]
             Syntax::Typescript(_) => true,
             _ => false,
         }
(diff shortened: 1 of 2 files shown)


crates/swc_ecma_parser/src/lib.rs
```diff
@@ -229,7 +235,11 @@ impl Syntax {
     /// Should we parse typescript?
     #[cfg(feature = "typescript")]
     pub const fn typescript(self) -> bool {
-        matches!(self, Syntax::Typescript(..))
+        match self {
+            #[cfg(feature = "typescript")]
```
kdy1, line 239: Please remove this line

kdy1 (commented):
swc-bump:
 - swc_ecma_parser

crates/swc_ecma_parser/src/lib.rs
arHSM: apologies, overlooked the already present attribute

kdy1 (approved):

Issue #8356, opened 2023-11-29
Module parse failed: Cannot use 'yield' as identifier inside a generator


jimkyndemeyer:
### Describe the bug

The Input code below produces the following error:
```
Module parse failed: Cannot use 'yield' as identifier inside a generator (38:25)
 * ../../node_modules/next/dist/build/webpack/loaders/next-swc-loader.js
You may need an additional loader to handle the result of these loaders.
|         },
|         async *evaluateAsyncYield (signal) {
>             return await yield* nestedAsyncYield();
|         },

```

### Input code

```typescript
const obj = {
    async *asyncYield() {
        return await (yield* nestedAsyncYield());
    }
}
```

### Config

```json
No config. Using Next.js 14.0.3.
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.3.94-nightly-20231016.1&code=H4sIAAAAAAAAA0vOzysuUchPylKwVahWgAJeLhgrsbgyL1lBC0xFZqbmpGhoKlQjpEGgKLWktChPIbE8MbNEQaMSpEhLIS%2B1uCQ1xRFJm6Y1QlsthFlrDQDD%2BPwqfgAAAA%3D%3D&config=H4sIAAAAAAAAA32QTQ6CMBCF95yCzNoFuFDjHTyDmdSBVAttOpVICHe3BYqgxE1%2F3vteJ31dkqZwZwHntPNHfzFomex89wq3tcOXV8C1hlhYaRzsous4WAUqpkHqRwcc2pJcSBHvs%2Fw4JUBpzRQTk1bJWhbtcqbQlbHEvAYDinWpaIGGUdooakh9w956EJmrUMhcY0U%2Fz0Wi%2BOsaKxt05HdttiBJpw2VsUAr82zVju8nievQFFT69lx8aOh4bO0AHygWNA8ByZeYdPZJSf8G9gkShcsBAAA%3D

### Expected behavior

That the SWC output AST preserves the parenthesis instead of breaking the semantics of the code.

By not preserving the parenthesis, the return statement semantics become invalid. I've added TypeScript AST example to illustrate this problem, with the top being the original code, and the bottom the output from SWC:

![image](https://github.com/swc-project/swc/assets/15837712/d5333e0d-95b8-479d-9530-2cbe5b45e651)

### Actual behavior

See the error above which indicates an invalid AST is created as output.

### Version

@next/swc-win32-x64-msvc 14.0.3

PR #8357, opened 2023-11-29
fix(es/fixer): Wrap yield expression in await expression

magic-akari:
**Related issue:**

- Closes: #8356


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_transforms_base

PR #8364, opened 2023-12-01
feat(es/codegen): Add the option to ignore typescript

l4yton:
**Description:**
This change adds the option to ignore TS when printing.

Currently, to print an AST that contains TS nodes as valid JS, one would first have to transform it. This change makes it possible to directly print valid JS while also retaining the original AST with the TS nodes.

### DIFF
--- a/crates/swc_ecma_codegen/src/config.rs
+++ b/crates/swc_ecma_codegen/src/config.rs
@@ -39,6 +39,13 @@ pub struct Config {
 
     #[cfg_attr(feature = "serde-impl", serde(default))]
     pub inline_script: bool,
+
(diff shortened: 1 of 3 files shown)


dsherret (commented):
Not a member, but personally I think the SWC project not have a separate way of doing this and I believe that transformation is necessary anyway. For example, this currently doesn't handle stuff like class parameter properties, transforming enums, and transforming typescript namespaces. Also, usually when emitting other transformations will need to be done anyway like TSX -> JS or downleveling some JS features.

kdy1:
Closing as there are too many concerns

PR #8383, opened 2023-12-05
Use sha2 in place of sha-1

### LINKED ISSUE #8382: Use sha2 in swc_ecma_testing & swc_ecma_transforms_testing instead of sha-1
### Describe the feature

When writing custom plugins we need to pull in the swc_ecma_transforms_testing & swc_ecma_testing dependencies to run tests on the plugin. Many orgs have strict software supply chain checks, and will call out `sha1` as a deprecated/insecure hashing algorithm (My org did).

Since this is only pulled in only for checking compilation cache, we can easily swap it for `sha2`, without much issues. Let me know if you are open for this change, I will create a PR for this.

_No response_

### Additional context

_No response_

ronakj:
**Descrption:**

`sha-1` crate has been deprecated and the algorithm itself is not considered secure or collision-proof. Moving to `sha2` solves this.

**Related issue:**

 - Closes #8382

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4915,7 +4915,7 @@ dependencies = [
  "hex",
  "serde",
  "serde_json",
- "sha-1",
+ "sha2",
  "sourcemap",
  "swc_common",
  "swc_ecma_ast",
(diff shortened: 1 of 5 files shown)


kdy1 (approved):
swc-bump:
 - swc_ecma_testing
 - swc_ecma_transforms_testing

Issue #8398, opened 2023-12-07
Incorrect minification of `(obj?.a).b`


SukkaW:
### Describe the bug

As the title.

### Input code

```typescript
(obj?.a).b
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "target": "es2022",
    "minify": {
      "compress": true,
      "mangle": true
    },
    "loose": false
  },
  "isModule": true,
  "minify": true
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.3.101-nightly-20231207.1&code=H4sIAAAAAAAAA9PIT8qy10vU1EsCAAYRshYKAAAA&config=H4sIAAAAAAAAA22NPQ7DIAxG954Cec5QMfYOPQSiTkTEn2xHKopy9wJNUYds%2BD19j%2F2mFKxs4aH2%2BqxHNsRI466ESxTzrgSkZGRLLgtMPyvclNCGnRxfAWJoQWkjZH3X%2BhxAcNHN5T9vU8iEzGdlhIOJi8ertk%2BJm5iN52Y6B8fP9NrGorPxW28cHzjNEnLuAAAA

### Expected behavior

```js
(obj?.a).b;
```

### Actual behavior

```js
obj?.a.b;
```

### Version

1.3.101-nightly-20231207.1

PR #8399, opened 2023-12-07
fix(es/minify): Preserve parenthesis for optional chaining

magic-akari:
- Closes: #8398


kdy1 (approved):
Thank you!


---


swc-bump:
 - swc_ecma_transforms_base

kdy1:
It seems like this PR introduced a regression for some unknown reason. I'll investigate.
https://github.com/vercel/next.js/pull/59834#issuecomment-1866987810

crates/swc_ecma_transforms_optimization/src/simplify/expr/tests.rs
```diff
@@ -15,6 +15,7 @@ fn fold(src: &str, expected: &str) {
 
             chain!(
                 resolver(unresolved_mark, top_level_mark, false),
+                paren_remover(None),
```
magic-akari, line 18: We have added the `paren_remover` here, and I suppose we need a `fixer` for that. @kdy1

magic-akari: Never mind. This is a test pass

PR #8402, opened 2023-12-07
perf(common): Fix OOM of `inputSourceMap`

kdy1:
**Description:**

The previous code was very strange, and it could easily cause OOM.


kdy1 (commented):
swc-bump:
 - swc_common

Issue #7967, opened 2023-09-18
TypeScript identifier resolver incorrectly handles block scopes


eandre:
### Describe the bug

I'm using `swc_ecma_transforms_base::resolver` to determine which declaration a given identifier refers to.

I'm defining multiple TypeScript interfaces with the same name, each within its own block (using a top-level BlockStmt). However, `swc_ecma_transforms_base::resolver` maps them to the same `swc_ecma_ast::Id`. See example below

### Input code

TypeScript:
```typescript
{
    interface Foo {}
}

{
    interface Foo {}
}
```

Rust:
```rust
    let globals = Globals::new();
    GLOBALS.set(&globals, || {
        let lexer = Lexer::new(
            Syntax::Typescript(Default::default()),
            EsVersion::Es2022,
            StringInput::from(file.as_ref()),
            None,
        );
        let mut parser = Parser::new_from(lexer);
        let ast = parser.parse_module().unwrap();

        let mut resolver = swc_ecma_transforms_base::resolver(Mark::new(), Mark::new(), true);
        let ast = ast.fold_with(&mut resolver);

        // Here I expect the two separate `Foo` idents to have different ids,
        // since they're in different blocks.
    });
```

### Config

Using the Rust library.

### Playground link

View the AST tab:
https://play.swc.rs/?version=1.3.74&code=H4sIAAAAAAAAA6vmUgCCzLyS1KK0xORUBbf8fIXqWq5aLq5qHDIAgwgOgDIAAAA%3D&config=H4sIAAAAAAAAA1VPOw7DIAzdOQXy3KFi6NA79BCIOhERAYQdqSjK3QsJpM1mv4%2Ff8yqkhIkMPOVaxrJEnQjTuReEsmf9KQhwjkgm2chw6yxTpQbtCHdoOxhgnUbk6kJSd6WaA1wIhN3RsNl6O%2BT%2FTBPmmJDoKqxS7UeH10TRUmEO72Un2y%2B179HgAT9RDzsPg6VXd3JaUGxfBMLf3xcBAAA%3D

### Expected behavior

I expect the two `Identifier` AST nodes inside the `TsInterfaceDeclaration` nodes to have different `ctxt` values inside their spans.

### Actual behavior

The two `Identifier` nodes have the same `ctxt` value `2`, which means they get mapped to the same `swc_ecma_ast::Id` even though they're different types.

### Version

1.3.85

kdy1:
I think those are same

eandre:
[...]

PR #8403, opened 2023-12-07
fix(es/resolver): Handle `TsInterfaceDecl` and `UsingDecl` correctly

l4yton:
Related issue:
- Closes: https://github.com/swc-project/swc/issues/7967

### DIFF
--- a/crates/swc_ecma_transforms_base/src/resolver/mod.rs
+++ b/crates/swc_ecma_transforms_base/src/resolver/mod.rs
@@ -1656,6 +1656,10 @@ impl VisitMut for Hoister<'_, '_> {
         if self.resolver.config.handle_types {
             match decl {
                 Decl::TsInterface(i) => {
+                    if self.in_block {
+                        return;
+                    }
+
                     let old_in_type = self.resolver.in_type;
                     self.resolver.in_type = true;
                     self.resolver.modify(&mut i.id, DeclKind::Type);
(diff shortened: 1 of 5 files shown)


l4yton:
This should fix the issue, but I'm not sure how to properly add tests for this.

kdy1:
First of all, thank you for the contribution!

For testing, you can add a file named `input.js` to somewhere in https://github.com/swc-project/swc/tree/35fb6528f54b11192dd967297294cdc1852aa57b/crates/swc_ecma_transforms_base/tests/ts-resolver and run `UPDATE=1 cargo test` from your terminal.

Note that you may need to `touch` https://github.com/swc-project/swc/blob/35fb6528f54b11192dd967297294cdc1852aa57b/crates/swc_ecma_transforms_base/tests/ts_resolver.rs

kdy1:
Also, can you sign the CLA?

l4yton:
Sure, will sign the CLA. Just noticed that this fix doesn't work if typescript is `true` when using `resolver(Mark::new(), Mark::new(), true)`, though it does if it's `false`...

kdy1:
I requested review from other team members because I'm not sure about the changes

Austaras:
This looks fine. What concerns you?

crates/swc_ecma_transforms_base/src/resolver/mod.rs
```diff
@@ -344,35 +344,17 @@ impl<'a> Resolver<'a> {
...
-            return;
+        } else {
+            self.current
+                .declared_symbols
+                .insert(ident.sym.clone(), kind);
         }
 
         let mark = self.current.mark;
 
-        self.current
```
kdy1, line 363: Removal of this logic seems dangerous to me at that time, but now I'm not sure why I feel it dangerous.

kdy1 (approved):
Thank you!

---


swc-bump:
 - swc_ecma_transforms_base

PR #8333, opened 2023-11-23
refactor(es/ast): Improve type definitions of patterns

kdy1:
**Description:**

 - Copy `AssignmentTarget` from `oxc`.
 - Use `BindingIdentifier` in more places.

**Related issue:**
 
 - Closes #8026

PR #8417, opened 2023-12-12
refactor(es/utils): Use exact type for factory methods

kdy1:
**Description:**

This PR is extracted from #8333 because it's too large. 

**Breaking Changes**:
  - `ExprFactory::make_member()` now returns `MemberExpr` instead of `Box<Expr>`.
  - `member_expr!()` now returns `MemberExpr` instead of `Box<Expr>`.


kdy1 (commented):
swc-bump:
 - swc_ecma_utils --breaking

PR #8426, opened 2023-12-15
fix(es/codegen): Fix codegen of ts specific syntax in class props

dsherret:
**Description:** This fixes the codegen of private properties with `is_optional` or`definite` and class properties with `definite`.


Austaras (approved):

magic-akari (approved):

kdy1 (approved):
---

swc-bump:
 - swc_ecma_codegen

PR #8428, opened 2023-12-16
feat: add support for linux riscv64 gnu

kxxt:
**Description:**

This PR adds build for riscv64 linux gnu target.

`packages/html` and `packages/xml` seem to reference non existent crates like `bindings/binding_html_node` so I didn't regenerate the `binding.js` file for them. Looks like `@swc-lab/xml` and `@swc-lab/html` are not published to npm anyway.

### DIFF
--- a/.github/workflows/publish-core.yml
+++ b/.github/workflows/publish-core.yml
@@ -118,6 +118,17 @@ jobs:
               rm -rf target/release/.cargo-lock &&
               cp bindings/target/aarch64-unknown-linux-musl/release/swc . && chmod +x ./swc &&
               env RUSTFLAGS='-C target-feature=-crt-static' yarn build --target=aarch64-unknown-linux-musl
+          - host: ubuntu-latest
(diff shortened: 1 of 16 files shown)


kwonoj:
SWC currently have amount of platform targets and we actually want to reduce the numbers. What's the production system uses this target and how many users out there?

kxxt:
> SWC currently have amount of platform targets and we actually want to reduce the numbers. What's the production system uses this target and how many users out there?

Distro packagers are packaging some applications that uses swc. Frankly speaking I don't know if there is any production system uses swc on this target now. The user base is expected to grow in the future as RISC-V develops and matures.

kwonoj:
Expanding new target is not a one time workb it is continuous maintenance to sve compatibility for the dependencies / code in the swc's core. Please provide suffecient reasoning why swc/core have to manage this targets, happy to discuss internally if there are suffecient usecase.

Please use gh discussion with RFC for those. I'll close PR for now to avoid PR hanging without accepting.

PR #8434, opened 2023-12-19
refactor(css/parser): Remove value normalization

kdy1:
**Description:**

**Related issue :**
 - https://linear.app/vercel/issue/PACK-2154


kdy1:
@alexander-akait I removed value normalization because `transformY()` became `transformy()`

alexander-akait:
@kdy1 Hm, but it is valid change, what is the problem, also gzip is better when you have only lower characters

kdy1 (commented):
swc-bump:
 - swc_css_ast

PR #8443, opened 2023-12-21
fix(css/parser): Fix parsing of `hsla(var(--foo), 1)`

kdy1:
**Description:**

```css
.test {
    border-color: hsla(var(--ds-gray-200-value), 1);
}
```

should be compiled as

```css

.test {
    border-color: hsla(var(--ds-gray-200-value), 1);
}
```

Previously, it was

```
.test {
    border-color: hsla(var(--ds-gray-200-value),);
}

```

**Related issue:**

 - https://linear.app/vercel/issue/PACK-2175


kdy1 (commented):
swc-bump:
 - swc_css_parser

PR #8447, opened 2023-12-22
fix(es/codegen): Correct emit for type-only export declarations

dsherret:
**Description:**

This fixes the emit for `export type { } from "..."` and `export { type A } from "..."`.


kdy1 (approved):
Thank you!

---


swc-bump:
 - swc_ecma_codegen

PR #8449, opened 2023-12-22
fix(es/codegen): Emit `override` keyword

dsherret:
**Description:**

Emits the `override` keyword.


crates/swc_ecma_parser/tests/typescript/class/override/input.ts
```diff
@@ -0,0 +1,5 @@
+class C extends B {
+  override prop = 5;
+  override method() {
```
dsherret, line 3: I got a bit confused because the override keyword didn't show up expect when doing `extends`. This seems ok, but probably the AST should just represent the code as-is.

yarn.lock
```diff
@@ -1,8622 +1,6003 @@
...
-    wcwidth: "npm:^1.0.1"
-  checksum: 8d071828f40090a8e1c6e8f350c6eb065808e9ab2b3e57fa37e0d5ae78cb46dac00117c8f12c3c8b8da2923454afbd8265e08c10b69881170c5b269f451e7fef
-  languageName: node
-  linkType: hard
-
-"os-tmpdir@npm:~1.0.2":
-  version: 1.0.2
-  resolution: "os-tmpdir@npm:1.0.2"
-  checksum: 5666560f7b9f10182548bf7013883265be33620b1c1b4a4d405c25be2636f970c5488ff3e6c48de75b55d02bde037249fe5dbfbb4c0fb7714953d56aed062e6d
-  languageName: n
```
kdy1, line 1: Please revert this

dsherret: Done. Didn't notice. Must have been when I ran `yarn install` for some reason.

kdy1 (approved):
Thanks!

swc-bump:
 - swc_ecma_codegen

Issue #8374, opened 2023-12-04
Missing syntax error for `await` in class static block


nicolo-ribaudo:
### Describe the bug

The following script should be a syntax error, but SWC transforms it to valid code

### Input code

```typescript
class C {
  static {
    var await = 2;
  }
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "decorators": true
    },
    "target": "es2017",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    },
    "transform": {
      "decoratorVersion": "2022-03"
    },
    "assumptions": {
      "iterableIsArray": true
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": false
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.3.100&code=H4sIAAAAAAAAA0vOSSwuVnBWqOZSUCguSSzJTAYzFRTKEosUEssTM0sUbBWMrIFCtVy1XACXq5OLLgAAAA%3D%3D&config=H4sIAAAAAAAAA2VQu27DQAzb8xWG5hZw3aFAt44dunZXHTm4wPeAKAMxgvx77xL7aqObJIoUqeuhaeiMnt6bay5zk1ghWvs8wRyML3lCNidBry4ZPa2ooUADj5A6O0oflS0qMmQ6yR24PXAy1pNY0RN07cvbokVjjJC9FnkX3DBv3fTRJxXg31HP4TRW%2Fv6icsAQ1W%2BFqslvUbgYiqGu7brn9pV2bAYmnyyvYMt3Jso%2Fo3ziQ5XnXdLDwiYfj9Pd1PLd8sFyKKfwMZxBf5tr0pqKHL5W%2BiPT7RfDgTCFrwEAAA%3D%3D

### Expected behavior

Error

### Actual behavior

```js
class C {
}
(()=>{
    var _await = 2;
})();
```

### Version

1.3.100

PR #8450, opened 2023-12-22
fix(es/parser): emit error if await used as ident in class static ini…

iantanwx:
…tializer

**Description:**

This PR fixes https://github.com/swc-project/swc/issues/8374

**BREAKING CHANGE:**

https://github.com/swc-project/swc/issues/8374


magic-akari:
This case should be added.

```JavaScript
class Foo {
    static {
        function foo() {
            var await = 'bar';
        }
    }
}
```

iantanwx:
> This case should be added.
> 
> ```js
> class Foo {
>     static {
>         function foo() {
>             var await = 'bar';
>         }
>     }
> }
> ```

Thanks for this, I've added it.

kdy1 (dismissed):
Thanks!

---

swc-bump:
 - swc_ecma_parser

crates/swc_ecma_parser/src/parser/stmt.rs
```diff
@@ -2407,6 +2407,19 @@ export default function waitUntil(callback, options = {}) {
         test_parser(src, Syntax::Es(Default::default()), |p| p.parse_expr());
     }
 
+    #[test]
+    #[should_panic(expected = "Expected ident")]
```
magic-akari, line 2411: No. It should not panic. It's valid JavaScript.

iantanwx: @magic-akari i didn't realised 😓  

fixed it, but i'm not sure if the solution accounts for all corner cases...could you take a look?

magic-akari:
```JavaScript
class Foo {
    static {
        var await = "bar"; // SyntaxError: Unexpected reserved word

        function foo() {
            var await = "bar"; // ok

            class Bar {
                static {
                    var await = "bar"; // SyntaxError: Unexpected reserved word
                }
            }
        }
    }
}
```

I recommend setting the variable `in_static_block` to `false` upon entering the function body. 
Please note that I cannot guarantee that all scenarios have been considered.

iantanwx: Fixed, your suggested solution seems much better to me.

iantanwx: @magic-akari bump

magic-akari (approved):

sweetliquid (approved):

Issue #8459, opened 2023-12-29
Allow self-closing `/>` on HTML non-void elements via a flag


phoenix-ru:
Syntax like that is often used in Vue templates:
```html
<template>
  <slot />
  <div class="bad" />
</template>
```

Since `swc_html_parser` is spec-compliant, it acknowledges the `<slot />` and `<div />`, but does not actually pop the DOM Tree stack.
It makes parsing such templates impossible: https://github.com/phoenix-ru/fervid/issues/6

Can we add a flag to `ParserConfig` or a crate feature to treat such starting tags as self-closing? I can open a ticket and work on a PR.

_Originally posted by @phoenix-ru in https://github.com/swc-project/swc/discussions/8424_

PR #8460, opened 2023-12-29
feat(html/parser): Allow self-closing `/>` on non-void HTML elements via a flag

phoenix-ru:
**Description:**
This Pull Request allows authoring HTML like that, which is common and [encouraged](https://vuejs.org/guide/essentials/component-basics.html#self-closing-tags) in Vue templates:
```html
<slot />
<div class="self-closing" />
<p v-text="jsText" />
```

To avoid breaking changes, this behaviour is enabled using `allow_self_closing` flag in `ParserConfig`, which defaults to `false`. Consumers of `ParserConfig::default()` will not notice the change.

Thanks to this change, there will be significantly less friction migrating existing Vue codebases to a native SWC-based compiler.

Afaik, other spec non-conformance in an existing Vue parser (all mentioned [here](https://vuejs.org/guide/essentials/component-basics.html#in-dom-template-parsing-caveats), plus some undocumented) is either covered by SWC already (such as namespaces support) or [can be supported outside](https://github.com/phoenix-ru/fervid/blob/f29ea46192393a65c2cb5e320119dea0bf3a203e/crates/fervid_parser/src/template.rs#L51-L58) SWC.

https://github.com/swc-project/swc/issues/8459

### DIFF
--- a/crates/swc_html_parser/src/parser/mod.rs
+++ b/crates/swc_html_parser/src/parser/mod.rs
@@ -28,6 +28,8 @@ pub type PResult<T> = Result<T, Error>;
 pub struct ParserConfig {
     pub scripting_enabled: bool,
     pub iframe_srcdoc: bool,
+    // #8459
+    pub allow_self_closing: bool,
 }
 
 enum Bookmark<RcNode> {
@@ -1110,6 +1112,27 @@ where
             _ => self.insertion_mode.clone(),
         };
 
+        /// Convenience: allow non-void HTML elements to self-close when
+        /// a relevant config flag is set. It is achieved by processing the
+        /// matching end tag right after the starting self-closing tag.
+        macro_rules! maybe_allow_self_closing {
+            ($is_self_closing: ident, $tag_name: ident) => {
+                if self.config.allow_self_closing && *$is_self_closing {
+                    let mut end_token_and_info = TokenAndInfo {
+                        span: token_and_info.span,
+                        acknowledged: false,
+                        token: Token::EndTag {
+                            tag_name: $tag_name.to_owned(),
+                            raw_tag_name: None,
+                            is_self_closing: false,
+                            attributes: vec![],
+                        },
+                    };
+                    self.process_token(&mut end_token_and_info, None)?;
+                }
+            };
+        }
+
         match insertion_mode {
             // The "initial" insertion mode
             InsertionMode::Initial => {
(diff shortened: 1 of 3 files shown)


crates/swc_html_parser/src/parser/mod.rs
```diff
@@ -1828,6 +1855,7 @@ where
                         self.input.set_input_state(State::ScriptData);
                         self.original_insertion_mode = self.insertion_mode.clone();
                         self.insertion_mode = InsertionMode::Text;
+                        maybe_allow_self_closing!(is_self_closing, tag_name);
```
alexander-akait, line 1858: I think we don't need it here, because it is the `script` tag, I don't think it is possible to have `<script />`

crates/swc_html_parser/src/parser/mod.rs
```diff
@@ -1861,13 +1889,18 @@ where
...
+                        is_self_closing,
+                        ..
+                    } if tag_name == "template" => {
                         self.insert_html_element(token_and_info)?;
                         self.active_formatting_elements.insert_marker();
                         self.frameset_ok = false;
                         self.insertion_mode = InsertionMode::InTemplate;
                         self.template_insertion_mode_stack
                             .push(InsertionMode::InTemplate);
+                        maybe_allow_self_closing!(is_self_closing, tag_name);
```
alexander-akait, line 1903: Same here, so let's avoid do it there it unnecessary for perf reasons and avoid complex code where it unnecessary

crates/swc_html_parser/src/parser/mod.rs
phoenix-ru: I was generally thinking the same before I recalled this usage: https://vuejs.org/api/sfc-spec.html#src-imports.

TBH, I'd rather agree with you and make `<script />` and `<style />` unsupported. I've rarely seen src imports ever being used.

crates/swc_html_parser/src/parser/mod.rs
phoenix-ru: Found only a single test with this usage.
https://github.com/vuejs/core/blob/a41c5f1f4367a9f41bcdb8c4e02f54b2378e577d/packages/compiler-sfc/__tests__/parse.spec.ts#L196

In general I was trying to match the `Vue's template parser respects /> as an indication to end any tag, regardless of its type` premise. It is pretty convenient from a developer standpoint to not think about what can and what cannot be self-closed at all.

Regarding performance, I'd expect `rustc` to drop this branch altogether after static analysis (because `allow_self_closing` is set to `false` everywhere and condition always evaluates to `false`). But it's just a wild guess.

Performance is partially why I suggested making it a feature flag. But because features are merged, it is probably not a good idea.

Let's measure and see.

kdy1 (approved):
@alexander-akait  I think it's good enough because the performance will not drop significantly.


---


swc-bump:
 - swc_html_parser

Issue #8467, opened 2024-01-01
The build option `jsc.transform` does not have a type hint for `decoratorVersion`


SokuRitszZ:
### Describe the bug

There is no hint when I use jsc.transform.decoratorVersion.

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.3.102&code=H4sIAAAAAAAAA1WMvQrCQAzH9zzFf2xB2l2pKPgATu5HGkrhuEgSaaXcu1sXwfn30fdwEXS%2BsPERd9OnWLwxCqulUHuI%2BawFs6NoIOWsi4wdEWvxwE0YA5oWwxkbASbxsoImkk0SB3CsPwZ8G83SZZ3%2BjNNOK1Wiy%2F5rWuKc3HHFVj%2BKdfEqoAAAAA%3D%3D&config=H4sIAAAAAAAAA3WQuw6DMAxFd74CeS7iUalD%2F6Fr9ygYlIo8ZKdSEeLfm0CgVGo3%2B%2FrYvvaU5Tk8WMI1n0IYEieIkfY8KDwaL15BAT86ZEnKeTht1RalJeEtcSA8PXGveI5NnRgYF2leK%2BAF9ejjPOSmapo0CwZrGbeOpGllVDce3UirHSHzNxhRYfoBf28kYbizpI%2BDduN3JFbWREPRTlGdITFl%2BReri7pZsTlLi0Db9rk4SK%2BM71rPvMAH2i7a3YPi29YZH5jNb%2FgW%2FkyWAQAA

### Expected behavior

Add the jsc.transform type for "2021-12" | "2022-03"

### Version

1.3.102

PR #8468, opened 2024-01-01
fix(binding/types): Add type for `decoratorVersion`

SokuRitszZ:
**Description:**

Add type hint `"2021-12" | "2023-03` for `jsc.transform.decoratorVersion`.

**BREAKING CHANGE:**

#8467


kdy1 (approved):
swc-bump:
 - dbg-swc

PR #8470, opened 2024-01-01
fix(es/compat): Set inserted var inside export class in destructing

Austaras:
**Description:**

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---

---

swc-bump:
 - swc_ecma_compat_es2015

PR #8469, opened 2024-01-01
fix(es/resolver): Resolve shadowing of `undefined`, `NaN`, and `Infinity`

magic-akari:
**Related issue:**
- Closes: #8465

PR #8471, opened 2024-01-02
fix(es/resolver): Resolve top-level `undefined`, `NaN`, and `Infinity` correctly

### LINKED ISSUE #8465: wrongly transform function Infinity to 1/0
### Describe the bug
Infinity is wrongly transformed to 1/0
### Input code

```typescript
function Infinity(){
    console.log('xxx');
}
export default Infinity;
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": true,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true,
        "const_to_let": true,
        "pristine_globals": true
      },
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

https://play.swc.rs/?version=1.3.100&code=H4sIAAAAAAAAA0srzUsuyczPU%2FDMS8vMyyyp1NCs5lIAguT8vOL8nFS9nPx0DfWKigp1TWuuWq7UioL8ohKFlNS0xNKcErgmawCsssEKRwAAAA%3D%3D&config=H4sIAAAAAAAAA32US5LjIAyG932KlNe9nVnMAXo3Z6AICIc0Ri4k0nF15e4jY%2BcxHTk7W59%2BCSSh77fdrjuS6%2F7svuVTfkZbCMrtXyw0ZbZnsXTgBkuuxJG79ys90oyCTQTNdFlIx7b0wE1Fv1b3LiESXN1X2xBzDNNjQofDWIDowSZWiVcHyEz%2F61dW8GsGXOqjfY%2BYwOYXxFgyMTP0ULTADlOyI4E52aJEmU9qSyTUUsywMngzFhxVnn3kiFlyPlMP1huHHhQUCziOJ9BkkktkmeR6yn0a9rCvfd%2Ba%2FEMNJ5uqZSUnnFtL5LRK1ANGYhNq1kq4wI0aLHAt7k9lDKYA15KfdUeMeaMnnwBSgWSJsh1Ai9s8gszTljq8VMYcZGR5UrjMt3bLDL0U1cQYlMrOlYHCUetmAV8dzJV12nFWvFE%2Bih4MhCCzooSmr8juoCXlaQQMCpD%2B2qBN1QLM7RVu8PlBvMAfckvWB2z1GCwftilNwx7TiwQD8AH9CwdpBeM2LrIlzuM2r9mDjAZ41aVSA89LQB4Ao0ltWT7NhjwPiWj6hPv7mlgdLrclPNjc39%2F7soffVoduQF8bXDf83N9lL%2F%2Fu7k7XLXw7eBfp71XZkl7%2BAVPL6HotBgAA

### SWC Info output

1.3.100

### Expected behavior

don't turn to 1/0 like terser 
```js
function o(){console.log("xxx")}export default o;
```

### Actual behavior
```js
export default 1 / 0;
```

### Version

1.3.100

### Additional context

_No response_

Austaras:
…aration correctly in es module

**Description:**

For following code
```js
var NaN = 1
console.log(NaN)
```
Result would be

|Envirnoment|Result|
|-|-|
|Non strict script(browser, nodejs repl)|NaN|
|Non strict script(nodejs script)|1|
|Strict script(browser, nodejs repl)|runtime error|
|Strict script(nodejs script)|1|
|ESM|1|

So SWC choose to behave like browser in script mode and confirm to esm standard.

**BREAKING CHANGE:**

Closes #8465


crates/swc_ecma_transforms_base/tests/resolver/issues/7685/output.js
```diff
@@ -2,4 +2,4 @@ var NaN__2 = 1;
 {
     let NaN__3 = 1;
     console.log(NaN__3);
-}console.log(NaN);
+}console.log(NaN__2);
```
Austaras, line 5: The difference between this and #8469 is #8469 would generate
```js
var NaN = 1;
{
    let NaN__3 = 1;
    console.log(NaN__3);
}console.log(NaN);
```
which I believe would introduce problems for following passes.

crates/swc_ecma_minifier/tests/terser/compress/issue_597/NaN_and_Infinity_should_not_be_replaced_when_they_are_redefined/output.mangleOnly.js
```diff
@@ -1,3 +1,3 @@
-var t, i;
-Infinity.toString();
-NaN.toString();
+var t, r;
```
magic-akari, line 1: Anyway, We should not rename it.

crates/swc_ecma_minifier/tests/terser/compress/issue_597/beautify_off_1/output.js
```diff
@@ -1,2 +1,2 @@
 var NaN1;
```
magic-akari, line 1: It should be `NaN`

crates/swc_ecma_minifier/tests/terser/compress/issue_597/beautify_on_1/output.mangleOnly.js
```diff
@@ -1,2 +1,2 @@
 var n;
```
magic-akari, line 1: Same here

crates/swc_ecma_minifier/tests/terser/compress/issue_597/NaN_and_Infinity_should_not_be_replaced_when_they_are_redefined/output.mangleOnly.js
Austaras: Why not? All terser test cases are run in esm, it's fine to rename anything not exported.

crates/swc_ecma_minifier/tests/terser/compress/issue_597/beautify_off_1/output.js
Austaras: It can't be `NaN`. In L2 there's reference to global `NaN`, the only thing can do is to rename this to avoid shadowing.

magic-akari:
```
❯ node
Welcome to Node.js v18.18.2.
Type ".help" for more information.
> var NaN;
undefined
> console.log(
...     null,
...     undefined,
...     Infinity,
...     NaN,
...     Infinity * undefined,
...     Infinity.toString(),
...     NaN.toString(),
...     (Infinity * undefined).toString()
... );
null undefined Infinity NaN NaN Infinity NaN NaN
undefined
```

magic-akari: Shadowing can have different outcomes, such as taking effect, causing silent errors, or resulting in an error message. Due to the lack of information, we can only determine the result of shadowing at runtime. The only correct approach is to avoid renaming.

magic-akari (changes requested):

crates/swc_ecma_minifier/tests/terser/compress/issue_597/beautify_off_1/output.js
kdy1: Are you talking about the ESM mode?

Basically, the input is https://github.com/Austaras/swc/blob/16eee5ce3f12154ea8f5e3e62d6b25805a3504b6/crates/swc_ecma_minifier/tests/terser/compress/issue_597/beautify_off_1/input.js and when I tried executing code, the result was different from the result of executing the original input file

Austaras: Node repl is not esm, so this is not relate to this PR, which only handles output in esm.

Austaras: > Are you talking about the ESM mode?
> 

Yes.

kdy1:
I think the problem here may be the fact that we use ESM mode for terser test suites

crates/swc_ecma_minifier/tests/terser/compress/issue_597/beautify_off_1/output.js
Austaras: But in non-strict non-esm js, the end result should be
```js
var NaN1;
console.log(null, void 0, 1 / 0, NaN, NaN, (1 / 0).toString(), NaN.toString(), NaN.toString());
```
which is consistent between nodejs and browser.

Austaras: Oh, and terser would generate
```js
var NaN1;
console.log(null, void 0, 1 / 0, NaN1, NaN, (1 / 0).toString(), NaN1.toString(), NaN.toString());
```
when `module: false`

Austaras:
> I think the problem here may be the fact that we use ESM mode for terser test suites

Yes. I would open a separate PR for this. Doesn't seem like to have any problem.

Austaras:
Some investigation, for following code
```js
var NaN = 1
console.log(NaN)
```
Result would be

|Envirnoment|Result|
|-|-|
|Non strict script(browser, nodejs repl)|1|
|Non strict script(nodejs)|1|
|Strict script(browser)|runtime error|
|Strict script(nodjes)|1|
|ESM|1|

Seems like the benefit of this mechanism is very limited. Maybe just remove it?

kdy1:
I agree that the benefit is small, but I also think code is trivial enough to justify the size of benefit

crates/swc_ecma_minifier/tests/terser/compress/issue_597/NaN_and_Infinity_should_not_be_replaced_when_they_are_redefined/output.mangleOnly.js
Austaras: For
```js
var NaN = "zabcded"

console.log(NaN, NaN)
```
terser in `module: false` and `mangle: { toplevel: true }` would generate
```js
var o="zabcded";console.log(o,o);
```
I believe this is the right solution.

kdy1: Is this output expected? I expected this to be preserved in non-ESM mode, but maybe I was wrong

Austaras:
Fine. Let's proceed this way.

crates/swc_ecma_minifier/tests/terser/compress/issue_597/NaN_and_Infinity_should_not_be_replaced_when_they_are_redefined/output.mangleOnly.js
Austaras: And for
```js
var NaN = 1
console.log(NaN)
```
terser in `module: false` and `compress: { toplevel: true }` would generate
```js
console.log(1)
```

kdy1 (dismissed):
Thank you all!

---

swc-bump:
 - swc_ecma_transforms_base

crates/swc_ecma_minifier/tests/terser/compress/issue_597/NaN_and_Infinity_should_not_be_replaced_when_they_are_redefined/output.mangleOnly.js
Austaras: > Is this output expected? I expected this to be preserved in non-ESM mode, but maybe I was wrong

No, this doesn't seem to be right.

magic-akari:
> Maybe just remove it?

What does this mean? Should SWC skip them when renaming?

Austaras:
> > Maybe just remove it?
> 
> What does this mean? Should SWC skip them when renaming?

Maybe just remove the special treatment for `undefined` etc like what babel do. But that's not very good for minifier

kdy1 (approved):
Thank you!

PR #8472, opened 2024-01-02
test(es/minifier): Enable non esm mode for tests

Austaras:
**Description:**

**BREAKING CHANGE:**


crates/swc_ecma_minifier/tests/terser_exec.rs
```diff
@@ -54,7 +54,7 @@ use testing::assert_eq;
         // should error
         "blocks/issue_1672_if_strict",
         "blocks/issue_1672_for_strict",
-        // need support for script mode
+        // annex B
```
Austaras, line 57: We still cannot enable these two cases because nodejs doesn't implement annex B.

kdy1 (approved):
Thank you!

---

swc-bump:
 - dbg-swc

PR #8474, opened 2024-01-02
test(es/minifier): Enable script mode for `terser_exec` and `mangle`

Austaras:
**Description:**

**BREAKING CHANGE:**


Austaras:
The added file is because await and yield is not a keyword in script mode.

kdy1 (approved):

kdy1 (approved):
swc-bump:
 - dbg-swc

PR #8477, opened 2024-01-02
fix(es/codegen): Emit implements clause with commas

dsherret:
**Description:**

Emitting `implements` was missing commas.

### DIFF
--- a/crates/swc_ecma_ast/src/list.rs
+++ b/crates/swc_ecma_ast/src/list.rs
@@ -143,7 +143,7 @@ bitflags! {
             | Self::SpaceBetweenSiblings.bits()
             | Self::SpaceBetweenBraces.bits();
         const MultiLineFunctionBodyStatements = Self::MultiLine.bits();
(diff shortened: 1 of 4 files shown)


dsherret:
It seems one of the swc_ecma_minifier tests is flaky. I also had this fail then pass on the other PR.

kdy1 (approved):
swc-bump:
 - swc_ecma_codegen

crates/swc_ecma_ast/src/list.rs
```diff
@@ -143,7 +143,7 @@ bitflags! {
             | Self::SpaceBetweenSiblings.bits()
             | Self::SpaceBetweenBraces.bits();
         const MultiLineFunctionBodyStatements = Self::MultiLine.bits();
-        const ClassHeritageClauses = Self::SingleLine.bits() | Self::SpaceBetweenSiblings.bits();
+        const ClassHeritageClauses = Self::CommaDelimited.bits() | Self::SingleLine.bits() | Self::SpaceBetweenSiblings.bits();
```
dsherret, line 146: This one will actually be fixed via a bump of swc_ecma_ast.

PR #8478, opened 2024-01-02
fix(es/codegen): Emit declare keyword for class properties

dsherret:
**Description:**

Emits the `declare` keyword in class properties.


kdy1 (dismissed):
swc-bump:
 - swc_ecma_codegen

dsherret:
Fixed conflict in output.min.js file.

kdy1 (approved):

PR #8479, opened 2024-01-02
fix(es/codegen): Emit abstract keyword

dsherret:
**Description:** Emits the abstract keyword on classes and their methods and properties.


kdy1 (approved):
swc-bump:
 - swc_ecma_codegen

Issue #7279, opened 2023-04-16
Different `optional` for `Ident` in class property when has private name


dsherret:
### Describe the bug

There is a different behaviour for `Ident` when it's a private class property vs a regular class property. I think maybe `optional` can be removed on `Ident`? Shouldn't that be a higher level concept (ex. part of the parameter or class property and not the ident like how TypeScript represents its AST... I feel like the same should be done for `type_ann` too)?

### Input code

```typescript
export class Response {
  #a?;
  b?;
}
```

### Playground link

https://play.swc.rs/?version=1.3.50&code=H4sIAAAAAAAAA0utKMgvKlFIzkksLlYISi0uyM8rTlWo5uVSUFBOtLcG0UkgqhYAQ0CvSikAAAA%3D&config=H4sIAAAAAAAAA1WQvW7DMAyE9z6FoTmoCw8ZsrbIlilPQCh0w0J%2FICkgRuB3jyTHcbqJ3%2FF04N0%2Fus78iTWH7l6eZUjAgvyaC5EpKNwKMTolFMuU1OxWVaVKyhkbmRfBKPAvajWhDF%2FDYFbOEGSM7N8TsuAPjhTwGPnbgciR0F3k%2Be8rihGsvvkKIp8i6zlmtljDrqpJDn2P4j%2Fl2qfFstsMnIOSb7uQNXpQsuYpz%2F8ucDFK3RvBSb2tcePjJTvc6qqVLFfuzbZEgcZp9TZEclqdrav5AeQIxLB6AQAA

### Expected behavior

Probably just remove `optional` from ident or both should be `true`.

### Actual behavior

```json
{
  "type": "Module",
  "span": {
    "start": 0,
    "end": 41,
    "ctxt": 0
  },
  "body": [
    {
      "type": "ExportDeclaration",
      "span": {
        "start": 0,
        "end": 41,
        "ctxt": 0
      },
      "declaration": {
        "type": "ClassDeclaration",
        "identifier": {
          "type": "Identifier",
          "span": {
            "start": 13,
            "end": 21,
            "ctxt": 2
          },
          "value": "Response",
          "optional": false
        },
        "declare": false,
        "span": {
          "start": 7,
          "end": 41,
          "ctxt": 0
        },
        "decorators": [],
        "body": [
          {
            "type": "PrivateProperty",
            "span": {
              "start": 27,
              "end": 31,
              "ctxt": 0
            },
            "key": {
              "type": "PrivateName",
              "span": {
[...]

PR #8481, opened 2024-01-03
fix(es/codegen): Emit `?` for an optional computed property

dsherret:
**Description:** Emits optional `?` for a computed property.

### DIFF
--- a/crates/swc_ecma_codegen/src/lib.rs
+++ b/crates/swc_ecma_codegen/src/lib.rs
@@ -1566,9 +1566,11 @@ where
         }
 
         emit!(n.key);
+
         if n.is_optional {
             punct!("?");
         }
+
         if let Some(type_ann) = &n.type_ann {
             if n.definite {
                 punct!("!");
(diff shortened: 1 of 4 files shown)


crates/swc_ecma_codegen/src/lib.rs
```diff
@@ -1635,6 +1635,11 @@ where
 
         emit!(n.key);
 
+        // emit for a computed property, but not an identifier already marked as optional
+        if n.is_optional && !n.key.as_ident().map(|i| i.optional).unwrap_or(false) {
+            punct!("?");
+        }
```
dsherret, line 1641: When parsed it will have `optional` as `true` for both `n.is_optional` and on the ident, so that's why this condition is here. Somewhat related: https://github.com/swc-project/swc/issues/7279

crates/swc_ecma_codegen/tests/fixture/typescript/class_prop/input.js
```diff
@@ -3,9 +3,11 @@ class MyClass extends Base {
     prop2!: string;
     #prop3?: string;
     #prop4?: string = "test";
+    #privateOptionalNoType?;
```
dsherret, line 6: Also fixes this.

kdy1 (approved):
Thanks!


---


swc-bump:
 - swc_ecma_codegen

Issue #8482, opened 2024-01-04
Syntax Error Reported for "let a = ''" When Using '\u00a0' (Non-breaking Space) Between 'let' and 'a'


y1j2x34:
### Describe the bug

When using the statement let a = '' with a non-breaking space ('\u00a0') between 'let' and 'a', SWC reports a syntax error. It's important to note that executing this statement directly in a browser environment does not result in any errors.

### Input code

```typescript
export let a = ''
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "externalHelpers": true,
    "minify": {
      "compress": false,
      "mangle": false
    },
    "keepClassNames": true,
    "loose": true
  },
  "minify": false,
  "isModule": true,
  "module": {
    "type": "umd"
  },
  "env": {
    "targets": "",
    "bugfixes": true
  }
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.3.100&code=H4sIAAAAAAAAA0utKMgvKlHISS05tCBRwVZBXR0ALTDuvhIAAAA%3D&config=H4sIAAAAAAAAA21QQQ7CMAy77xVTzryAKxcu8IcwvKmj7aqmQ5um%2FZ0WujIhbrEd21GWqq6pl4aO9RLHCBx7gS84MjLbwFNkCI1habxygQ6b2kuSWtaCN7V%2BFMIU4C3rM7SDl7gT%2FIgsGmVVO%2B9LmsE4D5Etq%2BQbtp3G34oH4E6aRa5s8NOgh0GQqSpbvrWlgpRchvuosXOT2Zj8kjC7hGg0dypZsM%2FdBvsOIZ1A%2BTN0G7tWTeWs5KvWFyqxYZJvAQAA

### Expected behavior

A syntax error is reported when a non-breaking space ('\u00a0') is used between 'let' and 'a' during static code analysis or linting.

### Actual behavior

  x Expected '{', got 'let'
   ,-[input.js:1:1]
 1 | export let a = ''
   :        ^^^
   `----

Caused by:
    0: failed to process js file
    1: Syntax Error

### Version

1.3.100

PR #8483, opened 2024-01-04
fix(es/parser): The non ident part should not affect the keyword evaluation

magic-akari:
**Related issue:**

- Closes: #8482


kdy1 (approved):
swc-bump:
 - swc_ecma_parser

PR #8513, opened 2024-01-17
fix(es/typescript): Fix panic on invalid jsx pragma

melbourne2991:
**Description:**

Currently a jsx pragma with an invalid js identifier (eg, with dashes: `@jsx bad-pragma` causes a panic.

This PR prevents a panic and will ignore an invalid pragma in a comment.

(There may be an argument for showing an error or warning instead, but given a jsx pragma is still valid ES I'm not sure whether that makes sense?)

[Original (Deno) issue here](https://github.com/denoland/deno/issues/21927)


kdy1 (approved):
Thanks!

---

swc-bump:

PR #8530, opened 2024-01-20
fix(common): Remove `<` and `>` from `Display` impl of `FileName::Custom`

kdy1:
**Description:**

It causes some problems for turbopack.


kdy1 (commented):
swc-bump:
  - swc_common

PR #8536, opened 2024-01-22
fix(es/quote): Add support for import phase to quote macro

JSerFeng:
**Description:**

When using quote! to generate import decl, compile error: missing field `phase` in initializer of 'swc_core::ecma::ast::Import'

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

---


swc-bump:
 - swc_ecma_quote_macros

PR #8547, opened 2024-01-24
feat(css/parser): Implement error reporting for `@value` of CSS Modules

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_css_parser

PR #8550, opened 2024-01-24
fix(es/loader): Make `tsc` resolver work for bare specifier

XHMM:
**Description:**
From the typescript [baseUrl](https://www.typescriptlang.org/docs/handbook/modules/reference.html#baseurl) doc: When using bare specifiers (module specifiers that don’t begin with ./, ../, or /), baseUrl has a higher precedence than node_modules package lookups. 

In the current tsc resolver implementation, when resolving bare module specifiers, baseUrl was not used except for `paths`, this cause the belowing resolution failed,  but it worked when used in typescript project.

`tsconfig.json` / `.swcrc`:
```json
{
    baseUrl: "."
    paths: {
        "@common/*": ["src/common/*"]
    }
}
```

File structure:
- ./src/common/helper.ts
- ./src/index.ts

./src/index.ts content:
```ts
 // tsc can resolve this, but tsc resolver cannot
import sth from "src/common/helper"
```

**BREAKING CHANGE:**


crates/swc_ecma_loader/src/resolvers/tsc.rs
```diff
@@ -310,6 +310,13 @@ where
...
         }
 
+        // https://www.typescriptlang.org/docs/handbook/modules/reference.html#baseurl
+        if let Ok(v) =
+            self.invoke_inner_resolver(base, self.base_url.join(module_specifier).to_str().unwrap())
+        {
+            return Ok(v);
+        }
+
         if let Ok(v) = self.invoke_inner_resolver(&self.base_url_filename, module_specifier) {
```
kdy1, line 320: Please remove this line

XHMM: Done

kdy1 (approved):
Thanks!

---


swc-bump:
 - swc_ecma_loader

kdy1 (changes requested):
CI failed

crates/swc_ecma_loader/tests/tsc_resolver.rs
```diff
@@ -97,6 +97,46 @@ fn pattern_1() {
     }
 }
 
+#[test]
+fn base_url_precedence() {
```
kdy1, line 101: The test name says that it's a test about precedence, but it's not testing precedence. Can you add some files to https://github.com/swc-project/swc/tree/039c1685c663cd138e4a0cf769b506d1fd419ba5/crates/swc_ecma_loader/tests and check if it's actually precedence?

XHMM: I'vs updated test cases into two cases for more clear. 😊

kdy1 (approved):
Thank you!

---

swc-bump: 
 - swc_ecma_loader

Issue #8551, opened 2024-01-24
Decorator transform puts variables in incorrect scope


0f-0b:
### Describe the bug

If a class is followed by a TypeScript namespace, the decorator transform places variables generated for the class in the namespace.

### Input code

```typescript
class C {
  [Symbol.iterator]() {}
}

namespace NS {
  export function f() {}
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "decorators": true
    },
    "transform": {
      "decoratorVersion": "2022-03"
    },
    "target": "es2022"
  },
  "isModule": true
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.3.105&code=H4sIAAAAAAAAA0vOSSwuVnBWqOZSUIgOrsxNys%2FRyyxJLUosyS%2BK1dBUqK7lquXiykvMTS0uSExOVfALBitNrSjILypRSCvNSy7JzM9TSIMpBQBdEy1bUAAAAA%3D%3D&config=H4sIAAAAAAAAA1WNwQrCQAxE7%2F2KJWeFst78B6%2FewzbKim6WJAVL6b%2BbrbToLTNvZjJ3IcBDE5zD7KeLiqIku3ZHp2L4dgdsqqRJcjU4bHSgxILGop4wGWkFy5eDCRa9sbx%2BB%2FfKlUQzlzYd%2BxiP%2FQn%2B2yh3soZJW6DRlUHWCw%2Fjk7afS%2FcBWAWB98oAAAA%3D

### Expected behavior

`_computedKey` is declared in the same scope as `C`.

```js
var _computedKey;
_computedKey = Symbol.iterator;
class C {
    [_computedKey]() {}
}
var NS;
(function(NS) {
    function f() {}
    NS.f = f;
})(NS || (NS = {}));
```

### Actual behavior

`_computedKey` is declared in the namespace.

```js
_computedKey = Symbol.iterator;
class C {
    [_computedKey]() {}
}
var NS;
(function(NS) {
    var _computedKey;
    function f() {}
    NS.f = f;
})(NS || (NS = {}));
```

### Version

1.3.105

PR #8556, opened 2024-01-25
fix(es/decorator): Preserve state while traversing the `module_items` scope

magic-akari:
**Related issue:**
- Closes: #8551


kdy1 (approved):
Thanks!

---

swc-bump:

PR #8574, opened 2024-01-30
fix(es/proposals): Support using `using` keyword with functions

### LINKED ISSUE #8570: Support the `using` keyword for functions
### Describe the feature

The following check for the `using` keyword currently only allows it for objects. These lines will prevent the `using` keyword from being used on functions which is allowed in tsc.

https://github.com/swc-project/swc/blob/d6cd2146faed9155ae6f2c625daa173ddc7c0064/packages/helpers/esm/_using.js#L5-L7

A function can have the `Symbol.dispose` or `Symbol.asyncDispose` added to them to make them disposable just like how an object can. Here is the implementation in TypeScript, showing that it should allow the `using` keyword for functions.

https://github.com/microsoft/TypeScript/blob/36f9e9e037a8bf715d96c8270732d5830719d7c8/tests/baselines/reference/usingDeclarations.11.js#L57
```js
if (typeof value !== "object" && typeof value !== "function") throw new TypeError("Object expected.");
``` 

_No response_

### Additional context

This is based on a discussion about making spies and stubs disposable so that the originals will automatically be restored after they are no longer needed. This is needed to implement the feature.

https://github.com/denoland/deno_std/issues/3966

kdy1:
**Description:**

**Related issue:**

 - Closes #8570
 - https://github.com/babel/babel/pull/16150

### DIFF
--- a/crates/swc_ecma_transforms_base/src/helpers/_using.js
+++ b/crates/swc_ecma_transforms_base/src/helpers/_using.js
@@ -1,9 +1,7 @@
 function _using(stack, value, isAwait) {
     if (value === null || value === void 0) return value;
-    if (typeof value !== "object") {
-        throw new TypeError(
-            "using declarations can only be used with objects, null, or undefined."
-        );
+    if (Object(value) !== value) {
+        throw new TypeError("using declarations can only be used with objects, functions, null, or undefined.");
     }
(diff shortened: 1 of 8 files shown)


kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base

Austaras (approved):

PR #8580, opened 2024-01-31
fix(es/parser): Utilize `TokenKind`

kdy1:
**Description:**

 - https://rustmagazine.org/issue-3/javascript-compiler/


kdy1:
Seems like this is a too big changed considering the profit

PR #8582, opened 2024-02-01
feat(es/minifier): Respect more options

kdy1:
**Description:**

This PR makes the ES minifier respect more options.

Changes:
 - `reduce_vars` now enables IIFE invoker.
 - `reduce_vars` now enables seq inliner.
 - `top_retain` is now respected by seq inliner.
 - `arguments` is now ignored by seq inliner.

PR #8583, opened 2024-02-01
fix(es/minifier): Fix top-level check

kdy1:
**Description:**

Our logic for checking top-level was wrong.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

Austaras (changes requested):
Everything else is good. If `module: true` and `toplevel: false`, we should still remove unused top level function.

kdy1:
top_retain will be fixed by https://github.com/swc-project/swc/pull/8582

Austaras (dismissed):

PR #8584, opened 2024-02-01
feat(es/minifier): Inline more IIFE arguments

kdy1:
**Description:**

We can inline more.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier

PR #8593, opened 2024-02-02
feat(es/minifier): Implement correct `hoist_props`

kdy1:
**Description:**

 - The option `hoist_props` now does what it's supposed to do.
 - Dropping of unused properties now does not drop properties too aggressively.
 - The initializer of a dropped variable declaration is now properly visited.
 - Indexing with string literals is not marked as a dynamic index anymore. This is required to handle codes like https://github.com/swc-project/swc/blob/c3f67ceb1eb0cab9ef4a47c321acf90684bf216a/crates/swc_ecma_minifier/tests/terser/compress/hoist_props/name_collision_1/input.js#L1-L7.


kdy1 (commented):
---

swc-bump:
 - swc_ecma_usage_analyzer

Austaras (approved):

PR #8603, opened 2024-02-05
fix(es/quote): Support `AssignTarget`, really

kdy1:


kdy1 (commented):
swc-bump: 
 - swc_ecma_quote_macros

PR #8608, opened 2024-02-05
refactor: Remove dependency on swc_cached from swc_config

dsherret:
**Description:**

swc_ecma_transforms_react depends on swc_config, which depends on swc_cache, which depends on ahash, which causes some pain when trying to use in a Wasm environment. Looking into it, it seems like it's cleaner to break up the dependency here and instead import directly from swc_cached when needed?

Edit: actually, I don't believe this helps me, but seems like this is good to do anyway?

**BREAKING CHANGE:**

Removes re-export from swc_config.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3969,7 +3969,6 @@ dependencies = [
  "serde",
  "serde_json",
  "sourcemap",
- "swc_cached",
  "swc_config_macro",
 ]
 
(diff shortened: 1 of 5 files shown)


kdy1:
Actually I want to remove `swc_cached` and merge it into `swc_config`

dsherret:
Ah, ok let's close this.

PR #8609, opened 2024-02-05
chore(config): Remove an empty file in `swc_config`

dsherret:
**Description:**

Noticed this can be removed because it's an empty file? I believe it was moved to swc_config_macro

**BREAKING CHANGE:**


kdy1 (commented):
swc-bump:
 - swc_config

kdy1 (approved):

PR #8610, opened 2024-02-06
fix(es/parser): Set class property to `abstract` or `override` even in error states

dsherret:
**Description:**

Ensures that a class property with `abstract` or `override` is marked as such in the AST, even in error states.

**Related issue (if exists):** Part of #1178 (somewhat related)


kdy1 (approved):
swc-bump:
 - swc_ecma_parser

PR #8611, opened 2024-02-06
fix(es/parser): Fix parsing of dynamic source phase imports

dsherret:
**Description:** Fixes parsing of dynamic source phase imports.

**BREAKING CHANGE:**


crates/swc_ecma_parser/src/parser/expr.rs
```diff
@@ -267,21 +267,6 @@ impl<I: Tokens> Parser<I> {
                     })));
                 }
 
-                tok!("import") => {
-                    let import = self.parse_ident_name()?;
-
-                    if is!(self, '.') {
```
dsherret, line 273: This was a red herring and didn't work. For example, this does `is!(self, '.')` then `expect!(self, '.')` below.

kdy1 (approved):
swc-bump:
 - swc_ecma_parser

magic-akari (approved):

PR #8620, opened 2024-02-07
fix(binding/types): Update typings for `options.envs` to match implementation

### LINKED ISSUE #8494: Should the `optimizer.global.envs` type include `Record<string, string>`?
`optimizer.global.envs` type is defined here:

https://github.com/swc-project/swc/blob/9cf91a375bbd87abb16bffa2d6ec44a32059d102/packages/types/index.ts#L868C8-L868C8

I don't know Rust, but with a little intuition and a lot of GPT, I think it should be something like:

```diff
-    envs?: string[];
+    envs?: string[] | Record<string, string>;
```

davidmurdoch:
Fixes #8494

### DIFF
--- a/bindings/binding_core_wasm/src/types.rs
+++ b/bindings/binding_core_wasm/src/types.rs
@@ -905,7 +905,7 @@ export interface GlobalPassOption {
    *
    * Defaults to `["NODE_ENV", "SWC_ENV"]`
    */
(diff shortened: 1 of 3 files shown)


kdy1 (approved):
swc-bump:
 - dbg-swc

PR #8655, opened 2024-02-22
perf(visit): Use qualified paths for generated codes

kdy1:
**Description:**

I'm not sure about the reason, but the impact on compile time was much bigger than I expected.


kdy1 (commented):
swc-bump:
 - swc_visit_macros

PR #8656, opened 2024-02-22
fix(es/codegen): Emit computed getter and setter signatures

dsherret:
**Description:**

Emits computed getter/setter signatures.

**BREAKING CHANGE:**

No


kdy1 (approved):
swc-bump:
 - swc_ecma_codegen

PR #8657, opened 2024-02-23
fix(css/parser): Fix parsing of `@import` without `;` or space

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_css_parser

PR #8659, opened 2024-02-23
fix(es/codegen): Fix replacement when `inline_script` is on

xc2:
**Description:**

The replacements for `inline_script` might be with mistake

https://github.com/swc-project/swc/blob/40682c8d1f55b42b0a086487e740e975244ed42a/crates/swc_ecma_codegen/src/lib.rs#L649

which in terser is:
image

https://github.com/terser/terser/blob/0ef05847fa8c892c109b6d8a9415e9463f3e0995/lib/output.js?plain=1#L414

**BREAKING CHANGE:**

https://github.com/web-infra-dev/rspack/issues/5757


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_codegen

PR #8661, opened 2024-02-25
feat(es/lints): Add `no-cond-assign` rule

ArturAralin:
**Description:**

Here is no-cond-assing rule implementation.

**BREAKING CHANGE:**

No braking changes


kdy1 (approved):
swc-bump:
 - swc_ecma_lints

PR #8676, opened 2024-02-28
fix(es/codegen): Emit `namespace` keyword of `TsModuleDecl` if possible

dsherret:
**Description:** Emits the namespace keyword for TsModuleDecls when able.

Been meaning to open a PR for this one for a while, but the module keyword may be deprecated in the future in TS in non-ambient modules, so it's better to do this.

**BREAKING CHANGE:** No.

**Related issue (if exists):** None.


kdy1 (approved):
swc-bump:
 - swc_ecma_codegen

PR #8677, opened 2024-02-28
fix(es/visit): Skip `TsExprWithTypeArgs` from `noop_visit/fold_type` macros

dsherret:
**Description:** See https://github.com/swc-project/swc/discussions/8669

**BREAKING CHANGE:**

`TsExprWithTypeArgs` is now skipped in these macros. This fix might be a breaking change for anyone who was relying on the previous behaviour even though this is more correct now.


magic-akari (approved):

kdy1 (approved):
swc-bump:
 - swc_ecma_visit

PR #8683, opened 2024-03-02
feat(es/minifier): Allow `expr_simplifier` to do arithmetic with string literals

### LINKED ISSUE #8682: expr_simplifier does not handle numeric operations involving strings
### Describe the feature

`swc_core::ecma::transforms::optimization::simplify::expr_simplifier` should simplify binary expressions when the left, right or both expression is a string literal if the operator isn't addition. Currently, these expressions are not simplified and are left as-is by the simplifier. Example:

```javascript
"10" - 5 // output: 5
"4" / 2 // output: 2
"11" % 2 // output: 1
"10" ** 2 // output: 100

10 - "5" // output: 5
"10" - "5" // output: 5
// same examples as above but with the right side being a string instead of left
```

The file responsible seems to be: https://github.com/swc-project/swc/blob/main/crates/swc_ecma_transforms_optimization/src/simplify/expr/mod.rs

The issue currently is that `get_type().casted_to_number_on_add()` is being checked regardless of the operator. This should only be checked if the operator isn't addition.

_No response_

### Additional context

Babel's `path.evaluate()` correctly handles this.

As far as I know, you simply need to parse the string as a 64-bit float and perform a numeric operation with the parsed float.

leafypout:
**Description:**
This PR implements the feature described for the issue in #8682.

**Related issue:**

 - Closes #8682.


kdy1 (commented):
Can you run `UPDATE=1 cargo test --test projects --test tsc` from `./crates/swc`?

levi-nz:
> Can you run `UPDATE=1 cargo test --test projects --test tsc` from `./crates/swc`?

Looks all good on my end:
`test result: ok. 4562 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 8.87s`

kdy1:
It will never fail with `UPDATE=1`, but it will modify files.

levi-nz:
> It will never fail with `UPDATE=1`, but it will modify files.

Got it. Two files have changed, here are the new versions:

`enumConstantMemberWithString.2.minified.js`:
```js
//// [enumConstantMemberWithString.ts]
var T1, T2, T3, T4, T5, T11, T21, T31;
(T11 = T1 || (T1 = {})).a = "1", T11.b = "12", T11.c = "123", T11[T11.d = NaN] = "d", T11.e = "a1", (T21 = T2 || (T2 = {})).a = "1", T21.b = "12", (T31 = T3 || (T3 = {})).a = "1", T31.b = "12", T31[T31.c = 1] = "c", T31[T31.d = 3] = "d", (T4 || (T4 = {})).a = "1", (T5 || (T5 = {})).a = "12";
```
The only thing that has changed here is `T11.d = "a" - "a"` -> `T11.d = NaN`. When running `"a" - "a"` in NodeJS it yields `NaN` so this seems to be correct, but I'm not sure if that's correct in terms of this unit test.

`enumConstantMemberWithTemplateLiterals.2.minified.js`:
```js
//// [enumConstantMemberWithTemplateLiterals.ts]
var T1, T2, T3, T4, T5, T6, T21, T41, T51, T61;
(T1 || (T1 = {})).a = "1", (T21 = T2 || (T2 = {})).a = "1", T21.b = "2", T21[T21.c = 3] = "c", (T3 || (T3 = {})).a = "11", (T41 = T4 || (T4 = {})).a = "1", T41.b = "11", T41.c = "12", T41.d = "21", T41.e = "211", (T51 = T5 || (T5 = {})).a = "1", T51.b = "12", T51.c = "123", T51[T51.d = 1] = "d", T51[T51.e = 0] = "e", T51.f = "11", T51.g = "123", T51[T51.h = 1] = "h", (T61 = T6 || (T6 = {}))[T61.a = 1] = "a", T61[T61.b = 2] = "b";
```
The only thing that has changed here is `T51.e = "1" - "1"` -> `T51.e = 0`, which also seems to be correct like the first case above, but again I'm not sure if that's correct in the context of this unit test.

kdy1:
Please commit those files

kdy1 (approved):
swc-bump:

PR #8684, opened 2024-03-03
feat(es/lints): Add `no-prototype-builtins` rule

ArturAralin:
**Description:**

Add no-prototype-builtins rule implementation

**BREAKING CHANGE:**

No breaking changes


crates/swc/tests/error_msg.rs
```diff
@@ -36,8 +36,8 @@ fn swcrc_simple() {
     println!("{}", f);
 }
 
-#[testing::fixture("tests/errors/**/input.js")]
-#[testing::fixture("tests/errors/**/input.ts")]
+#[testing::fixture("tests/errors/**/no-prototype-builtins/**/input.js")]
```
kdy1, line 39: Please revert this

kdy1 (changes requested):

crates/swc/tests/error_msg.rs
kdy1: If you want to run specific test only, you can provide argument to `cargo test` like 

`cargo test foo` to run only tests that contains `foo` in the test name

ArturAralin: > Please revert this

Oh, sorry. I'm forgot to discard unnecessary changes

crates/swc_ecma_lints/src/rules/no_prototype_builtins.rs
```diff
@@ -0,0 +1,151 @@
...
+                self.emit_error(self.call_info.method_span.unwrap(), method_name);
+            }
+        }
+
+        self.call_info = prev_call_info;
+    }
+}
+
+impl Visit for NoPrototypeBuiltins {
+    fn visit_opt_chain_base(&mut self, opt_chai_base: &OptChainBase) {
```
kdy1, line 136: This looks like a type

kdy1 (changes requested):

crates/swc_ecma_lints/src/rules/no_prototype_builtins.rs
ArturAralin: I've checked without this visitor and part of my test are failed.

I think here is bug, because `visit_opt_chain_base` actually never visited, but without defining `visit_opt_chain_base` visitor `visit_call_expr` skip reaction on case like `foo?.hasOwnProperty('bar')`

kdy1: I mean `opt_chai_base`

ArturAralin: Oh... I did reading u comment in a little hurry and read "type" instead of "typo" sorry

kdy1: 🤣 I wrote `type`, sorry. It was also a typo.

ArturAralin: got morning damage 🤣

ArturAralin: Anyway it was useful. I'd fixed code after broken merge and added noop_visit_type to prevent extra job

kdy1 (approved):
Not sure about the logic, but approving based on the tests.

---


swc-bump:
 - swc_ecma_lints

kdy1 (approved):
Not sure about the logic, but approving based on the tests.

---


swc-bump:
 - swc_ecma_lints

PR #8686, opened 2024-03-04
fix(es/module): Allow `FileName::Anon` from node resolver

### LINKED ISSUE #8674: `jsc.baseUrl` is not resolving the require/import path
### Discussed in https://github.com/swc-project/swc/discussions/8665

x-ref: #8660

<div type='discussions-op-text'>

<sup>Originally posted by **devjiwonchoi** February 26, 2024</sup>
Hello, fellow engineers and appreciatable SWC maintainers,

I wish to understand if the following behavior is intended and if so, the reason for it:

### TL;DR

Why is `jsc.baseUrl` not resolving the require path as either relative or absolute?

### Context

Using `transform` or `transformSync` from `@swc/core`.

Based on this error message:

```sh
Error: failed to handle: base_dir(`.`) must be absolute. Please ensure that `jsc.baseUrl` is specified correctly.
```

I've understood that the `baseUrl` path must be absolute which differs from `tsconfig.json`, so I'm passing `process.cwd()`.

### Code

```ts
import { foo } from "src/foo"
console.log(foo)
```

```ts
await transform(code, { jsx: { baseUrl: process.cwd() } })
```

Expected:

```js
import { foo } from "./src/foo" // or a resolved absolute path?
console.log(foo)
```

Output:

```ts
import { foo } from "src/foo"
console.log(foo)
```

## Repro

https://github.com/devjiwonchoi/repro-swc-base-url-paths

```sh
pnpm i
pnpm tsx .
```

kdy1:
**Description:**

**Related issue:**
 
 - Closes #8674


kdy1 (commented):
swc-bump:
 - swc_ecma_loader

PR #8711, opened 2024-03-08
perf(es): Do not create tokio runtime if not required

### LINKED ISSUE #8708: Performance regression after v1.4.1-nightly-20240210.1
### Describe the bug

Updating from v1.4.1-nightly-20240210.1 to any version thereafter results in a 10-20% increase in compilation time.

I think the update that caused the slow down is related to the latest wasmer update.

### Input code

```typescript
git clone https://github.com/MetaMask/metamask-extension.git
cd metamask-extension
git checkout webpack-perf-regression
nvm use # or use Node v20 if you don't use nvm for managing node version
corepack enable # makes it use correct yarn version
mv .metamaskrc.dist .metamaskrc
yarn
time yarn webpack --no-cache # take note of the time
yarn up @swc/core@1.4.1-nightly-20240210.1 # install faster version of @swc/core
time yarn webpack --no-cache # this should be about 10-20% faster
```
```


### Config

```json
n/a, this is a webpack-based project using a basic webpack loader.

Loader: https://github.com/MetaMask/metamask-extension/blob/webpack-perf-regression/development/webpack/utils/loaders/swcLoader.ts

Loader usage is here: https://github.com/MetaMask/metamask-extension/blob/e885236f1b42969106e8b88a0d7e66de3921e999/development/webpack/webpack.config.ts#L105
```

https://github.com/MetaMask/metamask-extension/tree/webpack-perf-regression

### SWC Info output

(I'm away from my desk, but will fill this out later)

### Expected behavior

Performance should remain roughly the same between versions.

### Actual behavior

It got slower 🥲

### Version

v1.4.5

### Additional context

Original discussion: https://discord.com/channels/889779439272075314/889780921002246154/1215428591945383936

kdy1:
**Description:**

**Related issue:**

 - Closes #8708.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3717,6 +3717,7 @@ dependencies = [
  "swc_timer",
  "swc_visit",
  "testing",
+ "tokio",
  "tracing",
  "url",
  "walkdir",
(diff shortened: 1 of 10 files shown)


kdy1 (commented):
swc-bump:
 - swc_plugin_runner

PR #8724, opened 2024-03-10
refactor(es): Prepare `wasm32-wasi-preview1-threads` target support

Brooooooklyn:
`PathBuf` and `rayon` is usable in `wasm32-wasi-preview1-threads` target, so we can enable it.

### DIFF
--- a/bindings/binding_core_node/Cargo.toml
+++ b/bindings/binding_core_node/Cargo.toml
@@ -63,5 +63,5 @@ swc_core = { version = "0.90.18", features = [
   "ecma_visit",
   "base_node",
   "base_concurrent",
-] }
+]}
 swc_malloc = "0.5.10"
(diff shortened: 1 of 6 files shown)


.yarnrc.yml
```diff
@@ -3,3 +3,5 @@ compressionLevel: mixed
 enableGlobalCache: false
 
 nodeLinker: node-modules
+
+yarnPath: .yarn/releases/yarn-4.1.1.cjs
```
kdy1, line 7: It seems like `.yarn` is ignored by git

kdy1 (changes requested):
CI failed

kdy1 (approved):

kdy1 (commented):
swc-bump:
 - swc_ecma_transforms_base

PR #8750, opened 2024-03-15
feat(es/minifier): Handle more indexing expression

### LINKED ISSUE #8747: `expr_simplifier` does not simplify computed `MemberExpr`s with non-literal properties
### Describe the feature

`expr_simplifier` does not simplify computed `MemberExpr`s that have an expression as the index.

These two examples are not simplified, but both resolve to `undefined`:
```js
[][[]] // undefined
[][0] // undefined
```

Both of these should be simplified to `undefined`

_No response_

### Additional context

_No response_

levi-nz:
**Description:**

This PR is a WIP that fixes all known issues in #8747.

**Related issue:**

 - Closes #8747


levi-nz:
Not totally sure if the changed code for array literals is correct -- the code for this is a little more complicated than strings.

`[][0]` works correctly for now, but not `[][[]]`. Will look into this later.

levi-nz:
Another broken test case:

```javascript
var _ = {0.5: 'test'}[0.5];
```

Currently simplifies to:
```javascript
var _ = {
    0.5: 'test'
}[0.5];
```

Should simplify to:
```javascript
var _ = 'test';
```

levi-nz:
Only one potential issue so far, that `KnownOp` no longer implements `Eq` because of `Index` now taking `f64` instead of `i64`, but seems OK so far to me.

levi-nz:
@kdy1 The only remaining issue in terms of code changes is this line. Currently `is_literal(props)` does not work with the unit test below because the `0.5` is not considered a literal. I'm not sure why this `fract()` check exists and I'm not sure if changing this would break something.

https://github.com/levi-nz/swc/blob/454378d7eaac82bae40c5a770888e84580a1381b/crates/swc_ecma_utils/src/lib.rs#L1883

```rust
fold("({0.5: 'a'})['0.5']", "'a';");
```

levi-nz:
I'm pretty sure `Index(f64)` can be changed back to `Index(i64)` as well, as `IndexStr` is always used for objects.

crates/swc_ecma_utils/src/lib.rs
```diff
@@ -1880,7 +1880,7 @@ impl Visit for LiteralVisitor {
             PropName::Str(ref s) => self.cost += 2 + s.value.len(),
             PropName::Ident(ref id) => self.cost += 2 + id.sym.len(),
             PropName::Num(n) => {
-                if n.value.fract() < 1e-10 {
+                if n.value.fract() < 1e-10 { // TODO: This breaks object simplifying. Why is this here?
```
kdy1, line 1883: I think we may remove this

kdy1: `LiteralVisitor` is used to convert a javascript literal to `JSON.parse` when the object is huge, as `JSON.parse` is faster than javascript object literal.

I'm not sure why I implemented it in such a way. Maybe I just wanted to make it work for easy cases.

levi-nz:
I have ran `UPDATE=1 cargo test --test projects --test tsc`, but not sure if all of these files should be changed, especially `crates/swc/tests/tsc-references/templateStringInIndexExpressionES6.2.minified.js`, which is now empty.

kdy1 (changes requested):
CI failed

levi-nz:
This is getting a little complicated with property names, since we need to either correctly handle stuff like `[].push` not being replaced, or simply not replace invalid properties at all.

The solution I've implemented works and I don't think there's any issues, besides the fact that code like this will break:
```js
Object.prototype.foobar = 5;
var _ = ({}).foobar; // returns 5, but will return undefined from the minifier
```

If there's no way to account for this correctly, we should ideally make this a configurable option where the user can make the minifier replace stuff like `[].invalid` with `void 0` or just make it not replace anything at all.

If we continue with this approach, should we include deprecated properties like `Object.watch` to avoid any confusion (currently replaced with `void 0`), even though these no longer exist? This exists in `vue.js`. Ref: https://stackoverflow.com/questions/1029241/object-watch-for-all-browsers

levi-nz:
Babel's minify doesn't modify expressions like `[][[]]`, but Babel's `path.evaluate()` API lets you transform these expressions. Should we perhaps make this some sort of API instead? Not sure where to go from here, but this should definitely be an option to make the simplifier as powerful as Babel's `path.evaluate()` option.

kdy1 (changes requested):
Please move logics to https://github.com/swc-project/swc/blob/8eac0bec49f4f260a261d87f174ed2913c4bee63/crates/swc_ecma_minifier/src/compress/pure/evaluate.rs

`swc_ecma_utils` is not a good place for this kind of optimizations

levi-nz:
> Please move logics to https://github.com/swc-project/swc/blob/8eac0bec49f4f260a261d87f174ed2913c4bee63/crates/swc_ecma_minifier/src/compress/pure/evaluate.rs
> 
> `swc_ecma_utils` is not a good place for this kind of optimizations

Do we not care about modified prototypes? For example, in the current release version: https://github.com/swc-project/swc/blob/56e03a19602b36a7ebac9ccc899065101bb385af/crates/swc_ecma_transforms_optimization/src/simplify/expr/mod.rs#L165

We replace `x[-1]` with `undefined`, but that would break this code if minified:
```js
Array.prototype[-1] = "foo";
[][-1] // outputs "foo", but swc replaces with void 0
```

Shouldn't this logic be moved to compress as well?

kdy1:
There's an option named `pristine_globals` to mark globals as pristine, but I didn't bother to follow it in all places. Some rules uses it, but not all.

levi-nz:
Got it, I will update the minifier accordingly to try use it -- I think that option should be checked here in the aforementioned case

levi-nz:
Not sure why this exists, perhaps a bug with side effects? https://github.com/swc-project/swc/blob/56e03a19602b36a7ebac9ccc899065101bb385af/crates/swc_ecma_transforms_optimization/src/simplify/expr/mod.rs#L258

Current test case `fold("[1,2,3,4,5,6,7,8,9,10][4 + []]", "1");` fails, resulting in `0, 5` instead of just `5`, removing that if block works but I think this may break things. Looking into why this is happening

levi-nz:
I seem to have fixed this but not sure if it will cause side effects anywhere, will try take a look. It appears as `ArrayLiteral`s being replaced were always being replaced with a `SeqExpr` and if no side effects existed then there was a dummy `0` being added to the front. It just gets replaced with the single `Expr` now instead.

levi-nz:
Also, CI currently fails because of this. Is this normal? These unit tests don't seem to work on Windows, will try this on Mac later.

```
---- fixture_tests__terser__compress__evaluate__unsafe_string_bad_index__input_js stdout ----
Input: /home/runner/work/swc/swc/crates/swc_ecma_minifier/tests/terser/compress/evaluate/unsafe_string_bad_index/input.js
---- Config -----
{
    "evaluate": true,
    "unsafe": true
}

---- Input -----
console.log("1234".a + 1, "1234"["a"] + 1, "1234"[3.14] + 1);

thread 'fixture_tests__terser__compress__evaluate__unsafe_string_bad_index__input_js' panicked at crates/swc_ecma_minifier/src/compress/mod.rs:170:17:
Infinite loop detected (current pass = 202)
Code    0:









console.log("1234".a + 1, "1234"["a"] + 1, "1234"[3.14] + 1);


failures:
    fixture_tests__terser__compress__evaluate__unsafe_string_bad_index__input_js

test result: FAILED. 2674 passed; 1 failed; 27 ignored; 0 measured; 0 filtered out; finished in 96.08s

error: test failed, to rerun pass `-p swc_ecma_minifier --test compress`
Error: Process completed with exit code 101.
```

levi-nz:
Not sure if I got the `span` stuff correct. There is some copy pasta going on here from `simplify::expr`, should probably move that stuff into their own util functions.

Where should I create unit tests for the new compress functionality?

kdy1:
You can create a directory and place an `input.js` and `config.json` in https://github.com/swc-project/swc/tree/56e03a19602b36a7ebac9ccc899065101bb385af/crates/swc_ecma_minifier/tests/fixture

Note that you need to touch (or save again) `tests/compress.rs` once after adding a test file

levi-nz:
Will add `watch` and `unwatch` as exclusions for `Object`, even though they don't exist anymore. This way users can still use pristine_globals=true when using scripts like this to polyfill: https://gist.github.com/eligrey/384583

Not sure if there's any other exclusions we should add for property/method polyfills that don't exist anymore

crates/swc_ecma_minifier/src/compress/pure/mod.rs
```diff
@@ -66,6 +67,8 @@ pub(crate) fn pure_optimizer<'a>(
         data,
         ctx: Default::default(),
         changed: Default::default(),
+        in_callee: false,
+        in_left_side_assign: false
```
kdy1, line 71: Please use `ctx` instead

crates/swc_ecma_minifier/src/compress/pure/mod.rs
```diff
@@ -79,6 +82,8 @@ struct Pure<'a> {
     data: Option<&'a ProgramData>,
     ctx: Ctx,
     changed: bool,
+    in_callee: bool,
+    in_left_side_assign: bool
```
kdy1, line 86: Same here

kdy1 (changes requested):

levi-nz:
To do:
- Move stuff to ctx
- Handle known object properties
- Possibly add more unit tests

Will comment here again when everything is finalised

crates/swc_ecma_minifier/tests/benches-full/vue.js
```diff
@@ -185,7 +185,7 @@
     UA && UA.indexOf('android');
     var isIOS = UA && /iphone|ipad|ipod|ios/.test(UA) || 'ios' === weexPlatform;
     UA && /chrome\/\d+/.test(UA), UA && /phantomjs/.test(UA);
-    var isFF = UA && UA.match(/firefox\/(\d+)/), nativeWatch = {}.watch, supportsPassive = !1;
```
kdy1, line 188: This looks like a bug

levi-nz: `{}.watch` is replaced with `void 0` since the property `watch` doesn't exist on `Object`'s prototype, but I think I'll add `watch` and `unwatch` anyway to avoid any issues as these properties used to exist but don't exist anymore (even the MDN documentation for them have been removed).

levi-nz:
Is there a reason why stuff like `[1][0]` was replaced with `(0, 1)` before instead of just `1`? The tests seem to pass fine but I want to be sure. I don't see why it should be replaced with a SeqExpr instead of just a Literal if there are no side effects.

levi-nz:
`cargo test -p swc_ecma_minifier terser_exec_tests__terser__compress__evaluate__unsafe_string_bad_index__input_js --features concurrent` currently fails with a weird infinite loop bug, doesn't seem to be related to `optimize_member_expr`, need to investigate further

kdy1:
`(0, foo)` is about making `this` undefined

levi-nz:
> `(0, foo)` is about making `this` undefined

Does this apply to objects too? Like should `({a: 5}).a` be replaced with `(0, 5)` or just `5`? Currently it's replaced with the latter but I want to be sure

levi-nz:
Need to handle known object properties in simplify's `optimize_member_expr`. Will require a function signature change to accept obj and prop instead of expr. This PR should be ready for review tomorrow.

kdy1:
> 
> (0, foo) is about making this undefined
>
> Does this apply to objects too? Like should ({a: 5}).a be replaced with (0, 5) or just 5? Currently it's replaced with the latter but I want to be sure


I don't think so. `this` only matters for member exprs or **direct** eval calls

levi-nz:
Running `cargo test run_fixture_test_tests__exec__issues_6xxx__6878__1__exec_js` results in this error on both main and issue-8747 branch. Yarn installed with `npm i -g yarn` and `corepack enable`. Not sure how to fix?

```
levi@levi-kde-neon:~/RustroverProjects/levi-nz/swc$ cargo test run_fixture_test_tests__exec__issues_6xxx__6878__1__exec_js
    Finished test [unoptimized + debuginfo] target(s) in 0.39s
     Running unittests src/lib.rs (target/debug/deps/ast_node-e9d29c534642bab4)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/debug/deps/better_scoped_tls-2dc2c5ea5a8d762c)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/debug/deps/binding_macros-c052f01c3c2f3382)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src/main.rs (target/debug/deps/dbg_swc-05912a335a982368)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/debug/deps/from_variant-86e39519e21950c5)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/debug/deps/jsdoc-f39585929aa7c829)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 9 filtered out; finished in 0.00s

     Running tests/fixture.rs (target/debug/deps/fixture-a6c989228dc4b47b)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 235 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/debug/deps/preset_env_base-e8b904940a0c4c1c)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/debug/deps/string_enum-f4111496889125a4)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/simple.rs (target/debug/deps/simple-07ac1a7eda7c6644)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/debug/deps/swc-b072fffd3fe75c6a)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 5 filtered out; finished in 0.00s

     Running tests/error_msg.rs (target/debug/deps/error_msg-423e285177753c6d)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 104 filtered out; finished in 0.00s

     Running tests/exec.rs (target/debug/deps/exec-ce806a5ca5bc9fa2)

running 1 test
➤ YN0000: · Yarn 4.0.2
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ YN0000: └ Completed
➤ YN0000: ┌ Link step
➤ YN0000: └ Completed
➤ YN0000: · Done in 0s 194ms
$ dprint fmt
Formatted 193 files.
$ dprint fmt "scripts/*.js" -c scripts/.dprint.json
test run_fixture_test_tests__exec__issues_6xxx__6878__1__exec_js has been running for over 60 seconds.4.2 want: ^5.3.0
npm WARN deprecated @types/terser@3.12.0: This is a stub types definition. terser provides its own type definitions, so you do not need this installed.
npm WARN deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm WARN deprecated @babel/plugin-proposal-class-properties@7.18.6: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-class-properties instead.
npm WARN deprecated @babel/plugin-proposal-object-rest-spread@7.20.7: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-object-rest-spread instead.
npm WARN deprecated core-js@2.6.12: core-js@<3.23.3 is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Some versions have web compatibility issues. Please, upgrade your dependencies to the actual version of core-js.
npm ERR! code 1
npm ERR! path /home/levi/RustroverProjects/levi-nz/swc/node_modules/wasm-sjlj
npm ERR! command failed
npm ERR! command sh -c node-gyp rebuild
npm ERR! gyp info it worked if it ends with ok
npm ERR! gyp info using node-gyp@10.0.1
npm ERR! gyp info using node@20.12.2 | linux | x64
npm ERR! gyp info find Python using Python version 3.10.12 found at "/usr/bin/python3"
npm ERR! gyp info spawn /usr/bin/python3
npm ERR! gyp info spawn args [
npm ERR! gyp info spawn args '/home/levi/.nvm/versions/node/v20.12.2/lib/node_modules/npm/node_modules/node-gyp/gyp/gyp_main.py',
npm ERR! gyp info spawn args 'binding.gyp',
npm ERR! gyp info spawn args '-f',
npm ERR! gyp info spawn args 'make',
npm ERR! gyp info spawn args '-I',
npm ERR! gyp info spawn args '/home/levi/RustroverProjects/levi-nz/swc/node_modules/wasm-sjlj/build/config.gypi',
npm ERR! gyp info spawn args '-I',
npm ERR! gyp info spawn args '/home/levi/.nvm/versions/node/v20.12.2/lib/node_modules/npm/node_modules/node-gyp/addon.gypi',
npm ERR! gyp info spawn args '-I',
npm ERR! gyp info spawn args '/home/levi/.cache/node-gyp/20.12.2/include/node/common.gypi',
npm ERR! gyp info spawn args '-Dlibrary=shared_library',
npm ERR! gyp info spawn args '-Dvisibility=default',
npm ERR! gyp info spawn args '-Dnode_root_dir=/home/levi/.cache/node-gyp/20.12.2',
npm ERR! gyp info spawn args '-Dnode_gyp_dir=/home/levi/.nvm/versions/node/v20.12.2/lib/node_modules/npm/node_modules/node-gyp',
npm ERR! gyp info spawn args '-Dnode_lib_file=/home/levi/.cache/node-gyp/20.12.2/<(target_arch)/node.lib',
npm ERR! gyp info spawn args '-Dmodule_root_dir=/home/levi/RustroverProjects/levi-nz/swc/node_modules/wasm-sjlj',
npm ERR! gyp info spawn args '-Dnode_engine=v8',
npm ERR! gyp info spawn args '--depth=.',
npm ERR! gyp info spawn args '--no-parallel',
npm ERR! gyp info spawn args '--generator-output',
npm ERR! gyp info spawn args 'build',
npm ERR! gyp info spawn args '-Goutput_dir=.'
npm ERR! gyp info spawn args ]
npm ERR! gyp: binding.gyp not found (cwd: /home/levi/RustroverProjects/levi-nz/swc/node_modules/wasm-sjlj) while trying to load binding.gyp
npm ERR! gyp ERR! configure error 
npm ERR! gyp ERR! stack Error: `gyp` failed with exit code: 1
npm ERR! gyp ERR! stack at ChildProcess.<anonymous> (/home/levi/.nvm/versions/node/v20.12.2/lib/node_modules/npm/node_modules/node-gyp/lib/configure.js:271:18)
npm ERR! gyp ERR! stack at ChildProcess.emit (node:events:518:28)
npm ERR! gyp ERR! stack at ChildProcess._handle.onexit (node:internal/child_process:294:12)
npm ERR! gyp ERR! System Linux 6.5.0-28-generic
npm ERR! gyp ERR! command "/home/levi/.nvm/versions/node/v20.12.2/bin/node" "/home/levi/.nvm/versions/node/v20.12.2/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
npm ERR! gyp ERR! cwd /home/levi/RustroverProjects/levi-nz/swc/node_modules/wasm-sjlj
npm ERR! gyp ERR! node -v v20.12.2
npm ERR! gyp ERR! node-gyp -v v10.0.1
npm ERR! gyp ERR! not ok

npm ERR! A complete log of this run can be found in: /home/levi/.npm/_logs/2024-05-24T21_56_02_426Z-debug-0.log
test run_fixture_test_tests__exec__issues_6xxx__6878__1__exec_js ... FAILED

failures:

---- run_fixture_test_tests__exec__issues_6xxx__6878__1__exec_js stdout ----
Input: /home/levi/RustroverProjects/levi-nz/swc/crates/swc/tests/exec/issues-6xxx/6878/1/exec.js
thread 'run_fixture_test_tests__exec__issues_6xxx__6878__1__exec_js' panicked at crates/swc/tests/exec.rs:109:13:
assertion failed: status.success()


failures:
    run_fixture_test_tests__exec__issues_6xxx__6878__1__exec_js

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 546 filtered out; finished in 108.41s

error: test failed, to rerun pass `-p swc --test exec`
```

kdy1:
What's `wasm-sjlj`? I don't think SWC uses it.

levi-nz:
`foo?.bar.baz?.()` should be optimized to `(void 0).baz?.()`, but this is ignored for now. For some reason the AST becomes this

```
OptChainExpr { span: 53..65#0, optional: false, base: Member(MemberExpr { span: 53..65#0, obj: Unary(UnaryExpr { span: 45..47#0, op: "void", arg: Lit(Num(Number { span: 45..47#0, value: 0.0, raw: None })) }), prop: Ident(Ident { span: 62..65#0, sym: "baz", optional: false }) }) }
```

instead of

```
OptChainExpr { span: 11..27#0, optional: true, base: Call(OptCall { span: 11..27#0, callee: Member(MemberExpr { span: 11..23#0, obj: Unary(UnaryExpr { span: 12..18#0, op: "void", arg: Lit(Num(Number { span: 17..18#0, value: 0.0, raw: Some("0") })) }), prop: Ident(Ident { span: 20..23#0, sym: "baz", optional: false }) }), args: [], type_args: None }) }
```

which results in it becoming `(void 0)?.()` (due to `eval_opt_chain`'s `OptChainBase::Member` optimization that removes pure undefined or null). If I check `self.ctx.is_opt_call` in `eval_opt_chain` it works as intended but I'm not sure if this would be correct since the AST appears to be messed up. Might come back to this later, otherwise will just have this as an ignored case.

crates/swc_ecma_minifier/Cargo.toml
```diff
@@ -49,6 +49,7 @@ ryu-js            = { workspace = true }
 serde             = { workspace = true, features = ["derive"] }
 serde_json        = { workspace = true }
 tracing           = { workspace = true }
+phf = { version = "0.11.2", features = ["macros"] }
```
kdy1, line 52: Can you apply https://github.com/mainmatter/cargo-autoinherit ?

kdy1:
Also, can you rebase? (And thank you so much for the PR!)

levi-nz:
> Also, can you rebase? (And thank you so much for the PR!)

Just saw this, will do

levi-nz:
Had to manually fix some files as rebasing messed a few up, hope that's OK

crates/swc_ecma_minifier/src/compress/pure/evaluate.rs
```diff
@@ -639,6 +641,29 @@ impl Pure<'_> {
...
+            "before: optimize_member_expr: {}",
+            dump(&*member_expr, false)
+        );
+
+        if let Some(replacement) =
+            self.optimize_member_expr(&mut member_expr.obj, &member_expr.prop)
+        {
+            *e = replacement;
+            self.changed = true;
+
```
kdy1, line 661: Can you add a `report_change!` call here?

crates/swc_ecma_transforms_optimization/src/simplify/expr/tests.rs
```diff
@@ -1575,3 +1575,53 @@ fn test_export_default_paren_expr() {
     fold_same("import fn from './b'; export default (function fn1 () {});");
     fold("export default ((foo));", "export default foo;");
 }
+
+#[test]
+fn test_issue8747() {
```
kdy1, line 1580:
```suggestion
fn test_issue_8747() {
```

kdy1 (commented):
Thank you!

---

swc-bump:
 - swc_ecma_utils
 - swc_ecma_minifier
 - swc_core

kdy1 (commented):
Thank you!

---

swc-bump:
 - swc_ecma_utils
 - swc_ecma_minifier
 - swc_core

kdy1 (approved):

Issue #8755, opened 2024-03-17
export default named function should report top  level name conflict


stormslowly:
### Describe the bug

```js
export default function x (){}
let x =1;
```

this code should report that x already defined,  but swc just compiles it as ok 

### Input code

```typescript
export default function x (){}
let x =1;
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.4.8&code=H4sIAAAAAAAAA0utKMgvKlFISU1LLM0pUUgrzUsuyczPU6hQ0NBUqK615spJLQFybBUMrbkAmMVZhSwAAAA%3D&config=H4sIAAAAAAAAA22QOw6DMAyGd06BPLN26daxQw8RpQaFxklkGwmEuHtDCpRK3eL%2F8TnyXNU19GLhWs%2F5mQccE7IjDGr8oWb9hZjulCLrTQRZXQySfeUBS2RpPv1kONvnpkyZNWYF0JIRyy4pNLvby2q1xssvRw13qKUlly0O5IJrpzPcRkqMIjvj4JIJnce%2FaB%2BjnJ2iA8XnUArbIXRK67RuoBh6gW9y%2F8WxEZw89nq5yFK9Abnt2jdaAQAA

### Expected behavior

should report error like below
```
Identifier 'x' has already been declared. (2:4)

  1 | export default function x(){}
> 2 | let x =1 ;
```

this error is copied from https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=KYDwDg9gTgLgBAE2AMwIYFcA29noHYDGMAlhHnCABQCUA3gL4BQmw8IcAvAIxwDcQA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2&prettier=false&targets=&version=7.24.0&externalPlugins=&assumptions=%7B%7D

### Version

1.4.8

### Additional context

[...]

PR #8760, opened 2024-03-18
Fix/binding dup in export default fn_expr or class_expr

stormslowly:
**Description:**
add DefaultDecl's FnExpr and ClassExpr's ident to bindings. 
because all of them should be treated as hoisted Fn/Class Declare

```txt
  × the name `x` is defined multiple times
   ╭─[examples/all.js:1:1]
 1 │ export default class x{}
   ·                      ┬
   ·                      ╰── previous definition of `x` here
 2 │
 3 │ let x = 1;
   ·     ┬
   ·     ╰── `x` redefined here
 4 │ let t = function x(){};
   ╰────
```

**BREAKING CHANGE:**
No.

https://github.com/swc-project/swc/issues/8755


magic-akari (commented):
It's better to add some test cases.

stormslowly:
> It's better to add some test cases.

cases added ，is there any other cases  needed

crates/swc_ecma_parser/tests/tsc/multipleDefaultExports03.ts
```diff
@@ -4,5 +4,5 @@
 export default class C {
 }
 
-export default class C {
```
kdy1, line 7: Please revert this

kdy1: `tsc` test suite is **copied** tests so we should not modify the input

stormslowly: but the duplicated-binding error  report before the export default twice
so i need to update the expect error to dup-binding error ?

kdy1: Yes

stormslowly: done

kdy1 (approved):
swc-bump:
 - swc_ecma_lints
 - swc_core

PR #8768, opened 2024-03-21
fix(es/helpers): Fix metadata of decorators being `undefined`

sschen86:
when use "decorator stage3" with this option that `context.metadata` will be `undefined`

```js
const option = {
    jsc: {
        parser: {
            syntax: 'typescript',
            decorators: true,
        },
        transform: {
            decoratorVersion: '2022-03',
        },
        target: 'es5',
        externalHelpers: true, // external @swc/helpers
    },
    module: {
        type: 'es6',
    }
}
```

if set `externalHelpers` to `false`, `context.metadata` is right

example code 

```ts
function Get() {
  return (value, context) => {
    console.info(context.metadata); // out undefind
  };
}

class MyClass{
    @Get()
    doSome(){
        //..
    }
}
```


kdy1 (changes requested):
Please fix https://github.com/swc-project/swc/blob/ad932f0921411364b801b32f60eaf98f8629e812/packages/helpers/esm/_apply_decs_2203_r.js, too

kdy1 (changes requested):
Why did you revert changes in `swc_ecma_transforms_base`?

sschen86:
> Why did you revert changes in `swc_ecma_transforms_base`?

What, pray, is the meaning of this?

I only modified this file（packages/helpers/esm/_apply_decs_2203_r.js） and added the missing variable `metadata`

kdy1 (changes requested):
I misread it. Sorry.
Please modify https://github.com/swc-project/swc/blob/b798632a829eca950ede021d3e2db57f8649523c/crates/swc_ecma_transforms_base/src/helpers/_apply_decs_2203_r.js, too

sschen86:
> I misread it. Sorry. Please modify https://github.com/swc-project/swc/blob/b798632a829eca950ede021d3e2db57f8649523c/crates/swc_ecma_transforms_base/src/helpers/_apply_decs_2203_r.js, too

This file is correct and there is no missing variable `metadata`. It does not need to be modified

kdy1 (approved):
swc-bump:
 - dbg-swc

PR #8777, opened 2024-03-26
build(bindings/node): Update `sourcemap`

kdy1:


kdy1 (commented):
swc-bump:
  - dbg-swc

PR #8784, opened 2024-03-27
fix(es/module): Fix `jsc.paths` with `resolveFully`

### LINKED ISSUE #8782: Directory import error on compiled output
### Describe the bug

Hi everyone,

I have created a small reproducible example of the error which you can find [here](https://github.com/stelescuraul/dep-cruiser-issue).

Using path aliases in `tsconfig` and `swcrc`, the index.js filename is stripped from the end of the import path when using ESM modules. 

```
> npx swc -d ./dist -D --delete-dir-on-start --only="**/*.js,**/*.ts,**/*.cjs,**/*.mjs" --ignore ./node_modules,./dist .
> node dist/projects/api/main.js

Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import 'workspaces/dist/libs/math' is not supported resolving ES modules imported from workspaces/dist/projects/api/main.js
```

The output of in the `dist/projects/api/main.js` looks like this:
```js
import { sum } from "../../libs/math"; <<-- should be /math/index.js
```

Renaming the file "index.js" to anything else resolves the issue and the path is added correctly. EG renaming it to `main.js`:
```js
import { sum } from "../../libs/math/main.js"; <<-- main.js is correctly appended
```

This issue seems to have been introduced in `1.3.106`. last version that worked with this setup was `1.3.105`

### Input code

```typescript
// projects/api/main.ts

import { sum } from "@foo/math";

console.log(sum(1, 2));
```

### Config

```json
{
  "isModule": true,
  "jsc": {
    "target": "es2022",
    "keepClassNames": true,
    "loose": false,
    "baseUrl": ".",
    "paths": {
      "@foo/math": ["libs/math/index.js"],
      "@foo/math/*": ["libs/math/*"],
      "@foo/common": ["libs/common/main.js"]
    },
    "parser": {
      "syntax": "typescript"
    }
  },
  "minify": false,
  "module": {
    "type": "es6",
    "resolveFully": true
  }
}
```

https://github.com/stelescuraul/dep-cruiser-issue

### SWC Info output

_No response_

### Expected behavior

Correctly append `/index.js` for path aliases.

### Actual behavior

_No response_

### Version

>1.3.105

### Additional context

_No response_

kdy1:
**Description:**

**Related issue:**

 - Closes #8782

### DIFF
--- a/crates/swc/tests/fixture/issues-7xxx/7829/1/input/.swcrc
+++ b/crates/swc/tests/fixture/issues-7xxx/7829/1/input/.swcrc
@@ -1,6 +1,7 @@
 {
     "module": {
-        "type": "es6"
+        "type": "es6",
+        "resolveFully": true
     },
     "jsc": {
         "baseUrl": ".",
--- a/crates/swc/tests/fixture/issues-7xxx/7829/1/output/1.js
+++ b/crates/swc/tests/fixture/issues-7xxx/7829/1/output/1.js
@@ -1,2 +1,2 @@
-import { fn } from "./libs/pkg/src/index.ts";
+import { fn } from "./libs/pkg/src/index.js";
 console.log(fn);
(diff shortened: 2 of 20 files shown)


kdy1 (commented):
swc-bump:
 - swc_ecma_loader

Issue #8788, opened 2024-03-28
Bugfix `bugfix/transform-safari-id-destructuring-collision-in-function-expression` does not work for transpiled class methods


AlfonzAlfonz:
### Describe the bug

`bugfix/transform-safari-id-destructuring-collision-in-function-expression` should rename arguments if they are the same as the name of the function. This works fine for functions but ignores functions which are transpiled from classes.

### Input code

```typescript
class Test {
  id(id) {
    return id;
  }
}

const test = new Test();

console.log(test.id("abc"));
```

### Config

```json
{
  "$schema": "https://json.schemastore.org/swcrc",
  "minify": false,
  "env": {
    "bugfixes": true,
    "debug": true,
    "targets": {
      "safari": "5"
    }
  }
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.4.11&code=H4sIAAAAAAAAA2WNMQ7DMAwDd72CY7s0Dwjyi3zAsJVAgCMBtjwV%2BXuVIUCBcCNPpHJNvWPl7vgSAdMEKUhtHwerQzrUHI01HVyCS3lJecctQo19NI1sDnvSSfSs39VsGj8CLtiGZhfT%2F7ELW%2BVPtf3K5hj7AWcpMXSbAAAA&config=H4sIAAAAAAAAA1WOTQqEMAyF9z1Fydqtm7lNrKlkGIskcVCkd59UYcRNeO%2FL3ztCjPDWBK94uHSzoCjJ3zvRvRhuToDSjJqEF4Pu7FavtUmYuXDefSjjR%2BlEVL732WGdMm%2BkTkxWutZhJOdPZCgTmT4CYEbhFqCH%2B22oPwVP9l29AAAA

### SWC Info output

    Operating System:
        Platform: darwin
        Arch: arm64
        Machine Type: arm64
        Version: Darwin Kernel Version 23.1.0: Mon Oct  9 21:28:12 PDT 2023; root:xnu-10002.41.9~6/RELEASE_ARM64_T8103
        CPU: (8 cores)
            Models: Apple M1

    Binaries:
        Node: 20.11.1
        npm: 10.2.4
        Yarn: 1.22.22
        pnpm: 8.14.1

    Relevant Packages:
        @swc/core: 1.4.11
        @swc/helpers: N/A
        @swc/types: N/A
        

    SWC Config:
        output: N/A
        .swcrc path: N/A

    Next.js info:
        output: N/A

[...]

PR #8796, opened 2024-04-01
fix(es/bugfixes): Fix fn transform in nameless fns

AlfonzAlfonz:
**Description:**

This pull request fixes `bugfix/transform-safari-id-destructuring-collision-in-function-expression` module. Previously the transform ignored code inside of function which did not have specified identifier. Now the visitor should go through the content of the nameless functions.

https://github.com/swc-project/swc/issues/8788


kdy1 (approved):
swc-bump:
 - swc_ecma_compat_bugfixes
 - swc_core

PR #8814, opened 2024-04-04
fix(es/minifier): Respect `top_retain` for top-level functions

kdy1:


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier
 - swc_core

PR #8818, opened 2024-04-05
fix(es/decorators): Fix capacity overflow with decorators

dsherret:
**Description:**

Was looking through the code again and noticed this.

**Related issue (if exists):** No


crates/swc_ecma_transforms_proposal/tests/decorators/issue-7358/2/output.js
```diff
@@ -0,0 +1,42 @@
+export function test() {
+    var _dec, _initClass, _dec1, _dec2, _initProto;
+    let _Foo;
+    _dec = decorate(), _dec1 = decorate(), _dec2 = decorate();
```
dsherret, line 4: I added a test for this because I think in my last PR this expression was accidentally above the `var` decls, which is not a big deal, but not ideal.

kdy1 (approved):
swc-bump:
 - swc_core

Issue #8817, opened 2024-04-05
Soundness issue in `swc_visit`


mmastrac:
`swc_visit` has a soundness issue, detected by miri. In `Map<T>`, a reference to a "forgotten" self is kept. This reference is no longer valid when we try to use it as a pointer. While it is likely this works in most cases, there's a chance the compiler may optimize this into code that causes a crash. Instead, `Box::into_raw` should be used to get a pointer soundly.

Miri error: 

```
error: Undefined Behavior: attempting a read access using <6189435> at alloc1599974[0x0], but that tag does not exist in the borrow stack for this location
   --> /Users/matt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/swc_visit-0.5.11/src/util/map.rs:29:29
    |
29  |             ptr::write(p, f(ptr::read(p)));
    |                             ^^^^^^^^^^^^
    |                             |
    |                             attempting a read access using <6189435> at alloc1599974[0x0], but that tag does not exist in the borrow stack for this location
    |                             this error occurs as part of an access at alloc1599974[0x0..0x60]
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <6189435> was created by a SharedReadWrite retag at offsets [0x0..0x60]
   --> src/transpiling/mod.rs:403:7
    |
```

Likely fix:

```
impl<T> Map<T> for Box<T> {
    fn map<F>(mut self, f: F) -> Self
    where
        F: FnOnce(T) -> T,
    {
        // Leak self in case of panic.
        let p = Box::into_raw(self);

        unsafe {
            ptr::write(p, f(ptr::read(p)));

            // Recreate self from the raw pointer.
            Box::from_raw(p)
        }
    }
}
```

PR #8819, opened 2024-04-05
fix(visit): Improve `Map` implementation for `Box`

dsherret:
**Description:** See #8817

**Related issue (if exists):** #8817


kdy1 (approved):
Thank you!


---

swc-bump:
 - swc_visit
 - swc_core

PR #8827, opened 2024-04-08
fix(plugin/runner): Pin version of `virtual-fs`

kdy1:


kdy1 (commented):
swc-bump:
 - swc_core
 - swc_plugin_runner

PR #8843, opened 2024-04-11
fix(es/helpers): Add missing helpers

sschen86:
when i use @swc/core to build my code and set tsconfig with `"importHelpers": true`

it has error `Missing "./_/_identity" specifier in "@swc/helpers" package`.

I found that the reason for the error is that this file is missing.


kdy1 (approved):
swc-bump:
 - dbg-swc

PR #8845, opened 2024-04-11
fix(es/plugin): Create `tokio` runtime only if necessary

h-a-n-a:
**Description:**

Create tokio runtime if it's not in an async context.

**BREAKING CHANGE:**

This maybe related to this issue, since creating new tokio runtime consumes a lot of system resource.
https://github.com/web-infra-dev/rspack/issues/6209


kdy1 (approved):
swc-bump:
 - swc_core
 - swc

PR #8821, opened 2024-04-08
build(cargo): Update rustc to `nightly-2024-04-03`

kdy1:
**Description:**

**Related issue (if exists):**

PR #8846, opened 2024-04-11
Update rust-toolchain version, and fix compilation error

bgw:
**Description:**

Updating to a newer toolchain gives the following error, due to a change in lint behavior:

```
error: the item `swc_ecma_codegen` is imported redundantly
   --> crates/swc/src/lib.rs:135:24
    |
135 | use swc_ecma_codegen::{self, Node};
    |                        ^^^^ the item `swc_ecma_codegen` is already defined by prelude
    |
note: the lint level is defined here
   --> crates/swc/src/lib.rs:107:9
    |
107 | #![deny(unused)]
    |         ^^^^^^
    = note: `#[deny(unused_imports)]` implied by `#[deny(unused)]`
```

N/A, I discovered this while trying to patch a newer version of swc into next.js which is using a newer toolchain.


kdy1:
I think it's quite similar to https://github.com/swc-project/swc/pull/8821

Issue #8813, opened 2024-04-04
minifier: invalid inlining case expression in switch statement with fall-through case


hyp3rflow:
### Describe the bug

According to the [CaseBlockEvaluation spec](https://tc39.es/ecma262/multipage/ecmascript-language-statements-and-declarations.html#sec-runtime-semantics-caseblockevaluation), case expressions in case clauses placed after `default` case should not be evaluated when there is any match case before `default` case.

### Input code

```typescript
const k = (() => {
    const x = 'asdf';
    switch (x) {
        case x:
        default:
        case console.log("hi"):
    }
})();
```

### Playground link (or link to the minimal reproduction)

[...]

PR #8854, opened 2024-04-12
fix(es/minifier): Handle switch cases

magic-akari:
**Related issue:**

- Closes: #8813

### DIFF
--- a/crates/swc_ecma_minifier/tests/terser/compress/functions/issue_2620_4/output.js
+++ b/crates/swc_ecma_minifier/tests/terser/compress/functions/issue_2620_4/output.js
@@ -1,2 +1,10 @@
 var c = "FAIL";
+(function(a, NaN) {
+    (function() {
+        switch(a){
+            case a:
+            case void (c = "PASS"):
+        }
+    })();
+})(0 / 0);
 console.log(c);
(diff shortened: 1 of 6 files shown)


crates/swc_ecma_minifier/tests/terser/compress/functions/issue_2620_4/output.js
```diff
@@ -1,4 +1,2 @@
 var c = "FAIL";
-var a, NaN;
-a = 0, c = "PASS";
```
magic-akari, line 3:
```js
var c = "FAIL";
(function () {
    function f(a, NaN) {
        function g() {
            switch (a) {
                case a:
                    break;
                case ((c = "PASS"), NaN):
                    break;
            }
        }
        g();
    }
    f(0 / 0);
})();
console.log(c);
```
This case failed for another reason.
It matched the first branch, but `NaN !== NaN`, so this is a bug.

crates/swc_ecma_transforms_optimization/src/simplify/branch/tests.rs
```diff
@@ -867,19 +867,16 @@ fn test_optimize_switch_with_default_case() {
...
-            "    break;",
-            "  default:",
-            "    if (a) { break; }",
-            "    bar();",
-            "}",
-        ),
-        "x; foo()",
-    );
+    test_same(concat!(
+        "switch (x) {",
```
magic-akari, line 871: This is a bugfix. 
Since we cannot determine whether x is `NaN` or not.

kdy1 (approved):
swc-bump:
 - swc_core

Issue #8837, opened 2024-04-09
duplicated identifier with `mangle.safari10`


taozhou-glean:
### Describe the bug

with `safari10` enabled, the generated code is throwing `duplicated identifier` errors

https://github.com/swc-project/swc/discussions/3373#discussioncomment-9064386

### Input code

```typescript
// code taken from https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/c9311d91794b7f4b3ae12b82754b1236746e1eb8/lib/msal-browser/src/broker/nativeBroker/NativeMessageHandler.ts#L106
export class NativeMessageHandler {
    static async createProvider(
        logger: Logger,
        handshakeTimeoutMs: number,
        performanceClient: IPerformanceClient
    ): Promise<NativeMessageHandler> {
        logger.trace("NativeMessageHandler - createProvider called.");
        try {
            const preferredProvider = new NativeMessageHandler(
                logger,
                handshakeTimeoutMs,
                performanceClient,
                NativeConstants.PREFERRED_EXTENSION_ID
            );
            await preferredProvider.sendHandshakeRequest();
            return preferredProvider;
        } catch (e) {
            // If preferred extension fails for whatever reason, fallback to using any installed extension
            const backupProvider = new NativeMessageHandler(
                logger,
                handshakeTimeoutMs,
                performanceClient
            );
            await backupProvider.sendHandshakeRequest();
            return backupProvider;
        }
    }
}
```

### Config

```json
{
  "jsc": {
    "externalHelpers": true,
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "transform": {
      "legacyDecorator": true,
      "decoratorMetadata": true
    },
    "loose": true,
    "minify": {
      "mangle": {
        "toplevel": false,
        "keep_classnames": false,
        "keep_fnames": false,
        "keep_private_props": false,
        "ie8": false,
        "safari10": true
      },
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
[...]

PR #8857, opened 2024-04-13
refactor(es/minifier): Remove `mangle.safari10`

Auspicus:
When the `mangle.safari10` option is set to true identifiers are re-used in the `try/catch` block which results in a parse error.

**Description:**

Using the `mangle.safari10` option causes identifiers to be re-used in a `try/catch` block.

**BREAKING CHANGE:**

https://github.com/swc-project/swc/issues/8837


crates/swc_ecma_transforms_base/src/rename/mod.rs
```diff
@@ -130,7 +130,7 @@ where
     {
         let mut scope = {
             let mut v = Analyzer {
-                safari_10: self.config.safari_10,
+                safari_10: false,
```
Auspicus, line 133: Can use `Default::default()` here?

kdy1:
You can do `UPDATE=1 cargo test` from `./crates/swc_ecma_minifier` to update test snapshot files

Auspicus:
@kdy1 I think we'll have divergence with the terser output for the `broken_safari_catch_scope` test if we remove this option. Is that OK?

kdy1:
Yeap, it's fine

kdy1 (approved):
swc-bump:
 - swc_ecma_transforms_base
 - swc_ecma_minifier
 - swc_core

kdy1:
Can you run `UPDATE=1 cargo test --test projects --test tsc` from `./crates/swc`?

Issue #8858, opened 2024-04-14
swc does not prioritize jsc.paths by matching prefix length like tsc specifies


yehonatanz:
### Describe the bug

Seems like `swc` does not prioritize `jsc.paths` matches by length but instead chooses the first (re: arbitrary, since JSON objects are unsorted) that matches.
This is opposed to the behavior specified by `tsc` [here](https://www.typescriptlang.org/docs/handbook/modules/reference.html#wildcard-substitutions):
> When multiple patterns match a module specifier, the pattern with the longest matching prefix before any * token is used

As far as I understand (zero rust experience, so please be forgiving), this can be rather easily fixed by sorting `paths` by index of `*` in [tsc.rs](https://github.com/swc-project/swc/blob/main/crates/swc_ecma_loader/src/resolvers/tsc.rs#L61-L93).

I'm open to implementing it myself if that is indeed classified as bug that needs fixing.

### Input code

```typescript
import { hello } from '@app/helpers/hello';

hello();
```

### Config

```json
{
  "module": {
    "type": "commonjs",
    "strictMode": true,
    "noInterop": false
  },
  "jsc": {
    "paths": {
      "@app/*": [
        "packages/*/src"
      ],
      "@app/helpers/*": [
        "packages/helpers/src/*"
      ]
    },
    "baseUrl": ".",
    "parser": {
      "syntax": "typescript"
    }
  }
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.4.13&code=H4sIAAAAAAAAA8vMLcgvKlGoVshIzcnJV6hVSCvKz1VQd0gsKNAHChWkFhXrg6XUrbm4wAwNTWsAh2tUWDUAAAA%3D&config=H4sIAAAAAAAAA3WQvQ7CMAyE9z5FlbGq6M7EysDIhBhCamghbSzblaiqvDtJm%2FIzMEX3nX3yZcryXHWuHiyobT4FFbSMGJUyrutcf2dVLpyFWiMHV0dXaIDEe7fvBchhwFdtGQL20VN3Np9Y1NLwWwaw04hVEcgpkXnIPPQNuCoqJqOScS5%2FdhqwCMR%2Fd1c%2FJISZNWN%2Bfbr5ohmOZGPNzdoPNTHQ94U89qKfcSj%2BCRtqUZY8HztmPnsBwh9FpEEBAAA%3D

### Expected behavior

[...]

PR #8875, opened 2024-04-18
fix(tsc-resolver): Prioritze jsc.paths by length in tsc resolver

yehonatanz:
**Description:**
See #8858 

**BREAKING CHANGE:**
`jsc.paths` are now prioritized differently to align with `tsc`. The order is:
1. Exact matches
2. Wildcard pattern matches, sorted by length of prefix before wildcard
3. 

**Related issue (if exists):** #8858


kdy1 (approved):
swc-bump:
 - swc_core
 - swc_ecma_loader

PR #8883, opened 2024-04-22
fix(es/minifier): don't reserve identifiers in expressions or patterns

### LINKED ISSUE #8622: variable is declared multiple times when variable is reserved and there are 2 passes and target is >=es2015
### Describe the bug

When `jsc.minify.compress.passes > 1` and `jsx.minify.mangle.reserved` includes a reserved name - SWC can produce code that semantically invalid with the same variable being definied multiple times.

### Input code

```typescript
export function foo(cond) {
  let reserved = 1;
  if (cond) {
    reserved = 2;
  }
  return [reserved, bar(cond)];
}

function bar(cond) {
  let reserved = 1;
  if (cond) {
    reserved = 2;
  }
  return reserved;
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2015",
    "loose": false,
    "minify": {
      "compress": {
        "passes": 2
      },
      "mangle": {
        "reserved": ["reserved"]
      }
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

https://play.swc.rs/?version=1.4.0&code=H4sIAAAAAAAAA0utKMgvKlFIK81LLsnMz1NIy8%2FXSM7PS9FUqOZSUMhJLVEoSi1OLSpLTVGwVTC0BoplpikgqVBAljcCyddygcRKSovyFKJhcjoKSYlFEF2x1ly1XFxw%2B%2BDi2OwzJmCfCZp9MCmQFQByNT4O2QAAAA%3D%3D&config=H4sIAAAAAAAAA1WOQQ6DIBBF956CzLqL1qRd9A49QdMFwdFgQMgMNhrj3YsoajeEeX%2F%2Bnz8VQkDLCp5iit84eEmMtM%2BR8NgFOUQCqKxkRdoHuGS15UWqpWFMaF4VCJIaDMnF5fV23xxgnGPMjo1Z3el6PN9UznpC5hNL3ZhxYeXG5r2GlV1j8H89BiB9sYr0fQyf7C3ym1LAuqo%2FJUAYPa71H3As5aZ7fdD8ys5APRbzD1wt6BJUAQAA

### SWC Info output

_No response_

### Expected behavior

SWC produces valid code

### Actual behavior

SWC produces broken code:
```js
export function foo(e) {
    let reserved, reserved = 1;
    return e && (reserved = 2), [
        reserved,
        (reserved = 1, e && (reserved = 2), reserved)
    ];
}
```

### Version

1.4.0

### Additional context

Note: When targetting es5 the output uses `var` so it does not crash in a browser.

For comparison the 1 pass code:
```js
export function foo(e) {
    let reserved = 1;
    return e && (reserved = 2), [
        reserved,
        function(e) {
            let reserved = 1;
            return e && (reserved = 2), reserved;
        }(e)
    ];
}
```

1 pass code with no reserved names:
```js
export function foo(t) {
    let n = 1;
    return t && (n = 2), [
        n,
        function(t) {
            let n = 1;
            return t && (n = 2), n;
        }(t)
    ];
}
```

2 pass code with no reserved names:
```js
export function foo(o) {
    let t, e = 1;
    return o && (e = 2), [
        e,
        (t = 1, o && (t = 2), t)
    ];
}
```

Auspicus:
**Description:**

https://github.com/swc-project/swc/issues/3424 added support for the `reserved` mangle option which prevents identifiers from being mangled when listed in the array. This is fine in most cases but can cause issues when the compressor brings two reserved variables into the same scope. When this occurs the generated code results in a parsing error due to the duplicate identifier.

This PR adds a test to cover this case and a proposed fix by ignoring reserved identifiers for patterns and expressions.

**BREAKING CHANGE:**

fixes #8622

### DIFF
--- a/crates/swc_ecma_minifier/src/pass/mangle_names/preserver.rs
+++ b/crates/swc_ecma_minifier/src/pass/mangle_names/preserver.rs
@@ -101,7 +101,7 @@ impl Visit for Preserver {
         n.visit_children_with(self);
 
         if let Expr::Ident(i) = n {
-            if self.should_preserve || self.is_reserved(i) {
+            if self.should_preserve {
                 self.preserved.insert(i.to_id());
             }
         }
@@ -141,7 +141,7 @@ impl Visit for Preserver {
         n.visit_children_with(self);
 
         if let Pat::Ident(i) = n {
-            if self.should_preserve || self.is_reserved(&i.id) {
+            if self.should_preserve {
                 self.preserved.insert(i.to_id());
             }
         }
--- a/crates/swc_ecma_minifier/tests/fixture/issues/8622/config.json
+++ b/crates/swc_ecma_minifier/tests/fixture/issues/8622/config.json
@@ -1,0 +1,4 @@
+{
+    "defaults": true,
+    "passes": 2
+}
--- a/crates/swc_ecma_minifier/tests/fixture/issues/8622/input.js
+++ b/crates/swc_ecma_minifier/tests/fixture/issues/8622/input.js
@@ -1,0 +1,15 @@
+export function foo(cond) {
+    let reserved = 1;
+    if (cond) {
+        reserved = 2;
+    }
+    return [reserved, bar(cond)];
+}
+
+function bar(cond) {
+    let reserved = 1;
+    if (cond) {
+        reserved = 2;
+    }
+    return reserved;
+}
--- a/crates/swc_ecma_minifier/tests/fixture/issues/8622/mangle.json
+++ b/crates/swc_ecma_minifier/tests/fixture/issues/8622/mangle.json
@@ -1,0 +1,3 @@
+{
+    "reserved": ["reserved"]
+}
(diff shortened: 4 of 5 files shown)


kdy1 (approved):
I think this change is good. Can you undraft it?

Austaras (changes requested):
I don't think this is the right solution. Simply ignore `reserved` contradicts to terser behaviour and documentation. Maybe when checking `may_add_ident`, we should check if it is reserved too?

Auspicus:
@Austaras I agree. This doesn't match Terser's behavior here and am open to other directions. Terser doesn't compress this down to one function so it's a bit hard to compare the behavior across the two. There's a valid use case for having variable declarations at the top level be reserved and unmangled so that they can be accessed externally which I think this change might prevent although AFAICT there isn't a test for that.

kdy1:
Closing in favor of https://github.com/swc-project/swc/pull/8939

Issue #8872, opened 2024-04-17
Regression in `v1.4.15`: `ReferenceError: resource is not defined` in strict mode


jeengbe:
### Describe the bug

Updating from `.14` -> `.15`, code with `using` fails in strict mode with `ReferenceError: resource is not defined`.

This is the generated `_using_ctx()`:

```js
        function _using_ctx() {
            var _disposeSuppressedError = typeof SuppressedError === "function" ? SuppressedError : function(error, suppressed) {
                var err = new Error();
                err.name = "SuppressedError";
                err.suppressed = suppressed;
                err.error = error;
                return err;
            }
              , empty = {}
              , stack = [];
            function using(isAwait, value) {
                if (value != null) {
                    if (Object(value) !== value) {
                        throw new TypeError("using declarations can only be used with objects, functions, null, or undefined.");
                    }
                    if (isAwait) {
                        var dispose = value[Symbol.asyncDispose || Symbol.for("Symbol.asyncDispose")];
                    }
                    if (dispose == null) {
                        dispose = value[Symbol.dispose || Symbol.for("Symbol.dispose")];
                    }
                    if (typeof dispose !== "function") {
                        throw new TypeError(`Property [Symbol.dispose] is not a function.`);
                    }
                    stack.push({
                        v: value,
                        d: dispose,
                        a: isAwait
                    });
                } else if (isAwait) {
                    stack.push({
                        d: value,
                        a: isAwait
                    });
                }
                return value;
            }
            return {
                e: empty,
                u: using.bind(null, false),
                a: using.bind(null, true),
                d: function() {
                    var error = this.e;
                    function next() {
                        while (resource = stack.pop()) {
                        //     ^^^^^ fails here
                            try {
                                var resource1, disposalResult = resource1.d && resource1.d.call(resource1.v);
                                if (resource1.a) {
                                    return Promise.resolve(disposalResult).then(next, err);
                                }
                            } catch (e) {
                                return err(e);
                            }
                        }
                        if (error !== empty)
                            throw error;
                    }
                    function err(e) {
                        error = error !== empty ? new _disposeSuppressedError(error,e) : e;
                        return next();
                    }
                    return next();
                }
            };
```

The test is run in strict mode.

### Config

```
      '@swc/jest',
      {
        jsc: {
          parser: {
            syntax: 'typescript',
            decorators: true,
          },
        },
      },
```

### Playground link (or link to the minimal reproduction)

[...]

PR #8891, opened 2024-04-24
Remove unused export from _using_ctx.js

anenger:
**Description:**

Fixes stray export in `_using_ctx.js` inside of `packages/helpers`. This was causing build failures in my project, and I believe it was an errant export from a previous version of the helpers package.

#8872


kdy1 (approved):
swc-bump:
 - dbg-swc

Issue #8515, opened 2024-01-18
SWC Decorators not working for `var X = @addThing class {}`


branks:
### Describe the bug

We are running our code through esbuild first to generate a bundle, and then running it through swc to add the SystemJS layer on top. We have noticed that esbuild converts some classes with decorators from
`@addX class C { ... }` to `var C = @addX class { ... }` (because of https://github.com/evanw/esbuild/issues/478)

This is all valid JS and works fine. However, SWC seems to fall over when then trying to handle the decorators 

### Input code

```typescript
import { addX, addY } from 'someLib'

var C = @addX @addY class extends Component {
}

@addX @addY
class OtherClass extends Component {

}
```

### Config

```json
{
    "sourceMaps": true,
    "inlineSourcesContent": true,
    "jsc": {
      "target": "es2016",
      "loose": false,
      "externalHelpers": false,
      "keepClassNames": true,
      "parser": {
        "syntax": "ecmascript",
        "jsx": true,
        "decorators": true,
        "dynamicImport": true
      },
      "transform": {
        "legacyDecorator": true,
        "useDefineForClassFields": true,
        "react": {
          "refresh": false,
          "runtime": "automatic"
        }
      }
    },
    "module": {
       "type": "systemjs"
    }
  }
```

### Playground link (or link to the minimal reproduction)

[...]

PR #8905, opened 2024-04-29
fix(es/decorators): Handle ref of decorated class

magic-akari:
**Related issue:**

Follow up #8515 
Ts Decorator does not always return a class; we need to give it the correct reference.


kdy1 (approved):
swc-bump:
 - swc_core

PR #8907, opened 2024-05-01
fix(es/minifier): Preserve flags while dropping elements of `SeqExpr`

devongovett:
In code like `/*#__PURE__*/ (0, forwardRef)(Foo);`, the `dead_branch_remover` pass will simplify this to `forwardRef(Foo)` and remove the `PURE` annotation which might be used in later passes. I added a test to show this, and commented out the line that causes it. However I'm not quite sure how to fix it.

I guess the comment is attached to the original sequence expression which was removed. So we need to move the comment to the resulting expression somehow, or maybe change the span of the expression to match the old one.

I tried passing in the comments to this pass so I could move it, but it was a bit hard because of the `Parallel` implementation that would require the comments to be `Send + Sync`.

Any suggestions on the best approach here?


crates/swc_ecma_transforms_optimization/src/simplify/branch/mod.rs
```diff
@@ -108,7 +108,10 @@ impl VisitMut for Remover {
                 if s.exprs.is_empty() {
                     *e = Expr::dummy();
                 } else if s.exprs.len() == 1 {
-                    *e = *s.exprs.pop().unwrap();
+                    // TODO: preserve PURE comment attached to original
+                    // sequence expression.
+
+                    // *e = *s.exprs.pop().unwrap();
```
kdy1, line 114: You can

1. assign the span of the node with `PURE` comment to the emitted node

2. Accept `C: Comments` or `&dyn Comments` and call `move_leading`

kdy1:
I'll take on this

Issue #8893, opened 2024-04-25
Keep `assert` and `with` keywords as is


petamoriken:
### Describe the bug

V8 will remove support for the `assert` keywords, so Node.js and Chrome will follow suit.

> Hey, quick update on this. Node.js is planning to remove support for assert in v22 (which will be released in April) and Chrome in v126 (which will be released in May).

https://github.com/denoland/deno/issues/17944#issuecomment-2027005952

I'm trying to implement this in Deno as well (https://github.com/denoland/deno/pull/23541), but it doesn't work because swc commonizes the `assert` and `with` keywords. Please add a new option to pass those keywords as is to V8.

### Input code

```typescript
import foo from "./foo.json" with { type: "json" };
import bar from "./bar.json" assert { type: "json" };
console.log(foo);
console.log(bar);
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2022",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    },
    "experimental": {
      "keepImportAttributes": true
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.5.0&code=H4sIAAAAAAAAA8vMLcgvKlFIy89XSCvKz1VQ0tMHsvWyivPzlBTKM0syFKoVSioLUq0UlCBitdZcmRA9SYlFcD1ANlRPYnFxKlASU1dyfl5xfk6qXk5%2BugbQCk1UEaABmtYAe1o9F40AAAA%3D&config=H4sIAAAAAAAAA21QOw6DMAzdOQXy3KFi6NCtY4ceIqUGpU1IZBsJhLg7CSSUSt3i9%2FNzpqIs4c01XMspPMPgFTHSPgeEx07UEBCQ0SPXpL3AKbPCkWqUYVyheWNAFLUo0YVcnasqOcA4x5gdCbO608143Fk76wmZf4VRqrrW4N%2BNOHgkbTG0NcesD6K%2FW%2B9IbiKkn71gzBXqk79IGWDdq1%2FD01%2FEe7cLLvAV5bJ7MdD8yM41dV4AwZk7Q1cBAAA%3D

### Expected behavior

```js
import foo from "./foo.json" with {
    type: "json"
};
import bar from "./bar.json" assert {
    type: "json"
};
[...]

PR #8913, opened 2024-05-03
feat(es/parser): Add `disallowAssertKeywords` option

petamoriken:
**Description:**

V8 will remove support for the `assert` keywords, so Node.js and Chrome will follow suit.

> Hey, quick update on this. Node.js is planning to remove support for `assert` in v22 (which will be released in April) and Chrome in v126 (which will be released in May).

https://github.com/denoland/deno/issues/17944#issuecomment-2027005952

Disallow the `assert` keywords in Deno by adding the `disallowAssertKeywords` option in this PR.

**BREAKING CHANGE:**

N/A

https://github.com/swc-project/swc/issues/8893#issuecomment-2080369150

### DIFF
--- a/crates/swc_ecma_parser/src/lib.rs
+++ b/crates/swc_ecma_parser/src/lib.rs
@@ -187,6 +187,20 @@ impl Syntax {
         }
     }
 
+    pub fn disallow_assert_keywords(self) -> bool {
+        match self {
+            Syntax::Es(EsConfig {
+                disallow_assert_keywords,
+                ..
+            }) => disallow_assert_keywords,
+            #[cfg(feature = "typescript")]
+            Syntax::Typescript(TsConfig {
+                disallow_assert_keywords,
+                ..
+            }) => disallow_assert_keywords,
+        }
+    }
+
     /// Should we parse jsx?
     pub fn jsx(self) -> bool {
         match self {
@@ -315,6 +329,9 @@ pub struct TsConfig {
     #[serde(default)]
     pub decorators: bool,
 
+    #[serde(default)]
+    pub disallow_assert_keywords: bool,
+
     /// `.d.ts`
     #[serde(skip, default)]
     pub dts: bool,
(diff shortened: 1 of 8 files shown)


petamoriken:
@kdy1 Where should I write the test codes?

kdy1:
You can add a test like https://github.com/swc-project/swc/blob/c1115a2ebbca4fdfb2d4dc589c2d82b878cf9f88/crates/swc_ecma_parser/src/parser/tests.rs#L321-L345

kdy1 (commented):
swc-bump:
 - swc_ecma_parser --breaking

crates/swc_ecma_parser/src/lib.rs
```diff
@@ -187,6 +187,20 @@ impl Syntax {
...
 
+    pub fn disallow_assert_keywords(self) -> bool {
+        match self {
+            Syntax::Es(EsConfig {
+                disallow_assert_keywords,
+                ..
+            }) => disallow_assert_keywords,
+            #[cfg(feature = "typescript")]
+            Syntax::Typescript(TsConfig {
+                disallow_assert_keywords,
```
kdy1, line 198: Please remove this

crates/swc_ecma_parser/src/lib.rs
```diff
@@ -315,6 +329,9 @@ pub struct TsConfig {
     #[serde(default)]
     pub decorators: bool,
 
+    #[serde(default)]
+    pub disallow_assert_keywords: bool,
```
kdy1, line 333: and this

kdy1 (changes requested):

crates/swc_ecma_parser/src/lib.rs
kdy1: We should unconditionally follow the behavior of tsc

petamoriken: Oh, I see. The issue I want to solve with this PR is that it prevents the `assert` keywords from being used in TypeScript, so I'll try to confirm if tsc can handle it.

petamoriken:
Since tsc has not decided on a direction, this pull request is closed for the time being.

Issue #3085, opened 2021-12-21
`.browserlistsrc` should be considered even if `.swcrc` exists


kdy1:
Currently, we don't check for `.browserlistsrc`, but it can be confusing.

g-plane:
Which config file should be preferred?

kdy1:
Oh... Maybe we should emit a warning if there's both `jsc.target`/`env` with `.browserslisttc`

g-plane:
Looking for `.browserslistrc` file can be complicated, how do we check it?

kdy1:
@g-plane 

Yeah, I also think it's a complicated task.
Does `browserslist-rs` supports checking `.browserlistsrc` or entry in the `packge.json`?


Maybe a command like `swc doctor` would be better.

g-plane:
Yes, but it didn't expose APIs about reading configuration files.

kdy1:
Is there a way to distinguish between a default value and configured value?
I think it will be enough.

g-plane:
What does "default value" and "configured value" mean?

kdy1:
If user specified anything related to `browserslist`, it's configured.

g-plane:
Maybe it can't.

There's a kind of query: `defaults`. If users don't specifying anything, it will use `defaults` as "default value". However, users can also configure it using `defaults` in config file manually.

kdy1:
Oh, I see. Then we can check if `env` and `jsc.target` are specified at the same time.

g-plane:
https://github.com/babel/babel/blob/cb3ebde8ceed47f4b6783c16137cc9ddc78998b7/packages/babel-helper-compilation-targets/src/index.ts#L194

I think we may need to expose an API about loading config.

zalishchuk:
@kdy1 @g-plane Hey! Are you sure `browserslist` config loading works fine now? The issue I have is that `swc-loader` doesn't load the config from `package.json`, so I have to do it manually, `.browserslistrc` doesn't seem to work either.

```js
// webpack swc-loader options
{
  env: {
    // FIXME: swc-loader doesn't load browserslist config from package.json for some reason
    targets: require('browserslist').loadConfig({
      path: paths.appPath,
      env: nodeEnv,
    }),
    mode: 'entry',
    coreJs: '3.22',
  },
  jsc: {
    target: 'es5',
    parser: {
      syntax: 'typescript',
      dynamicImport: true,
      tsx: true,
    },
    experimental: {
[...]

PR #8921, opened 2024-05-06
fix(es/preset-env): Consider `browserslist` config if `env.target` is not configured

### LINKED ISSUE #3365: Browserslist not being applied from .browserslistrc or package.json
### Describe the bug

Documentation at https://swc.rs/docs/configuration/supported-browsers#browserslist suggests that if `targets` property is omitted in `.swcrc`, the browserslist should be read from `.browserslistrc` or package.json `browserslist` key.

This does not seem to be the case, only way to seemingly enable different targets is to specify the "targets" key in ".swcrc".

Example file at `test.ts`:

```
const myTest = 'foobar'

const someFunc = (test: string) => {
  console.log(test)
}

const myBar = someFunc(myTest);
```

and `.browserslistrc`:

```
chrome 95
```

### Input code

```typescript
yarn run swc ./test.ts -o ./test.js
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true
    }
  },
  "env": {
    "debug": true,
  }
}
```

### Playground link

_No response_

### Expected behavior

Browser targets from `.browserslist` file applied. Debug output should match the one when `"targets"` is specified in `.swcrc`.

Debug output with `"targets": "chrome 95"` in `.swcrc`:

```
Targets: BrowserData { chrome: Some(Version { major: 95, minor: 0, patch: 0 }), and_chr: None, and_ff: None, op_mob: None, ie: None, edge: None, firefox: None, safari: None, node: None, ios: None, samsung: None, opera: None, android: None, electron: None, phantom: None, opera_mobile: None, rhino: None }
```

### Actual behavior

Targets from browserslist not being applied. A defaults (?) applied instead as per debug output:

```
Targets: BrowserData { chrome: Some(Version { major: 94, minor: 0, patch: 0 }), and_chr: None, and_ff: None, op_mob: None, ie: Some(Version { major: 11, minor: 0, patch: 0 }), edge: Some(Version { major: 95, minor: 0, patch: 0 }), firefox: Some(Version { major: 78, minor: 0, patch: 0 }), safari: Some(Version { major: 13, minor: 1, patch: 0 }), node: None, ios: Some(Version { major: 12, minor: 2, patch: 0 }), samsung: Some(Version { major: 14, minor: 0, patch: 0 }), opera: Some(Version { major: 80, minor: 0, patch: 0 }), android: Some(Version { major: 95, minor: 0, patch: 0 }), electron: None, phantom: None, opera_mobile: None, rhino: None }
```

### Version

@swc/cli: 0.1.55 @swc/core: 1.2.133

### Additional context

_No response_

egfx-notifications:
**Description:**

The `env.path` option in `.swcrc` was originally implemented for `browserslist`, but implementation was (unintentionally?) removed during the migration to `browserslist-rs` (#2845).
Searching the current directory for browserslist configuration was also dropped in the process.

This PR reimplements searching for browserslist configuration using the lookup functionality provided by `browserslist-rs`

**BREAKING CHANGE:**

This may be considered a breaking change since the previous (broken) behavior for unconfigured `env.targets` was to return a `BrowserData` struct with all `Option`s set to `None`. Now the result of a browserslist `defaults` query is returned instead. This probably was the original behavior with `browserslist`, though.
*Update:* Thinking about it again, `targets_to_versions()` is part of `preset_env_base`s public API so we should probably consider this a breaking change anyway.

**Things to consider:**
- only works for non wasm targets as `browserslist-rs` does not provide configuration loading for wasm32 targets (which is probably to be expected that we won't check environment variables or search a filesystem in that case)
- `feat(preset-env): Let browserslist-rs handle default config path` changes the default for an unconfigured `env.path` from an empty string for wasm targets / current dir for other targets to a `None` value. `browserslist-rs` already looks up the current dir if `path` is `None` and it felt better to let it handle this case itself
  The commit which migrated from `browserslist` to `browserslist-rs` mentioned fixing the `default_path`, but does not mention what was broken and I also did not find any other usage of the `path` property. If the previous implementation was intentional, just ignore that last commit in the PR, the first two sufficiently address the bug.

**Related issue:**
Fixes  #3365
Implements #3085


egfx-notifications:
It seems some of the tests fail due to them expecting a specific prerecorded results, but this result has now changed due to the previous default without a target being `BrowserData` with only `None` values and now it is `BrowserData` according to browserslist-rs `defaults` query, so some stuff that browser now understand natively is not transpiled.

Seems forcing a browserslist query when target is none is not the best approach. I won't try any fixes before hearing your feedback on how to handle this. Should we make the `env.path` option mandatory if a browserslist config lookup is desired?

crates/preset_env_base/src/query.rs
```diff
@@ -96,8 +96,31 @@ impl Query {
...
+                .to_owned();
+            if let Some(path) = path {
+                browserslist_opts.path(
+                    path.clone()
+                        .into_os_string()
+                        .to_str()
+                        .unwrap_or_else(|| panic!("Invalid path \"{:?}\"", path)),
+                );
+            }
+            let distribs = browserslist::execute(&browserslist_opts)
```
kdy1, line 115: Does this work with configs that execute javascript? cc @g-plane

g-plane: No, it doesn't execute any code.

kdy1 (commented):
Can you rebase?

jodaka:
Nice to see some action here! Thanks @egfx-notifications

egfx-notifications:
I think I addressed all of the changes between my original commits and now. Unfortunately I can't run `cargo test` locally right now as my GLIBC version is too old. Let's see what CI says. I may set up a newer Ubuntu VM tomorrow if necessary.

kdy1:
No worries, I‘ll update the tests if needed

.changeset/tidy-pots-report.md
```diff
@@ -0,0 +1,5 @@
+---
+preset_env_base: major
```
kdy1, line 2: As this is a breaking change of swc_core, I'll merge it later.

crates/preset_env_base/src/query.rs
```diff
@@ -96,8 +96,35 @@ impl Query {
...
+                mobile_to_desktop: true,
+                ignore_unknown_versions: true,
+                ..Default::default()
+            };
+            if let Some(path) = path {
+                browserslist_opts.path = path
+                    .clone()
+                    .into_os_string()
+                    .into_string()
+                    .unwrap_or_else(|_| panic!("Invalid path \"{:?}\"", path))
```
kdy1, line 116: I think we should not `panic!()` here. Can you return an instance of `anyhow::Error` instead?

crates/preset_env_base/src/query.rs
```diff
@@ -96,8 +96,35 @@ impl Query {
...
+                    .into();
+            }
+            let distribs = browserslist::execute(&browserslist_opts)
+                .with_context(|| "failed to resolve browserslist query from browserslist config")?;
+
+            let versions =
+                BrowserData::parse_versions(distribs).expect("failed to parse browser version");
```
kdy1, line 123: Same here. We don't have to panic.

kdy1 (changes requested):

kdy1 (dismissed):
Thank you, and sorry for a (very) late review!

kdy1 (commented):
I missed it while reviewing, but while trying to update tests to merge this PR, I found that this PR makes the `swc` crate ignore `jsc.target`.
Could you fix it?

You can run `UPDATE=1 cargo test --test projects --test tsc --test error_msg` from `./crates/swc` to update test snapshots.

egfx-notifications:
> I missed it while reviewing, but while trying to update tests to merge this PR, I found that this PR makes the `swc` crate ignore `jsc.target`. Could you fix it?

So, I looked into this and think I need a little more context. The only test failures I could find in the fixture tests were tests which explicitly **didn't** specify `jsc.target`. Those test cases have neither `env.target` nor `jsc.target` defined and as I mentioned the default if specifying no target is now the result of a browserslist `defaults` query, thus making this PR a breaking change. So those are test failures that were expected to happen until the fixtures are updated.
Note that the shopify tests 003-005 also fall into this "defaults" change since their `.swcrc` is wrong, they specify 
```json
{
    "jsc": {
        "parser": {
            "syntax": "ecmascript",
            "jsx": true
        }
    },
    "module": {
        "type": "commonjs"
    },
    "env": {
        "chrome": "50"
    }
}
```
when it should have been
```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": true
    }
  },
  "module": {
    "type": "commonjs"
  },
  "env": {
    "targets": {
      "chrome": "50"
    }
  }
}
```
So those fixtures weren't using that targeted chrome version from the start.

As far as I can tell `jsc.target` should still work as intended if specified.

kdy1:
Oh, I think then it's fine and we should specify `jsc.target: es5` for those tests I think

egfx-notifications:
So, since `jsc.target` and `env` can't be used together and I think the `env` configurations in those test cases may be relevant for the case, I tried another route to solve this. I put `.browserslistrc` files without any content into the crate directories of the failing tests. This way it is picked up by autodiscovery if the test doesn't specify `jsc.target` or `env.targets`, but an empty query results in applying all possible transformations and thus emulates the previous default. This also means we don't have to regenerate all fixtures everytime `browserslist-rs` updates the result of the `defaults` query.  
Let me know if this is a suitable solution.

I also added a new test to check if `env.path` actually works.

kdy1:
We don’t apply everything as default

egfx-notifications:
> We don’t apply everything as default

Ok, maybe this was phrased wrong. I mean that with the empty query in the autodiscovered files we take the same code path as previously without autodiscovery. So every transformation that was applied to the test cases before the PR without explicit configuration would also be applied now.

kdy1 (approved):
Thank you! I'll merge it once https://github.com/swc-project/swc/actions/runs/14299826346 is done!

kdy1 (commented):
~I'll invert this PR with https://github.com/swc-project/swc/pull/10351~

Sorry it was already opt-in.

Issue #8789, opened 2024-03-28
SWC may generate invalid source map when inputSourceMap is provided


noyobo:
### Describe the bug

https://stackblitz.com/~/github.com/noyobo/swc-issues
```bash
$ npm run build 

[Error: failed to read input source map from user-provided sourcemap

Caused by:
    bad reference to source #299] {
  code: 'GenericFailure'
}
```

### Config

```js
import {transform} from '@swc/core';
import fs from "node:fs";

const fileCode = fs.readFileSync("./a.js", "utf-8");
const fileMap = fs.readFileSync("./a.js.map", "utf-8");
// valid check https://evanw.github.io/source-map-visualization/


// transformed successfully
const result = await transform(fileCode, {
  filename: 'a.js',
  sourceMaps: true,
  inputSourceMap: fileMap,
  jsc: {target: 'es5'},
  module: {type: "commonjs", strictMode: false},
})

// fs.writeFileSync("./a.transformed.js", result.code);
// fs.writeFileSync("./a.transformed.js.map", result.map);

// Using the result again to transform throws an error
// and verification failed in https://evanw.github.io/source-map-visualization/
await transform(result.code, {
  filename: "a.js",
  sourceMaps: true,
  inputSourceMap: result.map,
  module: {type: "commonjs", strictMode: false},
}).catch((e) => {
  /**
   * [Error: failed to read input source map from user-provided sourcemap
   *
   * Caused by:
   *     bad reference to source #299] {
   *   code: 'GenericFailure'
   * }
   */
  console.log(e);
});

```

### Playground link (or link to the minimal reproduction)

https://stackblitz.com/~/github.com/noyobo/swc-issues

### Expected behavior

The transformed sourceMap content can be used again

### Version

1.4.11

noyobo:
@kdy1  Is it fixed in v1.4.12? I tried to use 1.4.12 and the problem still exists。

https://stackblitz.com/~/github.com/noyobo/swc-issues

```
> npx -y swc-info@latest && node ./transform-a.mjs


    Operating System:
        Platform: linux
        Arch: x64
        Machine Type: undefined
        Version: Ubuntu 20.04.0 LTS Tue Apr 09 2024 10:36:06 GMT+0800 (中国标准时间)
        CPU: (3 cores)
            Models: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz

    Binaries:
        Node: 18.18.0
        npm: 10.2.3
        Yarn: 1.22.19
[...]

PR #8951, opened 2024-05-13
fix(es): Allow input source map file to be omitted

kdy1:
**Description:**

Some libraries generate source maps but do not upload them to npm, which causes SWC to fail.

**Related issue:**

 - https://github.com/swc-project/swc/issues/8789#issuecomment-2105055772


kdy1 (commented):
swc-bump:
 - swc_core
 - swc

PR #8955, opened 2024-05-14
refactor(es/ast): Remove unused fields of `TsPropertySignature`

lucacasonato:
**Description:**

TsPropertySignature can not have params, type params, or an initializer.

**BREAKING CHANGE:**

Yes. Any users looking into these fields should stop - they are always empty.


lucacasonato:
I suggest you review the two commits independently. The second one is just test expectation updates caused by the first commit. All code changes are in the first commit.

kdy1 (approved):

kdy1 (approved):
swc-bump:
 - swc_ecma_ast --breaking

PR #8956, opened 2024-05-16
fix(es/minifier): Abort array property inliner if the array is used as a ref

kdy1:
**Description:**

Repro:
 - https://github.com/kdy1/repro-next-46887

**Related issue:**

 - https://github.com/vercel/next.js/issues/46887


kdy1 (commented):
swc-bump:
 - swc_core
 - swc_ecma_minifier
 - swc_ecma_usage_analyzer

PR #8967, opened 2024-05-22
test(es/decorators): Add tests written by evanw

kdy1:
**Description:**

Add https://github.com/evanw/decorator-tests as partially ignored tests so we can work on decorator fixes gradually.


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #8968, opened 2024-05-22
feat(es/codegen): Add `to_code`

kdy1:
**Description:**

This is a utility function for printing an AST node as a string


kdy1 (commented):
swc-bump:
 - swc_core
 - swc_ecma_codegen

PR #8969, opened 2024-05-23
fix(es/codegen): Accept `&impl Node` instead of `impl Node`

kdy1:


kdy1 (commented):
swc-bump:
 - swc_ecma_codegen
 - swc_core

PR #8987, opened 2024-05-28
fix(es/resolver): Fix hoisting of `const` and `let`

kdy1:
**Description:**

 - Repro: https://github.com/kdy1/repro-next-66237

**Related issue:**

 - https://github.com/vercel/next.js/issues/66237


kdy1 (commented):
swc-bump:
 - swc_core
 - swc_ecma_transforms_base

PR #8994, opened 2024-05-29
fix(es/jest): Handle `@jest/globals`

kdy1:
**Description:**

 - We have two `jest` pass. One in `@swc/core` (via _hidden.jest flag), and one in the plugin. This PR fixes only the core one.

However, the linked issue is wrong because the code tries to break the rule of ESM specification. So although this PR closes the issue, the final form is not the issue author wanted.

**Related issue:**

 - Closes https://github.com/swc-project/plugins/issues/310


kdy1 (commented):
swc-bump:
 - swc_ecma_ext_transforms
 - swc_core
 - swc_ecma_ast

magic-akari (approved):

PR #8999, opened 2024-05-30
feat(x): Initialize project

kdy1:
**Description:**

SWC X will be a cross-module optimizer that utilizes an enormous amount of information. I'll experiment with several approach in my mind after merging this.


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #9000, opened 2024-05-30
feat(es/transforms): Add experimental output API for transforms

kdy1:
**Description:**

This interface will provide the way to emit some outputs from transforms. This data will be emitted back to the JS caller.
 


kdy1 (commented):
swc-bump:
 - dbg-swc
 - jsdoc
 - preset_env_base
 - swc_ecma_preset_env
 - swc_transform_common

PR #9002, opened 2024-05-30
fix(ci): Restore disabled CI checks

kdy1:
**Description:**

**BREAKING CHANGE:**


kdy1:
cc @Brooooooklyn `node@20` tests are failing because of the `corepack`. Can you help me fixing the CI? I think I need a way to enable corepack in docker, but not sure about the way to do it.

kdy1 (commented):
swc-bump:
 - dbg-swc

PR #9005, opened 2024-05-31
fix(es/minifier): Preserve special properties even if it's not used

kdy1:
**Related issue:**

 - https://github.com/vercel/next.js/issues/66378


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier
 - swc_core

PR #9014, opened 2024-06-03
chore: switch from `browserslist-rs` to `oxc-browserslist`

cijiugechu:
**Description:**

[oxc-browserslist](https://github.com/oxc-project/oxc-browserslist?tab=readme-ov-file) is a fork of [browserslist-rs](https://github.com/browserslist/browserslist-rs) with improved performance, actively maintained by [oxc-project](https://github.com/oxc-project). The following are the changes made to it after the fork:

> * reduced compilation speed from one minute to a few seconds
> * removed all unnecessary, heavy or slow dependencies: `ahash`, `chrono`, `either`, `indexmap`, `itertools`, `once_cell`, `string_cache`
> * improved some runtime performance, e.g. [improve sort method](https://github.com/oxc-project/oxc-browserslist/pull/28), [precompute versions](https://github.com/oxc-project/oxc-browserslist/pull/10)

**BREAKING CHANGE:**


kdy1:
@g-plane Thoughts?

g-plane:
[oxc-browserslist](https://github.com/oxc-project/oxc-browserslist) generates browserslist data into .rs files and they commit them into Git, while [browserslist-rs](https://github.com/browserslist/browserslist-rs) uses `build.rs` to generate files to `target` directory at compilation time. That's the biggest difference.

kwonoj:
+ would like to know if there's a reason to choose fork instead of upstream, at least wasn't able to see discussion around upstream repo.

g-plane:
If using Browserslist is not a hot path of SWC, this change is subtle.

cijiugechu:
Considering that swc is the upstream repository for many projects, even minor changes can benefit many downstreams.

kwonoj:
That applies if any upstream / downstream packages relies on swc also switch their dependency / transitive dependencies to this forked pkg, isn't it? While the implication might not be measurable, that'll take a longer way than if upstream adapts some improvements. As shared above, would like to understand why fork has to be created.

kdy1:
I think the improvement of the compile time is important difference, but I want to hear about the reason of creating a fork instead of sending a patch to original project

PR #9026, opened 2024-06-07
feat(es/minifier): Support `module: "unknown"`

### LINKED ISSUE #8571: Could `minify()` support `module: "unknown"`?
### Describe the feature

I'm working on migrating a large codebase from Terser to SWC for minification. We have a mix of scripts and modules in our output. Most of the time we can differentiate between them based on their file extension (`.js` vs `.mjs`) and use that to set `minify()`'s `module` option correctly. I'm finding that we have a handful of files with a `.js` extension that are actually modules. Terser seems to process the module files without failing even when `module` is set to `false`. SWC (understandably) fails. 

image

Would it be possible to add an `"unknown"` option to `minify()` similar to `transform()`'s `isModule`? 

Related: 
* https://github.com/swc-project/swc/issues/2541
* https://github.com/swc-project/swc/issues/3493

_No response_

### Additional context

_No response_

kdy1:
**Description:**

**Related issue:**

 - Closes #8571

### DIFF
--- a/bindings/binding_core_wasm/__tests__/simple.js
+++ b/bindings/binding_core_wasm/__tests__/simple.js
@@ -190,7 +190,8 @@ describe("parse", () => {
 describe("minify", () => {
     it("should work", () => {
         const output = swc.minifySync(
-            "const somename = 1; console.log(somename);"
+            "const somename = 1; console.log(somename);",
+            { module: false }
         );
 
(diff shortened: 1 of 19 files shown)


kdy1 (commented):
swc-bump:
 - swc_config
 - swc_core

PR #9033, opened 2024-06-11
fix(es): Make `output` field optional

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_compiler_base
 - swc_core

PR #9036, opened 2024-06-11
perf(es/parser): Optimize lexing of template literals

kdy1:
**Description:**

We don't need to push one char at a time.

### DIFF
--- a/crates/swc_ecma_parser/src/lexer/mod.rs
+++ b/crates/swc_ecma_parser/src/lexer/mod.rs
@@ -1160,6 +1160,7 @@ impl<'a> Lexer<'a> {
 
         let mut cooked = Ok(String::new());
         let mut raw = SmartString::new();
+        let mut raw_slice_start = start;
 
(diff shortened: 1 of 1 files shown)


kdy1 (commented):
swc-bump:
 - swc_core
 - swc_ecma_parser

crates/swc_ecma_parser/src/lexer/mod.rs
```diff
@@ -1223,16 +1247,14 @@ impl<'a> Lexer<'a> {
...
                     cooked.push(c);
                 }
-
                 raw.push(c);
+                raw_slice_start = self.cur_pos();
             } else {
                 self.bump();
 
                 if let Ok(ref mut cooked) = cooked {
                     cooked.push(c);
```
sokra, line 1256: Should `cooked` have the same optimization?

kdy1: I think so. I'll patch it tomorrow.

PR #9037, opened 2024-06-11
perf(es/parser): Optimize lexing of template literals, again

kdy1:
**Description:**

Applies same trick as #9036, but for `cooked`


kdy1 (commented):
swc-bump:
 - swc_ecma_parser
 - swc_core

PR #9038, opened 2024-06-11
fix(es/codegen): Emit named type in mapped types

dsherret:
**Description:**

Fixes the emit of mapped types.

### DIFF
--- a/crates/swc_ecma_codegen/tests/fixture/typescript/mapped_type/input.ts
+++ b/crates/swc_ecma_codegen/tests/fixture/typescript/mapped_type/input.ts
@@ -1,0 +1,27 @@
+type NoTypeAnn = {
+    [value: string];
+};
+type MappedTypeWithNamedType = {
+    [K in T as U]: undefined;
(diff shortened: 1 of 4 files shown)


crates/swc_ecma_codegen/tests/fixture/typescript/mapped_type/input.ts
dsherret, line 1: You should consider switching these tests to use file_test_runner. It doesn't require recompiling when adding a test: https://github.com/denoland/file_test_runner (https://david.deno.dev/posts/file-test-runner/)

kdy1 (approved):
Sounds interesting. Does it support listing tests without running them? IIRC it was required for some reason.

---

swc-bump:
 - swc_core
 - swc_ecma_codegen

dsherret:
Yeah, the collection module allows for that: https://docs.rs/file_test_runner/0.7.2/file_test_runner/collection/index.html

It's two steps—collect and then run.

PR #9032, opened 2024-06-11
fix(es/minifier): Visit RHS while hoisting properties

kdy1:
**Description:**

Handles 

```js
var FRUITS = { MANGO: "mango" }, getMangoLabel = (label) => label[FRUITS.MANGO];
```

while hoisting properties. In the code above, the initializer of the second variable declarator should be visited after hoisting the first variable declarator. Otherwise `FRUITS.MANGO` cannot be handled.

**Related issue:**

 - Closes #9030

PR #9039, opened 2024-06-12
perf(es/minifier): Do not visit var init multiple times

kdy1:
**Description:**

I mistakenly introduced a performance regression with https://github.com/swc-project/swc/pull/9032. It makes the minifier visit the initializer of variables multiple times - once while normal visiting and once in `hoist_props_of_vars`. This PR fixes it.


kdy1 (commented):
swc-bump:
 - swc_ecma_minifier
 - swc_core

PR #9041, opened 2024-06-12
perf(es/lints): Avoid needless allocations in `no-dupe-args`

kdy1:
**Description:**

I introduced a zero-allocation variant for `swc_ecma_utils::DestructuringFinder` that does not allocate anything.

### DIFF
--- a/crates/swc_ecma_lints/src/rules/no_dupe_args.rs
+++ b/crates/swc_ecma_lints/src/rules/no_dupe_args.rs
@@ -16,7 +16,9 @@ struct NoDupeArgs {}
 /// is usually small.
 macro_rules! check {
     ($node:expr) => {{
-        // This allocates only if there are duplicate parameters.
+        // This vector allocates only if there are duplicate parameters.
+        // This is used to handle the case where the same parameter is used 3 or more
+        // times.
         let mut done = vec![];
 
         let mut i1 = 0;
--- a/crates/swc_ecma_lints/src/rules/no_dupe_args.rs
+++ b/crates/swc_ecma_lints/src/rules/no_dupe_args.rs
@@ -16,6 +16,9 @@ struct NoDupeArgs {}
 /// is usually small.
 macro_rules! check {
     ($node:expr) => {{
+        // This allocates only if there are duplicate parameters.
+        let mut done = vec![];
+
         let mut i1 = 0;
         for_each_binding_ident($node, |id1| {
             i1 += 1;
@@ -24,9 +27,11 @@ macro_rules! check {
             for_each_binding_ident($node, |id2| {
                 i2 += 1;
 
-                if i1 >= i2 {
+                if i1 >= i2 || done.contains(&i1) {
                     return;
                 }
+
+                done.push(i1);
 
                 if id1.span.ctxt == id2.span.ctxt && id1.sym == id2.sym {
                     HANDLER.with(|handler| {
--- a/crates/swc_ecma_lints/src/rules/no_dupe_args.rs
+++ b/crates/swc_ecma_lints/src/rules/no_dupe_args.rs
@@ -24,7 +24,7 @@ macro_rules! check {
             for_each_binding_ident($node, |id2| {
                 i2 += 1;
 
-                if i1 == i2 {
+                if i1 >= i2 {
                     return;
                 }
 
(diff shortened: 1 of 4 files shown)


kdy1 (commented):
swc-bump:
 - swc_ecma_lints
 - swc_ecma_utils

crates/swc_ecma_lints/src/rules/no_dupe_args.rs
```diff
@@ -13,54 +12,65 @@ pub fn no_dupe_args() -> Box<dyn Rule> {
...
-                        .emit();
-                });
-            }
-        });
-    }
+/// This has time complexity of O(n^2), but it's fine as the number of paramters
+/// is usually small.
+macro_rules! check {
+    ($node:expr) => {{
+        // This allocates only if there are duplicate parameters.
```
sokra, line 19: I don't think that's true...

crates/swc_ecma_lints/src/rules/no_dupe_args.rs
```diff
@@ -13,54 +12,65 @@ pub fn no_dupe_args() -> Box<dyn Rule> {
...
+
+        let mut i1 = 0;
+        for_each_binding_ident($node, |id1| {
+            i1 += 1;
+
+            let mut i2 = 0;
+            for_each_binding_ident($node, |id2| {
+                i2 += 1;
+
+                if i1 >= i2 || done.contains(&i1) {
```
sokra, line 30: The `Vec::contains` makes the whole algorithm `O(n^3)`

crates/swc_ecma_lints/src/rules/no_dupe_args.rs
```diff
@@ -13,54 +12,65 @@ pub fn no_dupe_args() -> Box<dyn Rule> {
...
-                        )
-                        .span_label(old_span, "previous definition here".to_string())
-                        .span_label(span, &"used as parameter more than once".to_string())
-                        .emit();
-                });
-            }
-        });
-    }
+/// This has time complexity of O(n^2), but it's fine as the number of paramters
+/// is usually small.
```
sokra, line 16: This probably holds true for 99.9%, but maybe we can add another implementation that is used for n > 100 based on a hashmap.

crates/swc_ecma_lints/src/rules/no_dupe_args.rs
```diff
@@ -13,54 +12,65 @@ pub fn no_dupe_args() -> Box<dyn Rule> {
...
+
+            let mut i2 = 0;
+            for_each_binding_ident($node, |id2| {
+                i2 += 1;
+
+                if i1 >= i2 || done.contains(&i1) {
+                    return;
+                }
+
+                done.push(i1);
```
sokra, line 34: I don't really get what this should do...

crates/swc_ecma_lints/src/rules/no_dupe_args.rs
kdy1: This comment was about `done,` but well I made a mistake.

crates/swc_ecma_lints/src/rules/no_dupe_args.rs
```diff
@@ -13,54 +12,65 @@ pub fn no_dupe_args() -> Box<dyn Rule> {
...
+
+                if id1.span.ctxt == id2.span.ctxt && id1.sym == id2.sym {
+                    HANDLER.with(|handler| {
+                        handler
+                            .struct_span_err(
+                                id2.span,
+                                &format!(
+                                    "the name `{}` is bound more than once in this parameter list",
```
sokra, line 42: If a binding is defined 3 times this error will be reported multiple times. That was already the case with the old implementation.

crates/swc_ecma_lints/src/rules/no_dupe_args.rs
kdy1: This was a mistake. It's now fixed.

crates/swc_ecma_lints/src/rules/no_dupe_args.rs
kdy1: That's handled by `done`. Actually I added `done` just to match the original behavior.

crates/swc_ecma_lints/src/rules/no_dupe_args.rs
kdy1: Now `done` is empty for almost all cases.

PR #9047, opened 2024-06-14
perf(es/parser): Do not track `raw` by hand

kdy1:
**Description:**

We can skip this logic completely, as it's `raw`.

~I'm under the impression that I'm super stupid~


kdy1 (commented):
swc-bump:
 - swc_ecma_ast --breaking
 - swc_ecma_codegen
 - swc_ecma_parser
 - swc_core

PR #9055, opened 2024-06-15
perf: Introduce `SourceSlice`

kdy1:
**Description:**

`SourceSlice` is sort of `&' input str` that is owned. It works by using `Lrc<str>`, which is the type of `SourceFile.src`. This type is used to avoid (enormous amount of) allocations for `raw` fields. With this PR, `raw` fields are now `SourceSlice`, which does not allocate.

**BREAKING CHANGE:**

 - `swc_common::SourceFile.src` is now `Lrc<str>` instead of `Lrc<String>`.
 - `swc_common::input::StringInput` now takes `&'a Lrc<str>` instead of `&'a str`.

### DIFF
--- a/crates/swc_common/src/input.rs
+++ b/crates/swc_common/src/input.rs
@@ -2,7 +2,10 @@ use std::str;
 
 use debug_unreachable::debug_unreachable;
 
-use crate::syntax_pos::{BytePos, SourceFile};
+use crate::{
+    source_slice::SourceSlice,
+    syntax_pos::{BytePos, SourceFile},
+};
 
 pub type SourceFileInput<'a> = StringInput<'a>;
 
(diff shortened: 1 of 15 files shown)


kdy1:
I'll postpone this PR. Performance difference is about ~2%

kdy1:
With boxing, the performance is much better.
The problem is now `rkyv` implementations for Wasm plugins...

```
Gnuplot not found, using plotters backend
es/lexer/cal-com        time:   [5.8715 ms 5.8856 ms 5.9024 ms]
                        change: [-4.0248% -3.7259% -3.4050%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) high mild
  5 (5.00%) high severe

     Running benches/parser.rs (/Users/kdy1/projects/s/perf/target/release/deps/parser-b0b086eaa04716f6)
Gnuplot not found, using plotters backend
es/parser/cal-com       time:   [14.626 ms 14.713 ms 14.815 ms]
                        change: [-3.0856% -2.3994% -1.6524%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 16 outliers among 100 measurements (16.00%)
  7 (7.00%) high mild
  9 (9.00%) high severe

```

PR #9056, opened 2024-06-15
perf(es/parser): Reduce allocations for `raw` while lexing numbers

kdy1:
**Description:**

We don't need to allocate one byte at a time while lexing numbers.


kdy1 (commented):
swc-bump:
 - swc_common --breaking

kdy1 (commented):
swc-bump:
 - swc_ecma_parser --breaking

PR #9057, opened 2024-06-16
perf(es/parser): Reduce allocations while lexing numbers

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_core
 - swc_ecma_parser

PR #9069, opened 2024-06-17
fix(types): add missing typeAnnotation field in Identifier node

maastrich:
**Description:**

I was working on a walker for SWC ast and found out that the Identifier can have a `typeAnnotation` field which is not documented in the `Identifier` type. I have added the `typeAnnotation` field to the `Identifier` type.

**BREAKING CHANGE:**

No breaking changes.

No related issue.

### DIFF
--- a/bindings/binding_core_wasm/src/types.rs
+++ b/bindings/binding_core_wasm/src/types.rs
@@ -1357,6 +1357,8 @@ export interface Identifier extends ExpressionBase {
 
   /// TypeScript only. Used in case of an optional parameter.
   optional: boolean;
+
+  typeAnnotation?: TsTypeAnnotation;
 }
 
 export interface OptionalChainingExpression extends ExpressionBase {
--- a/bindings/binding_core_wasm/src/types.rs
+++ b/bindings/binding_core_wasm/src/types.rs
@@ -1357,8 +1357,6 @@ export interface Identifier extends ExpressionBase {
 
   /// TypeScript only. Used in case of an optional parameter.
   optional: boolean;
-
-  typeAnnotation?: TsTypeAnnotation;
 }
 
 export interface OptionalChainingExpression extends ExpressionBase {
@@ -2014,6 +2012,8 @@ export interface BindingIdentifier extends PatternBase {
   type: "Identifier";
   value: string;
   optional: boolean;
+
+  typeAnnotation?: TsTypeAnnotation;
 }
 
 export interface ArrayPattern extends PatternBase {
--- a/bindings/binding_minifier_wasm/src/types.rs
+++ b/bindings/binding_minifier_wasm/src/types.rs
@@ -1357,6 +1357,8 @@ export interface Identifier extends ExpressionBase {
 
   /// TypeScript only. Used in case of an optional parameter.
   optional: boolean;
+
+  typeAnnotation?: TsTypeAnnotation;
 }
 
 export interface OptionalChainingExpression extends ExpressionBase {
--- a/bindings/binding_minifier_wasm/src/types.rs
+++ b/bindings/binding_minifier_wasm/src/types.rs
@@ -1357,8 +1357,6 @@ export interface Identifier extends ExpressionBase {
 
   /// TypeScript only. Used in case of an optional parameter.
   optional: boolean;
-
-  typeAnnotation?: TsTypeAnnotation;
 }
 
 export interface OptionalChainingExpression extends ExpressionBase {
@@ -2014,6 +2012,8 @@ export interface BindingIdentifier extends PatternBase {
   type: "Identifier";
   value: string;
   optional: boolean;
+
+  typeAnnotation?: TsTypeAnnotation;
 }
 
 export interface ArrayPattern extends PatternBase {
--- a/packages/types/index.ts
+++ b/packages/types/index.ts
@@ -1358,6 +1358,8 @@ export interface Identifier extends ExpressionBase {
 
     /// TypeScript only. Used in case of an optional parameter.
     optional: boolean;
+
+    typeAnnotation?: TsTypeAnnotation;
 }
 
 export interface OptionalChainingExpression extends ExpressionBase {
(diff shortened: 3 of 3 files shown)


kdy1 (approved):
swc-bump:
 - dbg-swc

kdy1 (changes requested):
Identifier does not have type annotation. Only binding identifiers have.

kdy1 (changes requested):
Identifier does not have type annotation. Only binding identifiers have.

maastrich:
@kdy1 not sure what you mean by that 🤔 
when using the parse function I end up with Identifiers with typeAnnotation

```ts
import { parseSync } from '@swc/core'

console.log(JSON.stringify(parseSync("const foo: string = 'bar'", { syntax: 'typescript' }), null, 2))
```

outputs: 
```json
{
  "type": "Module",
  "span": {
    "start": 6,
    "end": 31,
    "ctxt": 0
  },
  "body": [
    {
      "type": "VariableDeclaration",
      "span": {
        "start": 6,
        "end": 31,
        "ctxt": 0
      },
      "kind": "const",
      "declare": false,
      "declarations": [
        {
          "type": "VariableDeclarator",
          "span": {
            "start": 12,
            "end": 31,
            "ctxt": 0
          },
          "id": {
            "type": "Identifier",
            "span": {
              "start": 12,
              "end": 15,
              "ctxt": 2
            },
            "value": "foo",
            "optional": false,
            "typeAnnotation": {
              "type": "TsTypeAnnotation",
              "span": {
                "start": 15,
                "end": 23,
                "ctxt": 0
              },
              "typeAnnotation": {
                "type": "TsKeywordType",
                "span": {
                  "start": 17,
                  "end": 23,
                  "ctxt": 0
                },
                "kind": "string"
              }
            }
          },
          "init": {
            "type": "StringLiteral",
            "span": {
              "start": 26,
              "end": 31,
              "ctxt": 0
            },
            "value": "bar",
            "raw": "'bar'"
          },
          "definite": false
        }
      ]
    }
  ],
  "interpreter": null
}
```

kdy1:
Try 
```ts
const foo: string = bar;
```

`foo` here is a pattern / binding identifier, but `bar` is a identifier reference.

Select View: JSON AST from [the playground](https://play.swc.rs/?version=1.6.1&code=H4sIAAAAAAAAA0vOzysuUUjLz7dSKC4pysxLV7BVSEosAgBPza3dFwAAAA%3D%3D&config=H4sIAAAAAAAAA1WPSw7DIAxE9zkF8rrbbnqHHgJRJyLiJ%2BxIpVHuXiBAmx2eGfsN%2ByQErKTgIfb8zEOQkTCOOSuUHMt3VoBTQFJRB4Zbd5mKNUtDWKXjdIBlXJDLFtK9xcF4T9jjTbPa6Tn9A5W3ISLRNVii0i0Gr7ipIcH611bN9pFStuDzNevdSqOD%2FBQaxw1%2Fu73D4IGmZz9Yo8cXZg1gZysBAAA%3D). It's on the right top.


```json
{
  "type": "Module",
  "span": {
    "start": 0,
    "end": 23,
    "ctxt": 0
  },
  "body": [
    {
      "type": "VariableDeclaration",
      "span": {
        "start": 0,
        "end": 23,
        "ctxt": 0
      },
      "kind": "const",
      "declare": false,
      "declarations": [
        {
          "type": "VariableDeclarator",
          "span": {
            "start": 6,
            "end": 23,
            "ctxt": 0
          },
          "id": {
            "type": "Identifier",
            "span": {
              "start": 6,
              "end": 9,
              "ctxt": 2
            },
            "value": "foo",
            "optional": false,
            "typeAnnotation": {
              "type": "TsTypeAnnotation",
              "span": {
                "start": 9,
                "end": 17,
                "ctxt": 0
              },
              "typeAnnotation": {
                "type": "TsKeywordType",
                "span": {
                  "start": 11,
                  "end": 17,
                  "ctxt": 0
                },
                "kind": "string"
              }
            }
          },
          "init": {
            "type": "Identifier",
            "span": {
              "start": 20,
              "end": 23,
              "ctxt": 1
            },
            "value": "bar",
            "optional": false
          },
          "definite": false
        }
      ]
    }
  ],
  "interpreter": null
}
```

Note that 

```json
          "init": {
            "type": "Identifier",
            "span": {
              "start": 20,
              "end": 23,
              "ctxt": 1
            },

```


does not have `typeAnnotation`

maastrich:
@kdy1 actually it does, I guess you missmatched a bracket, the one at the end of the snippet you sent is closing the `"span"` property 🤔 
the full node is:

```json
          "id": {
            "type": "Identifier",
            "span": {
              "start": 6,
              "end": 9,
              "ctxt": 2
            },
            "value": "foo",
            "optional": false,
            "typeAnnotation": {
              "type": "TsTypeAnnotation",
              "span": {
                "start": 9,
                "end": 17,
                "ctxt": 0
              },
              "typeAnnotation": {
                "type": "TsKeywordType",
                "span": {
                  "start": 11,
                  "end": 17,
                  "ctxt": 0
                },
                "kind": "string"
              }
            }
          },
```

morganney:
@maastrich I believe @kdy1 is speaking only of the `init` property of a `VariableDeclarator` node that does not have a `typeAnnotation`. In your example the `init` is a `StringLiteral` node, whereas in his example the `init` is a an `Identifier`.

I'm wondering what source code would produce a `BindingIdentifier` node and how that is fundamentally different from an `Identifier` node? Looking at the definitions of both, the seem identical:

```js
interface BindingIdentifier extends PatternBase {
    type: "Identifier";
    value: string;
    optional: boolean;
}

interface Identifier extends ExpressionBase {
    type: "Identifier";
    value: string;
    optional: boolean;
}
```

A `VariableDeclarator.id` is a `Pattern` which includes both `BindingIdentifier` and `Identifier`, and both examples of source code produce a `typeAnnotation` on the `VariableDeclarator.id` node.

maastrich:
Ok, I understand it better now, there are few irregularities with the types, the type behind `TsTemplateLiteralType` and `TemplateLiteral` cannot be differentiated like `BindingIdentifier` and `Identifier` may be can we find a way to discriminate them easily ?

maastrich:
@kdy1 It should work better for you like this ?

maastrich:
Closing as it was a missunderstanding of how nodes of same node type can be defined differently

PS: @kdy1 is there a list of overlapping node types ?

PR #9073, opened 2024-06-18
perf(common): Use `Chars` in `StringInput`

kdy1:
**Description:**

Now it's possible


kdy1 (commented):
swc-bump:
 - swc_core
 - swc_common

Issue #3854, opened 2021-09-21
Incorrect Jest coverage 


stephane-arista:
When using @swc/jest to transpile ts to commonjs and running `jest --coverage` certain branches are shown as not covered (console logging in these branches show that tests do run the code path). Using babel to transpile and run the tests shows the correct coverage.

kdy1:
You need `sourceMaps: true` or `sourceMaps: "inline"`.
I verified that it's working

bobaaaaa:
@kdy1 Hm, I tested both `sourceMaps` in `.swcrc` and as an option in the `jest.config.json`. Both did not work for me. Even with `target: "es2021"`.

kdy1:
Hmm... Source maps are valid, maybe emitted tokens without sourcemap entry can be the cause I guess?
Some sourcemap libraries have bugs related to it.

bobaaaaa:
@kdy1 I updated the gist with the generated .js + .js.map files: https://gist.github.com/bobaaaaa/3649b3a7e6312793a257bf67c500128a

kdy1:
I'm not sure why does jest can't understand the sourcemap.
I need to dig into jest to know the source map library it uses...

bobaaaaa:
@kdy1 I tested the new version. Unfortunately, its still not fixed :(

Tested with `es2020` & `es2021` + `sourceMaps: true` & `sourceMaps: "inline"`. I updated all files in my gist build with `v1.2.155`:
https://gist.github.com/bobaaaaa/3649b3a7e6312793a257bf67c500128a

(Keep in mind, this is not a huge blocker/issue for us. Still thx for looking into this)

kdy1:
Can you try the latest version? (`v1.2.156`)
There was a bug fix for a module that only contains `export * from './foo'` and such module is super common, so I think it can be the cause of this.

wight554:
Can confirm same issue with vitest + swc

bobaaaaa:
@kdy1 problem is still there with `v1.2.156`

bobaaaaa:
@kdy1 I saw you created another fix (https://github.com/swc-project/swc/pull/4074) 👍 
Because in `1.2.157` the issue is still present.

kdy1:
I changed the milestone, thanks!

kdy1:
I triggered publishing of the new version

sscode02:
I see you fixed this in v1.2.158, but the problem persists after I update, is it my configuration problem?
```
transform: {
[...]

PR #9074, opened 2024-06-18
fix(es/decorators): Reduce the number of sourcemap mappings

kdy1:
**Description:**

Coverage gets better if we generate fewer source map entries. This PR only fixes the issue for `"coverageProvider": "v8"`. Much more work is required for the default coverage provider I guess.

**Related issue:**

 - #3854


kdy1 (commented):
swc-bump:
 - swc_ecma_ast
 - swc_core

PR #9075, opened 2024-06-18
perf(es/lexer): Optimize lexer

kdy1:
**Description:**

 - `TokenContext` is now 1 byte.
 - `read_word_with` now uses `&dyn` instead of `impl FnOnce`.


kdy1 (commented):
---

swc-bump:
 - swc_ecma_parser
 - swc_core

PR #9080, opened 2024-06-19
perf(es/lexer): Add a fast-path to template literal lexer

kdy1:
**Description:**

We don't need to allocate at all in the fast path.


kdy1 (commented):
swc-bump:
 - swc_core
 - swc_ecma_parser

PR #9081, opened 2024-06-19
perf(es/lexer): Add fast-path to jsx lexer

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_ecma_parser
 - swc_core

PR #9082, opened 2024-06-19
perf(es/lexer): Use `debug_unreachable!`

kdy1:
**Description:**

This will result in a smaller binary and a minor performance gain.


kdy1:
Closing as it was wrong

PR #9084, opened 2024-06-20
feat(es/typescript): Support `deferred` modifier

magic-akari:

- https://github.com/microsoft/TypeScript/pull/58729

### DIFF
--- a/crates/swc_ecma_ast/src/typescript.rs
+++ b/crates/swc_ecma_ast/src/typescript.rs
@@ -85,6 +85,7 @@ pub struct TsParamProp {
     pub accessibility: Option<Accessibility>,
     #[cfg_attr(feature = "serde-impl", serde(rename = "override"))]
     pub is_override: bool,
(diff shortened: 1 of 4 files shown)


magic-akari:
https://github.com/swc-project/swc/blob/bd2509bafa35baa52db10385f0aae720768e5538/crates/swc_ecma_ast/src/function.rs#L15-L17

I need to update the type of params from `Vec<Param>` to `Vec<ParamOrTsParamProp>`, which will cause a lot of changes.

kdy1:
Merging the two types sounds good to me

magic-akari:
I have observed that Babel also distinctly differentiates between `TSParameterProperty` and `Pattern`.
It might be necessary to communicate with Babel to explore better implementation methods.

magic-akari:
TypeScript has adopted the `immediate` keyword. Moreover, we should await the completion of #9313.

PR #9095, opened 2024-06-21
perf(es/lexer): Apply various optimizations

kdy1:
**Description:**

 - `TokenContexts` now uses smallvec, as the element type is very small (1 byte) now.
 - Some codes are moved to allow CPU to cache code correctly.


kdy1 (commented):
swc-bump:
 - swc_ecma_parser
 - swc_core

PR #9096, opened 2024-06-22
perf(es/lexer): Use `memchr::memmem` for lexing block comments

kdy1:
**Description:**

This PR also makes the lexer more CPU-cache friendly.


kdy1 (commented):
swc-bump:
 - swc_ecma_parser
 - swc_core

PR #9100, opened 2024-06-23
perf(es/parser): Optimize macros

kdy1:
**Description:**

I could optimize macros used enormous amount of times.


kdy1 (commented):
swc-bump:
 - swc_core
 - swc_ecma_parser

PR #9102, opened 2024-06-24
perf(es/parser): Remove unnecessary check

kdy1:
**Description:**

We can skip some checks for error recovery.


kdy1 (commented):
swc-bump:
 - swc_core
 - swc_ecma_parser

PR #9107, opened 2024-06-26
fix(es/minifier): Remove pure function used by variable initialization

CPunisher:
**Description:**

https://play.swc.rs/?version=1.6.5&code=H4sIAAAAAAAAA9PXVygvyixJVcgvUihILAYykvNTUhUyUotSubjKEosUwlNTs10SKxVsFfS1lOPjA0KDXOPjtfQVNNJK85JLMvPzFDSgSow0Faq5FGAajPQSgXqUgkvzUhIrlayBEkWpJaVFeXB5rlpNmE6FmhqF6lpNay4ADxpWp44AAAA%3D&config=H4sIAAAAAAAAA4WVO47jMAxA%2B5wiUD3FIsUWc4Dt5gyCYlGOsvoYIpWJMcjdl7aTiSdr2q4M8okUv%2Fra7ffqjI1633%2Fx7%2FSpzhSE8kPGUuwTmStLFTTRYFN8R%2BptTpxxUDsTEL7FtyehQs4ID2Imjz5517%2F6a3LsCiC%2ByFljSlsjJML%2Fbd31JX8OSir1VXfMOYBJG1ptUPtE0EKRnDQ5BNMh6IspgrUhAlM8ZsndAFQCq7uSO5FJ1pPPie%2BwTFgwVjfZgqD2BRryF5COs28%2BnpBDF2IdEQvH2rZjXyxYgYsJ1ZBwB7iOpeQoBA%2Bn7JG0q0lK9wSs5GkC7sVYsuCdLkC1pOXz5%2BzTSi3%2FAnCWgkFMJoLkY6Qc9%2BeaFbdpwSfHI0G9wPAcSVlI0HIRtPdOqMSQQSjkpW4oYGsDQyUa6Yp3ZCXV6C1ocI77TnCDn56ak3QJ6jvITlByjxgndeqk1N8bYIUZhm8D%2BcNZILlp71Q0dFonsI%2FHHDacRaBTthsQl4%2FyOlJ4a127daYmC9xiYEWs4qhcXkg8aJR1AJJ6jEeRres25ONzbc2g2%2FyEiia1z92z%2BHQQz9ToTgEefh0O%2FPDsZszzCZlFozx%2BZFvDz0lQ8SF7vCpju02mf6vJ6u72Dzzk3g4aBwAA

The above code will not be optimized in SWC but will be optimized in Terser. Terser analyzes variable usage step-by-step, so it first determines whether WeekDay is used at the top level. It does not consider the parameter in the call expression within the variable declaration as a reference to this variable. Additionally, Terser does not inline pure functions because inlining them would directly result in the variable being referenced at the top level.

See: https://github.com/terser/terser/blob/master/lib/compress/inline.js#L347

Perhaps this variable should be optimized during the DCE (Dead Code Elimination) phase? However, the DCE phase cannot determine the pure mark.

Maybe a more fundamental reason is that all side effect determinations should directly consider the pure mark, like: https://github.com/terser/terser/blob/9083832e6818eeda9b892c4140716a3d3a2632b7/lib/compress/inference.js#L884

I am not sure if the code I modified is correct, and I am somewhat confused about the determination of side effects and pure marks in the project...

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/optimize/iife.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/iife.rs
@@ -470,6 +470,13 @@ impl Optimizer<'_> {
             return;
         }
 
+        trace_op!("iife: Checking pure");
+
+        if self.has_pure(call.span) {
+            log_abort!("iife: Has pure mark");
+            return;
+        }
+
         trace_op!("iife: Checking callee");
 
         match callee {
--- a/crates/swc_ecma_minifier/src/compress/optimize/mod.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/mod.rs
@@ -833,7 +833,8 @@ impl Optimizer<'_> {
                     BlockStmtOrExpr::Expr(_) => false,
                 },
                 _ => false,
-            } && args.is_empty() =>
+            } && args.is_empty()
+                || e.span().has_mark(self.marks.pure) =>
             {
                 report_change!("ignore_return_value: Dropping a pure call");
                 self.changed = true;
--- a/crates/swc_ecma_minifier/src/compress/optimize/util.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/util.rs
@@ -77,6 +77,11 @@ impl<'b> Optimizer<'b> {
         span.has_mark(self.marks.noinline)
     }
 
+    /// Check for `/*#__PURE__*/`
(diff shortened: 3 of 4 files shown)


kdy1:
Can you run `UPDATE=1 cargo test -p swc -p swc_ecma_minifier --features concurrent` and commit/push the changes?

PR #9114, opened 2024-06-29
feat(es/parser): Support stripping flow with ES parser

kdy1:
**Description:**

This PR adds support for flow syntax without a feature to construct AST. The parser strips out extra syntax, and the returned AST is identical to ECMAScript AST.


rtrembecky:
hello, last week I found out I can't switch to @swc/jest for a React Native project, where react-native and expo dependencies need to be transformed, because they include the Flow types. do you think this will be merged in the foreseeable future? thanks.

rtrembecky:
BTW I found https://github.com/changwoolab/flow-aware-swc-jest and tried it but also got some error. just sharing for visibility/reference

rtrembecky:
@kdy1 hi, any update on this effort?

kdy1:
@rtrembecky Is there any concrete use plan?

rtrembecky:
well it's simple, our RN tests take too long to spin and one of the issues I identified is babel transpilation of RN modules, which are being shipped with Flow. I'd like to switch to a faster engine (SWC) but still support Flow-stripping.

jbroma:
In [Re.Pack 5](https://github.com/callstack/repack) we're using `swc` through Rspack and right now we're stripping flow types with `flow-remove-types` utility. 

It would be amazing to have a builtin support for doing this with `swc` 🙌

rtrembecky:
does swc support stripping Flow types now?

PR #9118, opened 2024-07-01
perf(es/lexer): Do not use `memchr::memmem` in parser

kdy1:
Reverts swc-project/swc#9096.
I'm reverting this as this regressed performance in `linux-x86_64`.


kdy1 (commented):
swc-bump:
 - swc_ecma_parser
 - swc_core

PR #9124, opened 2024-07-02
feat(binding): Create Wasm package for stripping only TypeScript

kdy1:
**Description:**

This PR adds a Wasm binding which is only capable of stripping TypeScript types.

**Related issue:**

 - https://github.com/marco-ippolito/node/pull/2


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #9127, opened 2024-07-03
refactor(es/typescript): Extract type annotation proposal out

kdy1:
**Description:**

The Node.js team wants a Wasm package that only implements type annotation proposals. So we need to refactor our TypeScript transform to allow it.

**Related issue:**

 - https://github.com/nodejs/loaders/issues/208#issuecomment-2205063110

### DIFF
--- a/bindings/binding_typescript_wasm/src/lib.rs
+++ b/bindings/binding_typescript_wasm/src/lib.rs
@@ -18,7 +18,7 @@ use swc_core::{
                 hygiene::hygiene,
                 resolver,
             },
-            typescript::typescript,
+            typescript::strip_type,
         },
         visit::VisitMutWith,
     },
@@ -55,9 +55,8 @@ pub struct Options {
     #[serde(default)]
     pub source_maps: bool,
 
-    #[serde(default)]
-    pub transform: swc_core::ecma::transforms::typescript::Config,
-
+    // #[serde(default)]
+    // pub transform: swc_core::ecma::transforms::typescript::Config,
     #[serde(default)]
     pub codegen: swc_core::ecma::codegen::Config,
 }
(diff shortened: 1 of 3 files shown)


kdy1:
@magic-akari Can you take a look? I looked at the TypeScript transform source code, but I'm not sure if applying `StripType::default()` is enough.

magic-akari:
> @magic-akari Can you take a look? I looked at the TypeScript transform source code, but I'm not sure if applying `StripType::default()` is enough.

I think it's almost ok.
Let's check the config in crates/swc_ecma_transforms_typescript/src/config.rs

Maybe we should change the config about how to handle TS import/export statements.

magic-akari:
> What I would like is for some option or collection of options where I can use SWC to implement the Type Annotations proposal, but nothing more. In particular, see https://github.com/tc39/proposal-type-annotations#intentional-omissions:

_Originally posted by @GeoffreyBooth in https://github.com/nodejs/loaders/issues/208#issuecomment-2205063110_

Based on this requirement, I believe we only need to export `StripType`. What do you think? @kdy1

kdy1:
I was thinking the same. I'll add a function for it.

kdy1:
Thank you!

kdy1 (commented):
swc-bump:
 - swc_core

magic-akari:
What happens when only using `StripType::default()`:

Enums, namespaces will be ignored and left behind. (Originally processed by Transform)
Implicit type imports will be ignored and left behind, only syntactic type imports will be processed by StripType. (Originally processed by StripImportExport)
`import foo = require('foo')`, `export = foo` will be ignored and left behind. (Originally processed by Transform)
Class constructor parameter properties will be ignored and left behind. (Originally processed by Transform)

We need to confirm with the node team whether we are aligned.

magic-akari (approved):

PR #9133, opened 2024-07-04
perf(es): Reduce allocations for dynamic stacks

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_ecma_parser
 - swc_ecma_utils
 - swc_core

PR #9134, opened 2024-07-04
fix(es/loader): Exclude `.json` from default extension list

### LINKED ISSUE #9121: SWC transpiles node_module import to JSON file of the same name
### Describe the bug

When setting up TSOA, there is a config file called `tsoa.json` that we must create in our root project. At the same time there is also some decorator imports from the `tsoa` package. When we try to build our project with SWC the following error happens when running the transpiled code with node

```
class TestController extends _tsoa.Controller {
                                   ^

TypeError: Class extends value undefined is not a constructor or null
```

This is happening because SWC is incorrectly transpiling the require into the JSON file and not the node_module package.
In this [codesandbox example](https://codesandbox.io/p/devbox/hidden-wave-hkjqfk), `build/controllers/test.js` has the following incorrect line `const _tsoa = require("../../tsoa");` instead of `const _tsoa = require("tsoa");`

By changing the file to something else (e.g `tsoa.config.json`) SWC will correctly transpile the code to the node_module package.

On TSC, this issue is not present

### Input code

_No response_

### Config

```json
{
    "$schema": "https://swc.rs/schema.json",
    "module": {
        "type": "commonjs"
    },
    "sourceMaps": true,
    "jsc": {
        "target": "es2022",
        "baseUrl": "./src",
        "parser": {
            "syntax": "typescript",
            "tsx": false,
            "decorators": true,
            "dynamicImport": false,
            "privateMethod": false,
            "functionBind": false,
            "exportDefaultFrom": false,
            "exportNamespaceFrom": false,
            "topLevelAwait": false,
            "importMeta": false
        }
    }
}
```

https://github.com/NullScope/swc-json-node_module-clash

### SWC Info output

_No response_

### Expected behavior

Whenever SWC encounters an import that match both a node_module and a JSON file, the node_module should take precedence

### Actual behavior

_No response_

### Version

1.6.6

### Additional context

_No response_

kdy1:
**Description:**

**Related issue:**

 - Closes #9121

### DIFF
--- a/crates/swc_ecma_loader/src/resolvers/node.rs
+++ b/crates/swc_ecma_loader/src/resolvers/node.rs
@@ -109,7 +109,7 @@ pub struct NodeModulesResolver {
     ignore_node_modules: bool,
 }
 
-static EXTENSIONS: &[&str] = &["ts", "tsx", "js", "jsx", "json", "node"];
+static EXTENSIONS: &[&str] = &["ts", "tsx", "js", "jsx", "node"];
 
 impl NodeModulesResolver {
     /// Create a node modules resolver for the target runtime environment.


kdy1 (commented):
swc-bump:
 - swc_core
 - swc_ecma_loader

PR #9138, opened 2024-07-04
feat(bindings/ts): Add transform/strip-only mode

kdy1:
**Description:**

This PR adds `strip-only`/`transform` mode to `@swc/wasm-typescript`.

- Both mode errors on decorator usages.
- In `strip-only` mode, `enum` and TypeScript parameter properties are treated as an invalid syntax.
- In `transform` mode, those are transpiled.

### DIFF
--- a/bindings/binding_typescript_wasm/Cargo.toml
+++ b/bindings/binding_typescript_wasm/Cargo.toml
@@ -13,9 +13,6 @@ bench      = false
 crate-type = ["cdylib"]
 
 [features]
-default = ["swc_v1"]
-swc_v1  = []
-swc_v2  = []
 
 [dependencies]
 anyhow = "1.0.66"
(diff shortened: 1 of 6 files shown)


kdy1 (commented):
swc-bump:
 - dbg-swc

GeoffreyBooth:
@kdy1 Thank you for adding this! Is there a way I can test this new mode in the SWC playground? When I go to https://swc.rs/playground and choose `1.6.8-nightly-20240706.1` and try to add `"mode": "strip-only"` at various places in the JSON config editor, they all error.

kdy1:
Fast TS strip is a separate tool, so if you want to try it on the web, we may need to build another playground. The fastest way to test output is adding a test file (or modify one) in https://github.com/swc-project/swc/tree/2fe2feeb1fb0d1776295da28b0af792e2a19da3f/crates/swc_fast_ts_strip/tests and run `UPDATE=1 cargo test -p swc_fast_ts_strip`. It requires `rustup` to run.

GeoffreyBooth:
@kdy1 Thanks. I think if `--experimental-strip-types` ships in Node, users will want a way to see what source code Node is evaluating. Adding this to the existing playground, or creating a new one, sounds like a great idea if you’re up for it.

kdy1:
@GeoffreyBooth I filed https://github.com/swc-project/swc-playground/issues/42

kdy1:
@GeoffreyBooth Now the SWC playground has one! 
image


Thank you, @g-plane !

PR #9142, opened 2024-07-04
perf(es/minifier): Make minifier less parallel

kdy1:
**Description:**

This PR remove needless parallelization points.


kdy1:
This was a wrong fix as there was threshold logic

PR #9143, opened 2024-07-05
feat(es/typescript): Add `swc_fast_ts_strip`

kdy1:
**Description:**

This PR also adds CI process for each Wasm bindings


kdy1 (commented):
swc-bump:
 - dbg-swc

kdy1 (commented):
swc-bump:
 - swc_fast_ts_strip

magic-akari (approved):

magic-akari:
There are still some details that need to be addressed, but we can merge first and then complete it in the following PR, let's keep moving forward quickly.

crates/swc_fast_ts_strip/tests/errors/modules.ts
```diff
@@ -0,0 +1 @@
+module 'foo' { }
```
robpalme, line 1: ### Ambient Module Declaration

This is an Ambient Module Declaration, not a `module`-keyword namespace.  It should be stripped as a type.  It should not error.

```ts
module 'myAmbientModuleDeclaration' { }   // this line disappears
```

### `module`-keyword namespace

The use of the legacy [`module`-keyword namespace](https://github.com/microsoft/TypeScript/issues/57913) should error, just like it does for `namespace`-keyword namespaces.

```ts
module aModuleKeywordNamespace { }  // this line errors
```

### type-only namespaces

The use of type-only namespaces should be permitted and stripped like any other type.

`declare namespace Foo { }    // this line disappears`

### type-only module-keyword namespaces

The use of type-only module-keyword namespaces is a grey/opinionated area and depends on whether you want to promote modern code or retain compatibility with legacy usage.  Personally I'd suggest erroring is healthiest because the fix is trivial - the user can replace `module` with `namespace`.

`declare module Foo { }    // probably this line errors`

kdy1: I resolved first three items, and I'll keep the fourth one as-is (for now)

PR #9144, opened 2024-07-05
feat(es/codegen): Implement `blank-space` codegen

kdy1:
**Description:**

[`ts-blank-space`](https://gist.github.com/acutmore/27444a9dbfa515a10b25f0d4707b5ea2) is about stripping TypeScript types without a need for a sourcemap. With this PR, the ES codegen implements `blank_space_mode`, and it will be used to generate ES code without source map.


kdy1 (commented):
swc-bump:
 - swc_common --breaking

magic-akari:
The fixer might move the brackets, how should this part be handled?

crates/swc/tests/fixture/blank-space-codegen/strip-only/output/5.ts
```diff
@@ -0,0 +1,5 @@
+
+
+const foo = 1;
+
+const bar = call        (foo);
```
kdy1, line 5: - `const bar = call<string>(((((foo)))));`
 - `const bar = call.             (foo);`

@magic-akari

magic-akari:
```JavaScript 
(function foo(){})();
(function bar(){}());
```

crates/swc/tests/fixture/blank-space-codegen/strip-only/output/6.ts
```diff
@@ -0,0 +1,4 @@
+(function foo() {}) ();
+
+
+(function bar() {})();
```
kdy1, line 4: @magic-akari  Is this what you meant?

magic-akari: Yes, for this situation, we may need to retain the parentheses as much as possible. 
Our current operation is to remove the parentheses first and then fix them.
Considering that the type strip does not rely on the AST structure, can we remove the fixer?

kdy1: Yeap, it would be better. I'll patch it soon.

kdy1: Yeap, it would be better. I'll patch it with a follow-up PR. 

Patching it in this PR is possible, but it's better to patch in the next PR because currently there's no `blank_space_mode` in published crates (and in`./bindings`)

magic-akari: Maybe I was wrong.
The structure like `(foo?.bar as X).foo` may still need to rely on a fixer when the type is removed.
We need to handle this situation more cautiously.

kdy1:
I found that we don't need this for blank-space codegen for TypeScript. We can just replace the source string using the locations of TypeScript AST nodes.

magic-akari:
> I found that we don't need this for blank-space codegen for TypeScript. We can just replace the source string using the locations of TypeScript AST nodes.

I agree, this sounds like a simpler, more feasible, and reliable method.

PR #9152, opened 2024-07-05
feat(es/typescript): Improve fast TS stripper

kdy1:
**Description:**

**Related issue:**

 - https://github.com/swc-project/swc/pull/9143#pullrequestreview-2160591522


kdy1 (commented):
swc-bump:
 - swc_core
 - swc_fast_ts_strip

bindings/binding_typescript_wasm/__tests__/__snapshots__/transform.js.snap
```diff
@@ -1,44 +1,44 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`transform in strip-only mode should remove declare enum 1`] = `""`;
+exports[`transform in strip-only mode should remove declare enum 1`] = `"                   "`;
```
robpalme, line 3: `declare enum` is fine and should be stripped without error.  It is a pure type-only concept.

kdy1: Yeah so the result is an empty string

PR #9159, opened 2024-07-06
fix(es/typescript): Fix tricky cases in TS fast strip

kdy1:
**Description:**

We can leverage lexer information to handle tricky cases.


kdy1:
@magic-akari Do you have any other tricky cases in your mind?

magic-akari:
I believe it is sufficient to handle edge cases.

kdy1 (commented):
swc-bump:
 - swc_core
 - swc_ecma_parser
 - swc_fast_ts_strip

magic-akari (approved):

PR #9160, opened 2024-07-06
chore(ci): Fix CI

kdy1:
**Description:**

`cargo test` action is not running for an unknown reason


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #9163, opened 2024-07-06
fix(es/typescript): Fix replacement logic of fast TS strip

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_fast_ts_strip

PR #9178, opened 2024-07-08
chore(es/typescript): Improve decorator handling of fast strip

magic-akari:
**Description:**

Given that decorators will be a part of JavaScript, as specified in the [proposal](https://github.com/tc39/proposal-decorators), we retain them in the output rather than reporting an error.

**BREAKING CHANGE:**


kdy1 (approved):
swc-bump:
 - swc_core
 - swc_fast_ts_strip

PR #9180, opened 2024-07-09
perf(es/ast): Rollup merge of breaking changes

kdy1:
**Description:**

 - `Span.ctxt` is removed to reduce the size of `Span`.
 - `ctxt: SyntaxContext` is added to various AST nodes.
 - `IdentName` is used instead of `Ident` in some places like `MemberProp`, because those fields does not need `ctxt` or `optional`.


kdy1 (commented):
swc-bump:
 - swc_common --breaking

PR #9190, opened 2024-07-10
perf(es/ast): Drop `raw` fields from AST

kdy1:
**Description:**

 - Remove `raw` fields from literals in favor of `span`.
 - Add `Stmt::Directive` to preserve APIs.


kdy1:
Closing as this will break too many things

PR #9191, opened 2024-07-10
perf(bindings/wasm): Shrink the size of the wasm file

magic-akari:
**Description:**

For `binding_typescript_wasm`, it reduces the WASM size by 48% .

**BREAKING CHANGE:**


kdy1 (approved):
swc-bump:
 - dbg-swc

PR #9195, opened 2024-07-10
feat(allocator): Initialize package

kdy1:
**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_allocator/src/lib.rs
+++ b/crates/swc_allocator/src/lib.rs
@@ -1,0 +1,140 @@
+use std::ops::{Deref, DerefMut};
+
+use bumpalo::Bump;
+
+#[derive(Default)]
+pub struct Allocator {
+    alloc: Bump,
+}
+
+impl From<Bump> for Allocator {
+    fn from(alloc: Bump) -> Self {
+        Self { alloc }
+    }
+}
+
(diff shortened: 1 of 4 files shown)


crates/swc_allocator/src/lib.rs
Boshen, line 1: You should probably reference our code :-(

kdy1: Actually, I depended on `oxc_allocator` but the problem was the lack of trait implementations of `Box`.
I prefer depending on `oxc_allocator,` but I created a new package instead of requesting implementing them mainly because of `rkyv`.  Traits like `PartialEq` may be also useful for `oxc`, but AFAIK, `oxc` has nothing to do with `rkyv`. If you are fine with adding some derives for `rkyv`, I'll use `oxc_alllocator` instead.

Boshen: It's better to have your own abstraction. I was saying you should put a comment saying where the code originated from.

kdy1: Oops. I added it with https://github.com/swc-project/swc/commit/be99ce0a570c41b3fd471c5d609bd63c25740b36

PR #9196, opened 2024-07-10
perf(es/lexer): Remove `raw` fields from `Token`

kdy1:
**Description:**

We can use `raw` to slice the input.


Boshen:
You can outline all the strings by getting them from the span, and for escaped strings store it else where.

```
pub escaped_strings: FxHashMap<u32, &'a str>
```

kdy1:
@Boshen Thank you! I'll try

kdy1:
I'll try a different approach

PR #9202, opened 2024-07-11
fix(es/typescript): Handle unicode for fast ts strip

magic-akari:
**Description:**

**BREAKING CHANGE:**


crates/swc_fast_ts_strip/src/lib.rs
```diff
@@ -611,6 +664,18 @@ impl Visit for TsStrip {
...
 
+trait U8Helper {
+    fn is_utf8_char_boundary(&self) -> bool;
+}
+
+impl U8Helper for u8 {
+    #[inline]
+    fn is_utf8_char_boundary(&self) -> bool {
+        // This is bit magic equivalent to: b < 128 || b >= 192
+        (*self as i8) >= -0x40
```
oxalica, line 675: Is this copied from some legacy code, or eyeball optimized? The expression in comment compiles to exact same code with the "optimized" one: https://godbolt.org/z/KsqjM8jad

IMO, readability is more important if they are actually the same.

magic-akari: Copy from std::core::num::u8

kdy1 (approved):
swc-bump:
 - swc_fast_ts_strip

PR #7112, opened 2023-03-21
fix(es/ast): Fix `EqIgnoreSpan` impl of `Number`

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/7111.

PR #9208, opened 2024-07-11
fix(es/minifier): Fix case matching

CPunisher:
**Description:**

Caused by https://github.com/swc-project/swc/pull/7112
Since swc minifier evaluates expressions in a separate pass, I make a temporary patch here.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/9176


kdy1 (approved):
Thank you!

---


swc-bump:
 - swc_ecma_minifier
 - swc_core

PR #9209, opened 2024-07-11
fix(es/fixer): Wrap `in` expr in for-in head

### LINKED ISSUE #9200: Missing parenthesis in initializer of for loop
### Describe the bug

SWC is removing some parentheses form this for loop, and the resulting code logs `0` instead of `1`

### Input code

```typescript
let count = 0;
for (var a = 1 || (2 in {}) in { x: 1 }) count++;
console.log(count);
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true,
      "decorators": true
    },
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    },
    "target": "es2022"
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": false
}
```

https://play.swc.rs/?version=1.6.7&code=H4sIAAAAAAAAA8tJLVFIzi%2FNK1GwVTCw5krLL1LQKEssUkgE8g0VamoUNIwUMvMUqms1wZRChRVQWAHIA2vS1rbmSs7PK87PSdXLyU%2FXAAtqWnMBADlkkHNWAAAA&config=H4sIAAAAAAAAA1WPMQ7DIAxF95wCee5QMXToHXoIRJyIKgFkO1KjKHcvJICSDf%2F3P9%2FeOqXgyxbeakvPNERDjNTmpPDqxfySArJGZEsuCjwqFc5IaMEm9WgDGQnEhRxgPzlMITAmMJiJSwZm592wXkttmCMh892YrcaPU8vffhZDI0peFFk%2FtYauMJhDvxyhcmQ%2B5PS9Lqa6RCsEx5%2BaPOv2P70hJh0xAQAA

### SWC Info output

_No response_

### Expected behavior

Same code

### Actual behavior

```js
let count = 0;
for(var a = 1 || 2 in {} in {
    x: 1
})count++;
console.log(count);
```

### Version

1.6.7

### Additional context

Babel has the same bug: https://github.com/babel/babel/issues/16628

magic-akari:
- Closes #9200


crates/swc/tests/fixture/issues-9xxx/9200/output/2.js
```diff
@@ -0,0 +1,72 @@
+for(var a = (b in c) in {});
+for(var a = 1 || (b in c) in {});
+for(var a = 1 + (2 || (b in c)) in {});
```
magic-akari, line 3: We have some false positives. It is no longer necessary to add parentheses around the `in` expression that is already in parentheses. 

We need to check the `in` expressions from the outside in. But currently, the `wrap` calls are from the inside out, which cannot be completed in a single pass.

Do you have any thoughts? @kdy1

kdy1: Can we proceed with some false positives for now and refactor them outside? If you prefer, rewriting the fixer to work outside in is also fine.

magic-akari: I plan to leave a TODO for now, as this is a very rare edge case.

kdy1 (approved):
---

swc-bump:
 - swc_ecma_transforms_base
 - swc_core

PR #9210, opened 2024-07-11
fix(es/minifier): Fix compress pow `NaN`

Fnll:
**Description:**

`f64::powf` for `NaN` returns 1 instead of `NaN`, see https://github.com/rust-lang/rust/issues/60468. We should handle that.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/9193


kdy1 (approved):
swc-bump:
 - swc_core
 - swc_ecma_minifier

kdy1:
Thank you!

crates/swc_ecma_minifier/tests/terser/compress/evaluate/pow_nan/input.js
```diff
@@ -0,0 +1 @@
+console.log(Math.pow(1, NaN));
```
magic-akari, line 1: Please add another test case:

```
console.log(Math.pow(NaN, 0));
```

magic-akari, line 1: Please add another test case:

```
console.log(Math.pow(NaN, 0));
```

magic-akari (changes requested):

crates/swc_ecma_minifier/tests/terser/compress/evaluate/pow_nan/input.js
Fnll: sure

Fnll: I added a condition branch for number to the power of 0, which should equal to 1. Is that ok?

magic-akari:
@Fnll Could you sign the CLA?

crates/swc_ecma_minifier/src/compress/util/mod.rs
```diff
@@ -511,18 +511,16 @@ pub(crate) fn eval_as_number(expr_ctx: &ExprCtx, e: &Expr) -> Option<f64> {
                             if args.len() != 2 {
                                 return None;
                             }
-                            let first = eval_as_number(expr_ctx, &args[0].expr)?;
-                            let second = eval_as_number(expr_ctx, &args[1].expr)?;
+                            let base = eval_as_number(expr_ctx, &args[0].expr)?;
+                            let exponent = eval_as_number(expr_ctx, &args[1].expr)?;
 
-                            if second == 0.0 {
```
Fnll: > 2. If exponent is either +0𝔽 or -0𝔽, return 1𝔽.

As step 2 of https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-numeric-types-number-exponentiate, can we simply return 1 instead of trying to evaluate in that condition?

magic-akari: It appears that the calculation path has seen minimal alteration.
We generally believe that low-level optimizations, such as those provided by LLVM, typically surpass the performance of manually added branches.

crates/swc_ecma_minifier/tests/terser/compress/evaluate/pow_spec/output.terser.js
```diff
@@ -0,0 +1 @@
+console.log(1);
```
magic-akari, line 1: @kdy1 Is there a way to update the output.terser.js ?

kdy1: You may update it manually, but being better than `output.terser.js` is fine.

Another way to update `output.terser.js` is replacing `compress.js` in terser test suite with https://github.com/swc-project/swc/blob/181320a5554a1b29a64650bdc2c1097b67fe6e9a/crates/swc_ecma_minifier/scripts/x/terser/compress.js and running the terser test.

crates/swc_ecma_minifier/tests/terser/compress/evaluate/pow_nan/output.js
```diff
@@ -1 +1 @@
-console.log(NaN);
\ No newline at end of file
+console.log(Math.pow(1,NaN));
```
Fnll: It breaks test, is this intentional?
https://github.com/swc-project/swc/actions/runs/9904675638/job/27362648382?pr=9210

magic-akari (approved):

PR #9212, opened 2024-07-11
feat(es/utils): Support for arrays using `cast_to_number`

levi-nz:
**Description:**
This PR allows `ArrayLit`s to be converted to numbers in the `cast_to_number` function. This allows expressions using arrays to be converted to numbers. See some example expressions below that were previously not able to be computed, but are now able to due to this change.

```js
+[] // 0
+[[]] // 0
+[1] // 1
+[undefined] // 0
+[null] // 0
+[[1]] // 1
+[,] // 0
+[,,] // NaN
```

Regarding the implementation, arrays are converted to strings, and the string is then parsed as a number. So arrays like `[]` and `[undefined]` return `""` which then return `0` when parsed as a string. This is also why arrays with more than one element can't be parsed because e.g. `[1, 2]` returns `"1,2"`. This procedure follows the ECMAScript specification.
https://262.ecma-international.org/6.0/#sec-tonumber
https://262.ecma-international.org/6.0/#sec-toprimitive

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_transforms_optimization/src/simplify/expr/tests.rs
+++ b/crates/swc_ecma_transforms_optimization/src/simplify/expr/tests.rs
@@ -489,6 +489,27 @@ fn test_unary_ops_4() {
 }
 
 #[test]
+fn test_unary_ops_5() {
+    // Empty arrays
+    fold("+[]", "0");
+    fold("+[[]]", "0");
+    fold("+[[[]]]", "0");
+    
+    // Arrays with one element
+    fold("+[1]", "1");
+    fold("+[[1]]", "1");
+    fold("+[undefined]", "0");
+    fold("+[null]", "0");
+    fold("+[,]", "0");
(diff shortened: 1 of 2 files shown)


levi-nz:
[ff2749d](https://github.com/swc-project/swc/pull/9212/commits/ff2749d5b693fc4e3beb691e3570fcc0b0233414) might be better for performance, and does the same thing as far as I'm aware, but the current code is more correct.

The performance change might also simply be because of arrays not being casted to numbers before.

kdy1:
No worries, those performance issues are not related to minifier

kdy1 (approved):
swc-bump:
 - swc_ecma_utils
 - swc_core

levi-nz:
Sounds good, also I'm not sure if the return statements are correct regarding purity but I assume you've looked over that already

Issue #9213, opened 2024-07-11
Multiple crashes when compiling javascript/typescript files


qarmin:
### Describe the bug

Using `swc compile` on this files
[Untitled Folder.zip](https://github.com/user-attachments/files/16181285/Untitled.Folder.zip)

causes several crashes/timeouts like
```
timeout: the monitored command dumped core
```
```
thread 'main' panicked at crates/swc_ecma_compat_es2015/src/destructuring.rs:1165:44:
called `Option::unwrap()` on a `None` value
```
```
thread 'main' panicked at /home/runner/work/Automated-Fuzzer/Automated-Fuzzer/swc-main/crates/swc_common/src/source_map.rs:662:60:
byte index 6 is not a char boundary; it is inside '\u{feff}' (bytes 4..7) of `void﻿#!shßpublicb`
```
```
thread 'main' panicked at crates/swc_ecma_compat_es2021/src/logical_assignments.rs:155:50:
called `Result::unwrap()` on an `Err` value: Invalid(Invalid { span: 0..0#0 })
```
```
thread 'main' panicked at crates/swc_ecma_compat_es2015/src/destructuring.rs:409:52:
internal error: entered unreachable code: Object rest pattern should be removed by es2018::object_rest_spread pass
```

### Playground link (or link to the minimal reproduction)

[...]

PR #9215, opened 2024-07-12
fix(es/compat): Add support for destructuring with BigInts

magic-akari:

- #9213


kdy1 (approved):
swc-bump:
 - swc_ecma_compat_es2015
 - swc_core

Issue #8919, opened 2024-05-06
minifier: invalid handling for `var` variable declared in switch statement


hyp3rflow:
### Describe the bug

when there is a `var` declaration in the default case, following case statements referencing x will be minified incorrectly.

### Input code

```typescript
"use strict";
const k = (() => {
switch ("") { 
    default: var x;
    case "": x; 
} 
return x;
})();
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2022",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": true,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true,
        "const_to_let": true,
        "pristine_globals": true
      },
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": false
}
```

### Playground link (or link to the minimal reproduction)

[...]

PR #9220, opened 2024-07-12
fix(es/minifier): Fix variable declaration in default branch

CPunisher:
**Description:**

When a default branch appears before an exact match, the variable declaration should be recorded before deleting the default branch.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/8919


kdy1 (approved):
Thanks!

---

swc-bump:
 - swc_ecma_minifier
 - swc_core

PR #9224, opened 2024-07-12
perf(bindings): Invert Wasm size shrink

kdy1:
Reverts swc-project/swc#9191

I'm reverting because it applies even to `@swc/core`. cc @magic-akari 
Instead, I used an environment variable.

### DIFF
--- a/bindings/Cargo.toml
+++ b/bindings/Cargo.toml
@@ -16,12 +16,11 @@ resolver = "2"
 # lto = "fat"
 
 # debug = true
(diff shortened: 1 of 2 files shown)


kdy1 (commented):
swc-bump:
 - dbg-swc

magic-akari (approved):

Boshen:
In oxc root Cargo.toml:

```
[profile.release.package.oxc_wasm]
opt-level = 'z'
```

PR #9225, opened 2024-07-12
refactor(es/helpers): Remove unnecessary exports

### LINKED ISSUE #9203: Helper methods do not need to be export functions
### Describe the feature

`_` export is enough, and we can rename them while exporting.

_No response_

### Additional context

_No response_

devjiwonchoi:
**Description:**

Helper methods were exporting twice, `_` and it's own function name.
We can map the build script to export `_` as it's own name (filename).
Therefore we export only once as `_`, and map the name on the build script.

```js
// helpers/_foo.js

function _foo() {
  // ...
}

export { _foo as _ }
```
```js
// index.js

// We know the func name will be `_foo` based on the filename.
export { _ as _foo } from '_foo.js'
```

Closes #9203


packages/helpers/esm/_array_without_holes.js
```diff
@@ -1,6 +1,6 @@
 import { _array_like_to_array } from "./_array_like_to_array.js";
```
magic-akari, line 1: All import statements also need to be corrected.

magic-akari (changes requested):

magic-akari (changes requested):
We also need to update code at packages/helpers/scripts/ast_grep.js, ensuring that the generation logic for CJS is correct.

devjiwonchoi:
Oh, yes! Working on it. 😆

magic-akari:
There is some code logic that needs to be referred to at https://ast-grep.github.io. If you need assistance, please let us know.

packages/helpers/scripts/ast_grep.js
```diff
@@ -87,7 +87,7 @@ export function ast_grep() {
                     source.prependLeft(range.start.index, `exports._ = exports.${func_name} = `);
                 });
 
-            const export_shortname = `export { ${func_name} as _}`;
+            const export_shortname = `export { ${func_name} as _ }`;
```
devjiwonchoi, line 90: this seemed to need a space, lmk if it is intended - no error thrown yet.

magic-akari: This does not matter since ast-grep works on the AST, not based on string matching. 
And inserting a space enhances the visual formatting.

devjiwonchoi: Ah, understood.

magic-akari: In fact, `export_alias` and the modified `match` have matched the same sentence. 
So I think we can omit one query.

packages/helpers/scripts/ast_grep.js
```diff
@@ -65,8 +63,10 @@ export function ast_grep() {
                 report_export_mismatch(tree.filename(), match);
             }
 
+            const export_as_lodash_length = `export { ${func_name} as _ }`.length;
```
devjiwonchoi, line 66: @magic-akari Am I good to omit this as well?

```js
task_queue.push(	
  fs.appendFile(tree.filename(), export_shortname, "utf-8"),	
);
```

packages/helpers/scripts/ast_grep.js
```diff
@@ -86,27 +86,14 @@ export function ast_grep() {
...
                 });
-
-            const export_shortname = `export { ${func_name} as _}`;
-
-            const export_alias = tree.root().find(export_shortname);
-
-            if (!export_alias) {
-                task_queue.push(
-                    fs.appendFile(tree.filename(), export_shortname, "utf-8"),
-                );
```
devjiwonchoi, line 97: @magic-akari Am I good to omit this as well?

```js
task_queue.push(	
  fs.appendFile(tree.filename(), export_shortname, "utf-8"),	
);
```

magic-akari: It's OK.

packages/helpers/scripts/ast_grep.js
```diff
@@ -86,27 +86,14 @@ export function ast_grep() {
 
                     source.prependLeft(range.start.index, `exports._ = exports.${func_name} = `);
```
magic-akari, line 87:
```suggestion
                    source.prependLeft(range.start.index, `exports._ = `);
```

There is also a similar statement on line 73.

packages/helpers/scripts/ast_grep.js
```diff
@@ -65,8 +63,10 @@ export function ast_grep() {
                 report_export_mismatch(tree.filename(), match);
             }
 
+            const export_as_lodash_length = `export { ${func_name} as _ }`.length;
+
             const export_start = match.range().start.index;
-            const export_end = export_start + export_lentgh;
+            const export_end = export_start + export_as_lodash_length;
```
magic-akari, line 69: You should directly use the end index of the match.

packages/helpers/scripts/ast_grep.js
devjiwonchoi: I think we can just remove the whole forEach part:

```js
match
  .findAll({
      rule: {
          pattern: func_name,
          kind: "identifier",
          inside: { kind: "assignment_expression", field: "left" },
      },
  })
  .forEach((match) => {
      const range = match.range();
  
      source.prependLeft(range.start.index, `exports._ = exports.${func_name} = `);
  });
```

This is doing the same thing:

```js
source.update(
  export_start,
  export_end,
  `exports._ = exports.${func_name} = ${func_name};`,
);
```

packages/helpers/scripts/ast_grep.js
devjiwonchoi: This is doing same as:

```js
source.prependLeft(range.start.index, `exports._ = exports.${func_name} = `);
```

If I use `match.range().end.index` it throws as:

```
file:///Users/devjiwonchoi/DevLife/forks/swc/node_modules/magic-string/dist/magic-string.es.mjs:761
                if (end > this.original.length) throw new Error('end is out of bounds');
                                                      ^

Error: end is out of bounds
    at MagicString.update (file:///Users/devjiwonchoi/DevLife/forks/swc/node_modules/magic-string/dist/magic-string.es.mjs:761:41)
    at file:///Users/devjiwonchoi/DevLife/forks/swc/packages/helpers/scripts/ast_grep.js:70:20
```

devjiwonchoi: Ah, you're right, sorry.

packages/helpers/scripts/ast_grep.js
devjiwonchoi: I think it can also be just:

```js
`exports._ = ${func_name};`
```

Since we are going to access `_` only?

magic-akari: Function declarations are hoisted. But using function expressions directly might cause issues. This is especially true in cases of self-recursion within the function. You need to verify it.

magic-akari: > I think we can just remove the whole forEach part:
> 
> ```js
> match
>   .findAll({
>       rule: {
>           pattern: func_name,
>           kind: "identifier",
>           inside: { kind: "assignment_expression", field: "left" },
>       },
>   })
>   .forEach((match) => {
>       const range = match.range();
>   
>       source.prependLeft(range.start.index, `exports._ = exports.${func_name} = `);
>   });
> ```
> 
> This is doing the same thing:
> 
> ```js
> source.update(
>   export_start,
>   export_end,
>   `exports._ = exports.${func_name} = ${func_name};`,
> );
> ```

This is used to modify the internal references within a function. Consider the following example.

https://github.com/swc-project/swc/blob/1ab406cd7aa19ea333a8462b0cd496ceb3e39ac1/packages/helpers/esm/_construct.js

```JavaScript
import { _is_native_reflect_construct } from "./_is_native_reflect_construct.js";
import { _set_prototype_of } from "./_set_prototype_of.js";
export function _construct(Parent, args, Class) {
    if (_is_native_reflect_construct()) _construct = Reflect.construct;
    else {
        _construct = function construct(Parent, args, Class) {
            var a = [null];
            a.push.apply(a, args);
            var Constructor = Function.bind.apply(Parent, a);
            var instance = new Constructor();

            if (Class) _set_prototype_of(instance, Class.prototype);

            return instance;
        };
    }

    return _construct.apply(null, arguments);
}
export { _construct as _ };
```

Please note that the exported `_construct` will be updated within the function body.

kdy1 (dismissed):
Thank you!

---

swc-bump: 
 - dbg-swc

kdy1:
I'll merge once @magic-akari approves this PR

packages/helpers/scripts/ast_grep.js
devjiwonchoi: Cool, had the same fix lol. :)

added comment why we find from `tree.root`!

magic-akari:
There are some helpers from the 'tslib', should they be processed in the same way? @kdy1

magic-akari (approved):

kdy1:
Those are fine. My point was that extra exports are useless, but creates extra entries for some tree shaking thing in turbopack.

devjiwonchoi:
@magic-akari Thank you for so detailed, kind review!!

PR #9233, opened 2024-07-14
chore(ci): Use dynamic matrix for `cargo-test` action

kdy1:
**Description:**

We can dynamically create a matrix to make managing CI simpler.


kdy1 (commented):
swc-bump:
 - dbg-swc

PR #9236, opened 2024-07-14
swc_ecma_parser: fix #9227: allow escaped keywords as an object property name

### LINKED ISSUE #9227: swc_ecma_parser: Unexpected escape sequence in reserved word (2)
### Describe the bug

Have swc_ecma_parser parse this test262 case: [test/language/import/import-assertions/json-extensibility-array.js](https://github.com/tc39/test262/blob/main/test/language/import/import-assertions/json-extensibility-array.js) (others also exhibit this behavior).

(This is a re-post of #9214, this time with a proper reproduction link that works on the latest parser version, 0.147.0)

### Input code

```typescript
var obj = ({ bre\u0061k: 42 });
```

### Config

```json
(I'm only using swc_ecma_parser in my toy/learning JS impl; I'm not using the full swc and I don't have an .swcrc)

fwiw:
Syntax::Es(Default::default())
EsVersion::Es2015
```

https://play.swc.rs/?version=1.6.7&code=H4sIAAAAAAAAAytLLFLIT8pSsFXQqFZIKkqNKTUwMDPMtlIwMVKo1bTmAgDLr8KQIAAAAA%3D%3D&config=H4sIAAAAAAAAA1VPSQ7CMAy89xWRzxwACQ78gUdYwa1SZVMcJKKqfyfNUujNs3nsZRACZpbwEEseM%2FAYmMKOM8PJRvxkBkgaZBmUj3Dq6sybNKJmKtRaFYgYJoolxdfz5dYSoJ1j6onGGWXVmP47pTM%2BEPPRuFnRTpqOjUNrBeNe7yK2X2LyVC%2B4w8%2FUy%2FbFoPjZk3Xt%2BgVne8%2FLGAEAAA%3D%3D

### SWC Info output

_No response_

### Expected behavior

The above code is parsed as an object literal with property "var" (which is legal JS). This works in node.

### Actual behavior

The following error is reported:

```
error: Unexpected escape sequence in reserved word: var
  --> <anon>:37:14
   |
37 | var obj = ({ v\u0061r: 42 });
   |              ^^^^^^^^
```

### Version

0.147.0

### Additional context

The relevant code that calls the parser is the following (just an extract; the original with full context is [here](https://github.com/sebastiano-barrera/mcjs/blob/1b2c4fdbf05e2da94179ce9508c0786604eb7c9a/mcjs_vm/src/bytecode_compiler.rs#L106)):

```js
use swc_ecma_ast::EsVersion;
use swc_ecma_parser::{lexer::Lexer, Parser, StringInput, Syntax};

// ...
// content: String
let source_file = source_map.new_source_file(swc_common::FileName::Anon, content);
let input = StringInput::from(source_file.as_ref());

// ...

let lexer = Lexer::new(
    Syntax::Es(Default::default()),
    EsVersion::Es2015,
    input,
    None,
);

// ...

let script_ast = parser
    .parse_script()
    .map_err(|e| {
        e.into_diagnostic(&err_handler).emit();
        error!("parse error")
    });
```

sebastiano-barrera:
**Description:**

With this PR, the lexer error `SyntaxError::EscapeInReservedWord` is un-done at the parser level if the reserved words appears as an object property name.

The bug in #9227 only happens because the first "peek" in the lexer happens when context has `in_property_name` that's still _false_, so there is no chance to disambiguate. (And I couldn't be sure that a change to this call didn't affect other cases; JavaScript's grammar is quite varied.)

[NOTE: I just noticed that there are a bunch more cases where \u escape codes should be allowed but aren't. This PR wont' be sufficient, adn the final fix will probably need a completely different approach.]
  
**This is a draft!**

I'm posting it to get some feedback on whether this is going in the right direction.

There are also a few open points:

- [ ] _Negative tests_. I'd like to add a negative test (expected to fail) under _tests/js_. Is this even possible?
- [ ] _TODO: obj.v\u0061r_. Feature: apply the same fix to property names in positions other than object literals (e.g. `obj.\u0063ontinue`)
- [ ] _"Unget token"_. Should we refactor the "replace token" logic that's currently implemented with p.input.cur() then p.input.store()? Something like `p.input.unget(token)`
- [ ] _More general design_. The bug is roughly in the pattern: "lex with ctx A; then start parsing with slightly different ctx B", but an error in ctx A is not an error in B. Is it worth it to make a more general change? I can think of:
  1. make the lexer fully independent of context (probably not practical)
  1. add a mechanism where the lexer can detect that an error was due to ctx A and the context has changed since, so the lexing is repeated with ctx B. Not sure how hard this is.
 
**Related issue (if exists):** #9227


crates/swc_ecma_parser/src/error.rs
```diff
@@ -83,7 +83,7 @@ pub enum SyntaxError {
     UnterminatedStrLit,
     ExpectedUnicodeEscape,
     EscapeInReservedWord {
-        word: JsWord,
```
sebastiano-barrera, line 86: This is done so that, in addition to printing the word in an error message, the whole `Token` that caused the error can be 'put back into the parser'.

crates/swc_ecma_parser/src/parser/object.rs
```diff
@@ -51,6 +51,23 @@ impl<I: Tokens> Parser<I> {
...
+                    let err_tok = err_tok.clone();
+                    let check = p.input.bump();
+                    debug_assert_eq!(err_tok, check);
+                    eprintln!(
+                        "  :: parser.object.parse_prop_name: recovered error {:?} -> {:?}",
+                        err_tok, word
+                    );
+                    p.input.store(Token::Word(word));
+                }
+            }
```
sebastiano-barrera, line 69: This entire logic could be refactored into a method in `Buffer`, something like `p.input.unget` (reusing the from the C; or maybe "unread" like in Go)

crates/swc_ecma_parser/src/parser/object.rs
```diff
@@ -51,6 +51,23 @@ impl<I: Tokens> Parser<I> {
         .parse_with(|p| {
             let start = cur_pos!(p);
 
+            // The lexer marks keywords with escapes as errors, but unlike in
+            // the general case, it's OK here as a property name. We turn the
+            // error back into a token
+            if let Some(err_tok @ Token::Error(e)) = p.input.cur() {
```
kdy1, line 57: This is the wrong direction. We should adjust `ctx` to ignore the error at the first place.

crates/swc_ecma_parser/src/error.rs
```diff
@@ -83,7 +83,7 @@ pub enum SyntaxError {
...
+        word: Word,
```
kdy1, line 86: This change makes this PR a breaking change, so I think it's better to revert this.

crates/swc_ecma_parser/src/parser/object.rs
sebastiano-barrera: If by "adjusting ctx" you mean adding `with_ctx({ in_property_name: true, ... })` at the call site of a parser, I had initially thought it could have been too coarse... I need to try again.

kdy1: I think adding a field for ignoring escapes in keywords to the context type is the best solution

kdy1:
Closing as stale.

PR #9240, opened 2024-07-15
perf(allocator): Drop `scoped_tls`

kdy1:
**Description:**

The performance is much better without it.


kdy1 (commented):
swc-bump:
 - swc_allocator
 - swc_core

PR #9247, opened 2024-07-15
test: Enable scoped API by default

kdy1:
**Description:**

This PR is to profile real performance using codspeed


kdy1 (commented):
swc-bump:
 - swc_allocator
 - swc_core
 - swc_common

PR #9252, opened 2024-07-15
fix(allocator): Remove wrong assertions and add tests

kdy1:
**Description:**


kdy1 (commented):
swc-bump:
 - swc_allocator
 - swc_core

PR #9254, opened 2024-07-15
refactor(allocator): Use RAII guard instead of `scope`

kdy1:
**Description:**

**Related issue:**

- https://github.com/swc-project/swc/issues/9253


kdy1 (commented):
swc-bump:
 - swc_allocator
 - swc_core

Issue #9184, opened 2024-07-09
compress optimize  escape `\\` in template string


stormslowly:
### Describe the bug

expect the output should be
`(foo - \*foo)`  or  `(bar - \*bar)`

but 

`(foo - *foo)`  or  `(bar - *bar)`

### Input code

```typescript
let pi= Math.random() >0.6 ? "foo": "bar";
let fi=`(${`${pi}`} - ${`\\*${pi}`})`

console.log(fi)
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2020",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": true,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true,
        "const_to_let": true,
        "pristine_globals": true,
        "passes": 8
      },
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link (or link to the minimal reproduction)

[...]

PR #9261, opened 2024-07-16
fix(es/minifier): Collect raw str values for new Tpl element

stormslowly:
**Description:**
convert `raw` value  to `cooked` for creating new Tpl element.

<del>
while optimizing nested tpl, after concatenating some cooked strs to assemble a new tpl element,
the `raw` of the tpl should be escaped cooked string value.
</del>

**BREAKING CHANGE:**

None

https://github.com/swc-project/swc/issues/9184

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/pure/strings.rs
+++ b/crates/swc_ecma_minifier/src/compress/pure/strings.rs
@@ -127,8 +127,8 @@ impl Pure<'_> {
         for idx in 0..(tpl.quasis.len() + tpl.exprs.len()) {
             if idx % 2 == 0 {
                 let q = tpl.quasis[idx / 2].take();
-
-                cur_cooked_str.push_str(q.cooked.as_deref().unwrap_or(&*q.raw));
+                
+                cur_cooked_str.push_str(&Str::from_tpl_raw(&q.raw));
                 cur_raw_str.push_str(&q.raw);
             } else {
                 let mut e = tpl.exprs[idx / 2].take();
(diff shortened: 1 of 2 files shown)


stormslowly:
@kdy1  Oops, a compress case failed, i am trying to fix

crates/swc_ecma_minifier/tests/exec.rs
```diff
@@ -11325,3 +11325,13 @@ fn issue_9010() {
         "#,
     );
 }
+
+#[test]
+fn issue_9184() {
+    run_default_exec_test(
+        r#"
+        let pi= Math.random() >1.1 ? "foo": "bar";
```
kdy1, line 11333: Can we have one more case for the reverse case? (`Math.random() < -1`)

stormslowly: I am not quite understand why adding the reverse case.
IMO, the key of this case is about  purify the nested tpl string. 
The `Math.random()` is just for the `pi` variable not being to replaced by its literal value in the tpl string.

kdy1: Ah, ok

crates/swc_ecma_minifier/src/compress/pure/strings.rs
```diff
@@ -121,13 +121,15 @@ impl Pure<'_> {
...
-        let mut cur_str_value = String::new();
+        let mut cur_cooked_str = String::new();
+        let mut cur_raw_str = String::new();
 
         for idx in 0..(tpl.quasis.len() + tpl.exprs.len()) {
             if idx % 2 == 0 {
                 let q = tpl.quasis[idx / 2].take();
 
-                cur_str_value.push_str(q.cooked.as_deref().unwrap_or(&*q.raw));
+                cur_cooked_str.push_str(q.cooked.as_deref().unwrap_or(&*q.raw));
```
kdy1, line 131: I think this logic is wrong. `cooked` or `raw` should never be the case, and we should run tpl string replace logic for `raw` if we want to use it as a `value`.

See https://rustdoc.swc.rs/swc_ecma_ast/struct.Str.html#method.from_tpl_raw

stormslowly: updated

kdy1 (approved):
Thanks!

---

swc-bump:

Issue #9101, opened 2024-06-24
Publish pre-codegened version of AST/visitor crates


kdy1:
### Describe the feature

I'm not sure about the way, but [`syn_codegen`](https://github.com/dtolnay/syn/blob/ecb0429297d3da2f20ab8559e41ddf8a2137d83b/codegen/README.md) looks like a good example

PR #9262, opened 2024-07-16
refactor(visit): Pre-generate visitor code

kdy1:
**Description:**

This would improve compile time by a margin, and this is a refactoring that should be done before applying `swc_allocator::boxed::Box` to ES AST.

This PR removes `swc_visit_macros` because we now use a simple binary to generate code before publishing.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/9101


kdy1 (commented):
swc-bump:
 - swc_visit --breaking

PR #9264, opened 2024-07-16
feat(es/testing): Parse test code as a `Program` instead of a `Module`

### LINKED ISSUE #8713: `test_transform` should probably use `parse_program` instead of `parse_module`
### Describe the feature

The problem right now with `test_transform` is that, because it uses `parse_module` to parse code internally, visitors that use `visit_script`/`visit_mut_script`/`fold_script` will not work because these functions are never called due to the parsed code being a module (thanks to `apply_transform`): https://github.com/swc-project/swc/blob/b0c00b96f18c391bea89059130d08484e3ac3506/crates/swc_ecma_transforms_testing/src/lib.rs#L175

To fix this, we could make `test_transform` use `parse_program` instead, but this might introduce breaking changes. A `Program` can be either a `Module` or `Script`, but if we make `test_transform` use `parse_program` it might break existing unit tests (1st or 3rd party) that use `visit_module` instead of `visit_program` (or the other two functions). Some of the code related to this would also need to be changed, but that doesn't seem too big of an issue.

Overall, making `test_transform` would make people's lives a bit easier when making visitors that use specific functions like `visit_script` if they know they will never need to use the program or module functions, but I'm curious what the maintainers think about this as I'm not sure if I'm right.

_No response_

### Additional context

_No response_

leafypout:
**Description:**
This PR addresses the issue described in #8713 

**BREAKING CHANGE:**
This will break existing unit tests that use `fold_module`/`visit_module`/`visit_mut_module` if the visitor is intended to work for both modules and scripts, instead of using `fold_program`/`visit_program`/`visit_mut_program`. This will also break existing unit tests if they're testing with input code that gets parsed as a script in `parse_program` if the visitor expects a module (they will need to update their `test!` calls to add `module` as the first argument, or use a function like `apply_module_transform`)

- Closes #8713

### DIFF
--- a/crates/swc_ecma_transforms/tests/decorators.rs
+++ b/crates/swc_ecma_transforms/tests/decorators.rs
@@ -91,6 +91,7 @@ fn transformation(t: &Tester) -> impl Fold {
 
 // transformation_declaration
 test!(
+    module,
     syntax(false),
     |t| transformation(t),
     transformation_declaration,
@@ -101,6 +102,7 @@ class A {}
 );
 // transformation_initialize_after_super_multiple
 test!(
+    module,
     syntax(false),
     |t| transformation(t),
     transformation_initialize_after_super_multiple,
@@ -134,6 +136,7 @@ export default @dec() class {}
 );
 // transformation_initialize_after_super_statement
 test!(
+    module,
     syntax(false),
     |t| transformation(t),
     transformation_initialize_after_super_statement,
@@ -211,6 +214,7 @@ expect(A).toBe(C);
 );
 // misc_method_name_not_shadow
 test!(
+    module,
     syntax(false),
     |t| tr(t),
     misc_method_name_not_shadow,
@@ -395,6 +399,7 @@ expect(log).toEqual(numsFrom0to9);
 );
 // transformation_initializer_after_super_bug_8808
 test!(
+    module,
     syntax(false),
     |t| transformation(t),
     transformation_initiailzer_after_super_bug_8808,
@@ -434,6 +439,7 @@ expect(A.prototype.method()).toBe(2);
 );
 // transformation_arguments
 test!(
+    module,
     syntax(false),
     |t| transformation(t),
     transformation_arguments,
@@ -532,6 +538,7 @@ expect(calls).toBe(1);
 // ordering
 // transformation_initialize_after_super_expression
 test!(
+    module,
     syntax(false),
     |t| transformation(t),
     transformation_initialize_after_super_expression,
@@ -914,6 +921,7 @@ test!(
 // finishers
 // transformation_extends_await
 test!(
+    module,
     syntax(false),
     |t| transformation(t),
     transformation_extends_await,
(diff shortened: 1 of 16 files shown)


levi-nz:
Several swc tests rely on the parsed test code being a module due to their visitors, going to have to fix these. The good news is that this will fix any of these issues for people using `parse_script`. e.g. some of the simplifier tests fail on scripts but not on modules.

levi-nz:
I will probably clean up the internal API a bit because I think it's a bit messy, but essentially how testing works now:

- Current `test` implementations will continue to work, unless the visitor(s) being tested is implemented wrongly. Since `parse_program` is the default now and not `parse_module`, any visitors using `visit_module` instead of `visit_program` won't work if the parsed AST ends up being a script (because `visit_module` won't be called, only `visit_script` will). Visitors should use `visit_program` if they want to handle both modules and scripts.
- Callers can force the test implementation to use `parse_module` or `parse_script` instead of `parse_program` by calling `test!(module, ...)` and `test!(script, ...)` respectively.
- Some tests need to use `test!(module, ...)` as mentioned above, since their visitors depend on the code being a module. A good example are the fixture tests in swc_ecma_transforms_module.

kdy1:
Sounds good, thank you!

levi-nz:
Not sure what we think of the internal api, `test!(module, ...)`, `test!(script, ...)`, `apply_module_transform` etc.

crates/swc_ecma_transforms_proposal/tests/decorators.rs
```diff
@@ -10,7 +10,7 @@ use swc_common::{chain, comments::SingleThreadedComments, Mark};
 use swc_ecma_parser::{EsSyntax, Syntax, TsSyntax};
 use swc_ecma_transforms_base::{assumptions::Assumptions, resolver};
 use swc_ecma_transforms_proposal::{decorator_2022_03::decorator_2022_03, DecoratorVersion};
-use swc_ecma_transforms_testing::{test_fixture, FixtureTestConfig};
+use swc_ecma_transforms_testing::{test_module_fixture, FixtureTestConfig};
```
kdy1, line 13: Can we have `FixtureTestConfig.module: Option<bool>`, which defaults to `None`, and 

 - `Some(true)`: Module (and applies visitor after wrapping as Program)
 - `Some(false)`: Script (and applies visitor after wrapping as Program)
 - `None`: Program

crates/swc_ecma_transforms/tests/decorators.rs
```diff
@@ -96,6 +96,7 @@ fn transformation(t: &Tester) -> impl Fold {
 
 // transformation_declaration
 test!(
+    module,
```
kdy1, line 99: I love this idea 👍

kdy1 (approved):
Thank you!

levi-nz:
@kdy1 can you un-merge? Not finished yet. Forgot to mark as draft

levi-nz:
I've pushed my changes (should be good to go now), but I'll need to rebase once you revert the commit.

kdy1:
@levi-nz Done. https://github.com/swc-project/swc/pull/9621

PR #9265, opened 2024-07-17
refactor(allocator): Feature gate `nightly`

kdy1:
**Description:**

Feature gate `nightly`, and make it **really identical to types from `std`** if it's not enabled.

We need to use `#[fundamental]` for `swc_allocator::boxed::Box`. Without it, it will be very hard to use the type. But it's not stabilized. So we feature gate them under a cargo feature flag.

**Related issue:**

 - Blocks https://github.com/swc-project/swc/pull/9262


kdy1:
This was the wrong way. We may need to drop support for stable rustc

PR #9272, opened 2024-07-18
fix(es/typescript): Workaround `wasm-bindgen`

magic-akari:
> [!NOTE]  
> We should remove this if https://github.com/rustwasm/wasm-bindgen/pull/4015 get mereged.


kdy1 (approved):
Thanks!

---

swc-bump:
 - dbg-swc

PR #9274, opened 2024-07-18
feat(allocator): Feature gate `nightly` via macros

kdy1:
**Description:**

We can make swc faster without dropping support for stable rustc.


kdy1 (commented):
swc-bump:
 - swc_allocator
 - swc_core
 - swc_common

PR #9281, opened 2024-07-19
perf(allocator): Use `std` instead of `allocator-api2`

kdy1:
**Description:**

`std` is faster.


kdy1 (commented):
swc-bump:
 - swc_allocator

Issue #9283, opened 2024-07-19
Add ESM exports to `@swc/wasm-typescript`


kdy1:
### Describe the feature

I'll do this this weekend.

https://github.com/swc-project/swc-playground/issues/42#issuecomment-2238481433

Requested by @g-plane

magic-akari:
maybe related
- https://github.com/rustwasm/wasm-bindgen/issues/3790

PR #9286, opened 2024-07-19
feat(es/typescript): Add esm build for fast ts strip

magic-akari:

- Closes: #9283

### DIFF
--- a/.github/workflows/publish-npm-package.yml
+++ b/.github/workflows/publish-npm-package.yml
@@ -588,12 +588,18 @@ jobs:
           - crate: "binding_core_wasm"
             npm: "@swc\\/wasm"
             target: nodejs
+            out: "pkg"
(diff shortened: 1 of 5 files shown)


magic-akari:
> How will this work while publishing? I think this will add `esm` Wasm binary to the original package, right? I think it may cause an issue for node.js team due to the size of the binary. Can we have another entry at
> 
> https://github.com/swc-project/swc/blob/c1cd0b99c14b03c250f2d278f10480da733e0dfa/.github/workflows/publish-npm-package.yml#L587-L596
> 
> instead?

I agree.

magic-akari:
I tweaked the CI logic.
This is the most troublesome because I don't know if it will work the first time.

kdy1 (approved):
I'll try using nightly

PR #9303, opened 2024-07-20
perf(es/ast): Use `swc_allocator`

kdy1:
**Description:**


kdy1:
I'll retry this after merging https://github.com/swc-project/swc/pull/9313

PR #9313, opened 2024-07-22
perf(es/parser): Rewrite parser

kdy1:
**Description:**

I'm going to rewrite the ES parser of SWC because

 - I didn't know anything about compiler or parser theories at the time of writing the parser (#1). 
 - @Boshen and oxc team investigated a lot and proved that the parser can be much faster than now. (Wonderful job, it's really nice)


Boshen:
Please be aware in oxc, there is a second ast pass for ast path, symbols and scopes to complete all syntax errors.

We also use the latest conformance suite from https://github.com/tc39/test262, you may need to consider replacing the outdated one you are using.

d3lm:
Would be cool if there was a way for it to also produce a red green syntax tree to allow modifications to the code that preserves formatting.

@Boshen Any thoughts on this?

I personally have a lot of use cases where I need to parse some module and then transform the code. The problem is that with an AST, since it's lossy in terms of formatting, any transformations to the tree means that if you re-print it in the end you'd loose the formatting of the module.

Boshen:
> Would be cool if there was a way for it to also produce a red green syntax tree to allow modifications to the code that preserves formatting.
> 
> @Boshen Any thoughts on this?
> 
> I personally have a lot of use cases where I need to parse some module and then transform the code. The problem is that with an AST, since it's lossy in terms of formatting, any transformations to the tree means that if you re-print it in the end you'd loose the formatting of the module.
> 
> It'd be cool if it used a lossless syntax tree instead of an AST.

CST is out of scope for swc / oxc. Take a look at https://crates.io/crates/biome_parser for CST.

d3lm:
Yea I was aware of Biome's parser that's inspired by the Rust Analyzer's parser I guess. Thanks!

Issue #9295, opened 2024-07-19
`@swc/wasm-typescript` handling `() => <T>{}`


acutmore:
### Describe the bug

TypeScript's (legacy) prefix style type assertions (`<Type>val` instead of `(val as Type)`) cannot always be safely removed in `"strip-only"` mode when they are at the start of the body of an arrow function.

### Input code

```typescript
import * as swc from "@swc/wasm-typescript";

const input = `(() => <any>{})`;
const output = swc.transformSync(input, { mode: "strip-only" });

console.log(output.code);
```

### Config

```json
{ "mode": "strip-only" }
```

### Playground link (or link to the minimal reproduction)

no playground support for wasm-typescript yet: https://github.com/swc-project/swc-playground/issues/42

### Expected behavior

Either error or (if possible) produce JS with matching semantics as the input i.e. the equivalent of: `() => ({})`;

### Actual behavior

Resulting code is: `(() =>      {})` which has different semantics to the input.

### Version

@swc/wasm-typescript@1.7.0

### Additional context

I _think_ it's not possible to solve this in all cases. So it might be best to error for arrow functions that start with `<Type>{`.

There are some cases that can work with `"strip-only"`:

```typescript
let f = () => <T>{};
```

can strip and keep the JS in place with:

```javascript
let f = () => 0||{};
```

But that approach does not work for:

```typescript
let f = () => <T>{p:null}.p ?? g();
```

which could not be changed to:

```javascript
let f = () => 0||{p:null}.p ?? g();
```

As it is invalid to have `||` and `??` at the same level.

I would be very interesting if there was a way to solve this, otherwise I think the solution is to error.

kdy1:
> TypeScript's (legacy) prefix style type assertions (<Type>val instead of (val as Type)) cannot always be safely removed in "strip-only" mode when they are at the start of the body of an arrow function.


That's why we don't touch it.

acutmore:
> That's why we don't touch it.

Hi @kdy1, thanks for taking a look. In my testing it looks like `"strip-types"` does remove it.

`(() => <any>{})()` becomes `(() =>      {})`. The first returns `{}` at runtime, the second returns `undefined`.

[...]

PR #9328, opened 2024-07-24
fix(es/typescript): Preserve type assertions

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/9295


crates/swc_fast_ts_strip/src/lib.rs
```diff
@@ -777,9 +777,10 @@ impl Visit for TsStrip {
         self.add_replacement(n.span);
     }
 
+    /// We do not strip type assertion because it's not safe.
+    ///
+    /// See https://github.com/swc-project/swc/issues/9295
     fn visit_ts_type_assertion(&mut self, n: &TsTypeAssertion) {
-        self.add_replacement(span(n.span.lo, n.expr.span().lo));
-
```
magic-akari, line 782: Should we display an error message? 

```
The angle-bracket syntax for type assertions, `<T>expr`, is not supported in type strip mode. Instead, use the 'as' syntax: `expr as T`.
```

kdy1: Sounds good. I added one.

magic-akari (approved):

PR #9342, opened 2024-07-29
fix(es/codegen): Emit question token for class methods

dsherret:
I'm in the process of upgrading swc and stumbled on this one.

**Description:** Regression in emitting class methods that are optional.

### DIFF
--- a/crates/swc_ecma_codegen/src/lib.rs
+++ b/crates/swc_ecma_codegen/src/lib.rs
@@ -1599,6 +1599,10 @@ where
             }
         }
 
+        if n.is_optional {
+            punct!("?");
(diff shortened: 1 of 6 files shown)


crates/swc_ecma_codegen/src/lib.rs
```diff
@@ -1599,6 +1599,10 @@ where
             }
         }
 
+        if n.is_optional {
+            punct!("?");
+        }
```
dsherret, line 1604: I think this regressed because previously it happened to emit from the binding ident being optional.

magic-akari (approved):

Issue #8269, opened 2023-11-11
passing `let let` to the parser should throw an error 


sapphi-red:
### Describe the bug

`let` is reserved in `let`/`const`/`class` declarations in non-strict mode, but SWC parser doesn't reject them.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#:~:text=(also%20reserved%20in%20const%2C%20let%2C%20and%20class%20declarations)

### Input code

```typescript
{
    let let
}
{
    const let = 0
}
{
    class let {}
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.3.97&code=H4sIAAAAAAAAA6vm5VIAgpzUEhDm5arl5aqGCCXn5xWDBRVsFQxQJHISi4vBEtW1YHEAdYxiO0QAAAA%3D&config=H4sIAAAAAAAAA1VPOw7DIAzdcwrkuUPF0CF36CEs6kRE%2FISpVBTl7iUB0mbz%2B%2FnZ6yAELKxgFGsZCwgYmeKJC8PZJfwUBkhZZBV1SHDr6sK7NKFhOqitKpAwzpSOFMu7lC0Bxnumnmic1U5P%2Bb9TeRsiMV%2BNuxXdbOjaOLRWsP71PsT2S8qB6gUP%2BJl62bkYND97sq7dvnp7734YAQAA

### Expected behavior

Throws an error like `Unexpected reserved word 'let'`.

### Actual behavior

No error happens.

### Version

1.3.97

### Additional context

This errors with most parsers on https://astexplorer.net/.

kdy1:
You selected `Source Type: Script`.

SWC correctly emits an error if the type is `Module`.

[...]

PR #9343, opened 2024-07-29
fix(es/parser): Fix `let` parsing in scripts to match spec

zrosenbauer:
**Description:**

Fix for #8269 

WIP - still need to tweak tests & verify running as `strict: false`

**BREAKING CHANGE:**


mischnic:
This is fixed now by https://github.com/swc-project/swc/pull/9484

Issue #9368, opened 2024-08-01
compilation doesn't omit import that is not used in runtime if typescript alias is created from imported type


layerok:
### Describe the bug

compilation doesn't omit import statement that is not used in runtime if typescript alias is created from imported type

### Input code

```typescript
import {Schemas} from './types.d';
import AddressResource = Schemas.AddressResource;
```

### Config

```json
{}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.7.4&code=H4sIAAAAAAAAA8vMLcgvKlGoDk7OSM1NLK5VSCvKz1VQ19MvqSxILdZLUbfm5cqEqHFMSSlKLS4OSi3OLy1KTlWwVYBq0kOTsQYAWJI5D1UAAAA%3D&config=H4sIAAAAAAAAA1VPOw7DIAzdOQXy3CHK0KF36CEQdSIqfsKOVBTl7oUE0maz38fveRVSwps0PORaxrJElQjTuReEsmf1KQhwjkg6mchw6yxTpSZlCXdoOxhglWbk6kIah3FoDrAhEHZHw5zxZsr%2FmTq4mJDoKqxS5WeL10TRUsGF17KT7Zfa92hwh5%2Boh52HwdCzOzktKLYvIO%2B2AhcBAAA%3D

### SWC Info output
<details>
  <summary><code>npx -y swc-info@latest</code></summary>
  
```
    Operating System:
        Platform: darwin
        Arch: arm64
        Machine Type: arm64
        Version: Darwin Kernel Version 22.6.0: Wed Jul  5 22:22:52 PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T8103
        CPU: (8 cores)
            Models: Apple M1

    Binaries:
        Node: 20.0.0
        npm: 9.6.4
        Yarn: 1.22.19
        pnpm: 9.4.0

    Relevant Packages:
        @swc/core: 1.3.89
        @swc/helpers: N/A
        @swc/types: 0.1.12
        

    SWC Config:
        output: N/A
        .swcrc path: N/A

    Next.js info:
        output: N/A
        
```
</details>

### Expected behavior

import statement that is unsused in runtime is omitted 

### Actual behavior

import statement that is unsused in runtime is preserved

### Version

1.7.4

### Additional context

I'm using swc in my Vite project. After updating swc to version 1.3.89 I started seeing this error in the browser console.
```
[...]

PR #9369, opened 2024-08-01
fix(es/typescript): Analyze import chain

magic-akari:
- Closes: #9368

### DIFF
--- a/crates/swc_ecma_transforms_typescript/src/strip_import_export.rs
+++ b/crates/swc_ecma_transforms_typescript/src/strip_import_export.rs
@@ -1,6 +1,6 @@
 use std::mem;
 
-use swc_common::collections::AHashSet;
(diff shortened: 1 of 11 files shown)


crates/swc_ecma_transforms_typescript/src/strip_import_export.rs
```diff
@@ -101,6 +113,25 @@ impl UsageCollect {
...
     }
+
+    fn analyze_import_chain(&mut self) {
+        if self.import_chain.is_empty() {
+            return;
+        }
+
+        let mut new_usage = AHashSet::default();
+        for id in &self.id_usage {
+            let mut id = id.clone();
```
magic-akari, line 124: Is there any way to avoid cloning?

kdy1 (approved):

PR #9388, opened 2024-08-07
fix(visit): Fix method signature for `Box<T>` and `Option<T>`

kdy1:
**Description:**


kdy1:
I think this is overkill

PR #9399, opened 2024-08-08
fix(es/typescript): Strip class modifiers

magic-akari:
**Description:**

**BREAKING CHANGE:**


crates/swc_fast_ts_strip/src/lib.rs
```diff
@@ -441,6 +442,59 @@ impl TsStrip {
...
+                }
+                Token::Word(Word::Ident(IdentLike::Known(
+                    KnownIdent::Readonly
+                    | KnownIdent::Public
+                    | KnownIdent::Protected
+                    | KnownIdent::Private,
+                ))) => {
+                    self.add_replacement(*span);
+                }
+                Token::Word(Word::Ident(IdentLike::Other(o))) if *o == "override" => {
```
magic-akari, line 489: Adding `override` to the `KnownIdent` is likely a wise move.

kdy1: I think we should add it

magic-akari: Let's add it to #9313.

kdy1: Sounds good 👍

kdy1 (approved):
I'll merge once the publish pipeline ends

kdy1:
I found that we can include this patch in 1.7.8

magic-akari:
The ts-type strip still needs one or two additional fix.

magic-akari:
> I found that we can include this patch in 1.7.8. Can you fix the CI failures?

patch-section.sh removed in c977c2bec2b0f9771031ed6ebc15bffb0f33eae0.
How to fix it?

kdy1:
You can do `cargo install cargo-link2` and do `cargo link ../` from `./bindings`. Sorry, I forgot documenting it.

magic-akari:
> error: the `#[global_allocator]` in swc_malloc conflicts with global allocator in: swc_malloc

The CI process has failed due to a conflict with swc_malloc. Could you help me? @kdy1

kdy1:
I'll take on this

kdy1 (approved):
I restored `patch-sections.sh` for now. Sorry for my mistake and I'll fix the bug of `cargo-link` after publishing this change.

PR #9417, opened 2024-08-11
chore: Mention swc in the license

marco-ippolito:
This change adds mention of SWC in the license otherwise it would be anonymous (and odd 😄 )


kdy1 (approved):
🤣

Issue #8944, opened 2024-05-11
SWC Core throws / prints (misleading) error


jd-carroll:
### Describe the bug

We are receiving hundreds of errors in our console using TSUP + @swc/core:
```
  ERROR  failed to read input source map: failed to parse inline source map url
index.js.map
Caused by:
    relative URL without a base
    at /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/swc-0.273.27/src/lib.rs:385
```

These errors seem to be thrown when the corresponding `.js.map` file is missing from the bundled sources. 

It would be nice if `@swc/core` printed a more meaningful error message.

Note: I have no idea how to produce a repro, but I've spent a fair bit of time debugging this with console.log and I am relatively confident that this is the case.

### Playground link (or link to the minimal reproduction)

https://github.com/egoist/tsup/blob/00188a0dc848c48fac45de245d1e021f370a84a3/src/esbuild/swc.ts#L31-L49

### Expected behavior

If an error is to be thrown / printed, print something more descriptive to the issue as well as a little more context.

### Version

1.5.5

### Additional context

Possibly related:
#8879
#8910
https://github.com/swc-project/swc/issues/8789#issuecomment-2105055772

Less likely related but will link anyways:
#8628

jd-carroll:
@davidmurdoch I am seeing the same behavior related to sentry. The issue is (of course) `opentelemetry`. A number of the published modules list source maps, such as:
```
//# sourceMappingURL=internal-types.js.map
```

But inspecting the installed module, the source maps do not exist.

kdy1:
Closing as duplicate

jd-carroll:
@kdy1 duplicate to what?

kdy1:
https://github.com/swc-project/swc/issues/8789

Issue #8910, opened 2024-05-03
Getting a ton of "relative URL without a base".


atrofimov-sc:
### Describe the bug

My team uses @swc/jest. We regenerated our package-lock.json and now we get thousands of these errors in the console when running tests:

```
  ERROR  failed to read input source map: failed to parse inline source map url
index.js.map

Caused by:
    relative URL without a base
    at /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/swc-0.273.28/src/lib.rs:386
```

### Input code

```typescript
N/A We didn't do anything weird. 

Our jest config:


module.exports = {
  // Makes Jest transpile TS/X files into JS/X files before executing them.
  transform: {
    // SWC config for being able to parse ts files.
    '^.+\\.(t|j)s$': [
      '@swc/jest',
      {
        sourceMaps: true,

        jsc: {
          target: 'es2020',
          parser: {
            syntax: 'typescript',
            tsx: true,
          },
        },
      },
    ],
    // SWC config for being able to parse tsx files.
    '^.+\\.(t|j)sx$': [
      '@swc/jest',
      {
        sourceMaps: true,

        jsc: {
          target: 'es2020',
          parser: {
            syntax: 'typescript',
            tsx: true,
          },

          transform: {
            react: {
              runtime: 'automatic',
            },
          },
        },
      },
    ],
  },
};
```
```


### Config

```json
N/A
```

### Playground link (or link to the minimal reproduction)

N/A

### SWC Info output

    Operating System:
        Platform: darwin
        Arch: x64
        Machine Type: x86_64
        Version: Darwin Kernel Version 23.4.0: Fri Mar 15 00:11:05 PDT 2024; root:xnu-10063.101.17~1/RELEASE_X86_64
        CPU: (12 cores)
            Models: Intel(R) Core(TM) i7-8700B CPU @ 3.20GHz

    Binaries:
        Node: 20.12.2
        npm: 10.5.0
        Yarn: N/A
        pnpm: N/A

    Relevant Packages:
        @swc/core: 1.4.17
        @swc/helpers: 0.5.11
        @swc/types: 0.1.6
        typescript: 5.4.5

    SWC Config:
        output: N/A
        .swcrc path: N/A

    Next.js info:
        output: N/A

### Expected behavior

Don't throw errors

### Actual behavior

Throws errors

### Version

@swc/jest@0.2.36

### Additional context

When I set `sourceMaps: false` I see the same errors.

PR #9422, opened 2024-08-12
fix(es): Improve sourcemap url error messages.

simonbuchan:
**Description:**

Parse based on the provided prefix: inline urls must start with "data:", so only trying one will ensure the correct and more helpful error message will be shown.

Currently, any error while trying to load a sourcemap file will give the unhelpful error:

```
  ERROR  failed to read input source map: failed to parse inline source map url
index.js.map

 Caused by:
     relative URL without a base
     at <snip>/swc-<ver>/src/lib.rs:386
```

Further, for the common "missing file" error, give a better message that hopefully describes the two paths tried, rather than increasing confusion by talking about a `.js.map.map` file if it was correct.

Now it will show, for example:

```
  ERROR  failed to read input source map: failed to find input source map file "index.js.map" in "file:///D:/github/skilitics/billing/api-client/lib/services/index.js" file as either "file:///D:/github/skilitics/billing/api-client/lib/services\\index.js.map" or with appended .map
    at D:\github\swc-project\swc\crates\swc\src\lib.rs:400
```

(This example appears to be an issue in swc-node, where it is passing a file URL which is treated as a path, and therefore never exists, I'm working on a PR for there too)

There are more improvements that could make this code clearer and more reliable, but this should resolve many of the confused users (including me!)

**Related issue:**

See (maybe can be considered to fix): #8944, #8910

[PR in swc-node to fix the originating issue](https://github.com/swc-project/swc-node/pull/840)


simonbuchan:
Should I be concerned about those performance changes? I haven't touched anything that should affect them, so I assume they are noise, but the messaging on the bot implies it's pretty confident...

kdy1:
Nope, it's quite flaky.

kdy1 (approved):
Thanks!

PR #9425, opened 2024-08-13
feat(html/minifier): Support using custom css minifier

ahabhgk:
**Description:**

Support use a custom css minifier when minimizing html, add `feature = "custom-css-minifier"` to enable this and remove the dependencies of `swc_css_*`.
And `feature = "default-css-minifier"` keep the original behavior, using `swc_css` when minimizing html, and this is the default feature.
(Rspack is now using LightningCSS to minimize CSS files, but the CSS insides html is still using `swc_css` (which called by `swc_html_minifier`) to do the minify, so we want to also use LightningCSS to minify the CSS insides html to keep the consistent with CSS files)

### DIFF
--- a/.changeset/ninety-tips-heal.md
+++ b/.changeset/ninety-tips-heal.md
@@ -1,0 +1,7 @@
+---
+swc_core: patch
+swc_html: patch
+swc_html_minifier: patch
+---
+
+feat(html/minifier): Support using custom css minifier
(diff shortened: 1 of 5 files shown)


kdy1 (changes requested):
Can you add `default-features = false` to https://github.com/swc-project/swc/blob/2cdab4102194444ed5796db55bdab06fb6b5b9ec/crates/swc_html/Cargo.toml#L27?
Otherwise you can't disable default features.

kdy1 (approved):

PR #9441, opened 2024-08-16
feat: make error format configurable

mertcanaltin:
- Added `ErrorFormatConfig` enum to allow configuration of the `ErrorFormat` used in the `minify_sync` function.
- Updated `minify_sync` to accept an optional `error_format` parameter, which can be set to "normal", "json", or "color".
- The `error_format` parameter defaults to `ErrorFormat::Normal` if not provided.
- This change resolves the TODO for making the `ErrorFormat` configurable, providing greater flexibility in error reporting formats.

can be used in the typeScript field : ```const result = minify_sync(codeBuffer, optsBuffer, "json");```


kdy1 (changes requested):
Build in CI failed

kdy1:
cc @kwonoj Do you have any idea about this?

bindings/binding_core_node/src/minify.rs
```diff
@@ -16,6 +17,12 @@ use swc_core::{
 
 use crate::{get_compiler, util::try_with};
 
+#[derive(Debug, Deserialize)]
+pub enum ErrorFormatConfig {
```
kdy1, line 21: Why is this option in binding?

bindings/binding_core_node/src/minify.rs
```diff
@@ -86,7 +109,11 @@ fn minify(code: Buffer, opts: Buffer, signal: Option<AbortSignal>) -> AsyncTask<
 }
 
 #[napi]
-pub fn minify_sync(code: Buffer, opts: Buffer) -> napi::Result<TransformOutput> {
+pub fn minify_sync(
+    code: Buffer,
+    opts: Buffer,
+    error_format: Option<ErrorFormatConfig>,
```
kdy1, line 115: Also, why is this an additional parameter, instead of additional field?

kdy1 (changes requested):

kwonoj:
> Do you have any idea about this?

Sorry, checking in late. Mind elaborate what you have in mind? the config option itself?

kdy1:
Yeap, and it being an additional parameter instead of a field.

kwonoj:
Yes, curious if there's a reason it have to be a 3rd arg instead of extending existing option.

bindings/binding_core_node/src/minify.rs
mertcanaltin: The ErrorFormatConfig was introduced to provide flexibility in formatting error outputs, allowing different representations (e.g., JSON) depending on the user's needs. It was included here to streamline error handling directly within the minification process

bindings/binding_core_node/src/minify.rs
mertcanaltin: ı think reasoning behind using an additional parameter instead of a field was to maintain the separation of concerns. By keeping ErrorFormatConfig as a separate argument, it avoids bloating the existing options structure and allows for easier future extensions or modifications

bindings/binding_core_node/src/minify.rs
kdy1: I mean, why is it stored in `bindings/*`? I think it's a quite general option and I believe `swc_error_reporters` is the better place to store this type.

mertcanaltin:
> Yes, curious if there's a reason it have to be a 3rd arg instead of extending existing option.

decision to introduce it as a third argument rather than extending the existing options was to avoid coupling the error formatting configuration with the core minify options. This modular approach allows users to control error output separately, which can be more intuitive and flexible

bindings/binding_core_node/src/minify.rs
mertcanaltin: now I understand better thank you I will make an update

mertcanaltin:
thank you for your suggestions I sent a commit

bindings/binding_core_node/src/minify.rs
```diff
@@ -19,7 +20,7 @@ use crate::{get_compiler, util::try_with};
 struct MinifyTask {
     c: Arc<swc_core::base::Compiler>,
     code: String,
-    options: String,
+    options: JsMinifyOptions,
```
kdy1, line 23: Please revert this

bindings/binding_core_node/src/minify.rs
```diff
@@ -57,7 +58,7 @@ impl Task for MinifyTask {
 
     fn compute(&mut self) -> napi::Result<Self::Output> {
         let input: MinifyTarget = deserialize_json(&self.code)?;
-        let options: JsMinifyOptions = deserialize_json(&self.options)?;
+        let options = self.options.clone();
```
kdy1, line 61: Same here

mertcanaltin: ı used, thanks

bindings/binding_core_node/src/minify.rs
mertcanaltin: ı used, thanks

bindings/binding_core_node/src/minify.rs
```diff
@@ -76,7 +77,8 @@ impl Task for MinifyTask {
 fn minify(code: Buffer, opts: Buffer, signal: Option<AbortSignal>) -> AsyncTask<MinifyTask> {
     crate::util::init_default_trace_subscriber();
     let code = String::from_utf8_lossy(code.as_ref()).to_string();
-    let options = String::from_utf8_lossy(opts.as_ref()).to_string();
+    let opts_str = String::from_utf8_lossy(opts.as_ref()).to_string();
```
kdy1, line 80: You should revert this too

bindings/binding_core_node/src/minify.rs
```diff
@@ -77,6 +78,7 @@ fn minify(code: Buffer, opts: Buffer, signal: Option<AbortSignal>) -> AsyncTask<
     crate::util::init_default_trace_subscriber();
     let code = String::from_utf8_lossy(code.as_ref()).to_string();
     let options = String::from_utf8_lossy(opts.as_ref()).to_string();
+    let options: JsMinifyOptions = deserialize_json(&opts_str).expect("Invalid options format");
```
kdy1, line 81: I mean you should never deserialize json from the JS thread

bindings/binding_core_node/src/minify.rs
```diff
@@ -89,18 +91,15 @@ fn minify(code: Buffer, opts: Buffer, signal: Option<AbortSignal>) -> AsyncTask<
 pub fn minify_sync(code: Buffer, opts: Buffer) -> napi::Result<TransformOutput> {
     crate::util::init_default_trace_subscriber();
     let code: MinifyTarget = get_deserialized(code)?;
-    let opts = get_deserialized(opts)?;
+    let opts_str = String::from_utf8_lossy(opts.as_ref()).to_string();
+    let opts: JsMinifyOptions = deserialize_json(&opts_str)?;
```
kdy1, line 95: Please revert this. `minifySync` need to do all operations on the JS thread, so we deserialize json from the JS thread, but `minify()` is about delegating the work to the non-main thread.

kdy1 (changes requested):
Please patch https://github.com/swc-project/swc/blob/b0b5e36675835ce5f98ad55528dddc6514064553/bindings/binding_core_node/src/minify.rs#L62-L66 too

bindings/binding_core_node/src/minify.rs
```diff
@@ -59,15 +60,18 @@ impl Task for MinifyTask {
...
 
-        try_with(self.c.cm.clone(), false, ErrorFormat::Normal, |handler| {
-            let fm = input.to_file(self.c.cm.clone());
-
+        try_with(self.c.cm.clone(), false, ErrorFormat::Normal, |handler| { 
+            let fm = input.to_file(self.c.cm.clone()); 
+        
             self.c.minify(fm, handler, &options)
         })
+        .map_err(|err| {
```
kdy1, line 68: Please revert this

kdy1 (changes requested):

bindings/binding_core_node/src/minify.rs
mertcanaltin: got it back thanks

bindings/binding_core_node/src/minify.rs
kdy1, line 23: It should be String

mertcanaltin: thanks 🙏

PR #9462, opened 2024-08-20
feat(bindings/html): Allow using `lightningcss` as minfiier

kdy1:
**Description:**


kdy1:
@hardfist I want to hear opinion about baking lightningcss part into `swc_html_minifier` itself.

kdy1:
cc @devongovett for visibility 👍

hardfist:
> @hardfist I want to hear opinion about baking lightningcss part into `swc_html_minifier` itself.

I think it's ok if can be controlled by feature flag

Issue #9460, opened 2024-08-20
Cannot remove unused variables declared by object destructuring with default value


wjw99830:
### Describe the bug

Cannot remove unused variables declared by object destructuring with default value.

However, it works without default value or declared by member accessing with nullish coalescing.

### Input code

```typescript
const { a, b = 'b' } = {};
const c = {}.c ?? 'c';
```

### Config

```json
{
  "minify": true,
  "jsc": {
    "target": "esnext",
    "minify": {
      "compress": {
        "module": true,
        "pure_getters": true
      },
      "mangle": false
    }
  }
}
```

### Playground link (or link to the minimal reproduction)

[...]

PR #9465, opened 2024-08-20
fix(es/minifier): take ident for ObjectPatProp::Assign with `value: Some(_)`

wjw99830:
**Description:** 
In order to remove unused variables declared by object destructuring with default value, call `take_ident_of_pat_if_unused` for `ObjectPatProp::Assign` whose `value` match `Some(_)`

**Related issue:** 

 - Closes https://github.com/swc-project/swc/issues/9460


crates/swc_ecma_minifier/tests/fixture/issues/9460/input.js
```diff
@@ -0,0 +1,3 @@
+const a = 1;
+const { b = "b", c, d } = {};
```
magic-akari, line 2: Even with `pure_getters`, the computation of the default value for assignment may still have side effects.
For example:

```JavaScript
const { b = console.log("b"), c, d } = {};
```

@Austaras What do you think?

wjw99830: `ignore_return_value` will check if the expr is side-effect-free.

wjw99830: > `ignore_return_value` will check if the expr is side-effect-free.

It seems that `take_ident_of_pat_if_unused` will return before calling `ignore_return_value`, so that the expression will always be removed if unused.

Austaras: > `ignore_return_value` will check if the expr is side-effect-free.

No. There're multiple places who would call `take_pat_if_unused` like `drop_unused_arrow_params`

Austaras: Maybe just check that assign value is side effect free?

wjw99830: > may_have_side_effects

I tried it. But expressions with pure annotation cannot be recognized so that there're still side-effect-free expressions in output code.

**Input code**
```js
const { a = /*#__PURE__*/ sideEffectFree(), b = "b" } = {};
```

**Output code**
```js
const { a = /*#__PURE__*/ sideEffectFree() } = {};
```

wjw99830: See the latest commit, expected output is
```js
const { a = sideEffect(), d } = {};
console.log(d);
```
but got
```js
const { a = /*#__PURE__*/ sideEffectFree(sideEffect()), d } = {};
console.log(d);
```
`is_pure_callee` doesn't consume the pure annotation.

kdy1: It's intentional, and if we want to drop it we can invoke `ignore_return_value` from the method.

wjw99830: It seems that `ignore_return_value` doesn't consume pure either. Its return expr is same as input expr.

kdy1 (changes requested):
Can we extract all side effects to somewhere, and drop useless bindings instead of checking for side-effect-free from here?

It's how other passes work.

kdy1:
I made https://github.com/swc-project/swc/pull/9478

wjw99830:
> drop useless bindings

Assign of object pattern only run when the binding's value is `undefined`. If we drop the binding, the expr will always run.

wjw99830:
Maybe we could invoke `has_mark` from `ignore_return_value`?
It's sure that the return value of expr is unused when calling `ignore_return_value`, and we need the assignment indeed.

kdy1:
Yeap it would be fine.

wjw99830:
Should we apply the optimize of `take_pat_if_unused` for unpredictable expression when `pure_getters` is `true`?

```js
export function Component(props) {
  const { a, b } = props;
}
```

crates/swc_ecma_minifier/src/compress/optimize/unused.rs
```diff
@@ -164,6 +164,53 @@ impl Optimizer<'_> {
         params.retain(|p| !p.is_invalid());
     }
 
+    #[cfg_attr(feature = "debug", tracing::instrument(skip_all))]
+    fn take_ident_of_obj_pat_if_unused_and_side_effect_free(
```
kdy1, line 168: I think this is unnecessarily complex.

See https://github.com/swc-project/swc/pull/9478/files#diff-0ff76c3ab1246166f0a7132a4da5346082038c9dd821ab81d95343aa31510aa0R399-R408

wjw99830: OK.
This pr could be closed and then follow #9478.

PR #9470, opened 2024-08-20
feat(plugin/runner): Try using `rmp_serde` instead of `rkyv`

kdy1:
**Description:**

Use `rmp_serde::to_vec_named` instead of `rkyv`.

### `rkyv`  (typescript.js)

```
# Invoke
es/plugin/invoke/1      time:   [328.21 ms 331.07 ms 334.07 ms]

# Native serialization + deserialization
es/plugin/serde/1       time:   [158.78 ms 159.99 ms 161.31 ms]
```

### `rmp_serde`  (typescript.js)

```
# Invoke
es/plugin/invoke/1      time:   [1.7162 s 1.7230 s 1.7309 s]

# Native serialization + deserialization
es/plugin/serde/1       time:   [449.86 ms 452.11 ms 454.59 ms]
```


kdy1:
cc @kwonoj @magic-akari @Austaras 

We can have `default` values for `rmp_serde`, so we can avoid breaking Wasm plugins while adding fields (mainly for newer TypeScript syntax).

Do you think the gain is enough to justify the performance cost?

kdy1:
I'm also considering other serialization libraries like capnproto

kdy1:
I updated the timing, and it seems like `rkyv` is super fast

Issue #9459, opened 2024-08-20
Provide a compress option like `pure_getters` to remove unused variables by array destructuring


wjw99830:
### Describe the feature

With `pure_getters`, variables by object destructuring will be removed if unused.

Provide a compress option like `pure_getters` to remove unused variables by array destructuring.

**Input code**
```js
const [state, setState] = /*#__PURE__*/ useState();
```

**Output code**
```js
// all code should be removed
```

PR #9478, opened 2024-08-21
feat(es/minifier): Drop more patterns with `PURE` marker

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/9459
 - Closes https://github.com/swc-project/swc/issues/9460
 - Closes https://github.com/swc-project/swc/pull/9465

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/optimize/unused.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/unused.rs
@@ -318,7 +318,15 @@ impl Optimizer<'_> {
         if !name.is_ident() {
             // TODO: Use smart logic
             if self.options.pure_getters != PureGetterOption::Bool(true) {
-                return;
+                let has_puer_ann = match init {
+                    Some(Expr::Call(c)) => c.ctxt.has_mark(self.marks.pure),
+                    Some(Expr::New(n)) => n.ctxt.has_mark(self.marks.pure),
+                    Some(Expr::TaggedTpl(t)) => t.ctxt.has_mark(self.marks.pure),
+                    _ => false,
+                };
+                if !has_puer_ann {
+                    return;
+                }
             }
 
             if let Some(init) = init.as_mut() {
--- a/crates/swc_ecma_minifier/src/compress/optimize/unused.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/unused.rs
@@ -1,3 +1,4 @@
+use once_cell::sync::Lazy;
 use rustc_hash::FxHashSet;
 use swc_atoms::JsWord;
 use swc_common::{util::take::Take, DUMMY_SP};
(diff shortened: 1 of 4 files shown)


crates/swc_ecma_minifier/src/compress/optimize/unused.rs
```diff
@@ -382,10 +396,16 @@ impl Optimizer<'_> {
 
                             self.take_pat_if_unused(&mut p.value, None, is_var_decl);
                         }
-                        ObjectPatProp::Assign(AssignPatProp {
-                            key, value: None, ..
-                        }) => {
-                            self.take_ident_of_pat_if_unused(key, None);
+                        ObjectPatProp::Assign(AssignPatProp { key, value, .. }) => {
+                            if has_pure_ann {
```
wjw99830, line 400: `has_pure_ann` only represents that `init` is pure. Consider that `init` isn't pure, but the `value` is pure

kdy1: I don't expect such case to exist in real world code

wjw99830: The `value` is an independent and unpredictable expression which inits the value of `undefined` key. I think it has nothing to do with `init`

kdy1: It's not simply dropping `value`; it's calling `ignore_return_value` on the value because we can drop the initializer.

wjw99830: Yeap.
But if the `init` can't be dropped while the `value` can, this logic will skip dropping `value` because `has_pure_ann` is `false`.

kdy1: > I don't expect such case to exist in real world code

wjw99830: for example:
```js
function getDefaultTextInfo() {
  return { a: 'a', b: 'b' };
}

const { textInfo = getDefaultTextInfo() } = useAtomValue(DataAtom);
```

wjw99830: It's very common but not generated by tool. The example was just from my private project:
```js
const { state = {} } = getRuntimeInfo();
```

`state` is unused after tree shaking, but it cannot be removed.

kdy1: This is not removable unless pure_getters is true. A member access can have a side effect.

kdy1: If you are talking about `pure_getters: true`, feel free to send a PR

wjw99830: > This is not removable unless pure_getters is true. A member access can have a side effect.

It can't be removed while `pure_getters` is true.

See [playground](https://play.swc.rs/?version=1.7.14&code=H4sIAAAAAAAAA8vMLcgvKlGoVkhPLQkqzSvJzE31zEvLV6hVSCvKz1VQL4KIqVtzcSXn5xWDVCbqKCQp2CpU1wIV2aLp09C0BqvLz0nVy8lP10jUtAYAWiuesmIAAAA%3D&config=H4sIAAAAAAAAA31Vy47bMAy85ysCn%2Few2EMPvRUFeus3CIpEOUol0RCpbIJF%2Fr2047w2dC6BoyE5fGn0tVqvux257uf6Sz7lz1CBoO7hV0q%2FMWcoTAIGmwjeZgtbxeLqISd0LGwPctKBy5ZcjQN3bxd0R4fHCK9ZJpPTzMW29sBTZPp4%2F3ifo3YJkeBbXjmWGI73eTnMIxHdnY126Fsanbm2a0ZyLlRNqXfGKn7Ss88GMYEtLxBjycTC0EPVAjtMyQ4EZm%2BrEmWswNZIqFGMYGPwZqg4qHjxkSMW4XxGPVhvHHqlEz5WcBz3oLkJl7gVwgRKPRPsYdP6ftqRb96wt6lZVjjhMI1KslWibjESm9CK1sIzuNCDMzg397tnDKYCt1qe%2FXYYy8JM%2FgFIB5IlKjaDFneyCLJPS97hpWcsQVaZjwoue69VWaCXppoYg9LZsTNQOWrTrOCbg7GzTktnhhfaR9GDgRBkV5TQ9BnZbTVSPg6AQQFkvjZoW3UGzPUWLuDjhXgB%2F5EqWV%2Bw2SJb3i6jdMwbTC8IMvAW%2FQsDGQXjMlxFJQ7DMt6KB1kN8KpJowl4FgG5AIwmTTr6tBtyPSSi6RNuVJkYWhUUmOF6F2bwdFXzbEt%2FU4NHAa%2BigQFrvtflCtbxoyjXVjjmMUZnG6MMIrruQrS6%2FE5RbwI%2BP1njPp2fiB%2FdzejyGlzq6SL9vVf%2B1ek%2FN4d5Zv0GAAA%3D)

kdy1: I know

wjw99830: The closed #9460 was from this.

PR #9530, opened 2024-09-04
fix(es/minifier): Prevent removing side effects from accessing getter

CPunisher:
Closes https://github.com/swc-project/swc/issues/9500

Caused by https://github.com/swc-project/swc/blob/c7fdd6b69b129a11465125d4e11a898326b7e884/crates/swc_ecma_minifier/src/compress/pure/misc.rs#L1547. When the object with getters pass to `self.ignore_return_value`, 
https://github.com/swc-project/swc/blob/c7fdd6b69b129a11465125d4e11a898326b7e884/crates/swc_ecma_minifier/src/compress/pure/misc.rs#L966
converts the object to `0` because the object is side-effect-free according to
https://github.com/swc-project/swc/blob/c7fdd6b69b129a11465125d4e11a898326b7e884/crates/swc_ecma_utils/src/lib.rs#L1496

We should skip this process to fix the issue.

As is known only accessing getters and setters may cause side effect, we can safely do the transformation when none of them appears in the object. More precision is possible if comparing the lit prop names. I also collect computed keys of getters and setters in the object, is there any bad case?

The reason why only numeric (string) key removes the statement is that string key (`Computed`) is converted to `Ident` in other phases, e.g. `{}['a']` => `{}.a`, which does not matching the pattern.


Austaras (changes requested):
Better to add tests at issues/9500, otherwise good.

Austaras (dismissed):

kdy1 (approved):

Issue #9429, opened 2024-08-14
Cannot implement decorator as a private method


bryfox:
### Describe the bug

SWC does not appear to support defining a decorator with a private method

### Input code

```typescript
See playground link
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "decorators": true,
      "decoratorsBeforeExport": true
    },
    "transform": {
      "legacyDecorator": true
    },
    "target": "es2022"
  },
  "isModule": true
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.7.11&code=H4sIAAAAAAAAA3WSz0rEMBDG7%2FsU43pJoeQBun9QEC8iiHgTkWw77QazSUimrkX67k7abl2FvaTpzJfffDNJaVSMcO8cfC8AIinSJVwb1zRYPSLtXbW%2BDU0E%2FCK0VYTWflh3tK9vOTwjbQUfAnj3wZGjzmMBaXV1Iso5ms8qj4G6B%2BwKLhW0bcZMhbEM2pMLBbywvnqalHdzYi1or2ORuDlIKRWbKiBZy2CzBT5w0BHXydM2QbOhH4DS2UhwGDqBzVkl%2BalMi1erfwbGMCvr1paknQUxVTsRASh0836s4AxKnplY7rB2AZfZak4HpDbYyYFU3ptuaCWHAXoS9lBrq4y5SFY1YfgF98O3T7%2F9gpcbkSb%2B5%2BIyDvt2Z%2FhCm4BIIitA2e5sMDN8r0dyv2DYOLKan8QGLB7TyAVnOSAnzuoH9oK%2FCzYCAAA%3D&config=H4sIAAAAAAAAA1VPuw6DMAzc%2BQrkuUOVoUP%2FoR8RBYOCyEO2kYoQ%2F94EkrTd7LvznW%2Fv%2Bh5mNvDs9zSmJWpipLYnhDcv%2Bp0QkC0iG7JR4FZZ4UwJrdigAU0gLYG4MCdxXDyIpgkl2yGru1LFCpYQGBM86oWLFzjr7bj9PmOCi4TM%2F8Is1X5a2v2V2JVUcGFYT7KUzEWuDx7wFdWwZgyWX%2FXy7HF8AF26uLkwAQAA

### Expected behavior

No error when running the emitted code; decorator executes as expected.

`tsc` output for comparison:

```ts
class Foo {
    // ...
    static {
[...]

PR #9535, opened 2024-09-05
fix(es/decorator): Add support for private access expressions in legacy decorators

magic-akari:
**Related issue :**

- Closes: #9429


magic-akari:
I have no clue about this performance regression.

kdy1 (approved):

PR #9547, opened 2024-09-11
test(es): Add tests for arm64 windows

kdy1:
**Description:**

### DIFF
--- a/packages/core/__tests__/issue-9520.mjs
+++ b/packages/core/__tests__/issue-9520.mjs
@@ -7,7 +7,14 @@ it("should transform synchronously", () => {
 	return 'foo';
 };`);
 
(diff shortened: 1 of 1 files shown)


packages/core/__tests__/issue-9520.mjs
```diff
@@ -0,0 +1,21 @@
+import swc from "..";
+
+it("should transform synchronously", () => {
+    const { code } = swc.transformSync(`export function foo() {
+	if (false) {
+	}
+	return 'foo';
+};`);
+
+    expect(code).toMatchInlineSnapshot();
```
wojtekmaj, line 10: It seems like there is no snapshot included? 🤔

PR #9562, opened 2024-09-18
fix(plugin): Don't panic when ast byte not match

jerrykingxyz:
**Description:**

1. Disable rkyv strict feature, it is useless in swc.
2. Archived_root will panic if the wasm plugin ast binary not match core ast, so use safe api to avoid panic when mismatch. Implementation reference: https://github.com/rkyv/rkyv/blob/v0.7.43/examples/json/src/main.rs#L57-L76

**BREAKING CHANGE:**

No

### DIFF
--- a/.changeset/four-tips-impress.md
+++ b/.changeset/four-tips-impress.md
@@ -1,0 +1,8 @@
+---
+ast_node: major
+swc_atoms: major
+swc_common: minor
+swc_plugin_proxy: minor
+---
+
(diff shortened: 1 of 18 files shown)


kdy1:
@jerrykingxyz This PR seems to cause some issues with `swc_html.` Can you take a look?

image

jerrykingxyz:
@kdy1 sure, i will create a new PR to fix it

PR #9574, opened 2024-09-20
fix(ast): Add `archive(check_bytes)` to all relevant AST types

jerrykingxyz:
**Description:**

1. all struct that support rkyv add `archive(check_bytes)`
2. all recursive struct that support rkyv add check_bytes bound

this PR will fix the error reported in https://github.com/swc-project/swc/pull/9562

**BREAKING CHANGE:**
No


kdy1 (approved):
Thank you!

Issue #9567, opened 2024-09-19
Wrong source map generated for minifiy


SyMind:
### Describe the bug

`\n` in template strings will be converted to a newline, but the source map is incorrect.

The generated source map is also broken with Sentry. For more details, please refer to this issue: [https://github.com/web-infra-dev/rspack/issues/7914](https://github.com/web-infra-dev/rspack/issues/7914).

I have created a repository to reproduce the bug.

![image](https://github.com/user-attachments/assets/1153659f-49d8-471e-b438-24a1fd0d94f9)

### Input code

```typescript
function foo(span, error) {
    span.setStatus({
        message: `${error.message} ${error.code ? `\nMongoose Error Code: ${error.code}` : ''}`,
    });
}
foo();
```

### Config

```json
import * as path from 'path';
import * as fs from "fs";
import swc from "@swc/core";
 
const src = fs.readFileSync(path.join(import.meta.dirname, 'index.js'), 'utf-8');
const { code, map } = await swc.minify(src,
  {
    module: true,
    compress: true,
    mangle: true,
    sourceMap: true,
  }
);
```

### Playground link (or link to the minimal reproduction)

[...]

PR #9578, opened 2024-09-22
fix(es/codegen): Fix wrong sourcemap when there are new lines in tpl

CPunisher:
**Description:**

1. split quasi by '\n' 
2. add the mapping between original string and generated string one by one.

image

closes https://github.com/swc-project/swc/issues/9567


kdy1 (approved):

cjnoname:
@kdy1 Hey mate, when the new version with this fix will be released? It's quite a serious bug.

PR #9605, opened 2024-10-02
perf(es/typescript): Reduce unnecessary visits

magic-akari:


kdy1 (approved):

magic-akari:
`visit_mut_module_items` is not only called in es modules but also in ts modules.
Moving the related methods to `visit_mut_module` ensures that our `UsageCollect` and `DeclareCollect` are called only once.

Issue #8718, opened 2024-03-08
Minifier should check type of assignment target before merging assignments


hyp3rflow:
### Describe the bug

Currently in `replace_seq_assignment` step, we don't check the type of target.
however, if the type of target differs, the total count of actual operations matters.

### Input code

```typescript
let a = '';
console.log((a += 1, a += 2))
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2022",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": true,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true,
        "const_to_let": true,
        "pristine_globals": true,
        "passes": 3
      },
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link (or link to the minimal reproduction)

[...]

PR #9617, opened 2024-10-04
fix(es/minifier): Check type of assignment target before merging assignments

CPunisher:
**Description:**

Collect types of vars, maybe other optimization could benefit from this: `merged_var_type: Option<Value<Type>>`

When the variable is reassigned, the we merge the types with some simple rules:
`None` + `None` = `None`
`None` + `Some(ty)` = `Some(ty)`
`Some(ty1)` + `Some(ty2)` if `ty1` == `ty2` = `Some(ty1)`
Otherwise = Unknown

closes https://github.com/swc-project/swc/issues/8718


kdy1:
Is collecting initial types enough? I think it can be useful for `const` but not for others

CPunisher:
> Is collecting initial types enough? I think it can be useful for `const` but not for others

No, it's not enough. I realize that the usage analyzer is flow-insensitive, so it's hard to know the accurate type of a variable and I get stuck.

CPunisher:
@kdy1 What do you think of this implementation? This looks more sound than the main branch at the cost of losing some completeness (some cases that can be optimized are not).

kdy1:
I think it's quite a clever idea. Yeah, we have a merging function for merging characteristics of variables, and reusing it for types make sense I think.

PR #9623, opened 2024-10-08
feat(es/testing): Parse test code as a `Program` instead of a `Module`

leafypout:
**Description:**
This PR addresses the issue described in https://github.com/swc-project/swc/issues/8713

**BREAKING CHANGE:**
Will break unit tests that use `fold_module`/`visit_module`/`visit_mut_module` if the visitor is intended to work for both modules and scripts instead of using `fold_program`/`visit_program`/`visit_mut_program`.

When creating visitors, you should use `fold_program`/`visit_program`/`visit_mut_program` if you simply want to visit the top-level node.

When creating tests, the input source code will be parsed using `parse_program` by default. If you need to parse it as a `Module`, you can use `module: Some(true)` in `FixtureTestConfig` (or with `test!(module, ..)`), which will parse it as a `Program::Module`, or `Some(false)` for `Program::Script`. `None` will use `parse_program` (`parse_program` will auto-detect the underlying type).

### DIFF
--- a/.changeset/nine-trees-beam.md
+++ b/.changeset/nine-trees-beam.md
@@ -1,5 +1,6 @@
 ---
 swc_ecma_transforms_testing: breaking
+swc_ecma_transforms_optimization: patch
 ---
 
 feat(es/testing): Parse test code as a `Program` instead of a `Module`
--- a/.changeset/nine-trees-beam.md
+++ b/.changeset/nine-trees-beam.md
@@ -1,0 +1,5 @@
+---
+swc_ecma_transforms_testing: breaking
+---
+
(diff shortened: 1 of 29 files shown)


levi-nz:
Also, this removes `_ok_if_code_is_eq` parameter because it wasn't used or referenced anywhere.

levi-nz:
Should be good to go now. CI should pass (I hope).

levi-nz:
Also `exec_tr` and `compare_stdout` do not have this new feature. It seems like these functions are only used for modules and changing them messed some of the tests up.

kdy1 (approved):

PR #9625, opened 2024-10-08
feat(bindings/html): Accept `Buffer|string` instead of `Buffer`

kdy1:
**Description:**


kdy1:
@slorber Once it's published, `@swc/html` will accept `string | Buffer` for `minify*` APIs. I'll fix the publish pipeline asap and publish a new version soon.

slorber:
Thanks :)

PR #9642, opened 2024-10-14
feat(es/minifier): Support unary negate in `cast_to_number`

levi-nz:
**Description:**
Adds support for `-` in `cast_to_number`. Before it only worked if `arg` is `Infinity`. Now it uses a recursive call on `arg` so it works for expressions like `-5`, `-[]` etc.

This change is important because negative number literals (e.g. `-5`) are a `UnaryExpr` with `op`=`-` & `arg`=`5`, unless you apply `expr_simplifier` pass or something else that uses it.

**BREAKING CHANGE:**

### DIFF
--- a/.changeset/lazy-ladybugs-jump.md
+++ b/.changeset/lazy-ladybugs-jump.md
@@ -1,0 +1,6 @@
+---
+swc_ecma_utils: patch
+swc_core: patch
+---
+
+feat(es/minifier): Support unary negate in `cast_to_number`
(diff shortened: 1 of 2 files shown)


CPunisher:
Could you please add some tests?

levi-nz:
> Could you please add some tests?

There doesn't seem to be any tests for `cast_to_number` in this file, or anywhere else that I can see. So I don't think tests are required. Also this change is simply a recursive call.

CPunisher (dismissed):

kdy1 (approved):
Thank you!

PR #9657, opened 2024-10-17
feat(es/minifier): Implement optional catch binding

### LINKED ISSUE #8966: Minifier should use optional catch binding
### Describe the feature

Input:
```js
try {
    console.log("hello");
} catch (e) {}
```

Output:
```js
try{console.log("hello")}catch(e){}
```

Expected output (more minified):
```js
try{console.log("hello")}catch{}
```

I turned on minify & compress, and set the target to ES2022, so the target supports optional catch binding and the minifier can take advantage of it and remove the catch binding `(e)`.

_No response_

### Additional context

_No response_

CPunisher:
**Description:**

**BREAKING CHANGE:**

closes https://github.com/swc-project/swc/issues/8966

I think removing `!v.declared_as_catch_param` is safe, since I can't think of a bad case:
1. for the vars in the params of catch-clause, we prevent the removal if `self.options.ecma < EsVersion::Es2019 || !self.options.unused` is not satisfied.
2. for the usages or re-declarations with the same var name in the body of catch-clause, I think it's safe to remove them.

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/optimize/unused.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/unused.rs
@@ -178,7 +178,6 @@ impl Optimizer<'_> {
                 && v.usage_count == 0
                 && !v.reassigned
                 && v.property_mutation_count == 0
-                && !v.declared_as_catch_param
             {
                 self.changed = true;
                 report_change!(
--- a/.changeset/plenty-frogs-live.md
+++ b/.changeset/plenty-frogs-live.md
@@ -1,0 +1,6 @@
+---
+swc_ecma_minifier: patch
+swc_core: patch
+---
+
+feat(es/minifier): Implement optional catch binding
--- a/crates/swc_ecma_minifier/src/compress/optimize/mod.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/mod.rs
@@ -2876,15 +2876,17 @@ impl VisitMut for Optimizer<'_> {
         n.finalizer.visit_mut_with(self);
     }
 
-    fn visit_mut_catch_clause(&mut self, node: &mut CatchClause) {
+    fn visit_mut_catch_clause(&mut self, n: &mut CatchClause) {
+        n.visit_mut_children_with(self);
+
         if self.options.ecma < EsVersion::Es2019 || !self.options.unused {
             return;
         }
 
-        if let Some(param) = &mut node.param {
+        if let Some(param) = &mut n.param {
             self.take_pat_if_unused(param, None, false);
             if param.is_invalid() {
-                node.param = None;
+                n.param = None;
             }
         }
     }
--- a/crates/swc_ecma_minifier/src/compress/optimize/mod.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/mod.rs
@@ -2876,6 +2876,19 @@ impl VisitMut for Optimizer<'_> {
         n.finalizer.visit_mut_with(self);
     }
 
+    fn visit_mut_catch_clause(&mut self, node: &mut CatchClause) {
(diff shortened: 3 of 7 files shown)


crates/swc_ecma_minifier/src/compress/optimize/unused.rs
```diff
@@ -178,7 +178,6 @@ impl Optimizer<'_> {
                 && v.usage_count == 0
                 && !v.reassigned
                 && v.property_mutation_count == 0
-                && !v.declared_as_catch_param
```
Austaras, line 181: I believe we should still check `declared_as_catch_param`. The problem of related test cases arise because `a` in catch param and `a` in catch body shares same syntax context, which should be fixed in resolve pass.

CPunisher: Sorry, I do not quite get your point. Do you mean something related to variable hoisting? I run the related tests and whether removing `var a;` or not leads to the same result: `Uncaught ReferenceError: a is not defined`. Could do please tell me a more comprehensible example?

Austaras:
```js
function f() {
    try {
        x();
    } catch (a) {
        var a;
    }
}
```
both a resolve to `#3` which is wrong.

CPunisher: You mean the `a` in the catch param and the `a` declared in the catch block is actually different? Although I am still not clear why we should preserve `var a` in this case, I find https://tc39.es/ecma262/#sec-variablestatements-in-catch-blocks. Does this spec mean they are actually not different?
> They do not shadow the same-named bindings introduced by the [CatchParameter](https://tc39.es/ecma262/#prod-CatchParameter)

I also try:
```js
!function f() {
    var a = 1;
    try {
        throw "error"
    } catch (a) {
        var a = 2;
    }
    console.log(a) // 1
}()
```

kdy1: IIRC there was a browser that has a bug related to it. Exactly the 

> They do not shadow the same-named bindings introduced by the [CatchParameter](https://tc39.es/ecma262/#prod-CatchParameter)

thing

Austaras: > there was a browser

Highly likely to be Safari

Austaras: > You mean the `a` in the catch param and the `a` declared in the catch block is actually different? Although I am still not clear why we should preserve `var a` in this case, I find https://tc39.es/ecma262/#sec-variablestatements-in-catch-blocks. Does this spec mean they are actually not different?
> 
> > They do not shadow the same-named bindings introduced by the [CatchParameter](https://tc39.es/ecma262/#prod-CatchParameter)
> 
> I also try:
> 
> ```js
> !function f() {
>     var a = 1;
>     try {
>         throw "error"
>     } catch (a) {
>         var a = 2;
>     }
>     console.log(a) // 1
> }()
> ```

Shadow is what we need. But currently they're considered by SWC as same identifier, which may cause problem.

kdy1: Does this PR trigger the bug? If not, I think it can be fixed by another PR

Austaras: No but I believe we should still check `declared_as_catch_param`

Austaras: Ok I'm wrong. They should be the same context.

Austaras (dismissed):

PR #9658, opened 2024-10-17
fix(es/compat): Add missing visit children for `destructuring`

stormslowly:
**Description:**

add missing `visit_mut_children_with` in deconstructing  transform 

**BREAKING CHANGE:**
NO

ref: https://github.com/swc-project/swc/discussions/9647


magic-akari:
I believe the fix is correct, but I can't reproduce the issue in the [swc playground](https://play.swc.rs/?version=1.7.36&code=H4sIAAAAAAAAA0vLL1LQSM7PKy5RiM421FEoM4xVyE9T8E%2FKSk0u0UvNKynKTC3WyNfUVKjmUlBIQ1ZtBFRthE%2B1ggJIaX5Oql5OfroGxHQdBYg%2BTWugglquWpAqBQA3330PhQAAAA%3D%3D&config=H4sIAAAAAAAAA02NSw7CMAxE95wi8potLDgBGw5hhSkKSktku1KjqnfHJPx2njcfr7sQ6K6RTmH100VhUchXO9E6GS9OCHFkjZKKUTO3fe8Yyw3WEnqgN8RikInzGblA1F2TGS%2Bz1Wh8XOeM32OrBX3iSH%2BpNKWhOh84a2tT0sun2he3J5HUMBTGAAAA).

I suspect our transformer has redundant elsewhere.

stormslowly:
in my opinion, in swc playground run the  `for_of`  transform then `destructing` , so it can't be reproduced there.

@magic-akari  you can try comment out the visit children call, then run the new added case, panic will appear

kdy1:
If so, the order of transform is wrong. The compatibility transforms are expected to run in the correct order, and the correct order is.. well it's not documented but `esxxxx` functions encode the correct orders

magic-akari (dismissed):
Overall, this fix is in the right direction, let's merge this fix first.

PR #9664, opened 2024-10-21
chore(ci): Remove miri ci for `swc_ecma_codegen` and `swc_ecma_minifier`

CPunisher:
**Description:**

**BREAKING CHANGE:**

Running miri for `swc_ecma_codegen` and `swc_ecma_minifier` is too slow.
Miri on `swc_ecma_codegen` consumes up to 4 hours, and miri on `swc_ecma_minifier` consumes up to 6 hours, which exceeded github's 6 hour limit and makes ci bound to error.

Since they do not introduce too much unsafe code, I think we'd better remove the miri tests for them or run partial tests in a reasonable time.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -501,8 +501,8 @@ jobs:
           - string_enum
           - swc
           - swc_bundler
-          - swc_ecma_codegen
-          - swc_ecma_minifier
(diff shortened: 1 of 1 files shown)


kdy1 (approved):
Thanks!

Issue #8868, opened 2024-04-16
`swc.minifySync` panics on some inputs


s100:
### Describe the bug

When trying to minify certain (not very well-formed) strings, `swc.minifySync(...)` panics and terminates the Node.js process.

### Input code

```typescript
import swc from '@swc/core'
try {
  swc.minifySync('\u001a\u0000', {
    mangle: true,
    module: true
  })
} finally {
  console.log('this should be printed no matter what, but is not')
}
```

### Config

```json
{
  "//": "I have none"
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.3.100&code=H4sIAAAAAAAAAzWMsQoCMRBE%2B0D%2BYbs94Thja2VvYXGtTYjJEUh2ZS9Bgty%2Fa%2FBshuHNY2J%2BshRYXw6CcAa8fOvRsXjUqkiDt1bQ5ylHiqHNjdyA92rMyfY0BsefA5AtLcmfoUj14474Uf%2Bok%2B2g1QYhkk1p%2F3ZMKyc%2FJV4GvF2xG1p9AAzNzlaYAAAA&config=H4sIAAAAAAAAA1WPSw7DIAwF9zkF8rrbdtE79BAWdSIifrKJVBTl7iUE0maH3xsz8jooBbNoeKq1PMsQkYX4nEsi2Sf8lARIOxTNJia49XaWvRrRCtVoOxpIyBOluiX3hoMNQajjLXPGmzH%2FC3VwkUnkCu4o%2BsnSVTc0JbjwXmrZDkk50qF%2FwA%2FqsvNjMPLqm4kXGrYvhlQioBQBAAA%3D

### SWC Info output

    Operating System:
        Platform: win32
        Arch: x64
        Machine Type: x86_64
        Version: Windows 10 Enterprise
        CPU: (16 cores)
            Models: 11th Gen Intel(R) Core(TM) i7-11850H @ 2.50GHz

    Binaries:
        Node: 18.17.1
        npm: N/A
        Yarn: N/A
        pnpm: N/A

    Relevant Packages:
        @swc/core: 1.4.14
        @swc/helpers: N/A
        @swc/types: 0.1.6

    SWC Config:
        output: N/A
        .swcrc path: N/A

    Next.js info:
        output: N/A

### Expected behavior

In this case I believe the correct behaviour would be to throw an exception which can be caught, so that execution can continue.

### Actual behavior

```
[...]

PR #9665, opened 2024-10-21
fix(bindings): Update napi to handle string with `\0`

Brooooooklyn:
- Close https://github.com/swc-project/swc/issues/8868


kdy1 (approved):
Thanks!

PR #9677, opened 2024-10-24
feat(es/minifier): Optimize switch with side effect and termination tests

CPunisher:
**Description:**

This is learnt from https://github.com/terser/terser/pull/1044
**Key point:** all the cases (statements) after the last side-effect and non-terminated one are useless.

There are also some points on whether a statement is side-effect free:
- **Var declaration:** I think var declaration always contains side effect. 
- **Function declaration:** In non-strict mode, function declaration is same as var declaration. In strict mode, function is declared in enclosing block scope, so it's side-effect free. But there is a bad case when we call the function before it is declared in the switch case:
```js
"use strict";
const k = (() => {
    let x = 1;
    switch (x) {
        case y():
        case x: {
            async function y() {
                console.log(123);
            }
        }
    }
    return x;
})();
```

**This is an existing bug.** I'm not sure whether we should fix it since it is a very bad code and I also find some similar bad cases with terser.

~I'm also not sure it's appropriate to introduce context variable `in_strict` for `StmtExt:: may_have_side_effects`, because `in_strict` could change from `false` to `true`. But this is a **conservative** mistake and does not break the program. Maybe we should use visitor for context-aware side effect checker in the future.~

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/8953

### DIFF
--- a/.changeset/happy-planes-dance.md
+++ b/.changeset/happy-planes-dance.md
@@ -1,0 +1,5 @@
+---
+swc_ecma_utils: major
+---
+
+feat(es/minifier): Optimize switch and replace empty test with side effect and termination tests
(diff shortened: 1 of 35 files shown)


kdy1 (commented):
Can we add `in_strict` to `ExprCtx` instead? Does it involve too much changes? I introduced `ExprCtx` to reduce changes required when we are adding fields.

Issue #9110, opened 2024-06-28
@swc/helper _ts_generator wrong code exported


yf-yang:
### Describe the bug

@swc/helpers unexpectedly merges two different code paths.

### Input code

```typescript
function* test() {
  while (!False()) {
    // execute this line
    while (!False()) {
      // execute this line
      break;
    }
    // execute this line
    if (False()) {
      // NOT execute this line
      break;
    }

    // execute this line
    yield "correct";
    return;
  }

  // NOT execute this line
  yield "wrong";
  return;
}

function False() {
  return false;
}

for (const result of test()) {
  console.log(result);
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "externalHelpers": true
  },
  "env": {
    "targets": "Chrome >= 48"
  }
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.6.5&code=H4sIAAAAAAAAA42RPQ7CMAyFd07x6JQy0AOwM8LCBSA4bUSUSImjghB3p%2Flhg4gt8Xvfs2WraCVrZzdgCix6PFfAPGlDEOv92QQSfSkCwwC6k4xM4EkHGG0pCz%2F8DQK4eDrfdvnzaqdrBfEt%2BXA8%2FZfejn9oMld00nlPkrsCeeLobXpnvNGt4rN3dszwB11AVZeLOn6evuhQqVRczkNIZwMvWoiG4VS9RiGS5gxtjRtFcfQJfANPEoyouwEAAA%3D%3D&config=H4sIAAAAAAAAA1WPSw7DIAwF9zkF8rrbdtE79BCIOhERP9mOVBTl7oUE0maH3xszsA5KwcwGnmotxzIkTYx0ziXhHER%2FSgKSE7IhmwRuvRWu1agd4x5tRwOiaUKpW8j3hoOLkbHjLfM22DH%2FC030iZD5ClZUh8nhVTc0Jfj4XvayfaQ%2B9tA%2F4Ad12XkxWH71TaEFh%2B0LYuVI0xQBAAA%3D

### SWC Info output

  Operating System:
        Platform: darwin
        Arch: x64
        Machine Type: x86_64
        Version: Darwin Kernel Version 22.3.0: Mon Jan 30 20:42:11 PST 2023; root:xnu-8792.81.3~2/RELEASE_X86_64
        CPU: (16 cores)
[...]

PR #9684, opened 2024-10-29
fix(es/generator): Fix code generation for `break` in nested while

CPunisher:
**Description:**

**BREAKING CHANGE:**

closes https://github.com/swc-project/swc/issues/9110


kdy1 (approved):
Nice work! How did you investigate it? Actually I couldn't find a good way to investigate generator pass issues

kdy1 (approved):

PR #9690, opened 2024-10-30
perf(es/transform): Add inline annotation

kdy1:
**Description:**

### swc_core@v3.0.1, LTO = false

```
Benchmarking es/full/all/es5: Collecting 100 samples in estimated 54.860 s (100 iterations)
Benchmarking es/full/all/es5: Analyzing
es/full/all/es5         time:   [525.54 ms 526.92 ms 528.33 ms]
```

### swc_core@v3.0.1, LTO = false + `#[inline(always)]`

```
Benchmarking es/full/all/es5: Collecting 100 samples in estimated 53.857 s (100 iterations)
Benchmarking es/full/all/es5: Analyzing
es/full/all/es5         time:   [510.40 ms 512.59 ms 514.88 ms]
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -150,7 +150,8 @@ resolver = "2"
   wasmer-wasix              = { version = "0.27.0", default-features = false }
 
 [profile.release]
-lto = true
+debug = true
+# lto = true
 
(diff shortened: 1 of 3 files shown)


kdy1:
Oops I uploaded wrong files

PR #9701, opened 2024-11-01
feat(es/plugin): Introduce `manual-tokio-runtmie` to `swc` crate

kdy1:
**Description:**

This can cause a problem for Wasm plugin runtimes.


prisis:
Stupid question but shouldnt it be "manual-tokio-runtime" and not "manual-tokio-runtmie"?

kdy1:
Yeah you are right. It’a a typo and we need to add a feature with correct name (to avoid a breaking change)

Issue #9698, opened 2024-10-31
Strip-types handing async arrow with multi-line type parameter


acutmore:
### Describe the bug

Using `@swc/wasm-typescript` (strip-types mode), with input that contains an async arrow function that also has a type parameter that spans multiple lines the output JavaScript includes a syntax error.

### Input code

```typescript
let f = async <
   T
>(v: T) => v;
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2022",
    "loose": true,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.7.40&code=H4sIAAAAAAAAA8tJLVFIU7BVSCyuzEtWsOFSUFAI4bLTKLNSCNFUsLVTKLPmAgD0%2BerKIwAAAA%3D%3D&config=H4sIAAAAAAAAA1WPSw7DIAxE9zkF8rqLikUXvUMPgagTUfETdqSiKHcvJEDbHZ5h%2FMbbJAS8SMNdbOVZhqgSYRpzUSh7Vu%2BiAOeIpJOJDJfuMlVrVpbwkPbTAVZpQa4pJHmVsiXAhkBYZE4rNskZb%2Bb8i9TBxYREffOgOeUXi%2F%2FAqUHBhed6mO2UWvcscIPvpw4bi8HQoydrq2n%2FALtj80kWAQAA&strip-types=

### SWC Info output

N/A

### Expected behavior

The output to not contain a new line after the `async` keyword. e.g:

```js
let f = async ( 
    
  v   ) => v;
```

### Actual behavior

```js
let f = async  
    
 (v   ) => v;
```

### Version

1.7.40

### Additional context

This edge case was originally found by @seralexeev and reported on `ts-blank-space` which hit the same issue: https://github.com/bloomberg/ts-blank-space/issues/19

PR #9704, opened 2024-11-01
fix(es/typescript): Handle multiline type parameters in async arrow functions

magic-akari:
**Related issue:**

- Closes: #9698


kdy1 (approved):
Thank you!

Issue #9705, opened 2024-11-01
[dts]: remove unused import in isolatedDeclaration 


SoonIter:
### Describe the feature

some differences between swc and tsc

### source code

```typescript
import { useState } from "react";
import reactLogo from "./assets/react.svg";
import "./App.css";

function App(): JSX.Element {
	const [count, setCount] = useState(0);

	return (
		<div className="App">
			<div>
				<a href="https://reactjs.org" target="_blank" rel="noreferrer">
					<img src={reactLogo} className="logo react" alt="React logo" />
				</a>
			</div>
			<h1>Rspack + React + TypeScript</h1>
			<div className="card">
				<button type="button" onClick={() => setCount(count => count + 1)}>
					count is {count}
				</button>
				<p>
					Edit <code>src/App.tsx</code> and save to test HMR
				</p>
			</div>
			<p className="read-the-docs">
				Click on the Rspack and React logos to learn more
			</p>
		</div>
	);
}

export default App;
```

### expected output

the same as tsc

```typescript
import "./App.css";
declare function App(): JSX.Element;
export default App;
```

### actual output

```typescript
import { useState } from "react"; // unused
import reactLogo from "./assets/react.svg"; // unused and would emit type error
import "./App.css";

declare function App(): JSX.Element;
export default App;
```

### Additional context

# Related links

https://github.com/oxc-project/oxc/blob/0bb1aa4c6472d2553154fe2f1d8939abb69ed915/crates/oxc_isolated_declarations/src/lib.rs#L359
https://github.com/web-infra-dev/rspack/pull/8097

PR #9715, opened 2024-11-05
feat(typescript): Align `isolatedDeclaration` implementation with tsc

CPunisher:
**Description:**
- [x] **Basic implementation.** This also includes diagnostics and type inference.
    - [x] Function
    - [x] Class
    - [x] Variable declaration
    - [x] Enum
- [x] **Remove unused imports and declarations.** Based on the result ast of fast dts transformation, we collect the usage of ids which also contain syntax context. Then we prune the ast to cut off the unused imports and declarations.
- [x] **Port tests and bug fixes.** To ensure correctness, I'd like to port the fixtures as oxc does https://github.com/oxc-project/oxc/tree/main/crates/oxc_isolated_declarations/tests/fixtures.
- [x] **Benmarks**
- [x] Strip internal

**Implementation:** 
Most code is direct translation of https://github.com/oxc-project/oxc/tree/main/crates/oxc_isolated_declarations. The differences come from:
1. swc and oxc use **different ast**.
2. The transformation of swc is **mutation-based**, which directly mutates the whole cloned ast, while the transformation of oxc is **immutation-based**, which constructs and copies the ast nodes on demand. Maybe mutation-based transformation is not better, but I think it's also annoyed to construct ast nodes and it could be easy to refactor if there is some demands in the future.
3. oxc transforms while collecting references information to prune unused idents. However, the ast nodes of swc contain **syntax context**, we don't need to collect the information again. So I build a reference graph and calculate the reachability of identifiers.
4. I think oxc could have bugs. I manually compare the result with `tsc --declaration --isolatedDeclarations --emitDeclarationOnly --isolatedModules --noResolve --noCheck --strictNullChecks test.ts`. I also reguard those cases where tsc can't compile as undefined behaviors.

😅 Actually I find my implementation of point 2 is not too good either. I meet many bad cases after my basic implementation and write many ugly patches. So issues are welcome.

**Benchmarks:**
Roughly test with https://raw.githubusercontent.com/oxc-project/benchmark-files/main/vue-id.ts under M3Pro.
|  | parse without syntax context resolution | isolated declarations |
| -------------- | ------ | --------------------------------- |
| swc | 282.68ms | 335.44ms |
| oxc | 53.286ms | 63.917ms | 
| tsc | - | 3.875s |

**Known problem:**
- Due to the comment emitting algorithm in swc, isolated declaration may also generate comments with wrong positions.

closes https://github.com/swc-project/swc/issues/9705
closes https://github.com/swc-project/swc/issues/9718


SoonIter:
thanks for your help❤️️
 
😂 got an another blocking point of isolatedDeclaration

should generate the same error as tsc

```
src/App.tsx(8,10): error TS9007: Function must have an explicit return type annotation with --isolatedDeclarations.
```

I'll open another issue

Boshen:
FYI oxc uses immutable AST for a single transform pipeline.

CPunisher:
> FYI oxc uses immutable AST for a single transform pipeline.

I see. And I believe immutable AST gets better performance and other engineering benefits in isolated declaration transformation. After finishing, I have a little bit of regret... But maybe in the next pr 😁.

crates/swc/Cargo.toml
```diff
@@ -48,8 +48,8 @@ plugin_transform_host_native = [
   "swc_plugin_runner/plugin_transform_host_native",
 ]
 # Do not inject tokio runtime while running plugin transforms
-manual-tokio-runtmie = ["manual-tokio-runtime"]
 manual-tokio-runtime = []
+manual-tokio-runtmie = ["manual-tokio-runtime"]
```
kdy1, line 52: Can you revert this?

crates/swc/benches/isolated_declarations.rs
```diff
@@ -0,0 +1,50 @@
...
+    swc::Compiler::new(cm)
+}
+
+fn bench_isolated_declarations(criterion: &mut Criterion) {
+    criterion.bench_function("isolated-declarations", |b| {
+        let c = mk();
+        let fm =
+            c.cm.load_file(Path::new("benches/assets/vue-id.ts"))
+                .unwrap();
+        b.iter_with_large_drop(|| {
```
kdy1, line 20: `iter_with_large_drop` is effective only if you return the variable with large drop.

CPunisher: Yes, I find there is no difference if I change it to `iter`.

kdy1:
Sorry for the delay. I'm still reviewing this.

kdy1 (approved):

kdy1 (approved):
Thank you so much for a great work!

PR #9725, opened 2024-11-09
refactor(es/compat): Use special span instead of passing `static_blocks_mark`

magic-akari:
**Description:**

Let's make an assumption in the code: the use of dummy span as a private class field name is generated by us, it is there simply because it is necessary to maintain the order of side effect execution, its name is not referenced elsewhere, so we can safely remove its name.

This is a breaking change for Rust users.

**BREAKING CHANGE:**


kdy1 (commented):
What if plugin authors create a private name with dummy spans?

I think some plugins related to decorators may generate some private names, and if it's the case, most of the plugin authors will simply use dummy span (at least initially)

magic-akari:
Unfortunately, we have already used a similar trick here

https://github.com/swc-project/swc/blob/a417ff4d868b45a2157154e2334b8e1177c369e1/crates/swc_ecma_compat_es2015/src/classes/mod.rs#L631-L636

Or here

https://github.com/swc-project/swc/blob/49a8b3e756b4f99d2b958e68aa61e88e7d01d94e/crates/swc_ecma_compat_es2015/src/classes/constructor.rs#L23-L28

kdy1:
Hmm. Makes sense. I'll review this PR later today.

magic-akari:
Or should we add another special span?

By the way: We added PURE_SP last time, but we didn't add the documentation properly.

kdy1:
I didn't think deeply enough while writing the `DUMMY_SP` hack because there was no Wasm plugin API, and I introduced `static_block_mark` because there's a custom plugin API at the moment of writing.

How about using [`SYNTHESIZED`](https://rustdoc.swc.rs/swc_common/source_map/struct.BytePos.html#associatedconstant.SYNTHESIZED)? Maybe we should declare a constant for `Span`, too.


Also, can you add a bit of documentation?

PR #9740, opened 2024-11-16
fix(es/plugin): Migrate `swc plugin new` to use `.cargo/config.toml`

SpanishPear:
**Description:**
As per https://doc.rust-lang.org/cargo/reference/config.html we should be generating a `.cargo/config.toml` instead of `.cargo/config` when creating a new plugin

Motivation - console warnings 😆 

**BREAKING CHANGE:**
Technically this means users **must** have a toolchain > 1.38 (released september, 2019 - so five years ago). 
Given that:
 - the [public documentation for swc plugins](https://swc.rs/docs/plugin/ecmascript/getting-started#install-rust) link directly to the latest version of the rust toolchain (currently 1.81)
 - 1.38 was > **five years ago** 
 - This only impacts **new** usages of `swc plugin new ..`
 - README.md specifies that the `swc-project/swc` MSRV is 1.73 anyways - whilst that doesn't **specifically** refer to **generated** code, it shows the intent of the project :)

I don't believe it will have material impact. 

I will raise a separate PR to generate the correct MSRV for both `swc-project/swc` **and** the generated plugin itself.

### DIFF
--- a/.changeset/sixty-walls-hang.md
+++ b/.changeset/sixty-walls-hang.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_cli_impl: patch
+---
+
+Migrate  to config.toml to fix warning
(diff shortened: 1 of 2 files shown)


kdy1 (dismissed):
Thank you!

PR #9760, opened 2024-11-26
fix(es/decorators): Fix init order of `2022-03` impl 

### LINKED ISSUE #9669: [Bug] 2022-03 decorator initialise class instance in wrong order
### Describe the bug

As per tc39/proposal-decorators [[Adding initialization logic with addInitializer](https://github.com/tc39/proposal-decorators?tab=readme-ov-file#adding-initialization-logic-with-addinitializer)]
Method and Getter/Setter decorators should be initialised before any class fields are initialised

### Input code

```typescript
const myDecorator = (value, context) => {
  context.addInitializer(() => {
    console.log("init")
  })
  return value
}

class A {
  value = this.getNumber();
  value2 = this.getNumber();
  
  @myDecorator
  getNumber() {
    console.log("getNumber")
    return 1
  }
}

const a = new A()
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "decorators": true,
      "tsx": true
    },
    "transform": {
      "decoratorVersion": "2022-03"
    },
    "target": "es2021",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

https://play.swc.rs/?version=1.7.39&code=H4sIAAAAAAAAA3WQzwrCMAzG732KsFMLMtCrTBx48eI71C7MQtdCm%2FmXvbtdYGUHhRBIvi%2FJj5jgE8HwOqEJUVOI0IC8azfiBkzwhE9S0BzgI2Cpa911Z2%2FJamffGKUsBrak4LB2oZeVzaZK5f40p4g0Rg%2B8W0xCGKdTgpYHuZkv082muke6jMM1b1b7Rdv9EXMcV%2FC5XBl%2BMRWZwQrVdqZkKv6Hzuc8PqCV6gsyc1fOIQEAAA%3D%3D&config=H4sIAAAAAAAAA1VQQQ6DMAy78wqU8yYxJu2wP%2By6ewUBdaJNlRRpCPH3tVA6uCW2E1uei7KEjzTwLOcwhsUpFuS8B0Qm69U3IOAnh9Kwdh4uO9tiQ6w8sQSF5xEz4%2BWboBVZNgI8KysdsTl65C9vZNFko1td1fW1usP5WnGPPtIoQXBLQWAgEgxwpwZJEcBoq7vpaNOQcYwiZ2GUKtsP%2BX5zLJIrGGrHlUwVxRq2BA%2F4i3az%2FBi0vPbLtYXlByFGoz9uAQAA

### SWC Info output

_No response_

### Expected behavior
When I run output code, I should see console log
```
init
getNumber
getNumber
```

This is [tsc playground link](https://www.typescriptlang.org/play/?target=8#code/MYewdgzgLgBAtgTwCIFNQCcCGUTpgXhgAoA3TAGwFcUAaGUMKFADygEoCA+GAbwCgY9cE1YA6TABMJASTABLKHIpyAXinREiHfN36DBDCCHIpR5EAHMiAIjnyo1tgJgBfJ4PQooldGBhkqFD4XPj5gckwICBgAQV5nAOoCGCgACzkIUQsvADlKOAAjdS0AbgSKagAmZLSMrNz8oo02MsFnAAFEVAxsXGdsqDzC4o49fUNjU3MrawGhpsdnDy8fPwBGZxCQsPBoGExksBQAd1itIA), and if I run the output code, it's as expected
image

### Actual behavior
Wehn I run output code, I see console log
```
getNumber
getNumber
init
```
image

### Version

1.7.39

### Additional context

_No response_

jzhan-canva:
**Description:**
Fix the order of initialization for 2022-03 decorators.
TBH the [proposal](https://github.com/tc39/proposal-decorators) isn't very clear on how `init` is executed for [auto-accessor](https://github.com/tc39/proposal-decorators) decorator

**Related issue:**

 - Closes #9669


jzhan-canva:
@kdy1 please take a look on this draft

jzhan-canva:
the only thing I changed in this PR is to move `_init_proto` to before field initialization, I didn't change any logic of how `_init_proto` calls each initializer internally

I can double check the 2022-03 specification

jzhan-canva:
so I checked with babel playground with 2022-03 (assume it is the "correct" version)
between babel and my branch, there are two difference
1. how class is initialized by `addInitializer`: this is nothing to do with this pr, as this pr only change the class fields initialization 
2. how auto-accessor is initialized by `addInitializer` and `init`: babel use `addInitializer` first, my branch uses `init` first. this is controlled by the internal logic of `_init_proto`. and the proposal didn't specify which should run first

kdy1 (changes requested):
Can you add some tests?

jzhan-canva:
> Can you add some tests?
will do

another question @kdy1, does swc have corporate CLA that my employer can sign?

kdy1:
No, but I don’t think signing is necessary for this changes. I’m not going to monetize SWC or change license forever. I just want to be free while moving source codes in GitHub. I moved plugin codes from vercel/next.js to swc-project/plugins, and some npm packages within swc org, but that’s all.

crates/swc_ecma_transforms_proposal/src/decorator_impl.rs
```diff
@@ -718,10 +717,10 @@ impl DecoratorPass {
             }
         }
         body.visit_mut_with(self);
-        c.visit_mut_with(self);
         c.ident = preserved_class_name.clone();
         replace_ident(&mut c.class, c.ident.to_id(), &preserved_class_name);
         c.class.body.extend(body);
+        c.visit_mut_with(self);
```
kdy1, line 723: This is wrong because of the time complexity. You should select the exact node before calling `visit_mut_with`

kdy1 (changes requested):

crates/swc_ecma_transforms_proposal/src/decorator_impl.rs
jzhan-canva: sorry I don't get it, I don't see why moving this from line 721 to line 723 increases the time complexity

kdy1: Ah sorry, it's fine if it's about moving an existing line

kdy1 (approved):

Issue #9743, opened 2024-11-18
`loose` transpiles class properties


eps1lon:
### Describe the bug

When `env.targets` targets an environment with support for class properties, SWC will still transpile in a `constructor` implementation. 

This is problematic since throwing an error from the parent class constructor would create a stackframe that can't be fully sourcemapped since the `super(...args)` call in the generated code does not appear in the original code.

Alternatively, https://github.com/swc-project/swc/issues/9742 would fix the main motivation for this ticket.

### Input code

```typescript
class Foo extends Bar {
  handleScroll = () => {}
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "loose": true,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true,
  "env": {
    "targets": "node >= 18.18"
  }
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.9.2&code=H4sIAAAAAAAAA0vOSSwuVnDLz1dIrShJzUspVnBKLFKo5lJQyEjMS8lJDU4uys%2FJUbBV0NBUsLVTqK7lqgUAn5sc8jMAAAA%3D&config=H4sIAAAAAAAAA1WOTQrDIBBG955CZl0K2ZRQaG%2FQQ4iZBEP8wTGlEnL3GBttupv5nvP8FsY5jCThzpc0psUJT%2BjrnhKKJohPSgClFiS9cgEuhY60o15MhDlavwQmawkTCX7GI9LKqD6e1dJq55GoGKpVCzNM%2BC9mhxy07eYMj8ohOsz16Aa%2FR%2BWzKgZFr3JZWgGa90kk%2FIBhbwPGdsifD96016bNVrZud4rKzC0BAAA%3D

### Expected behavior

Input code is not modified.

### Actual behavior

Transpiled code contains `super(...args)` call that can't be sourcemapped to original code due to https://github.com/swc-project/swc/issues/9742.

### Version

1.9.2

kdy1:
There are two bugs here, one for documentation and one for the default value of the `Assumptions` API.

[...]

PR #9774, opened 2024-12-02
fix(es/preset-env): Disable all class passes for target with native class support

kdy1:
**Description:**

This is just PoC to ask for opinion @magic-akari. I think it sounds reasonable to disable all class properties transform where it's natively supported. So I made this, but I want to hear opinion about it.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/9743

### DIFF
--- a/crates/swc/tests/fixture/issues-1xxx/1456/case1/output/index.ts
+++ b/crates/swc/tests/fixture/issues-1xxx/1456/case1/output/index.ts
@@ -3,6 +3,7 @@ var _ts_decorate = require("@swc/helpers/_/_ts_decorate");
 var _ts_metadata = require("@swc/helpers/_/_ts_metadata");
 var _ts_param = require("@swc/helpers/_/_ts_param");
 class MyClass1 {
+    param1;
     constructor(param1){
         this.param1 = param1;
     }
@@ -15,6 +16,7 @@ MyClass1 = _ts_decorate._([
     ])
 ], MyClass1);
 class MyClass2 {
+    param1;
     constructor(param1){
         this.param1 = param1;
     }
@@ -27,6 +29,7 @@ MyClass2 = _ts_decorate._([
     ])
 ], MyClass2);
 class MyClass3 {
+    param1;
     constructor(param1){
         this.param1 = param1;
     }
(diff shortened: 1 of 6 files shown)


kdy1:
Does it work even for the targets with native class support?

magic-akari:
> Does it work even for the targets with native class support?

Yes, in TypeScript, even with native support for class fields, `useDefineForClassFields` can still influence the generated code.

input: 
```TypeScript
class Foo {
    bar: any;
}
```


output:
https://www.typescriptlang.org/play/?useDefineForClassFields=true&target=99#code/PTAEAEFcGcFMBFYDMCWA7WAxA9gJwMIA2AhtNJirIQCbQBcoALrpLAFAgSPG4DmsjBrGgYAHozZsAxiTKgc2UAG82oNaABGPBsTQBPANxsAvkA
```JavaScript
"use strict";
// @useDefineForClassFields: true
// @target: esnext
class Foo {
    bar;
}
```

output:
https://www.typescriptlang.org/play/?target=99&useDefineForClassFields=true#code/PTAEAEFcGcFMBFYDMCWA7WAxA9gJwMIA2AhtNJirIQCbQBcoSxhcAUCBAC7G4DmsnBrGgYAHp1asAxiTKgc2UAG9WoNaABGPBsTQBPANysAvkA
```JavaScript
"use strict";
// @useDefineForClassFields: false
// @target: esnext
class Foo {
}

```

magic-akari:
Perhaps we should redesign the transformation options for the class fields;  I suggest we include at least three options:

- spec `[[DEFINE]]`
- `[[SET]]`
- loose (preserve class fields if native support is available, otherwise use `[[SET]]` semantics)

Austaras:
> Perhaps we should redesign the transformation options for the class fields; I suggest we include at least three options:
> 
>     * spec `[[DEFINE]]`
> 
>     * `[[SET]]`
> 
>     * loose (preserve class fields if native support is available, otherwise use `[[SET]]` semantics)

Maybe 1 and 3 would be enough?

kdy1:
Closing as stale. I'll take another look in the future.

PR #9784, opened 2024-12-06
feat(es/transforms): Add `module.outFileExtension`

hanseltime:
**Description:**

Summary of Changes:

1. Adds an `outFileExtension` property to every module configuration option struct
2. passes this configuration through to the resolver function and changes all hard-coded `"js"` checks to use the new parameter
3. As a matter of backwards compatibility, structures default to "js" if nothing is provided
4. `@swc/core` snapshot tests for each module using `resolveFully` and `outFileExtension` (Please check the rendered output of the snapshots to make sure you don't have any issues with it).
1. QOL - updates `@swc/core` jest config to use explicit module names (this was because the mapping util in jest was exploding when it would try to load in the .node files that were over 512MB)
5. QOL - updates the test script to pass through any existing NODE_OPTIONS - this allows debugging tools to add their own inspect configurations while preserving our esm module flag

**Note**

This is a rudimentary fix to:  https://github.com/swc-project/swc/issues/3067#issuecomment-2516108434.  I had to implement it locally since I could not find a workaround in my timeline.

I do know that @kdy1 said they would work on it a few days ago, so I am opening this to either provide some visibility to a solution that works for me if they haven't started or to let them switch to PR critique if that feels more valuable for their time on this issue.  Let me know if this should be closed due to a better implementation already being in the works.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/3067

### DIFF
--- a/crates/swc/src/config/mod.rs
+++ b/crates/swc/src/config/mod.rs
@@ -48,7 +48,7 @@ use swc_ecma_transforms::{
         self,
         path::{ImportResolver, NodeImportResolver, Resolver},
         rewriter::import_rewriter,
-        EsModuleConfig,
+        util, EsModuleConfig,
     },
     optimization::{const_modules, json_parse, simplifier},
     proposals::{
@@ -1400,7 +1400,7 @@ impl ModuleConfig {
 
         let base_url = base_url.to_path_buf();
         let resolver = match config {
-            None => build_resolver(base_url, paths, false, &"js".to_string()),
+            None => build_resolver(base_url, paths, false, &util::Config::default_js_ext()),
             Some(ModuleConfig::Es6(config)) | Some(ModuleConfig::NodeNext(config)) => {
                 build_resolver(
                     base_url,
@@ -1750,7 +1750,7 @@ fn build_resolver(
     mut base_url: PathBuf,
     paths: CompiledPaths,
     resolve_fully: bool,
-    file_extension: &String,
+    file_extension: &str,
 ) -> SwcImportResolver {
     static CACHE: Lazy<DashMap<(PathBuf, CompiledPaths, bool), SwcImportResolver, ARandomState>> =
         Lazy::new(Default::default);
--- a/crates/swc/src/config/mod.rs
+++ b/crates/swc/src/config/mod.rs
@@ -1405,7 +1405,7 @@ impl ModuleConfig {
                 build_resolver(
                     base_url,
                     paths,
-                    config.resolve_fully,
+                    config.config.resolve_fully,
                     &config.config.out_file_extension,
                 )
             }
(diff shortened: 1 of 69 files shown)


hanseltime:
Noting here a few things that I'm unsure on:

1. The Miri tests are failing, but they seem to fail on things like test_fixtures macro, and "unimplemented functions".  If someone could set me straight on it, I can sit down and debug it if asked.
2. If I should be submitting a documentation change to the website in tandem with this

Thanks in advance!

crates/swc/src/config/mod.rs
```diff
@@ -1733,6 +1750,7 @@ fn build_resolver(
     mut base_url: PathBuf,
     paths: CompiledPaths,
     resolve_fully: bool,
+    file_extension: &String,
```
kdy1, line 1753: I think clippy will angry at this

kdy1:
```suggestion
    file_extension: &str,
```

crates/swc_ecma_transforms_module/src/system_js.rs
```diff
@@ -21,6 +22,9 @@ pub struct Config {
 
     #[serde(default)]
```
kdy1, line 23: I think you should remove this

crates/swc_ecma_transforms_module/src/util.rs
```diff
@@ -46,6 +46,13 @@ pub struct Config {
...
     #[serde(default)]
     pub resolve_fully: bool,
+
+    #[serde(default = "default_js_ext")]
+    pub out_file_extension: String,
+}
+
+pub fn default_js_ext() -> String {
+    "js".to_string()
 }
```
kdy1, line 56:
```suggestion
impl Config {
    pub fn default_js_ext() -> String {
        "js".to_string()
    }
}
```

kdy1 (commented):
Thank you!

crates/swc_ecma_transforms_module/src/system_js.rs
hanseltime: @kdy1 - I didn't add this attribute.  Did you mean the inner Config attribute below it or were you hoping I would clean this up?  Also, I'm assuming you would want this changed across all Module Shapes or just this one?

crates/swc/src/config/mod.rs
hanseltime: Ah thanks.  I have updated it!

crates/swc_ecma_transforms_module/src/system_js.rs
kdy1: I mean the whole `resolve_fully: bool`. It's now duplicated as you added a field below.

hanseltime: Oh I see.  Thanks for clarifying. Okay, I've updated that for the ESModuleConfig and the SystemJS Config.

crates/swc_ecma_transforms_module/src/util.rs
hanseltime: I've switched the function to an impl on Config and all calls to the function to use the `Config::default_js_ext()` as well.

Issue #9377, opened 2024-08-04
Fail to parse import attributes in TSTypeAlias


EliLichtblau:
### Describe the bug

Parse failure on assert in importType

[Oxc had the same bug](https://github.com/oxc-project/oxc/issues/4601)

### Input code

```typescript
type Vite =  typeof import("vite", { assert: { "resolution-mode": "import" } });
```

### Config

```json
{
  "jsc": {
    "transform": {
        "legacyDecorator": true,
        "decoratorMetadata": true,
        "react": {
          // runtime: "automatic", // or classic
          "runtime": "automatic", //
          "throwIfNamespace": true,
          "useBuiltins": true,
          "development": false,
        },
      },
      "loose": false,
      "parser": {
        "syntax": "typescript",
        "tsx": true
      },
      "target": "es2020",
  },
  "module": {
    "type": "commonjs"
  },
  "isModule": true
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.7.5&code=H4sIAAAAAAAAAyupLEhVCMssSVWwVVAoAXLy0xQycwvyi0o0lMqAwko6CtUKicXFqUUlVkCWUlFqcX5OaUlmfp5ubn5KqpKVghJEuZJCrUKtpjUA6VCkqlAAAAA%3D&config=H4sIAAAAAAAAA22RwW7CMAyG73uKyGckEEeOE5cd2DtYqQtBSVzZzkaFePc1Fe3Kupvj%2F7P927m%2FOQdX9XBw9yEcHiaYtWVJc2pMRzqj74%2FkWdBYBtGk0OYXaCbpRIYNGq4RIfT20ta57dZJyRYSHRxgMU5owcOmCizOR1QNflEATxxW%2FBKyi%2FD3R%2FuJibRDTyszA1OU3kuIFrL%2BJzf0RZG7RLl6bjHqQn9M4RxAZFb6S0KHoiSvt9Q%2BG97qAtZ3pF5CZ7C4k%2BntaWg1xFDOVP0A6X63341lowyJmxJp8ZFD7wp6TonzVWEmg54mdhzy%2BAHqvjFtBgIAAA%3D%3D

### Expected behavior

Should compile

### Actual behavior

Fails to compile

### Version

1.7.6

PR #9796, opened 2024-12-12
feat(es/ast): Add import attributes to `TsImportType`

dsherret:
**Description:**

This parses, emits, and visits import attributes in TsImportType.

**BREAKING CHANGE:**

 - Adds an additional property to `TsImportType`

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/9377


crates/swc_ecma_ast/src/typescript.rs
```diff
@@ -546,6 +546,8 @@ pub struct TsImportType {
     pub qualifier: Option<TsEntityName>,
     #[cfg_attr(feature = "serde-impl", serde(rename = "typeArguments"))]
     pub type_args: Option<Box<TsTypeParamInstantiation>>,
+    #[cfg_attr(feature = "serde-impl", serde(default))]
+    pub with: Option<Box<ObjectLit>>,
```
dsherret, line 550: I'm not sure about this name because this is the full `{ with: { "module-resolution": "import" } }` (which I need for dprint-plugin-typescript). Maybe something else would be better here?

crates/swc_ecma_parser/src/parser/typescript.rs
```diff
@@ -316,6 +316,18 @@ impl<I: Tokens> Parser<I> {
             }
         };
 
+        // the "assert" keyword is deprecated and this syntax is niche, so
+        // don't support it
```
dsherret, line 320: I don't think we should support the "assert" keyword for these reasons.

crates/swc_ecma_visit/src/generated.rs
dsherret, line 1: This file was difficult to figure out how to run (`cargo run --package generate-code -- --input-dir crates/swc_ecma_ast --output crates/swc_ecma_visit/src/generated.rs`). It might be good to add instructions somewhere.

Also, running that command generates a bunch of other things that I reverted because they were unrelated to this PR. I think maybe this file is out of date on main. It might be good to add some CI verification to ensure it stays up to date.

crates/swc_ecma_ast/src/typescript.rs
dsherret: I'll wait to update the failing tests until getting feedback on this.

kdy1: So `annotation`? I think we may need to use separate type for this kind of annotations. How do you think? cc @swc-project/core-es

magic-akari: Is `with` hardcoded?
I hope there is no situation like `{ with: { "foo": "bar" }, other: "value" }`.

CPunisher: > The expected type of that second argument is defined by a type called `ImportCallOptions`, which by default just expects a property called `with`.

And typescript playground says `Object literal may only specify known properties, and 'xxx' does not exist in type 'ImportCallOptions'.`

So I think some names related to `ImportCallOptions` are ok as well.

kdy1: I think we need a breaking change anyway if there's a situation like `{ with: { "foo": "bar" }, other: "value" }`

dsherret: I updated to this (I think)

dsherret:
Could someone re-run the test expectations for me? I had so much trouble getting yarn 4 running on my machine and gave up.

kdy1:
I’ll do it later today

crates/swc_ecma_visit/src/generated.rs
kdy1: I know this is not intuitive, but you can run `UPDATE=1 cargo test` from `tools/generate-code`.
I think we should add an alias for this and document it.

dsherret:
Thanks! I think this PR should be good to go now.

kdy1:
Is this urgent? As this breaks Wasm ABI I prefer to merge it along with other breaking changes.

dsherret:
Nope, not urgent.

kdy1 (dismissed):

kdy1 (approved):

kdy1 (approved):

PR #9803, opened 2024-12-18
fix: don't panic when wasm bytecheck faild

hardfist:
**Description:**
Currently when wasm plugin is not compatible with swc version it will panic which cause it hard to report useful error info from caller except use catch_unwind.
change the behavior to not panic and let the user to decide error handling 

**BREAKING CHANGE:**
No

### DIFF
--- a/.changeset/new-apricots-exercise.md
+++ b/.changeset/new-apricots-exercise.md
@@ -1,0 +1,6 @@
+---
+swc: patch
+swc_core: patch
+---
+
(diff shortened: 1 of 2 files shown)


kwonoj (approved):

hardfist:
https://github.com/swc-project/swc/actions/runs/12397595140/job/34608511623?pr=9803#step:13:3559 
it seems the test failing cause removing the extra comma(don't know why there's a extra comma here)

PR #9807, opened 2024-12-19
perf(es/lexer): Use `logos` lexer as a sub-lexer

kdy1:
**Description:**

 - Define `RawToken`.
   - It should be declared in a standalone crate because it requires heavy code generation. `logos` generates a deterministic finite state machine that consists of lookup tables and jump tables.
   - In the future, `RawToken` will be renamed to `Token` and used directly by the parser. I need to investigate much more about API.
   - For performance, `RawToken` should be single-byte-sized.

 - Adjust lexer to work on `RawToken` instead of `char`.
   - `logos` takes `&str` and generates `Iterator<Item = RawToken>`.
   - Some ECMAScript code cannot be lexed using FSM. `logos` provides [callback API](https://logos.maciej.codes/callbacks.html), and that's how we should handle ambiguous tokens.
   - The regex syntax of `logos` is a bit inferior to that of `regex`. In other words, even the regex is valid, the logos lexer may not generate matching `RawToken`. This is for performance, and it's documented [here](https://logos.maciej.codes/common-regex.html).

 - Wrapper: `logos::Lexer` => `RawLexer` => `Lexer` => `Parser`
   - The callback API of logos is not enough for lexing ECMAScript. We have to wrap the logos lexer with our lexer and call logos lexer only if it's valid to do. Well, it's valid most of the time, but currently, we handle regex and template literals using a separate method.
   - I introduced `RawLexer` as a sort of _buffer_ (based on [`peek_nth`](https://docs.rs/itertools/latest/itertools/fn.peek_nth.html) from [`itertools`](https://docs.rs/itertools/latest/itertools/)), but  I found that it's a good place to have various lexing methods, so I added `read_regexp`. `read_regexp` uses another logos token definition, so it should be in the `swc_ecma_raw_lexer` crate.

- [ ] Fix tests
  - [ ] Currently, some of spans are wrong.
  - [ ] Currently, _processed_ values of AST are simply filled with the _raw_ value from the lexer. These should have the correct values. For example, `Str` has the value of `\\\\` and the raw value of `\\\\` for input string `\\\\`.  But the value field should be `\\` instead.


crates/swc_ecma_parser/src/lexer/state.rs
```diff
@@ -372,7 +353,7 @@ impl Iterator for Lexer<'_> {
             }
 
             self.state.update(start, token.kind());
-            self.state.prev_hi = self.last_pos();
+            self.state.prev_hi = self.input.cur_pos();
```
GiveMe-A-Name, line 356: We do not need maintain the cur_pos message in parser.lexer. How about we can get `span` from logos.lexer ?  It's a better way to reduce ecma.parser complexity.

kdy1: I agree that it can be a better design. You would only need to add `start_pos` to those.

crates/swc_ecma_raw_lexer/src/jsx.rs
```diff
@@ -0,0 +1,7 @@
+use logos::Logos;
+
+use crate::LogosError;
+
+#[derive(Logos, Debug, Clone, Copy, PartialEq, Eq)]
+#[logos(error = LogosError)]
+pub enum JsxToken {}
```
GiveMe-A-Name, line 7: `<` in jsx maybe `BinOpToken::Lt` or `JSXTagStart`, logos lexer is so hard using a kind enum to express two token kind. So I think we don't generate jsx token by logos lexer. 
We can think of logos lexer as a basic lexer that generate basic token kind such as `LtAngle` ....
Lexer using logos and state to generate more specific Token.
Do you think it a better way?

kdy1: Yeap, that's the way I was thinking of.

PR #9818, opened 2024-12-28
fix(es/minifier): Fix minification of `framer-motion` by checking `cons.termniates()`

sjchmiela:
**Description:**

Next.js 15 seems to minify parts of `framer-motion`'s code in an invalid way, changing the meaning. The original function modifies the single `definition` variable as it changes meaning (see `input.js`):
- if `definition` is a function, it calls it and saves the return value as `definition`
- if `definition` is a string, it resolves animation variant based on value as name and saves the variant as `definition`
- if `definition` is again a function, it calls it again and saves the return value as `definition` again

However, `swc` seems to be not noticing that `definition` may be mutated between the calls and collapses all `if`s into one:

```
// `resolveVariantFromProps` in framer-motion
// https://github.com/motiondivision/motion/blob/ecd97f7dce8954be300aa73ab6a96208437941c5/packages/framer-motion/src/render/utils/resolve-variants.ts#L31-L72
export function resolveVariantFromProps(props, definition, custom, visualElement) {
    if ("function" == typeof definition || ("string" == typeof definition && (definition = props.variants && props.variants[definition]), "function" == typeof definition)) {
        const [current, velocity] = getValueState(visualElement);
        definition = definition(void 0 !== custom ? custom : props.custom, current, velocity);
    }
    return definition;
}
```

which skips the first call to `definition`, first "resolution" step.

Note: I suspect the `output.js` may not be precisely what SWC may output as minified code (maybe there is a way to minify it further). It definitely differs from the current output though by the initial call and reassignment to `definition`.

### DIFF
--- a/.changeset/gold-shoes-cross.md
+++ b/.changeset/gold-shoes-cross.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_minifier: patch
+---
+
+fix(es/minifier): Add failing test case for an example in `framer-motion`
--- a/crates/swc_ecma_minifier/src/compress/optimize/conditionals.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/conditionals.rs
@@ -121,7 +121,9 @@ impl Optimizer<'_> {
                     (
                         Some(Stmt::If(l @ IfStmt { alt: None, .. })),
                         Some(Stmt::If(r @ IfStmt { alt: None, .. })),
-                    ) => SyntaxContext::within_ignored_ctxt(|| l.cons.eq_ignore_span(&r.cons)),
+                    ) => SyntaxContext::within_ignored_ctxt(|| {
+                        l.cons.eq_ignore_span(&r.cons) && l.cons.terminates()
+                    }),
                     _ => false,
                 });
         if !has_work {
--- a/crates/swc_ecma_minifier/tests/fixture/issues/framer-motion/1/input.js
+++ b/crates/swc_ecma_minifier/tests/fixture/issues/framer-motion/1/input.js
@@ -1,0 +1,28 @@
+// `resolveVariantFromProps` in framer-motion
+// https://github.com/motiondivision/motion/blob/ecd97f7dce8954be300aa73ab6a96208437941c5/packages/framer-motion/src/render/utils/resolve-variants.ts#L31-L72
+
+export function resolveVariantFromProps(props, definition, custom, visualElement) {
+    if (typeof definition === "function") {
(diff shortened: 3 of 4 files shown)


kdy1:
Thank you for reducing the edge case! Actually it's the most hard part of minifier work...

Austaras:
A more simple [example](https://play.swc.rs/?version=1.10.4&code=H4sIAAAAAAAAA0utKMgvKlFIK81LLsnMz1NI1EjSUUjWVKjmUgCCzDQFjSQYBwSS8%2FOKSxRSFGwVkhS0FQxRxPNzUvVy8tM1UnQUUjTBMrVccFOSKTGlFgAly6YYpgAAAA%3D%3D&config=H4sIAAAAAAAAA32VO5LbMAxA%2Bz2FR3WKjIsUOUC6nIFDk6BMRyQ0BOi1Zsd3D%2FTzZw25s%2FDwBwh%2Ffex2zYlc83v3JT%2Flo7eFoNy%2BRUJDZnsRSQMuWXIl9tz8WOmJRhRsRzCJrjNp2JYWeLKi%2Fc%2F9vlnlxWYKWNJjiALW8YNgFNXMMcHowFbGZDm6ZsHXp0gdIsGawyJLMccwPIZwmPoCRM9RJMmaIDM92y%2Bs4OcIuNRH%2BQGxAylimxhLJmaGForm2GHX2Z7AnG1RvIyZ2hIJtRAjrAze9AV7lWcfOWKWmK%2FUg%2FXGoQcFxQKO4xk0M4klZpmkPKWeCXs41LadNuebNZxtVy0rMeEyjUSyVbweMRKbULPWwhlu9GCGS3O%2FW8ZgCnAt%2BdXuhDFvzOQfgHSgs0TZJtD8ThpB9mnLOry1jDnIyvKgcNlvrcoMrTTVxBiUzo6dgcJRm2YBXx2MnXVaOgveaB9FDwZCkF1RXNNnZHfUgvLQAwYFyHxt0LZqBub2Cjf4%2BCDe4D9SJesLtmjIZTluUxrSAbs3ARLwEf0bBRkF4zYuciUu%2FTav2YOsBnhVpdIEXo%2BAPABG000X%2BGU35HmIR9N2eLififW23i57srm9v%2Ff55H4sCk1CXye4%2FG2M852P%2Fa%2FmrrRe4VviTaS%2Fq%2BUU9PofNvYYEYIGAAA%3D) of why we need to check if `cons` terminates.

kdy1 (approved):
Thank you!

PR #9843, opened 2025-01-06
perf(es/minifier): Speed up `merge_sequences_in_exprs` by caching computation

CPunisher:
## Problem

In https://github.com/chenjiahan/rspack-swc-minify-test:
Rspack + SWC minify: 15.89 s
Rspack + esbuild minify: 1.74 s

This is caused by `merge_sequences_in_exprs` which tries to merge every pair of exprs with time complexity of O(n^2). In the given example, a vec of 5003 exprs (react component declarations and other declarations) are passed to this function. 

From the profiling result below we can see `visit_children_with` takes a lot of times, which is called by `IdentUsageFinder` and `idents_used_by` to check whether an ident is used in an ast node. However, in the O(n^2) loops, this part of the result is heavily recalculated.

Example:
```js
// input.js
export function source() {
    let c = 0, a = 1;
    c += a, a += 5;
    let b = c;
    console.log(a, b, c);
}

// esbuild/terser output
export function source(){let o=0,e=1;o+=e,e+=5,console.log(e,o,o)}

// swc output
export function source(){console.log(6,1,1)}
```

## Solution

To be honest, I don't come up with a better algorithm to reduce time complexity or pruning. But caching part of the computation does work very well in this bad case. Anyway, I'm proposing this pr first.

## Result

**Before: 10128ms**  
image

**After: 1013ms(~10x)**
image

### DIFF
--- a/.changeset/forty-maps-smoke.md
+++ b/.changeset/forty-maps-smoke.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_minifier: patch
+---
+
+perf(es/minifier): Speed up `merge_sequences_in_exprs` by caching computation
--- a/crates/swc_ecma_minifier/src/compress/optimize/sequences.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/sequences.rs
@@ -987,11 +987,11 @@ impl Optimizer<'_> {
                                 if !self.is_skippable_for_seq(Some(a), e2) {
                                     break;
                                 }
-                            }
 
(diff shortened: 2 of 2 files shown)


kdy1 (approved):
CI failed, and I verified locally that CI failure does not happen on the main branch.

I love the idea behind this PR. I think it's fine to simply update test cases of the `swc` crate in this case, considering the performance difference and the amount of difference in outputs. Can you update or fix them? You can do either of them.

kdy1 (approved):
Thank you!

PR #9846, opened 2025-01-06
feat(ts/fast-strip): Distinguish invalid vs unsupported

kdy1:
**Description:**

I tried adding a property to the error thrown, but `wasm_bindgen` does not support adding property to an error nor creating different kinds of errors.

So I throw an object with `code` and `message` instead.


marco-ippolito:
So when it has unsupported syntax it will have the label `help: unsupported syntax`.
That means I have to match with a regex which I think is not very efficient.
I'd rather not have it honestly if I cannot access it efficiently

kdy1:
@marco-ippolito Is it fine to throw an object? I think `wasm_bindgen` would allow creating object with arbitrary properties.

(Not sure if it supports throwing them, though)

marco-ippolito:
> @marco-ippolito Is it fine to throw an object? I think `wasm_bindgen` would allow creating object with arbitrary properties.
> 
> (Not sure if it supports throwing them, though)

I think so since we always wrap what is thrown into a Node error. Let's give it a try

marco-ippolito:
> @marco-ippolito Can you take a look on the new error output?

So I'd expect always an object with
code and message
Code could be either InvalidSyntax or UnsupportedSyntax
Message the usual error message

kdy1:
Yeap, I added some test cases

marco-ippolito (approved):

bindings/binding_typescript_wasm/__tests__/transform.js
```diff
@@ -138,5 +138,24 @@ describe("transform", () => {
                 })
             ).rejects.toMatchSnapshot();
         });
+
```
robpalme, line 141: Unrelated to this PR, I'm surprised there is no test for Parameter Properties or "import/export equals" or angle-brackets-type-assertion in this file, given the enum and namespace cases are covered here.

crates/swc_fast_ts_strip/src/lib.rs
```diff
@@ -1062,28 +1121,34 @@ impl Visit for TsStrip {
...
     fn visit_ts_namespace_decl(&mut self, n: &TsNamespaceDecl) {
         HANDLER.with(|handler| {
-            handler.span_err(
-                n.span(),
-                "TypeScript module declaration is not supported in strip-only mode",
-            );
+            handler
+                .struct_span_err(
+                    n.span(),
+                    "TypeScript module declaration is not supported in strip-only mode",
```
robpalme, line 1149: I don't see any test fixture for this case. And I'm confused why this is needed.

crates/swc_fast_ts_strip/tests/errors/enums.swc-stderr
```diff
@@ -3,9 +3,11 @@
  1 | enum Foo { }
    : ^^^^^^^^^^^^
    `----
+  help: Unsupported syntax
```
robpalme, line 6: This whole directory of tests is missing a file to cover Parameter Properties.

bindings/binding_typescript_wasm/__tests__/transform.js
kdy1: Actually main tests are done at https://github.com/swc-project/swc/tree/caa7f370ff2003983a3448a2b7e6f0f6d6224b86/crates/swc_fast_ts_strip/tests/fixture

This file tests only the JS interface

crates/swc_fast_ts_strip/src/lib.rs
robpalme: This whole visit still seems redundant given that it does not appear in any of the test fixtures.

But that is a pre-existing condition, so I am approving anyway.

crates/swc_fast_ts_strip/tests/errors/parameter-property.swc-stderr
```diff
@@ -0,0 +1,14 @@
+  x TypeScript parameter property is not supported in strip-only mode
+   ,-[5:1]
+ 4 | class Foo {
+ 5 |     constructor(public id: string) { }
+   :                        ^^^^^^^^^^
```
robpalme, line 5: The ^^^ should point to the `public` keyword only.

robpalme (approved):
Thanks for adding the new tests.

My comments are non-blocking.

marco-ippolito:
Is this change included in the latest nightly?

kdy1:
I triggered a full publish action for `1.10.7` (stable).

https://github.com/swc-project/swc/actions/runs/12703698875

kdy1:
@marco-ippolito It's published

PR #9864, opened 2025-01-11
perf(es/minifier): Use `FastId` for minifier

kdy1:
**Description:**


kdy1:
This looks like a bad idea

PR #9865, opened 2025-01-11
perf(es/minifier): Make DCE analyzer parallel

kdy1:
**Description:**

The analyzer of the DCE pass can be parallelized in a similar way to mark-sweep GC. It currently uses `&mut petgraph::DiGraphMap` to find reachable bindings, but we can split the graph creation into multiple threads.

Also, I decided to amort allocations because `swc_parallel::join` is called an enormous amount of time, meaning that it will be slower if I allocate one hashmap from each closure. I'll use [thread_local **crate**](https://docs.rs/thread_local/latest/thread_local/) to reuse the hashmap and collect all of them after the traversal is done.

### Profiling result
```
```

### DIFF
--- a/crates/swc_ecma_transforms_optimization/src/simplify/dce/mod.rs
+++ b/crates/swc_ecma_transforms_optimization/src/simplify/dce/mod.rs
@@ -262,7 +262,14 @@ impl Parallel for Analyzer<'_> {
 
         self.scope
             .bindings_affected_by_eval
+            .reserve(other.scope.bindings_affected_by_eval.len());
+        self.scope
+            .bindings_affected_by_eval
             .extend(other.scope.bindings_affected_by_eval);
+
+        self.scope
+            .bindings_affected_by_arguements
+            .reserve(other.scope.bindings_affected_by_arguements.len());
         self.scope
             .bindings_affected_by_arguements
             .extend(other.scope.bindings_affected_by_arguements);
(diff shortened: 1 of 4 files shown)


crates/swc_ecma_transforms_optimization/src/simplify/dce/mod.rs
```diff
@@ -1111,6 +1203,87 @@ impl VisitMut for TreeShaker {
     }
 }
 
+fn merge_data(data: Arc<ThreadLocal<RefCell<Data>>>) -> Data {
```
kdy1, line 1206: Actually, this function is slow and I want to optimize this function further.

kdy1 (commented):
Any ideas to improve the performance even more?

kdy1 (commented):
Any ideas to improve the performance even more?

kdy1:
I tried profiling, but it seems to have almost no effect on current benchmark suite.

kdy1:
With this PR:

```
Using 1970 files
Took 7.039294875s
```

CPunisher:
I just comp up with some subtle points:
1. Wrap TreeShaker's config in `Arc` to reduce the allocation of `Parallel::create`
2. Pre-allocate memory for `entries` in `substract_cycles` with `self.entry_ids.len()`
3. Let `j` start after `i` in `substract_cycles` to reduce the number of cycles by half

From the profiling result, the main cost comes from too many HashMaps with their hash calculation. But I didn't figure out any optimization.

kdy1:
This can be optimized by merging values of the hashmap in parallel and extending the map later. I'll implement it later.

kdy1:
Closing in favor of https://github.com/swc-project/swc/pull/10076 (I want to use graphite)

PR #9874, opened 2025-01-13
perf(es/minifier): Make infection collector parallel

kdy1:
**Description:**

It seems like an easy one.


kdy1:
Closing due to perf regression

PR #9902, opened 2025-01-19
perf(es/minifier): Do heavy operation only if required

kdy1:
**Description:**

In `store_var_for_inlining`, we had called `inline_with_multi_replacer` needlessly. This PR fixes it.


crates/swc_ecma_minifier/src/compress/optimize/inline.rs
```diff
@@ -89,7 +89,7 @@ impl Optimizer<'_> {
             let is_inline_enabled =
                 self.options.reduce_vars || self.options.collapse_vars || self.options.inline != 0;
 
-            self.vars.inline_with_multi_replacer(init);
```
kdy1, line 92: I'm not sure if this is really not required

kdy1: Okay, this covers all branches that we store AST in `self.vars*`

PR #9915, opened 2025-01-21
fix(es/parser): Parse `yield<T> (v: T)=>v`

### LINKED ISSUE #9914: SWC should parse `yield<T> (v: T)=>v`
### Describe the bug

sub issue of #9878 

### Input code

```typescript
(function* () {
    yield <T>(v: T) => v;
});
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2022"
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

https://play.swc.rs/?version=1.10.9&code=H4sIAAAAAAAAA9NIK81LLsnMz9NS0NBUqOZSAILKzNScFAWbEDuNMiuFEE0FWzuFMmuuWk1rAMYyBAQtAAAA&config=H4sIAAAAAAAAA02NOQ6EMAxF%2B5wick2BUlBwBw4RgUFBLJFtpIkQd58kkBk6%2F%2FcXn0prmLmHVp%2FxjMJbYqSfjoTDJvYTCUjwyD05L1AVVzhZo10YM7puB8TShJJayKY2BtTjwboPx4L%2Fj2n1zjWvkNvcGMpyRo670hQ6UF1fU15err0AAAA%3D

### SWC Info output

_No response_

### Expected behavior

SWC should parse successfully.

### Actual behavior

_No response_

### Version

all

### Additional context

_No response_

magic-akari:
**Related issue:**

- Closes: #9914

### DIFF
--- a/crates/swc_ecma_parser/src/parser/expr.rs
+++ b/crates/swc_ecma_parser/src/parser/expr.rs
@@ -39,34 +39,32 @@ impl<I: Tokens> Parser<I> {
     pub(super) fn parse_assignment_expr(&mut self) -> PResult<Box<Expr>> {
         trace_cur!(self, parse_assignment_expr);
 
-        if self.input.syntax().typescript() && self.input.syntax().jsx() {
+        if self.input.syntax().typescript() && is!(self, JSXTagStart) {
             // Note: When the JSX plugin is on, type assertions (`<T> x`) aren't valid
             // syntax.
 
-            if is!(self, JSXTagStart) {
-                let cur_context = self.input.token_context().current();
-                debug_assert_eq!(cur_context, Some(TokenContext::JSXOpeningTag));
-                // Only time j_oTag is pushed is right after j_expr.
-                debug_assert_eq!(
-                    self.input.token_context().0[self.input.token_context().len() - 2],
-                    TokenContext::JSXExpr
-                );
+            let cur_context = self.input.token_context().current();
+            debug_assert_eq!(cur_context, Some(TokenContext::JSXOpeningTag));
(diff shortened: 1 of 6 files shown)


crates/swc_ecma_parser/src/parser/expr.rs
```diff
@@ -39,34 +39,32 @@ impl<I: Tokens> Parser<I> {
     pub(super) fn parse_assignment_expr(&mut self) -> PResult<Box<Expr>> {
         trace_cur!(self, parse_assignment_expr);
 
-        if self.input.syntax().typescript() && self.input.syntax().jsx() {
+        if self.input.syntax().typescript() && is!(self, JSXTagStart) {
```
kdy1, line 42: Does this mean we now produce `Token::JSXTagStart` even when `jsx()` is not enabled?

magic-akari: Actually, when we encounter `JSXTagStart`, it indicates that JSX is enabled. The current implementation has just one fewer condition compared to the previous code, but logically, it should be equivalent in functionality.

This code ended up in its current state due to my rebase and squash process. However, upon reviewing it, I realize this part doesn’t actually need to be changed. If needed, I’m happy to revert this change. Please let me know your thoughts!

kdy1: I think the previous check is needless, as the existence of a such token proves `self.input.syntax().jsx() == true`. Thank you for clarification!

kdy1 (approved):

PR #9924, opened 2025-01-22
perf(es/minifier): Make analyzer not call `collect_infects_from` recursively

kdy1:
**Description:**

We don't need to make time complexity non-linear by recursively visiting the effects of infection. Instead, suppose we stop the infection analyzer when the usage analyzer invokes the infection analyzer. In that case, we will have practically identical analysis results because the points are named using an id, and it points to all the infections that occurred by referencing the id.

**Related issue:**
 - Closes https://github.com/swc-project/swc/issues/9927


kdy1:
cc @CPunisher Do you have an idea for better caching? `*const Function` does not look good to me.

CPunisher:
> cc @CPunisher Do you have an idea for better caching? `*const Function` does not look good to me.

The point is to determine an **unique** id for a function node. Maybe `Ident`, `Span`...
But does it handle a node many times?

kdy1:
@CPunisher Yeap. I wrote about them in https://github.com/swc-project/swc/issues/9927.

While visiting nodes for usage analysis, it calls `collect_infects_from` on the node, but while recursing into a node from the usage analyzer, it calls `collect_infects_from` on a child node again.

CPunisher:
> @CPunisher Yeap. I wrote about them in #9927.
> 
> While visiting nodes for usage analysis, it calls `collect_infects_from` on the node, but while recursing into a node from the usage analyzer, it calls `collect_infects_from` on a child node again.

I see. If it is hard to cache the result, can we **avoid** repeated analysis? That means we can stop collect_infects_from going deeper since we know we will handle inner nodes later.

kdy1:
@CPunisher  I think I managed to find something close to not recursively calling them.

kdy1:
Diff: 
image

PR #9942, opened 2025-01-24
perf(es/minifier): Make character frequency calculation more parallel

kdy1:
**Description:**

We join two operations. The first operation is codegen, and the second operation is visiting. The second operation is done in parallel, but it may stall in some cases. If we perform codegen in parallel, the threads that have nothing to do while visiting will get some tasks of codegen.


kdy1:
Closing as this should not have thing to do with tests

PR #9980, opened 2025-01-31
chore: run generate-code

dsherret:
**Description:**

Runs `cargo run --package generate-code -- --input-dir crates/swc_ecma_ast --output crates/swc_ecma_visit/src/generated.rs` which seemed to be outdated.

https://github.com/swc-project/swc/pull/9796#discussion_r1882527713


dsherret:
Seems the code generation is wrong. I'm just going to revert the output in my PR to focus only on the changes I'm making.

Issue #9904, opened 2025-01-20
ReferenceError: Must call super constructor in derived class before accessing 'this' or returning from derived constructor


nanianlisao:
### Describe the bug

Problem syntax error caused by using this before Super

### Input code

```typescript
class StartNode extends React.Component {
  constructor() {
    const s = <Component />;
    super()
  }
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": true,
    },
    "transform": {
      "react": {
        "runtime": "classic",
        "development": true,
      },
    },
  },
  "env": {
    "targets": "chrome 100",
  },

  "module": {
    "type": "commonjs",
  },
  "sourceMaps": false,
  "filename": "/Users/view.tsx",
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.10.8&code=H4sIAAAAAAAAA0vOSSwuVgguSSwq8ctPSVVIrShJzUspVghKTUwu0XPOzy3Iz0vNK1Go5lJQSM7PKy4pKk0uyS%2FS0ASLQMUUihVsFWwQivXtrMGSxaUFqUClQHYtVy0Ak6yhH2oAAAA%3D&config=H4sIAAAAAAAAA1WQTRKCMAyF956CydpR3HoHlx4gU4KW6d8kAWUY724LirhpmvfSL2mmXVVBJwbO1ZSvOUnIQrzmWZExKD6zAmQ8imGbFPZft5NiKfe0SK8lgDIGaSP7LYoJjW6EIvVBraeCNw5FrFnZ2W1oIBeTp6B%2FXdY%2Bc5gPoDD8vqHIN1KZsXeOnqpTXcOnuFT72PSONg%2FGtAwRvY%2BhE%2FiBJfZs6IKp4Fp0Mg8BrXUUcBn9eM1Lk%2BNg6XHQvJH97vUGRsh9mlsBAAA%3D

### SWC Info output

"use strict";
class StartNode extends React.Component {
    constructor(){
        const s = /*#__PURE__*/ React.createElement(Component, {
            __source: {
                fileName: "input.jsx",
                lineNumber: 3,
                columnNumber: 15
            },
            __self: this
        });
        super();
    }
}

### Expected behavior

Hopefully, like babel, there will be no __self

var _jsxFileName = "/Users/view.tsx";
class StartNode extends React.Component {
  constructor() {
    const s = /*#__PURE__*/React.createElement(Component, {
[...]

PR #9987, opened 2025-02-03
fix(es/transforms): Avoid adding `__self` in constructors of derived class in the `jsx_self` react transform

CPunisher:
fixes https://github.com/swc-project/swc/issues/9904

1. port babel test: https://github.com/babel/babel/tree/main/packages/babel-plugin-transform-react-jsx-self/test/fixtures/react-source
2. check if the jsx is in a constructor of a derived class: https://github.com/babel/babel/blob/main/packages/babel-plugin-transform-react-jsx-self/src/index.ts#L77


kdy1 (approved):
THank you!

kdy1 (approved):

Issue #9981, opened 2025-01-31
optimizer.simplify: true is wrongly stripping code it shouldn't


dalimian:
### Describe the bug

given this code `const someVar = someNonExistentVarSoThisCodeShouldTriggerAnError;`

when compiled with `jsc.transform.optimizer.simplify` set to true in `.swcrc` the above code is dropped from the compiled output.  that seems wrong, optimization should not change any behavior. The end effect is that SWC is essentially silencing what would have been a loud failure.

side question, is `optimizer.simplify`  feature considered stable

### Input code

```typescript

```

### Config

```json
{
  "$schema": "https://swc.rs/schema.json",
  "env": {
    "loose": true,
    "targets": [
      "current node"
    ],
  },
  "jsc": {
    "baseUrl": "../../../",
    "parser": {
      "syntax": "ecmascript",
      "jsx": true
    },
    "transform": {
      "react": {
        "development": false,
      },
      "optimizer": {
        "simplify": true,
        "globals": {
          "vars": {
            "NODE_ENV": "true"
          }
        }
      }
    },
    "paths": {
      "site-modules/*": [
        "./client/site-modules/*"
      ],
      "client/*": [
        "./client/*"
      ],
      "server/*": [
        "./server/*"
      ],
      "shared/*": [
        "./shared/*"
      ],
      "stubs/*": [
        "./stubs/*"
      ],
      "archetype/*": [
        "./archetype/*"
      ],
      "package.json/*": [
        "./package.json/*"
      ]
    },
  },
  "module": {
    "type": "commonjs",
  }
}
```

### Playground link (or link to the minimal reproduction)

[...]

PR #9989, opened 2025-02-03
fix(es/transforms): Pass `unresolved_mark` to `simplifier` instead of `top_level_mark`

CPunisher:
fixes https://github.com/swc-project/swc/issues/9981

I'm not sure if this mistake is intentional.


kdy1 (dismissed):
Thanks! It's a mistake from old days

PR #10016, opened 2025-02-08
perf(ecma/parser): swc ecma lexer

GiveMe-A-Name:
**Description:**
TODO:

**BREAKING CHANGE:**


GiveMe-A-Name:
refactor swc parser on swc_ecma_fast_parser

PR #10022, opened 2025-02-11
fix(ts/fast-strip): Handle unsupported `module` keyword

### LINKED ISSUE #10017: Option to error on `module` namespaces
### Describe the feature

This feature request is motivated by Node's TypeScript integration. See https://github.com/nodejs/amaro/issues/168 for the backstory.

The request is for SWC to introduce a TypeScript-specific option (example:  `no_module_keyword_for_ts_namespace`) to error when the source text includes any form of namespace using the legacy `module` keyword:

- `module Foo {}`
- `declare module Foo {}`

This applies to both the ambient and non-ambient forms.  And applies to both instantiated and uninstantiated forms.

https://github.com/nodejs/amaro/issues/168

### Exclusions

_Ambient module declarations_ (`declare module "specifier" {}`) are fine and should continue to be erased and not error.

### Amaro Usage

The initial user of this feature will be Amaro in Node.  Node needs to be able to show a helpful message to the user.  The returned error should identify this specific error so that we can display a human-readable message such as:

>`module` keyword is not supported. Use `namespace` instead.

magic-akari:
**Related issue:**

- Closes: #10017


magic-akari:
The implementation is currently in `type-strip` as it allows us to directly access tokens and source code text, minimizing the risk of AST breaking changes.

It’s unclear whether we can achieve the same in the `Transform` pass, but it is something we need to try.

@kdy Please let me know your thoughts. Thank you!

kdy1:
I think I need to make a AST-breaking change tomorrow, so please feel free to work on it. Please tell me if you prefer to make the change with another PR

magic-akari:
I've found a solution that preserves our current AST structure and avoids introducing any breaking changes. It's straightforward to pass the source text to the Transform pass.

However, if we decide to update the AST, we might consider adding a field to specify the `module` or `namespace`. This could also benefit other libraries (such as dprint), though I'm not completely certain about its overall impact.

Which do you prefer?

magic-akari:
I’ve created PR #10023. Please note that this PR is independent of #10023. Once #10023 is merged, we can simplify the namespace detection method accordingly.

magic-akari:
This is a bit awkward. We could have simply removed the top-level namespace, but we had to dig into the internal nodes to catch the error.

```TypeScript
namespace Foo {
    namespace Bar {
        module Boom { }
    }
}
```

crates/swc_fast_ts_strip/src/lib.rs
```diff
@@ -239,14 +243,24 @@ pub fn operate(
...
+    let deprecated_ts_module_as_error = options.deprecated_ts_module_as_error.unwrap_or_default();
+
     match options.mode {
         Mode::StripOnly => {
             let mut tokens = RefCell::into_inner(Rc::try_unwrap(tokens).unwrap());
 
             tokens.sort_by_key(|t| t.span);
 
+            if deprecated_ts_module_as_error {
+                program.visit_with(&mut ErrorOnTsModule {
```
kdy1, line 255: Is this really worth it? I'm asking because visitors have problems with the performance and the binary size.

kdy1: Feel free to resolve this if you think it's too complex to merge the visitor logic of `ErrorOnTsModule` into the original visitor

magic-akari: Let me give this a try.

magic-akari: Currently, our TsStrip only visits top-level namespaces/modules and skips nested ones, while our ErrorOnTsModule visits both top-level and nested namespaces/modules and skips all other structures.

These behaviors are contradictory.

I attempted to wrap both in an external Visitor that encapsulates TsStrip and ErrorOnTsModule and switches state when entering or leaving a namespace. However, this approach would require proxying all methods of TsStrip and ErrorOnTsModule—TsStrip alone has 46 visitor methods that need to be proxied. This adds considerable complexity and increases the potential for errors.

magic-akari: Give up trying and leave things as they are.

kdy1: Fair enough 👍

magic-akari: We need to take a systematic approach to addressing this issue in #9984.

kdy1 (approved):

Issue #10002, opened 2025-02-06
`es/ast`: `ExprOrSpead#spread` can be `Option<BytePos>`


kdy1:
And after then, we can use `SmallVec` for free.

PR #10026, opened 2025-02-12
perf(es/ast): Use smallvec and reduce size if possible

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/10001
 - Closes https://github.com/swc-project/swc/issues/10002
 - Example of update for Wasm plugin authors: https://github.com/swc-project/plugins/pull/412
 - Example of update for custom SWC binary: https://github.com/vercel/next.js/pull/76104


kdy1:
Numbers for RSS of the SWC minifier as standalone:

`main`:
```
   process: user: 5.075 s, system: 1.249 s, faults: 2, rss: 277.2 MiB, commit: 1.9 GiB
```

With small vector optimization:
```
   process: user: 4.944 s, system: 1.193 s, faults: 0, rss: 267.2 MiB, commit: 2.0 GiB
```

The gain is only 10MB, so I think this optimization does not worth the effort

Issue #9887, opened 2025-01-16
[swc plugin] The handler.struct_span_warn does not emit warnings if the plugin transform completes without errors


gaoachao:
### How to run the demo

demo: https://github.com/gaoachao/swc-plugin-demo

```shell
pnpm install
pnpm run build
npx swc ./index.js
```

### Current behavior

- When the transform is successful, warnings are not printed.
- When there is an error during the transform, warnings are printed.

### Expected behavior

When `npx swc index.js` succeeds, warnings should still be printed in the terminal.

## How to Reproduce the Issue
Here is my code, and no warnings are printed.
```rust
"createSelectorQuery" => {
    HANDLER.with(|handler| {
        handler
            .struct_span_warn(
                n.span,
                format!(
                    "WARNING: {} is deprecated.",
                    id.sym.to_string()
                )
                .as_str(),
            )
            .emit()
    });
}
"getElementById" => {
    HANDLER.with(|handler| {
        handler
            .struct_span_warn(
                n.span,
                format!(
                    "WARNING: {} is deprecated.",
                    id.sym.to_string()
                )
                .as_str(),
            )
            .emit()
    });
}
```

If I change one of the `struct_span_warn` calls to `struct_span_err`, both warnings and errors will be printed on the terminal.

```rust
"createSelectorQuery" => {
    HANDLER.with(|handler| {
        handler
            .struct_span_err(
                n.span,
                format!(
                    "WARNING: {} is deprecated.",
                    id.sym.to_string()
                )
                .as_str(),
            )
            .emit()
    });
}
"getElementById" => {
    HANDLER.with(|handler| {
        handler
            .struct_span_warn(
                n.span,
                format!(
                    "WARNING: {} is deprecated.",
                    id.sym.to_string()
                )
                .as_str(),
            )
            .emit()
    });
}
```

# Input code

```typescript
this.createSelectorQuery();
this.getElementById();
```

# Config

```json
{
  "$schema": "https://swc.rs/schema.json",
  "isModule": true,
  "jsc": {
    "parser": {
      "syntax": "ecmascript"
    },
    "target": "es2022",
    "experimental": {
      "plugins": [["./my_first_plugin.wasm", {}]]
    }
  }
}
```

## Playground link (or link to the minimal reproduction)

https://github.com/gaoachao/swc-plugin-demo

## Expected behavior

When the transform is successful, warnings should still be printed in the terminal.

## Version

[...]

PR #10027, opened 2025-02-12
fix(error-reporters): Store diagnostics in `TransformOutput`

GiveMe-A-Name:
**Description:**

SWC will return a `Result<Error>` to js, then js console message then throw error, when swc apply pass occur errors.
But SWC would not print warning message if `emit_handler` was only had warning message. So we print warning message in rust.

In the future, It might be more appropriate to let rust print the information and let the js side end the process

**BREAKING CHANGE:**

https://github.com/swc-project/swc/issues/9887


crates/swc_error_reporters/src/handler.rs
```diff
@@ -114,6 +114,15 @@ where
             Ok(_) => Err(anyhow::anyhow!(msg)),
             Err(err) => Err(err.context(msg)),
         }
+    } else if handler.has_warning() {
+        let mut lock = wr.0.lock();
+        let error = take(&mut *lock);
+
+        let msg = String::from_utf8(error).expect("error string should be utf8");
+
+        println!("{}", msg);
```
kdy1, line 123: We should not print it. Assuming that the user is going to print is too much of an assumption.
I think one solution is to return the warning string in the `TransformOutput` instead, although it may require a breaking change.

GiveMe-A-Name: Yes, your worry is considerate.
Emit warning string in the `TransformOutput` instead is the best way to resolve it. But we need more discuss it that avoid swc import uncomfortable breaking change. 
I will try another way to fix it.

kdy1: How about defining a trait in `swc_error_reporters` and implementing it for `TransformOutput`?

GiveMe-A-Name: We need create a new fields in `TransformOutput` to save diagnostics, then node cli print it using js api.

kdy1: Yeah that's what I intended. You can't avoid a breaking change in this case, but you can define a trait in `swc_error_reporters` to mitigate the dependency issue

crates/swc_compiler_base/src/lib.rs
```diff
@@ -37,17 +37,28 @@ pub struct TransformOutput {
...
 pub struct TransformOutput {
     pub code: String,
+
     #[serde(skip_serializing_if = "Option::is_none")]
     pub map: Option<String>,
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub output: Option<String>,
+
+    pub diagnostics: std::vec::Vec<String>,
```
kdy1, line 55: This is a breaking change anyway

crates/swc_common/src/errors/emitter.rs
```diff
@@ -41,6 +43,8 @@ pub trait Emitter: crate::sync::Send {
     fn should_show_explain(&self) -> bool {
         true
     }
+
+    fn emit_diagnostics(&mut self, _de: &RefCell<dyn DiagnosticEmitter>) {}
```
kdy1, line 47: `RefCell` this trait is too much assumption about the trait and `DiagnosticEmitter`

crates/swc_compiler_base/src/lib.rs
```diff
@@ -37,17 +37,28 @@ pub struct TransformOutput {
...
+}
+
+impl DiagnosticEmitter for TransformOutput {
+    fn emit(&mut self, diagnostic: String) {
+        self.diagnostics.push(diagnostic);
```
kdy1, line 60: I think this is inefficient without a good reason.

kdy1 (changes requested):
This PR is inevitably breaking change, so I think it's overly complicated.

crates/swc_compiler_base/src/lib.rs
GiveMe-A-Name: Maybe  transformOutput take diagnostics handler directly is a better way.
```rust 
transformOutput.diagnostics = handler.take_diagnostics();
```

crates/swc_compiler_base/src/lib.rs
GiveMe-A-Name: `TransformOutput` may need include [diagnostics](https://github.com/swc-project/swc/blob/main/crates/swc_common/src/errors/diagnostic.rs#L29-L48) struct.
What do you think?

kdy1: This should be json string I think, so it's fine. But we should not render the diagnostic in the current way.

GiveMe-A-Name: > But we should not render the diagnostic in the current way.

You means [here](https://github.com/swc-project/swc/blob/5efe61828a5372b066dc473de078be115157bed8/crates/swc_error_reporters/src/lib.rs#L181-L189)?

kdy1: Can we have an option to set the formatting of error messages? Like `--message-format=json` of various tools.

This may be out of the scope of the PR, but I think we cannot render it (`miette`/`PrettyEmitter`) or serialize it (json) without such option

kdy1: Hopefully in `HandlerOpts`

GiveMe-A-Name: Got it, i will implement it recently.

crates/swc_common/src/errors/mod.rs
```diff
@@ -874,6 +874,10 @@ impl Handler {
             }
         }
     }
+
+    pub fn take_diagnostics(&self) -> Vec<String> {
```
kdy1, line 878: This sounds good. Nice work!

crates/swc_error_reporters/src/lib.rs
```diff
@@ -175,9 +178,19 @@ impl Emitter for PrettyEmitter {
             children,
         };
 
+        let mut format_result = String::new();
+
         self.reporter
-            .render_report(&mut self.wr, &diagnostic)
+            .render_report(&mut format_result, &diagnostic)
```
kdy1, line 184: I think this may be fine without an additional option, considering we were already formating as string (rendering using `miette`) error messages, and `json` support is out of the scope of this PR anyway.

kdy1 (approved):
Thank you so much for the effort! I think this PR is fine to be merged, but please tell me if you want to add json support to this PR.

crates/swc_common/src/errors/mod.rs
GiveMe-A-Name: Great, you can merge it at any time.
> Can we have an option to set the formatting of error messages?

About `option to set the formatting of error messages`, i can implement in another PR.

PR #10036, opened 2025-02-14
perf(hstr): Disable interning for global store

kdy1:
**Description:**

This PR disables interning for **global** atom store, which is invoked by using `From<_>` implementation of `Atom`, like  `"foo".into()`. I think it's possible that interning them does not worth the cost, so I want to profile this.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/10032


kdy1:
Closing as https://github.com/swc-project/swc/issues/10032 is fixed

PR #10038, opened 2025-02-14
fix(typescript): Improve type inferring for undefined and null

CPunisher:
I remember doing that by design, but I forget why. Now let's align it with tsc


kdy1 (approved):
Thank you!

PR #10046, opened 2025-02-18
build(es): Select optimization level for each crates

kdy1:
**Description:**

I used `ddt cargo bin-size select-per-crate --compare --features plugin --release`, the command added by https://github.com/swc-project/ddt/pull/46


kdy1:
With this PR:

```
❯ ll swc.darwin-arm64.node
-rwxr-xr-x  1 kdy1  staff    38M  2 18 17:17 swc.darwin-arm64.node

```

Main:

```
❯ ll swc.darwin-arm64.node
-rwxr-xr-x  1 kdy1  staff    41M  2 18 17:19 swc.darwin-arm64.node
```

PR #10052, opened 2025-02-19
perf(es/minifier): Use core visitors

kdy1:
**Description:**

I'll try to profile the performance difference and the binary size difference.


kdy1:
Closing as this didn't reduce the size greatly

PR #10077, opened 2025-02-21
feat(es/minifier): Seq inline into var without init

Austaras:
**Description:**

**BREAKING CHANGE:**


kdy1 (approved):
Cool, thanks!

PR #10098, opened 2025-02-25
refactor(es/react): Remove `Lrc` from `parse_expr_for_jsx`

dsherret:
**Description:**

It being in an Lrc is a higher level concern.

**BREAKING CHANGE:**

`parse_expr_for_jsx` no longer returns an `Lrc`.

### DIFF
--- a/crates/swc_ecma_transforms_react/src/jsx/mod.rs
+++ b/crates/swc_ecma_transforms_react/src/jsx/mod.rs
@@ -136,7 +136,7 @@ pub fn parse_expr_for_jsx(
     name: &str,
     src: Lrc<String>,
     top_level_mark: Mark,
-) -> Lrc<Box<Expr>> {
+) -> Box<Expr> {
     let fm = cm.new_source_file_from(
         FileName::Internal(format!("jsx-config-{}.js", name)).into(),
         src,
@@ -163,7 +163,6 @@ pub fn parse_expr_for_jsx(
         apply_mark(&mut expr, top_level_mark);
         expr
     })
-    .map(Lrc::new)
     .unwrap_or_else(|()| {
         panic!(
             "failed to parse jsx option {}: '{}' is not an expression",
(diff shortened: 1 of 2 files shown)


crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -360,13 +359,12 @@ impl JsxDirectives {
                             if let Some(src) = val {
                                 if is_valid_for_pragma(src) {
                                     // TODO: Optimize
```
dsherret, line 361: Maybe this todo was talking about the cloning? Perhaps we can remove this now?

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -136,7 +136,7 @@ pub fn parse_expr_for_jsx(
     name: &str,
     src: Lrc<String>,
     top_level_mark: Mark,
-) -> Lrc<Box<Expr>> {
+) -> Box<Expr> {
```
dsherret, line 139: Sidenote: this actually probably should just return an Expr then someone could put it in an Lrc or Box after?

kdy1 (changes requested):
It was necessary for turbopack because of caching and allocations.
What's the exact problem?

dsherret:
> It was necessary for turbopack because of caching and allocations.
What's the exact problem?

In this case, the `parse_expr_for_jsx` function doesn't need the return type to be in an Arc/Rc. The caller of the function should decide on that which makes this function more flexible. For example, this removes a clone here: https://github.com/swc-project/swc/pull/10098/files#diff-8abacd70370fd7f85bb3452458bd542918f7b0a9dc6074dce36d726593c4efe0R362

crates/swc_ecma_transforms_react/src/jsx/mod.rs
dsherret: It seems that the codebase too often pre-emptively boxes stuff. For example, `parse_file_as_expr` should probably return an `Expr` instead of `Box<Expr>` and then this could return an `Expr` and then the caller could decide to put this in an Lrc or Box. This would be a big change though.

kdy1: Returning `Box<Expr>` is better if you can't be 100% sure that `rustc` will see `box some_expression()` for all instance of `some_expression()`

kdy1:
It seems like I got confused with another `Lrc`. I'll review this later today.

crates/swc_ecma_transforms_react/src/jsx/mod.rs
kdy1: Did you profile the performance using a profiler? You will notice that it becomes slower without boxing, although the situation is a bit different from the time I made decisions.

crates/swc_ecma_transforms_react/src/jsx/mod.rs
kdy1: I think so

kdy1:
Hmm
Unboxing Expr may worth a try, mainly because now it’s smaller than 64 bytes.

kdy1:
Nevermind. It's still 96 bytes so unboxing is just bad.

Issue #10099, opened 2025-02-25
resolver no longer analyzes `declare var`, but it does analyze `declare function`


dsherret:
### Describe the bug

Since https://github.com/swc-project/swc/pull/9734 (swc 1.9.3) it's no longer possible to use the resolver to figure out the scope of a `declare var`.

Even if this was intentional, it seems inconsistent with `declare function` which still has the correct scope set?

### Input code

```typescript
// file.d.ts
export class Test { // ctxt: 2
  method(): typeof publicVar;
}

declare function random(): number; // ctxt: 2
export declare var publicVar: number; // ctxt: 0 (previously 2 as well)
```

### Config

```json

```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.9.3&code=H4sIAAAAAAAAA0XMQQqDMBBG4X0gd%2FiX7RXqNcR9TEYqmEyYTIoi3l2lqPvvPZozi8JPrhS0VBSrNUAk%2FXJ4vT%2FQJRMPyLWfRt85aazZrLEm0NEIYajJ68gJ4lLgeCapxp5OSP%2F5RX9Ons%2FNdnIgB2WCAAAA&config=H4sIAAAAAAAAA1WQvW7DMAyE9z6FoTmoCw8ZsrbIlilPQCh0w0J%2FICkgRuB3jyTHcbqJ3%2FF04N0%2Fus78iTWH7l6eZUjAgvyaC5EpKNwKMTolFMuU1OxWVaVKyhkbmRfBKPAvajWhDF%2FDYFbOEGSM7N8TsuAPjhTwGPnbgciR0F3k%2Be8rihGsvvkKIp8i6zlmtljDrqpJDn2P4j%2Fl2qfFstsMnIOSb7uQNXpQsuYpz%2F8ucDFK3RvBSb2tcePjJTvc6qqVLFfuzbZEgcZp9TZEclqdrav5AeQIxLB6AQAA

### Expected behavior

Previously, the above said the scope of `publicVar` was the same scope as `random`.

### Actual behavior

Now it says the ctxt of `random` is `2` and `publicVar` is `0`.

### Version

1.11.1

CPunisher:
Here is another same problem: https://github.com/swc-project/swc/issues/9892

PR #10101, opened 2025-02-25
perf: Remove some vector allocations in resolver

dsherret:
**Description:**

I was looking into https://github.com/swc-project/swc/issues/10099 and noticed this.

**BREAKING CHANGE:** No


kdy1 (approved):
Thank you!

Issue #10097, opened 2025-02-25
`const enum` transpiles to empty statement


dsherret:
### Describe the bug

Recent versions of swc have started transpiling const enums as an empty statement. I'm not sure if this is intentional, but the TypeScript compiler does not emit an empty statement in this case.

### Input code

```typescript
const enum E {
  A
}
```

### Config

```json

```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.10.18&code=H4sIAAAAAAAAA0vOzysuUUjNK81VcFWo5uVSUHDUAZFOQLKWlwsAQ0wgaB8AAAA%3D&config=H4sIAAAAAAAAA1WQvW7DMAyE9zyFwTmoAw8ZsjbIlilPQCh0o1Z%2FICm0RuB3j2THcbqJ3%2FF0ON43TQPfYuDQ3MuzDAlZiF9zITIExb9CQIdEYtgmhe2iqlRJOdNExlkARf4irSaSbtd1sHDGIH1k%2F56QhY7U20CnyJ8ORU6W3FWe%2F76imNDom68g61NkvcTMhmrYTTXJoW1J%2FIfc2jRbtquBc1Drp13MGj2qNfCUx38NXIxS93p0UrtNHHy8ZkfruepJ5pZ7WJdssP2weCdk5bw4IYefEH8DbMYHRewXq38BAAA%3D

### Expected behavior

Should be removed and output nothing.

```js
```

### Actual behavior

```js
;
```

### Version

1.11.1

PR #10128, opened 2025-03-01
fix(es/typescript): Remove empty statements that const enum decls are folded into

CPunisher:
closes https://github.com/swc-project/swc/issues/10097

By searching `asEmbeddedStatement` in https://github.com/microsoft/TypeScript/blob/main/src/compiler/factory/nodeFactory.ts, we can know there are four cases where typescript **doesn't** insert an empty statement:
1. Stmt in ModuleItem
2. Stmts in Script's body
3. Stmts in BlockStmt
4. Stmts in SwitchCase

Case 2, 3, 4 can be merged in `visit_mut_stmts`, while case 1 should be handled in `visit_mut_module_items`

### DIFF
--- a/.changeset/gold-hotels-camp.md
+++ b/.changeset/gold-hotels-camp.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_transforms_typescript: patch
+---
+
+fix(es/typescript): Remove empty statements that const enum decls are folded into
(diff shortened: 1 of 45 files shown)


magic-akari (dismissed):

kdy1 (approved):

kdy1:
I discarded the changes of stack traces because CI failed

Issue #10133, opened 2025-03-03
v1.11 breaks ordering with inlined functions


jdelStrother:
### Describe the bug

I'm using chart.js, which has a [function](https://github.com/chartjs/Chart.js/blob/a647e0d007ff43e9ddc397603cc9ba1547ca0791/src/helpers/helpers.curve.ts#L186) that, simplified, looks like

```js
function _updateBezierControlPoints(points, options, area, loop, indexAxis) {
    let i, ilen, point, controlPoints;
    for(i = 0, ilen = points.length; i < ilen; ++i){
         point = points[i];
         controlPoints = splineCurve(point)
         point.cp1x = controlPoints.previous.x;
         point.cp1y = controlPoints.previous.y;
    }
}
```

as of v1.11+ with compression enabled, it compiles to something along the lines of

```js
function _updateBezierControlPoints(points, options, area, loop, indexAxis) {
    let i, ilen, point, controlPoints;
    for(i = 0, ilen = points.length; i < ilen; ++i){
         point.cp1x = (controlPoints = function(point) {
             // inlined splineCurve
         })(point = points[i]).previous.x;
         point.cp1y = controlPoints.previous.y;
    }
}
```

which explodes on the first iteration because `point` hasn't been assigned

### Input code

```typescript
function splineCurve(firstPoint, middlePoint, afterPoint) {
    const previous = firstPoint.skip ? middlePoint : firstPoint;
    const current = middlePoint;
    const next = afterPoint;
    return {
            x: current.x - (next.x - previous.x),
            y: current.y - (next.y - previous.y)
    };
}

function _updateBezierControlPoints(points) {
    let i, ilen, point, controlPoints;
        let prev = points[0];
        for(i = 0, ilen = points.length; i < ilen; ++i){
            point = points[i];
            controlPoints = splineCurve(prev, point, points[Math.min(i + 1, ilen - 1)]);
            point.cp1x = controlPoints.x;
            prev = point;
        }
}
let points = [{x: 1, y: 2}, {x: 2, y: 1}, {x: 3, y: 2}];
_updateBezierControlPoints(points);
console.log(points)

export { _updateBezierControlPoints };

```

### Config

```json

```

### Playground link (or link to the minimal reproduction)

[...]

PR #10134, opened 2025-03-04
fix(es/minifier): Fix regression due to #10056

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/10133


kdy1:
```
❯ git bisect bad
19d01d718aeb7a211384944dd2d9858cc4f04da2 is the first bad commit
commit 19d01d718aeb7a211384944dd2d9858cc4f04da2
Author: Austaras <austaras@outlook.com>
Date:   Thu Feb 20 23:44:38 2025 +0800

    feat(es/minifier): Inline across side-effect-free member exprs, (#10056)

    **Description:**

    This is safe even in the presence of an impure setter because JS eval rhs **before** setter

    ```js
    ({ set a(v) { console.log(v) }}).a = (console.log(1), 2)
    ```

    would output `1 2`, and gcc has this optimization too.
``

PR #10145, opened 2025-03-04
refactor(es/parser): Rewrite lexer and parser

kdy1:
**Description:**

This is a rewrite from scratch

### DIFF
--- a/crates/swc_ecma_fast_parser/src/lexer/cursor.rs
+++ b/crates/swc_ecma_fast_parser/src/lexer/cursor.rs
@@ -1,0 +1,192 @@
+//! Byte-level cursor for fast input traversal
+//!
+//! This cursor operates directly on UTF-8 bytes for maximum performance.
+
+use std::slice;
+
+use swc_common::BytePos;
+
+/// High-performance cursor for traversing input bytes
+pub struct Cursor<'a> {
+    /// Input source as bytes
+    input: &'a [u8],
+
+    /// Current position in bytes
+    pos: usize,
+
+    /// Length of the input in bytes
+    len: usize,
+}
+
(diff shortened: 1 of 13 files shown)


kdy1:
The benchmark result looks promising.

Lexer only benchmark:

Although it fails in some cases, it's much faster than the original implementation

 - Original:
스크린샷 2025-03-05 오후 9 56 53

 - New:
스크린샷 2025-03-05 오후 9 55 47

kdy1:
With another AI-based optimization:
image

kdy1:
I'll create follow-up PRs for other tasks

crates/swc_ecma_fast_parser/src/lexer/cursor.rs
```diff
@@ -0,0 +1,341 @@
...
+use crate::util::{likely, unlikely};
+
+/// High-performance cursor for traversing input bytes
+#[repr(C)] // Ensure predictable memory layout for better cache behavior
+pub struct Cursor<'a> {
+    /// Input source as bytes
+    input: &'a [u8],
+
+    /// Current position in bytes
+    pos: usize,
```
magic-akari, line 17: Would `u32` be better here? It natively handles JavaScript inputs up to 4GB and avoids unnecessary casts, making the data flow more explicit. Thoughts?

kdy1: Yeap. I missed it. Thank you!
(I forgot it while doing `git reset --hard` due to regression)

kdy1: The point is that, `pos` and `len` can be fetched at once on 64 bit machines

PR #10102, opened 2025-02-25
fix(es/resolver): Analyze variable declarations with `declare`

dsherret:
**Description:**

Tests are going to fail here because I haven't updated them. Let me know if this is the right approach and I'll look into it.

This change makes the resolver analyze ambient declarations when `handle_types` is `true` and also makes the behaviour consistent across all declaration kinds and not just variable declaration.

**Related issue:**

Closes https://github.com/swc-project/swc/issues/9892

PR #10147, opened 2025-03-05
fix(es/resolver): Invert #10102

kdy1:
**Description:**

 - Fixes regression of correctness caused by https://github.com/swc-project/swc/pull/10102


crates/swc_ecma_transforms_base/tests/ts-resolver/declare/output.ts
```diff
@@ -1,14 +1,14 @@
-declare const MY_MAGIC_VARIABLE__2: string;
-declare function myFunction__2(): string;
+declare const MY_MAGIC_VARIABLE__0: string;
```
dsherret, line 1: These aren't global identifiers.

kdy1: I added `resolver_resolving_declare`

kdy1:
I'll see if https://github.com/vercel/next.js/pull/75346 by @CPunisher works

crates/swc_ecma_transforms_base/src/resolver/mod.rs
```diff
@@ -153,6 +153,39 @@ pub fn resolver(
             handle_types: typescript,
             unresolved_mark,
             top_level_mark,
+            resolve_declare: false,
+        },
+    })
+}
+
+/// See [resolver]
+pub fn resolver_resolving_declare(
```
dsherret, line 162: This works fine, thanks!

kdy1:
My logistic was that the variable renamer (`hygiene`) is the inherent pass of SWC, so it should not collide with the resolver pass, but let's see, as it sounds hacky to the core team member.

kdy1:
Sorry. I misunderstood the problem of turbopack. I noticed it while looking at the content of https://github.com/vercel/next.js/pull/75346

(The changes were quite different than I thought)

PR #10154, opened 2025-03-06
perf(es/fast-lexer): Use lookup table for lexing operators

kdy1:
**Description:**

**Related issue:**

 - This is follow-up of https://github.com/swc-project/swc/pull/10153


kdy1:
Closing as there's no perf diff

PR #10156, opened 2025-03-06
perf(es/minifier): Try new DCE

kdy1:
**Description:**

Let's run benchmark...


kdy1:
This was not a correct impl

PR #10159, opened 2025-03-07
fix(es/parser): Rescan `<<` in new expression

magic-akari:
**Related issue:**

- https://github.com/nodejs/amaro/issues/194


kdy1 (approved):
Thanks!

PR #10160, opened 2025-03-07
perf(es/fast-lexer): Cache `splat` of SIMD

kdy1:
**Description:**

I'm not sure if this would improve performance, so let's look at the benchmark result


kdy1:
Closing as it's not an optimization

Issue #10150, opened 2025-03-05
@swc/wasm-typescript errors are now strings


marco-ippolito:
### Describe the bug

I think this was a breaking change since the error is emitted as a string, which is unexpected and slightly unconvinient. The expectation was to keep emitting an js object error.
The main issue is that:
- json parsing adds some overhead and it has to be done ALWAYS since string is not useful
- makes error handling weird/inconsistent since swc wasm crashes are also strings so we try to json.parse and it fails while right now if its a string we know its a wasm crash

### Input code

```typescript
const inputCode = "module F { export type x = number }";
try {
	transformSync(inputCode, {
		mode: "transform",
	});
} catch (error) {
	console.log("TYPEOF OF ERROR", typeof error);
	assert.strictEqual(error.code, "UnsupportedSyntax");
}



TYPEOF OF ERROR string
```

### Config

```json

```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.11.7&code=H4sIAAAAAAAAA8vNTynNSVVwU6hWSK0oyC8qUSipLEhVqFCwVcgrzU1KLVKoBQBGDE3yIwAAAA%3D%3D&config=H4sIAAAAAAAAA1WPSw7DIAwF9zkF8rrbdtE79BAWdSIifrKJVBTl7iUE0maH3xsz8jooBbNoeKq1PMsQkYX4nEsi2Sf8lARIOxTNJia49XaWvRrRCtVoOxpIyBOluiX3hoMNQajjLXPGmzH%2FC3VwkUnkCu4o%2BsnSVTc0JbjwXmrZDkk50qF%2FwA%2FqsvNjMPLqm4kXGrYvhlQioBQBAAA%3D&strip-types=

### Expected behavior

Emit an object

### Actual behavior

Emits a string

### Version

1.11.7

kdy1:
I'll resolve this tomorrow

kdy1:
What should it throw? I'm asking because it may throw more than one errors. Is throwing an array of json object fine?

marco-ippolito:
> What should it throw? I'm asking because it may throw more than one error. Is throwing an array of json object fine?

Why more than one error? I would expect the first error encountered by the parser

kdy1:
Because there can be more than one modules, but I'll throw others out

marco-ippolito:
From my pov i would need the previous behavior where just one object was thrown, with the location

PR #10162, opened 2025-03-07
feat(ts/fast-strip): Throw an object instead of string

kdy1:
**Description:**

Throw an object from `@swc/wasm-typescript`.

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/10150


marco-ippolito:
@kdy1 I noticed that the object is a javascript Map object, is it possible to make into a js object rather than a map?

PR #10167, opened 2025-03-08
fix(es/compat): Hoist `arguments` in object method while lowering async functions

### LINKED ISSUE #10149: `arguments` is not transform correct in async function
### Describe the bug

After the code transform, it breaks the original behavior.

### Input code

```typescript
var foo = {
    async bar({ name }) {
        console.log('arguments.length', arguments.length)
    }
}

foo.bar({})
```

### Config

```json

```

https://play.swc.rs/?version=1.11.7&code=H4sIAAAAAAAAAytLLFJIy89XsFWo5lIAgsTiyrxkhaTEIo1qhbzE3FSFWk2oDAgk5%2BcV5%2Bek6uXkp2uoJxall%2Bam5pUU6%2BWk5qWXZKjrKKALaYJ11nLVcnEBLdEDG1urCQBsC3CYdAAAAA%3D%3D&config=H4sIAAAAAAAAA1WPSw7DIAxE95wCed1tN71DD2FRJwLxEyZSUZS7Fyikzc4z8%2ByRdyElGFbwkHsdq4iYmNKpq8PFZ3xXB0g5ZJV0zHCbqeEWLWiZunV8E8iYVsp9i%2B8DBxsC08SH57TXS%2FkvVMHFRMxXsKHoV0vXOjEqwYXX1sPxSC6xqXbNBW8YfuRsPK%2BD5udcz2kjcXwAmJf30BkBAAA%3D

### SWC Info output

_No response_

### Expected behavior

Output: 1

The `arguments` should get a correct value

### Actual behavior

Output: 0

The `arguments` should be correct

### Version

1.11.7

### Additional context

Note: once remove `async`, it works fine like expect

Image

magic-akari:
**Related issue:**

- Closes #10149

### DIFF
--- a/.changeset/foo-bar.md
+++ b/.changeset/foo-bar.md
@@ -1,0 +1,6 @@
+---
+swc_ecma_compat_es2017: patch
+swc_core: patch
+---
+
+fix(es/compat): Hoist arguments in object method while lowering async functions
--- a/crates/swc/tests/fixture/issues-10xxx/10149/1/input/.swcrc
+++ b/crates/swc/tests/fixture/issues-10xxx/10149/1/input/.swcrc
@@ -1,0 +1,11 @@
+{
+    "jsc": {
+        "parser": {
+            "syntax": "ecmascript",
+            "jsx": false
+        },
+        "target": "es5"
+    },
+    "minify": false,
+    "isModule": true
+}
--- a/crates/swc/tests/fixture/issues-10xxx/10149/1/input/index.ts
+++ b/crates/swc/tests/fixture/issues-10xxx/10149/1/input/index.ts
@@ -1,0 +1,15 @@
+var foo = {
+    async bar({ name }) {
+        console.log("arguments.length", arguments.length);
+    },
+};
+
+class Foo {
+    async bar({ name }) {
+        console.log("arguments.length", arguments.length);
+    }
+}
+
+async function bar({ name }) {
+    console.log("arguments.length", arguments.length);
+}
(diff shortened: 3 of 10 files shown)


magic-akari:
During my investigation of this issue, I discovered that esbuild employs a more streamlined approach compared to Babel. Rather than hoisting `this` and `arguments`, esbuild directly passes the current `this` and `arguments` to the function  `__async`, then utilizes `apply` internally to properly bind them.

Here's a code example demonstrating this approach:
```JavaScript
var foo = {
  bar(_0) {
    return __async(this, arguments, function* ({ name }) {
      console.log("arguments.length", arguments.length);
    });
  }
};
```

This optimization eliminates the hoisting process for `this` and `arguments`, though `super` references still require hoisting. I'm currently evaluating the possibility of adopting esbuild's transformation strategy for our implementation.

kdy1 (dismissed):
Changeset bot is refusing to work 🤣 
I'll add them manually

Austaras:
Why isn't this implemented using `FnEnvHoister`?

kdy1 (approved):

kdy1:
Oh I forgot the comment. @magic-akari Is there a particular reason?

magic-akari:
> Oh I forgot the comment. @magic-akari Is there a particular reason?

Are you referring to esbuild's approach? 
From my observations, esbuild's output tends to be more consistent and simpler than Babel's across various scenarios.
I generally favor simpler implementations as they tend to be easier to maintain and less prone to bugs.

magic-akari:
> Why isn't this implemented using `FnEnvHoister`?

Do you mean `wasn't`? It is implemented using `FnEnvHoister` now.
I have no idea why it wasn't implemented that way originally. It might have simply been overlooked.

kdy1:
Sorry, you are right. Now it uses the `FnEnvHoister` and I think the previous implementation is from old days.

Austaras:
> Do you mean `wasn't`? It is implemented using `FnEnvHoister` now. I have no idea why it wasn't implemented that way originally. It might have simply been overlooked.

That's good.

> From my observations, esbuild's output tends to be more consistent

I don't like apply and bind. IIRC they may introduce some nasty problems.

PR #10181, opened 2025-03-12
perf(es/fast-lexer): replace PHF with static keyword lookup table

GiveMe-A-Name:
**Description:**

Replaces the perfect hash function (PHF) keyword lookup with a more performant static keyword lookup table. The new implementation:

- Removes dependency on `phf` crate
- Adds a custom static keyword lookup mechanism
- Introduces a more cache-friendly keyword search algorithm
- Adds support for the `module` keyword

**BREAKING CHANGE:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5313,7 +5313,6 @@ dependencies = [
  "criterion",
  "memchr",
  "num-bigint",
- "phf",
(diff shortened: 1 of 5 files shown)


kdy1 (approved):
Wow, thank you!

PR #10196, opened 2025-03-14
perf(es/minifier): Do not repeat applying pure minifier on last

kdy1:
**Description:**

This is mostly waste of time


crates/swc/tests/vercel/full/ms/1/output/index.ts
```diff
@@ -51,10 +51,12 @@ export default function(c, a) {
             }
         }(c);
         if ("number" == typeof c && isFinite(c)) {
+            ;
```
kdy1, line 54: Can someone help me fix the regression caused by this? We have https://github.com/swc-project/swc/blob/fcc68842422c0a46eee4d47956e77a727f202998/crates/swc_ecma_minifier/src/compress/pure/mod.rs#L157 but it seems like this semicolon is created after invoking `retain`

cc @swc-project/core-es

magic-akari: Understanding the traversal order remains challenging. A visualization tool that highlights AST node spans in the code editor during traversal, while logging corresponding visitor function locations, would be valuable. Implementing a step-through visualization that replays the traversal sequence within the editor interface could greatly enhance our debugging workflow.

magic-akari: https://github.com/swc-project/swc/blob/08cd363ac81e43ff09daac40d48abcc1fdab5146/crates/swc_ecma_minifier/src/compress/pure/vars.rs#L250-L255

The key point is that the VarMover here will delve deeper into the AST to modify code at nested inner layers, while our upper-layer visitor currently operates at the outer level.

```JavaScript
export default function(a, c) { // <- Pure::Visit_mut_stmts 
    try {
        if ("string" == typeof a && a.length > 0) return function(e) {
// ...
        }(a);
        if ("number" == typeof a && isFinite(a)) {
            var r, n, t, o;   // <- VarMover changed this
            return (null == c ? void 0 : c.long) ? (r = a, (n = Math.abs(r)) >= 86400000 ? s(r, n, 86400000, "day") : n >= 3600000 ? s(r, n, 3600000, "hour") : n >= 60000 ? s(r, n, 60000, "minute") : n >= 1000 ? s(r, n, 1000, "second") : "".concat(r, " ms")) : (t = a, (o = Math.abs(t)) >= 86400000 ? "".concat(Math.round(t / 86400000), "d") : o >= 3600000 ? "".concat(Math.round(t / 3600000), "h") : o >= 60000 ? "".concat(Math.round(t / 60000), "m") : o >= 1000 ? "".concat(Math.round(t / 1000), "s") : "".concat(t, "ms"));
        }
        throw Error("Value is not a string or number.");
    } catch (s) {
        var u; // <- VarMover changed this
        throw Error((void 0 === (u = s) ? "undefined" : e(u)) === "object" && null !== u && "message" in u ? "".concat(s.message, ". value=").concat(JSON.stringify(a)) : "An unknown error has occurred.");
    }
}
```

magic-akari (approved):

crates/swc/tests/vercel/full/ms/1/output/index.ts
kdy1: > Understanding the traversal order remains challenging. A visualization tool that highlights AST node spans in the code editor during traversal, while logging corresponding visitor function locations, would be valuable. Implementing a step-through visualization that replays the traversal sequence within the editor interface could greatly enhance our debugging workflow.

Yeah, I love the idea, but I need to think about the way to implement it.

---

> The key point is that the VarMover here will delve deeper into the AST to modify code at nested inner layers, while our upper-layer visitor currently operates at the outer level.

Yeap, thank you! I fixed it.

PR #10203, opened 2025-03-15
perf(es/minifier): Make pure optimizer more parallel

kdy1:
**Description:**

Note: Benchmark results may not represent the actual value for build machines with more CPU cores.


kdy1:
Closing as this resulted in regression of the performance in real-world apps with a powerful build machine

PR #10208, opened 2025-03-17
test(es/minifier): Test only sizes for large inputs

kdy1:
**Description:**

Large output results are not helpful for catching errors


Austaras:
Wait, I believe it's really helpful to have these full sized js in test. They could reveal many corner cases in complex scenario. @kdy1 I believe we shall revert this.

Austaras:
We should removed minified tests instead like mapbox and some other.

kdy1:
@Austaras I'll add back them via another PR, thanks for the notice!

PR #10218, opened 2025-03-17
feat(es/minifier): Default to the smallest size

kdy1:
**Description:**

AFAIK all bundlers used in production support extracting license comments. We can default to the smallest size, I think.


kdy1:
Side note: If this looks like a problem, feel free to tell me. I can revert/invert this PR.

Issue #10229, opened 2025-03-19
Reduce escaped newline for `TemplateElement`


wjw99830:
### Describe the feature

Escaped newline will be replaced to a newline character in esbuild.

Example:
```js
// input.js
const str = `a\nb\nc`;

// output.js
const str = `a
b
c`;
```

### Babel plugin or link to the feature description

https://esbuild.github.io/faq/#minified-newlines

PR #10232, opened 2025-03-19
feat(es/minifier): support `reduce_escaped_newline`

wjw99830:
**Description:** Add `compress.experimental.reduce_escaped_newline` option to convert `StringLiteral` with escaped newline `\n` to `TemplateLiteral` with newline character(only for `compress.ecma` >= 2015).

Default is `true`.

**BREAKING CHANGE:** None.

**Related issue (if exists):** #10229


kdy1:
I think we should not add a new option for this

wjw99830:
> I think we should not add a new option for this

You mean enable the feature by default and cannot be turned off ?

kdy1:
If so, can we have an experimental config instead? Like `compress.experimental.reduce_escaped_newline`? You may need to define `CompressExperimentalOptions` struct. (And please make it `#[non_exhaustive]`)

wjw99830:
Good idea.I'll try it later.

wjw99830:
I find that `swc_ecma_codegen` will generate `\n` for `\\n` within `TemplateElement.raw`.
```rs
let unescape = match c {
    '\\' => match iter.next() {
        Some(c) => match c {
            'n' => Some('\n'),
            // ...
        },
        // ...
    },
    _ => Some(c),
};

match unescape {
    Some('\n') => buf.push('\n'),
    // ...
}
```
It’s same as what I do inside `reduce_escaped_newline_for_tpl_element`, so I will remove it to prevent redundant work.

It still makes sense to retain `reduce_escaped_newline_for_str_lit`, means that `reduce_escaped_newline` works only when `compress.ecma` >= 2015. Do you have any better ideas about the design of this option?

kdy1:
Maybe add an option to [the codegen option](https://rustdoc.swc.rs/swc_ecma_codegen/struct.Config.html)?
It's already `#[non_exhaustive]`, so you would need to add one method and call it from https://github.com/swc-project/swc/blob/b192dc82e6a84bd30f159fb12ca8a216f41e8efb/crates/swc/src/lib.rs#L910-L917

wjw99830:
OK. Maybe we should call it `reduce_escaped_controls` to including `\t`(or possible others in future).

crates/swc_ecma_codegen/src/config.rs
```diff
@@ -39,6 +44,9 @@ pub struct Config {
 
     #[cfg_attr(feature = "serde-impl", serde(default))]
     pub inline_script: bool,
+
+    #[cfg_attr(feature = "serde-impl", serde(default = "true_by_default"))]
+    pub reduce_escaped_newline: bool,
```
kdy1, line 49: Can you document that this is `true` by default?

crates/swc_ecma_codegen/src/config.rs
```diff
@@ -87,4 +96,9 @@ impl Config {
         self.inline_script = inline_script;
         self
     }
+
+    pub fn with_reduce_escaped_newline(mut self, reduce_escaped_newline: bool) -> Self {
```
kdy1, line 100: And maybe also here

crates/swc_ecma_minifier/src/compress/optimize/strings.rs
```diff
@@ -137,4 +137,29 @@ impl Optimizer<'_> {
             _ => {}
         }
     }
+
+    pub(super) fn reduce_escaped_newline_for_str_lit(&mut self, expr: &mut Expr) {
```
kdy1, line 141: Do we still need this, even with the codegen config?

kdy1 (changes requested):

crates/swc_ecma_codegen/src/config.rs
wjw99830: Got

crates/swc_ecma_minifier/src/compress/optimize/strings.rs
wjw99830: This is used to transform string literal to template literal. Maybe should rename the method name, it's not clear enough.

wjw99830: Document added.

crates/swc_ecma_minifier/src/compress/optimize/strings.rs
```diff
@@ -137,4 +137,31 @@ impl Optimizer<'_> {
...
+
+        if let Expr::Lit(Lit::Str(s)) = expr {
+            if s.value.contains('\n') {
+                *expr = Expr::Tpl(Tpl {
+                    span: s.span,
+                    exprs: Default::default(),
+                    quasis: vec![TplElement {
+                        span: s.span,
+                        cooked: Some(s.value.clone()),
+                        raw: s.value.clone(),
```
kdy1, line 158: This may cause an issue.

kdy1: The function for `Tpl#raw` => `Str#value` is at https://rustdoc.swc.rs/swc_core/ecma/utils/swc_ecma_ast/struct.Str.html#method.from_tpl_raw , but let me search for `Str` => `Tpl` function...

kdy1: You should use functions at https://github.com/swc-project/swc/blob/f33b0bc2d1474232f608847a6a26eaa068f2c106/crates/swc_ecma_minifier/src/compress/pure/strings.rs#L573-L593

crates/swc_ecma_minifier/src/option/terser.rs
```diff
@@ -400,6 +421,11 @@ impl TerserCompressorOptions {
                     })
                 })
                 .collect(),
+            experimental: self.experimental.map(From::from).unwrap_or(
+                CompressExperimentalOptions {
+                    reduce_escaped_newline: self.defaults,
```
kdy1, line 426: I think this will enable this optimization by default for most of the users, but I think it's not intended.

crates/swc_ecma_minifier/tests/fixture/next/target-es2015/static/chunks/main-04b5934c26266542/output.js
```diff
@@ -923,7 +923,10 @@
                         return (cn = navigator.connection) && (cn.saveData || /2g/.test(cn.effectiveType)) ? Promise.resolve() : getFilesForRoute(assetPrefix, route).then((output)=>Promise.all(canPrefetch ? output.scripts.map((script)=>{
                                 var href, as, link;
                                 return href = script.toString(), as = "script", new Promise((res, rej)=>{
-                                    const selector = '\n      link[rel="prefetch"][href^="'.concat(href, '"],\n      link[rel="preload"][href^="').concat(href, '"],\n      script[src^="').concat(href, '"]');
+                                    const selector = `
```
kdy1, line 926: And I think you may need to revert the change in this file. I think it's caused by `defaults: true`.

crates/swc_ecma_minifier/src/compress/optimize/strings.rs
wjw99830: This function replace `'\n'` to `"\\n"` for `Tpl#raw`, but I just want to preserve the `'\n'`. It seems little bit redundant if replace `"\\n"` back to `'\n'`.

crates/swc_ecma_minifier/src/option/terser.rs
wjw99830: It's intentional to enable this optimization by default. Because it's already the default value in codegen for the same option.

crates/swc_ecma_minifier/src/compress/optimize/strings.rs
kdy1: If so, can you add test cases for inputs that contain those strings, respectively?

One test per the search text in "", like one for "\\$" in `convert_str_value_to_tpl_cooked` (you need 3 test cases for the function) and   and "$" for `convert_str_value_to_tpl_raw`?

wjw99830: `convert_str_value_to_tpl_cooked` will replace `"\\\\"` to `"\\"`. This behavior results in `Tpl#cooked` missing a `'\\'` character, is it expected?

kdy1:
```js
"\\\\"
```

is _not_ identical to

```js
`\\`
```


Oh? I think it's a bug of the function. Could you fix it?

kdy1: Maybe I need to search for the usage of the function because it can be a function with the wrong name. (e.g., Minifier is invoking the function with `Str#raw`, just if it's the case)

wjw99830: In `pure/strings.rs` and `pure/misc.rs`, it appears to be invoked with `Str#value`...

crates/swc_ecma_minifier/src/option/terser.rs
wjw99830: If we disable it by default, how to control the existed minification in codegen?

crates/swc_ecma_minifier/src/compress/optimize/strings.rs
wjw99830: I use a new fn to convert `Str#value` to `Tpl#raw` without replacing for `\n` and test for some edge cases.

crates/swc_ecma_minifier/src/option/terser.rs
kdy1: I think it's fine to enable it by default. I just misunderstood your intention.

crates/swc_ecma_minifier/src/compress/optimize/strings.rs
wjw99830: Does the original function need to be fixed in this PR?

kdy1: It’s fine to skip for this PR

wjw99830: Any other suggestions for this review(or above)?

crates/swc_ecma_minifier/src/option/terser.rs
```diff
@@ -69,6 +71,22 @@ pub enum TerserTopRetainOption {
     Seq(Vec<Atom>),
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+#[non_exhaustive]
+pub struct TerserExperimentalOptions {
+    #[serde(default = "true_by_default")]
+    pub reduce_escaped_newline: bool,
```
kdy1, line 79:
```suggestion
    pub reduce_escaped_newline: Option<bool>,
```
Sorry, I missed this one in the previous review.

kdy1: `true_by_default` with `bool` means

 - `experimental: {}` => `true`, regardless of `defaults`
 - `experimental: null` => Uses `defaults`

kdy1: I think those two should behave identically because the user didn't specify the  `reduce_escaped_newline`

kdy1 (commented):
Other things look good. Only one concern about consistency. Thanks!

crates/swc_ecma_minifier/src/option/terser.rs
wjw99830: It's my oversight. Thanks to point out.

kdy1 (dismissed):
Thank you!

PR #10237, opened 2025-03-20
feat(swc_parallel): Introduce `rayon` mode

kdy1:
**Description:**

It enables parallelism based on `rayon` instead of `chili`


kdy1:
I'll revert other PRs to publish this as a separate `swc_core`.

Issue #10192, opened 2025-03-14
[Enhance]: Improving SWC's Error Recovery for more friendly experience


GiveMe-A-Name:
### Describe the feature

The `try_with_handler` should return a SwcError instead of anyhow!(msg), it's helper upper application such as `deno`, `rspack` can get more detail about error.
Swc have many kind error with `invoke wasm plugin error`, `wasm plugin emit custom error`, `js parser error` ...., So upper application need more detail to distinguish what the error kind.
## Example
```rust
// crate  swc_error_reporter
pub struct Error {
    inner: Box<InnerError>
}

struct InnerError {
   kind: ErrorKind,
   message: Cow<str>,
   ....
}

impl Display for Error { .. }

fn try_with_handler(...) -> Result<T, Error> { ... }

```

```rust
// rspack/core

match swc_try_with_handler {
      Ok(program) => program,
      Err(error) => {
          match error.kind() {
             // do somethings for different errors
          }
      },
}
```

```rust
// node_binding

#[napi] 
fn transform_sync(...) -> Result<T> {
   swc_core::transform(...).map_err(|e: SwcError| anyhow!(e.to_string()))
}

```

kdy1:
I like the idea. I used anyhow::Error because `swc` was designed as a bridge for nodejs and rust, but it's used from the Rust side now.

GiveMe-A-Name:
> I like the idea. I used anyhow::Error because `swc` was designed as a bridge for nodejs and rust, but it's used from the Rust side now.

Yes. if we need pass error to nodejs, we also can use anyhow::Error, we can can `anyhow!(SwcError.to_string())` in node_binding.

PR #10241, opened 2025-03-21
enhance(errors): Integrate miette for enhanced diagnostic reporting

GiveMe-A-Name:
# Description:
### Enhance errors reporting

SWC api `try_with_handler` only return anyhow!(msg) Err, it'a will lose many message about error stack, level ..., 
To express more error message in swc, we can return `Vec<Diagnostic>` in `try_with_handler`;

try_with_handler_diagnostic return a `Result<Ret, Vec<Diagnostic>>`. Layer user can using `diagnostic.code` to distinguish what kind error is.

- Added `miette` as a dependency for improved error handling and reporting.
- Updated diagnostic handling to utilize `miette` for pretty printing errors.
- Introduced a new `ThreadSafetyDiagnostics` to manage and store diagnostics.
- Refactored emitters to support the new diagnostic structure and ensure compatibility with `miette`.
- Enhanced error handling in various components to provide clearer and more informative error messages.

# BREAKING CHANGE:

The api `try_with_handler` would not return anyhow!(msg).
It's will return WrapperDiagnostics as it's Err. Then you can take diagnostics from WrapperDiagnostics or get pretty string from it.

```diff
- pub fn try_with_handler<F, Ret>(cm: Lrc<SourceMap>, config: HandlerOpts, op:F) -> Result<Ret, anyhow::error>;
+ pub fn try_with_handler<F, Ret>(cm: Lrc<SourceMap>, config: HandlerOpts, op:F) -> Result<Ret, WrapperDiagnostics>;
```

```rust
/// A wrapper around a value that also contains a list of diagnostics.
/// It helps swc error system migrate to the new error reporting system.
pub struct WrapperDiagnostics {
    diagnostics: Vec<SwcDignostic>,
   ....
}

impl WrapperDiagnostics {
    pub fn diagnostics(&self) -> &[Diagnostic] {
        &self.diagnostics
    }

    pub fn to_pretty_error(&self) -> anyhow::Error {
        let error_msg = self.to_pretty_string();

        anyhow::anyhow!(error_msg)
    }


   pub fn to_pretty_string(&self) -> String {
      .... 
    }
}
```

### formatting

In SWC, the `Syntax Error` message formatting dependents anyhow. 
```rust
// input: Foo {}
// https://play.swc.rs/?version=1.11.13&code=H4sIAAAAAAAAA3PLz1eorgUAR6TWygYAAAA%3D&config=H4sIAAAAAAAAA1WPSw7DIAwF9zkF8rrbdtE79BAWdSIifrKJVBTl7iUE0maH3xsz8jooBbNoeKq1PMsQkYX4nEsi2Sf8lARIOxTNJia49XaWvRrRCtVoOxpIyBOluiX3hoMNQajjLXPGmzH%2FC3VwkUnkCu4o%2BsnSVTc0JbjwXmrZDkk50qF%2FwA%2FqsvNjMPLqm4kXGrYvhlQioBQBAAA%3D

  x Expected ';', '}' or <eof>
   ,-[input.js:1:1]
 1 | Foo {}
   : ^|^ ^
   :  `-- This is the expression part of an expression statement
   `----


Caused by:
    0: failed to process js file
    1: Syntax Error
```

So if we return `Vec<Diagnostic>` instead of `anyhow:Error`, The output formatting would have a litter breaking change.
```diff
  x Expected ';', '}' or <eof>
   ,-[input.js:1:1]
 1 | Foo {}
   : ^|^ ^
   :  `-- This is the expression part of an expression statement
   `----

- Caused by:
-    0: failed to process js file
-    1: Syntax Error
+ x Syntax Error
```

# Related issue (if exists):

https://github.com/swc-project/swc/issues/10192


GiveMe-A-Name:
> Actually, a breaking change is fine until next version of `swc_core` is a breaking change anyway

Got，I will update code later.

crates/swc/tests/tsc-references/1.0lib-noErrors.1.normal.js
```diff
@@ -10,3 +10,4 @@
 //!  31 | /**
 //!  32 |   * Converts A string to an integer.
 //!     `----
+//!   x Syntax Error
```
kdy1, line 13: We should not have this, because we throw a `SyntaxError` from JS

crates/swc_error_reporters/src/handler.rs
```diff
@@ -87,70 +135,75 @@ pub fn try_with_handler<F, Ret>(
...
-    op: F,
-) -> Result<Ret, Error>
-where
-    F: FnOnce(&Handler) -> Result<Ret, Error>,
-{
-    try_with_handler_inner(cm, config, op, true)
-}
+// Try operation with a [Handler] and prints the errors as a [String] wrapped
+// by [Err].
+// pub fn try_with_json_handler<F, Ret>(
```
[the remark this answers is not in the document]
GiveMe-A-Name: Got it

crates/swc/tests/tsc-references/1.0lib-noErrors.1.normal.js
GiveMe-A-Name: Emm, i think that we should have this.
Before this, `try_with_handler` get a function: `op: FnOnce(&Handler) -> Result<Ret, anyhow::Error>`, then return a `Result<Ret, anyhow:Error>`.
it would like this code
```rust
pub fn try_with_handler(...) {

   let ret = op(&handler);
   match ret {
     ...
      Err(err) => {
          return err.context(diagnostic_msg);
      }
   }   
}
```
In `tsc_references` tests, it's would print it without anyhow:error inner only print err's context by using api [err.to_string()](https://github.com/swc-project/swc/blob/main/crates/swc/tests/tsc.rs#L460-L462), So it would not print `Syntax Error`. Actually, it's should be printed.

crates/swc_error_reporters/src/handler.rs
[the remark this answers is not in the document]
GiveMe-A-Name: Done

crates/swc/tests/tsc-references/1.0lib-noErrors.1.normal.js
kdy1: We had printed `SyntaxError` using context at the past, and I removed it because of an exact requirement

GiveMe-A-Name: I’m a bit confused now. So, in some situations we should print the SyntaxError message, while in other situations we should not print it? Otherwise, why not just drop it

kdy1: No, we should not print it. I mean I intentionally removed it.

GiveMe-A-Name: Ok, I got it.

GiveMe-A-Name: So we completely ignore the error return by op function? I fell a bit weird  :confused:.
```rust
    let handler = Handler::with_emitter(true, false, emitter);

    let ret = HANDLER.set(&handler, || op(&handler));

    match ret {
        Ok(ret) => {
            if diagnostics.contains_error() {
                return Err(diagnostics.as_wrapper(cm.clone(), config));
            }

            Ok(ret)
        }
        // ignore the error, only return diagnostics collected from handler.
        Err(_err) => Err(diagnostics.as_wrapper(cm.clone(), config)),
    }
```

kdy1: I don’t think we should, but we need a way to remove ‘SyntaxError’ from the error message of JS ‘SyntaxError’

GiveMe-A-Name: Ok, i got it.

kdy1: I think we may ignore empty errors, or create a special error type (unit struct like `struct OnlyDiagnostics;` in `swc_common` or `swc_error_reporters`.

GiveMe-A-Name: I don't get what you means. maybe you can give us a pseudo-code

kdy1: You can define an error struct named `OnlyDiagnostics` in the `swc_error_reporters`, and you can then [downcast](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.downcast) the error object to `OnlyDiagnostics`.

In `swc_error_reporters` crate

```rust
struct OnlyDiagnostics;

impl Error for OnlyDiagnostics {

}

// implement more traits, like Display


```

From `swc` crate (IIRC)

```rust

// Instead of `bail!("Syntax Error");`
anyhow::bail!(swc_error_reporters::OnlyDiagnostics);

```

From `try_xxx`

```rust
if let Some(..) = err.downcast_ref::<OnlyDiagnostics>() {
 // Drop contexts, and emit only diagnostics
} else {
 // Use err.to_string() to include contexts
}
```

GiveMe-A-Name: > You can define an error struct named OnlyDiagnostics in the swc_error_reporters, and you can then [downcast](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.downcast) the error object to OnlyDiagnostics.

I will try to implement it later

kdy1: Okay, I'll review it later today or tomorrow.

I'll change the target branch to `dev` instead of `main`, and I'll rebase-merge the `dev` branch into the main branch once all tasks are done. In this way all PRs will be a separate commit in the main branch git history.

For context of using another branch, merging the PR that makes JS throw `SyntaxError` with `Syntax Error` in error message will block updating `swc_core` (in next.js) or publishing a new version, so I would want to avoid it.

crates/swc_error_reporters/src/error.rs
```diff
@@ -0,0 +1,16 @@
+/// ParseSyntaxError is a custom error type that represents a syntax error in
+/// swc parse
+///
+/// We will return this error if swc parser occur syntax error, but we will
+/// ignore this error in rust side, because we will return this error in js
+/// side.
+#[derive(Debug)]
+pub struct ParseSyntaxError;
```
kdy1, line 8: This should have a more general name, I think?

kdy1: But not sure at the moment. I'm not sure if it's correct.

GiveMe-A-Name: how about `OnlyDiagnosticsError`? I'm not sure if it's suitable too.

kdy1: I think it would be better for now. (Mainly because I'm not sure about the naming at the moment)

GiveMe-A-Name: Got it. I update it right now.

crates/swc_common/src/errors/diagnostic.rs
```diff
@@ -496,6 +505,8 @@ impl Diagnostic {
     }
 }
 
+// impl From<>
```
kdy1, line 508:
```suggestion
```

crates/swc_common/Cargo.toml
```diff
@@ -39,11 +39,12 @@ shrink-to-fit = ["dep:shrink-to-fit", "swc_atoms/shrink-to-fit"]
 
 
 [dependencies]
-anyhow                = { workspace = true, optional = true }
+anyhow                = { workspace = true }
 arbitrary             = { workspace = true, features = ["derive"], optional = true }
 bytecheck             = { workspace = true, optional = true }
 cfg-if                = { workspace = true }
 either                = { workspace = true }
+miette                = { workspace = true, features = ["fancy-no-syscall"] }
```
kdy1, line 47: It should not depend on `miette` directly. It's too heavy to be a common dependency

kdy1 (changes requested):

crates/swc_common/Cargo.toml
GiveMe-A-Name: I can move it to `swc_error_reports` crate

kdy1: Thank you! Also, I think `fancy-no-syscall` should be optionally enabled (using a cargo feature) because binary size matters for some users of SWC

bindings/binding_core_wasm/__tests__/error.js
```diff
@@ -1,9 +1,6 @@
 const swc = require("../pkg");
 
-it("properly reports error", function () {
-    expect(() => {
-        swc.transformSync("Foo {}", {});
-    }).toThrow("Syntax Error");
+it("properly reports error", () => {
```
kdy1, line 3: Please revert this. The `SyntaxError` type includes the text `Syntax Error`

GiveMe-A-Name: Got it, i should add `Syntax Error` text in js side `@swc/core` package.

kdy1: To be exact, we should throw the JS SyntaxError type insted of adding text

GiveMe-A-Name: I found it's hard to do it, js side just receive a error from napi. But js do not known what kind error is.

kdy1: `napi` has an API to throw syntax errors.
https://docs.rs/napi/latest/napi/struct.Env.html#method.throw_syntax_error

GiveMe-A-Name: I have a idea that we maintain syntax error in `try_with_handler`, we only need find a ways to control print formatting way.
For example: continue using `Debug` trait and `Display` trait to control the print formatting. 
I found swc using `format!("{}", err)` or `format!("{:?}", err)` to control formatting before this PR.

kdy1: Do you mean we should include Syntax Error text from diagnostics from the swc crate?

GiveMe-A-Name: Yes.Stay the same as before. If we need report syntax Error in js not rust, maybe we need have other PR.
And another reason is Syntax Error emit by js side the mainly aim is to resolve the print formatting.

kdy1: SyntaxError should be thrown from Rust code using napi

GiveMe-A-Name: Got it, i will update it later

crates/swc_compiler_base/Cargo.toml
```diff
@@ -33,6 +33,7 @@ swc_ecma_codegen = { version = "8.0.2", path = "../swc_ecma_codegen" }
 swc_ecma_minifier = { version = "12.4.0", path = "../swc_ecma_minifier" }
 swc_ecma_parser = { version = "10.0.2", path = "../swc_ecma_parser" }
 swc_ecma_visit = { version = "8.0.0", path = "../swc_ecma_visit" }
+swc_error_reporters = { path = "../swc_error_reporters" }
```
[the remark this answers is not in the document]
kdy1: I think it's very strange, but `cargo` denies publishing a library if it has a path dependency. I don't like the behavior of `cargo`, but there's no option :(

crates/swc_error_reporters/src/error.rs
```diff
@@ -0,0 +1,15 @@
...
+/// parsing phase.
+///
+/// Although a syntax error is reported by the SWC parser, this error is handled
+/// on the JavaScript side. The Rust side will ignore it during processing.
+#[derive(Debug)]
+pub struct OnlyDiagnosticsError;
+
+impl std::fmt::Display for OnlyDiagnosticsError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Syntax Error")
```
kdy1, line 11: If this is the error message, I think it's better to name is `SyntaxError`, although it can be done in a next PR anyway

packages/core/__tests__/transform/api_test.js
```diff
@@ -145,7 +145,7 @@ it("should respect isModule = true", async () => {
         swc.transformSync(`const five = 005`, {
             isModule: true,
         });
-    expect(f).toThrowError(/Syntax Error/);
```
kdy1, line 148: This should be reverted

crates/swc_compiler_base/Cargo.toml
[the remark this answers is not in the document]
kdy1: But actually I think it's better to do within the final PR (the final PR for `dev` branch)

crates/swc_error_reporters/tests/fixture/2.ans
```diff
@@ -1,21 +1,18 @@
-
-  [38;2;255;30;30m×[0m test
+  [31m×[0m test
```
kdy1, line 1: FYI.
image

crates/swc_error_reporters/tests/fixture/1.ans
```diff
@@ -1,7 +1,6 @@
-
-  [38;2;255;30;30m×[0m test
+  [31m×[0m test
```
kdy1, line 1: image

kdy1 (commented):
Btw, I love this PR :)

crates/swc_compiler_base/Cargo.toml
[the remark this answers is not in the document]
GiveMe-A-Name: Yes, i agree

packages/core/__tests__/transform/api_test.js
GiveMe-A-Name: It was reverted.

bindings/binding_core_wasm/__tests__/error.js
GiveMe-A-Name: I remove extra error struct. 
And I kept the original writing, continue using anyhow::Error::context o ensure print correctly, reference to https://github.com/swc-project/swc/blob/main/crates/swc_error_reporters/src/handler.rs#L153

kdy1 (approved):
Thank you so much!

magic-akari:
Hi, I noticed this thread mentioned throwing a SyntaxError through the N-API. Was this feature implemented?

From the results, it appears it wasn't. (Correct me if I'm wrong)
Were there technical constraints preventing implementation?

PR #10283, opened 2025-03-27
feat(es/minifier): Merge expression with empty return

Austaras:
**Description:**

**BREAKING CHANGE:**


crates/swc/tests/libs-size.snapshot.md
```diff
@@ -1,14 +1,14 @@
...
-| terser.js | 1.08 MiB | 444.91 KiB | 120.44 KiB |
-| three.js | 1.19 MiB | 627.90 KiB | 154.58 KiB |
-| typescript.js | 10.45 MiB | 3.17 MiB | 846.04 KiB |
-| victory.js | 2.30 MiB | 689.65 KiB | 153.64 KiB |
-| vue.js | 334.13 KiB | 112.96 KiB | 41.69 KiB |
+| terser.js | 1.08 MiB | 444.95 KiB | 120.45 KiB |
+| three.js | 1.19 MiB | 627.99 KiB | 154.58 KiB |
+| typescript.js | 10.45 MiB | 3.17 MiB | 846.08 KiB |
+| victory.js | 2.30 MiB | 689.66 KiB | 153.64 KiB |
+| vue.js | 334.13 KiB | 112.98 KiB | 41.69 KiB |
```
Austaras, line 14: Why there're two size test and their results are different

crates/swc_ecma_minifier/src/compress/optimize/sequences.rs
```diff
@@ -190,16 +190,8 @@ impl Optimizer<'_> {
                         }
 
                         Stmt::Return(mut stmt @ ReturnStmt { arg: Some(..), .. }) => {
-                            match stmt.arg.as_deref_mut() {
-                                Some(e) => {
-                                    e.prepend_exprs(take(&mut exprs));
-                                }
-                                _ => {
```
Austaras, line 197: This never get called.

crates/swc_ecma_minifier/src/compress/optimize/mod.rs
```diff
@@ -2578,6 +2578,37 @@ impl VisitMut for Optimizer<'_> {
         debug_assert_eq!(self.prepend_stmts.len(), prepend_len);
         debug_assert_eq!(self.append_stmts.len(), append_len);
         debug_assert_valid(s);
+
+        if let Stmt::Block(block) = s {
```
kdy1, line 2582: Does it need to be in the full optimizer?
It seems like it's not using any field from `self`

crates/swc/tests/libs-size.snapshot.md
kdy1: I think the difference comes from codegen option difference. It would be fine to remove the one in `swc_ecma_minifier` IMO

crates/swc_ecma_minifier/src/compress/optimize/mod.rs
Austaras: No. It's just because the original code is in optimize

crates/swc_ecma_minifier/tests/terser_exec.rs
```diff
@@ -57,7 +57,10 @@ use testing::assert_eq;
         "blocks/issue_1672_for",
         // parser error
         "arrow/async_identifiers",
-        "async/async_identifiers"
+        "async/async_identifiers",
+        // lint error
+        "sequences/delete_seq_4/input.js",
+        "sequences/delete_seq_5/input.js"
```
Austaras, line 63: Way out of scope of this PR.

crates/swc_ecma_minifier/tests/fixture/issues/moment/1/output.js
```diff
@@ -910,10 +904,7 @@ function(global, factory) {
         }) : ("string" == typeof input && (config._i = input = config._locale.preparse(input)), isMoment(input)) ? new Moment(checkOverflow(input)) : (isDate(input) ? config._d = input : isArray(format) ? !// date from string and array of format strings
         function(config) {
             var tempConfig, bestMoment, scoreToBeat, i, currentScore, validFormatFound, bestFormatIsValid = !1;
-            if (0 === config._f.length) {
-                getParsingFlags(config).invalidFormat = !0, config._d = new Date(NaN);
-                return;
-            }
+            if (0 === config._f.length) return getParsingFlags(config).invalidFormat = !0, config._d = new Date(NaN);
```
Le0Developer, line 907: This function no longer returns `undefined` but an Invalid Date

crates/swc_ecma_minifier/tests/fixture/next/47005/output.js
```diff
@@ -327,10 +327,7 @@
                     r,
                     _
                 ]), (0, A.useEffect)(()=>{
-                    if (P) {
-                        en(P);
-                        return;
-                    }
+                    if (P) return void en(P);
```
Le0Developer, line 330: This became longer (like most occurrences). I tested the gzip output as well and it's one byte longer. However with brotli it's 3 bytes shorter (thanks to saving whitespace). After minification (by esbuild), this example is consistently worse.

```js
if(P){en(P);return} // before
if(P)return void en(P); // after
```

| version | no compression | gzip | brotli |
| --- | --- | --- | --- |
| before | 19177  | **4308** | 3750
| after this pr | **19109** [1] | 4309 (+1) | **3747** (-3)
| |
| before + minified [2] | **8421** | **3480** | **3099**
| after + minified [2] | 8425 (+4) | 3481 (+1) | 3103 (+4)

[1]: The uncompressed size after this PR is smaller because whitespace was removed.
[2]: Minified using esbuild.

TLDR: I don't see how this optimization is worth it. While the `return void` looks like a juicy back-reference for compressors, the `void` would normally not be there, so we're saving a semicolon at most. Sometimes we also save the BlockStatement for an if-statement, but as the example above shows, even with this it's not worth it.
The juicy back-reference is also lost for comma expressions (`return a, void b`).

The few improved benchmarks seem to be related to better inlining (see jquery.js).

Le0Developer (commented):
I can't comment on the rust code, but the output js. Found ~~one behavior regression and~~ overall size is also worse unless the code gets better inlined.

This seems to be the only pattern where code size improves thanks to the ternary:
```js
// input
if(cond) {
  doSomething();
  return
}
return somethingElse();

// output
return cond ? void doSomething() : somethingElse();
```

crates/swc_ecma_minifier/tests/fixture/next/47005/output.js
Austaras: The general idea is that after gzip, void wouldn't take as much as 4 bytes and there's enough occurence of void because undefined would be turn into void 0

crates/swc_ecma_minifier/tests/fixture/issues/moment/1/output.js
Austaras: This is caused by `ignore_return_value`

crates/swc_ecma_minifier/tests/fixture/next/47005/output.js
Austaras: The situation where an extra paren is needed has alreay been exclued.

Austaras: We're not blindly replace empty return with return void like terser. This compress is only done when a block can be eliminated.

Le0Developer: Yeah but it looks like it's still not worth it unless it gets transformed into a ternary return.

Austaras: The benefit could also come from `ignore_return_value` as mentioned before. And `d3.js` showcase where this optimizaion itself brings the size down.

Austaras: And we should have a better `if_return` anyway.

Le0Developer: Oh absolutely there's benefits, but currently it's a mixed bag:
image

crates/swc/tests/libs-size.snapshot.md
kdy1: I'll remove one in `swc` with another PR after merging this

kdy1 (approved):
Thanks!

kdy1:
I'll merge this after waiting for https://github.com/swc-project/swc/actions/runs/14141123198/job/39622522587

PR #10289, opened 2025-03-28
fix(html): remove `async` from `minifyFragmentSync`

GeorgeTaveras1231:
**Description:**

`minifyFragmentSync` returns a Promise which seems wrong

**BREAKING CHANGE:**

N/A - Technically breaks APIs but this seems like an unintentional API

### DIFF
--- a/packages/html/index.ts
+++ b/packages/html/index.ts
@@ -65,7 +65,7 @@ export function minifySync(
     return binding.minifySync(content, toBuffer(options ?? {}));
 }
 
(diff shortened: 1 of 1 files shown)


kdy1 (approved):
Thanks! I missed it. 

Internal type is correct. https://github.com/swc-project/swc/blob/b2a48bc08dcc31f66bb7d9d0745f4f6ffcdb38ca/bindings/binding_html_node/src/lib.rs#L751-L761

PR #10292, opened 2025-03-29
feat(es/minifier): Optimize code generated by optional chaining

Austaras:
**Description:**

**BREAKING CHANGE:**


crates/swc_ecma_minifier/src/compress/optimize/bools.rs
```diff
@@ -167,34 +184,25 @@ impl Optimizer<'_> {
...
-                                op: op!("!="),
-                                left: cmp.take(),
-                                right: Lit::Null(Null { span: DUMMY_SP }).into(),
-                            });
-                        }
-                    }
-                    _ => {}
-                }
+        if let (Value::Known(Type::Undefined), Value::Known(Type::Null))
+        | (Value::Known(Type::Null), Value::Known(Type::Undefined)) = (lt, rt)
```
kdy1, line 188:
```suggestion
        | (Value::Known(Type::Null | Type::Undefined)) = (lt, rt)
```

crates/swc_ecma_minifier/src/compress/optimize/bools.rs
```diff
@@ -167,34 +184,25 @@ impl Optimizer<'_> {
...
-                                span,
-                                op: op!("!="),
-                                left: cmp.take(),
-                                right: Lit::Null(Null { span: DUMMY_SP }).into(),
-                            });
-                        }
-                    }
-                    _ => {}
-                }
+        if let (Value::Known(Type::Undefined), Value::Known(Type::Null))
```
kdy1, line 187:
```suggestion
        if let (Value::Known(Type::Undefined | Type::Null))
```

kdy1 (commented):
It looks good in general, but I think this would be more readable

Austaras:
Yeah I would also like to give it a better name.

crates/swc_ecma_minifier/src/compress/optimize/bools.rs
Austaras: This is to ensure they're `undefined | null` or `null | undefined`, but never `undefined | undefined` or `null | null`.

kdy1:
Sounds good. Can you rename it within this PR?

crates/swc_ecma_minifier/src/compress/optimize/bools.rs
Austaras: Of course there would not be any real code like this.

kdy1: Ah. Thanks. I misread it.

kdy1 (dismissed):

kdy1:
Do you want to rename the function in this PR? I'm fine with either of the options of creating another PR or modifying within this PR

Austaras:
> Do you want to rename the function in this PR? I'm fine with either of the options of creating another PR or modifying within this PR

I've renamed it.

kdy1:
I think you made a mistake? Maybe it is a bug of Github, but to be sure. (It's not showing on the diff view)

Austaras:
Sorry I pushed wrong commit.

kdy1 (dismissed):

PR #10306, opened 2025-04-02
fix: worker_threads issue on glibc platform

Brooooooklyn:
Ref:
- https://github.com/napi-rs/napi-rs/pull/2494
- https://github.com/rust-lang/rust/issues/91979


kdy1 (approved):
Thank you!

PR #10317, opened 2025-04-05
 fix(es/compat): Remove one promise tick in yield* (tc39/ecma262#2819)

magic-akari:
**Related issue:**

- babel/babel#14877
- tc39/ecma262#2819


magic-akari:
> [!NOTE] 
The new release of swc/core requires a corresponding version bump for swc/helpers.

magic-akari:
The tests failed on CI for the newly added exec test I added, though they run successfully on my local machine.
Is it possible to clear the cache and rerun the tests?

kdy1:
Oh, the cache does not take the version or content helpers into account. I think that's the cause. I'll try to fix it tomorrow.

kdy1 (approved):
Thank you!

kdy1:
I published a new version of `@swc/helpers` via https://github.com/swc-project/swc/commit/8b0d42bdce54797f045f07441218faef7a35ad45

magic-akari:
> I published a new version of `@swc/helpers` via [8b0d42b](https://github.com/swc-project/swc/commit/8b0d42bdce54797f045f07441218faef7a35ad45)

To ensure compatibility, we should enforce a minimum version requirement of swc/helpers@0.5.17 for the new swc/core release. However, I noticed there's currently no version constraint specified in the package.json here:
https://github.com/swc-project/swc/blob/8b0d42bdce54797f045f07441218faef7a35ad45/packages/core/package.json#L61

Is there a way to implement this version restriction, or do we have an alternative strategy for managing dependencies in this case?

kdy1:
We can use `peerDependencies` and `peerDependenciesMeta`.
The version we specify in `dependencies` is only for `@swc/core` package, but if we specify `peerDependencies` the package manager will install the correct version. If we also use peerDependenciesMeta, we can mark it optional, too. Can you send a PR?

PR #10320, opened 2025-04-07
chore: use const fn in Span

### LINKED ISSUE #9962: `swc_common`: Make `Span::new` const fn

bvanjoi:
**Description:**

Using `const fn` can generally help lower both runtime overhead.

**BREAKING CHANGE:**

nothing.

Fixes #9962


crates/swc_common/src/syntax_pos.rs
```diff
@@ -385,71 +385,71 @@ extern "C" {
...
 impl Span {
     #[inline]
-    pub fn lo(self) -> BytePos {
+    pub const fn lo(self) -> BytePos {
         self.lo
     }
 
     #[inline]
-    pub fn new(mut lo: BytePos, mut hi: BytePos) -> Self {
-        if lo > hi {
```
kdy1, line 394: Actually #9962 is about removing this swap

crates/swc_common/src/syntax_pos.rs
```diff
@@ -385,71 +385,71 @@ extern "C" {
...
-            std::mem::swap(&mut lo, &mut hi);
+    pub const fn new(lo: BytePos, hi: BytePos) -> Self {
+        if lo.0 > hi.0 {
```
kdy1, line 394: This should be removed and callers should be fmodified to call Span::new with correct argument

crates/swc_common/src/syntax_pos.rs
bvanjoi: I see. But I have two questions:
• Does this change break existing code?
• Should we add `assert!(lo < hi)` or `debug_assert!(lo < hi)`?

kdy1: Yes for the first question. (It’s a bug IMO)

And debug assert would be enough considering the number of parser tests

kdy1: Moving `swap` to the caller code will make it slower. That's why simply changing it to a `const fn` slowed it. (swap operation is inlined into call side with constant swap)

image

bvanjoi:
The latest benchmark report doesn’t show any regressions (I’m guessing it’s just some noise in the `es/lints/libs/terser` metrics).

Not sure why the `run_fixture_test_tests__exec__generator__ecma262_pr_2819__exec_js` failed though – can’t reproduce it locally. Happy to dig deeper if we decide to move forward with landing this.

@kdy1 What do you think? Should we keep pursuing this change?

kdy1:
Are you still willing to work on removal of `std::mem::swap` calls?

PR #10323, opened 2025-04-08
perf(es/ast): Reduce redundant string comparison for `Atom`s

bvanjoi:
This PR addresses redundant string hashing operations in the `PhfHash` implementation, potentially optimizing runtime performance by eliminating duplicate hash computations


kdy1:
You don't need to rebase unless there's a merge conflict

bvanjoi:
> You don't need to rebase unless there's a merge conflict

Got it :)

---- 

Do these quantified performance gains validate the implementation approach, or would you recommend architectural refinements to optimize maintainability?

crates/swc_ecma_ast/src/ident.rs
```diff
@@ -630,6 +632,48 @@ static RESERVED: phf::Set<&str> = phf_set!(
     "with",
 );
 
+static RESERVED_ATOMS: Lazy<FxHashSet<Atom>> = Lazy::new(|| {
```
kdy1, line 635: Can we define a macro instead?
One macro invocation per each group of strings.

```rust
define_set!(RESERVED, RESERVED_ATOMS, ["break", "case", "catch", ..., "with" ]);
```

I think we can keep maintainability good as well with this simple trick

kdy1: It can expand to a `phf_set` call and `static RESERVED_ATOMS: Lazy<FxHashSet<Atom>> = Lazy::new(|| {` call

chenjiahan (commented):
Should we change `chore(ast): ...` to `perf(ast): ...`?

bvanjoi:
> Should we change chore(ast): ... to perf(ast): ...?

done

kdy1 (approved):

PR #10340, opened 2025-04-09
avoid compute len for StringInput::as_str

bvanjoi:


bvanjoi:
The performance regression stems from the `bstr::utf8::decode`, I will seek other methods

PR #10341, opened 2025-04-09
refactor(es/compat): Simplify async_to_generator

magic-akari:


magic-akari:
This refactor reduces reliance on FnEnvHoister by explicitly passing this and arguments as parameters to helper functions instead of using variable hoisting. However, when encountering super references, FnEnvHoister remains necessary to handle such cases, falling back to a slow path.

---

Compiler Implementation Comparisons

Input:

```JavaScript
async function foo() {}
class Foo {
    async foo() {}
    async use_this() {
        return await this.bar();
    }
    async use_arguments() {
        return await arguments[0].bar();
    }
}
```

[esbuild](https://esbuild.github.io/try/#dAAwLjI1LjIALS10YXJnZXQ9ZXMyMDE2AGFzeW5jIGZ1bmN0aW9uIGZvbygpIHt9CmNsYXNzIEZvbyB7CiAgICBhc3luYyBmb28oKSB7fQogICAgYXN5bmMgdXNlX3RoaXMoKSB7CiAgICAgICAgcmV0dXJuIGF3YWl0IHRoaXMuYmFyKCk7CiAgICB9CiAgICBhc3luYyB1c2VfYXJndW1lbnRzKCkgewogICAgICAgIHJldHVybiBhd2FpdCBhcmd1bWVudHNbMF0uYmFyKCk7CiAgICB9Cn0):

```JavaScript
function foo() {
  return __async(this, null, function* () {
  });
}
class Foo {
  foo() {
    return __async(this, null, function* () {
    });
  }
  use_this() {
    return __async(this, null, function* () {
      return yield this.bar();
    });
  }
  use_arguments() {
    return __async(this, arguments, function* () {
      return yield arguments[0].bar();
    });
  }
}
```

[TypeScript](https://www.typescriptlang.org/play/?target=3#code/PTAEAEBcEMCcHMCmkBcpEGcBMAGAjAGwBQAxgDbQYagBiA9naAN5GhuiUCeAdiaAGYMAFAEpmAX1bsuvUAFcMiAPqQAFgEsMo5lPbtYyObG4cA7tHWRQazQDoARnFEBuXaEl6ZfBcrjw5ALaI3JBaYix6egaQRibQ5pYcCIHBoQDaOAC6Dk4irnqS4kA):

```JavaScript
function foo() {
    return __awaiter(this, void 0, void 0, function* () { });
}
class Foo {
    foo() {
        return __awaiter(this, void 0, void 0, function* () { });
    }
    use_this() {
        return __awaiter(this, void 0, void 0, function* () {
            return yield this.bar();
        });
    }
    use_arguments() {
        var arguments_1 = arguments;
        return __awaiter(this, void 0, void 0, function* () {
            return yield arguments_1[0].bar();
        });
    }
}
```

[Babel](https://babeljs.io/repl#?browsers=&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=MYGwhgzhAEBiD29oG8BQ0PUgTwHbGgDNEAKAShQF91Md9oBXCAUwH0AXACwEsJyUamTACdm7BsNxYA7mG7toXXgDoARmGHkA3IOjUhdAkzYaA5gwC2zXOz4U0QoaPGSZchWcvXbAbQAMALpqGtq61NRAA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=true&targets=&version=7.27.0&externalPlugins=%40babel%2Fplugin-transform-async-to-generator%407.25.9&assumptions=%7B%7D):

```JavaScript
function foo() {
  return _foo.apply(this, arguments);
}
function _foo() {
  _foo = _asyncToGenerator(function* () {});
  return _foo.apply(this, arguments);
}
class Foo {
  foo() {
    return _asyncToGenerator(function* () {})();
  }
  use_this() {
    var _this = this;
    return _asyncToGenerator(function* () {
      return yield _this.bar();
    })();
  }
  use_arguments() {
    var _arguments = arguments;
    return _asyncToGenerator(function* () {
      return yield _arguments[0].bar();
    })();
  }
}
```

[SWC Previous Implementation](https://play.swc.rs/?version=1.11.18&code=H4sIAAAAAAAAA0ssrsxLVkgrzUsuyczPU0jLz9fQVKiu5UrOSSwuVnDLz1eo5lIAgkSIOpg0Qqi0ODW%2BJCOzGCQOFgaBotSS0qI8hcTyxMwSBZCsXlJikYamNVgBuu7EovTS3NS8EjxGwJVEG8SiGlULACCfNDPBAAAA&config=H4sIAAAAAAAAA02OMQ7DIAxF95wCee7QdsjQE3TpIRB1KiICyHakoCh3L5DQdPP%2Fz%2F%2Fba6cUjGzgodY8ZhE1MdJPZ4eTF71kB9BMmg3ZKHBpdOSCBu0Yq7XtBHARJK%2FdE11E4rwjNOMBRdMHpVby%2FXrrjzpwITC2uuLVNpjCe3Z4%2Figp4h7u4Vyy3g7pPwyWXy1ZjnfbF0BTCw3vAAAA):

```JavaScript
function foo() {
    return _foo.apply(this, arguments);
}
function _foo() {
    _foo = _async_to_generator(function*() {});
    return _foo.apply(this, arguments);
}
class Foo {
    foo() {
        return _async_to_generator(function*() {})();
    }
    use_this() {
        var _this = this;
        return _async_to_generator(function*() {
            return yield _this.bar();
        })();
    }
    use_arguments() {
        var _arguments = arguments;
        return _async_to_generator(function*() {
            return yield _arguments[0].bar();
        })();
    }
}

```

SWC Implementation in this PR:

```JavaScript
function _foo() {
    return _async_to_generator(function*() {})();
}
class Foo {
    foo() {
        return _async_to_generator(function*() {})();
    }
    use_this() {
        return _async_to_generator(function*() {
            return yield this.bar();
        }).call(this);
    }
    use_arguments() {
        return _async_to_generator(function*() {
            return yield arguments[0].bar();
        }).apply(this, arguments);
    }
}
```

---

> [!NOTE]  
> 1. **Parameter Passing Strategies:**
> 
>     - TypeScript/esbuild inject `this` and `arguments` via helper parameters.
>     - SWC uses `call()`/`apply()` to forward `this` and `arguments`.
>     - TypeScript hoists `arguments` (renamed to `arguments_1`), while esbuild directly uses the parameter. I expect no functional differences between them.
> 
> 2. **Babel's specific Behaviour:**
> 
>     - Always hoists `this`/`arguments` but uses `apply()` for async function declarations.
>     - Babel caches transformed async functions (e.g., `_foo`), which may improve runtime performance, though no benchmarks validate this. Esbuild/TypeScript omit caching, and I align with their simpler approach.
> 
> 3. **IIFE async function:**
>     - Previous SWC handled async IIFEs (e.g., `(async()=>{})()`) via `_async_to_generator(...)()`.
>     - However, the new implementation currently omits handling this case, producing `(()=> _async_to_generator(...)())()`, as it aligns more naturally with optimizations typically handled during minification.
> 
---

**Open Questions:**

1. Should we add back Babel-style runtime caching functions?
2. Should async IIFEs be explicitly handled, or can these optimizations be deferred to the minification phase?

---

crates/swc_ecma_transforms_compat/tests/__swc_snapshots__/tests/es2017_async_to_generator.rs/issue_8452.js
```diff
@@ -2,14 +2,8 @@ class Test0 {
...
-        var _this;
-        super(), _this = this, console.log(/*#__PURE__*/ function() {
-            var _ref = _async_to_generator(function*(e) {
-                yield _this.test();
-            });
-            return function(e) {
-                return _ref.apply(this, arguments);
-            };
-        }());
+        super(), console.log((e)=>/*#__PURE__*/ _async_to_generator(function*() {
```
magic-akari, line 5: By injecting `this` at the call site instead of hoisting it, we avoid potential order issues of `this` and `super`.

magic-akari:
The performance regression is caused by `fast_path` removal. 
While I don't know why the fast_path implementation was effective initially, **could this suggest that `visit` operations generally demonstrate better performance than `visit_mut` implementations?** 

Blindly restoring the fast_path without modifications would reintroduce certain traversal bugs, which would require additional handling logic to address.

kdy1:
> could this suggest that visit operations generally demonstrate better performance than visit_mut implementations?


Yes, `visit` is faster than `visit_mut`. My guess about the reason is compiler optimizations, but I don't think we should add `#[fast_path] ` back.

kdy1:
> Should we add back Babel-style runtime caching functions?


I think I don't need it unless someone files an issue. I don't expect such a cache to make a noticeable difference for end-users.

kdy1 (approved):

PR #10342, opened 2025-04-10
chore(ci): Fix CI

kdy1:


crates/swc/tests/exec/generator/ecma262-pr-2819/.exec.js
kdy1, line 1: /cc @magic-akari Can you take a look?

PR #10345, opened 2025-04-10
avoid compute len for StringInput::as_str

bvanjoi:
**Description:**

**BREAKING CHANGE:**


kdy1:
I'll wait for the codspeed to report the numbers for the latest changes

bvanjoi:
This isn’t panning out like I hoped, so let’s wrap it up.

PR #10348, opened 2025-04-10
test(es): Unignore tests

kdy1:
**Description:**

I wiped the Github Actions cache for the swc repository, so I'm re-enabling this.


kdy1:
cc @magic-akari It still fails

magic-akari:
I encountered the same error when I switched my Node version to 18 locally.

For our exec tests, we execute the code both before and after transformation and then compare their outputs.
Therefore, my guess is that the error is happening in the original (pre-transformation) code execution.

kdy1:
Oh then I'll update nodejs to 20

PR #10350, opened 2025-04-10
perf(es/parser): remove ascii check for no-ascii

bvanjoi:
Skip that ascii check for no-ascii char.

It probably won’t move the needle in benchmarks since non-ASCII chars are pretty rare in our data, but I still think it’s a valid optimization to make.


crates/swc_ecma_ast/src/ident.rs
```diff
@@ -275,16 +275,27 @@ impl Ident {
...
+    }
+
     /// Returns true if `c` is a valid character for an identifier start.
     #[inline]
     pub fn is_valid_start(c: char) -> bool {
         if c.is_ascii() {
             Self::is_valid_ascii_start(c as u8)
         } else {
-            unicode_id_start::is_id_start_unicode(c)
+            Self::is_valid_non_assicc_start(c)
```
kdy1, line 289: It looks like a typo?

PR #10358, opened 2025-04-11
perf(common): less bound check for chars

bvanjoi:


bvanjoi:
It's crazy that this is actually slower.

PR #10360, opened 2025-04-11
perf: more atoms

bvanjoi:
Let's cut down on string comparisons where possible.


bvanjoi:
Not seeing enough gains here

PR #10361, opened 2025-04-11
chore(es/parser): init cap for some vec

bvanjoi:


kdy1 (approved):
Thanks! I verified that it's 0.x seconds faster overall

PR #10364, opened 2025-04-12
perf(es/parser): Reduce clone of token contexts

bvanjoi:
I will close this if it's not shown any improvements.


kdy1 (approved):

kdy1:
It seems like it's about 0.3% ~ 0.5% faster.

PR #10367, opened 2025-04-12
perf(es/parser): Use `bitflags` to reduce parser context size

CPunisher:
`Context` size: 30 bytes -> 4 bytes


kdy1 (approved):
Wow! It's HUGE. Thank you!

kdy1 (changes requested):
Oops. CI failed

kdy1 (approved):

PR #10369, opened 2025-04-13
perf(es/parser): use `arrayvec` and unsafe `push` to optimize escaped string parsing

CPunisher:
Since we can make sure that `buf` has enough capacity, we can avoid bound checks to improve performance.

This could only take effects when parsing escaped chars.


kdy1:
Can you rebase?

kdy1 (dismissed):

kdy1 (approved):

PR #10370, opened 2025-04-13
perf(es/parser): replace byte arguments with generics

CPunisher:
I think this could introduce more compiler optimization.

Let's wait for the benchmarks


kdy1 (approved):
image

Good idea, it's slightly faster

Issue #9778, opened 2024-12-02
When unknown browser version used, all polyfills are added to the bundle


erykpiast:
### Describe the bug

When I use a browser version unknown for SWC in the browserslist selector, instead of adding no transipilation and polyfills (this is the default @babel/preset-env behavior), everything is transpilled and polyfilled. That's a bit unfortunate, as the browserslist-rs tool used under the hood recognizes only browsers released before June 2024. https://github.com/browserslist/browserslist-rs/pull/25. I'm afraid changing how to handle defaults it's beyond my Rust knowledge 😅

### Input code

```typescript
async function foo({ ...props }) {
  await console.log({ ...props });
}

foo();
```

### Config

```json
{
  "minify": false,
  "env": {
    "targets": ["Chrome > 130"]
  },
  "jsc": {
    "preserveAllComments": true,
    "parser": {
      "syntax": "ecmascript",
      "jsx": true
    },
    "minify": {
      "mangle": false,
      "format": {
              "asciiOnly": false,
        "comments": "some"
      },
      "compress": {
        "defaults": false,
              "unused": true,
              "dead_code": true,
              "toplevel": true
      }
    },
    "loose": false
  }
}
```

### Playground link (or link to the minimal reproduction)

https://play.swc.rs/?version=1.9.3&code=H4sIAAAAAAAAA0ssrsxLVkgrzUsuyczPU0jLz9eoVtDT0ysoyi8oVqjVVKjmUlBILE%2FMLFFIzs8rzs9J1cvJT0dVY81Vy8UF0qlpDQAfwt5ETwAAAA%3D%3D&config=H4sIAAAAAAAAA3VRQYrDMAy89xVG5x627K2HQukD9gFLWYyttC6yHSwnNJT8vYpJ3aSwPlkzo7E0fmyUAu%2BCawbYq0YT43aCMPRSP%2BQqRdbpgpkF%2BIXTNUWP6qB2319wFn4s%2Bhubt75NyJh6PBKdovcYSm9OXfGeBDqJoDYIwkPI%2Bi4IoPGaTXJthu2LvfF9NijIONvUuauN1%2BFCuNykwE1MXueF8HVAnnLuJ9Dw2SOcec8OLEvDTI3VVxTTrrxyBouN7qj0fVjOgi50jHYVyYK1qO2fiRb%2FE%2BTYEvZIy0hkqlU0FCPXHKZf2oxPuR%2FOf%2BsBAAA%3D

### SWC Info output

[...]

PR #10376, opened 2025-04-15
fix: Skip all transpilation and adding any polyfills when unknown browser version used

Guocork:
**Description:**

Skip all transpilation and adding any polyfills when unknown browser version used

**BREAKING CHANGE:**
I need some help. Right now, my approach is: 
when an unsupported browser version is set in browserslist-rs — for example:
```
"env": {
  "targets": ["Chrome > 130"]
}
```
Because the [`targets_to_versions`](https://github.com/swc-project/swc/blob/main/crates/swc_ecma_preset_env/src/lib.rs#L48C19-L48C38) has the  [`ignore_unkonwn_versions`](https://github.com/swc-project/swc/blob/main/crates/preset_env_base/src/query.rs#L57) option enabled in  [`browserslist_opts`](https://github.com/swc-project/swc/blob/main/crates/preset_env_base/src/query.rs#L106), it ends up returning an empty result, and [`is_any_target`](https://github.com/swc-project/swc/blob/main/crates/swc_ecma_preset_env/src/lib.rs#L49) becomes true.

So, I added a check to skip processing when the `targets` list is empty. But when I run `cargo test`, not all the tests pass. I'm not sure if there's a problem with my approach, or if I’m missing some edge cases.

#9778

### DIFF
--- a/crates/swc_ecma_preset_env/src/lib.rs
+++ b/crates/swc_ecma_preset_env/src/lib.rs
@@ -58,7 +58,7 @@ where
             let f = transform_data::Feature::$feature;
             !exclude.contains(&f)
                 && (c.force_all_transforms
-                    || (is_any_target
+                    || (!is_any_target
                         || include.contains(&f)
                         || f.should_enable(&targets, c.bugfixes, $default)))
         }};
@@ -369,6 +369,9 @@ impl Polyfills {
             + VisitMutWith<corejs2::Entry>
             + VisitMutWith<corejs3::Entry>,
     {
+        if self.targets.is_any_target() {
+            return Default::default();
+        }
         let required = match self.mode {
             None => Default::default(),
             Some(Mode::Usage) => {


crates/swc_ecma_preset_env/src/lib.rs
```diff
@@ -58,7 +58,7 @@ where
             let f = transform_data::Feature::$feature;
             !exclude.contains(&f)
                 && (c.force_all_transforms
-                    || (is_any_target
+                    || (!is_any_target
```
kdy1, line 61: `!is_any_target` is wrong way to detect unknown browser version

kdy1, line 61: `!is_any_target` is not a way to detect an unknown browser version

kdy1 (changes requested):
image

You have to look at failing test cases. You would be able to adjust the condition to create the correct condition for the input case.

PR #10377, opened 2025-04-15
refactor(ecma/lexer): split lexer of parser

bvanjoi:
We are not ensure each token is correct during parse js/ts code, so let's split `parse/lexer` into single crates and this crate can ensure lexer is correctly


bvanjoi:
It's surprising that these failing test cases parse correctly in the playground environment, yet were passing in our earlier test suite.

For instance, the failing test case in `test262::error_reporting::fail::0ff3826356c94f67.js` containing `({function} = 0)` was parsed successfully but generated no error reports before.

crates/swc_ecma_lexer/benches/files/angular-1.2.5.js
kdy1, line 1: Can we remove the files in `swc_ecma_lexer/benches/files` and reference the files in `swc_ecma_parser` using `../swc_ecma_parser/benches/files/xxx` instead of copying these files?

bvanjoi:
> Please feel free to fix CI in any direction 

The issue occurs because `swc_ecma_lexer` doesn't receive the `feature="verify"` flag. Moving `is_reserved_word` to `swc_ecma_parser` resolves this.

kdy1:
You can do

```toml
[features]
verify = ["swc_ecma_lexer/verify"]

```

from the swc_ecma_parser/Cargo.toml

bvanjoi:
Appreciate the clarification! The design pattern makes sense now.

kdy1 (approved):
Thanks!

Issue #10379, opened 2025-04-16
[Tracking] Use bitflags to reduce context size in the project


CPunisher:
It has been proved in https://github.com/swc-project/swc/pull/10367 that reducing context struct size can bring considerable performance improvement. Since there are lots of `Context` with similar programming pattern in swc project, we can refactor them one by one.

I will list all the tasks here and finish them in orders. cc @kdy1 

- [x] `swc_ecma_usage_analyzer::analyzer`: https://github.com/swc-project/swc/pull/10380
- [x] `swc_ecma_usage_analyzer::alias`: https://github.com/swc-project/swc/pull/10387
- [x] `swc_ecma_minifier::compress::optimize`: https://github.com/swc-project/swc/pull/10381

kdy1:
You may not need to optimize CSS crates, as those are not actively maintained.

PR #10380, opened 2025-04-16
perf(es/minifier): Use bigflags to reduce context size of analyzer

CPunisher:
Related https://github.com/swc-project/swc/issues/10379

context 16 bytes -> 8 bytes

Let's wait for the benchmark results

### DIFF
--- a/.changeset/five-jobs-teach.md
+++ b/.changeset/five-jobs-teach.md
@@ -1,0 +1,7 @@
+---
+swc_ecma_minifier: minor
+swc_ecma_usage_analyzer: minor
+swc_core: minor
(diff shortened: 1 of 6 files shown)


CPunisher:
image

It seems ok, although I'm not sure if the issue icons matter.

kdy1:
Those issue icons are fine. To remove it we should use the default malloc
but it would make benchmark too slow.

kdy1:
Those issue icons are fine. To remove it we should use the default malloc but it would make benchmark too slow.

kdy1 (approved):

PR #10381, opened 2025-04-16
perf(es/minifier): Use bitflags to reduce compress context size

CPunisher:
Related https://github.com/swc-project/swc/issues/10379

context 48 bytes -> 24 bytes

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/optimize/props.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/props.rs
@@ -2,7 +2,7 @@ use swc_common::{util::take::Take, DUMMY_SP};
 use swc_ecma_ast::*;
 use swc_ecma_utils::{contains_this_expr, private_ident, prop_name_eq, ExprExt};
 
-use super::{unused::PropertyAccessOpts, Optimizer};
+use super::{unused::PropertyAccessOpts, BitCtx, Optimizer};
 use crate::util::deeply_contains_this_expr;
 
 /// Methods related to the option `hoist_props`.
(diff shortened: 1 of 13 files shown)


crates/swc_ecma_minifier/src/compress/optimize/props.rs
```diff
@@ -248,7 +248,10 @@ impl Optimizer<'_> {
             return;
         }
 
-        if self.ctx.is_update_arg || self.ctx.is_callee || self.ctx.is_exact_lhs_of_assign {
+        if self.ctx.bit_ctx.contains(BitCtx::IsUpdateArg)
+            || self.ctx.bit_ctx.contains(BitCtx::IsCallee)
+            || self.ctx.bit_ctx.contains(BitCtx::IsExactLhsOfAssign)
+        {
```
kdy1, line 254:
```suggestion
        if self.ctx.bit_ctx.contains(BitCtx::IsUpdateArg | BitCtx::IsCallee | BitCtx::IsExactLhsOfAssign)
        {
```

Not sure if rustc will be able to optimize this, but this would be better in worst case

magic-akari:
```suggestion
        if self.ctx.bit_ctx.intersects(BitCtx::IsUpdateArg | BitCtx::IsCallee | BitCtx::IsExactLhsOfAssign)
        {
```

Replacing `contains` with `intersects` better matches the original implementation.

kdy1: Oh you are right. Sorry I was too sleepy at the moment because of the zet lag.

kdy1 (approved):

PR #10388, opened 2025-04-17
perf(es/parser): move `found_module_item` to `Parser`

CPunisher:
I'm not sure whether this is correct. I just think `found_module_item` is a property of the whole Program rather than one function scope. This may reduce the cost of creating `State`(72 bytes -> 64 bytes) in `parse_fn_body`.


kdy1 (approved):
It seems like it's slightly faster! 
image

PR #10394, opened 2025-04-20
dont mark line break for tpl token

bvanjoi:
This change seems theoretically sound (though needs confirmation) – we should verify through CI testing.


bvanjoi:
Closing this PR as it:
- Does not address any known defects;
- Delivers no measurable performance improvements;
- Fails CI checks (though the failure appears unrelated to these changes).

PR #10397, opened 2025-04-20
perf(es/parser): Defne `Token` as `TokenType` + `TokenValue`

kdy1:
**Description:**

I expect this to improve the performance of the parser greatly.

cc @bvanjoi 
Can you help with this task?

### DIFF
--- a/crates/swc_ecma_lexer/src/token.rs
+++ b/crates/swc_ecma_lexer/src/token.rs
@@ -11,147 +11,17 @@ use swc_atoms::{atom, Atom, AtomStore};
 use swc_common::{Span, Spanned};
 pub(crate) use swc_ecma_ast::{AssignOp, BinaryOp};
 
-pub(crate) use self::{Keyword::*, TokenType::*};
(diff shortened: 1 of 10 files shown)


bvanjoi:
OK, I understand the meaning of this change. I will create a new PR to implement it.

crates/swc_ecma_lexer/src/token.rs
```diff
@@ -11,211 +11,120 @@ use swc_atoms::{atom, Atom, AtomStore};
...
-                }
-            }
-        }
-    };
+#[derive(Clone, PartialEq)]
+pub struct Token {
+    pub kind: TokenType,
+    pub span: Span,
+    /// Had a line break before this token?
+    pub had_line_break: bool,
```
bvanjoi, line 21: `had_line_break` also can move into `Parser::state`, see: https://github.com/microsoft/TypeScript/blob/main/src/compiler/scanner.ts#L1071

kdy1:
Closing in favor of https://github.com/swc-project/swc/pull/10399

PR #10399, opened 2025-04-21
refactor(es/parser): Split parser into also-lex/parse-only

bvanjoi:
This constitutes a substantial code revision. We should first evaluate whether it yields measurable performance enhancements.


bvanjoi:
The performance gains demonstrate a compelling case for implementation. 

Let me explain my idea:
- First, split `Token` into `TokenKind` (the type of token, which is still referred to as `Token` in this PR) and `TokenValue`. This means we can't retrieve the correct combination of `TokenKind` and `TokenValue` after parsing, because the `TokenValue` has already been consumed during parsing.
- And then, these changes mean there are now two different types of `Token` between `swc_ecma_lexer` and `swc_ecma_parser`: one contains the `TokenValue`, while the other does not.
- The points above mean that using a single Parser between `swc_ecma_lexer` and `swc_ecma_parser` is more complicated. We would need to create a trait to handle different types of `Token` and manage their behaviors, and the ROI for this seems low.
- So, I made the following changes: I copied the `swc_ecma_parser/parser` into `swc_ecma_lexer`, meaning that `swc_ecma_lexer` now includes the `Parser` itself. I also created a `swc_ecma_parser/lexer` folder, which contains a lexer that serves only the `swc_ecma_parser/Parser` and is not exposed externally. With these changes, if you want to generate the correct tokens, you should use `swc_ecma_lexer/{Lexer, Parser}`, while the `swc_ecma_parser/Parser` is now only for quickly generating the AST.
- Based on these changes, I’ve removed some tests that ensured the lexer worked correctly in `swc_ecma_parser` and switched to using `swc_ecma_lexer` in `swc_fast_ts_strip`.
- Another change is simply replacing `Token` with `TokenKind` and `TokenValue`.

@kdy1  Do you agree with these changes? If so, I will fix CI and make it ready for review.

CPunisher:
I have seen that the two lexers and two parsers in `swc_ecma_lexer` and `swc_ecma_parser`. Do we need maintain all of them or one pair is WIP while another is released?

bvanjoi:
> The performance gains demonstrate a compelling case for implementation.
> 
> Let me explain my idea:
> 
> * First, split `Token` into `TokenKind` (the type of token, which is still referred to as `Token` in this PR) and `TokenValue`. This means we can't retrieve the correct combination of `TokenKind` and `TokenValue` after parsing, because the `TokenValue` has already been consumed during parsing. (edit: Unless the `TokenValue` is stored elsewhere, but this is unnecessary for parsing.)
> * And then, these changes mean there are now two different types of `Token` between `swc_ecma_lexer` and `swc_ecma_parser`: one contains the `TokenValue`, while the other does not.
> * The points above mean that using a single Parser between `swc_ecma_lexer` and `swc_ecma_parser` is more complicated. We would need to create a trait to handle different types of `Token` and manage their behaviors, and the ROI for this seems low.
> * So, I made the following changes: I copied the `swc_ecma_parser/parser` into `swc_ecma_lexer`, meaning that `swc_ecma_lexer` now includes the `Parser` itself. I also created a `swc_ecma_parser/lexer` folder, which contains a lexer that serves only the `swc_ecma_parser/Parser` and is not exposed externally. With these changes, if you want to generate the correct tokens, you should use `swc_ecma_lexer/{Lexer, Parser}`, while the `swc_ecma_parser/Parser` is now only for quickly generating the AST.
> * Based on these changes, I’ve removed some tests that ensured the lexer worked correctly in `swc_ecma_parser` and switched to using `swc_ecma_lexer` in `swc_fast_ts_strip`.
> * Another change is simply replacing `Token` with `TokenKind` and `TokenValue`.
> 
> @kdy1 Do you agree with these changes? If so, I will fix CI and make it ready for review.

bvanjoi:
> Do we need maintain all of them

Yes, I think so. Perhaps we can find a better way to integrate it.

kdy1:
Why do we need two separate parser/lexer/token types? 
What's the problem you see in the token definition in https://github.com/swc-project/swc/pull/10397 ?

bvanjoi:
> Why do we need two separate **token** types?

The `value` field defined in `Token`([source](https://github.com/swc-project/swc/pull/10397/files#diff-b9ff6e1704d57c6d063326c9d5a20588f5b9816bbdccb6e8a05ec664c6f481beR22)) is not required for parsing. Moving it to the outer struct would reduce memory usage.

Additionally, the `had_line_break` field in `Token` during parsing also unnecessary and can be removed in the future.

> Why do we need two separate ** lexer** ?

Since the lexical analysis implementations differ, `swc_ecma_lexer/lexer` guarantees token correctness while `swc_ecma_parser/lexer` does not maintain the same level of validation.

> Why do we need two separate **parser**?

Using a single parser to handle different lexical logic introduces unnecessary development complexity. While I've separated them for now, I believe there should be a better way to consolidate this logic.

CPunisher:
> The `value` field defined in `Token`([source](https://github.com/swc-project/swc/pull/10397/files#diff-b9ff6e1704d57c6d063326c9d5a20588f5b9816bbdccb6e8a05ec664c6f481beR22)) is not required for parsing.

I'm not sure why `TokenValue` is not required for parsing. For example, doesn't an identifier token required for constructing identifier ast node?

CPunisher:
> * This means we can't retrieve the correct combination of `TokenKind` and `TokenValue` after parsing, because the `TokenValue` has already been consumed during parsing(edit: Unless the `TokenValue` is stored elsewhere, but this is unnecessary for parsing)

I guess this is the root cause of the splitting. But I'm unclear about it. Could you please provide more detailed examples?

bvanjoi:
> I'm not sure why TokenValue is not required for parsing. 

`TokenValue` is not required **in the `Token`**. For example:

```
a + b
^        --> `token: ident`, `lexer.state.token_value: a`
    ^    --> `token: ident`, `lexer.state.token_value: b`
    
// rather than
a + b
^        --> `token: token_kind: ident, token_value: a`
    ^    --> `token: token_kind: ident, token_value: a` // we dont store `token_value` in each `token`
```

kdy1:
> The value field defined in Token([source](https://github.com/swc-project/swc/pull/10397/files#diff-b9ff6e1704d57c6d063326c9d5a20588f5b9816bbdccb6e8a05ec664c6f481beR22)) is not required for parsing. Moving it to the outer struct would reduce memory usage.
>
> Additionally, the had_line_break field in Token during parsing also unnecessary and can be removed in the future.

I understood, but I think the gain is too small if we have to maintain two tokens/lexers/parsers. My guess is that #10397 is the sweet point, but not sure.

bvanjoi:
>  In other words, what's the difference between them now

The `swc_ecma_parser/Parser` implementation prioritizes AST generation performance at the cost of potential token validation gaps, whereas `swc_ecma_lexer/Parser` enforces strict token-level correctness guarantees during AST construction.

> I think the gain is too small if we have to maintain two tokens/lexers/parsers

I suggest considering landing this change for three reasons:
- ​Using a separate lexer and parser between `swc_ecma_parser` and `swc_ecma_lexer` is the foundation for future optimizations (such as removing `TokenContexts`, which requires using a custom lexer). I think with this change we can eventually achieve the same performance as `oxc/parser`.
​- Strict token validation is not commonly needed, and sacrificing performance improvements for it may not be a good trade-off.
- ​We can reuse code by creating a shared crate (e.g., `swc_ecma_lexer_common`) to store common functions and traits.

bvanjoi:
> > In other words, what's the difference between them now
> 
> The `swc_ecma_parser/Parser` implementation prioritizes AST generation performance at the cost of potential token validation gaps, whereas `swc_ecma_lexer/Parser` enforces strict token-level correctness guarantees during AST construction.
> 
> > I think the gain is too small if we have to maintain two tokens/lexers/parsers
> 
> I suggest considering landing this change for three reasons:
> 
> * ​Using a separate lexer and parser between `swc_ecma_parser` and `swc_ecma_lexer` is the foundation for future optimizations (such as removing `TokenContexts`, which requires using a custom lexer; and remove the `had_line_break ` field in `Token`, which need different `Token` type). I think with this change we can eventually achieve the same performance as `oxc/parser`.
> * Strict token validation is not commonly needed, and sacrificing performance improvements for it may not be a good trade-off.
> * ​We can reuse code by creating a shared crate (e.g., `swc_ecma_lexer_common`) to store common functions and traits.

bvanjoi:
> I'm not too clear about Strict token validation. Generating correct AST is always the foundation for parser.

Consider the tokenization of `<` in `<div/>` when parsing a JSX file. Under strict parsing rules, this would be classified as a `Token::JsxTagStart`. However, in practice, treating it as a simple `Token::LessThan` is sufficient for generating a correct AST. The stricter approach requiring `Token::JsxTagStart` forces us to maintain unnecessary contextual information purely for AST construction purposes.

> why we still keep `swc_ecma_lexer/Parser` in the project

The primary purpose of `swc_ecma_lexer/Parser` is to ensure the lexer can function reliably, particularly for cases involving `try_parse` methods that may require multiple lexing attempts. Unlike a single-pass lexer, this implementation needs to handle scenarios where tokens cannot be fully determined in one pass.

(Alternatively, we might consider eventually removing `swc_ecma_lexer/Parser` entirely and simply exposing a `lexical_analysis` function that handles tokenization independently.)

> Do you mean copy the new Lexer and Parser, optimize on top of the new ones, and replace the original ones once they are completed

I'm having difficulty understanding the intended meaning here. Could you please clarify?

kdy1:
Ok, I got the idea. Would it be possible to use a single struct with const generic or generic to make maintenance easy? I'm worried about the maintenance burden.

CPunisher:
> Ok, I got the idea and I like it. Would it be possible to use a single struct with const generic or generic to make maintenance easy? I'm worried about the maintenance burden.

Agree. I expect finally we should eliminate all duplication and get a clear lexer and a clear parser (just like now, and faster), but we still make the code in the process as maintainable as possible.

kdy1:
If it's inevitable at the moment, I think we can merge this PR as is and refactor these with other PRs, but I think we should at least have concrete plan for future deduplication.

CPunisher:
Currently most of the code in these two set of lexers and parsers are identical, which is exactly not elegant...

bvanjoi:
> I'm worried about the maintenance burden.

I am also worried about the maintenance cost: At present, it seems troublesome to abstract `swc_ecma_lexer/Parser` and `swc_ecma_parser/Parser` into an independent trait. My suggestion is:
- Create a library `swc_ecma_lexer_common`, and then put the common parts of `swc_ecma_lexer` and `swc_ecma_parser` into it, and `swc_ecma_parser` will no longer depend on `swc_ecma_lexer`.
- Gradually sort out the reusable logic of the two and implement it in `swc_ecma_lexer_common`.

What do you think?

kdy1:
It sounds acceptable for now, but I want to hear opinion from @CPunisher, too.

CPunisher:
> > I'm worried about the maintenance burden.
> 
> I am also worried about the maintenance cost: At present, it seems troublesome to abstract `swc_ecma_lexer/Parser` and `swc_ecma_parser/Parser` into an independent trait. My suggestion is:
> 
> * Create a library `swc_ecma_lexer_common`, and then put the common parts of `swc_ecma_lexer` and `swc_ecma_parser` into it, and `swc_ecma_parser` will no longer depend on `swc_ecma_lexer`.
> * Gradually sort out the reusable logic of the two and implement it in `swc_ecma_lexer_common`.
> 
> What do you think?

It's a not bad way. But I **personally** prefer merging everything back into one crate and control the public items (for example, export `swc_ecma_lexer/{Lexer, Parser}` and `swc_ecma_parser/Parser`). In my opinion, this makes code reuse easier, narrow down the scope of code changes and avoid spreading things across the whole project. I'm not sure if this is a more reasonable way though.

bvanjoi:
At this stage, I don’t have a strong preference (though I’ll provide suggestions if technical advantages arise). Do you think we could use `swc_ecma_lexer/src/common` to store these shared structures?

kdy1:
So... @CPunisher, do you think this PR is good to merge in its current state, or should we wait for the `swc_ecma_lexer/src/common` refactoring?

CPunisher:
> So... @CPunisher, do you think this PR is good to merge in its current state, or should we wait for the `swc_ecma_lexer/src/common` refactoring?

Considering the improvement vs the code complexity, I think it's better to refactor first.

bvanjoi:
I plan to refactor both the `Lexer` and `Parser` into trait-based implementations, and will remove the draft status once completed.

bvanjoi:
I've identified minor performance regressions in both the `lexer` and `fast-strip`. Based on preliminary analysis, these appear to be related to trait code generation. However, I'm choosing to disregard these regressions for two reasons: (1) the impact is negligible (nanosecond-level differences), and (2) the variations may simply reflect differences between Rust compiler versions rather than actual performance issues

kdy1:
I think small regression of them is acceptable at this time. I can optimizer further at the future.

bvanjoi:
Completed, marked as ready for review.

bvanjoi:
While all tests have passed, this significant change may introduce some unintended side effects. However, we can address issues that arise on a case-by-case basis.

bvanjoi:
@kdy1  It seems Copilot isn't providing helpful review comments. Do you have any suggestions? Alternatively, can we proceed with this change and I can make further improvements base this?

kdy1:
You can ignore Copilot. I wanted to see if it would generate a good review

crates/swc_ecma_parser/src/lexer/util.rs
```diff
@@ -0,0 +1,119 @@
...
+            }
+            // Check for Unicode line breaks (rare case)
+            else if b > 127 {
+                let remaining = &input_str[pos..];
+                if let Some(c) = remaining.chars().next() {
+                    if c == '\u{2028}' || c == '\u{2029}' {
+                        self.state.had_line_break = true;
+                    }
+                    // Skip multibyte characters
+                    pos += c.len_utf8() - 1; // -1은 아래 증가분 고려
```
kdy1, line 105: Can you remove Korean comments?

crates/swc_ecma_parser/src/lexer/util.rs
```diff
@@ -0,0 +1,119 @@
...
+        let mut pos = 0;
+        let len = bytes.len();
+
+        // Byte-based scanning for faster search
+        while pos < len {
+            let b = bytes[pos];
+
+            if was_star && b == b'/' {
+                // Found comment end: "*/"
+                self.input.bump_bytes(pos + 1); // 종료 '/' 포함해서 이동
```
kdy1, line 55: Please delete korean comments

crates/swc_ecma_parser/src/lexer/token.rs
```diff
@@ -0,0 +1,1493 @@
...
+    Str,
+    Num,
+    BigInt,
+    Regex,
+    Template,
+    JSXName,
+    JSXText,
+    // Identifiers and keyword
+    Ident,
+    // Reserved keyword tokens (starting from 100)
```
kdy1, line 188: Does this start from 100?

crates/swc_ecma_parser/src/lexer/token.rs
```diff
@@ -0,0 +1,1493 @@
...
+    Try,
+    TypeOf,
+    Var,
+    Void,
+    While,
+    With,
+    Yield,
+    Module,
+
+    // TypeScript-related keywords (starting from 150)
```
kdy1, line 229: Same Q here

crates/swc_ecma_parser/src/lexer/token.rs
```diff
@@ -0,0 +1,1493 @@
...
+        Some(atom)
+    }
+
+    #[inline(always)]
+    pub(crate) const fn is_known_ident(self) -> bool {
+        let t = self as u8;
+        t >= Token::Abstract as u8 && t <= Token::Target as u8
+    }
+
+    pub(crate) const fn is_word(&self) -> bool {
```
kdy1, line 1247: Is there a reason you used `&self `instead of `self` here?

crates/swc_ecma_lexer/src/common/lexer/token.rs
```diff
@@ -0,0 +1,569 @@
...
+    fn unknown_ident(value: Atom, lexer: &mut Self::Lexer) -> Self;
+    fn is_unknown_ident(&self) -> bool;
+    fn take_unknown_ident(self, buffer: &mut Self::Buffer) -> Atom;
+    fn take_unknown_ident_ref<'b>(&'b self, buffer: &'b mut Self::Buffer) -> &'b Atom;
+
+    fn is_known_ident(&self) -> bool;
+    fn take_known_ident(&self) -> Atom;
+
+    fn starts_expr(&self) -> bool;
+    fn to_string(&self, buffer: &Self::Buffer) -> String;
```
kdy1, line 190: This does not seem to be used

crates/swc_ecma_parser/src/lexer/token.rs
```diff
@@ -0,0 +1,1493 @@
...
+                if ctx.contains(Context::Strict) =>
+            {
+                true
+            }
+
+            _ => false,
+        }
+    }
+
+    pub(crate) fn to_string(self, value: Option<&TokenValue>) -> String {
```
kdy1, line 935: Can you add `#[cold]`?

crates/swc_ecma_parser/src/lexer/token.rs
```diff
@@ -0,0 +1,1493 @@
...
+
+    pub(crate) const fn is_word(&self) -> bool {
+        matches!(
+            self,
+            Token::Null | Token::True | Token::False | Token::Ident
+        ) || (*self).is_known_ident()
+            || (*self).is_keyword()
+    }
+
+    pub(crate) fn as_word_atom(&self, value: Option<&TokenValue>) -> Option<Atom> {
```
kdy1, line 1255: Same here, `Token` is `: Copy`. You don't need `&self` and needless referencing of Copy types end to make things slower

crates/swc_ecma_parser/src/lexer/token.rs
```diff
@@ -0,0 +1,1493 @@
...
+    }
+
+    #[inline(always)]
+    pub(crate) const fn is_bin_op(self) -> bool {
+        let t = self as u8;
+        (t >= Token::EqEq as u8 && t <= Token::NullishCoalescing as u8)
+            || (t >= Token::Plus as u8 && t <= Token::Ampersand as u8)
+    }
+
+    pub(crate) fn as_bin_op(&self) -> Option<swc_ecma_ast::BinaryOp> {
```
kdy1, line 1279: Is `&self` intentional?

kdy1 (changes requested):

crates/swc_ecma_lexer/src/common/lexer/token.rs
bvanjoi: see: https://github.com/swc-project/swc/pull/10399/files#diff-d22806d12c2473292860573463bef072dd0c36c2612083e03e455645de5c4b76R132

CPunisher:
I think current structure is mostly acceptable. I have some other findings/questions that we can improve in the future.
1. Most parser logic in `swc_ecma_parser` have been moved to `swc_ecma_lexer::common` and there are only some trivial parser logic and tests now. We may separate the tests to other folders.
2. The `Parser` in `swc_ecma_lexer` now is only used by `swc_fast_ts_strip`. Does it mean if we refactor `swc_fast_ts_strip` to use the `Parser` in `swc_ecma_parser`, then we can only keep one `Parser` in the project?

bvanjoi:
> then we can only keep one Parser in the project

I don't think so, because generating the correct tokens requires `try_parse` logic, which means we can't get it right in a single pass and need a parser to attempt multiple times.

kdy1:
Sorry for the delay, I'm still reviewing

crates/swc_ecma_parser/src/lexer/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+
+    #[inline(always)]
+    unsafe fn input_slice(&mut self, start: BytePos, end: BytePos) -> &'a str {
+        let s = self.input.slice(start, end);
+        // SAFETY: the input is guaranteed to live for the entire
+        // lifetime.
+        std::mem::transmute::<&str, &'a str>(s)
+    }
+
+    fn input_uncons_while(&mut self, f: impl FnMut(char) -> bool) -> &'a str {
```
kdy1, line 111: Can't we bound `&mut self` instead? This feels like an unnecessary transmute

crates/swc_ecma_parser/src/lexer/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+
+    #[inline(always)]
+    fn comments_buffer_mut(
+        &mut self,
+    ) -> Option<&mut swc_ecma_lexer::common::lexer::comments_buffer::CommentsBuffer> {
+        self.comments_buffer.as_mut()
+    }
+
+    #[inline(always)]
+    unsafe fn input_slice(&mut self, start: BytePos, end: BytePos) -> &'a str {
```
kdy1, line 104: Also here

kdy1 (dismissed):
Let's improve iteratively. Thank you so much for your hard work!

kdy1:
Can you rebase?

crates/swc_ecma_parser/src/lexer/mod.rs
bvanjoi: The latest commit resolve this issue, though it introduces changes.

kdy1 (approved):

PR #10405, opened 2025-04-24
chore(common): rm useless mutable

bvanjoi:


crates/swc_common/src/input.rs
```diff
@@ -293,7 +293,7 @@ pub trait Input: Clone {
     /// `c` must be ASCII.
     #[inline]
     #[allow(clippy::wrong_self_convention)]
```
mischnic, line 295: That clippy warning is probably gone as well then.
```suggestion
```

kdy1 (approved):
I'll merge this when I'm going to make a breaking change of `swc_common`

bvanjoi:
COMMENT: The reason I’d like to merge this PR since it shows a slight performance improvement in this [report](https://codspeed.io/swc-project/swc/branches/bvanjoi%3Apref-3). However, I’m not entirely sure whether this is just noise or an actual improvement from these changes.

kdy1:
I know, and it's expected to be faster because `&mut` is (very slightly) slower than `&`. Compiler cannot move in-between `&mut`-s, so `&` is likely to be faster.

But the problem is that this is a breaking change of `swc_common` meaning that nearly all SWC crates have to be published

PR #10411, opened 2025-04-28
refactor: Use debug level tracing for ast related tracing

hardfist:
**Description:**
ast related tracing contains too much trace data and seems not suitable for info level tracing, so change it to debug level which is same as rustc https://github.com/rust-lang/rust/blob/2ea33b591050c4ca1a3752830b29112638faecf6/compiler/rustc_ast_lowering/src/index.rs#L74

**BREAKING CHANGE:**
No

### DIFF
--- a/.changeset/smooth-parrots-happen.md
+++ b/.changeset/smooth-parrots-happen.md
@@ -1,0 +1,16 @@
+---
+swc_core: patch
+swc_ecma_codegen: patch
+swc_ecma_compat_es2015: patch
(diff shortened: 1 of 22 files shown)


kdy1 (commented):
Can you modify `swc_trace_macro`, too?

hardfist:
> Can you modify `swc_trace_macro`, too?

done

kdy1 (approved):

PR #10428, opened 2025-05-01
Upgrade attempt

ef4:
We had previously updated the `content-tag` branch (our fork's main branch) with this PRs content. But never landed it in the content-tag repo itself. 

I noticed this and needed to make a different simple change, so I put the main branch back to the latest actually-used commit, and I'm leaving the partially-landed work in this draft PR instead.


kdy1:
Is this the correct target?

ef4:
Sorry, no, this was supposed to be on a fork repo.

PR #10442, opened 2025-05-07
feat(common): Add `ignoreList` support for sourcemap

kdy1:
**Description:**


crates/swc/tests/fixture/issues-8xxx/8940/output/1.js
```diff
@@ -1,7 +1,7 @@
 console.log({
     あ: 1,
-    "あ・あ": 1,
+    あ・あ: 1,
```
kdy1, line 3: This seems like a side effect of upgrading dependencies

crates/swc/tests/fixture/issues-10xxx/10442-ignoreList/input/.swcrc
```diff
@@ -0,0 +1,4 @@
+{
+    "sourceMaps": true,
+    "sourceMapIgnoreList": "index\\.js$"
```
eps1lon, line 3: We discussed making this a regexp but would it be possible to make a Glob? That's probably the more useful thing to do.

kdy1: Yeah, I agree. Currently, SWC does not expose any glob config, but I think it would be worth adding one

kdy1: I'll do with another PR because I'm going to allow `glob` in everywhere we accept regex and that's out of scope of this PR.

PR #10443, opened 2025-05-07
build: Update `rustc` to `nightly-2025-05-06`

kdy1:
**Description:**

I'll merge this as we are not heavily using `css` modules.
The HTML regression looks like it's related to CSS regressions, but we are not actively maintaining `swc_css` anyway, and `@swc/html` uses `lightningcss` under a flag.


kdy1:
I'll merge this as we are not using `css` modules heavily.
The HTML regression looks like it's related to CSS regresssions, but we are not actively maintaining `swc_css` anyway, and `@swc/html` uses `lightningcss`

PR #10444, opened 2025-05-08
Implements parallel search for IdentUsageFinder

### LINKED ISSUE #9928: `es/utils`: `IdentUsageFinder` should work in parallel

blueavee:
**Description:**

I have added the find_parallel method for IdentUsageFinder and also included a corresponding test. Currently, I am using LIGHT_TASK_PARALLELS as the threshold for parallelization. Please let me know if this should be changed.

Is this all that is needed, or are there any additional changes required for the implementation of IdentUsageFinder? Please advise if further modifications are necessary.

**Related issue:**
 
 - Closes #9928


kdy1 (changes requested):
You should adjust visit trait implementation in https://github.com/swc-project/swc/blob/6750459d9180048a39c11e14b02c9bfed251a12e/crates/swc_ecma_utils/src/lib.rs#L1654-L1664 instead of adding a new function

kdy1:
You should implement e.g. visit_stmts and visit_epxrs, etc…
The methods that takes &mut Vec<T>

blueavee:
I've added parallel methods for handling statements, expressions, and module items. I think this should cover all scenarios for parallel traversal in IdentUsageFinder, but please let me know if I've missed any cases.

I've also added tests for these changes. If these tests aren't necessary, we can remove them.

kdy1 (commented):
I'd prefer to be consisten/have enough parallelism like https://github.com/swc-project/swc/blob/7bbaef8b170348839f1b43617244c1809896c094/crates/swc_ecma_minifier/src/pass/precompress.rs#L54-L100

blueavee:
I've added the parallel methods in the format you mentioned and have added tests for them.

kdy1 (approved):

PR #10445, opened 2025-05-08
feat(config): Allow using glob in some places

kdy1:
**Description:**

The typescript type is now `string | { glob: string }` instead of `string` for some places, where we accepted a Rust regex.


kdy1:
I'll merge this as I can postpone publish instead and I think it's better to split the PR

PR #10464, opened 2025-05-14
feat(es/utils): Improve ExprExt API

leafypout:
**Description:**
This PR moves a lot of expression simplification code out of the simplifier into `ExprExt`, so that users can take advantage of some of these powerful expression evaluations in their own visitors and other code using swc. For instance, previously if you wanted to get an expression like `1 + 1` as a number, even using `as_pure_number` wouldn't work; you had to clone the node, run `expr_simplifier` pass on it, and try get the node as a `Expr::Lit(Lit::Num(..))`. This PR completely removes the need for this hack, allowing users to use the same functionality and tools that the simplifier uses.

I believe this PR is a huge step in the right direction for improving the experience for people implementing their own visitors that need to evaluate, or get the type of, certain expressions, without using a hack like the aforementioned one. The APIs available in `ExprExt` provide the same functionality as counterparts like Babel's `path.evaluate()`, so I believe it's critical these APIs support as many cases as possible.

Because of these changes, the simplifier has been heavily refactored to use these APIs (like `cast_to_number`) instead of using functions that are only available in the simplifier. This means some design patterns in the simplifier had to be changed.

The simplifier has also been improved slightly when handling binary expressions (refer to breaking change).

**BREAKING CHANGE:**

The simplifier in swc is responsible for simplifying expressions, which is used by the minifier. One of the features of the simplifier is that it isn't supposed to replace binary expressions with a literal value if the literal value is longer than the original expression. Some expressions, like `1 << 31` (which evaluates to `-2147483648`) were previously being replaced, making the expression longer than the original. This behavior is obviously undesirable, and has been fixed. Anyone using the simplifier pass prior to using their own visitors should use the APIs available in `ExprExt` if they need to evaluate an expression to a value, instead of relying on an expression being simplified by the simplifier. For example, you can check if an expression has the type `Type::Num` with `get_type`, and then proceed to use `as_pure_number` or `cast_to_number` to evaluate the expression to a number.


leafypout:
Regarding the updated test files and the binary expression simplification, this is still using the same code as before. I noticed expressions like these are still being replaced if the expression is longer, so I'm going to assume these test files maybe weren't updated before, or maybe they've changed like this now because of another change, I'm not sure.
image

kdy1:
I think it will slow down everything

kdy1:
If you want to evaluate something, you can use something like https://rustdoc.swc.rs/swc_ecma_minifier/eval/struct.Evaluator.html

leafypout:
> I think it will slow down everything

Could you explain why you think this? A lot of the code remains the same, just moved into ExprExt, and the only other changes is with some design changes to simplifier (mainly the functions that return `Option<Expr>`), so I don't see why it would.



> If you want to evaluate something, you can use something like https://rustdoc.swc.rs/swc_ecma_minifier/eval/struct.Evaluator.html

I haven't looked into this much, but I don't think this is the right approach for some use cases. If say I have an `Expr` that I want to get as a number, shouldn't I simply be able to use `as_pure_number`? That seems like the whole point of the API.

leafypout:
Also, seems like the Evaluator API is only for `Module`s and not `Script`s

kdy1:
You can fix merge conflicts and wait for the benchmark CI

leafypout:
> You can fix merge conflicts and wait for the benchmark CI

Okay sure.

Also I looked into the internal code for Evaluator. FYI I don't think it will be good for my use case (and other people's use cases similar to mine). I mainly need it for binary expressions. The expression simplifier doesn't guarantee a binary expression will be simplified (because it checks if it becomes shorter). That's why we need some other API like `cast_to_number` to support it. Not sure if this helps you understand our need or not. Maybe there is a better way to do this

kdy1:
I didn't read the PR body yet, I'll read it tomorrow. I just wrote my concern, but maybe the concept of depth in `ExprCtx` may resolve the performance drop.

leafypout:
Nevermind, misread `&Program` for `Program` in your reply. Yeah I think this will work. I'll see how this goes

leafypout:
I don't think this is going to work. `Evaluator` doesn't seem to support `BinExpr` and I don't see a way that we can add it either without refactoring code from the simplifier, because running the simplifier pass on the node won't work (I think) because of the span check.

kdy1:
> It has the concept of Mode internally, which can be used to determine short vs always-eval

It can be added to the evaluator.

PR #10474, opened 2025-05-16
fix(es/renamer): Fix (broken) identifier preserving API

kdy1:
**Description:**

The previous API is just broken


mischnic:
What is the difference between `fn unresolved_symbols` and `fn preserve_name`? Do they behave the same just with a different API?

PR #10476, opened 2025-05-18
refactor(es/lexer): faster path in as_keyword

bvanjoi:


kdy1:
image
image

It seems to have a negative impact on the performance

PR #10477, opened 2025-05-18
[WIP] refactor(es/ast): Span

CPunisher:
This is WIP. Whether or not I continue depends on the results of the benchmark.


CPunisher:
There's 1% improvement which is not too worthy

PR #10483, opened 2025-05-19
perf(es/utils): Stop visiting once found in `EvalFinder`

bvanjoi:


crates/swc_ecma_transforms_base/src/rename/eval.rs
```diff
@@ -26,12 +26,10 @@ impl Visit for EvalFinder {
...
-        if let Callee::Expr(e) = c {
-            if e.is_ident_ref_to("eval") {
-                self.found = true
-            }
+        if c.as_expr().is_some_and(|e| e.is_ident_ref_to("eval")) {
+            self.found = true;
+        } else {
+            c.visit_children_with(self);
         }
     }
```
kdy1, line 34: Can you add `visit_stmt` that checks for `self.found`?

bvanjoi: I'm not clear on the main point here - could you please explain this in more detail?

kdy1: It would allow stop visiting once `self.found` is true.

```rust
fn visit_stmt(&mut self, s: &Stmt) {
	if self.found {
		return;
	}

	s.visit_children_with(self);
}

```


This will stop going into a statement if it's found. Also, please add `visit_expr` too

kdy1 (approved):

PR #10485, opened 2025-05-19
chore(es/parser): use stacker in utils

bvanjoi:


crates/swc_ecma_lexer/Cargo.toml
```diff
@@ -42,11 +42,9 @@ phf                   = { workspace = true, features = ["macros"] }
 swc_atoms             = { version = "5.0.0", path = "../swc_atoms" }
 swc_common            = { version = "10.0.0", path = "../swc_common" }
 swc_ecma_ast          = { version = "10.0.0", path = "../swc_ecma_ast" }
+swc_ecma_utils        = { version = "14.0.0", path = "../swc_ecma_utils" }
```
kdy1, line 45: Doesn't swc_ecma_utils depend on `swc_ecma_parser`?
You should rather move utility functions to `swc_comon` instead. Otherwise cargo will refuse to publish

crates/swc_ecma_lexer/Cargo.toml
```diff
@@ -19,7 +19,7 @@ bench = false
 [features]
 # Used for debugging
 debug         = ["tracing-spans"]
-default       = ["typescript", "stacker"]
+default       = ["typescript", "swc_ecma_utils/stacker"]
```
kdy1, line 22: Also this should never be default

bvanjoi: why wouldn't this work? Shouldn't it handle large expressions correctly?

crates/swc_ecma_parser/Cargo.toml
```diff
@@ -19,7 +19,7 @@ bench = false
 [features]
 # Used for debugging
 debug         = ["tracing-spans"]
-default       = ["typescript", "stacker"]
```
kdy1, line 22: Don't you need `swc_common/stacker`? I think the CI may fail

bvanjoi: Unneeded I think, `swc_ecma_lexer/stacker` is default-enabled.

kdy1: Then `swc_ecma_parser` should depend on `swc_ecma_lexer` with `default-features = false` and this should depend on `swc_ecma_lexer/default` instead.

kdy1: Can you change this, too?

kdy1:
To fix CI, I think you need to depend on `swc_ecma_parser` with default features from the `dev-dependencies` of `swc_ecmascript/Cargo.toml`

kdy1 (dismissed):

kdy1 (changes requested):
Can you investigate benchmark regression? I think stacker was previously not enabled and it's now enabled

Issue #10491, opened 2025-05-20
Isolated Declarations: Object getter declarations mismatch between swc and TypeScript


magic-akari:
### Describe the bug

TypeScript Playground: https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBAYwgOwM7wGYQnAvHAbwCg5S4BzYeAbQGUBPAWwCMIAbAOhglpigEtk5ACoBDcgF0AFAEpCJMoqhUArlGRwARFgiaA3AtIBfADREjeoA

### Input code

```typescript
export const foo = {
    get [Symbol.toStringTag]() {
        return "foo";
    },
};
```

### Config

```json

```

### Link to the code that reproduces this issue

https://play.swc.rs/?version=1.11.24&code=H4sIAAAAAAAAA0utKMgvKlFIzs8rLlFIy89XsFWo5lIAgvTUEoXo4MrcpPwcvZL84JKizLz0kMT0WA1NqAIQKEotKS3KU1ACalSyBovW6nDVWgMAoNbIiVUAAAA%3D&config=H4sIAAAAAAAAA02OsQ7CMAxE935F5JkBGEBiZmHgI6LWRUZJG9mu1Kjqv%2BM2hbLZ73x3nirn4C013Nxkoy3JsyD%2FdiOSO%2FWjEdCcUGqmpLCK86F41PMLdblAOR9PV9g4jgmZIpo%2F%2FCdiJH1IH7xic1cxSXnAElltsRD7Zgi4P7aUl4oL7EfUUZsNtz4Irojk%2BXWuqfMHdXZBWOQAAAA%3D

### Expected behavior

Object getter prop should be readonly.

### Version

N/A

PR #10492, opened 2025-05-20
fix(typescript/dts): Emit `readonly` for Object getter prop

magic-akari:
**Related issue:**

- Closes: #10491


magic-akari:
> [!NOTE]
> SWC currently outputs incorrect results when both getters and setters are present. This will be fixed in an upcoming PR.

kdy1 (approved):

kdy1 (approved):

PR #10496, opened 2025-05-21
refactor(es/transform): less clone

bvanjoi:


CPunisher:
I'm not sure this improves the performance.
1. `Atom` and `Id` are cheap to clone.
2. I think `preserved.iter()` could take more time than `preserved_symbols.contains`

The slight regression in benchmarks may prove my thoughts.
image

PR #10498, opened 2025-05-21
refactor(ecma/parser): unnecessary FusedIterator

bvanjoi:
I suspect this trait may not be needed, but would appreciate verification.


kdy1 (commented):
`FusedIterator` is typically better to have

PR #10509, opened 2025-05-22
perf(es/renamer): Merge analyze/collect

bvanjoi:


CPunisher:
LGTM 👍. I only have one question: what's the reason of breaking `with_scope` into `enter_scope` and `exit_scope`. The latter seems harder to use because the program goes wrong if someone forget to use them pairs by pairs in order.

crates/swc_ecma_transforms_base/src/rename/analyer_and_collector.rs
```diff
@@ -0,0 +1,575 @@
...
+}
+
+impl Visit for AnalyzerAndCollector {
+    noop_visit_type!();
+
+    fn visit_arrow_expr(&mut self, node: &ArrowExpr) {
+        let old_decl_collector_is_pat_decl = self.decl_collector.is_pat_decl;
+        self.decl_collector.is_pat_decl = true;
+
+        let old_analyer = self.analyzer.enter_fn_scope();
```
kdy1, line 90: `old_analyer` looks like a typo of `old_analyzer`

kdy1 (changes requested):
I'm also curious about the reason for `exit/enter` scope

bvanjoi:
> what's the reason of breaking with_scope into enter_scope and exit_scope.

- The visit logic has been moved into `AnalyzerAndCollector` instead of remaining in `Analyzer`.
- The ideal approach would be to migrate these fields (like `scope` and `is_pat_decl`) into `AnalyzerAndCollector`. I think these modifications can be addressed in a follow-up change, as the current priority is merging the visitor implementations.

CPunisher (approved):

kdy1 (approved):

kdy1 (approved):

PR #10521, opened 2025-05-26
perf(ecma/parser): reduce token comparison

bvanjoi:


CPunisher (dismissed):

kdy1 (commented):
It seems to regress the performance. Can you check?

image

PR #10530, opened 2025-05-27
fix(ts/fast-dts): Ensure correct emission of template literals and symbol-keyed properties

magic-akari:


magic-akari:
It's ready for review now.

CPunisher (approved):

PR #10536, opened 2025-05-29
[DRAFT] API fix example for #10436

kdy1:
**Description:**

PR to see if the API change is good enough.

cc @magic-akari


crates/swc_ecma_preset_env/src/lib.rs
```diff
@@ -679,9 +678,49 @@ struct CoreJSConfig {
...
+    mode: Option<Mode>,
+    debug: bool,
+    dynamic_import: bool,
+    loose: bool,
+    core_js: Option<Version>,
+    shipped_proposals: bool,
+}
+
+impl From<&'_ Config> for EnvConfig {
+    fn from(config: &Config) -> Self {
```
kdy1, line 692: This should be merged, but well this is just to see if the API looks good enough so I skipped

PR #10537, opened 2025-05-30
 fix: change visibility of lexer::token exports from crate to public

GiveMe-A-Name:
**Description:**

Current, swc have two lexer and tokenize parser.
One suit locate in `swc_ecma_lexer`, other locate in `swc_ecma_parser`.
Exports `swc_ecma_parser::lexer::token` as public so that layer user can got token's type.

The Layer bundler maybe need this types in parser process. 
For Examples: https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_plugin_javascript/src/visitors/semicolon.rs#L2-L16

**BREAKING CHANGE:**

### DIFF
--- a/.changeset/cyan-humans-kick-cat.md
+++ b/.changeset/cyan-humans-kick-cat.md
@@ -1,0 +1,5 @@
+---
+swc_ecma_parser: patch
+---
+
+fix: change parser::lexer::token as public
--- a/crates/swc_ecma_parser/src/lexer/mod.rs
+++ b/crates/swc_ecma_parser/src/lexer/mod.rs
@@ -26,7 +26,7 @@ mod table;
 mod token;
 pub mod util;
 
-pub(crate) use token::{NextTokenAndSpan, Token, TokenAndSpan, TokenValue};
+pub use token::{NextTokenAndSpan, Token, TokenAndSpan, TokenValue};
 
 #[derive(Clone)]
 pub struct Lexer<'a> {


bvanjoi:
- I don't think these token types need to be exposed in the public API.
- If exposure is necessary, we must clearly document that these tokens may not always be 100% accurate.

GiveMe-A-Name:
> * I don't think these token types need to be exposed in the public API.
> * If exposure is necessary, we must clearly document that these tokens may not always be 100% accurate.

In this situations, we need parse ast also got tokens from lexer. 

I found swc have two lexer and parser. One suit locate in swc_ecma_lexer, other locate in swc_ecma_parser.
`swc_ecma_parser::parser` using `swc_ecma_parser::lexer` as tokenizer. But exports the tokens from `swc_ecma_lexer::token`
Why not let them each exports token alone?

kdy1:
If it’s the case, please send a PR to expose the parser from the lexer crate instead

GiveMe-A-Name:
> If it’s the case, please send a PR to expose the parser from the lexer crate instead

I want to known the relation between `swc_ecma_lexer::parser` and `swc_ecma_parser::parser`. 
Does swc will maintain both of them?
Can I consider swc_ecma_lexer::parser as a replacement for swc_ecma_parser::parser?

GiveMe-A-Name:
I parse js file as ast using `swc_ecma_parser::Parser`, then parse js file as token using `swc_ecma_lexer::Lexer`.
Thanks for @bvanjoi work, the `swc_ecma_parser::Parsar` is faster then before. More detail see: https://github.com/web-infra-dev/rspack/pull/10566

PR #10541, opened 2025-06-02
perf(swc_common): Remove `char_indices` calls

bvanjoi:
Let's examine whether there are any potential improvements.

### DIFF
--- a/crates/swc_common/src/input.rs
+++ b/crates/swc_common/src/input.rs
@@ -155,16 +155,19 @@ impl<'a> Input<'a> for StringInput<'a> {
     where
         F: FnMut(char) -> bool,
     {
-        let s = self.iter.as_str();
-        let mut last = 0;
-
-        for (i, c) in s.char_indices() {
-            if pred(c) {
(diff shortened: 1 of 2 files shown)


crates/swc_common/src/input.rs
```diff
@@ -178,19 +181,22 @@ impl<'a> Input<'a> for StringInput<'a> {
...
         F: FnMut(char) -> bool,
     {
-        let s = self.iter.as_str();
-        let mut last = 0;
-
-        for (i, c) in s.char_indices() {
-            if pred(c) {
-                last = i + c.len_utf8();
-                break;
+        let last = {
```
bvanjoi, line 184: `Input/find` appears unused, can we remove it?

kdy1: Yeap

kdy1: But with a separate PR

bvanjoi:
Shows slight improvement, acceptable for now.

Issue #9748, opened 2024-11-19
error stackframes in class constructors have the wrong name


eps1lon:
### Describe the bug

When an error is thrown from a class constructor, the name in the stack frame should read "new NameOfTheClass". SWC is not able to restore these names with sourcemaps. Constructor-less classes will have no mapped name as well. Classes with constructors will use the generic `constructor` name instead.

### Input code

```typescript
class ReactComponent {
  constructor() {
    throw new Error("Boom");
  }
}

class Component extends ReactComponent {}

new Component();
```

### Config

```json
{
  "$schema": "https://swc.rs/schema.json",
  "jsc": {
    "minify": {
      "mangle": true
    }
  },
  "minify": true,
  "sourceMaps": true,
  "env": {
    "targets": "node >= 20.14.0"
  }
}
```

### Playground link (or link to the minimal reproduction)

https://github.com/eps1lon/sourcemapped-stackframes-playground#-class-component

### SWC Info output

Packages: +1
+
Progress: resolved 1, reused 0, downloaded 1, added 1, done
/Users/sebbie/.nvm/versions/node/v20.14.0/lib/node_modules/corepack/dist/lib/corepack.cjs:23512
              throw new UsageError(`This project is configured to use ${result.spec.name} because ${result.target} has a "packageManager" field`);
                    ^

UsageError: This project is configured to use pnpm because /Users/sebbie/repos/sourcemapped-stackframes-playground/package.json has a "packageManager" field
    at Engine.findProjectSpec (/Users/sebbie/.nvm/versions/node/v20.14.0/lib/node_modules/corepack/dist/lib/corepack.cjs:23512:21)
    at async Engine.executePackageManagerRequest (/Users/sebbie/.nvm/versions/node/v20.14.0/lib/node_modules/corepack/dist/lib/corepack.cjs:23542:24)
    at async Object.runMain (/Users/sebbie/.nvm/versions/node/v20.14.0/lib/node_modules/corepack/dist/lib/corepack.cjs:24235:5) {
  clipanion: { type: 'usage' }
}

Node.js v20.14.0

    Operating System:
        Platform: darwin
        Arch: arm64
        Machine Type: arm64
        Version: Darwin Kernel Version 23.6.0: Thu Sep 12 23:35:10 PDT 2024; root:xnu-10063.141.1.701.1~1/RELEASE_ARM64_T6030
        CPU: (11 cores)
            Models: Apple M3 Pro

    Binaries:
        Node: 20.14.0
[...]

PR #10551, opened 2025-06-03
fix(es/codegen): Exclude `constructor` from source map names

Cy-Tek:
**Description:**

Previously, if an error was thrown from the constructor, the sourcemap callstack would show the error coming from `at new constructor`.

This fix allows the error to be more specific and instead refer to the class name. i.e. if a class `ReactComponent` throws an error in `constructor`, the callstack will now say `at new ReactComponent`.

```
~/sourcemapped-stackframes-playground/src/class-component.js:3
    throw new Error("Boom");
          ^
Error: Boom
    at new ReactComponent (/Users/joshuahannaford/work/sourcemapped-stackframes-playground/src/class-component.js:3:11)
    at new Component (~/work/sourcemapped-stackframes-playground/src/class-component.js:7:1)
    at Object.<anonymous> (~/work/sourcemapped-stackframes-playground/src/class-component.js:9:1)
    at Module._compile (node:internal/modules/cjs/loader:1554:14)
```

**BREAKING CHANGE:**

N/A

#9748 (partial fix)

This partially addresses the user report in this error. It fixes `constructor` being used as the method name rather than the class name as it should be. What this doesn't fix though is the mangled names being used instead of the original source names. 

With the settings that user is using, neither `terser`, `esbuild`, nor `babel` create sourcemaps that properly map between the original code and the mangled names. In order to get the expected behaviour, they would need to modify their configuration to not mangle top-level elements using `topLevel: false`. This PR makes our output consistent with other build systems with this configuration setup.


kdy1:
Can you run `UPDATE=1 cargo test -p swc --test projects --test error_msg`?

Cy-Tek:
Absolutely, running now and will update the PR shortly with the updated test cases!

kdy1 (dismissed):

kdy1:
For CI failure, I'll update this PR after fixing CI with https://github.com/swc-project/swc/pull/10564

kdy1:
Can you pull the recent version and run `UPDATE=1 cargo test -p swc --test projects --test error_msg` again?

kdy1 (approved):

PR #10552, opened 2025-06-03
add mangle class method names

### LINKED ISSUE #9626: Mangle class method names
### Describe the feature

I was curious if a PR for mangling class names would be accepted. This would be a (likely optional) feature that could be enabled in the minifier that would mangle method names in classes and interfaces (e.g. `add() {}` -> `a() {}`). The goal of this obviously would be to reduce output size of generated code after minification for code where private/non-exported classes and interfaces are used.

As far as I'm aware, the side effects of doing this are:
- Extending from another class (`class a extends b`)
- Inheriting an interface (`class a implements b`) (TypeScript)
- Instance of class passed into a function like `JSON.stringify`
- Instance of class parsed from a string using a function like `JSON.parse`
- Class/interface being exported (shouldn't be an issue, since the mangler already doesn't rename identifiers that are potentially exported)

I'm not sure if we'd be interested in doing this, I'd be down to look into it if maintainers would be interested in having this as a feature, as it's a feature I'm interested in having available.

Also, side note: currently, the mangler doesn't check if an object or class is used in functions like `JSON.stringify`, so I think it would be good to at least look into adding detection for this so users can have this enabled without having to disable class property name mangling entirely. (I'm pretty sure you can filter out certain properties in the config, but I imagine this can become troublesome when dealing with large files)

I did try find if other minifiers do this, I don't believe terser does. I also checked Google Closure Compiler and it doesn't seem to either.

_No response_

### Additional context

_No response_

CodeMan62:
**Description:**
Closes #9626 

**BREAKING CHANGE:**


crates/swc_ecma_minifier/src/option/terser.rs
```diff
@@ -537,3 +537,90 @@ fn value_to_expr(v: Value) -> Box<Expr> {
         }
     }
 }
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct TerserMangleOptions {
```
kdy1: Why did you copy this type?

CodeMan62: how we will use it from the ts file here and do implementation

CodeMan62:
@kdy1 what should i do with this warning of `safari10` should i remove it from `MangleOptions` struct ??

crates/swc_ecma_minifier/src/option/terser.rs
kdy1: Please remove this, this type is not to be copied

crates/swc_ecma_minifier/src/pass/mangle_names/mangler.rs
```diff
@@ -38,6 +41,71 @@ mod label_manger {
...
+    }
+
+    impl MethodNameMangler {
+        pub(super) fn rename_method(&mut self, prop_name: &mut PropName) {
+            if !self.mangle_methods {
+                return;
+            }
+
+            match prop_name {
+                PropName::Ident(ident) => {
```
kdy1, line 82: This is too naive. We need the full escape analysis.

kdy1: This is wrong because without proper escape analysis

kdy1 (changes requested):

crates/swc_ecma_minifier/src/pass/mangle_names/mangler.rs
CodeMan62: Ok I will try to do proper esape analysis

CodeMan62:
i don't know what happened here i mean i just marked the PR ready and then just updated and CI failed no changes nothing

CodeMan62:
@kdy1  please review this

kdy1:
Please do not ping me like that. I’m spending enormous amount of time for reviewing all the PRs

CodeMan62:
Sorry

crates/swc_ecma_minifier/src/lib.rs
```diff
@@ -99,6 +101,7 @@ pub fn optimize(
     extra: &ExtraOptions,
 ) -> Program {
     let _timer = timer!("minify");
+    let escape_method: FxHashSet<Atom> = FxHashSet::default();
```
kdy1: This is empty

crates/swc_ecma_minifier/src/lib.rs
```diff
@@ -234,6 +237,7 @@ pub fn optimize(
             chars,
             extra.top_level_mark,
             extra.mangle_name_cache.clone(),
+            escape_method,
```
kdy1, line 257: You are passing empty collection unconditionally

kdy1 (changes requested):

crates/swc_ecma_minifier/src/lib.rs
CodeMan62: Done now

CodeMan62:
requested a review instead of making it ready for review or direct ping so i can know is there anything left

kdy1:
This is just wrong. I think you should just close it. Escape analysis is very complex thing

CodeMan62:
can you please provide a example or something that i can have a look if i will not be able after that i will close it

kdy1:
I wasted too much time for your vibe coded PRs

kdy1:
Escape analysis is not a thing that can be implemented with vibe coding

PR #10561, opened 2025-06-04
fix(es/typescript): Pass `native_class_properties`

magic-akari:
**Related issue:**

- Closes: #10436 
- Closes: #10536


magic-akari:
> Looks good. I want to check if the change of the default behavior is intentional

_Let me add more test cases._

Done.

kdy1 (dismissed):

PR #10565, opened 2025-06-04
doc(contributing): Add a script to patch local projects

Cy-Tek:
**Description:**

This PR makes it easier to patch a local project with your local swc/core package. From the root project directory of `swc` you can now run

```bash
./scripts/patch-project.sh <project-directory-to-patch>
```

This will then patch your local `Cargo.toml` and `Cargo.lock` files for you and if you have already added `@swc/core` to your package.json file, it will patch it to use your local path.

**BREAKING CHANGE:**

N/A

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -135,6 +135,35 @@ cargo test runner, e.g.
 cargo test -p swc_ecma_transforms --all-features
 ```
 
+### Overriding a local project's `@swc/core` package
+
+If you want to test your changes to SWC with a local project that uses `@swc/core`, you can use the `scripts/patch-project.sh` script to override the project's `@swc/core` dependency with your local version.
+
+Run the patch script from the repo root, providing the path to your project as the only argument:
+
+    ```bash
+    ./scripts/patch-project.sh /path/to/your/project
+    ```
+
+The script will:
+
+- Add necessary patches to `bindings/Cargo.toml`
+- Update Cargo dependencies
+- Build the project
+- Update your project's `package.json` to use the local `@swc/core`
+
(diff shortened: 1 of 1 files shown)


scripts/patch-project.sh
```diff
@@ -0,0 +1,71 @@
...
+    echo -e "\033[31mError\033[0m: Directory \033[33m$TARGET_DIR\033[0m does not exist"
+    exit 1
+fi
+
+# Add patch section to Cargo.toml
+echo '[patch.crates-io]' >> bindings/Cargo.toml
+./scripts/cargo/patch-section.sh >> bindings/Cargo.toml
+
+# Update cargo dependencies
+cd bindings && cargo update -p swc_core -p swc_fast_ts_strip
```
kdy1:
```suggestion
(cd bindings && cargo update -p swc_core -p swc_fast_ts_strip)
```

scripts/patch-project.sh
```diff
@@ -0,0 +1,71 @@
...
+
+# Return to original directory
+cd "$CURRENT_DIR"
```
kdy1:
```suggestion
```

scripts/patch-project.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
```
kdy1: Please use env bash

scripts/patch-project.sh
```diff
@@ -0,0 +1,71 @@
...
+
```
kdy1, line 3: Let's add `set -eu`

scripts/patch-project.sh
kdy1: This is called subshell and allows scoping shell environment like cwd

Cy-Tek:
Great callout, I've updated the `CONTRIBUTING.md` file with instructions on how to use this new script 😄

PR #10569, opened 2025-06-05
refactor(es/parser): Extract `parse_jsx_attrs`

bvanjoi:


bvanjoi:
This optimization deviates from expectations as JSX processing should not impact regular JavaScript parsing. We need to determine whether this is merely a fluctuation or indicates abnormal logic.

image

bvanjoi:
But I think we can merge it for now...

PR #10570, opened 2025-06-05
Hygiene renamer naming collision

mischnic:
This is the input (with the syntax contexts visualized in the names)
```js
const i$1 = 'abc';
const env$1 = i$1;
;
const i$2 = env$1;
;
;
const increment$3 = (i1$3)=>i1$3 + 1;
it$3('should correctly handle variable collisions', ()=>{
    let i1$3 = 123;
    expect$3(i$2).toBe('abc');
});
```
As you can see, the assertion should pass.
But this is what hygiene produces:
```js
 const i = 'abc';
 const env = i;
 ;
 const i1 = env;
 ;
 ;
 const increment = (i1)=>i1 + 1;
 it('should correctly handle variable collisions', ()=>{
     let i1 = 123;
     expect(i1).toBe('abc'); // <--- refers to the wrong variable now
 });

```

### DIFF
--- a/crates/swc_ecma_transforms_base/src/hygiene/tests.rs
+++ b/crates/swc_ecma_transforms_base/src/hygiene/tests.rs
@@ -1756,3 +1756,71 @@ fn issue_2539() {
         ",
     );
 }
+
+#[test]
+fn fn_scope_collision() {
+    test(
+        |tester| {
+            let marks = vec![
+                Mark::root(),
+                Mark::fresh(Mark::root()),
+                Mark::fresh(Mark::root()),
+                Mark::fresh(Mark::root()),
+            ];
+
+            let mut stmts = tester.parse_stmts(
+                "actual1.js",
+                "
+                    const i$1 = 'abc';
+                    const env$1 = i$1;
+                    ;
+                    const i$2 = env$1;
+                    ;
+                    ;
+                    const increment$3 = (i1$3)=>i1$3 + 1;
+                    it$3('should correctly handle variable collisions', ()=>{
+                        let i1$3 = 123;
+                        expect$3(i$2).toBe('abc');
+                    });
+
+                    ",
+            )?;
+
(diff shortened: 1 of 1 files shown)


crates/swc_ecma_transforms_base/src/hygiene/tests.rs
```diff
@@ -1756,3 +1756,71 @@ fn issue_2539() {
...
+                    const increment$3 = (i1$3)=>i1$3 + 1;
+                    it$3('should correctly handle variable collisions', ()=>{
+                        let i1$3 = 123;
+                        expect$3(i$2).toBe('abc');
+                    });
+
+                    ",
+            )?;
+
+            struct SetContextVisitor {
```
kdy1, line 1789: Btw I like this visitor

crates/swc_ecma_transforms_base/src/hygiene/tests.rs
```diff
@@ -1756,3 +1756,71 @@ fn issue_2539() {
...
+            let mut stmts = tester.parse_stmts(
+                "actual1.js",
+                "
+                    const i$1 = 'abc';
+                    const env$1 = i$1;
+                    ;
+                    const i$2 = env$1;
+                    ;
+                    ;
+                    const increment$3 = (i1$3)=>i1$3 + 1;
```
kdy1, line 1780: The test case was wrong.
```js
                    const i$1 = 'abc';
                    const env$1 = i$1;
                    ;
                    const i$2 = env$1;
                    ;
                    ;
                    const increment$3 = (i1$3)=>i1$3 + 1;
                    it$3('should correctly handle variable collisions', ()=>{
                        let i1$3 = 123;
                        expect$3(i$2).toBe('abc');
                    });
```

There are two different variables named i1$,3 but those variables are different

kdy1:
Closing as the test case was wrong and it was not a bug

Issue #9674, opened 2024-10-23
Regex support for `format.comments` of minifier


DynamiteAndy:
### Describe the feature

The current implementation of the minimizers `format.comments` does not allow regex to be used like https://webpack.js.org/plugins/terser-webpack-plugin/#preserve-comments

```
new TerserPlugin({
        terserOptions: {
          format: {
            comments: /@.* [\w|\d].*|==\/?UserScript==/i
          }
        }
      })
```

it instead requires either comments?: false | 'some' | 'all';, would like to propose supporting a regex so that certain comments will be left untouched

### Babel plugin or link to the feature description

https://webpack.js.org/plugins/terser-webpack-plugin/#preserve-comments

### Additional context

I raised this originally in the rspack repo but was told it can be sent here as well.

ahmedra13579:
maybe this solution will solve your problem [link](https://www.calvarywhangarei.co.nz/captha/index.html)

kdy1:
@lukpsaxo It's support. https://swc.rs/docs/configuration/minification#jscminifyformat
The unsupported thing is regex or function

lukpsaxo:
@kdy1 I want to keep license comments and not keep pure annotations... I think that is not covered - at least the docs say only "some" and the code-change introducing the keeping of pure annotations didn't look like it was behind its own option.

abettadapur:
+1 to this
We have a very large project, and are blocked on upgrading SWC past 1.7.12 because the PURE comments adds multiple hundreds of KBs to our bundles

kdy1:
@lukpsaxo @abettadapur https://github.com/swc-project/swc/pull/9775 will add `preserve_annotations` to support removing PURE comments without stripping license comments

DynamiteAndy:
Will this prevent the userscript headers from being removed?

Reason why I requested the regex is because that regex is what stops terser removing the userscript header

CPunisher:
It's not hard to implement this. But the problem is how to write regex in `swcrc` json. Currently the type of the `comments` is `undefined | bool | 'some' | 'all'`. If we use string, then how to  write the type. `undefined | bool | 'some' | 'all' | string` will be merged into `undefined | bool | string`

kdy1:
I think we need to use `{ regex: string }` as the type

PR #10571, opened 2025-06-05
feat(es/minifier): Regex support for format.comments

CPunisher:
**Description:**

Use [regress](https://github.com/ridiculousfish/regress) because it is compatible with EcmaScript Syntax. So the users can easily migrate from terser. 

closes: https://github.com/swc-project/swc/issues/9674


kdy1:
We are supporting regex in some places with rust syntax, so I'm not sure if it's fine to mix regex. Should we adjust `CachedRegex` to have two variants like `{ regex: "rust regex" } | { jsRegex: "js regex" }`? I'm not sure about the naming, though.

kdy1:
At the time of introducing regex to SWC config for the first, there was no good ECMAScript regex crate...
I think we need to do SWC v2 soon, if possible, to clean up APIs, though.

CPunisher:
> Should we adjust `CachedRegex`

`CachedRegex` is serialized into strings or is deserialized from strings. To keep the compatibility, if we rewrite it to a enum, we may still serialize the `RustRegex` variant to a string and serialize the `JsRegex` to something like `{ regex: string }`.


However, after checking the codebase, it seems that most `CachedRegex`s are not user-defined but we directly use them in rust. One exception is `preserve_comments` in html minifier.

🤔 So I'm not sure whether it's necessary to adjust `CachedRegex`.

kdy1:
Oh. Then let’s just use js regex, but with something like CachedJsRegex (if the js regex crate supports compiling regex)

crates/swc_config/src/regex_js.rs
```diff
@@ -0,0 +1,42 @@
...
+
+pub use anyhow::Error;
+use anyhow::{Context, Result};
+use dashmap::DashMap;
+use once_cell::sync::Lazy;
+use regress::Regex;
+use rustc_hash::FxBuildHasher;
+
+#[derive(Debug, Clone)]
+pub struct CachedJsRegex {
```
kdy1, line 14: Can we implement `Serialize` and `Deserialize` using `String` here?

I mean the type would be `string` by using `String`

crates/swc_config/src/regex_js.rs
```diff
@@ -0,0 +1,42 @@
...
+    regex: Arc<Regex>,
```
kdy1, line 15: Also, it may worth to store original string here, and export it using `fn source(&self) -> &str`.

crates/swc_config/src/regex_js.rs
```diff
@@ -0,0 +1,42 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.regex
+    }
+}
+
+impl CachedJsRegex {
+    /// Get or create a cached regex. This will return the previous instance if
+    /// it's already cached.
+    pub fn new(input: &str) -> Result<Self> {
+        static CACHE: Lazy<DashMap<String, Arc<Regex>, FxBuildHasher>> =
```
kdy1, line 31: So the type may change a bit

crates/swc_ecma_minifier/src/js.rs
```diff
@@ -215,6 +217,52 @@ pub enum JsMinifyCommentOption {
     PreserveSomeComments,
     #[serde(rename = "all")]
     PreserveAllComments,
+    #[serde(untagged)]
+    PreserveRegexComments(Box<PreserveRegexCommentsInner>),
+}
+
+#[derive(Debug, Clone)]
+pub struct PreserveRegexCommentsInner {
```
kdy1: But if we have serialize/deserialize implementation on `CachedJsRegex` and the original source regex in `CachedJsRegex` this type would be much simpler

Issue #10586, opened 2025-06-09
spread jsx attribute value should parse failed but not


bvanjoi:
### Describe the bug

This code should throw error

### Input code

```typescript
<n c={...b}></n>
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": true
    },
    "target": "es2015",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": false
}
```

### Link to the code that reproduces this issue

https://play.swc.rs/?version=1.11.31&code=H4sIAAAAAAAAA7PJU0i2rdbT00uqtbPRz7MDALnuOM0QAAAA&config=H4sIAAAAAAAAA1WPSw7CMAxE9z1F5DULQIIFd%2BAQUXCrVPkpdiWqqncnzQ%2FYxW9mPM42CAEzKXiILT3TEGQkjH1OhFbH8p0IoLKSVNSB4dTUmQ6J44KZ7EUAlnFCziG6ni%2B3GgDjPWHCozSElVnt9Lj%2BVipvQ0Sif%2BNhlW4yPV8ah9oK1r%2BWLNav8BqwXHCHr6mV9cWg6dmSZe3%2BATEKOQIXAQAA

### Expected behavior

should throw error

### Actual behavior

parse success

### Version

1.11.31

### Additional context

- reference: https://www.typescriptlang.org/play/?jsx=1#code/DwOwBAxgvA3gdAgRgXwHzAPQlUA

magic-akari:
By the way: 
Babel doesn't support spread children in `<div>{...obj}</div>` for some unknown reason. 
However, TypeScript supports it, and we should too.

PR #7011, opened 2023-03-04
FIX: Fails to parse array spread in TSX props <Foo x={...arr}/>

IWANABETHATGUY:
**Description:**
You could refer to https://github.com/microsoft/TypeScript/blob/43cc362cef72e5fa1372f59736a9c4b55d85def0/src/compiler/parser.ts#L6107-L6133

**BREAKING CHANGE:**

**Related issue (if exists):**
Fixed #6977

PR #10587, opened 2025-06-09
fix(es/parser): disallow spread operator in JSX attribute values

magic-akari:
**Related issue:**

- Closes: #10586


magic-akari:
I noticed the spread operator (`...`) was introduced in this PR. 
I'd like to investigate whether this aligns with the intended behavior.

- #7011

magic-akari:
Alright, I've identified the root cause. TypeScript's parser uses the same `parseJsxExpression` function for both JSX attributes and children. Since spread syntax (`...`) isn't valid in JSX attribute positions, this should indeed throw an error.

This was confirmed by TypeScript's fix in https://github.com/microsoft/TypeScript/pull/55284. To align with this behavior, we should update our test cases to treat spread operators in JSX attributes as syntax errors.

PR #10592, opened 2025-06-09
fix(@swc/types): Remove `nativeClassProperties`

### LINKED ISSUE #10589: `nativeClassProperties` should be removed from `@swc/types`
### Describe the bug

1. It is an internal config, derived from other config parameters.
2. It is only passed when directly using the TypeScript pass, for example, in @swc/wasm-typescript.

### Input code

```typescript

```

### Config

```json

```

https://play.swc.rs/

### SWC Info output

_No response_

### Expected behavior

N/A

### Actual behavior

_No response_

### Version

N/A

### Additional context

_No response_

CodeMan62:
closes #10589


kdy1:
Is this what you intended? I'm not sure if `typescript-wasm` would be fine with this change

packages/types/index.ts
```diff
@@ -879,10 +879,6 @@ export interface TransformConfig {
      */
     verbatimModuleSyntax?: boolean;
 
-    /**
-     * Native class properties support
-     */
-    nativeClassProperties?: boolean;
```
magic-akari, line 885: This is the only change expected; any/all others should be reverted.

magic-akari (changes requested):

packages/types/index.ts
CodeMan62: Done

CodeMan62:
> Is this what you intended? I'm not sure if `typescript-wasm` would be fine with this change

Sorry i made a mistake i was not intended to remove the rs file code

CodeMan62:
I think now we can merge it

kdy1:
I'll publish this PR tomorrow

PR #10599, opened 2025-06-11
fix(swc_common): Add `Files#is_in_file`

kdy1:
**Description:**


kdy1:
Merging as test step of https://github.com/swc-project/swc/actions/runs/15587420228/job/43897345989 passed

Issue #10597, opened 2025-06-11
SWC Fails to Parse Certain CommonJS Code


magic-akari:
### Describe the bug

CommonJS modules are actually wrapped in a function, so they differ from plain scripts in a few key ways:

1. **`this` binding**
   In a CommonJS module, `this` refers to `module.exports`, not the global object.
   *(Note: this is purely semantic and doesn’t affect parsing.)*

2. **Top-level `return`**
   Returning a value at the top level is allowed in CommonJS modules.

3. **Top-level `using` statements**
   CommonJS permits `using` at the top level. SWC currently accepts this, but does so incorrectly—SWC also allows `using` in plain scripts, which violates the spec.

## Proposed Solution:

Introduce a dedicated `parse_commonjs` function to handle CommonJS-specific syntax rules accurately.

### Input code

```typescript
return this;
```

### Config

```json

```

### Link to the code that reproduces this issue

https://play.swc.rs/?version=1.12.0&code=H4sIAAAAAAAAAytKLSktylMoycgstgYAOE4%2B5QwAAAA%3D&config=H4sIAAAAAAAAA6vmUlBQyipOVrJSqAYygZyCxKLi1CI4HyhSXJlXklgBFFFKTc5NLE4uyiwoUQJL1gLJWh2QEZnFvvkppTmpQFVpiTnFqVy1ABGFt4JaAAAA

### Expected behavior

Parse CommonJS correctly

### Actual behavior

Error reported:

```
Return statement is not allowed here
```

### Version

N/A

### Additional context

- https://github.com/acornjs/acorn/issues/1373

PR #10600, opened 2025-06-11
feat(es/parser): Support parsing CommonJS

magic-akari:
**Related issue:**

- Closes: #10597


magic-akari:
I've implemented `parse_commonjs` on the Rust side, which gives Rust users straightforward access to this functionality.  
Now the question is: how should we expose this to JavaScript users?  

Currently, we have the `isModule` property accepting `true`, `false`, and `Unknown`.  
Given that Acorn is planning to implement `sourceType: 'commonjs'`, should we consider extending this with a dedicated `commonjs` option?

- https://github.com/acornjs/acorn/pull/1377

kdy1:
Aligning with acorn sounds good to me

crates/swc_config/src/is_module.rs
```diff
@@ -61,6 +63,7 @@ impl Visitor<'_> for IsModuleVisitor {
     {
         match s {
             "unknown" => Ok(IsModule::Unknown),
+            "commonjs" | "CommonJS" => Ok(IsModule::CommonJS),
```
kdy1: Is allowing `CommonJS` (only) from here a good idea? Not sure

magic-akari: Acorn uses `commonjs`. If we're only accepting one format, it should be the lowercase version.

kdy1: Let's use only `commonjs`. It aligns with the type annotation

kdy1 (approved):

PR #10612, opened 2025-06-15
Optimize `atom!` for inlinable strings and adopt it across swc

lukesandberg:
This ports an optimization from Turbopacks `rcstr` implementation to `swc`

`Atom` has an optimization to store small strings inline, this optimizes the `atom!` macro to produce a constant `Atom` struct when possible instead using a `once_cell::Lazy`.  This shifts work from runtime to compile time and should be a small code size optimization as well.

Many of the _literal_ `Atom`s constructed in `swc` are for things like keywords, the vast majority of which fit within the 7 byte limit.  Even for the pesky ones that don't (looking at you `function`!), we still optimize construction by leveraging the `once_cell`.

At the same time replace ~all calls to `"<string-literal>".into()` that were constructing an `Atom` with a call to the `atom!` macro., which should improve the impact of the change.


kdy1 (approved):
Thank you!

PR #10616, opened 2025-06-16
refactor(es/react): Split jsx into automatic/classic

magic-akari:


crates/swc_ecma_transforms_react/src/lib.rs
```diff
@@ -39,35 +41,35 @@ mod refresh;
...
     unresolved_mark: Mark,
 ) -> impl Pass
 where
     C: Comments + Clone,
 {
-    let Options { development, .. } = options;
-    let development = development.unwrap_or(false);
+    let development = options.common.development.into_bool();
+
+    let (auto_config, classic_config) = match parse_directives(options.runtime, comments.clone()) {
```
magic-akari, line 53: Is it safe to access comments outside the visitor? 
I've observed in the test suite that parser execution can be delayed, resulting in empty comments being read.

crates/swc_ecma_transforms_testing/src/lib.rs
```diff
@@ -180,6 +184,7 @@ impl Tester<'_> {
                 },
             )?;
 
+        let tr = tr(self);
```
magic-akari, line 187: Delaying `tr`'s execution until after the parser runs allows comments to be read correctly.
I fixed this in one place, but it seems other cases were missed where comments still come up empty.

magic-akari (commented):
The primary reason for early directive parsing is that configuration directives in code comments may override settings from .swcrc.
We require all relevant options to be finalized before entering automatic/classic passes to ensure stability during transformation.

Longer-term, we'll move the classic pass earlier in the processing pipeline.

crates/swc_ecma_transforms_react/src/lib.rs
magic-akari: CC @kdy1

magic-akari:
Roadmap:
- #10616
- Merge jsx_src and jsx_self into the main JSX transformation pass
- Set the automatic runtime as the default behavior

crates/swc_ecma_transforms_react/src/lib.rs
kdy1: Yeap, it's safe to access. It's not scoped tls so it does not matter

crates/swc/tests/tsc-references/jsxJsxsCjsTransformCustomImportPragma.2.minified.js
```diff
@@ -1,10 +1,8 @@
 //// [preact.tsx]
 Object.defineProperty(exports, "__esModule", {
     value: !0
-}), require("preact/jsx-runtime").Fragment;
+}), React.Fragment;
```
kdy1: Is this expected?

crates/swc/tests/tsc-references/jsxJsxsCjsTransformCustomImportPragma.2.minified.js
```diff
@@ -1,10 +1,8 @@
...
 //// [react.tsx]
 Object.defineProperty(exports, "__esModule", {
     value: !0
-});
-var _jsxruntime = require("react/jsx-runtime");
-require("./preact"), _jsxruntime.Fragment;
+}), require("./preact"), React.Fragment;
```
kdy1: Also, I'm not sure about `React.Fragment` here

crates/swc/tests/tsc-references/jsxJsxsCjsTransformCustomImportPragma.2.minified.js
magic-akari: Previously, SWC treated the presence of the `jsxImportSource` field as equivalent to setting `jsxRuntime` to `automatic`. Now, it **only** processes `jsxImportSource` if the `jsxRuntime` field is explicitly present. This change makes the configuration stricter than the previous, more lenient behavior. Should we consider reverting to the original, less strict approach?

kdy1: I think it‘s fine.

kdy1:
Are you fine with using `dev` branch as the target instead of `main` while everything is implemented?

magic-akari:
> Are you fine with using `dev` branch as the target instead of `main` while everything is implemented?

I'll introduce the breaking change in Step 3 when switching the default JSX runtime. Given that a breaking change is unavoidable regardless, It's OK to use `dev` branch.

crates/swc_ecma_transforms_react/src/jsx/automatic.rs
```diff
@@ -0,0 +1,666 @@
...
+pub fn automatic<C>(
+    options: AutomaticConfig,
+    common: CommonConfig,
+    unresolved_mark: Mark,
+    comments: Option<C>,
+) -> impl Pass + VisitMut
+where
+    C: Comments + 'static,
+{
+    let add_pure_comment: Lrc<dyn Fn(BytePos)> = match comments {
```
kdy1, line 34: I like this idea

crates/swc_ecma_transforms_react/src/jsx/automatic.rs
```diff
@@ -0,0 +1,666 @@
...
+    throw_if_namespace: bool,
+
+    add_pure_comment: Lrc<dyn Fn(BytePos)>,
+}
+
+impl Automatic {
+    fn inject_runtime<T, F>(&mut self, body: &mut Vec<T>, inject: F)
+    where
+        T: StmtLike,
+        // Fn(Vec<(local, imported)>, src, body)
```
kdy1:
```suggestion
```

kdy1 (commented):
Looks good to me

kdy1 (approved):

Issue #10610, opened 2025-06-13
minify Node binding wastes cycles performing string -> JSON -> string


dmichon-msft:
### Describe the feature

In the simple case of minifying a block of text, the current `minify()` projection into NodeJS wastes a fair amount of CPU invoking `JSON.stringify` on the raw text before passing it into the minifier, which then deserializes the JSON back into a string.

It seems like there should be a better way to handle differentiating the `{ [fileName: string]: string }` input format from `code: string` that doesn't require such an expensive operation. Projecting separate binding entry points and switching between them in the Node API seems like it would be cleaner? Or having which data type be an option in one of the other parameters?

### Additional context

I'm currently trying to diagnose why switching from running Terser in 8 worker threads to the SWC minifier isn't showing the performance gains that I expected, and so I'm poking into bottlenecks.

kdy1:
Are you using swc/core?

kdy1:
Also you don’t need js worker threads when you are using SWC

dmichon-msft:
Using @swc/core@1.12.1
Old method: 8 workers of Terser, communicated via postMessage
New method: Invocation of `minify` in `@swc/core` on main thread, currently using the binding directly so that I can serialize the `options` buffer exactly once for the thousands of invocations of the minifier.

Admittedly in the profiler the time in JSON processing only shows as 1.5 seconds of a 203 second, but it's 1.5 seconds of unnecessary work and also quite GC intensive.

kdy1:
I’ll fix this today

PR #10628, opened 2025-06-19
perf(es/minifier): Remove needless JSON conversion

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/swc-project/swc/issues/10610


packages/core/src/index.ts
```diff
@@ -63,18 +63,18 @@ export function plugins(ps: Plugin[]): Plugin {
 export class Compiler {
     private fallbackBindingsPluginWarningDisplayed = false;
 
-    async minify(src: string, opts?: JsMinifyOptions, extras?: NapiMinifyExtra): Promise<Output> {
+    async minify(src: string | Buffer, opts?: JsMinifyOptions, extras?: NapiMinifyExtra): Promise<Output> {
         if (bindings) {
-            return bindings.minify(toBuffer(src), toBuffer(opts ?? {}), extras ?? {});
+            return bindings.minify(Buffer.from(typeof src === 'object' ? JSON.stringify(src) : src), toBuffer(opts ?? {}), typeof src === 'object', extras ?? {});
```
dmichon-msft, line 68: @kdy1 
You widened the allowed type of `src` to include `Buffer`, in which case `typeof src === 'object'`, which means if a caller passes a `Buffer` it will get stringified, which is going to make a mess. If you want to include `Buffer` in the allowed input types you'll want to handle it explicitly.

kdy1: Ah you are right
Thank you!

PR #10638, opened 2025-06-19
feat(plugin/runner): Use wasmer-wasix sys-minimal feature to reduce binary size

hardfist:
**Description:**
In https://github.com/wasmerio/wasmer/issues/5612 we introduce sys-minimal feature to support only enable used wasmer-wasix feature to reduce binary size
This PR upgrade wasmer related crate to latest version and use sys-minimal feature instead of sys feature

**BREAKING CHANGE:**
it should pass all tests in swc_plugin repo, but not 100% sure will pass all existing wasm plugins

### DIFF
--- a/crates/swc_plugin_runner/Cargo.toml
+++ b/crates/swc_plugin_runner/Cargo.toml
@@ -23,6 +23,7 @@ plugin_transform_host_native = [
   "wasmer/cranelift",
   "wasmer/sys",
   "wasmer-wasix/sys-minimal",
+  "wasmer-wasix/host-fs",
   "wasmer-compiler-cranelift/default",
 ]
 plugin_transform_host_native_shared_runtime = ["tokio"]
--- a/crates/swc_plugin_runner/Cargo.toml
+++ b/crates/swc_plugin_runner/Cargo.toml
@@ -22,12 +22,7 @@ plugin_transform_host_js = [
 plugin_transform_host_native = [
   "wasmer/cranelift",
   "wasmer/sys",
-  "wasmer-wasix/sys",
-  "wasmer-wasix/logging",
-  "wasmer-wasix/host-fs",
-  "wasmer-wasix/sys-poll",
-  "wasmer-wasix/sys-thread",
-  "wasmer-wasix/host-threads",
+  "wasmer-wasix/sys-minimal",
   "wasmer-compiler-cranelift/default",
 ]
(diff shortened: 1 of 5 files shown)


crates/swc_plugin_runner/Cargo.toml
```diff
@@ -22,11 +22,9 @@ plugin_transform_host_js = [
 plugin_transform_host_native = [
   "wasmer/default",
   "wasmer-wasix/sys",
```
hardfist: I removed all default enabled feature for wasmer-wasix/sys and manually enabled it here so it wouldn't enable unnecessary features like networking

bgw: Is it possible to upstream something to wasmer to make this configurable without needing a fork?

hardfist: yeah, that's the plan

hardfist: https://github.com/wasmerio/wasmer/issues/5612 feature request for wasmer

hardfist: @bgw @kdy1  turns out it's not a trivial feature for wamser-wasix to implement subset of wasi features(https://github.com/wasmerio/wasmer/issues/5612#issuecomment-2996232778), so there're some options here
* waiting until wasmer team  implement subset of wasi features
* temporarily fork and use in swc and when wasmer support subset of wasi then change to official wamer version
* consider move to wasmtime + wasi-common(that's what zed's using for wasm plugin), it's 14M vs 23M(wasmer + wasmer-wasix)

kdy1: I think we should try wasmtime, but I'm not sure about the performance.

hardfist: I can give wasmtime a try and compare the performance,considering they both use Cranelift as backend so maybe no big performance expected

kdy1: cc @bgw Any thoughts?

syrusakbary:
Hey @hardfist , if you could open a PR with your changes to Wasmer we would happily merge it.

Note: we would need this to be feature-gated

crates/swc_plugin_runner/Cargo.toml
hardfist: since https://github.com/wasmerio/wasmer/pull/5621 is merged, we don't need fork any more

PR #10639, opened 2025-06-19
refactor(es/react): Merge jsx-dev pass into jsx pass

magic-akari:


magic-akari:
Roadmap:
- #10616
- #10639
- Set the automatic runtime as the default behavior

crates/swc/tests/fixture/issues-8xxx/8210/output/1.js
```diff
@@ -5,7 +5,7 @@ const Component = ()=>{
             fileName: "$DIR/tests/fixture/issues-8xxx/8210/input/1.js",
             lineNumber: 2,
             columnNumber: 23
-        }, void 0),
+        }, this),
```
magic-akari, line 8: https://www.typescriptlang.org/play/?jsx=5#code/PTAEAECsGcA8C5QCcCmBDAxgFwLQ1gCYoBuAUBgPYB20WoAwhQLYAO1KVdAvKABQCUoLgD5QAb1KgpyFFgCuSKnwA8LUFgAWASyoBzLmOVpQwYQF9hACRQAbGxWXAWw-qTNA

This is a bugfix. Based on TypeScript's output, the correct value should be `this`. 
Not sure why SWC was outputting `undefined`.

crates/swc_ecma_transforms_react/src/jsx/development.rs
```diff
@@ -0,0 +1,94 @@
...
+        if self.is_this_allowed() {
+            Expr::This(ThisExpr::dummy())
+        } else {
+            *Expr::undefined(DUMMY_SP)
+        }
+    }
+}
+
+#[macro_export]
+macro_rules! visit_mut_development {
```
kdy1, line 71: Can you add a comment that this is not a public API?

magic-akari: Removed the `#[macro_export]` attribute.

PR #10644, opened 2025-06-20
perf(ecma/parser): simd fin non whitespace char

bvanjoi:


crates/swc_ecma_lexer/src/lib.rs
```diff
@@ -14,6 +14,7 @@
 #![allow(clippy::wrong_self_convention)]
 #![allow(clippy::match_like_matches_macro)]
 #![allow(unexpected_cfgs)]
+#![feature(portable_simd)]
```
kdy1, line 17: We can’t use this

bvanjoi:
Closing this investigation as no significant performance improvements were identified.

PR #10648, opened 2025-06-20
perf(es/lexer): Introduce CowUtils

magic-akari:


magic-akari:
Closing this PR as it does not demonstrate any noticeable performance improvements.

magic-akari:
However, I believe this could help reduce memory allocations.

Issue #10651, opened 2025-06-20
Regression in handling &middot;


ibacher:
### Describe the bug

Say I have a React component like this:

```jsx
const a = () => {
    return (
        <>&middot;</>
    )
}
```

In SWC up to 1.12.1, this produces:

```js
var a = function() {
    return React.createElement(React.Fragment, null, "\xb7");
};
```

From 1.12.2 on, this same code produces:

```js
var a = function() {
    return React.createElement(React.Fragment, null, "&middot;");
};
```

### Input code

```typescript
const a = () => {
    return (
        <>&middot;</>
    )
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": true
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": {
        "toplevel": false,
        "keep_classnames": false,
        "keep_fnames": false,
        "keep_private_props": false,
        "ie8": false,
        "safari10": false
      }
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Link to the code that reproduces this issue

https://play.swc.rs/?version=1.12.1&code=H4sIAAAAAAAAA0vOzysuUUhUsFXQ0FSwtVOo5lIAgqLUktKiPAUNMAcEbOzUcjNTUvJLrG307cCimly1APVuwO48AAAA&config=H4sIAAAAAAAAA32QTQ7CIBCF9z1FM2sXutAY7%2BAZzASHhgqFMLSxaXp36Q%2B1auMGmPe%2BmQmvy%2FIcShZwybv4jIVDz%2BSXOircVgGfUQESBll45QLsklvyYAVf06j0kwEBfUFhbOLjTIO2lilqEjXTrBlVKdmu9wlrnCfmT3BAsSo0rdBhj3WaGtLfcLQeRO4mNDJXaOhnXCLkX9d51WCgeFu3BSk6b6iMEr067JM1O32WzjEmMPZerz4EoXU0RXaCN5QCWpaA4mvqHIPvXzdUBA3HAQAA

### Expected behavior

At least without a major version change, this code should continue to produce in 1.12.x:

```js
var a = function() {
    return React.createElement(React.Fragment, null, "\xb7");
};
```

### Version

1.12.2

PR #10652, opened 2025-06-20
fix(es/parser): Parse jsx entity

magic-akari:
**Related issue:**
- Closes: #10651


magic-akari:
> [!NOTE]  
> In the future, we should postpone the escaping of JSX entities until the transformation phase and remove the escaped value from AST

PR #10665, opened 2025-06-21
perf(es/lexer): Optimize lexing of numbers with separators

kdy1:
**Description:**

`str.replace('ascii_char', "")` can be specialized.


magic-akari:
We should implement our own version of `cow_replace`.

kdy1:
@magic-akari I made `cow-replace` crate with https://github.com/dudykr/ddbase/pull/89

PR #10667, opened 2025-06-21
perf(swc/lexer): less alloc in read num

bvanjoi:


bvanjoi:
I've found optimization patterns similar to those implemented in #10655. Let's verify this by comparing benchmark results.

bvanjoi:
The benchmark indicates this implementation acceptable.

crates/swc_ecma_lexer/src/common/lexer/mod.rs
```diff
@@ -585,80 +587,79 @@ pub trait Lexer<'a, TokenAndSpan>: Tokens<TokenAndSpan> + Sized {
...
-                _ => None,
-            };
-
-            if let Some(value) = fast_value {
-                let raw_number_str = raw_str.replace('_', "");
-
-                return Ok((value, LazyBigInt::new(raw_number_str), non_octal));
-            }
-        }
+        // let mut non_octal = false;
```
kdy1: Please remove these commented lines

PR #10670, opened 2025-06-21
pref(es/parser): try use fast parse number

bvanjoi:
To evaluate whether this implementation offers performance advantages over the standard library


bvanjoi:
Closing as the benchmarks show no measurable improvement:

image

PR #10672, opened 2025-06-21
pref(es/parser): batch iter for skip_whitespace

bvanjoi:


kdy1:
Performance regressed for many lexer benchmark cases. Is it expected?

PR #10673, opened 2025-06-21
Fix 10662: make the inline optimization compatible with earlier rust versions

### LINKED ISSUE #10662: hstr does not compile with rust 1.86.0
### Describe the bug

We have dependencies which are dependent on hstr, which are now fetching 1.1.0.  We are currently using version 1.86.0 of the toolchain, and hstr fails to compile with the following error:

```
error: `core::slice::<impl [T]>::copy_from_slice` is not yet stable as a const fn
   --> /workspace/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hstr-1.1.0/src/dynamic.rs:161:13
    |
161 | /             unsafe_data
162 | |                 .data_mut()
163 | |                 .split_at_mut(len)
164 | |                 .0
165 | |                 .copy_from_slice(text.as_bytes());
    | |_________________________________________________^
    |
help: add `#![feature(const_copy_from_slice)]` to the crate attributes to enable
   --> /workspace/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hstr-1.1.0/src/lib.rs:4:1
    |
4   + #![feature(const_copy_from_slice)]
    |
```

### Input code

```typescript

```

### Config

```json

```

N/A

### SWC Info output

_No response_

### Expected behavior

For hstr 1.1.0 to compile with rustc 1.86.0

### Actual behavior

hstr 1.1.0 fails to compile with the error:

### Version

hstr 1.1.0

### Additional context

This is a nested dependency issue

lukesandberg:
The `copy_from_slice` method was not stabilized as `const` until 1.90. This replaces that method with a simple while loop which should be much more 'const compatible'

Tested with `rustup run 1.86 cargo test   -p hstr`  (note that also required removing the `-Zshare-generics=y` compiler flag).

Fixes #10662

### DIFF
--- a/.changeset/tidy-tigers-lick.md
+++ b/.changeset/tidy-tigers-lick.md
@@ -1,0 +1,6 @@
+---
+hstr: patch
+swc_core: patch
+---
+
+Fix 10662: make the inline optimization compatible with earlier rust versions
--- a/crates/hstr/src/dynamic.rs
+++ b/crates/hstr/src/dynamic.rs
@@ -155,14 +155,16 @@ pub(crate) const fn inline_atom(text: &str) -> Option<Atom> {
         // INLINE_TAG ensures this is never zero
         let tag = INLINE_TAG | ((len as u8) << LEN_OFFSET);
         let mut unsafe_data = TaggedValue::new_tag(NonZeroU8::new(tag).unwrap());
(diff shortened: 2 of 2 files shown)


bwilliams-sequence:
Thanks for acting on this so quickly!

PR #10677, opened 2025-06-22
perf(ts/fast-strip): Use `swc_ecma_parser::Lexer`

CPunisher:
**Description:**

`ts_fast_strip` can benefit from the new lexer, see: https://github.com/swc-project/swc/discussions/10683

### DIFF
--- a/.changeset/swift-socks-reflect.md
+++ b/.changeset/swift-socks-reflect.md
@@ -1,0 +1,7 @@
+---
+swc_ecma_parser: minor
+swc_ts_fast_strip: minor
+swc_core: minor
+---
+
+refactor(ts/fast-strip): use `swc_ecma_parser::Lexer`
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6563,7 +6563,7 @@ dependencies = [
  "swc_common",
  "swc_ecma_ast",
  "swc_ecma_codegen",
- "swc_ecma_lexer",
+ "swc_ecma_parser",
  "swc_ecma_transforms_base",
  "swc_ecma_transforms_react",
  "swc_ecma_transforms_typescript",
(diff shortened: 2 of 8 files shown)


kdy1 (commented):
Is the goal of the PR improving the performance of fast-strip?

CPunisher:
> Is the goal of the PR improving the performance of fast-strip?

- This pr is a trial because it also changes lexer and tokens which I think should be changed in a separate pr.
- Improving the performance of fast-strip is one of the reasons.
- Another important reason is that I find that `fast-strip` is the only crate that depends on the old compatible `swc_ecma_lexer::Lexer` in swc projects. We may have chance to remove the old lexer in the future if we solve the following problems:
   1. Since we have made the old lexer public, downstream crates like `rspack` depends on it. 
   2. Although I think it's not appropriate to expose any lexer, due to the problem 1, we have to provide alternatives for the users, which means we also have to make the new lexer public.
   3. `fast-strip` depends on one of the lexers, so we have to make one of the lexers public.
   4. Based on the problem 1 and 2, we should make sure that `swc_ecma_parser::Lexer` is the exact equivalent of `swc_ecma_lexer::Lexer`. It could be hard to prove because `swc_ecma_parser::Lexer` doesn't have strict token validation as we discussed in https://github.com/swc-project/swc/pull/10399

CPunisher:
Well, I think it's worth to open a discussion to clarify my ideas.

kdy1:
image

Wow it was huge

kdy1:
I think 4 may not be possible to resolve. cc @bvanjoi Any ideas?

bvanjoi:
I'm bot sure whether any unsolvable cases exist in this.

kdy1:
> Since we have made the old lexer public, downstream crates like rspack depends on it.

The main reason I think we should keep the old lexer is drpint. It’s a formatter, so my guess is that it requires correct token.

cc @dsherret Is this the case?

I think all other usecase would work well with the lexer in swc_ecma_parser too

CPunisher:
> Well, I think it's worth to open a discussion to clarify my ideas.

Done in https://github.com/swc-project/swc/discussions/10683. Let's discuss there.

kdy1:
Can you rebase / undraft this?

CPunisher:
> Can you rebase / undraft this?

I will first do some works in other prs for this.

kdy1 (approved):

PR #10679, opened 2025-06-22
perf(es/minifier): Make `CharFreq::scan` in mangler table-based

CPunisher:
**Description:**

Use table to reduce comparison cost

Compare: https://godbolt.org/z/ME57vnsMz


crates/swc_ecma_minifier/src/util/base54.rs
```diff
@@ -14,7 +14,7 @@ use swc_ecma_visit::{noop_visit_type, visit_obj_and_computed, Visit, VisitWith};
 
 #[derive(Clone, Copy)]
 
-pub(crate) struct CharFreq([i32; 64]);
+pub(crate) struct CharFreq([i32; 65]);
```
kdy1: Is this change reall required?

CPunisher: I extend the array to avoid extra checks in the table. Otherwise what should we do with:

```rs
_ => {
  table[i] = 64;
}
```

kdy1: I think 64 would perform better because of the CPu cache on x86_64 cpus (cache line is typically 64 bytes)

CPunisher: > I think 64 would perform better because of the CPu cache on x86_64 cpus (cache line is typically 64 bytes)

you are right, let me think of another way

crates/swc_ecma_minifier/src/util/base54.rs
```diff
@@ -193,37 +193,42 @@ impl CharFreq {
...
+                i += 1;
+                if i == 256 {
+                    break;
+                }
             }
+            table
+        };
+
+        for &c in s.as_bytes() {
+            self.0[SCAN_INDEX_TABLE[c as usize] as usize] += delta;
```
dmichon-msft: Since you're having to look up every character in a 256-entry table, wouldn't it be cheaper to make the CharFreq working structure be a 256-entry array, add every character to its own bucket directly, and then extract the 64 relevant entries at the end? It avoids a lookup during the critical path in exchange for a one-time hit of 64 lookups and writes at the end, but it would use more cache to store the table during processing.

Alternatively, all valid values are in the range 0x24-0x7a (inclusive), so if you are willing to pay the three instructions per iteration (unsigned subtract 0x24, check < 0x56, branch) to bounds check the value you only need an 86-entry working array.

CPunisher: Awesome, I will try your idea!

kdy1:
Is this PR ready for a review?

CPunisher:
> Is this PR ready for a review?

Yes (if there's performance improvement), but there seems something wrong with the codspeed report.

kdy1:
Ah I'll fix codspeed and rebase this PR

crates/swc_ecma_minifier/src/util/base54.rs
kdy1: It's slightly faster than the code in the main branch.
image

CPunisher: @kdy1 Can we merge it?

kdy1: Yeap, I’ll merge it soon

kdy1 (approved):

PR #10682, opened 2025-06-23
fix(es/react-compiler): Use tsx syntax for parser

GiveMe-A-Name:
**Description:**

We using `is_react_compiler_required` in ts or tsx file. But it's not support parse tsx syntax, so it's always return false.

https://github.com/swc-project/swc/blob/main/bindings/binding_react_compiler_node/src/support.rs#L37-L39

**BREAKING CHANGE:**

### DIFF
--- a/.changeset/sweet-up-time.md
+++ b/.changeset/sweet-up-time.md
@@ -1,0 +1,5 @@
+---
+binding_react_compiler_node: patch
+---
+
+fix(binding_react_compiler_node): Should parse tsx or ts file correctly, otherwise it always return false.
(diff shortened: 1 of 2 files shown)


kdy1 (commented):
Would it be fine to unconditionlly use tsx syntax?

GiveMe-A-Name:
> Would it be fine to unconditionlly use tsx syntax?

Have any effect? I don't sure. In my impression, it's will parse a bit slower if using tsx syntax. 
And I notice it's always using tsx syntax in [next.js](https://github.com/vercel/next.js/blob/canary/crates/napi/src/react_compiler.rs#L33.) .
And ensure it will parse correctly if user using decorators, so i add decorators attributes.

kdy1 (approved):

PR #10684, opened 2025-06-23
perf(es/minifier): Use phf instead of lazy json

quininer:
**Description:**

This attempts to use phf instead of lazy json parsing, which should help optimize startup time. I expect some regressions for single query time (due to using xxhash instead of fxhash), but I think there is some optimization space for here. I opened this PR in advance to get some feedback and see if we are happy with this direction.

I use the precomputed-map crate here instead of the phf crate for [better performance, size, and compile time](https://github.com/quininer/precomputed-map?tab=readme-ov-file#what-and-why).

**BREAKING CHANGE:**


kdy1:
I think code is bit dirtry and not sure if it worths.
Do you have any profiling result?
codspeed would not help as it iterates a lot. My guess is that you should use hyperfine for this kind of profiling

quininer:
I haven't really profiling yet, but I haven't observed serious regressions in real projects. I'll try more.

I could use proc macros like the phf crate does to hide the complexity, but this would affect compilation performance. we still need a `build.rs` or xtask that generates code like this:

```rust
pub const PROPS: &Set<&str, _> = set!{
    "s",
    // ...
};

**Comment by kdy1** [comment]
Are you sure that compilation is faster?

https://github.com/swc-project/swc/actions/runs/15816420064/job/44576012094 : main branch, 5m 5s (3m 19s compile)
https://github.com/swc-project/swc/actions/runs/15814223739/job/44570156183?pr=10684 : this PR, 5m22s, (3m 17s compile)

**Comment by quininer** [comment]
@kdy1 This won't be faster than just include json from the main branch, but it will be faster than generating a lot of rust code.

**Comment by quininer** [comment]
I tried to do some profiling, and as expected, it is hard to beat the current lazy json in terms of single query. The only advantages are startup time and avoid memory allocation.

startup:
```
Benchmark 1: ../../target/release/examples/lazy_props
  Time (mean ± σ):       1.6 ms ±   0.2 ms    [User: 1.0 ms, System: 0.6 ms]
  Range (min … max):     1.1 ms …   2.0 ms    1780 runs

Benchmark 1: ../../target/release/examples/phf_props
  Time (mean ± σ):     442.0 µs ±  49.6 µs    [User: 275.6 µs, System: 117.3 µs]
  Range (min … max):   307.6 µs … 1133.4 µs    6342 runs
```
```
     Running `/swc/target/release/examples/lazy_props`
1.153285ms
```
on my i7-13700H, the current lazy json has roughly 1ms startup overhead.

single query:
```
single-query/generated  time:   [100.11 ns 100.25 ns 100.44 ns]
single-query/lazy-json  time:   [16.300 ns 16.336 ns 16.384 ns]
```
Currently, `Atom` hashmap is much faster than `&str` phf because `Atom` provides pre-computed hash.

```
single-query/generated  time:   [109.64 ns 109.79 ns 109.95 ns]
single-query/lazy-json  time:   [117.78 ns 117.97 ns 118.19 ns]
```
But if we consider the cost of create `Atom`, then phf has no degradation.

Given that we have atoms everywhere, I think this is a degradation overall. A lazy map of strings might be a better start, like https://github.com/swc-project/swc/blob/v1.12.6/crates/swc_html_utils/src/lib.rs#L11 and https://github.com/swc-project/swc/blob/v1.12.6/crates/swc_ecma_preset_env/src/corejs3/data.rs#L7

PR #10690, opened 2025-06-24
fix(es/parser): friendly error tip for tpl

bvanjoi:


crates/swc_ecma_lexer/src/common/parser/mod.rs
```diff
@@ -513,6 +513,7 @@ pub fn parse_shebang<'a>(p: &mut impl Parser<'a>) -> PResult<Option<Atom>> {
     })
 }
 
+#[inline(always)]
```
magic-akari: Since we always expect parsing to succeed, this path should be marked `#[cold]` and `#[inline(never)]`, right?

Issue #9929, opened 2025-01-22
Tsconfig JSX = "preserve" Support


geospatialweb:
### Describe the feature

As I understand it, SWC only supports the React options available in tsconfig JSX - which therefore excludes Vue JSX that I'm using.  Thus no support for SWC tsconfig `emitDecoratorMetadata = true` which is not supported by Vite esbuild - hence my desire to switch to SWC (plus Rust performance gains) but I cannot because I am not using React. Any thoughts or possibilities?

XiNiHa:
This also blocks the use of SWC (and maybe Rspack families?) in building Solid libraries since Solid libraries need to ship files with JSX preserved.

chenjiahan:
@kdy1 Could you release "@swc/types" for https://github.com/swc-project/swc/pull/10691? Thank you!

NunoCruzSW:
I agree—I’m very disappointed that neither the SWC team nor the SolidJS team seems to be addressing this. I’m stuck on Webpack, and while I tried Vite, it was difficult to integrate with my codebase. Right now I don’t feel like I have a solid toolchain to use Solid with SWC or anything that offers similar performance.

chenjiahan:
@NunoCruzSW https://github.com/swc-project/swc/pull/10691 has been merged to the dev branch. When this change is released (probably SWC v2), you will be able to preserve Solid JSX in Rspack or [Rslib](https://github.com/web-infra-dev/rslib).

kdy1:
@magic-akari Do you think we should resolve this issue before v2? I think v2 will take a long time to come out

magic-akari:
Here are a few possible approaches:

1.  Cherry-pick only the commits related to the `preserve` option and merge them into the main branch.
2.  Merge all the JSX-related changes into the main branch (this might be considered if option 1 proves difficult, **although I don't recommend it**).
3.  Consider releasing an alpha version of v2 earlier than planned.

Option 1 may introduce some additional migration overhead.
Both options 2 and 3 represent **breaking changes**. They could be viable **if users don't require other community SWC plugins, or exclusively use their own custom-built SWC plugins**.

kdy1:
I think cherry picking is the way to go

NunoCruzSW:
Looking forward for this. I don't use react and even for react devs trying to move to other jsx libraries it will be a plus

luoye-fe:
[...]

PR #10688, opened 2025-06-23
refactor(es/react): Set `automatc` as default jsx runtime

magic-akari:

PR #10691, opened 2025-06-24
refactor(es/react): Add `preserve` supports and ts config style preset

magic-akari:
- #9929

### DIFF
--- a/bindings/binding_core_wasm/src/types.rs
+++ b/bindings/binding_core_wasm/src/types.rs
@@ -793,7 +793,7 @@ export interface EsParserConfig {
   importAssertions?: boolean;
 }
 
-type JSXPreset = "react" | "react-jsx" | "react-jsxdev" | "preserve";
+type JSXPreset = "react" | "react-jsx" | "react-jsxdev" | "preserve" | "react-native";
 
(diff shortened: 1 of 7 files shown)


magic-akari:
Roadmap:
- #10616
- #10639
- #10688
- #10691

bindings/binding_core_wasm/src/types.rs
```diff
@@ -793,14 +793,16 @@ export interface EsParserConfig {
...
 /**
  * Options for transform.
  */
 export interface TransformConfig {
   /**
-   * Effective only if `syntax` supports ƒ.
+   * Effective only if `syntax` supports.
    */
-  react?: ReactConfig;
+  react?: JSXPreset | ReactConfig;
```
magic-akari, line 805: The `react` field now supports both object-style configuration (maintaining Babel compatibility) and string-based presets (like TypeScript's JSX presets). 
However, the current config `"react": "react"` appears slightly odd. I propose renaming this field to `jsx` during future configuration refinements.

kdy1: 👍

Issue #10692, opened 2025-06-24
`=` after closing JSX tag breaks parser


mischnic:
### Describe the bug

The input below, particularly the `</span>=`, breaks the parser

[works fine in Babel](https://babeljs.io/repl#?browsers=Chrome%2040&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABAFQKYGcoAoCUiDeAUIogE6pQilJbEmIA8AmnffagDaoC2qYU6ALz4A2qzaMAfOIn0G6AA4BDMJIAeDAPSKVkwTNmMARqU3TDjMwBoDJKKRCpbiALoBfcWbo4A3IQ9AA&forceAllTransforms=false&modules=false&shippedProposals=false&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=react&prettier=true&targets=&version=7.27.6&externalPlugins=&assumptions=%7B%7D)

### Input code

```typescript
function Test() {
  return (
    <Y
        elements={[
        <>
            <span>x</span>=
            <br/>
        </>,
        true
        ]}
    />
  );
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": true
    },
    "target": "es2024",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Link to the code that reproduces this issue

[...]

PR #10693, opened 2025-06-24
fix(es/parser): Rescan `>=` for JSX closing tag

magic-akari:
**Related issue:**

- Closes: #10692

### DIFF
--- a/.changeset/funny-kangaroos-sort.md
+++ b/.changeset/funny-kangaroos-sort.md
@@ -1,0 +1,7 @@
+---
+swc_ecma_lexer: patch
+swc_core: patch
+swc_ecma_parser: patch
+---
+
+fix(es/parser): Rescan `>=` for JSX closing tag


crates/swc_ecma_parser/src/lexer/token.rs
```diff
@@ -1469,6 +1469,17 @@ impl Token {
             _ => false,
         }
     }
+
+    pub(crate) fn should_rescan_into_gt_in_jsx(self) -> bool {
```
bvanjoi, line 1473: The current implementation works well, but for reference, TypeScript employs an alternative approach:
- Only eat the greater symbol in `>`/`>=`/`>>`/`>>=`.
- And then rescan again when parsing binary expression.

The approach means we don't need to rescan much time during parse jsx expression

kdy1 (approved):

crates/swc_ecma_parser/src/lexer/token.rs
magic-akari: Appreciate the insight on TypeScript's approach! If their implementation proves more efficient, we should definitely consider migrating. Two key cases come to mind: 

1. Binary expressions like `<div/>>1` (which your comment addresses)  
2. In JSX contexts where operators appear inside elements, like `<> <div>=</div> </>`. Here the `=` must be parsed as JSX text, not a binary operator — **which also requires rescanning.**

Before making changes, we'd need to deep-dive into TypeScript's implementation specifics – particularly how they handle lookahead/rescanning during JSX expression parsing. Their scanner/parser interaction model would be critical to evaluate for both correctness and performance. Let's investigate further!

PR #10695, opened 2025-06-25
feat(es/parser): add `override` and `out` keyword

CPunisher:
**Description:**

This is two missing keywords about typescript modifiers. Currently the parser parses the modifier flags by comparing the str, which should be refactor in the future (https://github.com/swc-project/swc/blob/ec0bfc7315ed12c30bbe36ae929d1f767108e885/crates/swc_ecma_lexer/src/common/parser/typescript.rs#L342).

I do this change first because `fast-strip` relies on this.


magic-akari (approved):

magic-akari:
Surprisingly, adding new tokens didn’t require any parse changes. This just works thanks to our existing "everything to string" handling—a fortunate accident.
Moving to token comparison later is the right call.

Issue #10553, opened 2025-06-03
jsx classic does not use provided identifier when identifier is via a parameter


dsherret:
### Describe the bug

When using the classic JSX transform and receiving the jsx factory via a parameter, swc will rename the parameter.

### Input code

```typescript
export function Home(React) {
  return <div>Hello World {count}</div>
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "exportDefaultFrom": true,
      "jsx": true
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Link to the code that reproduces this issue

https://play.swc.rs/?version=1.11.29&code=H4sIAAAAAAAAA0utKMgvKlFIK81LLsnMz1PwyM9N1QhKTUwu0VSo5lJQKEotKS3KU7BJySyz80jNyclXCM8vyklRqE7OL80rqbXRB0lw1XIBANTYLa9IAAAA&config=H4sIAAAAAAAAA1WPsQ6DMAxEd74Cee7aDp2rbv0IKzUoKCGRbSQQ4t9JQ0jbLb57vovXpm1hEAP3dk3PNERkIa5zUmQZFeekABmPYthGhcvp0hwD64M6nJw%2BOfgEKk9UgUHmImVlOwxQ5J40p8q1xIELQShpHTopCeDtaLvl90Mm%2BMgk8g9%2BUBx7V%2FePuqZUgg%2FvKZvlUF0iHfU3%2BEJnWQ0GK69zMx%2Bx7Z%2FdIds0AQAA

### Expected behavior

```ts
export function Home(React) {
    return React.createElement("div", null, "Hello World ", count);
}
```

### Actual behavior

```ts
export function Home(React1) {
    return React.createElement("div", null, "Hello World ", count);
}
```

### Version

1.11.29

magic-akari:
One possible solution:  

Move the `react` pass before the `resolver`.

PR #10697, opened 2025-06-25
fix(es/react): Run classic jsx transform before resolver pass

magic-akari:
**Related issue:**

- #10553

### DIFF
--- a/crates/swc_ecma_transforms_react/tests/__swc_snapshots__/src/refresh/tests.rs/include_hook_signature_in_commonjs.js
+++ b/crates/swc_ecma_transforms_react/tests/__swc_snapshots__/src/refresh/tests.rs/include_hook_signature_in_commonjs.js
@@ -1,5 +1,5 @@
-var _s = $RefreshSig$();
 import { jsx as _jsx } from "react/jsx-runtime";
+var _s = $RefreshSig$();
 import { useFancyState } from './hooks';
 import useFoo from './foo';
 export default function App() {


magic-akari:
Roadmap:
- #10616
- #10639
- #10688
- #10691
- #10697

magic-akari:
**Required Documentation Updates:**

1. **Custom JSX Transform Plugins (e.g., vue-jsx):**  
   When using custom JSX transformation plugins, users must disable SWC’s built-in JSX transform by configuring `runtime: "preserve"`.

2. **Special Note for `runtime: "preserve"`:**  
   If using `runtime: "preserve"`, enable `verbatim_module_syntax: true`. This is required because TypeScript’s transform cannot resolve JSX-style pragma references **– Without this, JSX pragmas appearing as imports will be treated as type-only imports and incorrectly removed.**

magic-akari:
**Since we've set `automatic` as the default, the classic JSX transform is disabled by default.** This means with minimal configuration, plugins can still access JSX AST nodes. 
However, explicitly declaring `preserve` remains the recommended approach for clarity.

Austaras:
Would it not be easier if we just change how we compare idents in jsx transform code?

magic-akari:
> Would it not be easier if we just change how we compare idents in jsx transform code?

- The classic JSX transform must be applied before the TypeScript processing pass. Otherwise, we'd need to reimplement equivalent logic within the TypeScript pass.  
- If we place the JSX transform after the resolver, the required implementation would largely duplicate the resolver's existing logic.

Austaras:
> - If we place the JSX transform after the resolver, the required implementation would largely duplicate the resolver's existing logic.  

Why since if I'm reading it right you're basically abandoning name resolve here.

magic-akari:
> > * If we place the JSX transform after the resolver, the required implementation would largely duplicate the resolver's existing logic.
> 
> Why since if I'm reading it right you're basically abandoning name resolve here.

The details have been provided in the comments.
https://github.com/swc-project/swc/blob/414b2f1873501a46c179d537cc3c4f5771add9d6/crates/swc/src/config/mod.rs#L309-L314

kdy1 (approved):

mischnic:
Isn't this a pretty breaking API change as well? Because users of the Rust API would have to change the pass order?

magic-akari:
> Isn't this a pretty breaking API change as well? Because users of the Rust API would have to change the pass order?

It depends on how our Rust API users interact with SWC. To my knowledge, Rspack uses the `build_as_input` API – our changes are internal to this function. So if Rspack upgrades to this new SWC version, it will automatically pick up the changes without issues.  

I checked Turbopack’s implementation: it calls more granular functions (including `react::react`), so Turbopack would need adjustments on their side to adopt this.  

If backward compatibility is desired, we *can* preserve the old behavior. The current PR breaks compatibility, but we can revert it.

magic-akari:
The most breaking change is that we removed a parameter from `react::react`— the `top_level_ctxt`. This parameter was used to mark JSX factory references as top-level. Now, all contexts is marked as unresolved and are handled by the resolver pass.  

To preserve the old behavior, we could reintroduce this parameter. Internally, SWC would pass `unresolved_ctxt`, while third-party Rust users during the transition period could continue passing `top_level_ctxt`.  

The second change is that `react::react` now returns `(impl Pass, impl Pass)` (a tuple of passes) instead of a single `impl Pass`. However, since a tuple of passes still satisfies the `Pass` trait, this introduces **no observable difference** for consumers—except that we can now opt to destructure the tuple if needed.  

---

That said, I lean towards keeping this breaking change. **Why?** Because it forces Rust’s compiler to flag all existing `react::react` call sites via errors—effectively *surfacing every usage that needs upgrading*. This makes the migration explicit and systematic.

kdy1:
I also think we should keep this change, although I'm thinking about simply ignoring `SyntaxContext` from the classic JSX pass and running it after the resolver (and Wasm plugins).

PR #10699, opened 2025-06-25
feat(es/parser): expose Token API with unstable feature flag

CPunisher:

See https://github.com/swc-project/swc/discussions/10683


magic-akari:
Should this also be exposed in `swc_core`?

kdy1:
I think so? With an `unstable_` prefix

Issue #10698, opened 2025-06-25
keywords in jsx element name breaks parser in >=1.12.2


freben:
### Describe the bug

In 1.12.2 and newer, you get an error when trying to use keywods like `type`, `interface` etc in a JSX element name. Example: `<a.b.type />`. This seems like an unintended regression since it worked in earlier versions.

### Input code

```typescript

```

### Config

```json

```

### Link to the code that reproduces this issue

https://play.swc.rs/?version=1.12.6&code=H4sIAAAAAAAAA0vOzysuUSipLEhVsFXQ0FSwtVOo5lIAgqLUktKiPIW80pwca65aLq5ksMLyzJT01BKgUoii5Pzcgvy81LwSK6gACIAM0wHzahEai1LzUlKLsNthAzFVD26aHtg9%2BnZctQCD%2F%2FcenwAAAA%3D%3D&config=H4sIAAAAAAAAA1WPSQ7CMAxF9z1F5DVbWHCHHsIKbpUqk%2BJUIqp6d9JMwM7%2Bg598TELAxhKe4shjXjwGpjD2rHCyEd9ZAZIGWQblI9y6u%2FFlxbBTUc5qQMSwUiwlvrc0aOeYsragZmqaUVYt6ZcnnfGBmP%2BDVxTtqke%2F4qaGBONeezHbHzF5qvgHfEMdNg6D4rk3yxPnB%2BvJL0oTAQAA

### Expected behavior

1.12.1 gives:

```
var type = function() {
    return null;
};
var widget = {
    component: {
        type: type
    }
};
var render = function() {
    return React.createElement(widget.component.type, null);
};
```

### Actual behavior

1.12.2 gives:

```
  x Unexpected token `type`. Expected jsx identifier
    ,-[input.jsx:12:1]
  9 | }
 10 | 
 11 | const render = () => {
 12 |     return <widget.component.type />
    :                              ^^^^
 13 | }
    `----


Caused by:
    0: failed to process js file
    1: Syntax Error
```

### Version

1.12.2

PR #10701, opened 2025-06-25
fix(es/parser): Support keywords as JSX member expression properties

magic-akari:
**Related issue:**

- Closes: #10698

### DIFF
--- a/crates/swc_ecma_parser/tests/jsx/basic/issue-10698/input.js
+++ b/crates/swc_ecma_parser/tests/jsx/basic/issue-10698/input.js
@@ -1,0 +1,7 @@
+<widget.component.type />;
+
+<widget.string>
+    <Foo.interface />
+</widget.string>;
+
+<widget:number></widget:number>;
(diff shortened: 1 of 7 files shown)


crates/swc_ecma_parser/tests/jsx/basic/issue-10698/input.js
```diff
@@ -0,0 +1,7 @@
+<widget.component.type />;
+
+<widget.string>
+    <Foo.interface />
+</widget.string>;
+
+<widget:number></widget:number>;
```
bvanjoi, line 7: Could you add these cases:

```tsx
<a-b.type />;
<a-type.c />;
<a-type.type />;
<type-type.type />;
<type-b.type />;
<type-type.c />;
```

magic-akari: > Could you add these cases:
> 
> ```tsx
> <a-b.type />;
> <a-type.c />;
> <a-type.type />;
> <type-type.type />;
> <type-b.type />;
> <type-type.c />;
> ```

The parser works. But clearly, the transformer isn't functioning correctly—it outputs incorrect code. I can't really come up with what a sensible correct output *should* look like either, though. To be honest, I think this is good enough for now.

kdy1 (approved):

PR #10712, opened 2025-06-26
refactor(es/preset-env): Use phf for corejs3 entry

quininer:
**Description:**

This is a follow-up to https://github.com/swc-project/swc/pull/10684. I simply chose an example to experiment with.

I did some simple testing and I believe there is no large regression in single query performance.
```
// before
es/preset-env/entry/import
                        time:   [531.47 ns 532.42 ns 533.55 ns]

// after                        
es/preset-env/entry/import
                        time:   [549.62 ns 551.02 ns 552.55 ns]
```

Since we do dedup when packing strings, it is more compact than the original json and has some size optimization.

```
// before
16397965	../../target/release/deps/polyfills-d11a2ed3340dd897

// after
16070261	../../target/release/deps/polyfills-4b49b166ca0322eb
```

Since `../../data/core-js-compat/entries.json` is 500k large, I believe it also saves close to 500k of memory.

**BREAKING CHANGE:**


crates/swc_ecma_preset_env/src/corejs3/entry.rs
```diff
@@ -1,35 +1,47 @@
...
-        (
-            k,
-            v.into_iter()
-                .map(|s: String| &*Box::leak(s.into_boxed_str()))
-                .collect::<Vec<_>>(),
-        )
-    })
-    .collect()
-});
+include!("../generated/corejs3_entries.rs");
```
kdy1: I'm not sure about committing binary files into the repository

quininer: In some ways this is not worse than include the original json, since these binary files are more compact. but I agree there must be a better way.

One solution is that we can use `.gitignore` to hide generated files, but we have to generate them manually when cargo publish. This is similar to the behavior of `browserslist-rs`.
Another solution is that we can use `git-lfs` to manage binary files. This is a common practice when you have to track binaries.

kdy1: A binary file is hard to inspect so I consider it as a security risk

kdy1: I expect many people to have similar opinion as me

quininer: The binary file is generated deterministically, and we can verify its consistency on ci.

1. Use `.gitignore` to prevent generated files from be stored. generate them only in publish-crates ci. provide tools for ci and anyone to verify the consistency of files in git tags and crates.
2. Store generated files in repository. Use ci to verify that the binary file modifications in the PR are consistent with the original json file.

crates/swc_ecma_preset_env/benches/polyfills.rs
kdy1, line 1: Can you split this file as a separate PR (and I'll merge it first) so we can profile the difference using codspeed?

quininer: I'm not looking forward to the performance of this implementation, I know there is a lot of space for optimization. The main purpose here is to see if we are happy with the integration in this form.

kdy1:
I think this PR regresses the performance, but can you rebase?

https://github.com/swc-project/swc/pull/10722#issuecomment-3011181456 is 28.1 us

kdy1:
It regresses the performance

quininer:
Sure. That matches my local testing. I think it's within the range that micro-optimizations can make up for. The main purpose of this PR is to see if we're happy with this integration.

kdy1:
Is the goal reducing the binary size? Or is the goal reducing the compile time? If it's the former, I'm fine with generating binary file with a build script at the build time.

quininer:
Our goal is binary size, but I don't want to significantly affect the compilation speed.

Current our scale is acceptable to put it to build script, but should be note that build phf is a loop solve process, which cannot be guaranteed to be completed. If there is a larger scale of data in the future, then the phf solve in build script will significantly affect compilation time.

I made some optimizations and I believe we are now faster than the main branch.
1. Adjusted hot path
2. Eliminated indirect address access
3. Eliminated one bounds check (but still kept most of them to ensure handling of untrusted binary data)
3. Eliminated utf8 check

but the biggest speedup came from the hasher replacement. I used `foldhash` (It is the current default hasher for hashbrown), which is still slower than fxhash, but fxhash currently has no suitable crate. The `rustc-hash` crate does not guarantee the stable of output across architectures, and the other `fxhash` crate is unmaintained.


```
main
$ taskset -c 0 env RUST_LOG="off" cargo bench -- import
es/preset-env/entry/import
                        time:   [510.44 ns 511.30 ns 512.85 ns]
                        
phf
$ taskset -c 0 env RUST_LOG="off" cargo bench -- import
es/preset-env/entry/import
                        time:   [505.31 ns 506.69 ns 509.38 ns]
```

kdy1:
Then I think we should commit the binary file to the repository, but I think we need a way to verify it, like adding many snapshot tests for the embeded binary so if it's modified the test fails. (Without cryptic numbers in code, ideally. Is it possible?)

quininer:
> You expect foldhash to have a consistent output across versions or platforms, such as for persistent file formats or communication protocols.

Checking the README for foldhash, it doesn't seem to be suitable for this either. I think in one of the issues they mentioned ensure the hash output is consistent with the semvar version.

I realized another benefit of the build script, which is that it will not be affected by the stable of hasher.

quininer:
> Then I think we should commit the binary file to the repository, but I think we need a way to verify it, like adding many snapshot tests for the embeded binary so if it's modified the test fails.

This is ok because our codegen is completely deterministic. we can generate data to a temporary directory when test, and then compare it with the data in commit to see if it is consistent.

> (Without cryptic numbers in code, ideally. Is it possible?)

The only cryptic numbers I suggest keep are the seed used in last successful phf build, to ensure fast builds.

kdy1:
Ah sounds good. So the seed numbers should be stored in the build scripts?

quininer:
Yes, it is stored in build script. I'm trying to change to build script. At least for now it doesn't affect compilation performance too much.

quininer:
I'll release version precomputed-map 0.2 tonight so it can be merged. if we're happy with it, I'll start replace other static maps.

kdy1:
Can we use phf crate instead? What’s the difference?

kdy1:
It has a crate to generate code AFAIK

quininer:
@kdy1 The phf crate uses a different algorithm and requires the use of siphasher, and it does not do any string pack. This makes it have no advantages in performance, compilation speed, and size.

quininer:
I have an [post](https://github.com/orgs/web-infra-dev/discussions/26) about this (English coming soon)

Cargo.toml
```diff
@@ -126,6 +126,9 @@ resolver = "2"
   wasmer                    = { version = "6.0.0", default-features = false }
   wasmer-wasix              = { version = "0.600.0", default-features = false }
 
+  precomputed-map           = { git = "https://github.com/quininer/precomputed-map" }
```
kdy1: We cannot use git dep. `cargo` will block publishing if we ever use one.

kdy1 (changes requested):

Cargo.toml
quininer: I will release version 0.2 of it tonight, then it can be merged

PR #10713, opened 2025-06-26
refactor(es/parser): less compare

bvanjoi:


kdy1:
The performance regressed.
image

kdy1:
Maybe it's a just flake?

bvanjoi:
> The performance regressed.

Let's investigate this anomaly by breaking it down into a list of PRs to identify which one introduced these regressions.

PR #10716, opened 2025-06-26
feat: expose tsEnumIsMutable to js API

ahabhgk:
**Description:**

Expose tsEnumIsMutable to js API

**BREAKING CHANGE:**

### DIFF
--- a/.changeset/rich-phones-explode.md
+++ b/.changeset/rich-phones-explode.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc: patch
+---
+
+feat: expose tsEnumIsMutable to js API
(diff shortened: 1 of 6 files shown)


magic-akari:
What is the background or context that necessitates exposing this configuration?

ahabhgk:
When using swc-loader (or @swc/core directly), users should able to opt-out inline enum optimization if their code contains mutating a enum, so they can have the expected result

magic-akari:
I can see two likely scenarios here:

1.  The user writes an `enum` but doesn't use `tsc` for type checking.
2.  The user bypasses checks using `@ts-ignore`.

Neither of these approaches is considered a best practice. As a compiler, SWC aims to optimize and encourage best practices, but it can't enforce them. Users might occasionally need an escape hatch from our defaults. While I wouldn't recommend it, exposing such an escape mechanism might be necessary. 

@kdy1 What are your thoughts on this?

kdy1:
I agree with you. Even `tsc` allows a way to bypass the strict check (`@ts-ignore`) so I think we should follow them.

magic-akari (dismissed):

PR #10732, opened 2025-06-28
pref(es/lexer): faster match known ident

bvanjoi:
Return early if the length is not acceptable.


bvanjoi:
Close as there doesn't has any improve

PR #10733, opened 2025-06-29
fix(es/parser): Allow non-prop operand in delete

bvanjoi:
Expressions like `delete 1` and `delete []` should not trigger errors in a JavaScript/TypeScript parser, as these represent semantic errors rather than syntax errors.


kdy1 (commented):
Can you write some description?

bvanjoi:
> Can you write some description?

Done.

kdy1 (dismissed):

PR #10734, opened 2025-06-30
fix(es): Fix unicode handling

kdy1:
**Description:**


crates/swc_ecma_lexer/src/common/lexer/mod.rs
```diff
@@ -1112,26 +1112,31 @@ pub trait Lexer<'a, TokenAndSpan>: Tokens<TokenAndSpan> + Sized {
         op(self, &mut buf)
     }
 
-    fn read_unicode_escape(&mut self) -> LexResult<Vec<Char>> {
```
kdy1, line 1115: cc @bvanjoi Do you think this change is correct?

bvanjoi: If I understand correctly, this implementation appears incorrect. For `read_unicode_escape('\u0041')`, we expect the raw character sequence `['\\', 'u', '0', '0', '4', '1']` as output, not the decoded `'A'` directly.

kdy1: Thank you! I think you are right

PR #10676, opened 2025-06-22
pref(es/lexer): use bitflags for syntax

bvanjoi:

PR #10735, opened 2025-06-30
pref(ecma/parser): use reference for syntax

bvanjoi:


bvanjoi:
This shows only a minor improvement (0.3%).

bvanjoi:
In favor of #10676

PR #10754, opened 2025-07-02
fix(preset-env): Fix `default` value for `caniuse`

kdy1:
**Description:**

We should use the passed `default` value instead of static `false`.


kdy1:
I'll fix CI later

PR #10759, opened 2025-07-02
refactor(es/preset-env): Remove deprecated modules

magic-akari:
**Description:**

**BREAKING CHANGE:**

This is a breaking change for Rust users.


kdy1 (approved):
I'll merge this along with other Rust-side breaking changes

PR #10762, opened 2025-07-02
chore(common): Remove `clone()` in proc macro

bvanjoi:


crates/swc_ecma_lexer/src/common/parser/stmt.rs
```diff
@@ -1129,14 +1129,20 @@ fn parse_stmt_internal<'a, P: Parser<'a>>(
         };
         let span = p.span(start);
         if is_break {
-            if label.is_some() && !p.state().labels.contains(&label.as_ref().unwrap().sym) {
```
kdy1: Please extract this change to another PR

PR #10765, opened 2025-07-03
chore(deps): Remove unused deps with cargo-shear

CPunisher:
**Description:**

This could be helpful to the binary size


.github/workflows/CI.yml
```diff
@@ -78,6 +78,28 @@ jobs:
         run: |
           cargo deny check
 
+  cargo-shear:
```
kdy1: Can you add `cargo-shear` to `needs` array of the CI-done action?

kdy1 (dismissed):

kdy1:
Seems like there's a rebase mistake?

CPunisher:
> Seems like there's a rebase mistake?

Yes I just have rebased everything. BTW currently we cannot enable it in until the new `cargo-shear` version is released because of this bug https://github.com/Boshen/cargo-shear/pull/214

CPunisher:
But it seems that there's still something wrong. I will turn this pr into draft until I finish all the works.

kdy1 (approved):
I'll merge this as-is and add changeset from my local clone

PR #10791, opened 2025-07-07
pref(hstr): Do not compare string during creating atoms

bvanjoi:


kdy1 (approved):

crates/hstr/src/dynamic.rs
```diff
@@ -189,9 +189,7 @@ impl Storage for &'_ mut AtomStore {
         let (entry, _) = self
             .data
             .raw_entry_mut()
-            .from_hash(hash, |key| {
-                key.header.header.header.hash == hash && key.slice == *text.as_bytes()
-            })
```
bvanjoi, line 194: This means the change was made incorrectly due to collision concerns, so we should revert it. @kdy1

bvanjoi: I will revert this in #10813

PR #10792, opened 2025-07-07
fix(es): Make `swc_typescript` optional

hardfist:
**Description:**
swc_typescript should be optional cause it's not needed for all consumers

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc/Cargo.toml
+++ b/crates/swc/Cargo.toml
@@ -25,8 +25,9 @@ concurrent = [
 ]
 
 debug = ["swc_ecma_visit/debug", "swc_ecma_minifier/debug"]
-default = ["es3", "lint", "swc_typescript"]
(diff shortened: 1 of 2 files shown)


crates/swc/Cargo.toml
```diff
@@ -25,7 +25,7 @@ concurrent = [
 ]
 
 debug = ["swc_ecma_visit/debug", "swc_ecma_minifier/debug"]
-default = ["es3", "lint"]
+default = ["es3", "lint", "swc_typescript"]
```
kdy1: Can you rename the feature to `isolated_dts`? (or `isolated-dts` if it's more consistent)

hardfist: yeah isolated_dts is more meaningful

PR #10795, opened 2025-07-08
refactor(es/react): Remove redundant `replace` call

magic-akari:
**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/.changeset/bright-insects-jam.md
+++ b/.changeset/bright-insects-jam.md
@@ -1,0 +1,7 @@
+---
+swc_ecma_transforms_react: minor
+swc_ecma_utils: minor
+swc_core: minor
+---
+
+refactor(es/react): Remove redundant `replace` call
--- a/crates/swc_ecma_transforms_react/src/jsx/mod.rs
+++ b/crates/swc_ecma_transforms_react/src/jsx/mod.rs
@@ -1,7 +1,6 @@
 #![allow(clippy::redundant_allocation)]
 
 use std::{
-    borrow::Cow,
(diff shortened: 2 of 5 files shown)


magic-akari:
> JSX removes whitespace at the beginning and ending of a line. It also removes blank lines. New lines adjacent to tags are removed; new lines that occur in the middle of string literals are condensed into a single space. So these all render to the same thing:

https://legacy.reactjs.org/docs/jsx-in-depth.html#string-literals-1

Babel converts tabs (`\t`) to spaces, whereas TypeScript only strips leading/trailing whitespace (including both spaces and tabs). Since there's no rendering difference between these approaches, I prefer TypeScript's behavior: it eliminates a `replace()` call, yielding performance gains and reduced memory allocations.

- [TypeScript playground](https://www.typescriptlang.org/play/?#code/DwEwlgbgfAUAFgUwDZIPYEgDuqBOSQzAD040QA)
- [Babel playground](https://babeljs.io/repl/#?browsers=chrome%2050&build=&builtIns=false&corejs=3.42&spec=false&loose=false&code_lz=DwEwlgbgfAUAFgUwDZIPYEgDuqBOSQzAD040QA&forceAllTransforms=false&modules=false&shippedProposals=false&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2&prettier=true&targets=&version=7.28.0&externalPlugins=&assumptions=%7B%7D)

kdy1 (approved):

PR #10810, opened 2025-07-09
pref(hstr): only compare hash

bvanjoi:


kdy1:
preset-env is flaky

mischnic:
Is that actually correct though? Can't you have a hash collision?

bvanjoi:
> Can't you have a hash collision?

While theoretically possible, this collision is extremely unlikely in compiler scenarios. For reference, even [rustc/symbol](https://github.com/rust-lang/rust/blob/master/compiler/rustc_span/src/symbol.rs#L2767-L2785) - which handles more complex cases - only compares hash values.

bvanjoi:
If there are any concerns, I recommend adding this assertion before the return statement. If anyone reports a hash collision case, you can revert this change:

```rs
debug_assert!(
    (te_hash == oe_hash && te_slice == oe_slice) 
    || (te_hash != oe_hash && te_slice != oe_slice)
)
```

kdy1:
Does this have a performance improvement for real world test cases?
Asking because this is an optimization that improves performance only if the hash is same.

bvanjoi:
I made a mistake here: the code `inner.byte_strs.get_index_of(byte_str)` in rustc already handles hash collisions.

The following code demonstrates that `inner.byte_strs.get_index_of(byte_str)` is safe:

```rs
use std::hash::BuildHasher;

struct BuildSameHasher;

impl BuildHasher for BuildSameHasher {
    type Hasher = SameHasher;
    fn build_hasher(&self) -> Self::Hasher {
        SameHasher
    }
}

struct SameHasher;

impl std::hash::Hasher for SameHasher {
    fn finish(&self) -> u64 {
        123
    }

    fn write(&mut self, bytes: &[u8]) {
        ()
    }
}

fn main() {
    let mut set = indexmap::IndexSet::<&'static str, BuildSameHasher>::with_capacity_and_hasher(
        8,
        BuildSameHasher,
    );

    dbg!(set.len()); // 0
    set.insert("abc");
    dbg!(set.len()); // 1

    set.insert("abcd");
    dbg!(set.len()); // 2, means collision is handled correctly
}
```

PR #10813, opened 2025-07-09
refactor(hstr): Make the deallocation of `Atom`s explicit

bvanjoi:
**Description:**

Make the deallocation of `Atom`s explicit.

**BREAKING CHANGES:**

If you are developing a long-running process, you need to call `AtomStore::gc()`  for `Atom`s created by `AtomStore` API, and you need to call `global_atom_store_gc()` for strings created by other APIs, like `"".into()`.

### DIFF
--- a/crates/hstr/src/dynamic.rs
+++ b/crates/hstr/src/dynamic.rs
@@ -74,7 +74,7 @@ pub struct AtomStore {
 impl Default for AtomStore {
     fn default() -> Self {
         Self {
-            data: hashbrown::HashMap::with_capacity_and_hasher(64, Default::default()),
+            data: hashbrown::HashMap::with_capacity_and_hasher(64, BuildEntryHasher::default()),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


crates/hstr/src/dynamic.rs
```diff
@@ -88,18 +91,20 @@ impl AtomStore {
...
-            let v = GLOBAL_DATA.try_with(|global| {
-                let mut store = global.borrow_mut();
-                store.data.remove_entry(self)
-            });
-
-            if let Ok(Some((v, _))) = v {
-                v.into_inner();
-            }
-        }
+        // // If we are going to drop the last reference, we need to remove the
```
kdy1: This will cause some problems for nextjs dev servers

bvanjoi: Could you share more details about this? It seems there might be room for improvement.

kdy1: It leaks memory, so it's problematic for long-running processes.

cc @mischnic Can you double-check?

bvanjoi: Does 'leak memory' mean new strings are generated and inserted into this store, causing memory growth?(Of course, in this case, it shouldn't increment if no new strings are added.)

In this case, a better approach might be to add a new API called `AtomsStore::clean/AtomStore::gc` and call it at the start of compilation, rather than implementing reference counting and cleanup.

kdy1: I like the idea of `AtomStore::clean/gc`, given the performance improvement. So when it is called, all atoms in `AtomStore` is removed?
Then how to deal with the existing `Atom`s?

bvanjoi: I think:

- ​`clean`: Removes all atoms from `GlobalAtomStore`.
- ​`gc`: Only removes atoms that are no longer referenced anywhere.

In any case, creating and dropping atoms in the `AtomStore` just-in-time shows a performance disadvantage.

kdy1: I don’t expect such API to be trivial to implement. Problem is the dealloc itself, not just-in-time

kdy1: Can you implement actual dropping API within this PR?

kdy1 (commented):
Can you document this GC requirement on the rustdoc of `hstr::Atom` type?

It looks good, but I think we need to document it.

kdy1 (approved):

kdy1:
Thank you so much for all the hard work!

PR #10823, opened 2025-07-10
feat(hstr): with_capacity for atom store

bvanjoi:


kdy1 (commented):
It regresses the perf of ES lexer/parser.
Maybe the problem of initial capacity size?

Issue #10838, opened 2025-07-11
Add `UnknownBytes` to AST nodes to make Wasm plugin backward compatible (in the correct way)


kdy1:
https://github.com/swc-project/swc/pull/10842 is a hack, and https://github.com/swc-project/swc/pull/10841 will act as a guard for the hack.

Regardless, I'm going to add `UnknownBytes` to all extensible AST node type definitions. In that way, the Wasm plugin authors can return the unknown bytes as-is.

With #10842, unknown bytes will be dropped silently because those bytes are treated as a padding in some `ArchivedXXX` type.

PR #10842, opened 2025-07-12
feat(plugin): Remove `bytecheck` to make Wasm plugins backward compatible

### LINKED ISSUE #5060: Wasm plugins are not backward compatible

kdy1:
**Description:**

By default, and for obvious reasons, AST plugins are not compatible only if the memory layout changes. But with bytecheck, any field modification is an ABI-breaking change because `bytecheck` exists to do that.

**Related issues:**

 - Closes https://github.com/swc-project/swc/issues/5060

### DIFF
--- a/.changeset/nice-planes-hide.md
+++ b/.changeset/nice-planes-hide.md
@@ -1,0 +1,5 @@
+---
+swc_common: major
+---
+
+feat(plugin): Remove `bytecheck` to make Wasm plugins backward compatible
--- a/crates/swc_common/src/plugin/serialized.rs
+++ b/crates/swc_common/src/plugin/serialized.rs
@@ -126,22 +126,14 @@ impl PluginSerializedBytes {
     where
         W: rkyv::Archive,
         W::Archived: rkyv::Deserialize<W, rancor::Strategy<rkyv::de::Pool, rancor::Error>>,
-        for<'a> W::Archived: bytecheck::CheckBytes<
-            rancor::Strategy<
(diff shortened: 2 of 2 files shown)


kdy1:
For the _correct_ way to resolve Wasm plugin backward compatibility issue, see https://github.com/swc-project/swc/issues/10838

PR #10850, opened 2025-07-14
refactor(es/parser): inline parse stmt

bvanjoi:


kdy1:
I used `dev/rust` branch because I already published breaking changes recently

kdy1 (approved):

Issue #10863, opened 2025-07-15
bigint starts with 0 should not parsed


bvanjoi:
### Describe the bug

`0123n` should parse failed

### Input code

```typescript
0123n
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2015",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": true,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true,
        "const_to_let": true,
        "pristine_globals": true
      },
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": true,
  "isModule": true
}
```

### Link to the code that reproduces this issue

[...]

PR #10867, opened 2025-07-16
fix(es/lexer): Disallow legacy octal literals as BigInt

magic-akari:
**Description:**

**BREAKING CHANGE:**

**Related issue:**

- Closes: #10863


crates/swc_ecma_lexer/src/common/lexer/mod.rs
```diff
@@ -674,7 +674,10 @@ pub trait Lexer<'a, TokenAndSpan>: Tokens<TokenAndSpan> + Sized {
                 self.input_slice(lazy_integer.start, lazy_integer.end)
             };
 
-            if self.eat(b'n') {
+            // legacy octal number is not allowed in bigint.
+            if (!START_WITH_ZERO || lazy_integer.end - lazy_integer.start == BytePos(1))
```
magic-akari, line 678: The condition `lazy_integer.end - lazy_integer.start == BytePos(1)` ensures that only a single `0` character is allowed when `START_WITH_ZERO` is true, which correctly permits `0n` while blocking legacy octal cases like `0123n`.

Would appreciate a review to see if I've missed any edge cases here.

kdy1: Is 00n disallowed by the specification?

magic-akari: `00n` is not allowed because `00` is a legacy octal literal.

kdy1: I have thought about it for a while but failed to find a counterexample.
So I think you are right.

kdy1:
I'll wait for a review from @bvanjoi

bvanjoi (approved):

PR #10868, opened 2025-07-16
refactor(es/lexer): Avoid passing `convert`

CPunisher:
**Description:**

Since we have to use `Cow` there, I'm not sure which one is performant. But the code is more intuitive and clearer


CPunisher:
It's strange that the lexer gets faster while the parser gets slower 😂

kdy1:
It means this PR delegates the work done by lexer previously to the parser, but the total work is increased

kdy1 (approved):

Issue #10876, opened 2025-07-16
SWC 1.12.1+ incorrectly transforms closure-creating functions causing shared state across class instances


pgoldberg:
### Describe the bug

Starting in 1.12.1, SWC compression output is incorrect when using a function factory to construct a class method

### Input code

```typescript
const createCounter = () =>  {
    let count = 0;
    return  (numToAdd) => {
        count += numToAdd;
        return count;
    }
}

class Counter {
    add = createCounter();
}

const counter1 = new Counter();
const counter2 = new Counter();

console.log(counter1.add(1));
console.log(counter2.add(1));
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2024",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": false,
        "booleans": false,
        "booleans_as_integers": false,
        "collapse_vars": false,
        "comparisons": false,
        "computed_props": false,
        "conditionals": false,
        "dead_code": false,
        "directives": false,
        "drop_console": false,
        "drop_debugger": false,
        "evaluate": false,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": false,
        "hoist_vars": false,
        "if_return": false,
        "join_vars": false,
        "keep_classnames": false,
        "keep_fargs": false,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": false,
        "negate_iife": false,
        "properties": false,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": false,
        "switches": false,
        "typeofs": false,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": false,
        "const_to_let": false,
        "pristine_globals": false
      },
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Link to the code that reproduces this issue

[...]

Issue #10824, opened 2025-07-10
Minifier causing `this` to refer to last instantiated instance


Zylphrex:
### Describe the bug

When a class method is defined as an arrow function with a default argument, the `this` inside the method always refers to the most recently instantiated instance of the class.

In the example below, after instantiating `a1` and `a2`, calling `a1.c()` should have incremented the `a1.b`. However, it's incrementing `a2.b` instead.

I've narrowed it down to any combination of the following minifier options being enabled

- inline
- reduce_vars

Only when both options are disabled do we get the expected behaviour.

### Input code

```typescript
class A {
  b: number;

  constructor(b: number) {
    this.b = b;
  }

  c = (i = 1) => {
    this.b += i;
  };
}

const a1 = new A(1);
const a2 = new A(2);
a1.c();
console.assert(a1.b === 2);
console.assert(a2.b === 2);
```

### Config

```json
{
  "$schema": "https://swc.rs/schema.json",
  "env": {
    "mode": "usage",
    "coreJs": "3.41.0",
    "targets": [
      "last 10 Chrome versions",
      "last 10 Firefox versions",
      "last 4 Safari major versions",
      "last 3 iOS major versions",
      "Firefox ESR"
    ],
    "shippedProposals": true
  },
  "jsc": {
    "minify": {
      "compress": {
        "reduce_vars": true,
        "inline": 1
      }
    },
    "parser": {
      "syntax": "typescript"
    },
  },
}
```

### Link to the code that reproduces this issue

[...]

PR #10879, opened 2025-07-17
fix(es/minifier): Restrict IIFE inlining in class contexts

magic-akari:
**Description:**

Prevent function inlining in class field contexts to avoid variable scope issues.

**BREAKING CHANGE:**

**Related issue:**

- Closes: #10876 
- Closes: #10824


crates/swc_ecma_minifier/tests/fixture/next/36127/2/3/output.js
```diff
@@ -117,9 +119,11 @@ class t extends Text {
...
     replace(i, s, h) {
         if (!(h instanceof t)) return super.replace(i, s, h);
-        let l = n(this.text, n(h.text, n(this.text, [
-            ""
-        ], 0, i)), s), r = this.length + h.length - (s - i);
+        let l = n(this.text, n(h.text, function(t, e, i) {
+            return n(t, [
+                ""
+            ], 0, i);
+        }(this.text, 0, i)), s), r = this.length + h.length - (s - i);
```
Austaras: We should be possible to invoke IIFE in class methods

crates/swc_ecma_minifier/src/compress/optimize/mod.rs
```diff
@@ -179,6 +179,12 @@ bitflags! {
         const InWithStmt = 1 << 24;
 
         const InParam = 1 << 25;
+
+        /// `true` while we are inside a class body.
+        const InClass = 1 << 26;
+
+        /// `true` while we are in a static context (static method, static property, or static block).
+        const InStaticContext = 1 << 27;
```
Austaras: Maybe we only need `InNonStaticClass`?

crates/swc_ecma_minifier/tests/fixture/next/36127/2/3/output.js
magic-akari: I should try a different approach to fix this.

magic-akari: Actually, the key is to determine where we can safely inline declarations. They can be placed inside function scope or block scope – those locations are valid.

However, in places like class fields or anywhere that only accepts expressions, inlining declarations isn't possible. Attempting it there would cause the declarations to be hoisted outside to an enclosing scope, which is risky.

Austaras: But this approach is generally correct. You only need to unset `InClass` for function and arrow function

Austaras:
@magic-akari it's better to check whether in class body in [here](https://github.com/swc-project/swc/blob/main/crates/swc_ecma_minifier/src/compress/optimize/iife.rs#L823)

PR #10880, opened 2025-07-17
refactor(es/lexer): token eof

bvanjoi:


magic-akari:
This appears to be an unexpected closure triggered by the deletion of the target branch.

kdy1:
Nice catch, I was wondering where did the PR go 🤣

kdy1:
I think this is a good improvement. What matters is the performance of the parser.

It seems like this PR moves some work of the parser to the lexer, but the total work done by the parser is reduced.

crates/swc_ecma_transforms_base/tests/ts-resolver/ts_resolver_nested_enum/output.ts
```diff
@@ -1,11 +1,9 @@
 enum Foo__2 {
-    name__0,
-    string__0
+    name__0
```
bvanjoi, line 2: I think this change is acceptable because the input(`ts_resolver_nested_enum`) contains invalid syntax.

Issue #10849, opened 2025-07-14
Minifier bug with iife and undefined variable


ahabhgk:
### Describe the bug

Throw a runtime error that `obj` variable is not defined after minify the file with SWC

### Input code

```typescript
(function () {

const obj = { value: 42 };
console.log(obj === null || obj === void 0 ? void 0 : obj.value)

})()
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "transform": {
      "verbatimModuleSyntax": true
    },
    "target": "es2024",
    "externalHelpers": false,
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": true,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": true,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true,
        "const_to_let": true,
        "pristine_globals": true
      },
      "mangle": false
    },
    "experimental": {
      "emitIsolatedDts": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": true,
  "isModule": "unknown"
}
```

### Link to the code that reproduces this issue

[...]

PR #10891, opened 2025-07-18
fix(es/minifier): hoist props when only props are used

CPunisher:
**Description:**

We should not break the obj declaration into var declarations of its props if the obj var is directly used somewhere.

Not sure if the condition `accessed_props_count < usage.ref_count` is too strong or too weak.

fixes https://github.com/swc-project/swc/issues/10849

### DIFF
--- a/.changeset/selfish-colts-smell.md
+++ b/.changeset/selfish-colts-smell.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_minifier: patch
+---
+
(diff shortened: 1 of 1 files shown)


kdy1 (approved):

kdy1 (dismissed):

CPunisher:
Update a testcase which is actually wrong in the main branch

kdy1 (approved):

PR #10899, opened 2025-07-19
refactor(es/parser): less input mut

bvanjoi:


kdy1:
It seems to have no perf improvement. Is it expected?

bvanjoi:
yes, it's just a refactor.

PR #10907, opened 2025-07-21
feat(): Add `renamer_keep_contexts`

mischnic:
**Description:**

When using `renamer` outside of `hygiene` (i.e. you want to continue using the AST), want correct context contexts in some cases.

**BREAKING CHANGE:**


crates/swc_ecma_transforms_base/src/rename/mod.rs
```diff
@@ -57,25 +56,38 @@ pub fn rename(map: &RenameMap) -> impl '_ + Pass + VisitMut {
     rename_with_config(map, Default::default())
 }
 
-pub fn rename_with_config(map: &RenameMap, config: Config) -> impl '_ + Pass + VisitMut {
+pub fn rename_with_config<V: RenamedVariable + std::marker::Sync>(
```
mischnic: When using the imported `self::renamer_{concurrent,single}::Sync`, I got compile errors. Not sure how this is intended to work.

kdy1: You can just use `Sync` in `rename/mod.rs`.

It has

```rs
#[cfg(feature = "concurrent-renamer")]
use self::renamer_concurrent::{Send, Sync};
#[cfg(not(feature = "concurrent-renamer"))]
use self::renamer_single::{Send, Sync};
```

mischnic: > You can just use Sync in rename/mod.rs.

That's exactly what I tried first. And it didn't work. There were errors about the pass not being Sync.

mischnic: I've essentially replaced `IdentLike` with `RenamedVariable`, and `IdentLike` also had `std::marker::Sync` as a supertrait. So I've changed `RenamedVariable` to also declare that supertrait.

crates/swc_ecma_transforms_base/src/tests.rs
```diff
@@ -141,7 +141,7 @@ pub(crate) struct HygieneVisualizer;
 impl Fold for HygieneVisualizer {
     fn fold_ident(&mut self, ident: Ident) -> Ident {
         Ident {
-            sym: format!("{}{:?}", ident.sym, ident.ctxt).into(),
+            sym: format!("{}${}", ident.sym, ident.ctxt.as_u32()).into(),
```
kdy1, line 144: I expect this to create some test failures... but well...

kdy1 (commented):
I think I need to configure a binary size CI action before merging this PR.
I'll copy one from [rspack](https://github.com/web-infra-dev/rspack/tree/b9ec3fd639fdaeb89f353922ae548f4dc297889c/.github/actions/binary-limit) later this week

kdy1:
In case you have a local patch of SWC for turbopack, can you profile the binary size of turbopack with this patch?

mischnic:
I can't measure a difference:

swc/main 4eab8e89fd: 
```
$ ll packages/next-swc/native/next-swc.darwin-arm64.node
-rwxr-xr-x  1 niklas  staff   214M Jul 22 11:09 packages/next-swc/native/next-swc.darwin-arm64.node*
$ strip -x packages/next-swc/native/next-swc.darwin-arm64.node
$ ll packages/next-swc/native/next-swc.darwin-arm64.node
-rwxr-xr-x  1 niklas  staff   125M Jul 22 11:09 packages/next-swc/native/next-swc.darwin-arm64.node*
```

my fix f98cad2822 but without any code change on Turbopack side:
```
$ ll packages/next-swc/native/next-swc.darwin-arm64.node
-rwxr-xr-x  1 niklas  staff   214M Jul 22 11:13 packages/next-swc/native/next-swc.darwin-arm64.node*
$ strip -x packages/next-swc/native/next-swc.darwin-arm64.node
$ ll packages/next-swc/native/next-swc.darwin-arm64.node
-rwxr-xr-x  1 niklas  staff   125M Jul 22 11:14 packages/next-swc/native/next-swc.darwin-arm64.node*
```

my fix f98cad2822 and using renamer_keep_contexts:
```
$ ll packages/next-swc/native/next-swc.darwin-arm64.node
-rwxr-xr-x  1 niklas  staff   214M Jul 22 11:16 packages/next-swc/native/next-swc.darwin-arm64.node*
$ strip -x packages/next-swc/native/next-swc.darwin-arm64.node
$ ll packages/next-swc/native/next-swc.darwin-arm64.node
-rwxr-xr-x  1 niklas  staff   125M Jul 22 11:18 packages/next-swc/native/next-swc.darwin-arm64.node*
```

crates/swc_ecma_transforms_base/src/rename/mod.rs
```diff
@@ -84,33 +96,67 @@ where
         unresolved: Default::default(),
         previous_cache: Default::default(),
         total_map: None,
+        marker: std::marker::PhantomData::<Id>,
     })
 }
 
+pub trait RenamedVariable: Clone + Sized + std::marker::Send + std::marker::Sync + 'static {
```
kdy1: Can you seal this type?



```rust

mod private {
	pub trait Sealed {}
	impl Sealed for Id {}
	impl Sealed for Atom {}
};

```

mischnic:
Also, there is no overhead at all unless you're actually using `renamer_keep_contexts` because it compiles to the same code as before if you're only using `type Target = Atom`

crates/swc_ecma_transforms_base/src/rename/mod.rs
mischnic: done

crates/swc_ecma_transforms_base/src/hygiene/tests.rs
```diff
@@ -1756,3 +1777,74 @@ fn issue_2539() {
...
+        let mut program = Program::Module(tester.parse_module(
+            "actual1.js",
+            "
+                    const bar$1 = (patch$2)=>patch$2 + 1;
+                    const foo$1 = (patch$3)=>patch$3 !== patch$4;
+                ",
+        )?);
+
+        let expected = "
+        const bar$1 = (patch1$7)=>patch1$7 + 1;
```
kdy1, line 1802: I don't get why this is `$7`.

crates/swc_ecma_transforms_base/src/rename/analyzer/scope.rs
```diff
@@ -171,14 +169,18 @@ impl Scope {
                 }
 
                 if self.can_rename(&id, &sym, reverse) {
+                    let renamed = V::new_private(sym.clone());
```
kdy1, line 172: My guess is that it happens to be `$7` because of some `V::new_private` calls here, right?

kdy1: In current PR, `Renamer` use `Id`, while silently dropping `Id.1` and using `V::new_private` instead, which is defined as

```rust
impl RenamedVariable for Id {
    fn new_private(sym: Atom) -> Self {
        (sym, SyntaxContext::empty().apply_mark(Mark::new()))
    }

    fn to_id(&self) -> Id {
        self.clone()
    }
}
```

kdy1: I don't get the reason `V::new_private` is required. I think 

`let renamed = V::to_id(something)` is expected here

kdy1 (commented):
Can you write some descriptions as comments? (in the source code)_

crates/swc_ecma_transforms_base/src/rename/analyzer/scope.rs
mischnic: The map `to` which contains the `old_id -> new_id_or_atom` mapping needs to contain the syntax context for my usecase, so that all occurences of `old_id` get replaced with the same new value.

That is why I added `RenamedVariable`, so add `new_private` so that the generic `V` can store the syntax context in this case.

I added all of these generics because you said you were concerned about performance.

The alternatives would indeed be, as you said in the last comment, always have a `HashMap<Id, Id>` instead of the current `HashMap<Id, Atom>`, and use `V::to_id()` instead.
That would store slightly more data in the hashmap for the `hygiene` usecase.
With the current version, `hygiene` stores exactly the same datastructures as before (after monomorphizing the generic to `Atom`), but `renamer_keep_contexts` has slightly bigger hashmap entries.

kdy1: Can you add some comments to the code? I think future devs will not be able to understand the reasoning behind the API

crates/swc_ecma_transforms_base/src/tests.rs
mischnic: This is only used for the verbose logging.

The reason why I did this change was so because the form `name$i` is valid JS accepted by the parser, so I can use this in my test case as the `expected` fixture.

crates/swc_ecma_transforms_base/src/rename/analyzer/scope.rs
mischnic: I added some comments. Please let me know if I should go into further detail anywhere.

kdy1 (approved):

PR #10921, opened 2025-07-22
refactor(es/preset-env): Apply swc_ecma_compiler in preset env

magic-akari:
**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/.changeset/gentle-spies-complain.md
+++ b/.changeset/gentle-spies-complain.md
@@ -1,0 +1,5 @@
+---
+swc_ecma_preset_env: major
+---
+
+refactor(es/preset-env): Apply swc_ecma_compiler in preset env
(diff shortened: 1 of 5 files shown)


magic-akari:
> This may change the order of transform and may break many things.

Yes, this requires the compiler internals' execution order to align with the order declared in preset-env. Achieving this consistency is challenging, but it seems like the most solid iterative goal we can aim for. If the orders don't match, are there potential advantages or trade-offs we should consider that I might be missing?

kdy1:
It sounds right, but I think it **might** worth to publish a new version without this PR and merge this PR to make bisecting easier. What do you think?

kdy1:
I think alternatively, we can just merge this PR as the CI passed

kdy1 (approved):

Issue #10918, opened 2025-07-22
swc compress causes SyntaxError: Illegal break statement (2)


kenrick95:
### Describe the bug

Same as https://github.com/swc-project/swc/issues/10885

Seems not fixed yet in [1.13.1](https://github.com/swc-project/swc/blob/main/CHANGELOG.md#1131---2025-07-19)

### Input code

```typescript
import { useState } from 'react';
import { getCondition, doSomething } from './utils';

export default function useMeow() { 
    const [state, setState] = useState('init');
  const onMeow = async () => {
      switch (state) {
        case 'init': {
          const innerCondition = getCondition();
          switch (innerCondition) {
            case 'a':
              break;
            case 'b':
              break;
            default:
              await doSomething();
          }
          break;
        }
        default: {
          await doSomething();
          break;
        }
      }
    };
    return {
      onMeow,
    }
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "target": "es2022",
    "externalHelpers": true,
    "loose": false,
    "minify": {
        "compress": {
          "arguments": false,
          "arrows": false,
          "booleans": false,
          "booleans_as_integers": false,
          "collapse_vars": false,
          "comparisons": false,
          "computed_props": false,
          "conditionals": false,
          "dead_code": false,
          "directives": false,
          "drop_console": false,
          "drop_debugger": false,
          "evaluate": false,
          "expression": false,
          "hoist_funs": false,
          "hoist_props": false,
          "hoist_vars": false,
          "if_return": false,
          "join_vars": false,
          "keep_classnames": false,
          "keep_fargs": false,
          "keep_fnames": false,
          "keep_infinity": false,
          "loops": false,
          "negate_iife": false,
          "properties": false,
          "reduce_funcs": false,
          "reduce_vars": false,
          "side_effects": false,
          "switches": true,
          "typeofs": false,
          "unsafe": false,
          "unsafe_arrows": false,
          "unsafe_comps": false,
          "unsafe_Function": false,
          "unsafe_math": false,
          "unsafe_symbols": false,
          "unsafe_methods": false,
          "unsafe_proto": false,
          "unsafe_regexp": false,
          "unsafe_undefined": false,
          "unused": false,
          "const_to_let": false,
          "pristine_globals": false
        },
        "mangle": false,
      
      "format": {
        "beautify": true,
        "semicolons": false,
      },
    }
  },
  "module": {
    "type": "es6"
[...]

PR #10923, opened 2025-07-22
fix(es/minifier): Remove the last break in the second switch

nilptr:
**Description:**

fix: remove last the `break` in the 2nd case when turning 2 cases switch into if else

**BREAKING CHANGE:**

**Related issue (if exists):** #10918


kdy1 (approved):
Thank you! Can you sign the CLA?

Issue #10885, opened 2025-07-18
swc compress causes SyntaxError: Illegal break statement


kenrick95:
### Describe the bug

The produced code from the following input are:

```js
"use strict";
Object.defineProperty(exports, "__esModule", {
    value: !0
}), Object.defineProperty(exports, "default", {
    enumerable: !0,
    get: function() {
        return useMeow;
    }
});
const _react = require("react"), _utils = require("./utils");
function useMeow() {
    let [state, setState] = (0, _react.useState)('init');
    return {
        onMeow: async ()=>{
            if ('init' === state) {
                switch((0, _utils.getCondition)()){
                    case 'a':
                    case 'b':
                        break;
                    default:
                        await (0, _utils.doSomething)();
                }
                break; // <--- SyntaxError here!
            }
            await (0, _utils.doSomething)();
        }
    };
}
```

The highlighted point will throw a syntax error, because one cannot `break` outside a switch statement.

---

Workaround: Set `compress.switches` to `false` and it will generate code with valid syntax:

```js
"use strict";
Object.defineProperty(exports, "__esModule", {
    value: !0
}), Object.defineProperty(exports, "default", {
    enumerable: !0,
    get: function() {
        return useMeow;
    }
});
const _react = require("react"), _utils = require("./utils");
function useMeow() {
    let [state, setState] = (0, _react.useState)('init');
    return {
        onMeow: async ()=>{
            switch(state){
                case 'init':
                    switch((0, _utils.getCondition)()){
                        case 'a':
                            break;
                        case 'b':
                            break;
                        default:
                            await (0, _utils.doSomething)();
                    }
                    break;
                default:
                    await (0, _utils.doSomething)();
                    break;
            }
        }
    };
}
```

### Input code

```typescript
import { useState } from 'react';
import { getCondition, doSomething } from './utils';

export default function useMeow() { 
    const [state, setState] = useState('init');
  const onMeow = async () => {
      switch (state) {
        case 'init': {
          const innerCondition = getCondition();
          switch (innerCondition) {
            case 'a':
              break;
            case 'b':
              break;
            default:
              await doSomething();
          }
          break;
        }
        default: {
          await doSomething();
          break;
        }
      }
    };
    return {
      onMeow,
    }
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "externalHelpers": true,
    "loose": false,
    "minify": {
        "compress": {
          "arguments": false,
[...]

PR #10924, opened 2025-07-22
test(es/minifier): update fixture/issue/10918 test case

nilptr:
**Description:**

I notice that the test case in my previous PR #10923 (for issue #10918) is the same as PR #10898 (for issue #10885).

Issue #10918 was not caught because the `break` was removed by other minify options.

This PR is to update the minifier fixture/issue/10918 test case.

**BREAKING CHANGE:**

**Related issue (if exists):** #10918

### DIFF
--- a/crates/swc_ecma_minifier/tests/fixture/issues/10918/config.json
+++ b/crates/swc_ecma_minifier/tests/fixture/issues/10918/config.json
@@ -1,0 +1,3 @@
+{
+  "switches": true
+}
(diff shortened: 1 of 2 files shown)


kdy1 (approved):

Austaras:
1. #10918 is still not resolved
2. updated test would still function the same since we merge case with same body
3. we don't need to remove last break from last case here because it would be removed in other passes as it's useless

@nilptr @kdy1

kdy1:
I reopened it. Thank you!

PR #10925, opened 2025-07-22
chore(ci): Bump codspeed rust integrations to 3.0.4

art049:
**Description:**

Bumps the codspeed integration, bringing notably library detection, file path and better accuracy in microbenchmarks


art049:
~There is something wrong~
There was an issue with `cargo-codspeed@3.0.4` we're working on fixing it but everything works fine with 3.0.2!

kdy1 (approved):

PR #10943, opened 2025-07-25
refactor(plugin/runner): Introduce abstract runtime

quininer:
**Description:**

This PR introduces the `Runtime` trait abstraction to allow users to provide their own wasm runtime.

I use the dyn trait, which makes it have some runtime cost compared to before. but it should be almost invisible compared to wasm jit. The dyn trait allows us to switch runtime implement at runtime without bloat, so it is necessary.

**BREAKING CHANGE:**

It changes the public api of the `swc_plugin_runner` crate, but has no breaking changes to `swc` itself.


crates/swc/src/config/mod.rs
```diff
@@ -235,6 +235,10 @@ impl Options {
         config: Option<Config>,
         comments: Option<&'a SingleThreadedComments>,
         custom_before_pass: impl FnOnce(&Program) -> P,
+
```
kdy1: A parameter should never have `#[cfg]` flag. Otherwise every usage will break depedning on the `cfg` provided to the `swc` crate.

kdy1 (changes requested):

crates/swc/src/config/mod.rs
quininer: I can't find a good place to put it, do you have any suggestions?

kdy1: A builder method would be ideal, I guess?

kdy1: It would break as less as code as possible

quininer: Oh, I didn't notice that `Options::build_as_input` was a public method.

I'm considering two place:
1. Add a `serde(skip)` field in Options
2. Add a `plugin_runtime` field in `Handler`, but since `swc_common` is quite low-level, it needs to be a `Box<dyn Any>` to eliminate type.

kdy1: 1 would be better

.github/workflows/bench.yml
```diff
@@ -53,6 +53,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
+          target: wasm32-wasip1
```
kdy1: Why do we need this? Did we bench `swc_plugin_runner`?

kdy1: I disabled the automated benchmark for it because it's too flaky and too slow.

https://github.com/swc-project/swc/blob/64750702c98e2aca6e3c0d9e4a34974cc38b16e5/scripts/bench/list-crates-with-bench.sh

quininer: Oh, I forgot to delete this when I disable bench

crates/swc/src/plugin.rs
```diff
@@ -40,6 +42,7 @@ pub fn plugins(
     comments: Option<swc_common::comments::SingleThreadedComments>,
     source_map: std::sync::Arc<swc_common::SourceMap>,
     unresolved_mark: swc_common::Mark,
+    #[cfg(feature = "plugin")] plugin_runtime: Arc<dyn PluginRuntime>,
```
kdy1: `#[cfg]` should never be placed on a parameter

quininer: Ok. I noticed this isn't a public API and is only called from one place. so I'll put the function under cfg.

crates/swc_plugin_runner/src/host_environment.rs
```diff
@@ -11,12 +9,10 @@ use wasmer::Memory;
 ///
 /// ref: https://docs.wasmer.io/integrations/examples/host-functions#declaring-the-data
 #[derive(Clone)]
-pub struct BaseHostEnvironment {
-    pub memory: Option<Memory>,
-}
+pub struct BaseHostEnvironment {}
```
kdy1, line 12: Do we still need this type?

crates/swc_plugin_runner/src/runtime.rs
```diff
@@ -0,0 +1,131 @@
+//! Plugin runtime abstract
+
+use std::{any::Any, fmt, path::Path};
+
+/// Runtime FFI Value
+pub type Value = i32;
+
+/// Opaque module cache
+pub struct ModuleCache(pub Box<dyn Any + Send + Sync>);
```
kdy1, line 9: nit: I think it can be better if we do not expose `Any` as a public API as it allows user to specify anything, although it would be caught in runtime

crates/swc_plugin_backend_wasmer/Cargo.toml
```diff
@@ -0,0 +1,68 @@
+[package]
+name = "swc_plugin_backend_wasmer"
+version = "0.1.0"
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
```
kdy1:
```suggestion
edition = { workspace = true }
license = { workspace = true }
repository = { workspace = true }
```

It's a bug of `cargo-mono`, but this syntax is not supported.

crates/swc_plugin_backend_wasmer/benches/ecma_invoke.rs
```diff
@@ -44,26 +44,25 @@ fn plugin_group(c: &mut Criterion) {
 }
 
 fn bench_transform(b: &mut Bencher, plugin_dir: &Path) {
+    let runtime = Arc::new(WasmerRuntime);
```
kdy1, line 47: I think we may need to use a separate crate in the future PRs to avoid duplicating all the testing logics and benchmarking logics

crates/swc_plugin_runner/src/runtime.rs
quininer: It must be public so that users can access the custom type.

We could use a custom trait instead, but this would make the downcast unsafe and still not avoid runtime check.

crates/swc_plugin_runner/src/host_environment.rs
quininer: This isn't needed, but I'm keep it in case we might need to put something in future.

crates/swc_plugin_backend_wasmer/benches/ecma_invoke.rs
quininer: I'll do this in wasmtime PR.

Cargo.lock
```diff
@@ -7254,18 +7271,6 @@ dependencies = [
  "windows-sys 0.59.0",
 ]
 
-[[package]]
-name = "wasmer-cache"
```
kdy1, line 7258: Is this expected?

quininer: Yes, we use our own FileSystemCache instead of it. see https://github.com/swc-project/swc/pull/10943/files#diff-3bda5def6ce2b7553c3b3a5ad241c0bdb7021e67b7de1e594df4cd5a54d403b3R248

kdy1 (approved):

PR #10947, opened 2025-07-26
fix(es/minifier): Don't optimize Number properties when Number is shadowed

mertcanaltin:
Don't optimize Number properties when Number is shadowed


crates/swc_ecma_minifier/src/compress/optimize/evaluate.rs
```diff
@@ -145,6 +145,18 @@ impl Optimizer<'_> {
...
+                        .data
+                        .vars
+                        .get(&number_ident.to_id())
+                        .map(|var| var.flags.contains(VarUsageInfoFlags::DECLARED))
+                        .unwrap_or(false)
+                    {
+                        return;
+                    }
+                }
+
```
Austaras, line 153: Normally we just compare the ident's ctxt with `expr_ctx.unresolved_ctxt`

mertcanaltin: thanks, I edited

kdy1 (approved):

PR #10948, opened 2025-07-26
Mark `atom!` allocated Atoms as having a 'static' lifetime and stop refcounting them

lukesandberg:
The Clone and Drop routines will treat them like `inline` atoms but their storage will instead be in a `static` item constructed by the macro

This should speed up initialization and access to `atom!` values since there is no more lazy lock just. Instead we just take the address of a static and tag it. This does complicate `eq` and `hash` a bit as well as `as_str` to handle the new location. 

This follows a similar pattern used in turbopack as of https://github.com/vercel/next.js/pull/81994.  Things are different here due to the use of `ThinArc` for storage, which required me to introduce a separate struct for the `&'static str` case.

The motivation is simple, don't refcount values that will never be freed.

### DIFF
--- a/.changeset/itchy-cycles-compare.md
+++ b/.changeset/itchy-cycles-compare.md
@@ -1,0 +1,5 @@
+---
+hstr: patch
+---
+
+Mark `atom!` allocated Atoms as having a 'static' lifetime and stop refcounting them
--- a/crates/hstr/src/dynamic.rs
+++ b/crates/hstr/src/dynamic.rs
@@ -9,7 +9,6 @@ use std::{
     ptr::{self, NonNull},
 };
 
-use rustc_hash::FxHasher;
 use triomphe::ThinArc;
 
 use crate::{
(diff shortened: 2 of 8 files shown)


lukesandberg:
Benchmark results are interesting.  Looks like it is about a regression in the hash implementation which makes sense.  Will work on addressing that.

lukesandberg:
Whups, also a silly wasm32 breakage

lukesandberg:
well, it looks like the `as_str` and `eq` modifications are quite harmful.

The issue with `Atom` compared to `RcStr` is about the `ThinArc` representation.  My attempts to adopt this in turbopack were not fruitful. So im guessing the key issue is that turbopack spends a lot of time cloning and dropping, so optimizing that is worthwhile.  On the other hand many of the RcStr values we construct are 'de-novo' so being able to _move_ a `String` is important.  Perhaps in `swc` many of the atoms are copied out of an existing buffer, so the copy _into_ a `ThinArc` is required.

Reverting to a draft while i consider this

PR #10949, opened 2025-07-28
perf(es/minifier): Remove needless clone

bvanjoi:


kdy1:
image
It seems to regress the perf?

bvanjoi:
please see the latest change.

PR #10953, opened 2025-07-29
perf(es/minifier): Use bitflags for context of the `Pure` pass

bvanjoi:


kdy1 (approved):

bvanjoi:
@kdy1 Could we configure auto-merge to trigger only after successful benchmark verification? I want to prevent any unexpected performance regressions.

kdy1:
It’s not possible, but I can wait for it. I enabled auto merge for this because this is an obvious one.

PR #10954, opened 2025-07-29
refactor(hstr): use default inline strategy

bvanjoi:


CPunisher:
I think we'd better check the diff of binary size  

Take `binding_core_node` for example:
Before: 33390544 bytes
After: 33473168 bytes
~ +80KB

👍

kdy1:
I think we need a binary size CI action

kdy1:
@CPunisher Do you think we should merge this PR?

CPunisher:
> I think we need a binary size CI action

Yes. Actually rspack have done this. It upload the binding size to somewhere when finishing the ci in the main branch. And compare the binding size in each pr with the binding size of latest commit (in the main branch).

I plan to do this for swc, but there are two problems.
1. Where to upload the binding size data. Or we should build twice for each pr, which could be time-consuming.
2. I'm not sure if `binding_core_node` is a good bench case in terms of binary size.

CPunisher:
> @CPunisher Do you think we should merge this PR?

From my test on `binding_core_node `, I think it's acceptable.

kdy1 (approved):

kdy1:
Regarding the binary size, we may need a dedicated webapp?

Issue #10280, opened 2025-03-27
`es/minifier`: remaining cases where IIFE isn't invoked


Le0Developer:
### Describe the feature

More or less a tracking issue for the points in #10220

- [ ] [comma expressions (`return inlineMe(), 123`)](https://play.swc.rs/?version=1.11.13&code=H4sIAAAAAAAAAyvPzEvJL9dLVLBVSCvNSy7JzM%2FT0FSo5lJQKEotKS3KU8jMy8nMS%2FVN1dDUUTA0Muaq5eKCKUSSA%2BsoySjKL4coAgBRk2QXVwAAAA%3D%3D&config=H4sIAAAAAAAAA32US3LbMAyG9zmFR%2Bts00UP0F3PwKFFUGYqEhoCdKzJ%2BO4F9bCdGMxOxAcI5I%2FH58vh0L1T3%2F0%2BfMqnHCabCfLtLBaaE9uLWDqeJ6A%2Bh4m7150yVeTtSLCYrivp2OYBuEYBvW3u3YhIsLtvthhS8PNjwh7jlIHowSZW%2BV%2BJkLiaORd4fUQZPxT7EXEEm34gxpIJiWGArHj1OI52IjBnq%2BMoagVCLUOFhcGZKeOk8uQCB0wixTN1YJ3p0YGCQoaewxm0MMklYYnkdV9lvmMHxzIMS4m%2FRcPZjsWykhMuS0Hkts%2FshIHY%2BLKI8D3jChsSrFDXNniTgUtWEr5jSI2ofwDy%2FtESJRtBu9Di4aWXWtH%2Bx8iQvLQrzwqX3tYemWAQSU0IXtG1CgOZg1bLDK70UIXt21SXgYIDA95Ln2j0I3B%2F0lLW%2BUavACmu1e6%2F2k1r%2FjZcR6FN%2F8gDWe2szSFaPjUhzfGI2gTtwcAndG0uFWBs0iyL4TI1cUkOpB%2FAaR6FNHsdTjaMZlyW41M%2FyETID80w4vG%2BGDaH623pRpuG%2B4Sve%2Fdlc%2BgiurLAbaPXqq57%2BFd3d9q37q2Nu0B%2F98gl6fU%2FUub13x0GAAA%3D)
[...]

PR #10957, opened 2025-07-29
expand if statements before compression to allow iife inlining

gkgoat1:
**Description:**

Temporarily undoes if statement compression to allow IIFE inlining of branches. If statements are still recompressed after that occurs.

Solves scenario 2 in #10280 , where an iife can be inlined, but is nested within a ternary expression. By expanding them to if statements, the iife can be inlined before recompressing the statement to another ternary expression.

#10280


gkgoat1:
Converting to draft due to [potential performance issues](https://github.com/swc-project/swc/issues/10280#issuecomment-3134698808)

PR #10958, opened 2025-07-30
Introduce wasmtime plugin backend 

quininer:
**Description:**

This PR introduces wasmtime as an alternative plugin runtime backend for wasmer.

Currently, we simply copy test. maybe it would be better to use the same test for multiple backends in a separate test crate.

It depends https://github.com/swc-project/swc/pull/10943

**BREAKING CHANGE:**


crates/swc_plugin_backend_tests/tests/issues.rs
quininer, line 1: I didn't move this test to wasmtime because it appears to be a regression test that has nothing to do with wasm runtime.

PR #10967, opened 2025-07-31
pref(es/minifier): less clone atom

bvanjoi:


CPunisher:
Is it correct if two `Id` have conflicted `HashId`? Under the circumstance, for example, will `self.data.vars.get(&name.hashed_id())` return the usage data of other `Id`'s?

kdy1:
It’s possible and I think this approach is wrong

PR #10973, opened 2025-08-04
feat(es/ast): Add node id to `Ident`

bvanjoi:


Austaras:
Intersting. But why?

crates/swc_common/src/node_id.rs
```diff
@@ -0,0 +1,15 @@
...
+#[cfg_attr(
+    any(feature = "rkyv-impl"),
+    derive(rkyv::Archive, rkyv::Serialize, rkyv::Deserialize)
+)]
+#[cfg_attr(feature = "rkyv-impl", derive(bytecheck::CheckBytes))]
+#[cfg_attr(feature = "rkyv-impl", repr(C))]
+#[cfg_attr(feature = "shrink-to-fit", derive(shrink_to_fit::ShrinkToFit))]
+pub struct NodeId(#[cfg_attr(feature = "__rkyv", rkyv(omit_bounds))] u32);
+
+impl NodeId {
```
kdy1, line 13: Can we implement `Take` and `Default` using `NodeId::DUMMY`?
I think Default::default() being not dummy can be confusing

kdy1:
I used `dev/rust` because this is a Wasm-breaking change

mischnic:
What’s the usecase here?
And why only identifier and not every node type? 
In particular: what benefit does the nodeid have over the `atom, context`  tuple which is also unique for the variable?

kdy1:
@mischnic Maybe https://github.com/swc-project/swc/pull/10998#pullrequestreview-3105951361?

I assumed that he's going to use sequential node IDs to reimplement variable management, but it seems like he is trying to have a faster version of Id.

PR #10985, opened 2025-08-07
fix(es/module): Support optional chaining in `import.meta`

### LINKED ISSUE #10988: es/transpiler: optional chain on `import.meta` results in `import.meta` not getting transformed despite `preserveImportMeta: false`
When transpiling esm code that has `import.meta?.url`, it does not get transformed in any way, despite `import.meta.url` being transformed. Expected behavior is that it should be transformed in some way since `import.meta` is not supported in commonjs environments.

PeterCardenas:
**Description:**

`import.meta?.url` is not currently supported so it will not get transpiled. my thought is that it should get transpiled, so we are treating it the same as if the optional chain did not exist.

Example of this in `@cesium/engine` here: https://github.com/CesiumGS/cesium/blob/52f55d0e7063115b47b12fee20bfc1c291edcddb/packages/engine/Source/Core/buildModuleUrl.js#L44

closes #10988

### DIFF
--- a/.changeset/ninety-cobras-stare.md
+++ b/.changeset/ninety-cobras-stare.md
@@ -1,0 +1,5 @@
+---
+swc_ecma_transforms_module: patch
+---
+
+fix(es/module): Support optional chaining in `import.meta`
--- a/crates/swc_ecma_transforms_module/src/amd.rs
+++ b/crates/swc_ecma_transforms_module/src/amd.rs
@@ -286,6 +286,8 @@ where
             }
             Expr::OptChain(OptChainExpr { base, .. }) if !self.config.preserve_import_meta => {
                 let OptChainBase::Member(MemberExpr { obj, prop, span }) = &mut **base else {
+                    println!("OptChainBase::Member {:?}", base.span());
+                    n.visit_mut_children_with(self);
                     return;
                 };
                 if obj
@@ -301,6 +303,7 @@ where
                         },
                         MemberProp::PrivateName(..) => return,
                     };
+                    self.found_import_meta = true;
 
                     match p {
                         "url" => {
@@ -347,6 +350,8 @@ where
                         }
                         _ => {}
                     }
+                } else {
+                    n.visit_mut_children_with(self);
                 }
             }
             Expr::Member(MemberExpr { span, obj, prop })
--- a/crates/swc_ecma_transforms_module/src/amd.rs
+++ b/crates/swc_ecma_transforms_module/src/amd.rs
@@ -284,76 +284,6 @@ where
                     self.support_arrow,
                 );
             }
-            Expr::OptChain(OptChainExpr { base, .. }) if !self.config.preserve_import_meta => {
-                let OptChainBase::Member(MemberExpr { obj, prop, span }) = &mut **base else {
-                    println!("OptChainBase::Member {:?}", base.span());
-                    n.visit_mut_children_with(self);
-                    return;
-                };
-                if obj
-                    .as_meta_prop()
-                    .map(|p| p.kind == MetaPropKind::ImportMeta)
-                    .unwrap_or_default()
-                {
-                    let p = match prop {
-                        MemberProp::Ident(IdentName { sym, .. }) => &**sym,
-                        MemberProp::Computed(ComputedPropName { expr, .. }) => match &**expr {
-                            Expr::Lit(Lit::Str(s)) => &s.value,
-                            _ => return,
-                        },
-                        MemberProp::PrivateName(..) => return,
-                    };
-                    self.found_import_meta = true;
-
-                    match p {
-                        "url" => {
-                            *n = amd_import_meta_url(*span, self.module());
-                        }
-                        "resolve" => {
-                            let require = quote_ident!(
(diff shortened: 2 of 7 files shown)


PeterCardenas:
i'm a noob at rust so this is just me copy pasting code/following auto complete so bear with me please 🙏

magic-akari:
I'm curious about the semantics here.

Unlike regular property access, `import.meta` is treated as a single unit in the JavaScript specification. You can see hints of this — for example, `import["meta"]` is not a valid expression. So when you write `import.meta?.url`, it's more like how `foo?.url` works. It means:

```js
import.meta ? import.meta.url : undefined
```

Or, to be more precise — as Babel outputs it:

```js
var _import$meta;
(_import$meta = import.meta) === null || _import$meta === void 0 ? void 0 : _import$meta.url;
```

The key point is that the optional check applies to the entire `import.meta` expression. There's no equivalent behavior in CommonJS. Also, this kind of usage allows `undefined` as a possible value, and code using it is expected to handle that case properly.

However, this PR currently just ignores the optional chaining operator. I'm not sure that's the intended behavior.

magic-akari:
According to the JavaScript specification, we can see that `import.meta` is either already an object, or it's created starting from `Object.create(null)`.

```js
ImportMeta : import . meta
1. Let module be GetActiveScriptOrModule().
2. Assert: module is a Source Text Module Record.
3. Let importMeta be module.[[ImportMeta]].
4. If importMeta is empty, then
   a. Set importMeta to OrdinaryObjectCreate(null).
   b. Let importMetaValues be HostGetImportMetaProperties(module).
   c. For each Record { [[Key]], [[Value]] } p of importMetaValues, do
      i. Perform ! CreateDataPropertyOrThrow(importMeta, p.[[Key]], p.[[Value]]).
   d. Perform HostFinalizeImportMeta(importMeta, module).
   e. Set module.[[ImportMeta]] to importMeta.
   f. Return importMeta.
5. Else,
   a. Assert: importMeta is an Object.
   b. Return importMeta.
```

So from a semantic perspective, `import.meta` is always an object — it can never be a *falsy* value.

In older runtimes that don't support `import.meta` at all, trying to use it would throw a syntax error anyway, since `import` is a reserved keyword.

That means `import.meta?.url`, while syntactically valid, is always redundant in practice.

PeterCardenas:
just to be specific here, null is an object type right? so it can be false. you're right about not preserving the semantics of the ?. operator but during transpilation to commonjs I don't think the semantics matter. More specifically because of the runtime error case as well import.meta does not mean anything in commonjs and we get rid of it during transpilation, so it doesn't make sense to keep it around just to have the ?. semantics

magic-akari:
> just to be specific here, null is an object type right? so it can be false.

 No, “Object” here doesn’t mean values for which `typeof` returns `"object"` in JavaScript. The ECMA spec defines a distinct Null type that’s mutually exclusive with the Object type — I should have linked to the spec.

- https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-meta-properties
- https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-object-type
- https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-ecmascript-language-types-null-type

---

> More specifically because of the runtime error case as well import.meta does not mean anything in commonjs and we get rid of it during transpilation, so it doesn't make sense to keep it around just to have the ?. semantics

No, the semantics are actually quite important—otherwise we have no foundation to start from, since the original code is ESM and we need to respect its ESM semantics. During transpilation, using `{ url: expr }?.url` or even just `{}?.url` better preserves those semantics.

PeterCardenas:
that's fair, by ignoring the `?.` operator we are assuming what it actually means. which is fine in most if not all cases, but ultimately it's up to the runtime to decide what it means. thanks for the explanation! i'll try to do one of your recommendations when I have time.

magic-akari:
For those finding this via search who share this specific interest:
It's also worth noting that SWC's current implementation cannot fully transform `import.meta` while guaranteeing semantic correctness. 

My preferred approach here is to cover common usage patterns where we can reliably infer their semantics. For example, we clearly understand the meaning and transformation method for `import.meta?.url`.

However, certain cases remain problematic – like `foo(import.meta)`. What would we even transform this to? `{}`? Consider if `foo` were implemented as:

```javascript
function foo(x) {
  return x === import.meta;
}
```

This would return incorrect results after transformation. We currently have no reliable way to handle such patterns.

magic-akari:
> that's fair, by ignoring the `?.` operator we are assuming what it actually means. which is fine in most if not all cases, but ultimately it's up to the runtime to decide what it means. thanks for the explanation! i'll try to do one of your recommendations when I have time.

Let's leave this PR as-is for now until we conclude this discussion.
- #10989

magic-akari (changes requested):
Let's keep the current transpilation approach. Can you add this test file?

```JavaScript
foo?.bar(import.meta.url)
```

PeterCardenas:
good catch, looks like I missed continuing to iterate. fixed in the latest commit

kdy1 (approved):

Issue #10978, opened 2025-08-06
Non-Unicode Escape sequence incorrectly generated in string literals


h-a-n-a:
### Describe the bug

Non-Unicode Escape sequence got incorrectly generated in string literals.

For the following examples, the string literals contain some escaped Unicode Escape sequence and these sequences are generated in the results incorrectly.

With escaped (four slashes) unicode characters:

Input: `"\"\\\\uD83D\\\\uDD25\""`

With escaped (two slashes) unicode characters:

Input: `"\"\\uD83D\\uDD25\""`

### Input code

```typescript
"\"\\\\uD83D\\\\uDD25\""
"\"\\uD83D\\uDD25\""
```

### Link to the code that reproduces this issue

https://play.swc.rs/?version=1.13.3&code=H4sIAAAAAAAAA1OKUYoBglIXC2MXCMPFyDRGSYlLCSQBFYYJAgC%2F5A3NLQAAAA%3D%3D&config=H4sIAAAAAAAAA2WPSwrDMAxE9zlF0LrbdtE75BDCVYKDf0gONITcvf72Q3fWmxlrdAzjCKsouI9HeqYhIAvxe05EdhfxmQiQsiiKdYhw6eoqWZrRCBV0VgUi8kKxpOTa7GC8F%2Br2xqx2et6%2FFypvA5PIrzFb0S2G%2FvDs2WJedZy1wtBqgPWPrQTacXEPVCvd4GPqBSJv5VPQMvVgZsP5AtJpLxEnAQAA

### Expected behavior

**For the first case:**

Expected: `"\"\\\\uD83D\\\\uDD25\""`

**For the second case:**

Expected: `"\"\\uD83D\\uDD25\""`

### Actual behavior

**For the first case:**

Actual: `'"\\\uD83D\\\uDD25"';` ⬅️ This was un-escaped into three slashes

**For the second case:**

Actual: `'"\uD83D\uDD25"';` ⬅️ This was un-escaped into a single slash

### Version
1.13.3

h-a-n-a:
I investigated a bit and turned out that for string literals like below, the `value` field is the same, even though the `raw` field are different from each result.

`"\\\uD808\\\uDFAB"`:
```
[crates/swc_ecma_codegen/src/lit.rs:46:9] self = Str {
    span: 1..19,
    value: "\\\\uD808\\\\uDFAB",
    raw: Some(
        "\"\\\\\\uD808\\\\\\uDFAB\"",
    ),
}
```

[...]

Issue #10353, opened 2025-04-10
SWC minifier un-escapes unicode characters


slorber:
### Describe the bug

For some code inputs containing escaped unicode emojis, SWC minifier will output unescaped unicode emojis.

Example input:

```js
console.log("\\"\\\\uD83D\\\\uDE42\\"");
```

### Input code

```typescript
import swc from '@swc/core';

const input = `
console.log("\\"\\\\uD83D\\\\uDE42\\"");
`;

console.log('input');
console.log(input);
eval(input);

console.log('\n');
const output = await swc.minify(input);

console.log('output');
console.log(output.code);
eval(output.code);



Runnable playground: https://stackblitz.com/edit/stackblitz-starters-gpln2upu

Output:


❯ node index.js

input
console.log("\"\\uD83D\\uDE42\"");
"\uD83D\uDE42"


output
console.log('"\uD83D\uDE42"');
"🙂"
```

### Config

```json

```

### Link to the code that reproduces this issue

https://play.swc.rs/?version=1.11.18&code=H4sIAAAAAAAAA1OKiSl1sTB2UQIAWPmpCAkAAAA%3D&config=H4sIAAAAAAAAA1VQuW7DMAzd8xWC0KkorCBjpgxdM3YqMrAyfcGSDJJGawT590qOY8uT%2Fch3ifeDUsYo%2Fca2QQf6rHQjMvDZmI6DL55jlkBYBKoN%2F1qy%2BiOqNIeRLF5h4KgSGnGedmwjvMffCAYgRlpxEk1e4C%2FFyDQgW2oHme3mbYk2EMSw3FKpx%2FOjhcBzFcjljj3WYKfPl3InzC2vKFCCwIswr%2Ff2QDVK6oZ8Op6OS6%2F4Cmk4j7y4UI49snmP02%2B9odsae7HBV229MFaQEbgBwnIhrOC2a%2FQDjF%2FUp0qF4eXw83IJ3U6dzpl4NjgXfMd6Y7a%2Brab1Lo%2FDP5Tvfnb0AQAA

### SWC Info output

After being minified, if the unicode chars are escaped, they should also be escaped in the output

`\\uD83D` is not equivalent to `\uD83D`

### Expected behavior

`\\uD83D` should remain `\\uD83D`

### Actual behavior

`\\uD83D` is unescaped to `\uD83D`

### Version

1.11.18

### Additional context

Prettier website bug: https://github.com/prettier/prettier/pull/17308

[...]

Issue #7678, opened 2023-07-19
Minify option incorrectly strips escape characters


SarahAnnTolsma:
### Describe the bug

When escaping specific strings, SWC strips necessary escape characters in minified output. This results in the incorrect string being output. See the given input code and playground link for examples.

### Input code

```typescript
let str = "\\uD83D\\uDC68\\u200D\\uD83D\\uDE80";

let obj = {
    "\\uD83D\\uDC68\\u200D\\uD83D\\uDE80": "wrong"
};
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2015",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": true,
  "isModule": false
}
```

### Playground link

https://play.swc.rs/?version=1.3.70&code=H4sIAAAAAAAAA8tJLVEoLilSsFVQiokpdbEwdgFRzmYWQMrIwMAFSdDVwkDJmosrB6gjPykLqKOaSwEIiNJnpaBUXpSfl67EVWsNAEudP25zAAAA&config=H4sIAAAAAAAAA1VPOw7DIAzdcwrkuUNbqUvv0EMg6kRE%2FIQdqSjK3QME2maz39deByFgJgVPseYxL0FGwvjdM0LJsfxkBDgFJBV1YLh0lqlQozSEFdoOBljGCbm4kO7X26M5wHhP2B0Ns9rpMf13Km9DRKKzsEilmwyeG4fWCta%2Fl0q2X8q95YKcZr2bCX7K3shxqeGg6dXdR%2FS2AzVy5M4cAQAA

### Expected behavior

SWC should not strip escape characters and produce the following output:

```javascript
let str="\\uD83D\\uDC68\\u200D\\uD83D\\uDE80";let obj={"\\uD83D\\uDC68\\u200D\\uD83D\\uDE80":"wrong"};
```

### Actual behavior

SWC strips escape characters and produces the following output:

```javascript
let str="\uD83D\uDC68\\u200D\uD83D\uDE80";let obj={"\uD83D\uDC68\\u200D\uD83D\uDE80":"wrong"};
```

Note the missing `\` before `\uD83D` (twice), `\uDC68` and `\uDE80`.

### Version

1.3.70

### Additional context

As a workaround, the following produces the correct output:

```typescript
let workaround = {
[...]

PR #10987, opened 2025-08-07
fix(es/ast): Fix unicode lone surrogates handling

h-a-n-a:
**Description:**

This PR fixed an issue related to lone surrogates handling in Rust.

This fix's credits all go to Oxc team https://github.com/swc-project/swc/issues/10978#issuecomment-3160553130. What I'm doing is porting the fix that was made in Oxc and make it working under SWC.

### Problem:

The problem is related to the fundamental difference between how Rust and JavaScript handle Unicode, especially lone surrogates.

**JavaScript's Unicode Model**

```javascript
// JavaScript allows this - lone surrogates are stored in UTF-16
let str = "\uD800";  // High surrogate alone - technically invalid Unicode
let obj = { "\uD800": "value" };  // Works fine in JS
```
JavaScript uses UTF-16 internally and tolerates invalid Unicode sequences:
- Strings are UTF-16 code unit sequences, not Unicode scalar sequences
- Lone surrogates (U+D800-U+DFFF) are allowed and preserved
- No validation that surrogates come in proper high/low pairs
- Engine just stores the raw UTF-16 code units

**Rust's Unicode Model**

```rust
// This CANNOT exist in Rust:
let s = "\u{D800}";  // ❌ COMPILE ERROR - not a valid Unicode scalar
let c: char = '\u{D800}';  // ❌ COMPILE ERROR - char excludes surrogates
```
Rust enforces strict Unicode validity:
- String is UTF-8 and must contain valid Unicode scalar values
- char represents Unicode scalar values (U+0000-U+D7FF, U+E000-U+10FFFF)
- Surrogate code points (U+D800-U+DFFF) are explicitly excluded
- No way to represent lone surrogates in Rust's standard string types

###  Key Changes:
1. AST Structure: Added `lone_surrogates: bool` field to `Str` and `TplElement` structs to track when strings contain lone surrogates
2. Encoding Strategy: Lone surrogates are encoded using \u{FFFD} (replacement character) followed by the original hex digits for internal representation
3. Code Generation: Modified string output to properly escape lone surrogates back to \uXXXX format during codegen
4. Test: Also fixed some cases related to member expression optimizations and string concatenation optimizations

### TODOs:

1. Add support for serializing and deserializing literals with lone surrogates in `swc_estree_compat` 
2. Reflect AST changes in `binding` crates

### Breaking changes:

Breaks the AST by adding `lone_surrogates` field to `Str` and `TplElement` and breaks the `value` and `cooked` respectly in `Str` and `TplElement`. Both of the field is using `\u{FFFD}` (Replacement Character) as an escape if `lone_surrogates` set to `true`. 

To consume the real value, you need to first check if `lone_surrogates` is `true`, then unescape it by removing the char and construct it with the four trailing hexs(from `\u{FFFD}D800` to `\uD800`).

closes https://github.com/swc-project/swc/issues/10978
closes https://github.com/swc-project/swc/issues/10353

Fixed a regression of https://github.com/swc-project/swc/issues/7678


kdy1:
Thank you so much! Acutally I tried to fix this several times but it was very confusing :(

h-a-n-a:
@kdy1, Hi!
~I was quite confused about the CI failure in https://github.com/swc-project/swc/actions/runs/17063469624/job/48375172890?pr=10987 and I've also tried running the test locally on `main` but I have no luck. Would you please shed some light on this?~

~In addition to this, I've added `lone_surrogate` field to `StringLiteral` as a fix suggested in the original issue and we should probably also get `TplElement` fixed as well. But this would very likely to add some performance regression. The use case is that the lone surrogates in `TplElement` gets transformed to `StringLiteral` if the target is set to `<=es5`.
Failed case: https://github.com/swc-project/swc/actions/runs/17063469624/job/48375172821?pr=10987~

EDIT: Added support for `TemplateLiteral`

kdy1:
Do we really need to change AST?

kdy1:
Actually changing AST is not allowed in our case because for v2 we are going to aligh the AST with babel or typescript-eslint

h-a-n-a:
> Actually changing AST is not allowed in our case because for v2 we are going to aligh the AST with babel or typescript-eslint

Hi! @kdy1 
It's related to the fundamental difference between Rust string and JavaScript string. Without adding the custom escape sequence and the `lone_surrogates` sign, it's pretty hard to distinguish whether it's a lone surrogates or it's a raw string.

For example, for `\uD800`, you cannot escape it to `\\uD800` as it changes the semantic of the string. So I introduced `lone_surrogates` sign to indicate that we've used `\u{FFFD}` to escape the lone surrogates. Then we can convert the lone surrogates into valid code output and do the following string optimizations. Without this, I believe we cannot. Otherwise, maybe we could change the Atom type to support `UTF-16` characters.

> babel or typescript-eslint

We could probably just change the way it serialize and deserialize to align the AST with them? Adding another layer seems necessary to me as we know there's some difference between Rust and JavaScript that I mentioned above.

Boshen:
I'd appreciate a reference to Oxc because I assume the code is based on my comment https://github.com/swc-project/swc/issues/10978#issuecomment-3160553130, it took us a lot of time to understand the problem and then make the right fix.

> Actually changing AST is not allowed in our case because for v2 we are going to aligh the AST with babel or typescript-eslint

You probably want to bend the rule here, because the `String` type we use in Rust is UTF8, you need some extra information to make it UTF16.

h-a-n-a:
> I'd appreciate a reference to Oxc because I assume the code is based on my comment [#10978 (comment)](https://github.com/swc-project/swc/issues/10978#issuecomment-3160553130), it took us a lot of time to understand the problem and then make the right fix.

Will do! Really appreciate the job y'all have done ;-)

kdy1:
@h-a-n-a @Boshen  I see, thank you so much for the hard work and explanation! I think it might be possible to conform to their AST even with some changes... I'll review it later.

h-a-n-a:
@kdy1 Copilot seems broken with this huge code change ;-(.

Cargo.lock
```diff
@@ -5613,6 +5613,7 @@ dependencies = [
  "swc_atoms",
  "swc_common",
  "swc_ecma_ast",
+ "swc_ecma_utils",
```
kdy1: This will prevent publishing crates. Please remove cyclic dependencies

kdy1 (changes requested):
`cargo` denies publising if there's a cycle

Cargo.lock
h-a-n-a: I've hoisted `pair_to_code_point` to the `swc_ecma_lexer` level.

kdy1 (commented):
You deleted `./crates/swc_ecma_transforms_proposal/tests/decorator-tests`

Please restore it

crates/swc_ecma_transforms_proposal/tests/decorator-tests
kdy1, line 1: This one

kdy1:
The submodule change is the last request 👍.
Thank you so much for the hard work.

image

cc @hardfist to align the time to publish this, as this is a Wasm-ABI breaking change. I'll also talk with the next.js team

hardfist:
we can merge this breaking in Rspack 1.6 https://github.com/web-infra-dev/rspack/discussions/11554, I'm not sure whether we can make it compatible with codes that don't have lone surrogates

crates/swc_ecma_transforms_proposal/tests/decorator-tests
h-a-n-a: Fixed. Didn't know why I would ever delete that 🥲

kdy1 (approved):

h-a-n-a:
Thanks for reviewing this! ❤️

kdy1:
I'll revert this PR as requested by @CPunisher. Can you rebase this PR and resend to `dev/rust` branch?

kdy1:
I'll do it by myself

PR #10998, opened 2025-08-11
refactor(es/minifier): use node id

bvanjoi:


bvanjoi:
> If so, I think the field name is wrong.

I don't understand what this statement refers to. Could you please provide a detailed explanation?

Additionally, I should note that this approach is still under exploration, I can't draw definitive conclusions until testing is complete.

kdy1:
Okay, I'll wait for the result

PR #11000, opened 2025-08-12
perf(es/parser): remove `OneDirectionalList` and reduce allocation of `Vec`

CPunisher:
**Description:**

1. Separate `CommentsBuffer` into legacy and new ones with `CommentsBufferTrait`.
2. For the new one, remove `OneDirectionalList` and use `Vec`. Record positions in `LexerCheckpoint` for recovery.

There's regression for legacy lexer. I'm not sure the reason.


CPunisher:
😅 I will investigate the reasons

kdy1 (dismissed):
Can you rebase?

Issue #10994, opened 2025-08-08
`@swc/html` minifier option to disable removal of `<head>` and `</body>`


slorber:
### Describe the feature

The SWC html minifier removes the `<head>` and `</body>` closing tag attributes by default, as an optimization.

The following may be surprising, but it passes W3C validation, so SWC is technically right:

```html
<!doctype html>
<html lang=en>
<meta name=generator content="Docusaurus v3.7.0">
<title>Docusaurus changelog | Docusaurus</title>
<meta name=viewport content="width=device-width, initial-scale=1.0">
<body>
blabla
```

But it turns out that despite being valid HTML, this can cause troubles, notably for LinkedIn social cards. Apparently, the LinkedIn crawler expects those tags to be present to be able to find the `og:image` meta tag, and maybe this affects other sites as well.

For Docusaurus, we'd like to have an option in SWC minimizer to disable this built-in optimization. @kdy1 [suggested opening a feature request here](https://github.com/facebook/docusaurus/issues/10556#issuecomment-3159807304).

### Additional context

See also related Docusaurus discussions:
- https://github.com/facebook/docusaurus/pull/10554#issuecomment-2775197012
- https://github.com/facebook/docusaurus/issues/10556#issuecomment-3152116127

slorber:
cc @ramnivas who noticed the issue on his Docusaurus site

CPunisher:
I find that you can set `tagOmission` to `false` to keep all tag pairs. I'm not sure if we should add another option just for keep the tags of `head` and `body`.

alexander-akait:
We can improve it by `tagOmission: boolean | "keep-head-and-body"`

CPunisher:
> We can improve it by `tagOmission: boolean | "keep-head-and-body"`

🤔 I'm a bit worried about whether someone would like to prevent some other tags being omitted.

CPunisher:
@slorber Will this meet your expectation ?

<img width="1260" height="484" alt="Image" src="https://github.com/user-attachments/assets/ae9f4e92-d130-48fd-af5f-97ef5c79444a" />

slorber:
I guess it could yes, but I haven't tested much against the behavior of the LinkedIn crawler. Maybe removing the last `</html>` element could also cause troubles. I'm on holiday now so will check next week.

ramnivas:
I tried these options (added og:image to the html scraped from the images). With LinkedIn's post inspector, here are the results:

tagOmission: true -> didn't work
[...]

PR #11008, opened 2025-08-18
feat(html/minifier): Support `keep_head_and_body`

CPunisher:
**Description:**

We should add the option to binding and close that issue after publishing new version and upgrade the `swc_html` in node bindings,

related https://github.com/swc-project/swc/issues/10994


kdy1:
I'm publishing crates with https://github.com/swc-project/swc/commit/c659ccc0f767dc3d461a20f3e83d8681202bb52f

Issue #11034, opened 2025-08-27
es/minifier: const is transpiled to let


CPunisher:
### Describe the bug

If a const variable is reassigned in JavaScript, `Uncaught TypeError: Assignment to constant variable.` is thrown. But in some cases swc minifier transpiles it to `let`, and the runtime behavior is changed.

### Input code

```typescript
const foo = window.e ? 'bar' : 'baz'
foo = `it's not allowed`

console.log(`foo=${foo}`)
```

### Link to the code that reproduces this issue

[...]

PR #11035, opened 2025-08-28
fix(es/minifier): Avoid `const_to_let` if reassigned

CPunisher:
**Description:**

I move the process of `const_to_let` from `postcompress` to `Optimizer` because varinfo is need to check if the variable is reassigned.

I'm also curious why `const_to_let` is disabled in https://github.com/swc-project/swc/pull/11035/files#diff-dd713fdbfb1b9c77d20d446319255c3d41a169213b8825f43aacc0ee04f4c90f

fixes https://github.com/swc-project/swc/issues/11034

### DIFF
--- a/.changeset/famous-files-kneel.md
+++ b/.changeset/famous-files-kneel.md
@@ -1,0 +1,6 @@
+---
+swc_ecma_minifier: patch
+swc_core: patch
+---
+
+fix(es/minifier): avoid `const_to_let` if the const variable is reassigned


magic-akari:
I expected this PR to introduce a performance regression, since it adds an extra check. 
However, the benchmarks show no change.

CPunisher:
> I expected this PR to introduce a performance regression, since it adds an extra check. However, the benchmarks show no change.

Yes, it definitely makes minifier a bit slower. But I think the minifier does much more things so it's hard to inspect the change compared to the whole time cost.

bindings/binding_core_wasm/__tests__/simple.js
```diff
@@ -197,7 +197,7 @@ describe("minify", () => {
 
         expect(output).toMatchInlineSnapshot(`
             {
-              "code": "let somename=1;console.log(1);",
+              "code": "const somename=1;console.log(1);",
```
kdy1, line 200: Why is this changed?

kdy1:
Is this PR ready for a review?

bindings/binding_core_wasm/__tests__/simple.js
CPunisher: I'm considering why this test case should output `let` instead of `const`. It sets the `module` to `false` so the top level `const` in `Program::Script` should not be transformed into `let`.

kdy1: My bad, sorry

CPunisher: Actually after some changes it becomes `let` again...So I'm still investigating

PR #11069, opened 2025-09-07
feat(es/ast): Add backward compatibility to AST bytes

kdy1:
**Description:**

Things to note:

 - This will make performance noticeably worse.
 - This will make the API for Rust side users pretty bad.

I think we may need to use a proc macro for this, to copy the whole AST types into new types with the `extras: UnknownBytes<TYPE_NAME, PADDING_SIZE>` attribute.


kdy1:
Closing in favor of https://github.com/swc-project/swc/pull/11100

PR #11070, opened 2025-09-08
perf(es/transforms): split `paren_remover` and `fixer`

CPunisher:
**Description:**

I suppose `paren_remover` can be simpler but I'm not sure if this is correct


kdy1:
This will have a bad impact on the binary size

kdy1:
Can you patch locally to check if it has any impact on the rspack binary size?

CPunisher:
After: 38.3624267578 MB
Before: 38.3307952881 MB
+32.39 KB

kdy1:
Do you think it’s fine?

CPunisher:
> Do you think it’s fine?

I think that's totally fine. However, according to my local benches, there's no noticeable performance improvement. So I will close this pr.

PR #11100, opened 2025-09-18
refactor(ast): Introduce flexible serialization encoding for AST

quininer:
**Description:**

Following our previous discussion, I am trying to introduce a self-describing serialization encoding for ast to replace https://github.com/swc-project/swc/pull/11069 .

Since `serde` is currently used for json serialization, it is hard to modify serde to support both json and binary encodings (due to the `untagged` and `flatten` attr). So I made our own `Encode` and `Decode` proc macro, which also facilitate unknown variant support.

benchmark

I've done some simple benchmarks with a modified swc ast, and I believe we don't have an unacceptable performance loss relative to `rkyv`.
I'm using `cbor4ii` here, which I developed, so I'd prefer it if there's no performance difference. but it's easy to migrate to `bincode` or other scheme.

This also introduces an `unknown` feature for `ecma_ast` crate, which should only be enabled by the wasm plugin to ensure that old plugins can decode new AST.

**BREAKING CHANGE:**

This PR itself only introduces new encoding and will not cause any breakage. However, it may cause breakage when switch the wasm plugin communication scheme in future.

### DIFF
--- a/.changeset/chatty-brooms-drop.md
+++ b/.changeset/chatty-brooms-drop.md
@@ -1,0 +1,7 @@
+---
+ast_node: major
+from_variant: major
+swc_atoms: major
+---
+
+refactor(ast): Introduce flexible serialization encoding for ast


quininer:
This also requires a serialization test, include unknown variants.

Since I'm busy with other work, it will be a little slower.

kdy1:
Some CI failures are not fixable at the moment. https://github.com/swc-project/swc/pull/11197

crates/jsdoc/src/ast.rs
```diff
@@ -551,6 +607,10 @@ pub struct FileTag {
 #[ast_node]
 pub struct FunctionTag {
     pub span: Span,
+    #[cfg_attr(
+        feature = "encoding-impl",
+        encoding(with = "cbor4ii::core::types::Maybe")
```
kdy1, line 612: Do we really need this kind of annotations, for all of the `Option<T>` fields?

crates/ast_node/src/encoding/encode.rs
```diff
@@ -0,0 +1,235 @@
...
+pub fn expand(DeriveInput { ident, data, .. }: DeriveInput) -> syn::ItemImpl {
+    match data {
+        Data::Struct(data) => {
+            let fields = data
+                .fields
+                .iter()
+                .enumerate()
+                .map(|(idx, field)| -> syn::Stmt {
+                    let fieldpath: syn::ExprField = match field.ident.as_ref() {
+                        Some(name) => syn::parse_quote!(self.#name),
```
kdy1, line 14: I didn't get it. Can you add some comment what does this code do?
Is it specifying the `name` or `idx` to _describe_ the schema?

crates/jsdoc/src/ast.rs
quininer: Yes, because the current encode of `Option` doesn't meet our needs. I noticed that `Option` now also has `serde(default)`, so I think this is acceptable.

crates/ast_node/src/encoding/encode.rs
quininer: This will generate code like `self.0.encode(writer)` for `struct Bar(String)`.

kdy1: I see, thanks!

crates/ast_node/src/encoding/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+enum EnumType {
+    Unit,
+    One,
+    Struct,
+}
+
+pub(crate) fn is_unknown(attrs: &[syn::Attribute]) -> bool {
+    attrs
+        .iter()
+        .filter(|attr| attr.path().is_ident("encoding"))
```
kdy1, line 14: Can this function detect attributes like

```rust
    #[cfg_attr(
        feature = "encoding-impl",
        encoding(with = "cbor4ii::core::types::Maybe")
    )]
```
?

kdy1: Is it not a requirement?

crates/ast_node/src/encoding/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+            })
+            .unwrap();
+            is_unknown
+        })
+}
+
+fn is_with(attrs: &[syn::Attribute]) -> Option<syn::Path> {
+    attrs
+        .iter()
+        .filter(|attr| attr.path().is_ident("encoding"))
```
kdy1, line 29: Same here

crates/swc_common/src/lib.rs
```diff
@@ -85,3 +87,9 @@ pub use self::syntax_pos::{
     ArchivedBytePos, ArchivedCharPos, ArchivedFileName, ArchivedMultiSpan, ArchivedSourceFile,
     ArchivedSourceFileAndBytePos, ArchivedSpan, ArchivedSpanLinesError, ArchivedSpanSnippetError,
 };
+
+#[cfg(swc_ast_unknown)]
+#[track_caller]
+pub fn unknown() -> ! {
```
kdy1: We cannot use `!` type because of the MSRV

crates/ast_node/src/encoding/mod.rs
quininer: Okay, I'll add `encoding(none)`

crates/swc_common/src/lib.rs
quininer: Okay, I'll change it to a macro.

kdy1: You can use Infailible

crates/ast_node/src/encoding/mod.rs
kdy1: I'm not sure if it's required. Is detection via this function (in the proc-macro) important for the operation?

crates/jsdoc/src/ast.rs
kdy1: Makes sense. Thank you for the explanation!

crates/ast_node/src/encoding/mod.rs
quininer: Do you want to automatically check the `Option` in a macro? I think that's hard because macros are not type-aware.

I think explicit attributes are simpler and more reliable.

kdy1: I just wanted to check if it’s an intended behavior because it seems like this function will detect direct encoding attributes but not cfg_attr

crates/swc_common/src/lib.rs
quininer: I have replaced it with Infallible and macro

crates/swc_css_parser/tests/fixture/function/unknown/output.json
```diff
@@ -179,11 +179,13 @@
                     },
                     "token": {
                       "Dimension": {
-                        "value": 100.0,
-                        "raw_value": "100",
-                        "unit": "px",
-                        "type": "integer",
-                        "raw_unit": "px"
+                        "dimension": {
```
kdy1, line 182: Is this expected change?

quininer: This was originally intended to simplify the `encoding` macro, as it is only tuple variant in named enum. I can change it back.

PR #11101, opened 2025-09-18
Failing test for incomplete inlined function calls

eps1lon:
Failing test for https://github.com/vercel/next.js/issues/83925

### DIFF
--- a/crates/swc_ecma_minifier/tests/fixture/issues/react/hooks/6/input.js
+++ b/crates/swc_ecma_minifier/tests/fixture/issues/react/hooks/6/input.js
@@ -1,0 +1,16 @@
+const decideZoomByAccuracy = (range) => {
+    const isUnder = (accuracy) => {
+        return range <= accuracy;
+    };
+    if (isUnder(0)) {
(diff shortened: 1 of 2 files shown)


crates/swc_ecma_minifier/tests/fixture/issues/react/hooks/6/output.js
```diff
@@ -0,0 +1 @@
+export const zoom = 15;
```
eps1lon, line 1: SWC outputs `export const zoom = range <= 100 ? 15 : 11;` instead which triggers a `ReferenceError` since `range` is no longer defined.

crates/swc_ecma_minifier/tests/fixture/issues/react/hooks/6/input.js
eps1lon, line 1: It started out hook related but ultimately wasn't. I can't get this test to run locally if I put it in another folder so I left it here.

eps1lon:
Now passes after https://github.com/swc-project/swc/pull/11106

PR #11115, opened 2025-09-22
refactor(ast_node): Make AST enums `non_exhaustive`

quininer:
**Description:**

This is the split of https://github.com/swc-project/swc/pull/11100 . Since we need to add unknown variants to the ast enum, this requires a lot of code changes to handle non-exhaustive match. I'm doing this in a separate PR for review.

**BREAKING CHANGE:**

### DIFF
--- a/.changeset/smart-rats-smell.md
+++ b/.changeset/smart-rats-smell.md
@@ -1,0 +1,5 @@
+---
+ast_node: major
+---
+
+refactor(ast): Make ast enum `non_exhaustive`


kdy1:
Do you think this should go into the main branch?
Can you change the target to ‘dev/rust’?

quininer:
Sure, I can change it.

quininer:
I've switched to using a cfg flag instead of a feature to handle unknown variants, which makes it easier to ensure consistent handle.

Also, since all changes are behind the `swc_ast_unknown` cfg flag, it's not a breaking change. I think it can be safely merged into main branch.

.github/workflows/CI.yml
```diff
@@ -354,6 +354,11 @@ jobs:
         run: |
           ./scripts/github/run-cargo-hack.sh ${{ matrix.settings.crate }}
 
+      - name: Check unknown variant with transforms
```
quininer, line 357: I only added `transforms` ci since that covers most of sublibrary and is the primary use case for plugin.

crates/swc_ecma_lexer/src/common/parser/expr.rs
```diff
@@ -1962,6 +1974,8 @@ fn parse_args_or_pats_inner<'a, P: Parser<'a>>(
                     // creating `Invalid`, we don't have to emit a new
                     // error.
                 }
+                #[cfg(swc_ast_unknown)]
+                _ => (),
```
quininer, line 1978: I've ignored unknown in some places. there's no particular reason for this, maybe panick would be better.

kdy1: This is a breaking chnage regardless

crates/swc_ecma_ast/src/typescript.rs
```diff
@@ -410,6 +410,7 @@ pub struct TsKeywordType {
 #[cfg_attr(feature = "rkyv-impl", derive(bytecheck::CheckBytes))]
 #[cfg_attr(feature = "rkyv-impl", repr(u32))]
 #[cfg_attr(feature = "serde-impl", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(swc_ast_unknown, non_exhaustive)]
```
kdy1, line 413: Addition of this means `ast_node` crate has a breaking change

quininer: This will not cause any break unless the user configures `swc_ast_unknown` cfg.

crates/ast_node/src/lib.rs
```diff
@@ -158,29 +158,34 @@ pub fn ast_node(
...
+            let attrs = <syn::punctuated::Punctuated<syn::Ident, syn::Token![,]>>::parse_terminated
+                .parse(args)
+                .expect("failed to parse #[ast_node]");
+
+            let mut has_no_clone = false;
+            let mut has_no_unknown = false;
+            for attr in &attrs {
+                if attr == "no_clone" {
+                    has_no_clone = true;
+                } else if attr == "no_unknown" {
```
quininer, line 172: Maybe an `unknown` flag would be better than a `no_unknown` flag? This would keep the original behavior of `ast_node` macro.

crates/swc_bundler/src/bundler/chunk/cjs.rs
```diff
@@ -140,6 +140,8 @@ fn wrap_module(
                             unreachable!("module item found but is_es6 is false: {:?}", i)
                         }
                         ModuleItem::Stmt(s) => s,
+                        #[cfg(swc_ast_unknown)]
```
kdy1, line 143: I think usage of `cfg` like `cfg(swc_ast_unknown)` is really a bad idea for crates like SWC, which is used across various projects by multiple vendors. Can you remove this `cfg` flag completely?

This kind of `cfg` works well only for s single vendor where the defining side and using side is the same team. Otherwise this would require lots of _fix PRs_ for `cfg` flags. (And thus blocking users of SWC from iterating many times)

What was the reason you had to add this flag?

quininer: For reasons I describe [here](https://github.com/swc-project/swc/discussions/11121#:~:text=unknown%20variants%20will%20only%20be%20generated%20in%20swc%20plugin), I prefer to only enable the unknown variant in plugins, ensure no performance impact when no plugins are present.

This also avoids unnecessary breaking changes for use cases outside of plugins.

kdy1: Can you try local patching and building the bindings and https://github.com/swc-project/plugins with this? You will get what I meant if you try it.

You can use [`patch-section.sh`](https://github.com/swc-project/swc/blob/dfe96701beaf9096848ac696b18cc3cf27edfeb4/scripts/cargo/patch-section.sh) to get the patch section, and you can copy-paste it in the `[patch.crates-io]` section of `Cargo.toml`

kdy1:
I updated your branch to use patch commands

crates/swc_bundler/src/bundler/chunk/cjs.rs
kdy1: My concern was that every crate (except the final crate) using SWC would need to do 

```rust
#[cfg(swc_ast_unknown)]
_ => panic!("unable to access unknown nodes"),`
```

스크린샷 2025-10-10 오후 12 22 02


in nearly everywhere, and it may feel too cumbersome for the SWC users. 


Otherwise, a user won't be able to use it without upstream fixes. I attempted to demonstrate it via https://github.com/swc-project/plugins/pull/536 (the official plugins depend on some core transforms), but I found that you had already added the relevant `cfg` and `_ => panic!()` to the core transforms in the PR.

The problem is that, by default, **the library crate seems to work** without `#[cfg(swc_ast_unknown)]`, but actually it's not usable for the end-plugin authors. IMHO, public API should never introduce such cases.

quininer: This handle is optional; users can compile the plugin without the swc_ast_unknown cfg, which eliminates the need for this handling (but no forward compatibility).

This issue is also unrelated to the cfg; if remove cfg, all users must perform the `_ => panic!()` handle. The core issue here is handle unknown variants, not the presence or absence of a cfg. instead, the cfg is used to mitigate this issue.

.changeset/smart-rats-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+ast_node: major
```
kdy1, line 2: I added `ast_node: major` because I don't want to see complains about breaking changes published wrongly

Issue #11105, opened 2025-09-19
minifier: `var __proto__ = []; console.log({ __proto__ })` is wrongly compressed to `console.log({ __proto__: [] })`


sapphi-red:
### Describe the bug

`var __proto__ = []; console.log({ __proto__ })` is wrongly compressed to `console.log({ __proto__: [] })`.
This changes the behavior.
```js
var __proto__ = []; console.log({ __proto__ } instanceof Array) // false
```
```js
console.log({ __proto__: [] } instanceof Array); // true
```

### Input code

```typescript
var __proto__ = []; console.log({ __proto__ })
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": true,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true,
        "const_to_let": true,
        "pristine_globals": true
      }
    },
    "externalHelpers": false,
    "target": "es2024",
    "preserveAllComments": true
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true,
}
```

### Link to the code that reproduces this issue

[...]

PR #11123, opened 2025-09-25
fix(es/minifier): Preserve `__proto__` shorthand property behavior

magic-akari:
**Description:**

When minifying object shorthand properties, the `__proto__` property should be converted to a computed property `["__proto__"]` instead of a regular key-value pair to preserve the original object behavior. This prevents changing the object's prototype chain during minification.

**BREAKING CHANGE:**

- Closes: #11105

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/optimize/util.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/util.rs
@@ -590,10 +590,7 @@ impl VisitMut for NormalMultiReplacer<'_> {
                 } else {
                     PropName::Ident(IdentName::new(i.sym.clone(), i.span))
                 };
-                *p = Prop::KeyValue(KeyValueProp {
-                    key,
-                    value,
-                });
+                *p = Prop::KeyValue(KeyValueProp { key, value });
             }
         }
     }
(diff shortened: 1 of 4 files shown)


crates/swc_ecma_minifier/src/compress/optimize/util.rs
```diff
@@ -469,8 +469,20 @@ impl VisitMut for Finalizer<'_> {
...
+                    PropName::Computed(ComputedPropName {
+                        span: i.span,
+                        expr: Box::new(Expr::Lit(Lit::Str(Str {
+                            span: i.span,
+                            value: i.sym.clone(),
+                            raw: None,
+                        }))),
+                    })
+                } else {
+                    i.take().into()
```
CPunisher: No need to take the prop name and set it again. I think it's better to write everything in `if i.sym == "__proto__"` body and avoid unnecessary  `self.changed = true`

CPunisher: Sorry I make a mistake here

CPunisher (approved):

CPunisher (approved):

Austaras (approved):

Issue #11124, opened 2025-09-25
strip_types fails with void arrow function


robpalme:
### Describe the bug

This arrow function works in `tsc` and `ts-blank-space` but fails in SWC.

### Input code

```typescript
var t = (): (void) => {}
```

### Config

```json
strip_types: true
```

### Link to the code that reproduces this issue

https://play.swc.rs/?version=1.13.19&code=H4sIAAAAAAAAAytLLFIoUbBV0NC0UtAoy89M0VSwtVOorgUAw17kYRgAAAA%3D&config=H4sIAAAAAAAAA1VPOw7DIAzdOQXy3KFi6NA79BCIOhERAYQdqSjK3QsJpM1mv4%2Ff8yqkhIkMPOVaxrJEnQjTuReEsmf9KQhwjkgm2chw6yxTpQbtCHdoOxhgnUbk6kJSd6WaA1wIhN3RsNl6O%2BT%2FTBPmmJDoKqxS7UeH10TRUmEO72Un2y%2B179HgAT9RDzsPg6VXd3JaUGxfBMLf3xcBAAA%3D&strip-types=

### Expected behavior

Emits:

```ts
var t = ()         => { };
```

### Actual behavior

An error:

```
InvalidSyntax

Parenthesized expression cannot be empty
```

### Version

1.13.19

### Additional context

Found in:  https://github.com/nodejs/node/issues/60007

Relayed by @legendecas

PR #11125, opened 2025-09-26
fix(es/parser): Parse `(void)` correctly as arrow function return type

magic-akari:
**Description:**

**BREAKING CHANGE:**

- Closes: #11124

### DIFF
--- a/.changeset/gorgeous-masks-enjoy.md
+++ b/.changeset/gorgeous-masks-enjoy.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_lexer: patch
+---
+
+fix(es/parser): Parse `(void)` correctly as arrow function return type
--- a/crates/swc_ecma_lexer/src/common/parser/typescript.rs
+++ b/crates/swc_ecma_lexer/src/common/parser/typescript.rs
@@ -1171,7 +1171,10 @@ fn skip_ts_parameter_start<'a, P: Parser<'a>>(p: &mut P) -> PResult<bool> {
     let _ = eat_any_ts_modifier(p)?;
 
     let cur = p.input().cur();
-    if cur.is_word() || cur.is_this() {
+
+    if cur.is_void() {
+        Ok(false)
+    } else if cur.is_word() || cur.is_this() {
         p.bump();
         Ok(true)
     } else if (cur.is_lbrace() || cur.is_lbracket()) && parse_binding_pat_or_ident(p, false).is_ok()
--- a/crates/swc_ecma_parser/tests/typescript/issue-11124/input.ts
+++ b/crates/swc_ecma_parser/tests/typescript/issue-11124/input.ts
@@ -1,0 +1,6 @@
+var t = (): (void) => { }
+var t1 = (): (void | string) => { }
+var t2 = (): void => { }
+var t3 = (): void | number => { }
+type T = (void);
+function f(): (void) { }
--- a/crates/swc_ecma_parser/tests/typescript/issue-11124/input.ts.json
+++ b/crates/swc_ecma_parser/tests/typescript/issue-11124/input.ts.json
@@ -1,0 +1,415 @@
+{
+  "type": "Script",
+  "span": {
+    "start": 1,
+    "end": 163
+  },
+  "body": [
+    {
+      "type": "VariableDeclaration",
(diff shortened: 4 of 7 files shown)


magic-akari:
### 1. Problem Symptoms

-   Parser fails on `const f = (): (void) => { };` with error: "Parenthesized expression cannot be empty"
-   Error occurs at column 11, right at the `:` in `: (void)`

### 2. Initial Analysis

-   Confirmed via typescript-eslint's AST [output](https://ast-explorer.dev/#eNolyUEKxCAMAMCvhJxa2P2A0H3GnnIJNgeLVdEgFPHvTel1ZmBEhwd3br6GovjBYqBXkRe+0mJIj3vzzhUUNlhWB0vPYV9h+8GY1tl6UAIgFH/yX2oLORE6g8gqTQkpTZw3PYIkXw==) that `(void)` should be parsed as `TsParenthesizedType` containing `TsVoidKeyword`
-   SWC incorrectly attempts to parse it as an expression or function type

### 3. Tracing the Parse Flow

```
parse_paren_expr_or_arrow_fn (parsing arrow function)
  ↓
parse_ts_type_or_type_predicate_ann (parsing return type annotation after `:`)
  ↓
parse_ts_type
  ↓
parse_ts_non_conditional_type
  ↓
is_ts_start_of_fn_type (checking if it's a function type)
  ↓
is_ts_unambiguously_start_of_fn_type (lookahead check)
```

### 4. Root Cause Identification

The `is_ts_unambiguously_start_of_fn_type` function's logic:

1. Consumes `(`
2. Calls `skip_ts_parameter_start` to check for parameters
3. If `skip_ts_parameter_start` returns true, continues checking subsequent tokens
4. If it finds `)` followed by `=>`, it considers this a function type

### 5. The Core Issue

The bug in `skip_ts_parameter_start`:

-   When it encounters `void`, it matches the `cur.is_word()` condition
-   Incorrectly returns `true` and consumes the `void` token
-   Then sees `)` followed by `=>` (but this `=>` actually belongs to the arrow function body, not the type annotation)
-   Misidentifies `(void)` as a function type signature

### 6. The Fix

Modified `skip_ts_parameter_start` to recognize that `void` is not a valid parameter name:

-   `void` is a TypeScript type keyword and cannot be used as a parameter name
-   When encountering `void`, return `false` without consuming the token
-   This causes `is_ts_unambiguously_start_of_fn_type` to return `false`
-   The parser then correctly parses `(void)` as `TsParenthesizedType`

### 7. Interesting TypeScript Behavior Discovery

```typescript
type T = (void)   // OK
type X = (T)      // OK
type Y = (string) // OK

function foo(): (void) { } // OK
function bar(): (T) { }    // OK
function baz(): (string) { return '' } // OK

var f1 = (): (void)   => { }           // OK
var f3 = (): (T)      => {}            // 💥 TS7051:Parameter has a name but no type. Did you mean 'arg0: T'?
var f4 = (): (string) => { return '' } // 💥 TS7051:Parameter has a name but no type. Did you mean 'arg0: string'?
```

[Playground Link](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAKlC8UAUA3A9gSwCYEor4HoCoB5AaQChRIoANBZGPfFo0y66ATQaQGdgAJwwA7AOZ425ChQBmAVxEBjYBjQiostGiQ4AXMnTY8AbygBfKFMoLlq9VABGAQ0G6DSJlDOXWxabYqahouAF7uyALC4qZQghDA8oIaAOQpFlb+lBQorpoAjLz6hpi4LPAAfN4ZLLV+7Dl5sgDMRR5edZXevnV1bIC8G4ClO7AAygDsAAwArPl6AAquzgC2CRCCUAAWznxQzlAiy9CO8sD7aFCcAHRQACLYUCBo8lArzqmuYhMGMCkA-I3rWQAFjakSEogkCCqZniiWSUDSGQGwxg42mswWgkOwDWm22u32hycJzOF3AEGudywDyeLwgbwRHy+UCiEL+QA)

This shows an interesting TypeScript behavior where:

-   Parenthesized types work fine in type aliases and function return types
-   `(void)` works in arrow function return types (including unions with void)
-   But `(T)` and `(string)` fail in arrow function return types with TS7051 error, as TypeScript interprets them as parameter names without types rather than parenthesized return types

PR #11130, opened 2025-09-29
fix(es/transforms): Check errors::HANDLER.is_set() before failing

nayeemrmn:
- Will fix .. https://github.com/denoland/deno/issues/30858


kdy1:
Ah I forgot editing the commit title...

PR #11104, opened 2025-09-19
feat(hstr): Introduce `Wtf8Atom`

h-a-n-a:
**Description:**

Continue from https://github.com/swc-project/swc/pull/11085

This PR adds `Wtf8Atom` to represent unpaired surrogates (i.e. lone surrogates) in Rust.

**BREAKING CHANGE:**

**Related issue (if exists):**

Reimplemented a part of https://github.com/swc-project/swc/pull/10987

PR #11144, opened 2025-10-11
fix(es/ast): Fix unicode unpaired surrogates handling

h-a-n-a:
**Description:**

This PR reapplied https://github.com/swc-project/swc/pull/10987 and replaced `lone_surrogates: bool` mark with `Wtf8Atom`. `lone_surrogates` introduced in https://github.com/swc-project/swc/pull/10987 is more of an implicit mark, whereas the use of `Wtf8Atom` makes it more explicit where users need to explicitly call `to_string_lossy` to get the lossy UTF-8 result, making a huge difference from the original implementation where users need to convert `\uFFFDxxx` into the correct unicode. 

**BREAKING CHANGE:**

Both `cooked` in `TemplateElement` and `value` in `StringLiteral` are replaced with `Wtf8Atom` introduced in https://github.com/swc-project/swc/pull/11104.

`Wtf8Atom` does not expose a `to_string` method like the old `Atom` does. Internally, it stores the code points of the characters. You can call `code_points()` to get an iterator of the code points or call `to_string_lossy()` to get an lossy string in which all unpaired surrogates are replaced with `U+FFFD`(Replacement Character). 

### DIFF
--- a/.changeset/chatty-hotels-sort.md
+++ b/.changeset/chatty-hotels-sort.md
@@ -1,0 +1,6 @@
+---
+swc_atoms: major
+hstr: major
+---
+
+fix(es/ast): Fix unicode unpaired surrogates handling


h-a-n-a:
CI is failed here https://github.com/swc-project/swc/actions/runs/18711211814/job/53359904217?pr=11144 due to different layout of `ArchivedString` and `ArchivedVec<u8>`

kdy1:
@h-a-n-a If so, I think you can reflect the difference in the AST, too. I mean, CI failure seems to be related to mix of Vec<u8> and String, but I don't think we need to mix it, because this is a breaking change regardless

h-a-n-a:
> @h-a-n-a If so, I think you can reflect the difference in the AST, too. I mean, CI failure seems to be related to mix of Vec and String, but I don't think we need to mix it, because this is a breaking change regardless

@kdy1 I thought the breaking change would be only on the API side and not the ABI side, but for now I think it's a breaking change for all scenarios. You're right. We don't need to mix these things.

crates/swc/tests/fixture/issues-7xxx/7678/output/1.js
```diff
@@ -1 +1 @@
-let str="\uD83D\uDC68\\u200D\uD83D\uDE80";let obj={"\uD83D\uDC68\\u200D\uD83D\uDE80":"wrong"};
+let str="\\uD83D\\uDC68\\u200D\\uD83D\\uDE80";let obj={"\\uD83D\\uDC68\\u200D\\uD83D\\uDE80":"wrong"};
```
CPunisher, line 1: This is a wrong transformation, isn't it?

kdy1:
Is this PR ready for a review? Asking because CI failed

crates/swc_ecma_ast/src/typescript.rs
```diff
@@ -975,14 +974,14 @@ pub enum TsEnumMemberId {
...
-    }
-}
+// impl AsRef<Atom> for TsEnumMemberId {
+//     fn as_ref(&self) -> &Atom {
+//         match &self {
+//             TsEnumMemberId::Str(Str { value: ref sym, .. })
+//             | TsEnumMemberId::Ident(Ident { ref sym, .. }) => sym,
+//         }
+//     }
+// }
```
CPunisher, line 984: Will it be removed? Or are you missing something?

crates/swc_ecma_compat_es3/src/prop_lits.rs
```diff
@@ -46,11 +46,16 @@ impl Fold for PropertyLiteral {
         match n {
             PropName::Str(Str {
                 raw, value, span, ..
-            }) => {
-                if value.is_reserved() || !is_valid_ident(&value) {
+            }) if value.as_str().is_some() => {
+                let v = value.as_str().unwrap();
```
CPunisher: `Wtf8::as_str` could be a little costly I think...

h-a-n-a:
I think the prior CI failure is related to the wasm breaking. Looks like now the issue is gone? But I have no idea how it's fixed though.

And for the current CI failure, I was not able to reproduce the test failure of `terser_drop_unused_drop_toplevel_funcs_retain` locally. 

https://github.com/swc-project/swc/actions/runs/18831533377/job/53723863035?pr=11144

crates/swc/tests/fixture/issues-7xxx/7678/output/1.js
kdy1: Oops?
But IIRC the original output is wrong in some tests because of the lack of the proper surrogate handling (although I'm not sure if it was exactly this one)

kdy1:
I'll postpone publish of the PR until reviews are addressed

crates/swc/tests/fixture/issues-7xxx/7678/output/1.js
h-a-n-a: The original input is:

```
let str = "\\uD83D\\uDC68\\u200D\\uD83D\\uDE80";

let obj = {
    "\\uD83D\\uDC68\\u200D\\uD83D\\uDE80": "wrong"
};
```

CPunisher: ok that's great!

crates/swc_ecma_ast/src/typescript.rs
h-a-n-a: I think this can be removed as there will never be a lossless conversion as of `Str` to `Atom`

h-a-n-a: But I forgot to remove this comment and the PR is merged 🥲

crates/swc_atoms/src/wtf8_atom.rs
```diff
@@ -0,0 +1,237 @@
...
+    fn deserialize(
+        &self,
+        deserializer: &mut D,
+    ) -> Result<Wtf8Atom, <D as rancor::Fallible>::Error> {
+        let s: Vec<u8> = self.deserialize(deserializer)?;
+
+        Ok(Wtf8Atom::new(
+            // SAFETY: `ArchivedVec<u8>` is guaranteed to be serialized with `Wtf8Atom` byte
+            // sequence.  `Wtf8Atom` byte sequence is identical to `Wtf8` byte sequence.
+            unsafe { Wtf8::from_bytes_unchecked(&s) },
```
quininer, line 195: > `ArchivedVec<u8>` is guaranteed to be serialized with `Wtf8Atom` byte sequence.

This guarantee doesn't seem to be enforced, and users can construct a non-wtf8 `ArchivedVec<u8>`?

crates/swc_ecma_ast/src/typescript.rs
kdy1: I'll remove it with https://github.com/swc-project/swc/pull/11182

crates/swc_atoms/src/wtf8_atom.rs
kdy1: Documentation comment is a bit misleading, but in the compatible version of `swc_core`, the position of the field will ensure that Wtf8Atom is not mixed with other fields.

h-a-n-a: Would you please shed more light on the scenario that you're referring to?

Previously, I was thinking about the case of mixing old plugins with a newer core:

- a new `Str` core + old plugins: will deserialize it from `ArchivedString`, I assume it will return an `Result::Err` in `ArchivedString::deserialize` as the `ArchivedVec<u8>` is different from `ArchivedString`
- a new `Str` core + new plugins , will deserialize it from `ArchivedVec<u8>` and it will be a success
- an old `Str` core + new plugins, will deserialize it with `ArchivedVec<u8>`, and it will fail as the original value is an `ArchivedString`
- an old `Str` core + old plugins, will be fine.

quininer: I'm referring to the scenario where users construct malicious payload.

I realize that this seems outside of our security model due to the use of unchecked rkyv. However, this would make load untrusted plugins and untrusted caches a security risk.

Issue #11158, opened 2025-10-20
Minifier doesn't inline function but removes definition


mischnic:
### Describe the bug

The minifier removes the function `gen`, but doesn't inline it in the second occurence.

### Input code

```typescript
module.exports = () => {
  const gen = () => foo(12);
  function foo(length) {
    return length;
  }
  const a = `tmp-${gen()}-a`,
    b = `tmp-${gen()}-b`;
  console.log(a, b);
};
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2024",
    "loose": false,
    "minify": {
      "compress": true,
      "mangle": true
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": true,
  "isModule": true
}
```

### Link to the code that reproduces this issue

https://play.swc.rs/?version=1.13.20&code=H4sIAAAAAAAAA2WOQQrCMBBF9z3FLFwk0Aq6DfUsSeo0CulMSScglNzdGKUbt%2B%2FP%2B38WvueIZ3ytnGSDEZSG8QZ7BzAxbQIB6aAzs7pctanhnGmSJ1NjESnIQzcLIKHkRPCFn9tylLlaZWVZh9Nee5Uug7N9k%2Fxf4q35eVwfjByU68HX8WK6Nw5wbhi2AAAA&config=H4sIAAAAAAAAA02PSQrDMAxF9zmF0bqLEkoXvUMPYVwlOHhCUqAh5O51PaTdWe%2F766F9UAoWNvBQe37mIWlipHPOhLcg%2Bp0JoPGaDdkkcOnpwt9o0o6xoKMmIJpmlNLi8TreWgNcjIy90Zi3wU7bv9NEnwiZMxNa8bR5HWaHjVbf0Jzg42stWbtEtoTVf4ffp67qa8HysxfL0uMDc1MftxQBAAA%3D

### Expected behavior

Ideally
```js
module.exports=()=>{console.log("tmp-12-a",`tmp-12-b`)};
```

### Actual behavior

```js
module.exports=()=>{console.log("tmp-12-a",`tmp-${foo(12)}-b`)};
```

### Version

1.13.20

PR #11159, opened 2025-10-20
fix(es/minifier): fix inlining of hoisted functions

mischnic:
**Description:**

A regression from https://github.com/swc-project/swc/pull/11021

**BREAKING CHANGE:**

Closes https://github.com/swc-project/swc/issues/11158

### DIFF
--- a/.changeset/afraid-falcons-laugh.md
+++ b/.changeset/afraid-falcons-laugh.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_minifier: patch
+---
+
+fix(es/minifier): fix inlining of hoisted functions


crates/swc_ecma_minifier/src/compress/optimize/inline.rs
```diff
@@ -1026,3 +1032,47 @@ fn is_block_stmt_of_fn_simple_enough_for_copy(b: &BlockStmt) -> Option<u8> {
 
     None
 }
+
+fn arrow_simple_ident(arr: &ArrowExpr) -> Box<dyn Iterator<Item = &Ident> + '_> {
```
mischnic: This `Box<dyn` isn't ideal, but I couldn't think of a better way

PR #11186, opened 2025-10-27
refactor(es/transforms): Merge module transform visitors into single visitor

kdy1:
## Summary
Merged AMD, CommonJS, and UMD module transform visitors into a single unified visitor to reduce binary size.

## Changes
- Created new `module_transform.rs` with unified `ModuleTransform<C>` visitor
- Converted `amd.rs`, `common_js.rs`, and `umd.rs` to thin wrapper functions
- Consolidated all shared logic into single implementation:
  - Module declaration stripping
  - Import/export handling  
  - Dynamic import transformation
  - import.meta polyfills
  - Helper function calls

## Impact
- **Code reduction**: Removed ~1,728 lines of duplicate code
- **Binary size**: Eliminates duplicate visitor trait implementations and reduces monomorphization overhead
- **API compatibility**: Maintained - all existing factory functions work unchanged
- **Tests**: All 260 tests pass ✅

## Before/After
| File | Before | After | Reduction |
|------|--------|-------|-----------|
| amd.rs | 621 lines | 64 lines | 90% |
| common_js.rs | 697 lines | 50 lines | 93% |
| umd.rs | 494 lines | 51 lines | 90% |
| **New** module_transform.rs | - | 1,676 lines | - |

### DIFF
--- a/.github/workflows/binary-size-comment.yml
+++ b/.github/workflows/binary-size-comment.yml
@@ -1,0 +1,51 @@
+name: Binary Size Check - Comment
+
+on:
+  workflow_run:
+    workflows: ["Binary Size Check - Build"]
+    types:
+      - completed
+
+env:
+  CI: 1
+
+jobs:
+  comment-on-pr:
+    name: Comment on PR
+    runs-on: ubuntu-latest
+    # This workflow needs write permissions to comment on PRs
+    # It's safe because it doesn't execute any code from the PR
+    permissions:
+      pull-requests: write
+      contents: read
+    # Only run on successful builds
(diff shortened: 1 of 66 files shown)


kdy1:
Closing as this increased the binary size

magic-akari:
If the goal is to reduce binary size, we should aim to minimize generic code as much as possible.

One potential approach is to extract common logic from generic code into non-generic functions. This could help reduce monomorphization overhead, though it may come at the cost of some performance.

kdy1:
Yeah, that's generally true. I expect very minor (< 0.1%) perf regression for such changes, so I'm going to make those changes. Thanks!

PR #11192, opened 2025-10-28
refactor(es): Remove default features from `swc` crate

kdy1:
**Description:**

I'll add them back to `@swc/core`.


kdy1:
I'll retry this when SWC codebase is stabilized.

PR #11197, opened 2025-10-28
chore: Fix CI

kdy1:
# Closed as this CI failure is not fixable

`swc_core@v46.0.0` and `swc_core@v46.0.1` are yanked, so there's no way to go.
We need to publish a new version of `swc_core` first, and upgrade to it.

---

**Description:**

CI was broken due to cancellation of the publish pipeline


kdy1:
Closing as this issue is not fixable.

`swc_core@v46.0.0` and `swc_core@v46.0.1` are yanked

PR #11198, opened 2025-10-28
refactor(plugin): Switch plugin abi to flexible serialization

quininer:
**Description:**

This is a follow up to https://github.com/swc-project/swc/pull/11100, which will result in plugin abi break.

It currently passes the plugin tests, but I still need to do some code and configuration cleanup, and add a regression test.

plugin tests

**BREAKING CHANGE:**

Since we switch serialization schemes, this will result in breaking changes on the plugin abi and api.

### DIFF
--- a/crates/ast_node/src/encoding/decode.rs
+++ b/crates/ast_node/src/encoding/decode.rs
@@ -27,7 +27,7 @@ pub fn expand(DeriveInput { ident, data, .. }: DeriveInput) -> syn::ItemImpl {
                         (Some(with_type), false) => syn::parse_quote!(<#with_type<#ty> as cbor4ii::core::dec::Decode<'_>>::decode(reader)?.0),
                         (None, false) => syn::parse_quote!(<#ty as cbor4ii::core::dec::Decode<'_>>::decode(reader)?),
                         (None, true) => syn::parse_quote!(<#ty as Default>::default()),
-                        (Some(_), true) => panic!("unsupported with and ignore")
+                        (Some(_), true) => panic!("Cannot use both #[encoding(with)] and #[encoding(ignore)] attributes on the same field")
                     };
 
                     syn::parse_quote!{


quininer:
rkyv:
rkyv
cbor (with unknown):
cbor

I ran `swc_plugin_backend_tests/benches/ecma_invoke.rs`, and the new scheme is about 10% to 25% slower than rkyv scheme.

crates/swc_common/src/plugin/serialized.rs
```diff
@@ -48,8 +49,7 @@ impl PluginSerializedBytes {
      */
     #[tracing::instrument(level = "info", skip_all)]
     pub fn from_slice(bytes: &[u8]) -> PluginSerializedBytes {
```
kdy1: Can we remove this function and create a function that accepts `Vec<u8>` instead?

crates/swc_plugin_backend_tests/tests/fixture/swc_internal_plugin/src/lib.rs
```diff
@@ -76,9 +76,8 @@ pub fn process(mut program: Program, metadata: TransformPluginProgramMetadata) -
 
     dbg!();
 
-    let filename = metadata
-        .get_context(&TransformPluginMetadataContextKind::Filename)
-        .expect("Filename should exists");
```
kdy1, line 81: Please revert this. We need to check if `get_context` works

crates/swc_plugin_backend_tests/tests/fixture/.cargo/config.toml
```diff
@@ -0,0 +1,4 @@
+[target.wasm32-wasip1]
```
kdy1: Can you update the plugin template to have this?

kdy1: It's located at https://github.com/swc-project/swc/blob/622310055c59ee42b744038a33997e6f43cf4af0/crates/swc_cli_impl/src/commands/plugin.rs

crates/swc_plugin_backend_tests/tests/fixture/swc_internal_plugin/src/lib.rs
quininer: I remove it because the previous tests for `get_context` were incorrect, since `Filename` should be `None` (because `TransformPluginMetadataContext` was passed `None` https://github.com/swc-project/swc/blob/v1.14.0/crates/swc_plugin_backend_tests/tests/ecma_integration.rs#L183). but `rkyv` was return the string `0x0` before that.

crates/swc_common/src/plugin/serialized.rs
quininer: I think we could also directly consume `&[u8]` to avoid `vec`, but I'd rather leave that for later optimization.

kdy1:
Can you reflect some change requests about panic from `@Copilot` and fix the CI?

crates/swc_common/src/syntax_pos.rs
```diff
@@ -198,6 +198,107 @@ pub enum FileName {
...
+                Ok(FileName::MacroExpansion)
+            }
+            6 => {
+                let n = Array::len(reader)?;
+                debug_assert_eq!(n, Some(0));
+                Ok(FileName::ProcMacroSourceCode)
+            }
+            7 => {
+                let name = <&str>::decode(reader)?;
+                Ok(FileName::Url(Url::parse(name).map_err(|e| cbor4ii::core::error::DecodeError::Custom {
```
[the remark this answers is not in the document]
quininer: This aligns with the rkyv implement.

crates/swc_common/src/syntax_pos.rs
```diff
@@ -1392,6 +1626,101 @@ impl From<SourceMapLookupError> for Box<SpanSnippetError> {
...
+            (*self.0 as u64).encode(writer)
+        }
+    }
+
+    impl<'de> cbor4ii::core::dec::Decode<'de> for Usize<usize> {
+        fn decode<R: cbor4ii::core::dec::Read<'de>>(
+            reader: &mut R,
+        ) -> Result<Self, cbor4ii::core::dec::Error<R::Error>> {
+            <u64>::decode(reader)
+                .and_then(|n| n.try_into().map_err(|_| cbor4ii::core::dec::Error::Message("u64 value does not fit in usize")))
```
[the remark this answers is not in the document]
quininer: hallucination

crates/ast_node/src/encoding/decode.rs
```diff
@@ -23,9 +23,11 @@ pub fn expand(DeriveInput { ident, data, .. }: DeriveInput) -> syn::ItemImpl {
...
                 .map(|(field, field_name)| -> syn::Stmt {
                     let ty = &field.ty;
-                    let value: syn::Expr = match is_with(&field.attrs) {
-                        Some(with_type) => syn::parse_quote!(<#with_type<#ty> as cbor4ii::core::dec::Decode<'_>>::decode(reader)?.0),
-                        None => syn::parse_quote!(<#ty as cbor4ii::core::dec::Decode<'_>>::decode(reader)?)
+                    let value: syn::Expr = match (is_with(&field.attrs), is_ignore(&field.attrs)) {
+                        (Some(with_type), false) => syn::parse_quote!(<#with_type<#ty> as cbor4ii::core::dec::Decode<'_>>::decode(reader)?.0),
+                        (None, false) => syn::parse_quote!(<#ty as cbor4ii::core::dec::Decode<'_>>::decode(reader)?),
+                        (None, true) => syn::parse_quote!(<#ty as Default>::default()),
+                        (Some(_), true) => panic!("Cannot use both #[encoding(with)] and #[encoding(ignore)] attributes on the same field")
```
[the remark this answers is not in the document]
quininer: Panic is currently the common behavior for the `ast_node` macro.

quininer:
Copilot provides an illusion-based patch, which we need to revert.

kdy1:
Thank you so much for the hard work on the idea and implementing it!
I'll think about the way to merge this PR cleanly

PR #11199, opened 2025-10-28
mod(es/parser): Make the span of Program start at input start

### LINKED ISSUE #11167: Top-level comments should be placed after @swc/helpers imports
### Describe the bug

Hello, I'm working on some legacy project that is using `ng-annotate` package. 

I have a problem resulting from swc compilation, namely in the case of class definitions (and perhaps also in other situations that I have not encountered), the order of the JSDoc comment is shifted by placing the helper definition or helper import.

And this makes `ng-annotate` unable to process this structure. Furthermore, I suspect this state of affairs can also generate other problems, such as incorrect typing using JSDoc.

Is there any way to fix this bug, or force SWC to place these imports at the very top of the files rather than directly above the classes?

### Input code

```typescript
/** @ngInject */
class TestClass {
  constructor (private testField) {}
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false,
      "decorators": true
    },
    "target": "es2015",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    },
    "transform": {
      "decoratorVersion": "2022-03"
    },
    "preserveAllComments": true,
    "externalHelpers": true
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

https://play.swc.rs/?version=1.13.20&code=H4sIAAAAAAAAA9PXUnDIS%2FfMy0pNLlHQ0uflSs5JLC5WCEktLnEGs6p5uRQUkvPzikuKSpNL8osUNAqKMssSS1IVSoBq3DJTc1I0FapreblqAaNsl5RLAAAA&config=H4sIAAAAAAAAA2VQPY%2FCMAzd71dUng%2BpFN0NbKdbWFjZo%2BKioCSObBdRIf77JaUNrW6z%2Fez34cdHVcFVWthXj1SmJhoW5NKniQxBzT1NQIeI0rKNCp8zqpKhzjjBMjtjS2yUWBKk3OMIPF84qOELauZDaert18QFjkhwzQXeBtsNSzct%2Bcgo8k%2FUm3Bx5X6tyCZIR%2ByXRMXkCVkshWyoqZtmU%2B9gdZ3lkG%2F449wveY9B51jTAt4VORh3QBdxmXlkAE%2FnfjQ2fTh%2F8ZX%2BG95Lc9ASCqwc58uR7%2FkHjhFL0a0BAAA%3D

### Expected behavior

SWC should compile to:
```ts
import { _ as _define_property } from "@swc/helpers/_/_define_property";
/** @ngInject */
class TestClass {
    constructor(testField){
        _define_property(this, "testField", void 0);
        this.testField = testField;
    }
}
```

### Actual behavior

SWC returns:
```ts
/* @ngInject */ import { _ as _define_property } from "@swc/helpers/_/_define_property";
class TestClass {
    constructor(testField){
        _define_property(this, "testField", void 0);
        this.testField = testField;
    }
}
```

### Version

1.13.20

### Additional context

Also I found out that similar problem exists when using `@rsbuild/plugin-react` which is adding `react-runtime` imports to every file. This import is also placed before JSDoc comment. However I'm not sure if it's related.

For anyone looking for workaround  - I found out that preprocessing all JS/TS files to add semicolon (`;`) at the beginning of a file tricks compiler to work correctly. This trick does not resolve problem with `react-runtime`.

Austaras:
**Description:**

图片

**BREAKING CHANGE:**

Closes #11167

### DIFF
--- a/.changeset/nervous-years-sin.md
+++ b/.changeset/nervous-years-sin.md
@@ -1,0 +1,7 @@
+---
+swc_ecma_minifier: patch
+swc_ecma_parser: patch
+swc_core: patch
+---
+
+mod(es/parser): Program start at input start


Austaras:
Some investigation

For code
```js
// 123
class A extends B {

}
```

Current swc would output
```js
// 123
import { _ as _call_super } from "@swc/helpers/_/_call_super";
import { _ as _class_call_check } from "@swc/helpers/_/_class_call_check";
import { _ as _inherits } from "@swc/helpers/_/_inherits";
var A = /*#__PURE__*/ function(B1) {
    "use strict";
    _inherits(A, B1);
    function A() {
        _class_call_check(this, A);
        return _call_super(this, A, arguments);
    }
    return A;
}(B);
```
tsc would output
```js
"use strict";
var __extends = (this && this.__extends) || (function () {
    var extendStatics = function (d, b) {
        extendStatics = Object.setPrototypeOf ||
            ({ __proto__: [] } instanceof Array && function (d, b) { d.__proto__ = b; }) ||
            function (d, b) { for (var p in b) if (Object.prototype.hasOwnProperty.call(b, p)) d[p] = b[p]; };
        return extendStatics(d, b);
    };
    return function (d, b) {
        if (typeof b !== "function" && b !== null)
            throw new TypeError("Class extends value " + String(b) + " is not a constructor or null");
        extendStatics(d, b);
        function __() { this.constructor = d; }
        d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
    };
})();
// 123
var A = /** @class */ (function (_super) {
    __extends(A, _super);
    function A() {
        return _super !== null && _super.apply(this, arguments) || this;
    }
    return A;
}(B));
```
babel would output
```js
function _callSuper(t, o, e) { return o = babelHelpers.getPrototypeOf(o), babelHelpers.possibleConstructorReturn(t, _isNativeReflectConstruct() ? Reflect.construct(o, e || [], babelHelpers.getPrototypeOf(t).constructor) : o.apply(t, e)); }
function _isNativeReflectConstruct() { try { var t = !Boolean.prototype.valueOf.call(Reflect.construct(Boolean, [], function () {})); } catch (t) {} return (_isNativeReflectConstruct = function _isNativeReflectConstruct() { return !!t; })(); }
// 123
var A = /*#__PURE__*/function (_B) {
  function A() {
    babelHelpers.classCallCheck(this, A);
    return _callSuper(this, A, arguments);
  }
  babelHelpers.inherits(A, _B);
  return babelHelpers.createClass(A);
}(B);
```
This PR would change to babel output.

Austaras:
Some comments would be lost because they're now attached to typescript syntax node which would be eliminated later.

kdy1:
CI failures look like a real failure, but I think it may be better to wait until I finish https://github.com/swc-project/swc/issues/11205

crates/swc_typescript/tests/oxc_fixture/strip-internal.snap
```diff
@@ -1,8 +1,6 @@
 ```==================== .D.TS ====================
 
-/*
-@internal
-*/ declare class StripInternalClassFields {
```
Austaras, line 5: This comment **should** be removed.

kdy1:
I started reviewing this. Sorry for the late review.

crates/swc_ecma_minifier/tests/benches-full/lodash.js
```diff
@@ -1,11 +1,4 @@
-/**
- * @license
```
kdy1, line 2: Not sure if this is expected. I think we need to preserve some kinds of comments, and `@license` is one of them

Austaras: This happens because in original code
```js
/**
 * @license
 * Lodash <https://lodash.com/>
 * Copyright OpenJS Foundation and other contributors <https://openjsf.org/>
 * Released under MIT license <https://lodash.com/license>
 * Based on Underscore.js 1.8.3 <http://underscorejs.org/LICENSE>
 * Copyright Jeremy Ashkenas, DocumentCloud and Investigative Reporters & Editors
 */
;(function() {
```
It's considered as an leading comment for `EmptyStatement`, but that `EmptyStatement` is removed during minify. It's expected in the sense that this is how this PR should work, but we still need to get the comment back somehow.

Austaras: Similar case for [tsc](https://www.typescriptlang.org/play/?#code/PQKhCgAIUgBAbAlgYwKYDsDOqowDID2AJgIaYAWkAPOQC60AOmAXMMPMWeQHTIEC2wAHy5IAYQIMAngCdEAczqQA8gwwApAMqQAYgQCu6UrUQF0kEkcgFa5VDMh90tOQCN9tAjMzU6jFmySGABWmABm3F7ywqIASqjwqGSoRJCGRPaQALIAkgAqkEhoWKi+9Eys7JwUvALsKBjYItCQAELJqWaQAKpG9ph8MqjcoZAAjNwAHNwAzGWMlen9g6ihkTLReDliAKIAcpo7zTAS0nKKtJDq9qj8UpAAghQA1hhkADSQACIEyPr8GFoYg4+lSllSOXQADdUJgTPISCYYZB4gwvLR+pAAGSQHZERCeby4YDgRDOexhEhoXQEAiQADeUEgzPAAF8gA)

kdy1: I think we may need to fix the comment removal of minifier first (I guess? Not sure)

kdy1:
@claude Review this PR

crates/swc_ecma_minifier/tests/benches-full/lodash.js
Austaras: Yeah. I shall have a try

kdy1:
I'll merge after fixing CI with https://github.com/swc-project/swc/pull/11277

kdy1:
@claude Review this PR. Look at only `*.rs` files.

kdy1:
@claude Review this PR. Look at only `*.rs` files.

PR #11209, opened 2025-10-30
chore(es/ast): Use `opt-level = s` for AST

kdy1:
**Description:**

`swc_ecma_ast` only contains code for serialization/deserialization and some trait implementations like `Clone`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -237,7 +237,7 @@ opt-level = "s"
 opt-level = 3
 
 [profile.release.package.swc_ecma_ast]
-opt-level = 3
+opt-level = "s"
 
 [profile.release.package.swc_ecma_transforms_base]
 opt-level = 3


kdy1 (commented):
I think the perf regression caught by codspeed is all caused by wrong benchmarks.

I looked at the flamegraphs, and the only thing that slowed down is `clone()` (which exists just to run benchmark), which should not be profiled by those bencmarks

CPunisher:
@kdy1 I think we'd better only set `opt-level = s` for binding. It makes no sense to set it for root workspace because swc is usually a library. Doing so may only disrupt the benchmark result.

But I'm confused if it's worth to do so. Performance is also important.

kdy1:
It does not apply to end-user binary. It’s just to align our performance benchmark with the one of binding.

kdy1:
Actually I’d like to promote this configuration as recommended in the future, after verifying the numbers.

kdy1:
@CPunisher For example, I could verify that our opt-level configuration does not affect performance badly by copying them to the root workspace. https://github.com/swc-project/swc/pull/11210

And next.js is using a very similar opt-level configuration. https://github.com/vercel/next.js/blob/89cd2058a6b2ff711a7237b0ba5003b73b4649c1/Cargo.toml#L48-L274

I'd like to promote this as a documentation, but to do it, we need to verify that it does not make the performance worse.

CPunisher:
I'm still not too clear about the pros. IMO there's two aspects that disrupted the benchmark, although it actually does nothing with the rust api users of swc:
1. It adds a fake performance regression point in the dashboard of codpseed.
2. Since end-user usually (or by default) enable the most performant `opt-level` for swc, it causes our own benchmark results to deviate from theirs.

kdy1:
For 1, do you mean this PR? I can fix the benchmarks to not benchmark clone() calls first. Then it will not create a fake point.

For 2, I’m not sure if any rust user would look at the codspeed dashboard. It’s relative to previous commit, so typically it’s useful only for the main developers.

kdy1:
But about pros, swc/core is also a product and we need a way to profile the performance of it

CPunisher:
> For 1, do you mean this PR? I can fix the benchmarks to not benchmark clone() calls first. Then it will not create a fake point.

How will you avoid `clone()`s in the benchmark? I think they are everywhere in the crates. 

> For 2, I’m not sure if any rust user would look at the codspeed dashboard. It’s relative to previous commit, so typically it’s useful only for the main developers.

For example, we may find a optimization that makes parser 1% faster under `opt-level = s`. However, this optimization may not take effect with `opt-level = 3`.

kdy1:
> How will you avoid clone()s in the benchmark? I think they are everywhere in the crates.

`clone()` in this benchmark is one like https://github.com/swc-project/swc/blob/dba23f5a72d26b3b62fbafe2d8a65c69c3642669/crates/swc_ecma_transforms_base/benches/base.rs#L28. Those are not related to real workload.

> For example, we may find a optimization that makes parser 1% faster under opt-level = s. However, this optimization may not take effect with opt-level = 3.

This is a fair point, but I'm not going to slow down (opt-level = s) any crate where performance is important, and we need a way to profile the performance of `@swc/core`. I think being able to profile the performance of `@swc/core` outweighs the concern of weird optimization.

kdy1:
@claude resolve merge conflicts

PR #11225, opened 2025-11-03
perf(es/parser): Optimize `skip_space`

CPunisher:
**Description:**

1. Originally, the skip space function had two nested loops: the inner loop handled whitespace, and the outer loop checked for slash comments. The inner loop used `match` to deal with spaces and newlines, and used a jump table when encountering Unicode characters. Now, everything is handled in a single loop, all match statements are removed, and everything goes through the jump table.
2. After finding one space, it’s very likely there are more spaces following. Instead of going back to the jump table each time, it continuously tries to process the spaces in a row.
3. After finding a newline, the lexer's `had_line_break` is set to true. At this point, it can also skip over any following spaces and newlines continuously, since the `had_line_break` state only needs to be set once.
4. The original code handled Unicode at the byte level, for example by checking prefixes to determine Unicode length and manually combining multiple bytes into a full UTF-8 character. Now it simply reads the next character directly.

### DIFF
--- a/.changeset/fluffy-cobras-agree.md
+++ b/.changeset/fluffy-cobras-agree.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_parser: patch
+---
+
+perf(es/parser): optimize `skip_space`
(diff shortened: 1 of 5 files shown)


kdy1 (approved):
Nice improvement! Thanks!

magic-akari:
Impressive optimization!

PR #11237, opened 2025-11-06
chore: remove unconditioned instrument code

stormslowly:
To check if tracing has a performance cost

**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -123,7 +123,7 @@ resolver = "2"
   thiserror                 = "1.0.30"
   tokio                     = { version = "1", default-features = false }
   toml                      = "0.8.2"
-  tracing                   = "0.1.41"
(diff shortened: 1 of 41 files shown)


kdy1:
It has a noticeable cost. You can try enabling `tracing/max_level_off`

stormslowly:
In rspack-resolver, when I enable `max_level_off` and `release_max_level_off`, I still get 2~15% performance loss in benchmark.

kdy1:
I think it's a good idea to remove all tracing code, considering the perf difference

stormslowly:
According to codspeed's report, the most improved step is enabling `max_level_off`.
That's quite different from rspack-resolver's case.
I will do some digging to find the season and figure out a way to improve performance and keep trackability.

stormslowly:
close first, i will split this PR

PR #11252, opened 2025-11-10
perf(es/parser): Inline `skip_space`

CPunisher:
**Description:**

I think it's worth to inline `skip_space` after https://github.com/swc-project/swc/pull/11225, which made `skip_space` simple  and short.


kdy1 (approved):
Nice!

PR #11290, opened 2025-11-17
feat(test): Migrate from test262-parser-tests to full test262 suite

magic-akari:
**Description:**

**BREAKING CHANGE:**


magic-akari:
Close as stale

PR #11300, opened 2025-11-18
perf(parser): Use pratt parsing

magic-akari:
**Description:**

**BREAKING CHANGE:**


magic-akari:
Not sure whether it will improve performance, but I want to check the benchmark results first.

PR #11302, opened 2025-11-18
fix(es/codegen): Restore missing top-level comments

### LINKED ISSUE #11167: Top-level comments should be placed after @swc/helpers imports

kdy1:
This fixes an issue where top-level comments (like copyright headers) were being lost after PR #11199 changed Module/Script spans to start at position 0.

The fix skips emitting leading comments at Module/Script level when the body is non-empty. Instead, comments are emitted by the first body item with a real source span. Injected imports have DUMMY_SP, so they skip comment emission, and original source items emit their comments correctly after the imports.

Closes #11167

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.changeset/few-bees-crash.md
+++ b/.changeset/few-bees-crash.md
@@ -1,0 +1,7 @@
+---
+swc_ecma_codegen: patch
+swc_ecma_parser: patch
+swc_core: patch
+---
+
+fix(es/codegen): Restore missing top-level comments


Austaras:
Seems like this PR would still make some comments missing like in benchmark/{lodash, moment}.js? @kdy1

kdy1:
I'll try another fix

PR #11323, opened 2025-11-21
perf(es/parser): Optimize `byte_search!`

CPunisher:
**Description:**

Align the implementation to oxc. In terms of swc, this pr reduces the calls of `bump_bytes`.

PR #11328, opened 2025-11-25
fix(es/parser): Don't call `bump_bytes` in the `continue_if` of `byte_search!`

CPunisher:
**Description:**

In https://github.com/swc-project/swc/pull/11323, we only records the `pos` as an offset related the current remaining string and bumps the lexer at the end of `byte_search!` to reduce the calls to `bump_bytes`. However, if `bump_bytes` is manually called in the `continue_if`, the invariant is broken.


kdy1 (approved):
Thank you, as always!

PR #11329, opened 2025-11-25
chore(ci): Don't run claude code if there's no token

CPunisher:
**Description:**

Forked project doesn't have secrets


kdy1:
Thank you!

PR #11340, opened 2025-12-01
perf(es/parser): Small optimization after byte-based lexer

CPunisher:
**Description:**

This PR is the continuation of https://github.com/swc-project/swc/pull/11318


CPunisher:
No improvement. I think the compiler has done the optimiaztion.

PR #11341, opened 2025-12-01
perf(es/parser): Use `slice` rather than matching keywords

CPunisher:
**Description:**

The keyword tokens and source code are one-to-one, so we can directly slice them from the source code instead of using match to construct atoms.


kdy1 (approved):
Great work!

PR #11359, opened 2025-12-09
feat(es/parser): support `no_paren` parser option

CPunisher:
**Description:**

There's no `ParenExpr` ast type in estree spec: https://github.com/estree/estree/blob/master/es5.md

This PR is a simple and useful implementation to eliminate the usage of `paren_remover`.


kdy1 (dismissed):
I have some concerns regarding comments, but I think it can be resolved in the follow-up PR for actually using this feature.

PR #8144, opened 2023-10-18
feat(es/minifier): Minify `"undefined" == typeof x` to `"u" < typeof x`

magic-akari:
**Description:**

The only thing I'm worried about is IE returning "unknown" for some objects.
https://stackoverflow.com/questions/10982739/typeof-returning-unknown-in-ie

**BREAKING CHANGE:**

**Related issue (if exists):**

PR #11367, opened 2025-12-12
feat(es/minifier): Optimize `typeof x == "undefined"` to `typeof x > "u"`

magic-akari:
**Description:**

Optimize `typeof x == "undefined"` comparisons to shorter forms:

- `typeof x == "undefined"` → `typeof x > "u"` (for undeclared variables)
- `typeof x != "undefined"` → `typeof x < "u"` (for undeclared variables)
- `typeof x == "undefined"` → `void 0 === x` (for declared variables)

The `> "u"` optimization works because `"undefined"` is the only typeof result greater than `"u"` in lexicographic order:

| typeof result | > "u" |
|---------------|-------|
| "bigint"      | false |
| "boolean"     | false |
| "function"    | false |
| "number"      | false |
| "object"      | false |
| "string"      | false |
| "symbol"      | false |
| "undefined"   | true  |

### Performance

There were performance concerns with string comparison vs equality check, but these have been or are being addressed

**BREAKING CHANGE:**

**Related issue:**
- #8144
- https://crrev.com/c/5218959
- https://crrev.com/c/7218807
- WebKit/WebKit#50101
- WebKit/WebKit#50116

### DIFF
--- a/.changeset/proud-mirrors-report.md
+++ b/.changeset/proud-mirrors-report.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_minifier: patch
+---
+
+feat(es/minifier): Optimize `typeof x == "undefined"` to `typeof x > "u"`


TODO-negate-comparison-operators.md
kdy1, line 1: Can you remove these files?

magic-akari: Done

PR #11373, opened 2025-12-15
perf(es/parser): Remove `raw`s in `TokenValue`

CPunisher:
**Description:**

1. This PR removes all `raw: Atom`s in `TokenValue`, and the parser gets it from `self.input.iter.read_string(token_span)`.
2. For `TokenValue::Regex { exp, flags }`, I change it to `TokenValue::Regex(exp_end: u32)`. And the parser get the `exp` and `flags` with `token_span.lo + 1..exp_end` and `exp_end + 1..token_span.hi`.
3. You can see many snapshot updates in the diff. Because I simplify the `Token::to_string` logic, which will not print their detailed token values any more. I think this is reasonable because the diagnostic has already underlined them with the source code.

### DIFF
--- a/.changeset/rotten-ravens-decide.md
+++ b/.changeset/rotten-ravens-decide.md
@@ -1,0 +1,5 @@
+---
+swc_ecma_parser: major
+---
+
+perf(es/parser): Remove `raw`s in `TokenValue`


kdy1:
Thanks!

PR #11374, opened 2025-12-17
perf(es/minifier): Optimize data structures of `ProgramData`

CPunisher:
**Description:**

1. Change `scopes` from `FxHashMap<SyntaxContext, ScopeData>` to `Vec<ScopeData>` because we can know that the `SyntaxContext` is dense.
2. We can share one `scopes` and `property_atoms` between the analyzers to reduce the allocation.

### DIFF
--- a/.changeset/old-phones-bow.md
+++ b/.changeset/old-phones-bow.md
@@ -1,0 +1,5 @@
+---
+swc_ecma_usage_analyzer: major
+---
+
+perf(es/minifier): optimize data structures of `ProgramData`


crates/swc_ecma_minifier/src/program_data.rs
```diff
@@ -24,25 +24,34 @@ where
 {
     let data = if collect_property_atoms {
         ProgramData {
+            vars: FxHashMap::default(),
+            initialized_vars: IndexSet::default(),
+            top: ScopeData::default(),
+            scopes: Vec::new(),
             property_atoms: Some(Vec::with_capacity(128)),
-            ..Default::default()
```
kdy1, line 28: Why do we need these changes? (I mean the removal of `Default` impl`)

CPunisher: Sorry, I forget to change it back when I finished this PR.

PR #11387, opened 2025-12-19
refactor(es/parser): Distinguish JsxText from Str

CPunisher:
**Description:**

Since JSXText is always valid utf8 string, this pr adds `TokenValue::JsxText(Atom)` to avoid unnecessary conversion between `Wtf8Atom` and `Atom`


kdy1 (dismissed):
Thank you!

PR #11395, opened 2025-12-24
perf(es/parser): Remove `is_first` in lexer state

CPunisher:
**Description:**

We can use a dedicate `first_token` to remove the checks in `next_token`, which is triggered in every token bump. 

At the same time, I also cleaned up and refactor some simple APIs, which is introdced in the previous trait-based implementation with `swc_ecma_lexer`:
- prev_hi
- had_line_break


kdy1 (approved):
Great!

PR #11396, opened 2025-12-24
perf(es/parser): optimize identifier related lexer functions

CPunisher:
**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_parser/src/lexer/mod.rs
+++ b/crates/swc_ecma_parser/src/lexer/mod.rs
@@ -2219,7 +2219,9 @@ impl<'a> Lexer<'a> {
         let word = Token::unknown_ident(atom, self);
 
         if has_escape {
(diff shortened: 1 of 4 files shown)


crates/swc_ecma_parser/src/lexer/mod.rs
```diff
@@ -2300,3 +2305,8 @@ fn pos_span(p: BytePos) -> Span {
 fn fixed_len_span(p: BytePos, len: u32) -> Span {
     Span::new_with_checked(p, p + BytePos(len))
 }
+
+#[cold]
+fn code_branch<T, F: FnOnce() -> T>(f: F) -> T {
```
magic-akari:
```suggestion
fn cold_branch<T, F: FnOnce() -> T>(f: F) -> T {
```

PR #11398, opened 2025-12-26
perf(es/parser): Use `byte_search` to optimize `scan_jsx_token`

CPunisher:
**Description:**
1. Remove `allow_multiline_jsx_text` param from `scan_jsx_token`, because it's always `true`.
2. Divide `scan_jsx_token` into two paths. We assume there's no `&` in the fast path, so we can use `byte_search` to reduce the bump cost. And we fallbacks to the slow path, which is bascially same as the original implementation.

### DIFF
--- a/.changeset/dull-candles-fail.md
+++ b/.changeset/dull-candles-fail.md
@@ -1,0 +1,5 @@
+---
+swc_ecma_parser: major
+---
+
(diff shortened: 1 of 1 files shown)


kdy1 (dismissed):
Nice work!

PR #11400, opened 2025-12-29
perf(es/parser): Add inline(always) to hot path methods

kdy1:
## Summary
- Add `#[inline(always)]` to `SafeByteMatchTable::matches` method
- Add `#[inline(always)]` to `Buffer::bump` method

These are hot path methods in the parser that benefit from aggressive inlining to improve performance.

## Test plan
- Existing tests should pass
- Performance benchmarks should show improvement


kdy1:
Okay Claude Code did something dumb

PR #11406, opened 2025-12-29
fix(es/minifier): Evaluate TemplateLiteral in BinaryExpression

cuyl:
Make sure that `TemplateLiteral` are correctly handled when evaluating `BinaryExpression`.
    
 Refs #11390


kdy1:
Thanks!

PR #11411, opened 2025-12-30
refactor(es/minifier): Improve nested template literal evaluation

cuyl:
Refactors `eval_nested_tpl` to more robustly handle nested template literals by using `zip` for quasis and exprs, ensuring proper concatenation and merging.

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/pure/strings.rs
+++ b/crates/swc_ecma_minifier/src/compress/pure/strings.rs
@@ -155,7 +155,7 @@ impl Pure<'_> {
             };
         }
 
-        // Consume a quasis first to make sure it align with exprs
+        // Consume quasis first to make sure it align with exprs
         // quasis.len() == exprs.len() + 1
         if let Some(q) = q_iter.next() {
             push_str!(q);
@@ -166,21 +166,21 @@ impl Pure<'_> {
             match *e {
                 Expr::Tpl(mut tpl) => {
                     // For evaluated template only the first
-                    // and the last quasis could be concat with
+                    // and the last quasi could be concat with
                     // outside quasis.
                     let mut quasis_taken = tpl.quasis.take();
                     let l = quasis_taken.len();
 
-                    // Store the first quasis for later concat
+                    // Store the first quasi for later concat
                     let first = quasis_taken[0].take();
                     push_str!(first);
 
                     if l > 1 {
-                        // If there are more than one quasis
+                        // If there are more than one quasi
                         // Concat first with outside quasis
                         end_str!();
 
-                        // Store the last quasis for later concat
+                        // Store the last quasi for later concat
                         let last = quasis_taken.pop().unwrap();
                         push_str!(last);
 


cuyl:
I have read some of the codes here, I don't know much of the contexts here, and have some confusion need to clarify.
Just to make sure my changes here is reasonable.

1. Why using `Str::from_tpl_raw(&q.raw)` instead of using `&q.cooked` directly here.
2. `cooked` is Optional and when will it be `None`? is it always available after parse code?
    By reading `convert_tpl_to_str` it seems like cooked could be `None` and it fallback to using `Str::from_tpl_raw` but only in some condition.
3.  The following code `Str::from_tpl_raw` and `read_escaped_char` looks similar, maybe we can reuse some code.

https://github.com/swc-project/swc/blob/3554268dcb7c8af4abfe0a06e61a382a23c4a3eb/crates/swc_ecma_ast/src/lit.rs#L262-L309

https://github.com/swc-project/swc/blob/3554268dcb7c8af4abfe0a06e61a382a23c4a3eb/crates/swc_ecma_parser/src/lexer/mod.rs#L1541-L1609

https://github.com/swc-project/swc/blob/3554268dcb7c8af4abfe0a06e61a382a23c4a3eb/crates/swc_ecma_minifier/src/compress/pure/strings.rs#L202-L253


@kdy1

kdy1:
Good questions! Let me explain based on the codebase:

### 1. Why use `Str::from_tpl_raw(&q.raw)` instead of `&q.cooked`?

The code in `convert_tpl_to_str` (crates/swc_ecma_minifier/src/compress/pure/strings.rs:202-253) actually **tries to use `cooked` first** if it's available (lines 205-223). However, it falls back to `from_tpl_raw(&q.raw)` when:
- `cooked` is `None` (see question 2 below)
- `cooked` contains characters that shouldn't be in a regular string literal (like backslashes)

The fallback path (lines 226-249) uses `from_tpl_raw` to safely convert template-specific escape sequences.

### 2. When is `cooked` None?

According to the parser code (crates/swc_ecma_parser/src/parser/expr.rs:668-677):

```rust
let cooked = self.input.expect_template_token_value();
let cooked = match cooked {
    Ok(cooked) => Some(cooked),
    Err(err) => {
        if is_tagged_tpl {
            None  // <-- cooked is None here
        } else {
            return Err(err);
        }
    }
};
```

`cooked` is `None` when:
- The template is a **tagged template literal** (like ``tag\`string\```)
- AND it contains **invalid escape sequences** (like `\x`, `\u` without proper hex digits)

This follows the ES2018 spec which allows tagged templates to have invalid escape sequences. For non-tagged templates, invalid escapes cause a parse error instead.

### 3. Code similarity between `from_tpl_raw` and `read_escaped_char`

While they look similar, they serve different purposes:

- **`from_tpl_raw`** (crates/swc_ecma_ast/src/lit.rs:262-309): Handles a **limited subset** of escape sequences specific to templates (`\\\``, `\$`, `\\\\`, `\b`, `\f`, `\n`, `\r`, `\t`, `\v`). It's a **post-parse** transformation.

- **`read_escaped_char`** (crates/swc_ecma_parser/src/lexer/mod.rs:1541-1656): Handles **all escape sequences** including hex (`\x`), unicode (`\u`), octal, line continuations, etc. It's used during **lexing/parsing**.

The key difference:
- `read_escaped_char` parses escape sequences from source code during tokenization
- `from_tpl_raw` converts already-parsed template raw strings to regular strings

Reuse might be possible, but they operate at different stages with different requirements, so it would need careful consideration.

kdy1:
It's generated by Claude Code, but it seems legit.

cuyl:
Greate explanation,

So in here we only evaluate nest `Tpl` not `TaggedTpl`, so that I can assume cooked always available,
and change `Str::from_tpl_raw(&q.raw).as_ref()` to `&q.cooked.unwrap()`

cuyl:
Why `convert_tpl_to_str` only allows `'\u{0020}'..='\u{007e}'` that is visible ascii code here convert to str? I guess we can handle more cases.
@kdy1

kdy1:
Yeah I think it worths a separate PR

cuyl:
OK, I’ll do it when I have time

crates/swc_ecma_minifier/src/compress/pure/strings.rs
```diff
@@ -126,64 +126,73 @@ impl Pure<'_> {
         };
         let mut cur_cooked_str = Wtf8Buf::new();
         let mut cur_raw_str = String::new();
+        let mut q_iter = tpl.quasis.take().into_iter();
+        let e_iter = tpl.exprs.take().into_iter();
+
+        // Consume a quasis first to make sure it align with exprs
+        // quasis.len() == exprs.len() + 1
+        if let Some(q) = q_iter.next() {
+            cur_cooked_str.push_wtf8(&q.cooked.unwrap());
```
kdy1: This may panic

kdy1: `Tpl` might be synthesized from somewhere

cuyl: Did you mean `Tpl` can synthesized from somewhere that do not have `cooked`?

kdy1: Yeap. Maybe we can just document it and assume, though.

There are some assumptions made by the SWC minifier documented in swc.rs

cuyl: I think we can assume cooked always not `None` here, but anyway, I can fallback to from_tpl_raw here.

kdy1:
@claude Review this PR

cuyl:
Oops, a lot of grammar issues, already applied.

kdy1 (approved):
Thanks!

PR #11412, opened 2026-01-01
perf(es/codegen,es/utils): Migrate to dragonbox_ecma for faster Number::toString

magic-akari:
**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/.changeset/good-balloons-explode.md
+++ b/.changeset/good-balloons-explode.md
@@ -1,0 +1,7 @@
+---
+swc_ecma_codegen: patch
+swc_ecma_utils: patch
+swc_core: patch
+---
+
+perf(es/codegen,es/utils): Migrate to dragonbox_ecma for faster Number::toString


kdy1:
Is it really faster in real world workload? I can't find any differences in [codspeed](https://codspeed.io/swc-project/swc/branches/magic-akari%3Aperf%2Fuse-dragonbox-ecma?utm_source=github&utm_medium=comment&utm_content=header)

magic-akari:
oxc-project/oxc#12821 shows a ~4% codegen improvement when introducing dragonbox, but that result was based on dragonbox v0.0.5. This PR uses v0.1.0 instead, so the results are not directly comparable. 

The lack of a visible improvement could be due to a performance regression between versions or simply benchmark noise. I'll open a anther PR that introduces dragonbox at v0.0.5 and re-runs Codespeed for an apples-to-apples comparison.

kdy1 (approved):
As of https://github.com/swc-project/swc/pull/11413#issuecomment-3704368396

PR #11413, opened 2026-01-02
[DONT MERGE] dragonbox for codespeed

magic-akari:
**Description:**

**BREAKING CHANGE:**


magic-akari:
Still no measurable performance improvement with dragonbox v0.0.5. CC @kdy1

kdy1:
Then I think it might be CPU-specific (if oxc uses a different CPU than the default GHA one) or specific to memory access pattern of oxc. Do you think we should still merge it?

magic-akari:
I do lean toward migrating to dragonbox. Dragonbox is generally considered faster than Ryu and produces smaller code and lookup tables. As a result, it has been gradually replacing Ryu in newer C++ standard library implementations (e.g., the Microsoft STL) as well as in high-performance libraries such as {fmt}.

PR #11415, opened 2026-01-03
refactor(es/minifier): Improve tpl to str

cuyl:
**Description:**

- Enhance `Pure::convert_tpl_to_str` to prefer `cooked` values as much as possible.
- Improve `Str::from_tpl_raw` to return `Wtf8Atom` instead of `Atom`.
- Add more test cases.


cuyl:
I'm not sure why we have `force_str_for_tpl` here, I can see 2 definition:
1. `Minification::force_str_for_tpl` returns `false`
2. `Eval::force_str_for_tpl` returns `true`

There are some encoding logic in `convert_str_value_to_tpl_cooked`, why we need convert `Str::value` to `Tpl::cooked` using `convert_str_value_to_tpl_cooked`, Aren't they the same thing?
Can I just remove `convert_str_value_to_tpl_cooked`?
@kdy1

kdy1:
Does CI passes, if you remove it?

cuyl:
I have changed some outputs here, but looks not a bug.
@kdy1

kdy1:
@claude Review this PR. My concerns are

 - the correctness of `compress_array_join_as_tpl` in crates/swc_ecma_minifier/src/compress/pure/misc.rs
 - (as many as) edge cases you can think of, based on the code changes

kdy1:
@claude Review this PR. My concerns are

 - the correctness of `compress_array_join_as_tpl` in crates/swc_ecma_minifier/src/compress/pure/misc.rs
 - (as many as) edge cases you can think of, based on the code changes 

You should do `gh pr checkout 11415` to checkout the PR branch

kdy1 (commented):
## Review Summary

I've identified several critical issues that need to be addressed:

1. **CRITICAL**: Compilation error in  - calls removed function
2. **HIGH**: Surrogate pair handling logic has multiple bugs in 
3. **MEDIUM**: Error handling uses  instead of graceful degradation
4. **LOW**: Missing test coverage for edge cases

Detailed comments are provided on each file. Please address the critical and high priority issues before merging.

kdy1 (commented):
Review Summary

I have identified several critical issues that need to be addressed:

1. CRITICAL: Compilation error in misc.rs line 1429 - calls removed function
2. HIGH: Surrogate pair handling logic has multiple bugs in lit.rs
3. MEDIUM: Error handling uses panic instead of graceful degradation
4. LOW: Missing test coverage for edge cases

Detailed comments are provided on each file. Please address the critical and high priority issues before merging.

crates/swc_ecma_minifier/src/compress/pure/misc.rs
```diff
@@ -1426,7 +1423,7 @@ impl Pure<'_> {
                     }
                 }
                 Expr::Lit(Lit::Str(s)) => {
-                    cur_cooked.push_wtf8(&convert_str_value_to_tpl_cooked(&s.value));
+                    cur_cooked.push_wtf8(&Cow::Borrowed(&s.value));
                     cur_raw.push_str(&convert_str_value_to_tpl_raw(&s.value));
                 }
                 _ => {
```
kdy1, line 1429: **CRITICAL - Compilation Error**

This line calls `convert_str_value_to_tpl_cooked` which is removed in this PR (line 586 in strings.rs was deleted).

This will cause a compilation error. Should be changed to:
```rust
cur_cooked.push_wtf8(&Cow::Borrowed(&s.value));
```

This matches the changes made in strings.rs at lines 341, 405, and 441.

crates/swc_ecma_ast/src/lit.rs
```diff
@@ -259,47 +262,148 @@ impl Str {
...
+                                    }
+                                }
+                                if max_len == 2 && max_len != count {
+                                    panic!("invalid syntax");
+                                }
+                                if (0xd800..=0xdfff).contains(&result) {
+                                    // UTF-16 surrogate pair
+
+                                    if result & 0x400 == 0 {
+                                        // High surrogate pair
```
kdy1, line 365: **HIGH PRIORITY - Incorrect Surrogate Pair Detection**

The bitmask check `result & 0x400 == 0` is incorrect for distinguishing high vs low surrogates.

- High surrogates: 0xD800-0xDBFF
- Low surrogates: 0xDC00-0xDFFF

The correct check should be:
```rust
if result < 0xDC00 {
    // High surrogate (0xD800-0xDBFF)
} else {
    // Low surrogate (0xDC00-0xDFFF)
}
```

Or using bitmask:
```rust
if (result & 0xFC00) == 0xD800 {
    // High surrogate
}
```

crates/swc_ecma_ast/src/lit.rs
```diff
@@ -259,47 +262,148 @@ impl Str {
...
+                                        }
+                                    } else {
+                                        // Low surrogate pair
+                                        if prev_result != NO_PREV_RESULT {
+                                            // Syntax error: we do not have a
+                                            // high surrogate pair.
+                                            panic!("invalid syntax");
+                                        }
+                                        let cp = (result & 0x3ff) | (prev_result << 10);
+                                        prev_result = NO_PREV_RESULT;
```
kdy1, line 389: **HIGH PRIORITY - Incorrect Surrogate Pair Decoding Formula**

The surrogate pair decoding formula is missing the base offset. The current code:
```rust
let cp = (result & 0x3ff) | (prev_result << 10);
```

Should be:
```rust
let cp = 0x10000 + ((prev_result & 0x3ff) << 10) + (result & 0x3ff);
```

The formula for decoding UTF-16 surrogate pairs is:
- Code point = 0x10000 + (high_bits << 10) + low_bits

Where high_bits and low_bits are the lower 10 bits of each surrogate.

Also note: the condition check above seems inverted - if `result & 0x400 == 0` you are treating it as high surrogate and looking ahead, but this branch is inside the else clause for low surrogate handling based on the comment.

crates/swc_ecma_ast/src/lit.rs
```diff
@@ -259,47 +262,148 @@ impl Str {
...
+                                            count += 1;
+                                        }
+                                        'A'..='F' => {
+                                            result = (result << 4) | (c as u32 - 'A' as u32 + 10);
+                                            count += 1;
+                                        }
+                                        _ => panic!("invalid syntax"),
+                                    }
+                                }
+                                if max_len == 2 && max_len != count {
```
kdy1, line 356: **MEDIUM PRIORITY - Use of panic for error handling**

Using `panic!` for invalid escape sequences will crash the minifier at runtime. Consider these alternatives:

1. Output the raw escape sequence as-is (preserve what the user wrote)
2. Return a Result type and propagate errors gracefully
3. Use WTF-8 to preserve invalid sequences

Example malformed inputs that would panic:
- `` `\xF` `` - incomplete hex escape (only 1 digit instead of 2)
- `` `\u` `` - incomplete unicode escape
- `` `\u{110000}` `` - out of range unicode

For a production minifier, panicking on malformed input is not ideal. Consider graceful degradation.

crates/swc_ecma_minifier/src/compress/pure/misc.rs
kdy1: I think this is a halucination

crates/swc_ecma_ast/src/lit.rs
```diff
@@ -259,47 +262,148 @@ impl Str {
...
+                                    // Output raw value for syntax issue
+                                    buf.push_str(format!("\\u{result:x}").as_str());
+                                } else if result <= 0x10ffff {
+                                    // We can be sure result is a valid code point here
+                                    buf.push(unsafe { CodePoint::from_u32_unchecked(result) });
+                                } else {
+                                    panic!("invalid syntax");
+                                }
+                            }
+                            '0'..='7' => {
```
kdy1, line 409: **MEDIUM PRIORITY - Incomplete octal escape handling**

Template literals do not support octal escapes (they are a syntax error). The current code only handles `\0` followed by a non-octal digit, but does not handle other cases:

- `` `\0` `` followed by EOF or non-octal - Correctly handled ✓
- `` `\01` `` - Should be rejected but will panic ✗
- `` `\1` `` through `` `\7` `` - Should be rejected but will panic ✗

According to ES spec, octal escapes are not allowed in template literals (strict mode syntax error).

Recommendation: Either reject all octal-looking escapes with graceful error handling, or output them as-is in the WTF-8 buffer.

crates/swc_ecma_ast/src/lit.rs
```diff
@@ -259,47 +262,148 @@ impl Str {
...
+                                    panic!("invalid syntax");
+                                }
+                                if (0xd800..=0xdfff).contains(&result) {
+                                    // UTF-16 surrogate pair
+
+                                    if result & 0x400 == 0 {
+                                        // High surrogate pair
+                                        let mut iter = iter.clone();
+                                        if let Some('\\') = iter.next() {
+                                            if let Some('u') = iter.next() {
```
kdy1, line 375: **HIGH PRIORITY - Iterator not advanced after lookahead**

When detecting a high surrogate pair, the code does lookahead but does not advance the main iterator:

```rust
let mut iter = iter.clone();  // Creates a clone, shadows the binding
if let Some('\\') = iter.next() {
    if let Some('u') = iter.next() {
        prev_result = result & 0x3ff;
        continue;  // Main iterator not advanced!
    }
}
```

This will cause the main loop to re-process the same `\u` sequence, leading to incorrect output.

Fix: Do not shadow the iter binding - use a different name for the clone, then conditionally advance the main iter:
```rust
let mut peek_iter = iter.clone();
if let Some('\\') = peek_iter.next() {
    if let Some('u') = peek_iter.next() {
        prev_result = result & 0x3ff;
        // Now advance the main iterator to consume the lookahead
        iter.next(); // consume backslash
        iter.next(); // consume u
        continue;
    }
}
```

crates/swc_ecma_minifier/tests/eval.rs
```diff
@@ -101,6 +103,13 @@ fn eval_bin() {
 fn eval_lit() {
     assert_eq!(eval("", "true").unwrap(), "true");
     assert_eq!(eval("", "false").unwrap(), "false");
+    assert_eq!(eval("", "`\\xff`").unwrap(), "\u{ff}");
```
kdy1: **LOW PRIORITY - Expand test coverage**

The new tests are good but miss several important edge cases:

**Surrogate pairs:**
- Unpaired high surrogate: `` `\uD83D` ``
- Unpaired low surrogate: `` `\uDCA9` ``
- High surrogate followed by non-surrogate: `` `\uD83Dabc` ``

**Invalid escapes:**
- Incomplete hex: `` `\xF` ``, `` `\x` ``
- Incomplete unicode: `` `\u123` ``, `` `\u` ``
- Out of range: `` `\u{110000}` ``

**Line continuations:**
- Backslash CR: `` `a\<CR>b` ``
- Backslash LF: `` `a\<LF>b` ``
- Backslash CRLF: `` `a\<CR><LF>b` ``

**Octal escapes:**
- `` `\0` `` (null, valid)
- `` `\01` `` (octal, should error)
- `` `\7` `` (octal, should error)

Consider adding these tests to ensure the implementation handles all cases correctly.

kdy1:
**Correction on my first comment:**

You are absolutely right that my comment about line 1429 in misc.rs was incorrect. Looking at the diff more carefully, the change on that line is:

```diff
- cur_cooked.push_wtf8(&convert_str_value_to_tpl_cooked(&s.value));
+ cur_cooked.push_wtf8(&Cow::Borrowed(&s.value));
```

This shows you already fixed this correctly in the PR - the new version uses `Cow::Borrowed`. I misread the diff and thought the old function was still being called. My apologies for the confusion!

The other comments about surrogate pair handling, iterator advancement, and panic usage remain valid concerns for the `from_tpl_raw` implementation in lit.rs.

cuyl:
I assumed the panic is unreachable code, I will add error handing later.

crates/swc_ecma_ast/src/lit.rs
cuyl: This is intended, Here I want to check if next two char is  `\u`, so next loop will continue handle the logic.
otherwise just output result.

crates/swc_ecma_minifier/tests/eval.rs
cuyl: I guess put more tests as unit tests for ` Str::from_tpl_raw` will be better.

cuyl:
I will add more test cases later.

kdy1:
Thanks!

crates/swc_ecma_ast/src/lit.rs
```diff
@@ -253,53 +256,189 @@ impl<'a> arbitrary::Arbitrary<'a> for Str {
...
+                                            });
+                                        }
+                                        let mut iter = iter.clone();
+                                        if let Some('\\') = iter.next() {
+                                            if let Some('u') = iter.next() {
+                                                // less than 0xdc00
+                                                prev_result = result;
+                                                continue;
+                                            }
+                                        }
```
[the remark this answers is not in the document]
cuyl: Not missing, it should handled at line 400

kdy1:
## 1. CRITICAL - Incorrect Surrogate Pair Decoding Formula

**Location:** `crates/swc_ecma_ast/src/lit.rs:384-385`

The bitwise operation is incorrect:
```rust
result = 0x10000 | ((result & 0x3ff) | ((prev_result & 0x3ff) << 10));
```

**Problem:** Using bitwise OR (`|`) instead of addition (`+`) will produce incorrect results for certain surrogate pairs where the bits overlap.

**Should be:**
```rust
result = 0x10000 + ((prev_result & 0x3ff) << 10) + (result & 0x3ff);
```

This is the standard UTF-16 surrogate pair decoding formula. The current code may work for some cases due to bit alignment, but will fail for others.

**Example:** `\uD800\uDC00` should decode to U+10000 (65536), but the current formula may produce incorrect results.

kdy1:
## 2. MEDIUM - Overly Permissive Character Acceptance

**Location:** `crates/swc_ecma_minifier/src/compress/pure/strings.rs:228-231`

The new logic accepts ALL characters when `force_str_for_tpl` is false:
```rust
if c.chars().all(|c| match c {
    '\n' | '\r' => self.config.force_str_for_tpl,
    _ => true,  // ⚠️ This accepts everything\!
}) {
```

**Problem:** This is less conservative than the old code which only accepted ASCII printable characters:
```rust
// Old code
'\u{0020}'..='\u{007e}' => true,
_ => false,
```

**Impact:** May convert template literals with special characters to strings when they shouldn't be, potentially breaking code with unusual Unicode characters or control characters.

**Recommendation:** Review if this change is intentional. If so, document why it's safe. Otherwise, restore the ASCII printable check or add proper validation.

kdy1:
## 3. MEDIUM - Unsafe Code Needs Better Validation

**Location:** `crates/swc_ecma_ast/src/lit.rs:398-400`

```rust
if result <= 0x10ffff {
    // We can be sure result is a valid code point here
    buf.push(unsafe { CodePoint::from_u32_unchecked(result) });
}
```

**Problem:** The check `result <= 0x10ffff` alone doesn't guarantee a valid code point. The surrogate range (0xD800-0xDFFF) is also invalid for direct code points.

**Should be:**
```rust
if result <= 0x10ffff && !(0xd800..=0xdfff).contains(&result) {
    buf.push(unsafe { CodePoint::from_u32_unchecked(result) });
} else {
    emit_span_error(span, "Uncaught SyntaxError: Undefined Unicode code-point");
}
```

Or handle the surrogate case separately before reaching this code. Please verify the safety invariants are correct.

kdy1:
## 4. LOW - Test Structure Issue

**Location:** `crates/swc_ecma_ast/tests/lit.rs:104-110`

```rust
#[test]
#[should_panic]
fn should_panic() {
    test_from_tpl_raw("\\01", "");
    test_from_tpl_raw("\\2", "");   // ⚠️ Never reached
    test_from_tpl_raw("\\7", "");   // ⚠️ Never reached
}
```

**Problem:** Only the first assertion will run because the function panics and exits. The other two assertions are never executed.

**Should be:**
```rust
#[test]
#[should_panic]
fn should_panic_octal_01() {
    test_from_tpl_raw("\\01", "");
}

#[test]
#[should_panic]
fn should_panic_octal_2() {
    test_from_tpl_raw("\\2", "");
}

#[test]
#[should_panic]
fn should_panic_octal_7() {
    test_from_tpl_raw("\\7", "");
}
```

This ensures all three octal cases are actually tested.

kdy1:
## 5. LOW - Misleading Comment

**Location:** `crates/swc_ecma_ast/src/lit.rs:383`

```rust
// Low surrogate pair
// result is less than 0x1ffff here
```

**Problem:** The comment is misleading. At this point, `result` is a low surrogate in the range 0xDC00-0xDFFF, not "less than 0x1ffff". 

After the calculation on line 384-385, the final result would be in the range 0x10000-0x10FFFF (supplementary planes), but the comment appears before that calculation.

**Suggested fix:**
```rust
// Low surrogate pair (0xDC00-0xDFFF)
// Decode to supplementary plane code point (0x10000-0x10FFFF)
```

kdy1:
## ✅ Positive Aspects of This PR

Despite the issues noted above, this PR has many strengths:

1. **Excellent Test Coverage:** The 11 test cases cover a wide range of edge cases including emoji, surrogate pairs, hex escapes, and invalid sequences. This is exactly what's needed for Unicode handling code.

2. **Proper Error Handling:** Using `emit_span_error` instead of panicking is the right approach for a production minifier. This allows the tool to report errors gracefully without crashing.

3. **WTF-8 Support:** Returning `Wtf8Atom` instead of `Atom` is crucial for preserving invalid Unicode sequences, which is important for minification correctness.

4. **Code Simplification:** Removing the unnecessary `convert_str_value_to_tpl_cooked` function and using `Cow::Borrowed` directly eliminates redundant conversions and improves performance.

5. **Comprehensive Escape Sequence Handling:** The implementation handles line continuations (`\r`, `\n`), Unicode escapes (`\u`, `\x`), and octal validation correctly.

Once the critical issues (especially #1) are addressed, this will be a solid improvement to the minifier. Great work on tackling this complex Unicode handling! 🦀

kdy1:
## Recommended Additional Testing

After fixing the issues, please add these test cases to verify correctness:

**1. Test for surrogate pair decoding (to catch issue #1):**
```rust
#[test]
fn surrogate_pair_boundary_cases() {
    // First supplementary plane character
    test_from_tpl_raw("\\uD800\\uDC00", "\u{10000}");
    
    // Last valid character
    test_from_tpl_raw("\\uDBFF\\uDFFF", "\u{10FFFF}");
    
    // Various combinations to ensure formula is correct
    test_from_tpl_raw("\\uD801\\uDC37", "\u{10437}"); // 𐐷
    test_from_tpl_raw("\\uD852\\uDF62", "\u{24B62}"); // 𤭢
}
```

**2. Test for special character handling (related to issue #2):**
```rust
#[test]
fn special_chars_in_tpl() {
    // Test that control characters are handled correctly
    let result = minify("`\\u0000\\u0001\\u001f`");
    // Verify behavior matches expectations
}
```

**3. Run execution tests:**
According to `crates/swc_ecma_minifier/AGENTS.md`, you should run:
```bash
./scripts/exec.sh
```

This will verify the changes don't break actual JavaScript execution.

cuyl:
> ## 3. MEDIUM - Unsafe Code Needs Better Validation
> **Location:** `crates/swc_ecma_ast/src/lit.rs:398-400`
> 
> ```rust
> if result <= 0x10ffff {
>     // We can be sure result is a valid code point here
>     buf.push(unsafe { CodePoint::from_u32_unchecked(result) });
> }
> ```
> 
> **Problem:** The check `result <= 0x10ffff` alone doesn't guarantee a valid code point. The surrogate range (0xD800-0xDFFF) is also invalid for direct code points.
> 
> **Should be:**
> 
> ```rust
> if result <= 0x10ffff && !(0xd800..=0xdfff).contains(&result) {
>     buf.push(unsafe { CodePoint::from_u32_unchecked(result) });
> } else {
>     emit_span_error(span, "Uncaught SyntaxError: Undefined Unicode code-point");
> }
> ```
> 
> Or handle the surrogate case separately before reaching this code. Please verify the safety invariants are correct.

Refer to `CodePoint::from_u32` `result <= 0x10ffff` is a valid code points range
https://github.com/swc-project/swc/blob/9632cd102778d5dc07cca2fa1ce590b9cbd5328e/crates/hstr/src/wtf8/mod.rs#L80-L85
And we have handled the case `(0xd800..=0xdfff).contains(&result)` before in `crates/swc_ecma_ast/src/lit.rs:362-391`

cuyl:
I will add the edge test cases

cuyl:
Thanks, already applied some changes.

> ## 2. MEDIUM - Overly Permissive Character Acceptance
> **Location:** `crates/swc_ecma_minifier/src/compress/pure/strings.rs:228-231`
> 
> The new logic accepts ALL characters when `force_str_for_tpl` is false:
> 
> ```rust
> if c.chars().all(|c| match c {
>     '\n' | '\r' => self.config.force_str_for_tpl,
>     _ => true,  // ⚠️ This accepts everything\!
> }) {
> ```
> 
> **Problem:** This is less conservative than the old code which only accepted ASCII printable characters:
> 
> ```rust
> // Old code
> '\u{0020}'..='\u{007e}' => true,
> _ => false,
> ```
> 
> **Impact:** May convert template literals with special characters to strings when they shouldn't be, potentially breaking code with unusual Unicode characters or control characters.
> 
> **Recommendation:** Review if this change is intentional. If so, document why it's safe. Otherwise, restore the ASCII printable check or add proper validation.

I think the new logic accepts ALL characters when `force_str_for_tpl` is `true`.
Again, I don't really know the reason for the old code, or why we have `force_str_for_tpl`. Do you have any opinion on this?
@kdy1

kdy1:
IIRC, `force_str_for_tpl` was necessary for `Evaluator` mainly because the evaluator didn't work well with `Expr::Tpl`. But as https://github.com/swc-project/swc/pull/11411 is merged, I don't think `force_str_for_tpl` is required anymore

cuyl:
So, Can I try to remove `force_str_for_tpl` in this PR?

kdy1:
Yeap, it would be nice. Thank you!

cuyl:
Seems `Optimizer::reduce_escaped_newline_for_str_lit` is a inverse operation of `Pure::convert_tpl_to_str`, added some code to prevent infinity loop.

cuyl:
Remove force_str_for_tpl is complicated, I would open another PR instead.

kdy1:
I see, it sounds good. Thanks!

cuyl:
OK, now force_str_for_tpl has been removed, and all tests passed

kdy1:
@claude Review this PR

crates/swc_ecma_minifier/src/compress/pure/strings.rs
```diff
@@ -438,7 +454,7 @@ impl Pure<'_> {
 
                     if let Some(cooked) = &mut r_first.cooked {
                         let mut c = Wtf8Buf::new();
-                        c.push_wtf8(&convert_str_value_to_tpl_cooked(&ls.value));
```
kdy1, line 441: I feel like the removal of `convert_str_value_to_tpl_cooked` is wrong, but need to verify

kdy1: I mean, the removal is fine if the logic is correct, but I feel like the new logic is wrong although I'm not sure

kdy1:
Thank you!

PR #11432, opened 2026-01-07
refactor(es/typescript): Convert TypeScript transforms to VisitMutHook pattern

kdy1:
## Summary
- Converted `StripType` and `StripImportExport` to implement `VisitMutHook<()>` instead of `VisitMut`
- Updated `TypeScript` transform to use `CompositeHook` and `VisitMutWithHook` for composition
- Made `strip_type` and `strip_import_export` modules public for hook usage
- Fixed namespace handling in `StripImportExport` to properly track depth and process nested modules

## Technical Details
This refactoring aligns with the architectural guidelines in `crates/swc_ecma_transformer/AGENTS.md`:
- Transform types implement `VisitMutHook<TraverseCtx>` instead of `VisitMut`
- Hooks are composed using `CompositeHook` and executed via `VisitMutWithHook`
- Main `Transformer` orchestrates hook execution

Key changes:
- `StripType`: Converted all `visit_mut_*` methods to `enter_*` and `exit_*` hook callbacks
- `StripImportExport`: Properly handles namespace depth tracking and processes module items before children
- Integration: Uses `CompositeHook` to chain `strip_import_export::hook()` and `strip_type::hook()`

## Test plan
- [ ] Run existing TypeScript transform tests
- [ ] Verify namespace and module import/export handling
- [ ] Check that type stripping works correctly
- [ ] Ensure no behavioral changes from the refactoring

### DIFF
--- a/crates/swc_ecma_transforms_typescript/src/typescript.rs
+++ b/crates/swc_ecma_transforms_typescript/src/typescript.rs
@@ -20,12 +20,12 @@ macro_rules! static_str {
 pub fn typescript(config: Config, unresolved_mark: Mark, top_level_mark: Mark) -> impl Pass {
     debug_assert_ne!(unresolved_mark, top_level_mark);
 
-    visit_mut_pass(TypeScript {
+    TypeScript {
         config,
         unresolved_mark,
         top_level_mark,
         id_usage: Default::default(),
-    })
+    }
 }
 
 pub fn strip(unresolved_mark: Mark, top_level_mark: Mark) -> impl Pass {
@@ -40,8 +40,8 @@ pub(crate) struct TypeScript {
     id_usage: FxHashSet<Id>,
 }
 
-impl VisitMut for TypeScript {
-    fn visit_mut_program(&mut self, n: &mut Program) {
+impl Pass for TypeScript {
+    fn process(&mut self, n: &mut Program) {
         let was_module = n.as_module().and_then(|m| self.get_last_module_span(m));
 
         let hook = CompositeHook {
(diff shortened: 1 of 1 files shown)


kdy1:
@magic-akari I'm working to merge the whole typescript pass into a single visit using `VisitMutHook`, and this is a first PR for it. Any concern about this PR?

magic-akari:
> @magic-akari I'm working to merge the whole typescript pass into a single visit using `VisitMutHook`, and this is a first PR for it. Any concern about this PR?

This isn’t specifically about this PR, but about the broader direction of this migration. For context, this split-visit design was something I originally suggested, so this is more of a retrospective concern on my side.

Over time, I’ve found that the oxc-style pattern of splitting a single visit into two parts can feel ergonomically awkward and tends to make code review harder. I’ve experimented with a few alternative approaches and suspect there may be cleaner solutions, although I haven’t landed on a concrete replacement yet.

I’m curious what your thoughts are on this design today. If better approaches emerge, would we be open to revisiting or evolving this direction?

kdy1:
Yeap, of course. My primary concern of this migration is the performance overhead of the visitor pattern. If there's a better alternative, I'm 100% open to migrating to it.

Beside it, I agree that this kind of visitor hook system has bad impact on code review. It has far more interactions between passes compared to standalone visitors, and that's why I'm trying to do it gradually, but doing it gradually wasn't enough for code review.

magic-akari:
> My primary concern of this migration is the performance overhead of the visitor pattern.

I was initially expecting this migration to bring some performance improvements, but looking at the Codespeed results, there's no measurable change. I'm curious to understand your specific concerns about the visitor pattern overhead.

kdy1:
I'll try merging more visitors in this PR to see if it really improves the performance.
There are too many visitors in typescript operation, and I only merged two of them

kdy1:
Closing due to conflicts

PR #11445, opened 2026-01-12
refactor(es/typescript): Reimplement TypeScript pass

kdy1:
**Description:**

I'll check the performance difference of a new implementation using codspeed


magic-akari:
I believe TypeScript strip requires a two-pass approach at minimum. The first pass should be read-only to collect metadata regarding enums, namespaces, and imports. A subsequent pass can then perform the actual transformation based on the information gathered.

magic-akari:
I noticed that the current implementation collects information top-down using `collect_local_decls`, `collect_imports`, `collect_value_usages`, and `collect_const_enum_values`, instead of leveraging the existing Visitor pattern we already have.

kdy1:
Thanks for pointing that out!

kdy1:
Closing due to conflicts

Issue #11435, opened 2026-01-07
TS enum get PURE comment even if enum values aren't pure


mischnic:
### Describe the bug

Transpiling a TS enum shouldn't lead to a PURE IIFE, if the enum value assignments contain side-effects.

Here's what esbuild does: https://esbuild.github.io/try/#YgAwLjI3LjIALS1idW5kbGUAAGEudHMAZXhwb3J0IGVudW0gQSB7CiAgeCA9IChsaXN0LnB1c2goJ2IgcHVyZScpLCAxMjMpLAp9CgAAYi50cwBleHBvcnQgZW51bSBCIHsKICB4ID0gMTIzLAp9CgBlAGluZGV4LmpzAGltcG9ydCB7QX0gZnJvbSAiLi9hLnRzIgppbXBvcnQge0J9IGZyb20gIi4vYi50cyIKCmNvbnNvbGUubG9nKEEsIEIp

### Input code

```typescript
enum Unused {
    x = (SIDE_EFFECT(), 123),
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2024",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    },
    "experimental": {
      "emitIsolatedDts": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Link to the code that reproduces this issue

https://play.swc.rs/?version=1.15.8&code=H4sIAAAAAAAAA0vNK81VCM0rLU5NUajmUgCCCgVbBY1gTxfXeFc3N1fnEA1NHQVDI2NNHa5aLgAIhj4cLgAAAA%3D%3D&config=H4sIAAAAAAAAA22QMQ6DMAxFd06BPHeoUNWhc5cOPUQEBqVKSGQbCYS4exNIKEjd4v%2F97O%2FMRVnCh2t4lHN4hsIrYqS9DgpPvagxKCCTR65Je4FLdoWj1SrDuErL5oAo6lAihVxdq1siwDjHmImkWd3rdjrurJ31hMznxtiq%2Bs7g3404eiRtMaQ1x1lotbzYGSXYPIXPbJF4sK4Z1sHpH%2BKtW%2Fo7%2FJpy0D0UaH5nUmjAYvkCP9KtD1MBAAA%3D

### Expected behavior

No PURE comment in this case

### Actual behavior

```
[...]

PR #11452, opened 2026-01-18
fix(es/typescript): Don't mark enums with opaque members as pure

taletski:
**Description:**

This draft PR adjusts the transform so that enums containing at least one opaque member are not marked as pure.  

This matches esbuild’s behavior and avoids marking expressions that may have side effects as pure.  

Feedback welcome before converting to a full PR.

**Related issue:** #11435

### DIFF
--- a/.changeset/famous-plants-dress.md
+++ b/.changeset/famous-plants-dress.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_transforms_typescript: patch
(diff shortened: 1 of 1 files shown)


taletski:
FYI: This is just a proposal. Happy to submit as a full PR if that makes sense.

kdy1 (approved):

kdy1:
Thank you!

PR #11459, opened 2026-01-21
feat(es/minifier): Inline function parameters with consistent values

### LINKED ISSUE #10931: `es/minifier`: Inline parameters
### Describe the feature

See [playground](https://play.swc.rs/?version=1.13.2&code=H4sIAAAAAAAAA33NsQ7CMBAD0D1fYXXKSShlYYWJkY8IIWkjpXfVkVIE4t8JAxti9bNskxYONQsjyDSXeLdJZIPEhKcB%2Bh6zxlvkiswlc%2BahpTnBnnwdnXq%2ByGQJe2zdjlBHlRUcVxxVRS21ssa6KLfFg%2Ftsk3kZE4SvUqIrMtjvb9ewI%2FptZ69%2F7NHsDUQbW4fJAAAA&config=H4sIAAAAAAAAA41VS47bMAzd5xSB17OYDopB0QMU3fQMgmJRjlJZEkTKE2OQu5f%2BBc6ENrqz%2BcSPHh%2Bpz8PxWF2wrn4eP%2FmTf5LOCPn%2BzxbsA%2BkrWyqoW411domqlwW94ABZ7RFG021CKh8jwoLMttYFZ%2Ft17Dq2KQPiysZWnZvSQqDBTLnAyxrK8UOwn2L0oMMOojQqFwgayPhY11yJ9zohqE5nIcpQqM4Oo5RiAAuBUSnHJOLBOHIxcM5n1IA2qo4GBMhlqMl1ILlxLnYLyNcT7jPCBk6lacZ2fvGGTvuiScgJ17EjXK0Q9RwdkrIlSBRO4AYHEziT%2B9XTWZWBSg7PfpfowkZP%2FgIwA14jBt2CFHc8YVlOW95219MFy4qlXsBZ3tItAzRMqnLOCswOzEAmJ3Uzgyk1DMzWUjkzvEEfOgMKrGWtCKHxw1F9lpJSnyBaAeD%2BaiupagLUfQo38GEgduBffEuaBCYmVq2m87Y79u0p%2Bp34LdA5mp0D3AmK23DmJXFN23gJBlgZYMQjBUfgeQew%2FikqDyRJg6eDI6rGx5O4JRLLfOzh%2B2y73Vdwq0Mz7oDVCqWYPHTgt6T9H2Ozi3LB3aD0ZdqfJhp%2BSELVPIzu2%2BvDi8FXeXg54EqQeVX%2BBp%2BednVFPM0jgxXg2%2Bvb9%2Bowe1ZtNGVFw6ju6dz76tDyBC30Vg7%2FLI6z7XD7Byzd7fcXBwAA)

```
function complex(foo, fn) {
  // prevent inlining
  if (Math.random() > 0.5) throw new Error()
  return fn?.(foo)
}

console.log(complex("foo"))
console.log(complex("bar"))
console.log(complex("baz"))
```

If i remove the `if (Math.random)...` line then SWC will just inline this function as `void 0` at all callsites, however, if the function is not a trivial inlining candidate it backs off and preserves the function pretty much as is.

In that case, i would like swc to just inline the _parameter_, all callsites are implicitly passing `undefined` to the `fn` parameter.  If the minifier could delete that parameter and add a variable `const fn  = undefined` then another minifier pass would just simplify the `return` statement to `return undefined`.

This example is of course somewhat trivial, but my actual usecase is wanting to pass an optional callback through a few layers of shared functions.  If we could inline `undefined` when not passed then it would allow this callback to be eliminated altogether in the common case where it isn't passed.

The closure compiler has a pass like this: https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/OptimizeParameters.java

The semantics are that parameters will get inlined if all callsites pass the same value and that value is side effect free and 'not too big'.   There was a once a code size regression where a large string literal got inlined in a way that moved it into an earlier chunk.  I don't think swc needs to worry about that since it doesn't support cross chunk optimizations (does it?)

Since swc is happy to potentially inline this function and delete it, it could perhaps also be happy to modify the signature in the same situations.

_No response_

### Additional context

I was hoping for something like this as i am working on the turbopack runtime and there are a number of hooks needed by HMR that are not needed in production.  So to share more code i was hoping to have common functions take optional callbacks that could just be trimmed in the production usecase.

kdy1:
When all call sites of a function pass the same value (or undefined) to a parameter, this optimization converts the parameter to a local variable initialized with that value.

Closes #10931

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/optimize/mod.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/mod.rs
@@ -2112,6 +2112,10 @@ impl VisitMut for Optimizer<'_> {
 
         self.drop_unused_params(&mut f.function.params);
 
+        // Inline parameters that have consistent values across all call sites
+        let fn_id = f.ident.to_id();
+        self.inline_params_with_same_value(&fn_id, &mut f.function.params, &mut f.function.body);
+
         let ctx = self
             .ctx
             .clone()
--- a/crates/swc_ecma_minifier/src/compress/optimize/unused.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/unused.rs
@@ -14,7 +14,7 @@ use crate::debug::dump;
 use crate::{
     compress::optimize::{util::extract_class_side_effect, BitCtx},
     option::PureGetterOption,
-    program_data::{ScopeData, VarUsageInfoFlags},
+    program_data::{ParamInlineValue, ScopeData, VarUsageInfoFlags},
 };
 
 #[derive(Debug, Default, Clone, Copy)]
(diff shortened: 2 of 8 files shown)


kdy1:
Does not look correct

PR #11489, opened 2026-01-23
fix(es/module): preserve ./ prefix for hidden directory imports

### LINKED ISSUE #9551: SWC `transform` with `baseUrl` mangles imports of hidden directories
### Describe the bug

When `transform` is given a `baseUrl`, relative and base-relative imports of hidden (dot-prefixed) directories become mangled. The examples in the reproduction that are incorrect are `./.foo` becoming `.foo`, and `src/.foo` become `.foo`. Obviously `.foo` cannot be resolved from the `src/index.js` file. Unless I'm misunderstading how `baseUrl` works, this seems incorrect to me.

I was unable to reproduce this inside the SWC playground as the bug depends on there being a filesystem. I suppose there's some resolution going on inside SWC when `baseUrl` is provided. Deleting the `.foo` directory results in neither of the imports to the `.foo` directory being altered, regardless of whether or not `baseUrl` is provided. This is still the case for a regular directory like `foo` though, so I suppose this behaviour is at least consistent between hidden and non-hidden directories.

### Input code

_No response_

### Config

_No response_

https://github.com/askoufis/swc-baseurl-bug

### SWC Info output

Operating System:
        Platform: darwin
        Arch: arm64
        Machine Type: arm64
        Version: Darwin Kernel Version 23.6.0: Mon Jul 29 21:14:30 PDT 2024; root:xnu-10063.141.2~1/RELEASE_ARM64_T6000
        CPU: (10 cores)
            Models: Apple M1 Pro

    Binaries:
        Node: 20.17.0
        npm: 10.8.2
        Yarn: N/A
        pnpm: 9.10.0

    Relevant Packages:
        @swc/core: 1.7.26
        @swc/helpers: N/A
        @swc/types: N/A

    SWC Config:
        output: N/A
        .swcrc path: N/A

    Next.js info:
        output: N/A

### Expected behavior

Imports to hidden (dot-prefixed) directories should not be mangled when a `baseUrl` is present. In the reproduction I would expect the `./.foo` import path to be unchanged, and the `src/.foo` import path to be converted to `./.foo`. 

### Actual behavior

_No response_

### Version

1.7.26

### Additional context

_No response_

kdy1:
When resolving imports with `baseUrl`, hidden directories (dot-prefixed like `.foo`) were incorrectly losing their `./` prefix, turning `./.foo` into `.foo` which is not a valid relative import.

The issue was that the code checked `starts_with('.')` to determine if a path was already relative, but this matched hidden directory names (which start with `.`) even though they need a `./` prefix to be valid relative paths.

The fix changes the check to look for actual relative path markers (`./` or `../`) instead of just `.`.

Fixes #9551

### DIFF
--- a/.changeset/wicked-pens-smell.md
+++ b/.changeset/wicked-pens-smell.md
@@ -1,0 +1,7 @@
+---
+swc_ecma_transforms_module: patch
+swc_ecma_transforms_proposal: patch
+swc_core: patch
+---
+
+fix(es/module): preserve ./ prefix for hidden directory imports


kdy1:
@claude Fix CI

PR #11493, opened 2026-01-23
feat(es/minifier): Add `unsafe_hoist_static_method_alias` option

### LINKED ISSUE #9741: [Minification] Hoist static methods with alias
### Describe the feature

Given the following code:

```ts
const a = {};
Object.assign(a, {});
const b = {};
Object.assign(b, {});
Object.assign(b, a);
```

Currently both terser and swc can only minified the code into

```js
const a={};Object.assign(a,{});const b={};Object.assign(b,{}),Object.assign(b,a);
```

Here is my proposal, a new compress option `unsafe_hoist_static_method_alias`. In the compress phase, swc counts and replaces static built-in object methods with aliases:

```ts
const __ObjectAssign = Object.assign;

const a = {};
__ObjectAssign(a, {});
const b = {};
__ObjectAssign(b, {});
__ObjectAssign(b, a);
```

In the mangle phase, swc can mangle `__ObjectAssign` into a shorter name:

```js
const s=Object.assign,c={};s(c,{});const n={};s(n,{}),s(n,c);
```

Note that this optimization should only applied to built-in objects' known static methods (`Object.assign`, `Reflect.apply`, etc.). Also since overriding built-in objects' static methods is possible, this option should be prefixed with `unsafe` and should not be enabled by default.

_No response_

### Additional context

_No response_

kdy1:
## Summary

This PR adds two new minifier compress options for hoisting frequently-used built-in references to local aliases, enabling better minification after mangling.

#### `unsafe_hoist_static_method_alias` (default: `false`)

Hoists frequently-used static method calls to local variables:

```js
// Before
Object.assign(a, {}); Object.assign(b, {}); Object.assign(c, {});

// After
var _Object_assign = Object.assign;
_Object_assign(a, {}); _Object_assign(b, {}); _Object_assign(c, {});
```

**Supported built-ins:**
- **Object**: `assign`, `keys`, `values`, `entries`, `fromEntries`, `freeze`, `seal`, `create`, `defineProperty`, `defineProperties`, `getOwnPropertyDescriptor`, `getOwnPropertyDescriptors`, `getOwnPropertyNames`, `getOwnPropertySymbols`, `getPrototypeOf`, `setPrototypeOf`, `preventExtensions`, `isFrozen`, `isSealed`, `isExtensible`, `hasOwn`, `is`
- **Array**: `isArray`, `from`, `of`
- **Reflect**: `apply`, `construct`, `defineProperty`, `deleteProperty`, `get`, `getOwnPropertyDescriptor`, `getPrototypeOf`, `has`, `isExtensible`, `ownKeys`, `preventExtensions`, `set`, `setPrototypeOf`
- **Math**: `abs`, `acos`, `acosh`, `asin`, `asinh`, `atan`, `atan2`, `atanh`, `cbrt`, `ceil`, `clz32`, `cos`, `cosh`, `exp`, `expm1`, `floor`, `fround`, `hypot`, `imul`, `log`, `log10`, `log1p`, `log2`, `max`, `min`, `pow`, `random`, `round`, `sign`, `sin`, `sinh`, `sqrt`, `tan`, `tanh`, `trunc`
- **String**: `fromCharCode`, `fromCodePoint`, `raw`
- **Number**: `isFinite`, `isInteger`, `isNaN`, `isSafeInteger`, `parseFloat`, `parseInt`
- **JSON**: `parse`, `stringify`
- **Promise**: `all`, `allSettled`, `any`, `race`, `reject`, `resolve`, `withResolvers`
- **Symbol**: `for`, `keyFor`

#### `unsafe_hoist_global_objects_alias` (default: `false`)

Hoists frequently-used global object constructor calls to local variables:

```js
// Before
new Map(); new Map(); new Map();

// After
var _Map = Map;
new _Map(); new _Map(); new _Map();
```

**Supported global objects:**
- Collections: `Map`, `Set`, `WeakMap`, `WeakSet`
- Async: `Promise`, `Proxy`
- Types: `Date`, `RegExp`, `URL`, `URLSearchParams`
- Errors: `Error`, `TypeError`, `RangeError`, `SyntaxError`, `ReferenceError`, `URIError`, `EvalError`
- Typed Arrays: `ArrayBuffer`, `SharedArrayBuffer`, `DataView`, `Int8Array`, `Uint8Array`, `Uint8ClampedArray`, `Int16Array`, `Uint16Array`, `Int32Array`, `Uint32Array`, `Float32Array`, `Float64Array`, `BigInt64Array`, `BigUint64Array`
- Text: `TextEncoder`, `TextDecoder`

### Smart Thresholds

The optimizer uses a cost-benefit formula to determine when hoisting is worthwhile:

```
min_usages = ceil((name_len + 7) / (name_len - 1))
```

- Every hoist costs `name_len + 7` bytes (e.g., `var a=Set;` = 10 bytes)
- Every usage saves `name_len - 1` bytes (e.g., `new a()` vs `new Set()` saves 2 bytes)

Examples:
- `Object.assign` (13 chars): needs 2+ usages
- `Math.cos` (8 chars): needs 3+ usages  
- `Set` (3 chars): needs 5+ usages

### Why "unsafe"?

These options are marked as unsafe because they assume built-in globals haven't been modified. If user code overrides `Object.assign` or `Map` after the hoisted alias is captured, the aliased calls will still use the original built-in.

### Implementation

The static alias optimization is integrated directly into the `Optimizer` (following the SWC way of avoiding separate visitors):

- **`StaticAliasState`**: Stored in `Compressor` and passed to `Optimizer` by reference, persisting state across optimization passes
- **Pass 1**: Counts usages of static method calls and global object constructors during `visit_mut_call_expr` and `visit_mut_new_expr`. At the end of module/script processing, builds aliases for methods/objects that exceed the threshold.
- **Pass 2+**: Replaces call/new expressions with aliases. Inserts var declarations at the top of the module/script.

Uses `private_ident!` macro for hygiene-safe identifier generation, avoiding name collisions.

## Test Plan

- [x] Added 7 fixture tests covering:
  - Basic static method hoisting (`9741`)
  - Threshold behavior (`9741_threshold`)
  - Multiple methods (`9741_multiple_methods`)
  - Disabled mode (`9741_disabled`)
  - Variable collision handling (`9741_collision`)
  - Function collision handling (`9741_collision_function`)
  - Global object hoisting (`9741_global_objects`)
- [x] Added 4 execution tests for runtime correctness
- [x] All 2,116 execution tests pass
- [x] All 2,202 fixture tests pass

Closes #9741

---
🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/optimize/static_alias.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/static_alias.rs
@@ -34,13 +34,23 @@
 //! ```
 //!
 //! After mangling, this can result in smaller output.
+//!
+//! ## Integration with Optimizer
+//!
+//! This module integrates directly into the `Optimizer` rather than using
+//! separate visitors. The Optimizer calls helper methods during its traversal:
+//! - `try_record_static_method_call` during `visit_mut_call_expr`
+//! - `try_record_global_object` during `visit_mut_new_expr`
+//! - `try_replace_static_method_call` during `visit_mut_call_expr`
+//! - `try_replace_global_object` during `visit_mut_new_expr`
+//! - `build_aliases_from_counts` at the end of module/script processing
+//! - `take_alias_var_decls` to get var declarations for insertion
 
 use rustc_hash::FxHashMap;
 use swc_atoms::Atom;
 use swc_common::{SyntaxContext, DUMMY_SP};
 use swc_ecma_ast::*;
 use swc_ecma_utils::private_ident;
-use swc_ecma_visit::{noop_visit_type, Visit, VisitWith};
 
 /// Calculate the minimum number of usages needed to save bytes after hoisting.
 ///
@@ -387,55 +397,61 @@ fn get_global_object_min_usages(name: &str) -> Option<usize> {
 /// Key for tracking static method usage: (object_name, method_name)
 pub type StaticMethodKey = (Atom, Atom);
 
-/// Counter for static method and global object usages.
-pub struct UsageCounter {
-    unresolved_ctxt: SyntaxContext,
+/// State for static alias optimization, stored in Optimizer.
+#[derive(Default)]
+pub struct StaticAliasState {
     /// Count of usages for each static method
-    static_method_counts: FxHashMap<StaticMethodKey, usize>,
+    pub static_method_counts: FxHashMap<StaticMethodKey, usize>,
     /// Count of usages for each global object
-    global_object_counts: FxHashMap<Atom, usize>,
-    /// Whether to count static methods
-    count_static_methods: bool,
-    /// Whether to count global objects
-    count_global_objects: bool,
+    pub global_object_counts: FxHashMap<Atom, usize>,
+    /// Aliases for static methods (built after counting)
+    pub static_method_aliases: FxHashMap<StaticMethodKey, Ident>,
+    /// Aliases for global objects (built after counting)
+    pub global_object_aliases: FxHashMap<Atom, Ident>,
+    /// Var declarations to prepend (built after counting)
+    pub var_decls: Vec<VarDeclarator>,
+    /// Whether aliases have been built (happens once per program)
+    pub aliases_built: bool,
+    /// Whether var declarations have been inserted
+    pub var_decls_inserted: bool,
+    /// Whether any replacements were made (in this pass)
+    pub replacements_made: bool,
 }
 
-impl UsageCounter {
-    fn new(
+impl StaticAliasState {
+    /// Record a static method call usage.
+    /// Called during Optimizer's visit_mut_call_expr.
+    pub fn try_record_static_method_call(
+        &mut self,
+        callee: &Expr,
         unresolved_ctxt: SyntaxContext,
-        count_static_methods: bool,
-        count_global_objects: bool,
-    ) -> Self {
-        Self {
-            unresolved_ctxt,
-            static_method_counts: Default::default(),
-            global_object_counts: Default::default(),
-            count_static_methods,
-            count_global_objects,
-        }
-    }
-
-    fn try_record_static_method_call(&mut self, callee: &Expr) {
-        if !self.count_static_methods {
+        enabled: bool,
+    ) {
+        if !enabled || self.aliases_built {
             return;
         }
-        if let Some((obj, prop)) = extract_static_method(callee, self.unresolved_ctxt) {
-            // Only record if this is a known static method (lookup returns Some)
+        if let Some((obj, prop)) = extract_static_method(callee, unresolved_ctxt) {
+            // Only record if this is a known static method
             if get_static_method_min_usages(&obj, &prop).is_some() {
                 *self.static_method_counts.entry((obj, prop)).or_insert(0) += 1;
             }
         }
     }
 
-    fn try_record_global_object(&mut self, callee: &Expr) {
-        if !self.count_global_objects {
+    /// Record a global object usage (from new expression).
+    /// Called during Optimizer's visit_mut_new_expr.
+    pub fn try_record_global_object(
+        &mut self,
+        callee: &Expr,
+        unresolved_ctxt: SyntaxContext,
+        enabled: bool,
+    ) {
+        if !enabled || self.aliases_built {
             return;
         }
         if let Some(ident) = callee.as_ident() {
-            // Only record if this is a known global object (lookup returns Some)
-            if ident.ctxt == self.unresolved_ctxt
-                && get_global_object_min_usages(&ident.sym).is_some()
-            {
+            // Only record if this is a known global object
+            if ident.ctxt == unresolved_ctxt && get_global_object_min_usages(&ident.sym).is_some() {
                 *self
                     .global_object_counts
                     .entry(ident.sym.clone())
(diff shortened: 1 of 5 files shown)


kdy1:
@claude Add test cases for issue described in https://github.com/swc-project/swc/pull/11493#pullrequestreview-3703502429

kdy1:
@claude Add another option as described in https://github.com/swc-project/swc/issues/9741#issuecomment-3796241054

crates/swc_ecma_minifier/tests/fixture/issues/9741_collision/input.js
```diff
@@ -0,0 +1,7 @@
+const _Object_assign = [];
+
+const a = {};
+Object.assign(a, {});
+const b = {};
+Object.assign(b, {});
+Object.assign(b, a);
```
SukkaW, line 7: My orignal review commment might not be a good test case against this. The proper test case would be:

```js
function _Object_assign(a, b) {
  console.log(this, Math.exp(a, b));
};
_Object_assign(4, 2);

const a = {};
Object.assign(a, {});
const b = {};
Object.assign(b, {});
```

kdy1: @claude add one

kdy1:
@claude Remove identifier collector and rely on the SWC way (https://swc.rs/docs/contributing/es-commons/variable-management) for managing conflicts. i.e. Use `private_ident!()` to create a new private identifier.

crates/swc_ecma_minifier/tests/fixture/issues/9741_collision/input.js
SukkaW: @kdy1 Does swc have a built-in variable collision renaming handling implementation yet? IIRC, swc might have one already. If so, might just reuse that.

kdy1: I already did it :)
You can scroll down to see it.

crates/swc_ecma_minifier/src/pass/precompress.rs
```diff
@@ -0,0 +1,457 @@
...
+        let unresolved_ctxt = SyntaxContext::empty().apply_mark(unresolved_mark);
+        let mut static_method_aliases = FxHashMap::default();
+        let mut global_object_aliases = FxHashMap::default();
+        let mut var_decls = Vec::new();
+
+        // Handle static method aliases
+        for ((obj, prop), count) in static_method_counts {
+            if count >= MIN_USAGES_TO_HOIST {
+                // Create a private identifier - hygiene pass will handle conflicts
+                let alias_ident = private_ident!(format!("_{}_{}", obj, prop));
```
SukkaW: IMHO, it is possible to somewhat precisely estimate the proper threshold to avoid wastes.

Every static method usage that get hoisted then mangled saves at most by
"original method name length minus 1" bytes.

E.g., the length of `Object.assign` is 13 bytes, so every `Object.assign(a, b)` that get mangled to `m(a, b)` saves 12 bytes (13 - 1). And the mangled name can become more characters if we have enough variables lying around.

But on the other hand, every hoist itself costs at least "original method name length plus 7" bytes.

E.g, `let m=Object.assign;` costs 20 bytes (13 + 7).

With this simple estimation, we can easily calculate the proper threshold for other methods.

E.g., the length of `Math.cos` is 8 bytes, thus every alias saves 7 bytes (8 - 1), and the hoist itself costs 15 bytes (8 + 7), so we need at least 3 usages (instead of 2) would we ever save bytes.

crates/swc_ecma_minifier/src/pass/precompress.rs
```diff
@@ -0,0 +1,457 @@
...
+                    definite: false,
+                });
+
+                static_method_aliases.insert((obj, prop), alias_ident);
+            }
+        }
+
+        // Handle global object aliases
+        for (name, count) in global_object_counts {
+            if count >= MIN_USAGES_TO_HOIST {
```
SukkaW: The same estimate and equation can be added to here as well.

Every global objects we alias also saves at most "global object name length minus 1", i.e. for every `new Set()` becomes `new a()` it saves 2 bytes (3 - 1).

And every hoist itself also costs at least "global object name length plus 7" bytes, i.e. `let a=Set;` costs 10 bytes (3 + 7).

The numbers (minus 1 and plus 7) are exactly the same here.

So, at least 4 `Set` usage would we ever seen any saving bytes.

crates/swc_ecma_minifier/src/pass/precompress.rs
SukkaW: The equation would be:

```
let x = global object or static method name len
let min_usage = (x + 7) / (x - 1)
```

kdy1:
@claude Validate and reflect if legit: https://github.com/swc-project/swc/pull/11493#pullrequestreview-3706877813 and https://github.com/swc-project/swc/pull/11493#pullrequestreview-3706897319

kdy1:
@claude Reflect https://github.com/swc-project/swc/pull/11493#pullrequestreview-3706877813 and https://github.com/swc-project/swc/pull/11493#pullrequestreview-3706897319

crates/swc_ecma_minifier/src/pass/precompress.rs
```diff
@@ -0,0 +1,479 @@
...
+/// - `Object.assign` (13 chars): (13+7)/(13-1) = 20/12 = 2 usages
+/// - `Math.cos` (8 chars): (8+7)/(8-1) = 15/7 = 3 usages
+/// - `Set` (3 chars): (3+7)/(3-1) = 10/2 = 5 usages
+fn min_usages_for_name(name_len: usize) -> usize {
+    if name_len <= 1 {
+        return usize::MAX; // Never hoist single-char names
+    }
+    // ceil((name_len + 7) / (name_len - 1))
+    (name_len + 7 + (name_len - 2)) / (name_len - 1)
+}
```
SukkaW: A potential very small performance gain, don't know if rustc handles and inlines this already or not, but since all alias-able/hoist-able Global Object and Static Methods are already known, we can potentially pre-populate the min usage count instead of calling `min_usages_for_name` on demand in precompress phase.

kdy1:
@claude Implement the idea described in https://github.com/swc-project/swc/pull/11493#pullrequestreview-3712204373
Verify the idea before implementing it

SukkaW (approved):
LGTM overall. @kdy1 Would you mind reviewing this from an experienced Rust programmer's perspective and land this?

kdy1:
I'll review this on Monday

kdy1 (commented):
@claude Remove precompess optimizer and merge it into the stateful optimizer

kdy1:
@claude fix CI

kdy1:
@claude Update the PR body to summarize changes

crates/swc_ecma_minifier/src/compress/optimize/static_alias.rs
kdy1, line 1: Remove new implementation of `Visit` and `VisitMut`. Visitors are slow

PR #11509, opened 2026-01-27
feat(swc): Make module transforms optional via `module` feature

### LINKED ISSUE #11505: Make it easy to exclude `swc_ecma_transforms_module` from the final binary
By default, it should not be included even for the users who use the `swc` crate. SWC is widely used by the bundler projects, and the module transforms are completely useless for them. `swc_cli` and `bindings/core_*` should explicitly enable the module transform of the `swc` crate.

kdy1:
## Summary

Make `swc_ecma_transforms_module` an optional dependency of the `swc` crate. Bundlers typically handle module transforms themselves, so they don't need this code, reducing binary size.

## Changes

- Add `module` feature to `swc` crate that enables module transforms
- Gate all module-related code with `#[cfg(feature = "module")]`
- Add `base_module` feature to `swc_core` that enables `swc/module`
- Update `swc_cli_impl`, `binding_core_node`, and `binding_core_wasm` to explicitly enable the module feature

Closes #11505

### DIFF
--- a/.changeset/friendly-scissors-grab.md
+++ b/.changeset/friendly-scissors-grab.md
@@ -1,0 +1,5 @@
+---
+swc_ecma_transforms_module: major
+---
+
+feat(swc): Make module transforms optional via `module` feature


kdy1:
@claude fix ci

kdy1:
@claude fix ci

PR #11529, opened 2026-02-04
fix(es/minifier): Prevent convert_tpl_to_str when there's emoji under es5

CPunisher:
**Description:**
When the target environment is below ES2015, and non-BMP characters (like emojis) are encountered, we stop the conversion.

This is because:
1. Tpl: `🦀` (may output directly in source code or require minimal escaping) -> shorter
2. Str (in ES5 mode): `\uD83E\uDD80` (escape sequence for surrogate pair) -> extremely long

https://github.com/swc-project/swc/pull/11415


kdy1:
@claude Add enormous amount of unit tests in the swc_ecma_minifier that actually invokes swc_ecma_codegen to verify all possibly problematic input cases

kdy1:
This PR looks good to me, but I'd prefer to have more e2e tests for this.

kdy1:
@claude fix ci

PR #11532, opened 2026-02-06
refactor(es/typescript): Run typescript transform in two passes

kdy1:
## Summary
- reworked swc_ecma_transforms_typescript to run in two passes: one semantic analysis pass and one transform pass
- added semantic.rs to collect usage/type-value/export/enum information in a single analysis traversal
- updated the transform pass to consume semantic info and perform strip/runtime transformations in one traversal
- rewired TypeScript entrypoints to use the new pipeline

## Validation
- cargo fmt --all
- cargo test -p swc_ecma_transforms_typescript -p swc_ecma_transforms_react -p swc_ecma_transforms
- cargo test -p swc --test projects --test tsc

### DIFF
--- a/crates/swc_ecma_transforms_typescript/src/transform.rs
+++ b/crates/swc_ecma_transforms_typescript/src/transform.rs
@@ -77,6 +77,7 @@ pub(crate) struct Transform {
     in_class_prop_init: Vec<Box<Expr>>,
 }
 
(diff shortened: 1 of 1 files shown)


kdy1:
@claude Fix cargo clippy. Run `cargo fmt --all` after then

PR #11533, opened 2026-02-06
refactor(es/react): Inline dev jsx metadata into jsx hook

kdy1:
## Summary
- inline development JSX metadata generation (__source, __self) into the main jsx hook
- simplify react() hook composition by removing separate jsx_src and jsx_self passes from the react pipeline
- preserve automatic-runtime createElement fallback behavior for source/self metadata and constructor-safe this handling

## Testing
- cargo fmt --all
- cargo test -p swc_ecma_transforms_typescript -p swc_ecma_transforms_react -p swc_ecma_transforms
- cargo test -p swc --test projects --test tsc

### DIFF
--- a/crates/swc_ecma_transforms_react/src/jsx/mod.rs
+++ b/crates/swc_ecma_transforms_react/src/jsx/mod.rs
@@ -245,6 +245,7 @@ where
             .throw_if_namespace
             .unwrap_or_else(default_throw_if_namespace),
         top_level_node: true,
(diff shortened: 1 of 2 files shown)


kdy1:
I expected more perf gain, but it seems like this is not very helpful

kdy1:
Reopening as it has improvement of 1%

kdy1:
I'll review this later

PR #11546, opened 2026-02-14
fix(es/minifier): Prevent incorrect array.join("") to string concat

### LINKED ISSUE #11545: Incorrect transformation of array join with conditional undefined element when compression + evaluate are enabled
### Describe the bug

When "compress" is enabled with "evaluate", code that involves a join of an array with an element that may be `undefined` is transformed incorrectly to stringify `undefined` rather than coerce it to the empty string. According to the spec, `undefined` (and `null`) elements should be converted into the empty string before joining ([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join#description)). However, with the SWC transformation, something like `["abc", boolean ? undefined : "def"]` will get converted into `"abc" + (boolean ? void 0 : "def")`, which is incorrect because this will perform coercion of `undefined` to `"undefined"`, resulting in `"abcundefined"` instead of just `"abc"`. This behavior seems to only rely on the "evaluate" setting of "compress".

### Input code

```typescript
function joinArrayWithUndefined(bool) {
  return ["abc", bool ? undefined : "def"].join("");
}

console.log(joinArrayWithUndefined(true))
console.log(joinArrayWithUndefined(false))

const x = ["abc", undefined].join("");
console.log(x);
```

### Config

```json
{
  "swcrc": false,
  "module": {
    "type": "commonjs"
  },
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es5",
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": false,
        "booleans": false,
        "collapse_vars": false,
        "comparisons": false,
        "computed_props": false,
        "conditionals": false,
        "dead_code": false,
        "directives": false,
        "drop_console": false,
        "drop_debugger": false,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": false,
        "hoist_vars": false,
        "if_return": false,
        "join_vars": false,
        "keep_classnames": false,
        "keep_fargs": false,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": false,
        "negate_iife": false,
        "properties": false,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": false,
        "switches": false,
        "typeofs": false,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": false,
        "const_to_let": false,
        "pristine_globals": false
      },
      "mangle": false
    },
    "loose": false,
  },
  "isModule": true,
  "minify": false
}
```

https://play.swc.rs/?version=1.15.11&code=H4sIAAAAAAAAA42OQQrCMBRE9znF8FcJSA9gEfEQ4kK6aNNEI%2BF%2FSBOoiHfXCJZuBHcD83gzyhe2OQjjJoEPKfX3U8jXI4%2FOB3ajHkSiwUMByeWSGGfqB0sb1AJ7lC%2BJLegdqWuqSROZVj2VssKTRNdEuegfEzkVZ8w%2FpO%2FjVNEPmzFjt9xZfqz318rZtC%2BgzhH57QAAAA%3D%3D&config=H4sIAAAAAAAAA32VObLbMAyG%2B3cKj2q3aXKAdDkDhyYhmQ5FaAjQtuaN7x5Qi5cEUkl8%2BAkRm76%2FDoeGbi675uehtZHgWC09%2BhJBTN9ykjOPQz01Dvse04UaMT8mzwu5l9tgM0F%2BnuvVY2J7r9J6BbkcBm6OK2W6r2En0%2BO4hLO5A64qoB%2BLe9OHFNrx%2FXL5miED0ZtNrKItPSSm9xc9WcabBk6IEWzaQ8aSCYmhg6y5OYzRDgTmajd4L%2BkJhGqQSguDN0PGQXdIPnDAJHYFe7DeOPSgsZDBcbiCKpR4IkyEUdVW7OFUum4q7L8crjYWy1XKuXyQ%2B1Qb%2BWJFdsZAbNqi5mKGW4mY6UaSQ2sycMlazAuGtKX7AyBZiJYo2V7N0%2BTRSm9tw11pSK30L48Kj6i%2FNEEnmTUhtFplan4gc1BDZvDFQU2w28EbyaDgwUDbStOo%2BBbYndWwdcax1YhU2qrPmIHZHMyF1%2FnYwb%2Fkoay32uLRWz5vUxr7E6qDtcqBz%2Bh3HKQcjNs4y9q4D9u8JA%2FSIOBVl0IqqGPLhtHEaVn%2B3yEyKXKn6SKe3tbG4vF4ruHepu41%2FR%2BbWHqTPso2b%2F1Av9c%2FxDr2r%2F083%2FL4%2BguCkD%2F2XAYAAA%3D%3D

### SWC Info output

_No response_

### Expected behavior

The output code should correctly handle `undefined` elements in array joins and turn them into the empty strings.

### Actual behavior

_No response_

### Version

1.15.11

### Additional context

_No response_

veeceey:
Fixes #11545

When `compress.evaluate` is enabled, `compress_array_join_partial` converts `[...].join("")` to string concatenation with `+`. This is incorrect when array elements can be `undefined` or `null`, because `Array.prototype.join` converts these to empty strings, while `+` converts them to `"undefined"` / `"null"`.

For example:
```js
["abc", cond ? undefined : "def"].join("")
// Expected when cond=true: "abc"
// Was producing: "abcundefined" (via "abc" + (cond ? void 0 : "def"))
```

The fix adds a check in the empty-separator path of `compress_array_join_partial` that bails out when non-literal expressions aren't guaranteed to be strings (checked via `is_definitely_string`). This prevents the semantics-changing transformation while still allowing the optimization for safe cases like template literals and known string concatenations.

All 471 exec tests pass, including 2 new tests covering the exact scenario from the issue.

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/pure/misc.rs
+++ b/crates/swc_ecma_minifier/src/compress/pure/misc.rs
@@ -49,6 +49,7 @@ fn may_produce_null_or_undefined(expr_ctx: ExprCtx, expr: &Expr) -> bool {
             .last()
             .is_some_and(|e| may_produce_null_or_undefined(expr_ctx, e)),
         Expr::Paren(ParenExpr { expr, .. }) => may_produce_null_or_undefined(expr_ctx, expr),
+        Expr::Await(AwaitExpr { arg, .. }) => may_produce_null_or_undefined(expr_ctx, arg),
         _ => false,
     }
 }


kdy1 (changes requested):
CI failed. Can you fix it?
You can do `UPDATE=1 cargo test` from the minifier directory or `swc` directory to update the test refernces

veeceey:
Fixed the clippy issue — changed `.map_or(false, ...)` to `.is_some_and(...)`. Should be good now, thanks for the heads up!

veeceey:
CI is green now after the clippy fix — switched to is_some_and() as suggested. Would appreciate a re-review when you get a chance!

veeceey:
CI is green now after the clippy fix. Would you mind taking another look?

crates/swc_ecma_minifier/src/compress/pure/misc.rs
```diff
@@ -33,6 +33,26 @@ fn is_definitely_string(expr: &Expr) -> bool {
...
+    match expr {
+        Expr::Cond(CondExpr { cons, alt, .. }) => {
+            may_produce_null_or_undefined(expr_ctx, cons)
+                || may_produce_null_or_undefined(expr_ctx, alt)
+        }
+        Expr::Seq(SeqExpr { exprs, .. }) => exprs
+            .last()
+            .is_some_and(|e| may_produce_null_or_undefined(expr_ctx, e)),
+        Expr::Paren(ParenExpr { expr, .. }) => may_produce_null_or_undefined(expr_ctx, expr),
+        _ => false,
```
kdy1, line 53: I think this is wrong (or maybe a bad name?)

kdy1: `await null` may produce null

veeceey: You're right — the name could be clearer. The intent is to check if an expression is guaranteed to produce a non-null/non-undefined value (i.e., safe to fold in array.join). `await null` does indeed resolve to `null`, so that should not be treated as "known non-null".

I'll rename this to something more precise (like `is_definitely_non_nullish`) and make sure `await` expressions are excluded. Will push a fix.

veeceey:
Pushed a fix to handle `await` expressions in the null/undefined check — `Expr::Await` now recurses into its inner expression, so `await null` is correctly identified as potentially producing null. Thanks for the catch @kdy1!

kdy1:
@codex Review

crates/swc_ecma_minifier/src/compress/pure/misc.rs
```diff
@@ -33,6 +33,27 @@ fn is_definitely_string(expr: &Expr) -> bool {
     }
 }
 
+/// Check whether an expression may obviously produce `null` or `undefined`.
+/// This catches direct references (void 0, undefined, null) and conditional
+/// expressions where either branch may produce null/undefined.
+fn may_produce_null_or_undefined(expr_ctx: ExprCtx, expr: &Expr) -> bool {
```
kdy1, line 39: There are lots of edge cases left, so I think we should not define a function like this.
We should rather do allowlist

kdy1:
Closing as the issue is fixed by https://github.com/swc-project/swc/pull/11558

PR #11553, opened 2026-02-18
fix(regexp): transpile unicode property escapes in RegExp constructor strings

### LINKED ISSUE #11539: Transpile regexp unicode property escapes when using the `RegExp` constructor
Describe the feature

Transpile regexp unicode property escapes 

```js
const IDENTIFIER_RE = new RegExp('\\p{ID_Start}', 'u')
console.log(IDENTIFIER_RE)
```

kdy1:
## Summary
- transform unicode property escapes for static string patterns passed to RegExp(...) and new RegExp(...)
- keep dynamic/non-literal constructor arguments untouched
- add regression coverage for constructor-call and constructor-new cases

Fixes #11539.

## Testing
- cargo test -p swc --test simple unicode_property_regex_constructor -- --nocapture
- cargo test -p swc --test simple unicode_property_regex -- --nocapture

### DIFF
--- a/crates/swc/tests/simple.rs
+++ b/crates/swc/tests/simple.rs
@@ -232,3 +232,51 @@ fn test_unicode_property_regex_letter() {
     // Should NOT contain the original \p{L} escape
     assert!(!compiled.contains("\\p{L}"));
 }
+
+#[test]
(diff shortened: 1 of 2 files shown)


kdy1:
It worked, but I think it can be better

PR #11583, opened 2026-02-24
fix(es/minifier): Handle `toExponential(undefined)`

### LINKED ISSUE #11572: Should treat `undefined` the same as empty args when calling `toExponential` and `toPrecision`
### Describe the bug

The output should be the same.

### Input code

```typescript
console.log(1.23.toExponential());
console.log(1.23.toPrecision());
console.log(1.23.toExponential(undefined));
console.log(1.23.toPrecision(undefined));
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2024",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": false,
        "booleans": false,
        "collapse_vars": false,
        "comparisons": false,
        "computed_props": false,
        "conditionals": false,
        "dead_code": false,
        "directives": false,
        "drop_console": false,
        "drop_debugger": false,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": false,
        "hoist_vars": false,
        "if_return": false,
        "join_vars": false,
        "keep_classnames": false,
        "keep_fargs": false,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": false,
        "negate_iife": false,
        "properties": false,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": false,
        "switches": false,
        "typeofs": false,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": false,
        "const_to_let": false,
        "pristine_globals": false,
        "passes": 5
      },
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

https://play.swc.rs/?version=1.15.11&code=H4sIAAAAAAAAA0vOzyvOz0nVy8lP1zDUMzLWK8l3rSjIz0vNK8lMzNHQ1LTmSsZUElCUmpxZnJmfh0sBshmleSmpaZl5qSkEDUNRCQCMDqdKmgAAAA%3D%3D&config=H4sIAAAAAAAAA32VTY7bMAyF9zlF4HUXxaDtogformcQFIlylFqiIVKZGIPcvfRfZqalsov58ZnRI0W%2FHY7H7kKu%2B3l8k5%2FyMNpCUB7PEqEps71JpAOXLLkSR%2B6%2B7PRCMwp2IFhC95V0bEsPvKjo5evLt03RDYgEu2KLpZhjmD7WdJjGAkQfYhKVV9YEmemzfmMFXzVwQhzA5mfIWDIxM%2FRQtDSHw2BHAnO1DZ7EtUioFplpZfBmLDjqCdlHjpglrmAP1huHHjQWCziOV1CFUk%2BEmeSQLezhVPt%2B6fe%2FHK52qJZnKZf6idyW3sg%2FVmRnjMQmVNWLFbaMWGnD5BhMAa5Fq3nBmFu6PwDiwmCJsk2qT0tGkNlqw6fSmIPML08Kl2FXT5qhF2dNjEHrzOwPFI5qyQK%2BOpgNdk9wwwyKHgyEIEOj4tfI7qyW5WkEDBqRTlv1GCswzYu58fl%2BPMG%2F5KCsj9qWkSyf25SmdEL1Yu1y4DP6JwnSDsY2LrI2bmOb1%2BxBBgS8mlJJBfO1ZcNohmWH%2Fj8hclPknaYf8KSvjVEmfunk9y12f6zsZHP%2FvhLWrX3YErqEvi5w%2Bx7MjV%2B3%2BI%2FuPWlf2I%2BqXaTfu3JeF4f7Xzl5NbxbBgAA

### SWC Info output

_No response_

### Expected behavior

```javascript
console.log("1.23e+0"), console.log("1.23"), console.log("1.23e+0"), console.log("1.23");
```

### Actual behavior

```javascript
console.log("1.23e+0"), console.log("1.23"), console.log("1e+0"), console.log(1.23.toPrecision(void 0));
```

### Version

1.15.11

### Additional context

_No response_

cuyl:
- Evaluate when parameters is empty or `undefined`.

Closes #11572


crates/swc_ecma_minifier/src/compress/pure/evaluate.rs
```diff
@@ -523,14 +523,21 @@ impl Pure<'_> {
...
+                    true
+                } else {
+                    arg.expr.may_have_side_effects(self.expr_ctx)
+                }
+            })
         {
             return;
         }
 
+        let args = args.iter().map(|i| Some(i.clone())).collect::<Vec<_>>();
```
kdy1: This clone is needless

cuyl: Sure, I've removed the neddless clone

cuyl:
Opps, let me see if I can find a way to remove clone.

crates/swc_ecma_minifier/src/compress/pure/evaluate.rs
```diff
@@ -524,13 +524,20 @@ impl Pure<'_> {
             _ => return,
         };
 
-        if args
-            .iter()
-            .any(|arg| arg.expr.may_have_side_effects(self.expr_ctx))
+        if args.len() > 1
+            || args.first().is_some_and(|arg| {
+                if arg.spread.is_some() {
+                    true
```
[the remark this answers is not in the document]
cuyl: We do not need these as it already handled in https://github.com/swc-project/swc/blob/0f799d82e2ee470533ed97153853a20840bf71ee/crates/swc_ecma_minifier/src/compress/pure/evaluate.rs#L71C19-L71C36

cuyl:
> > **P1 Badge Do not reject all spread arguments in number call folding**
> > This guard returns immediately whenever the first argument is spread, which makes `get_first_arg` unreachable for cases like `(1.23).toExponential(...[])` and `(1.23).toExponential(...[undefined])`. Those calls are equivalent to passing no precision and are exactly the new cases added in `tests/terser/compress/evaluate/number_method_call/input.js`, so the current condition prevents the intended optimization from ever running.
> > Useful? React with 👍 / 👎.
> 
 We already have `eval_array_spread`, `eval_spread_array_in_args` and `eval_spread_array_in_array` runs before `eval_number_method_call` I belive most of cases like this are handled.

cuyl:
Does anything block this PR?
@kdy1

crates/swc_ecma_minifier/src/compress/pure/evaluate.rs
```diff
@@ -1171,3 +1174,52 @@ fn f64_to_exponential_with_precision(n: f64, prec: usize) -> String {
...
 }
+
+/// Helper function to get the first argument of a call expression,
+///  taking into account spread elements.
+///
+/// `Some(None)` means that the first argument is a spread element, but we can't
+/// evaluate it.
+///
+/// `None` means that there is no argument.
+fn get_first_arg<'a>(
```
kdy1: This function is very wrong

kdy1 (changes requested):
AI coding is fine, but well... the diff in this PR does not look fine

crates/swc_ecma_minifier/src/compress/pure/evaluate.rs
cuyl: Opps, Should I revert it? I just want to remove the clone as you requested.
At first time I tried to add lifetime to prevent the clone, but I got an error.
```
error: reached the recursion limit while instantiating `get_first_arg::<'_, std::iter::Map<std::slice::Iter<'_, std::option::Option<swc_ecma_ast::ExprOrSpread>>, {closure@...}>>`
```
The code is here.
```rs
fn get_first_arg<'a, I>(args: I, expr_ctx: &ExprCtx) -> Option<Option<f64>>

where

    I: IntoIterator<Item = Option<&'a ExprOrSpread>>,

{

    args.into_iter()

        .find_map(|arg| {

            if let Some(arg) = arg {

                if arg.spread.is_some() {

                    return match arg.expr.as_array() {

                        Some(args) => {

                            if args.elems.is_empty() {

                                // next argument is used if the first spread argument is empty

                                // so we can't evaluate the call.

                                None

                            } else {

                                Some(get_first_arg(

                                    args.elems.iter().map(|e| e.as_ref()),

                                    expr_ctx,

                                ))

                            }

                        }

                        // If the spread argument is not an array, we can't evaluate the call.

                        None => Some(None),

                    };

                } else if !arg.expr.is_undefined(*expr_ctx) {

                    return Some(Some(eval_as_number(*expr_ctx, &arg.expr)));

                }

            }

            Some(None)

        })

        .unwrap_or(None)

}
```

To resolve the error, I've using AI to refactor it to the non-recursion version.

cuyl: But AI can't make it right, so I changed some logic to make it right.

cuyl:
> > > **P1 Badge Do not reject all spread arguments in number call folding**
> > > This guard returns immediately whenever the first argument is spread, which makes `get_first_arg` unreachable for cases like `(1.23).toExponential(...[])` and `(1.23).toExponential(...[undefined])`. Those calls are equivalent to passing no precision and are exactly the new cases added in `tests/terser/compress/evaluate/number_method_call/input.js`, so the current condition prevents the intended optimization from ever running.
> > > Useful? React with 👍 / 👎.
> 
> We already have `eval_array_spread`, `eval_spread_array_in_args` and `eval_spread_array_in_array` runs before `eval_number_method_call` I belive most of cases like this are handled.

As I mentioned before,  I realized `get_first_arg` is unnecessary, so I removed it.
@kdy1

cuyl:
I simplified the diff to make it easier to review. Any suggestions?
@kdy1

kdy1:
I instructed Codex to add some tests to this PR, and I'll review this on Monday.

kdy1 (changes requested):
CI failed

cuyl:
Fixed @kdy1

kdy1 (approved):

kdy1:
Thank you for your hard work!

PR #11604, opened 2026-03-01
reface(es/minifier): use arguments data from scope

Austaras:
**Description:**

**BREAKING CHANGE:**


crates/swc_ecma_minifier/src/compress/optimize/sequences.rs
```diff
@@ -2395,10 +2395,6 @@ impl Optimizer<'_> {
                         return Ok(false);
                     }
 
-                    if contains_arguments(&a.function) {
-                        return Ok(false);
-                    }
-
```
Austaras, line 2401: This never takes effect because we should check `a.function.body` instead of `a.function`

kdy1 (dismissed):
Nice work, thanks!

PR #11628, opened 2026-03-07
feat: add swc_es_codegen for swc_es_ast

kdy1:
## Summary
- add a new `swc_es_codegen` crate for `swc_es_ast`
- implement a standalone emitter without runtime dependency on any `swc_ecma_*` crate
- add fixture-based tests for JS, module syntax, TypeScript, and JSX
- expose a small API surface (`Config`, `Emitter`, `emit_program`, `CodegenError`)

## Performance Notes
- use a single preallocated `String` output buffer
- avoid `format!` in hot paths and write directly with `push` / `push_str`
- use `ryu-js` for fast number serialization
- keep string/template escaping as single-pass scans

## Verification
- `UPDATE=1 cargo test -p swc_es_codegen`
- `cargo fmt --all`
- `cargo test -p swc_es_codegen`


kdy1:
I created several PRs to compare various approaches

PR #11687, opened 2026-03-16
feat(react-compiler): scaffold SWC port of Babel entrypoint

kdy1:
## Summary
- add swc_ecma_react_compiler public API surface for options, errors, and entrypoint
- port Program-level compile entrypoint with function selection, directives, suppression handling, runtime and gating imports, and pass adapter
- add pipeline stage modules (hir, ssa, inference, optimization, reactive_scopes, validation, transform, utils) and fixture harness plus upstream sync script
- keep fast_check::is_required behavior compatible

## Tests
- cargo test -p swc_ecma_react_compiler
- cargo clippy -p swc_ecma_react_compiler --all-targets -- -D warnings

## Notes
- workspace-wide cargo fmt --all and cargo clippy --all --all-targets -- -D warnings currently fail due an existing unrelated parse error in crates/swc_es_parser/tests/bench_parser_inputs.rs (unclosed delimiter).

### DIFF
--- a/.changeset/green-owls-jump.md
+++ b/.changeset/green-owls-jump.md
@@ -1,0 +1,5 @@
+---
+swc_ecma_react_compiler: major
+---
+
+feat(react-compiler): scaffold SWC port of Babel entrypoint
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6516,8 +6516,10 @@ dependencies = [
 name = "swc_ecma_react_compiler"
 version = "14.0.0"
 dependencies = [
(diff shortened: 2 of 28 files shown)


kdy1:
Addressed review https://github.com/swc-project/swc/pull/11687#pullrequestreview-3951758434 in 710d12cd3c:
- ignore type-only imports in memo-cache import detection and import reuse
- apply eslint-disable-next-line suppression to the following function (and consume it once)
- added regression tests for both cases

PR #11708, opened 2026-03-19
fix(core): honor cwd in .swcrc lookup for relative filenames

kdy1:
## Summary

This fixes `.swcrc` auto-discovery for programmatic transforms that pass a **relative** `filename` plus `cwd` (the pattern observed in loader/integration paths where explicit `configFile` was a workaround).

### What changed

- `Compiler::read_config` now resolves default lookup root as:
  - `opts.root` if provided
  - otherwise `opts.cwd`
- `.swcrc` lookup no longer gives up when `canonicalize()` fails; it falls back to the original path and continues upward search.
- `default_cwd()` no longer uses a process-global cached value, so deserialization reflects the current directory at call time.
- Added regression coverage in `packages/core` for:
  - `transformSync(source, { filename: relative, cwd })` loading `.swcrc`
  - output equivalence with explicit `configFile`

## Files

- `crates/swc/src/lib.rs`
- `crates/swc/src/config/mod.rs`
- `packages/core/__tests__/transform/issue_11680_test.mjs`
- `packages/core/tests/issue-11680/.swcrc`
- `packages/core/tests/issue-11680/src/index.tsx`

## Verification

- `cd packages/core && yarn test __tests__/transform/issue_11680_test.mjs`
- `cd packages/core && yarn test`
- `cargo fmt --all`
- `cargo test -p swc`
- `cd packages/core && yarn build:dev && yarn test`

### DIFF
--- a/crates/swc/src/config/mod.rs
+++ b/crates/swc/src/config/mod.rs
@@ -1025,9 +1025,7 @@ pub struct CallerOptions {
 
 #[cfg(not(all(target_arch = "wasm32", not(target_os = "wasi"))))]
 fn default_cwd() -> PathBuf {
-    static CWD: Lazy<PathBuf> = Lazy::new(|| ::std::env::current_dir().unwrap());
-
-    CWD.clone()
+    ::std::env::current_dir().unwrap()
 }
 
 /// `.swcrc` file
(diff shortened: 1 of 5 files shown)


themavik (commented):
Reviewed the changes — the implementation follows the existing patterns and looks correct.

kdy1:
Th

PR #11728, opened 2026-03-23
fix(react): escape newlines in JSX quoted attribute values

### LINKED ISSUE #11550: JSX parser collapses newline to space in quoted attribute values
### Describe the bug

JSX output differs from Babel/Typescript

Whereas with the given input code, SWC outputs: `bruh bruh`

Babel/TS output: `bruh\nbruh`

This can lead to hydration errors in Next.js apps, when the client receives React Compiler output.

### Input code

```typescript
"use client";

export default function Hello({ }) {
  return (
    <div
      data-anything="bruh
bruh"
    >
      hello
    </div>
  );
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": true
    },
    "target": "es2024",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

https://play.swc.rs/?version=1.15.11&code=H4sIAAAAAAAAAy2MTQ6CMBCF93OKl65gYTwAwtprVDrYJs3U1BmjIdxdim5evrw%2FZ0%2FGnBOLuoGI349SFYEXb1mxmMyaiuDKOZduxdZjJaCyWhV0OwKXkF4HAMGrP3n5aExyH92tWqQm7sinfyu2s9%2F0vG%2Bb3Q%2B00Rfc63fmjAAAAA%3D%3D&config=H4sIAAAAAAAAA1WPSw7CMAxE9z1F5TULVCEW3IFDRMGtgtoksl2JqOrdm2%2BBXfxmJmNvXd%2FDmzU8%2Bi0%2B4%2BAVMdI5R8LBivpEAhI8sibjBS5NFU6S0IqZ7EUAUTShpBDycB1uNQCzc4wRj2pmrGwx1ozht1K7xRMy%2FxuTVdlpPvOlsautsLjXmsV6Slq3bHCHr6mVnR%2BD4WdL5jv2A3LKYuEWAQAA

### SWC Info output

_No response_

### Expected behavior

Should align with Babel/TS, and output `bruh\nbruh`. 

Perhaps debatable, since we are starting from invalid JavaScript syntax, but it is JSX code, that a spec should've defined, but AFAIK, no such thing exists, instead Babel/TS seem to be setting the expected output.

### Actual behavior

_No response_

### Version

1.15.11

### Additional context

Reported because for those using this JSX syntax, this can lead to hydration errors with the React Compiler, but arguably JSX is being too forgiving here, both SWC, Babel and TS should perhaps claim that this is not valid syntax. Or I am just missing some spec, explanation of this magic.

teee32:
## Problem

Literal newlines inside quoted JSX attribute string values are incorrectly
collapsed to a single space in the JSX transform. For example:

\`\`\`jsx
// Input
<div data-anything="line1
line2">hello</div>

// Current (wrong) output
React.createElement("div", { "data-anything": "line1 line2" }, "hello");

// Expected output
React.createElement("div", { "data-anything": "line1\nline2" }, "hello");
\`\`\`

This causes real-world issues such as hydration mismatches in Next.js
applications (see [#11550](https://github.com/swc-project/swc/issues/11550))
where data attributes containing newlines are corrupted.

## Root Cause

In `transform_jsx_attr_str()` at
`crates/swc_ecma_transforms_react/src/jsx/mod.rs`, the `'\n' | '\r' | '\t'`
match arm collapses all three characters to a single space, consuming any
trailing spaces:

\`\`\`rust
'\n' | '\r' | '\t' => {
    buf.push_char(' ');
    while let Some(next) = iter.peek() {
        if next.to_char() == Some(' ') {
            iter.next();
        } else {
            break;
        }
    }
}
\`\`\`

Literal `\n` and `\r` in JavaScript strings must be represented as escape
sequences (`\n`, `\r`), not collapsed to whitespace. Previous PR attempts
([#11556](https://github.com/swc-project/swc/pull/11556),
[#11569](https://github.com/swc-project/swc/pull/11569)) tried pushing
literal characters but the outputs were interpreted as invalid JS (literal
newlines in string literals are a syntax error). The codegen at
`crates/swc_ecma_codegen/src/lit.rs` already handles escaping LINE_FEED
and CARRIAGE_RETURN — the transform should push **actual** newline characters
and let codegen produce the correct escape sequences.

## Change

Split the combined `'\n' | '\r' | '\t'` match arm into three separate arms:

\`\`\`rust
'\n' => buf.push_char('\n'),  // Push actual char; codegen escapes to \\\\n
'\r' => buf.push_char('\r'),  // Push actual char; codegen escapes to \\\\r
'\t' => {
    buf.push_char(' ');  // kept as-is
    while let Some(next) = iter.peek() {
        if next.to_char() == Some(' ') {
            iter.next();
        } else {
            break;
        }
    }
}
\`\`\`

- `'\n'`: now pushes the actual newline codepoint; the existing codegen at
  `lit.rs:489` handles escaping it to `\\n` (single-backslash escape)
- `'\r'`: same — codegen at `lit.rs:490` escapes to `\\r`
- `'\t'`: behaviour unchanged — tabs are horizontal whitespace and are
  intentionally collapsed to a single space with trailing-space consumption

## Why This Fix

1. **Correct escaping**: pushing actual newline chars means the codegen's
   existing escape logic (designed precisely for this purpose) handles the
   output correctly
2. **Produces valid JS**: the codegen emits `\n` escape sequences, which
   are syntactically valid in JS string literals
3. **Minimal**: only 2 lines change in one function
4. **Backward compatible for the important case**: tab collapsing is preserved

## Risk

- **Low**: the change is confined to `transform_jsx_attr_str()` which is only
  called during JSX attribute value transformation
- Existing tab-collapsing behaviour is preserved
- All affected test fixtures (3 existing + 1 new regression test) have been
  updated and verified

## Validation

- `cargo test -p swc_ecma_transforms_react jsx` — 227 tests pass
- `cargo test -p swc --test projects` — 869 tests pass
- `cargo clippy -p swc_ecma_transforms_react` — no warnings
- `cargo fmt --all` — formatted

## Test Coverage

- **New fixture**: `tests/jsx/fixture/issue-11550/` — regression test for the
  reported case
- **Updated fixtures**: `codegen/jsx-1`, `issues-1233/case-1`,
  `issues-2162/case4`, `vercel/1`

Closes [#11550](https://github.com/swc-project/swc/issues/11550)

### DIFF
--- a/crates/swc_ecma_transforms_react/src/jsx/mod.rs
+++ b/crates/swc_ecma_transforms_react/src/jsx/mod.rs
@@ -2192,9 +2192,10 @@ fn transform_jsx_attr_str(v: &Wtf8) -> Wtf8Buf {
                 '\u{000c}' => buf.push_str("\\f"),
                 ' ' => buf.push_char(' '),
 
-                '\n' | '\r' | '\t' => {
+                '\n' => buf.push_char('\n'),
+                '\r' => buf.push_char('\r'),
+                '\t' => {
                     buf.push_char(' ');
-
                     while let Some(next) = iter.peek() {
                         if next.to_char() == Some(' ') {
                             iter.next();
--- a/crates/swc/tests/fixture/codegen/jsx-1/output/index.js
+++ b/crates/swc/tests/fixture/codegen/jsx-1/output/index.js
@@ -1,5 +1,5 @@
 export default /*#__PURE__*/ React.createElement(A, {
     className: b,
     header: "C",
-    subheader: "D E"
+    subheader: "D\n                E"
 });
--- a/crates/swc/tests/fixture/issues-1xxx/1233/case-1/output/index.js
+++ b/crates/swc/tests/fixture/issues-1xxx/1233/case-1/output/index.js
@@ -1,5 +1,5 @@
 function Component() {
     return /*#__PURE__*/ React.createElement("div", {
-        name: "A B"
+        name: "A\n      B"
     });
 }
--- a/crates/swc/tests/fixture/issues-2xxx/2162/case4/output/index.js
+++ b/crates/swc/tests/fixture/issues-2xxx/2162/case4/output/index.js
@@ -1,5 +1,5 @@
 function test() {
     return /*#__PURE__*/ React.createElement(React.Fragment, null, /*#__PURE__*/ React.createElement(A, {
-        b: "\\ "
+        b: "\\\n            "
     }));
 }
--- a/crates/swc_ecma_parser/src/lexer/mod.rs
+++ b/crates/swc_ecma_parser/src/lexer/mod.rs
@@ -1502,7 +1502,10 @@ impl<'a> Lexer<'a> {
                     }
                 }
 
-                chunk_start = cur_pos + BytePos(ch.len_utf8() as _);
+                // `read_jsx_new_line` can consume an entire CRLF sequence, so
+                // restart from the actual current position instead of advancing
+                // by only the first code unit.
+                chunk_start = self.input().cur_pos();
             } else {
                 self.bump(ch.len_utf8());
             }
--- a/crates/swc_ecma_parser/src/parser/jsx.rs
+++ b/crates/swc_ecma_parser/src/parser/jsx.rs
@@ -582,6 +582,37 @@ mod tests {
     }
 
     #[test]
+    fn crlf_in_attr() {
+        assert_eq_ignore_span!(
+            jsx(concat!(
+                "<div data-anything=\"line1",
+                "\r\n",
+                "    line2\" />"
+            )),
+            Box::new(Expr::JSXElement(Box::new(JSXElement {
+                span,
+                opening: JSXOpeningElement {
+                    span,
+                    attrs: vec![JSXAttrOrSpread::JSXAttr(JSXAttr {
+                        span,
+                        name: JSXAttrName::Ident(IdentName::new(atom!("data-anything"), span)),
+                        value: Some(JSXAttrValue::Str(Str {
+                            span,
+                            value: atom!("line1\r\n    line2").into(),
+                            raw: Some(Atom::new(concat!("\"line1", "\r\n", "    line2\""))),
+                        })),
+                    })],
+                    name: JSXElementName::Ident(Ident::new_no_ctxt(atom!("div"), span)),
+                    self_closing: true,
+                    type_args: None,
+                },
+                children: Vec::new(),
+                closing: None
+            })))
+        );
+    }
+
+    #[test]
     fn issue_584() {
         assert_eq_ignore_span!(
             jsx(r#"<test other={4} />;"#),
--- a/crates/swc_ecma_transforms_react/src/jsx/tests.rs
+++ b/crates/swc_ecma_transforms_react/src/jsx/tests.rs
@@ -11,7 +11,9 @@ use swc_ecma_transforms_base::{fixer::fixer, hygiene, resolver};
 use swc_ecma_transforms_compat::es2015::{arrow, classes};
 #[cfg(feature = "es3")]
 use swc_ecma_transforms_compat::es3::property_literals;
-use swc_ecma_transforms_testing::{parse_options, test, test_fixture, FixtureTestConfig, Tester};
+use swc_ecma_transforms_testing::{
+    parse_options, test, test_fixture, test_inline, FixtureTestConfig, Tester,
+};
 use testing::NormalizedOutput;
 
 use super::*;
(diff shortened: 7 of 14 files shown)


crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -2192,9 +2192,10 @@ fn transform_jsx_attr_str(v: &Wtf8) -> Wtf8Buf {
                 '\u{000c}' => buf.push_str("\\f"),
                 ' ' => buf.push_char(' '),
 
-                '\n' | '\r' | '\t' => {
+                '\n' => buf.push_char('\n'),
+                '\r' => buf.push_char('\r'),
+                '\t' => {
                     buf.push_char(' ');
-
                     while let Some(next) = iter.peek() {
```
kdy1, line 2199: Multiline JSX attribute values still normalize `\t` to a space in this branch.

Because this arm does `buf.push_char(' ')` (and then consumes following spaces), an input like `<A x="a\n\tb" />` is lowered to a runtime value of `"a\n b"` instead of preserving the tab (`"a\n\tb"`, matching TypeScript/Babel behavior).

So the newline fix is correct, but tab-indented multiline attrs are still semantically changed. Could we preserve `\t` here (for example `buf.push_char('\t')`) and add a fixture for that case?

crates/swc_ecma_transforms_react/src/jsx/mod.rs
```diff
@@ -2192,9 +2192,10 @@ fn transform_jsx_attr_str(v: &Wtf8) -> Wtf8Buf {
                 '\u{000c}' => buf.push_str("\\f"),
                 ' ' => buf.push_char(' '),
 
-                '\n' | '\r' | '\t' => {
+                '\n' => buf.push_char('\n'),
```
kdy1, line 2195: Nice fix for LF, but this exposes a CRLF regression. In `read_jsx_str` (`crates/swc_ecma_parser/src/lexer/mod.rs`), when `read_jsx_new_line(false)` consumes `\r\n`, `chunk_start` is still set via `cur_pos + BytePos(ch.len_utf8() as _)` (line 1505), so the `\n` gets re-sliced and appended again. I reproduced this by changing the fixture input newline to CRLF: the emitted value becomes `"bruh\\r\\n\\nbruh"` (extra LF). Could we advance `chunk_start` to the current lexer position after consuming the newline sequence and add a CRLF fixture?

kdy1, line 2195: This change makes multiline JSX attribute strings diverge from Babel semantics. Babel applies `value.value.replace(/\n\s+/g, " ")` in `@babel/plugin-transform-react-jsx`, so inputs like `<div a="x\n    y" />` compile to `"x y"` (newline + indentation collapsed), while `<div a="x\ny" />` keeps the newline. With unconditional `'\n' => buf.push_char('\n')` here, SWC now emits `"x\n    y"`, changing runtime values for existing indented multiline attributes and regressing fixtures such as `codegen/jsx-1`, `issues-1233`, and `issues-2162`. Could we preserve `\n` only when it is not followed by horizontal whitespace, and keep collapsing `\n[ \t]+` to a single space?

afurm:
The separate `'\r'` arm mirrors the `'\n'` approach — pushing the actual character and relying on codegen to escape it. But CRLF sequences (`\r\n`) are a single logical newline in some parsers. When the input is `<div data-x="a\r\nb" />`, does the lexer correctly produce two separate `'\r'` and `'\n'` events, or could they be consumed as one unit by `read_jsx_new_line`?

kdy1:
@claude Verify https://github.com/swc-project/swc/pull/11728#issuecomment-4237855794

kdy1:
@afurm Good catch. For JSX attribute strings, `\r\n` is not surfaced as two independent newline events: `read_jsx_new_line(false)` consumes the CRLF pair as a single lexer step.

The important part is that it still preserves the actual sequence as `"\r\n"` in the parsed string, so by the time we reach `transform_jsx_attr_str`, the `Str.value` still contains both `'\r'` and `'\n'`. The separate match arms therefore preserve the CRLF sequence, and codegen escapes it back out as `\\r\\n`.

There is one parser-side detail that matters here: because `read_jsx_new_line(false)` can consume both code units at once, the lexer needs to restart from `self.input().cur_pos()` afterward instead of advancing by only the first code unit. This branch already includes that fix, plus regression coverage for parser CRLF handling and for the React transform's CR / CRLF cases.

kdy1:
Closing as fixed by https://github.com/swc-project/swc/pull/11796

PR #11777, opened 2026-04-08
feat(swc_common): add SourceMapper.map_raw_pos

mischnic:
**Description:**

#10560 added `Files.map_raw_pos` which can be used to translate BytePos before generating the sourcemap.

The same is also needed in `SourceMapper`, so that the `pos.is_pure()` in `emit_leading_comments` uses the decoded BytePos.

This came up in https://github.com/vercel/next.js/pull/88205 where my BytePos re-encoding mechanism caused `BytePos::PURE` to not codegen the pure comment anymore.

**BREAKING CHANGE:**

### DIFF
--- a/.changeset/stale-crews-peel.md
+++ b/.changeset/stale-crews-peel.md
@@ -1,0 +1,7 @@
+---
+swc_common: patch
+swc_ecma_codegen: patch
+swc_core: patch
+---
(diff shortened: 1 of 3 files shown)


kdy1 (approved):
Good catch!

PR #11778, opened 2026-04-09
docs: Document Flow strip support

kdy1:
## Summary

This PR documents the current Flow parse/strip path across the Rust and JS surfaces.

## What Changed

- added a Flow parsing example to `swc_ecma_parser` rustdoc
- documented Flow strip usage in `swc_ecma_transforms_typescript`
- added a `flow_to_js` example and Flow strip correctness coverage
- exposed `FlowParserConfig` in the public parser typings and wasm bindings

## Why

SWC already supports parsing Flow syntax and stripping Flow annotations through the existing TypeScript transform entry point with `flow_syntax: true`, but that path was difficult to discover and not fully represented in the public typings.

## Impact

- Rust users now have crate-level docs and an example showing how to parse Flow and strip it to JavaScript
- JS users can express Flow parser settings through `@swc/types`
- the new fixture-based coverage helps keep the parse -> strip -> reparse flow working

## Validation

- `cargo fmt --all`
- `cargo test -p swc_ecma_transforms_typescript`
- `(cd packages/types && yarn build)`
- `(cd packages/core && yarn build:ts)`

### DIFF
--- a/.changeset/nine-wombats-obey.md
+++ b/.changeset/nine-wombats-obey.md
@@ -1,0 +1,7 @@
+---
+swc_core: patch
+swc_ecma_parser: patch
+swc_ecma_transforms_typescript: patch
+---
+
+[codex] Document Flow strip support
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6831,6 +6831,7 @@ dependencies = [
  "codspeed-criterion-compat",
  "rustc-hash 2.1.1",
  "serde",
+ "serde_json",
  "swc_atoms",
  "swc_common",
  "swc_ecma_ast",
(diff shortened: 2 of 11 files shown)


gkz:
@kdy1 The Flow team appreciates this, thank you!
We're adopting TypeScript syntax where there is no reason for it to be kept different, so the parsing differences will decline over time (e.g. we've already replaced `%checks` with the `is` type guard syntax). There is just a lag between an old syntax being deprecated vs. support completely deleted. Of course unique features (`component` syntax, pattern matching) will always remain.

kdy1:
That sounds good, thank you for letting me know!

PR #11781, opened 2026-04-10
fix(es/decorators): preserve super in moved static members

### LINKED ISSUE #11780: [decorators 2023-11] static field with super reference incorrectly moved to wrapper class instance field
### Description

When a class has a **class decorator** and a **static field initializer that references `super`**, `@swc/core` with `decoratorVersion: "2023-11"` incorrectly moves the static field to an **instance field** in the `_identity` wrapper class. Because `super` in the wrapper class resolves to `_identity` (not the actual parent class), the initializer crashes at runtime.

### Versions

- `@swc/core`: `1.15.24`
- `jsc.transform.decoratorVersion`: `"2023-11"`
- `jsc.target`: `"es2022"`

### Reproduction

```js

const code = `
  function customElement(tag) {
    return (cls, ctx) => { ctx.addInitializer(() => {}); };
  }

  class Base {
    static styles = ["base"];
  }

  @customElement("derived")
  class Derived extends Base {
    static styles = [...super.styles, "derived"];
  }

  console.log(Derived.styles);
`;

const result = swc.transformSync(code, {
  jsc: {
    target: "es2022",
    parser: { syntax: "typescript", decorators: true },
    transform: { decoratorVersion: "2023-11" },
  },
});

eval(result.code);
// TypeError: (intermediate value).styles is not iterable
```

### Generated output (simplified)

The transform generates this pattern:

```js
let _Derived;
new class extends _identity {
  static [class Derived extends Base {
    static {
      ({ c: [_Derived, _initClass] } = _apply_decs_2311(this, [_dec], [], 0, void 0, Base));
    }
  }];
  styles = [...super.styles, "derived"]; // ❌ instance field, super = _identity
  constructor() {
    super(_Derived), _initClass();
  }
}();
```

The static field `static styles = [...super.styles, "derived"]` is moved to the wrapper class as an **instance field** `styles = [...]`. In this new context, `super` refers to `_identity` (a helper function), not `Base`. Since `_identity.styles` is `undefined`, the spread throws:

```
TypeError: (intermediate value).styles is not iterable
```

### Expected behaviour

`static styles = [...super.styles, "derived"]` should remain a **static field on `Derived`**, with `super.styles` correctly resolving to `Base.styles`. Expected output:

```
["base", "derived"]
```

### Workaround

Replacing `super.styles` with an explicit reference to the parent class avoids the issue:

```js
@customElement("derived")
class Derived extends Base {
  static styles = [...Base.styles, "derived"]; // explicit reference works
}
```

Note: `@babel/plugin-proposal-decorators@7.29.0` generates the **identical** wrapper class pattern and exhibits the same bug. Issue filed at babel/babel#17929.

kdy1:
## Summary
- preserve super semantics when 2023-11 class decorators move static members onto the _identity wrapper
- extend the moved-static rewrite trigger to cover super usage in static field/accessor initializers, private static methods, and trailing static blocks
- add a regression fixture for issue #11780 covering the reported static field case and related moved-static super paths

## Testing
- git submodule update --init --recursive
- cargo fmt --all
- cargo test -p swc_ecma_transforms_proposal
- cargo clippy --all --all-targets -- -D warnings

Fixes #11780

### DIFF
--- a/crates/swc_ecma_transforms_proposal/src/decorator_impl.rs
+++ b/crates/swc_ecma_transforms_proposal/src/decorator_impl.rs
@@ -162,6 +162,25 @@ impl Visit for CurrentClassNameFinder {
     }
 }
 
+#[derive(Default)]
+struct SuperFinder {
+    found: bool,
+}
+
+impl Visit for SuperFinder {
+    noop_visit_type!();
+
+    fn visit_class(&mut self, _: &Class) {
+        // `super` inside a nested class belongs to that class and should not
+        // affect the decision to rewrite the enclosing moved static
+        // member.
+    }
+
+    fn visit_super(&mut self, _: &Super) {
+        self.found = true;
+    }
+}
+
 impl DecoratorPass {
     fn is_2023_11(&self) -> bool {
         matches!(self.version, DecoratorVersion::V202311)
@@ -527,6 +546,12 @@ impl DecoratorPass {
         v.found
     }
 
+    fn expr_uses_super(&self, expr: &Expr) -> bool {
+        let mut v = SuperFinder::default();
+        expr.visit_with(&mut v);
+        v.found
+    }
+
     fn stmts_use_instance_private_names(
         &self,
         stmts: &[Stmt],
@@ -540,6 +565,12 @@ impl DecoratorPass {
             names: instance_private_names,
             found: false,
         };
+        stmts.visit_with(&mut v);
+        v.found
+    }
+
+    fn stmts_use_super(&self, stmts: &[Stmt]) -> bool {
+        let mut v = SuperFinder::default();
         stmts.visit_with(&mut v);
         v.found
     }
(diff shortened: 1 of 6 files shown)


crates/swc_ecma_transforms_proposal/src/decorator_impl.rs
```diff
@@ -162,6 +162,25 @@ impl Visit for CurrentClassNameFinder {
...
+impl Visit for SuperFinder {
+    noop_visit_type!();
+
+    fn visit_class(&mut self, _: &Class) {
+        // `super` inside a nested class belongs to that class and should not
+        // affect the decision to rewrite the enclosing moved static
+        // member.
+    }
+
+    fn visit_super(&mut self, _: &Super) {
```
JLHwung, line 179: The super finder should skip `super` within decorators / computed keys, because the `super` there is resolved to the upper class environment. For example, `super.prop1` and `super.prop2` here will be resolved to `A`.

```js
class A extends O1 {
  @dec class B extends O2 {
    @(super.prop1)
    static [super.prop2] = 0;
  }
}
```

Issue #11785, opened 2026-04-13
Flow parser fails on callable type with unnamed function-type parameter followed by additional parameters


oblador:
### Describe the bug

SWC's Flow parser fails to parse valid Flow callable types where an unnamed parameter is a function type (arrow syntax) and is followed by additional parameters.

```flow
type Fn = (() => A, b: B) => C;
```

This is valid Flow syntax — the callable type has two parameters: an anonymous `() => A` and a named `b: B`. The parser appears to interpret `() => A` as closing the outer parenthesized group, then chokes on the `,`.

This matters in practice because React Native's source code uses this pattern, e.g. in [`createAnimatedPropsMemoHook.js`](https://github.com/facebook/react-native/blob/main/packages/react-native/src/private/animated/createAnimatedPropsMemoHook.js):

```flow
type AnimatedPropsMemoHook = (
  () => AnimatedProps,
  props: Readonly<{[string]: unknown}>,
) => AnimatedProps;
```

### Input code

```typescript
type Fn = (() => A, b: B) => C;
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "flow",
      "all": true
    }
  }
}
```

### Link to the code that reproduces this issue

Playground doesn't support flow

### Expected behavior

The code parses and transforms successfully, stripping the Flow type:

```js
// empty output (type-only statement)
```

### Actual behavior

```
[Error:   × Expected ';', '}' or <eof>
   ╭─[input.js:1:1]
 1 │ type Fn = (() => A, b: B) => C;
   · ──┬─ ──
   ·   ╰── This is the expression part of an expression statement
   ╰────

Caused by:
    Syntax Error] {
  code: 'GenericFailure'
}
```

### Version

1.15.24

### Additional context

The issue is specifically with **unnamed** function-type parameters in callable types. Related patterns that **do** parse correctly:

| Pattern | Parses? |
|---|---|
| `type Fn = (() => A, b: B) => C;` | No |
| `type Fn = (fn: () => A, b: B) => C;` | Yes |
| `type Fn = (() => A) => C;` | Yes |
| `type Fn = (() => A) \| B;` | Yes |

So the failure only occurs when:
1. A parameter is an unnamed function type using arrow syntax, **and**
2. It is followed by a `,` and additional parameters

kdy1:
I'll fix this asap.

Are you using @swc/core or from Rust side? Asking to know which one I should publish after fixing

oblador:
[...]

PR #11787, opened 2026-04-14
fix(es/parser): Fix Flow callable type with unnamed fn-type param

oblador:
**Description:**

When a Flow callable type has an unnamed function-type parameter followed by named parameters (e.g. `(() => A, b: B) => C`), the parser now correctly handles this by falling back to a speculative parse via the regular function type parser, whose binding-list handler supports mixed anonymous and named parameters.

Previously, `try_parse_flow_anon_fn_type` would bail on encountering a named parameter, and `is_ts_start_of_fn_type` would fail because it didn't recognize `(` as a valid function parameter start. The parser would then try to parse `(() => A, ...)` as a parenthesized type and error on the comma.

Also defensively clear `DisallowFlowAnonFnType` context when parsing parameter types in `try_parse_flow_anon_fn_type` and `try_parse_flow_anon_signature_param` so nested anonymous function types like `() => A` are always allowed as parameters.

**BREAKING CHANGE:** No

**Related issue (if exists):** #11785

### DIFF
--- a/.changeset/fix-flow-anon-fn-param.md
+++ b/.changeset/fix-flow-anon-fn-param.md
@@ -1,0 +1,6 @@
+---
+swc_ecma_parser: patch
+swc_core: patch
+---
+
+fix(es/parser): Fix Flow callable type with unnamed function-type parameter followed by additional parameters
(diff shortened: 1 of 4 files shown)


oblador:
Oh there is already a draft PR for this, I misinterpreted the message from eliottreich.

oblador:
He's even a bot...

PR #11801, opened 2026-04-17
fix(es/module): add opt-in symlink-preserving resolver

### LINKED ISSUE #11584: jsc.baseUrl causes canonicalization of symlinked source files, breaking bundler symlinks: false
### Describe the bug

When `jsc.baseUrl` is set in `builtin:swc-loader`, SWC unconditionally calls `canonicalize()` on file paths, resolving symlinks to their real filesystem paths. This defeats rspack's `resolve.symlinks: false` setting — relative imports from a symlinked source file resolve from the **real** path instead of the **symlink** location.

This affects any project that uses symlinked source files with `baseUrl` (e.g. monorepos sharing code via symlinks). Without `baseUrl`, the same setup works correctly because SWC does not create its `NodeImportResolver` and no canonicalization occurs.

Discovered via Meteor framework: https://github.com/meteor/meteor/issues/14053

### Input code

Minimal reproduction: **https://github.com/perbergland/swc-symlinks**

```
shared/
  source.ts          # imports ./dep (dep.ts does NOT exist here)
server/
  source.ts          # symlink -> ../shared/source.ts
  dep.ts             # exists here
src/
  index.ts           # imports ../server/source
```

`shared/source.ts`:
```typescript
import { VALUE } from "./dep";
export const myMethod = () => VALUE;
```

`server/dep.ts`:
```typescript
export const VALUE = "hello from server/dep";
```

With `resolve.symlinks: false`, the import `./dep` in `server/source.ts` (symlink) should resolve to `server/dep.ts`. Instead, SWC canonicalizes the path and looks for `shared/dep.ts`.

### Config

**Works** (no `baseUrl`):
```json
{
  "jsc": {
    "parser": { "syntax": "typescript" }
  }
}
```

**Fails** (with `baseUrl`):
```json
{
  "jsc": {
    "baseUrl": ".",
    "parser": { "syntax": "typescript" }
  }
}
```

https://github.com/perbergland/swc-symlinks

```bash
git clone https://github.com/perbergland/swc-symlinks
cd swc-symlinks
npm install
npm run setup       # creates the symlink
npm run build       # works (no baseUrl)
npm run build:baseurl  # FAILS (with baseUrl)
```

### Expected behavior

Both builds succeed. `resolve.symlinks: false` should preserve symlink paths regardless of whether `jsc.baseUrl` is set.

### Actual behavior

`npm run build:baseurl` fails:

```
ERROR in ./shared/source.ts 1:0-30
  × Module not found: Can't resolve './dep' in '.../swc-symlinks/shared'
```

SWC resolves the symlink `server/source.ts` to its real path `shared/source.ts`, then `./dep` is looked up in `shared/` where it doesn't exist.

### Version

Reproduced with `@rspack/core` 1.6.7 (uses SWC internally via `builtin:swc-loader`).

### Additional context

**Root cause** — in [`swc_ecma_transforms_module/src/path.rs`](https://github.com/swc-project/swc/blob/main/crates/swc_ecma_transforms_module/src/path.rs) (around line 270):

```rust
// Bazel uses symlink
// https://github.com/swc-project/swc/issues/8265
if let FileName::Real(resolved) = &target.filename {
    if let Ok(orig) = canonicalize(resolved) {
        target.filename = FileName::Real(orig);
    }
}
```

`canonicalize()` unconditionally resolves symlinks. This was added to fix Bazel sandbox issues (#8265), but it breaks the common case of symlinked source files with `resolve.symlinks: false`.

**Why it only happens with `baseUrl`**: Without `baseUrl`, SWC skips creating its `NodeImportResolver` entirely, so no `canonicalize()` is called and symlink paths pass through unchanged.

**Related issues:**
- #8265 — Bazel sandbox symlinks (same `canonicalize()` code, different scenario)
- #4057 — Symlinked node_modules resolved incorrectly
- PR #6716 — Earlier fix attempt, reverted because it broke Bazel

**Suggested fix**: Add a `preserveSymlinks` option to the resolver config. When enabled, use `std::fs::read_link()` instead of `canonicalize()` — or skip canonicalization entirely — so that symlink paths are preserved. This would allow bundlers that set `resolve.symlinks: false` to work correctly while keeping the Bazel workaround available as the default behavior.

kdy1:
**Description:**

This adds an opt-in `NodeImportResolver` constructor that preserves symlink paths when rewriting module specifiers.

Today, `NodeImportResolver` always canonicalizes resolved real paths before converting them back into relative import specifiers. That behavior is still useful for the Bazel-related fixes around `jsc.paths`, so this PR keeps it as the default. However, it breaks callers that intentionally preserve symlink locations and need rewritten imports to stay relative to the symlink path instead of the canonicalized real path.

This PR keeps the existing default constructor unchanged and adds a dedicated preserve-symlinks constructor for callers that need `resolve.symlinks: false`-style behavior. It also adds a regression test that creates a symlinked source file and asserts the opt-in boundary explicitly: the default resolver rewrites to the canonical target path, while the new constructor preserves the symlink path.

Validation:
- `cargo fmt --all`

**BREAKING CHANGE:**

None.

Closes #11584

### DIFF
--- a/.changeset/good-lemons-study.md
+++ b/.changeset/good-lemons-study.md
@@ -1,0 +1,5 @@
+---
+swc_ecma_transforms_module: major
+---
+
+fix(es/module): add opt-in symlink-preserving resolver
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6718,6 +6718,7 @@ dependencies = [
  "swc_ecma_transforms_typescript",
  "swc_ecma_utils",
  "swc_ecma_visit",
+ "tempfile",
  "testing",
  "tracing",
 ]
--- a/crates/swc_ecma_transforms_module/Cargo.toml
+++ b/crates/swc_ecma_transforms_module/Cargo.toml
@@ -46,6 +46,7 @@ swc_ecma_visit = { version = "23.0.0", path = "../swc_ecma_visit" }
 [dev-dependencies]
 indexmap   = { workspace = true, features = ["serde"] }
 serde_json = { workspace = true }
+tempfile   = { workspace = true }
 
 swc_ecma_loader = { version = "21.0.0", path = "../swc_ecma_loader", features = [
   "node",
--- a/crates/swc_ecma_transforms_module/src/path.rs
+++ b/crates/swc_ecma_transforms_module/src/path.rs
@@ -94,6 +94,7 @@ where
 {
     resolver: R,
     config: Config,
+    preserve_symlinks: bool,
 }
 
 #[derive(Debug, Clone)]
@@ -118,6 +119,16 @@ where
     R: Resolve,
 {
     pub fn with_config(resolver: R, config: Config) -> Self {
+        Self::with_config_inner(resolver, config, false)
+    }
+
+    /// Creates a resolver that preserves symlink paths when rewriting module
+    /// specifiers.
+    pub fn with_config_preserving_symlinks(resolver: R, config: Config) -> Self {
+        Self::with_config_inner(resolver, config, true)
+    }
+
+    fn with_config_inner(resolver: R, config: Config, preserve_symlinks: bool) -> Self {
         #[cfg(not(all(target_arch = "wasm32", not(target_os = "wasi"))))]
         if let Some(base_dir) = &config.base_dir {
             assert!(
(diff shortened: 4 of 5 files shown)


perbergland:
Thank you!

PR #11805, opened 2026-04-18
perf(parser): reduce ts speculative parse overhead

kdy1:
**Description:**

This PR continues the `swc_ecma_parser` parser-core re-architecture by cutting TypeScript speculative parse overhead on the hot path without changing the public parser API.

- decouple token payload storage from cursor checkpoints so speculative parse no longer shuffles token payload ownership during backtracking
- add cursor-based checkpoint helpers and use them to reduce parser state restore work in TypeScript speculation
- split generic async arrow parsing into head detection plus committed parse to avoid nested `try_parse_ts` round-trips
- add a token-level fast reject for instantiation expression type arguments in non-JSX TypeScript while preserving the existing TSX ambiguity behavior
- extend the fixture suite with generic async arrow and instantiation-expression edge cases, including member/computed access and optional chaining

Validation:

- `cargo fmt --all`
- `cargo test -p swc_ecma_parser --test js`
- `cargo test -p swc_ecma_parser --test jsx`
- `cargo test -p swc_ecma_parser --test typescript`
- `cargo test -p swc_ecma_parser --test errors`
- `cargo test -p swc_ecma_parser --test comments`
- `cargo test -p swc_ecma_parser --test span`
- `cargo clippy -p swc_ecma_parser --all-targets -- -D warnings`
- `cargo run --example perf --quiet`

Perf notes:

- `es/parser/cal-com` improved from roughly `9.22-11.34ms` on the baseline run to `8.74-8.97ms`
- `es/parser/typescript` stayed effectively neutral at roughly `68.5-82.7ms` baseline vs `68.9-82.2ms` current

**BREAKING CHANGE:**

None.

N/A

### DIFF
--- a/crates/swc_ecma_parser/src/lexer/capturing.rs
+++ b/crates/swc_ecma_parser/src/lexer/capturing.rs
@@ -145,16 +145,8 @@ impl<I: Tokens> Tokens for Capturing<I> {
         self.inner.token_flags()
     }
 
-    fn clone_token_value(&self) -> Option<super::TokenValue> {
-        self.inner.clone_token_value()
-    }
-
-    fn take_token_value(&mut self) -> Option<super::TokenValue> {
-        self.inner.take_token_value()
-    }
-
-    fn get_token_value(&self) -> Option<&super::TokenValue> {
-        self.inner.get_token_value()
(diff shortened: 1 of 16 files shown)


kdy1:
Closing as a bad optimization

PR #11810, opened 2026-04-19
fix(es/transforms): strip `@flow` and `@noflow` pragma comments

oblador:
**Description:** When debugging some other issues I noticed that the flow pragma comments were retained after flow stripping. While technically regular JS can be parsed as flow, it still seems incorrect and I noticed some edge cases where some JS comments caused flow parser to trip up trying to parse them as types. This PR changes this behaviour to strip flow pragma either in its entirety or if multiline only the pragma itself, making it align with the existing [babel transform](https://babeljs.io/repl#?config_lz=N4IgZglgNgpgdgQwLYxALhAJxgBygOgCsBnEAGhB22JgBdS0BtRkeAN3NFoUwHM6GIACYwwCAK5R6ZAARwA9rRkQYMgIxrZCpSoD6SeQCN1a8iANDJMBmKg0Kh8b0gAPa-lqZxMAL4BdMhZibn4AWgAmTmEYAGN5TARaeOIANRhMYgh5OHQQcIAGcIBmUI0zHAgcGCgIOBgABUx5HHliBChcgAsEGIBrEH8KMCh5AHcQAJBieXFMGJgAFQBPKtyLKwGgA&code_lz=PQKhAIChwgZB7ANtCLwAEBmj4HdwDOALgE4CWAxkTMJEA&lineWrap=true&version=7.29.2)

**BREAKING CHANGE:** No.

### DIFF
--- a/.changeset/flow-pragma-strip.md
+++ b/.changeset/flow-pragma-strip.md
@@ -1,0 +1,7 @@
+---
+swc_core: patch
+swc: patch
+swc_ecma_transforms_typescript: patch
+---
+
+Strip `@flow` and `@noflow` pragma comments when transforming Flow syntax
--- a/.changeset/flow-pragma-strip.md
+++ b/.changeset/flow-pragma-strip.md
@@ -4,4 +4,4 @@ swc: patch
 swc_ecma_transforms_typescript: patch
 ---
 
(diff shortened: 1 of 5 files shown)


kdy1 (changes requested):
Please remove the new visitor and use another approach. The binary size and the runtime performance will explode

Issue #11704, opened 2026-03-18
Number minification as hex can result in larger output at 0x1000000, 0x10000000, 0x100000000 breakpoints


bmiddha:
### Describe the bug

When SWC minifies decimal numbers into hexadecimal in `minify_number` (crates/swc_ecma_codegen/src/lit.rs), the resulting minified hex representation is occasionally longer than the original decimal string. This happens at specific breakpoints, such as 0x1000000, 0x10000000, and 0x100000000. For example, 16777216 (0x1000000) is minified as `0x1000000` (9 chars) instead of `16777216` (8 chars), so the output is longer than the original number. See playground link for demo.

### Input code

```typescript
16777215 // same length after minification
16777216 // bigger after minification

268435455 // same length after minification
268435456 // bigger after minification

4294967295 // same length after minification
4294967296 // bigger after minification

68719476735 // same length after minification
68719476736 // bigger after minification
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": true,
  "isModule": true
}
// See full playground config in repro link.
```

### Link to the code that reproduces this issue

https://play.swc.rs/?version=1.15.18&code=H4sIAAAAAAAAA43MQQqAIBBA0X2nmBuENs7kcSzMBtKgvD8VUW0CXf%2FHV8TMWhloW9hd9LD4FPIMbsp%2BgyhJJhldljU16qZ00UFCOPuPajT12Bk0NcvHlp6oLVpibWumLy5dqWdlkYm7mu2nC98DIGJbbVMBAAA%3D&config=H4sIAAAAAAAAA1WPSw7DIAwF95wCed1tu%2BgdcghEnYiIn2wiFUW5e4FA2%2BzwvDFP3oWUsLKGp9zLswxRESN950I4%2B6TehQBqp1iTiQluI125RrOyjA0dZwJJ0YKpbfG962BDYBx6Z854M%2Bf%2FQh1cJGS%2BilVVfrF4rRO9Elx4bS3sh6Qc8ax%2FwE8aZYm29i8YnsZiZeL4ADoI3bUTAQAA

### Expected behavior

Hexadecimal minification should only be used when the minified representation is not longer than the original decimal string, ideally only when it is actually shorter. For example:
[...]

PR #11816, opened 2026-04-23
fix(es/codegen): Refine hex number minification

magic-akari:
**Description:**

**BREAKING CHANGE:**

- #11704


magic-akari:
I was surprised to find that, regardless of whether we narrow or expand the hex range, or exclude longer sequences, the Compressed Size may decrease while the Gzipped Size still increases.

The threshold we previously established achieves a well-balanced trade-off. Therefore, no changes are needed, and this PR should be closed.

PR #11828, opened 2026-04-27
fix(ts/fast-strip): reject unsafe assertion erasure in binary expressions

magic-akari:
**Description:**

**BREAKING CHANGE:**

- bloomberg/ts-blank-space#62


kdy1 (approved):
Thanks!

PR #11839, opened 2026-05-02
fix(es/minifier): Avoid generating mangled property names that collide with existing properties

### LINKED ISSUE #11027: es/minifier: Properties of inherited class might be incorrectly renamed when props manging is enabled
### Describe the bug

If a class has a one-letter property name, then properties of an inherited class might get renamed to the same name, even if they are unrelated.
This only happens if `jsc.minify.mangle.props` is enabled.

Input:
```js
class Class1 {
    e = 1;
}

class Class2 extends Class1 {
    someField = 2;
}

console.log(new Class2().e); // Writes 1, as expected
```

Output:
```js
class e {
    e = 1;
}
class s extends e {
    e = 2;
}
console.log(new s().e);  // Writes 2, because `someField` gets renamed to `e`, which already existed in the base class
```

### Input code

```typescript

```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es2024",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": {
        "props": {}
      }
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

https://play.swc.rs/?version=1.13.4&code=H4sIAAAAAAAAA0vOSSwuVnAGkYYK1bxcCkCQqmCrYGjNy1XLy8XLlYxQYKSQWlGSmpeCrr44PzfVLTM1JwWozwiuLz%2BvOD8nVS8nP10jL7UcaoKGpl6qJlAJAFlNOqN3AAAA&config=H4sIAAAAAAAAA1WOSw6DMAxE95wCed1FhaoueoceIkoNCspPdpCKEHcnX0o3cTwzT56t63uYWcKr3%2BI3Ll4QI517VHi1QXyjAiiNYEnKB7g1d%2BZkjUIzZmkvDgRBE4ZM8XAfHpUA7RxjI6pmlFXjer0pnfGEzP%2FBFBV20niJpsrkfEpue9XKTG9uA8Z9lgsEYfVYmj3hF2olzoOg%2BN3IQAt2%2BwGr9JarLwEAAA%3D%3D

### SWC Info output

    Operating System:
        Platform: win32
        Arch: x64
        Machine Type: x86_64
        Version: Windows 10 Pro
        CPU: (16 cores)
            Models: AMD Ryzen 7 9800X3D 8-Core Processor

    Binaries:
        Node: 23.10.0
        npm: N/A
        Yarn: N/A
        pnpm: N/A

    Relevant Packages:
        @swc/core: 1.13.4
        @swc/helpers: N/A
        @swc/types: 0.1.24

    SWC Config:
        output: N/A
        .swcrc path: N/A

    Next.js info:
        output: N/A

### Expected behavior

After renaming, different properties should have different names

### Actual behavior

_No response_

### Version

1.13.4

### Additional context

I could reproduce the issue with longer property names too, but that only happened after the minification of an entire project (which I can't include here). But I think that's the same issue as this one (with single letter property names).

baltasarblanco:
## What's the problem this PR addresses?

Closes #11027.

When `jsc.minify.mangle.props` is enabled, the property mangler can rename a property in a subclass to a name that already exists as a property of a base class (or of any other object in the program), producing observable shadowing at runtime.

Minimal repro from the issue:

```js
class Class1 { e = 1; }
class Class2 extends Class1 { someField = 2; }
console.log(new Class2().e); // expected: 1
```

Before the fix, `someField` was mangled to `e`, so `new Class2().e` returned `2` instead of the inherited `1`.

## How did you fix it?

Root cause is in `crates/swc_ecma_minifier/src/pass/mangle_props.rs`. The state already builds an `unmangleable` set containing both reserved DOM/JS property names (loaded from `domprops.json` / `jsprops.json`) and all property names seen in the program. But `gen_name` called `Base54Chars::encode` directly, which only skips JavaScript reserved words — it never consults `unmangleable`.

In the repro, `e` is present in `domprops.json`, so `Class1.e` is correctly preserved by `is_reserved`. The encoder then walks `a, b, c, ...` and may emit `e` for `someField`, colliding with the inherited property.

The fix advances the encoder in a loop until it produces a candidate not in `unmangleable`. The set already aggregates exactly what we need to avoid, so no extra bookkeeping is needed. In practice this is 0-2 extra `encode` calls per generated name.

## Testing

- Added `issue_11027_inherited_class_property_collision` in `crates/swc_ecma_minifier/tests/mangle.rs` using the existing `assert_mangled` harness with `props` mangling enabled (the fixture-based suite at `tests/mangle/**` does not enable property mangling, so an inline test is the right vehicle here, in line with the other `props`-related tests in this file).
- `cargo test -p swc_ecma_minifier --test mangle` passes (2217 tests).
- `./scripts/exec.sh` passes (2129/2130, 1 ignored, unchanged).
- `cargo fmt --all` and `cargo clippy -p swc_ecma_minifier --all-targets -- -D warnings` clean.


baltasarblanco:
Regarding the CodSpeed regression on `es/large/codegen/es2016` (-2.23%, 37.1ms → 37.9ms): this benchmark exercises the codegen path, while this PR only modifies `swc_ecma_minifier::pass::mangle_props` — the change is gated behind `mangle.props` and never executes in a pure codegen run. CodSpeed also flagged "Different runtime environments detected" and notes that no successful main run was available, so the comparison fell back to `16a56d0` which may contain unrelated changes. The variance is consistent with runner noise (~0.8ms on a 37ms bench). Acknowledging on CodSpeed.

kdy1 (changes requested):
Can you add some complex test cases?

baltasarblanco:
Added three additional test cases covering deeper inheritance chains, object-literal-and-class namespace sharing, and mixed shapes (getters / methods / fields). Marking as ready for review.

kdy1:
Sorry, I should have been more specific.

Please add test cases for lots of properties to mangle

baltasarblanco:
Added `issue_11027_many_properties_stress_test`: a base class with six fields whose names (`a` through `f`) are all in `domprops.json`, plus a derived class declaring 25 fresh fields. The mangled output confirms none of the 25 generated names collide with the reserved-or-inherited names — the encoder skipped the unavailable single-character candidates and continued onward (`i, l, o, t, m, s, g, ...`), exercising the collision-avoidance loop repeatedly within a single mangle pass.

baltasarblanco:
Thanks for refactoring the stress test to compare runtime stdout via `run_mangle_props_exec_test` — that's a much better way to validate semantic preservation than matching the mangled output textually. I'll keep this pattern in mind for future contributions.

kdy1 (approved):

kdy1:
Thanks!

PR #11849, opened 2026-05-12
security: harden PR workflow permissions

kdy1:
**Description:**

Hardens PR-related GitHub Actions workflows by making token permissions explicit and minimizing credential exposure during untrusted PR jobs.

This changes the PR/test workflow defaults to read-only where possible, disables persisted checkout credentials for jobs that do not push, removes the binary-size comment workflow's trust in PR-uploaded PR-number artifacts, and limits Claude-triggered workflows to trusted repository actors while preserving the existing trusted-user automation path.

The repository Actions default `GITHUB_TOKEN` permission was also changed from write to read.

**BREAKING CHANGE:**

None.

None.

**Validation:**

- Parsed all `.github/workflows/*.yml` files with Ruby YAML
- `git diff --check`
- `cargo fmt --all`

### DIFF
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -31,7 +31,7 @@ runs:
         echo "$HOME/.local/bin" >> "$GITHUB_PATH"
 
     - name: Install Node.js
-      uses: actions/setup-node@v4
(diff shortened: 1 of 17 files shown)


magic-akari:
Maybe we should pin the actions?

PR #11853, opened 2026-05-13
perf(es/minifier): reduce minifier profiling hotspots

hardfist:
## Summary

This draft PR contains incremental minifier performance work guided by local callgrind runs on the `swc_ecma_minifier` full benchmark binary.

Current changes:
- Skip empty finalizer replacement-map lookups.
- Avoid eager pure callee checks when member/call shape makes them unnecessary.
- Reserve mangle rename maps from known symbol counts.
- Skip pure boolean rewrites for non-logical expressions.
- Streamline pure member callee checks and share pure method-call evaluation.
- Guard pure-expression evaluation hooks and reuse sequence alias ids.
- Skip no-op invalid cleanup paths and guard logical rewrite calls.
- Skip non-binary invalid cleanup and unused logical bool casts.
- Avoid empty infection propagation, reuse merge entry access, and clone only infection ids when propagating program data.
- Handle replacer identifiers before generic child traversal.
- Return early from Finalizer for leaf identifiers after the literal replacement check.
- Skip redundant var-without-init counter walks once direct statement scanning already knows work is needed, and stop the counter after it finds a result.
- Skip DCE analyzer/dropper child traversal for leaf identifier, JSX identifier, pattern identifier, and shorthand property nodes.
- Skip usage-analysis child traversal for leaf identifiers in usage, alias infection, and local identifier collectors.
- Return early from the pure optimizer for identifier, literal, and invalid leaf expressions.
- Fast-path bare identifiers in the optimizer while preserving identifier inline/evaluate rewrites.
- Skip leaf expression and identifier-pattern traversal in the var-without-init counter.
- Narrow the optimizer identifier fast path to the global-variable evaluation work that can apply to an unchanged identifier.
- Avoid `data.vars` lookups for global identifier evaluation unless the identifier name is a candidate for rewriting.
- Guard invalid cleanup so the pure optimizer only walks sequence/binary expressions with direct invalid children, while the optimizer still catches shallow nested logical invalids.
- Prune additional pure optimizer hot paths: var-without-init visitors avoid expression/class subtrees, boolean shortening is guarded at recursive and call sites, member simplification is limited to literal object shapes, and binary simplification skips shapes that cannot fold.
- Guard pure optimizer expression passes by expression kind, and skip Finalizer expression replacement checks when literal and hoisted-property replacement maps are both empty.
- Treat `this` as a leaf expression in the pure optimizer so it skips the full expression-pass pipeline.
- Fast-path boolean casts for identifiers, `this`, and function expressions before falling back to generic side-effect analysis.
- Narrow numeric pure optimizer guards to the operators handled by the number/int rewrite passes.

## Measurements

Local callgrind instruction counts with `RUST_LOG=off CI=1`:

- `lodash`: `origin/main` `271,951,203` -> current branch `240,324,539` Ir (`11.6295%` reduction).
- Latest step, `lodash`: previous pushed head `6656765043` `240,514,814` -> `4051e8a768` `240,324,539` Ir (`0.0791%` reduction, `190,275` Ir fewer).
- Latest step, `terser`: previous pushed head `6656765043` `624,150,608` -> `4051e8a768` `623,398,377` Ir (`0.1205%` reduction, `752,231` Ir fewer).

Callgrind outputs for the latest step:
- `/tmp/callgrind.lodash.number-op-guards.out`
- `/tmp/callgrind.terser.number-op-guards.out`

CodSpeed/CI notes:
- CI has been restarted for the new head `4051e8a768`.
- On earlier heads in this PR, `Cargo fmt`, `Cargo clippy`, `Benchmark swc_ecma_minifier`, and `Benchmark swc_es_minifier` completed successfully.
- The dependency license failure seen on earlier heads is the existing `rkyv 0.8.13` / `RUSTSEC-2026-0122` issue and is unrelated to this PR; this branch does not touch dependency manifests.

## Validation

- `cargo fmt --all`
- `RUST_LOG=off CI=1 cargo test -p swc_ecma_utils`
- `RUST_LOG=off CI=1 cargo test -p swc_ecma_minifier`

## Breaking change

None.

## Related issue

None.

### DIFF
--- a/.changeset/wet-bobcats-relate.md
+++ b/.changeset/wet-bobcats-relate.md
@@ -1,0 +1,9 @@
+---
+swc_ecma_minifier: patch
+swc_ecma_transforms_base: patch
+swc_ecma_transforms_optimization: patch
+swc_ecma_utils: patch
+swc_core: patch
+---
+
+perf(es/minifier): reduce minifier profiling hotspots
--- a/crates/swc_ecma_minifier/src/compress/optimize/evaluate.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/evaluate.rs
@@ -29,6 +29,15 @@ impl Optimizer<'_> {
         self.eval_global_vars(e);
     }
 
+    #[inline]
+    fn is_declared_ident(&self, i: &Ident) -> bool {
+        self.data
+            .vars
+            .get(&i.to_id())
+            .map(|var| var.flags.contains(VarUsageInfoFlags::DECLARED))
+            .unwrap_or(false)
+    }
+
     fn eval_fn_props(&mut self, e: &mut Expr) -> Option<()> {
         if self
             .ctx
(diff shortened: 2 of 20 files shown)


kdy1 (approved):
Thanks!

Issue #11850, opened 2026-05-12
[bug] isolated declarations output differs from tsc for parameter default initializer and definite assignment assertion


Timeless0911:
### Describe the bug

When using `jsc.experimental.emitIsolatedDts`, the generated declaration output differs from TypeScript's `tsc --declaration --emitDeclarationOnly` output for the same input.

I noticed two differences:

1. A default parameter initializer is preserved in the `.d.ts` output.
2. A class property definite assignment assertion (`!`) is preserved in the `.d.ts` output.

Both are omitted by `tsc` in the generated declaration file.

### Input

```ts
export async function inspectConfig({
  inspectOptions = {},
}: {
  inspectOptions?: Record<string, unknown>;
}): Promise<void> {
  void inspectOptions;
}

export class BuildManager {
  public assetsMiddleware!: string;
}
```

### SWC script

```ts
import { readFile } from 'node:fs/promises';

import swc from '@swc/core';

const input = 'src/index.ts';
const code = await readFile(new URL(input, import.meta.url), 'utf8');

const result = await swc.transform(code, {
  filename: input,
  jsc: {
    parser: {
      syntax: 'typescript',
    },
    experimental: {
      emitIsolatedDts: true,
    },
  },
});

const isolatedDeclarations = JSON.parse(
  (result as { output?: string }).output!,
).__swc_isolated_declarations__;

console.log(isolatedDeclarations);
```

### Actual SWC output

```ts
export declare function inspectConfig({ inspectOptions = {} }: {
    inspectOptions?: Record<string, unknown>;
}): Promise<void>;
export declare class BuildManager {
    assetsMiddleware!: string;
}
```

### Expected output / tsc output

Generated with:

```sh
tsc src/index.ts --declaration --emitDeclarationOnly --outDir dist-tsc
```

```ts
export declare function inspectConfig({ inspectOptions, }: {
    inspectOptions?: Record<string, unknown>;
}): Promise<void>;
export declare class BuildManager {
    assetsMiddleware: string;
}
```

### Diff

```diff
-export declare function inspectConfig({ inspectOptions, }: {
+export declare function inspectConfig({ inspectOptions = {} }: {
     inspectOptions?: Record<string, unknown>;
 }): Promise<void>;
 export declare class BuildManager {
-    assetsMiddleware: string;
+    assetsMiddleware!: string;
 }
```

### Versions

```txt
@swc/core: 1.15.33
typescript: 6.0.3
node: v25.2.1
[...]

PR #11857, opened 2026-05-15
fix(typescript): strip parameter binding defaults in dts

Timeless0911:
**Description:**

This PR fixes isolated declaration emit for function parameters that use binding pattern default initializers. Declaration emit now recursively drops assignment/default expressions from formal parameter binding patterns while preserving the binding shape, matching tsc and OXC fixture behavior.

Related to #11850


kdy1 (changes requested):
Can we avoid creating a new visitor?

CPunisher:
The alternative is a function to handle `Pat`, matching the `Pat::Assign` to remove the rhs, and recursive handle other nested pattern cases. Not sure if it will be simpler than this visitor.

CPunisher:
Finished with the recursive funciton https://github.com/swc-project/swc/pull/11857/changes/bfce5d36c122aeeb72dbe3d4d545154c88481644#diff-428159ff0b82efc17fe1807aa964ffb7262e3e177df3e5938c57e705978477dfR258-R285

kdy1 (approved):
Great, thanks!

PR #11858, opened 2026-05-15
fix(ts/fast-dts): strip definite assertions in dts

Timeless0911:
**Description:**

This PR aligns isolated declaration emit with tsc for definite assignment assertions on class members. Definite assertions are stripped from class fields and auto-accessors, and public auto-accessor visibility is normalized away in generated d.ts output.

Related to #11850


kdy1 (approved):
Thx. I'll merge later

CPunisher (approved):

PR #11869, opened 2026-05-19
perf(es/codegen): remove JsWriter last_srcmap cache

magic-akari:
**Description:**

I removed the `last_srcmap` cache because its impact appears to be negligible.

In local A/B walltime benchmarks for `es/codegen/large`, removing `last_srcmap` did not produce a stable regression beyond benchmark noise. 
Since `srcmap_done` still handles source map deduplication, this keeps the writer simpler while preserving behavior.

Let's see what CI/CodSpeed reports for this change.

**BREAKING CHANGE:**

- #11867


kdy1 (approved):
Thanks!

PR #11870, opened 2026-05-19
perf: speed up swc_ecma_codegen JsWriter hot path

hardfist:
**Description:**

This PR improves `swc_ecma_codegen` performance in the `JsWriter` hot path without changing benchmark definitions or test fixtures.

Changes:
- replace global `HashSet<(BytePos, line, col)>` source-map dedupe with per-generated-position dedupe
- cache indentation text (`indent_cache`) and write it in one call instead of per-level loops
- streamline `write()` hot-path branching for `Some(span)` vs `None`

Motivation:
- remove hashing and repeated branching from frequent codegen paths
- reduce write calls for pretty-print indentation
- keep emitted output and source-map behavior unchanged while improving throughput

Benchmark evidence (`cargo bench -p swc_ecma_codegen --bench bench -- --warm-up-time 1 --measurement-time 2 --baseline pre_opt`):
- `es/codegen/colors`: `change: [-14.707% -14.454% -14.172%]`
- `es/codegen/large`: `change: [-22.255% -21.769% -21.330%]`

No benchmark files or tests were modified.

**BREAKING CHANGE:**

None.

None.

---

Validation run:
- `cargo fmt --all`
- `cargo test -p swc_ecma_codegen`

### DIFF
--- a/crates/swc_ecma_codegen/src/text_writer/basic_impl.rs
+++ b/crates/swc_ecma_codegen/src/text_writer/basic_impl.rs
@@ -1,8 +1,6 @@
 use std::io::Write;
 
 use memchr::memchr2;
-use rustc_hash::FxBuildHasher;
-use swc_allocator::api::global::HashSet;
 use swc_common::{sync::Lrc, BytePos, LineCol, SourceMap, Span};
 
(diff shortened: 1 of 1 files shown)


hardfist:
@codspeedbot explain why this is faster

PR #11873, opened 2026-05-20
fix(es/minifier): respect ecma for iife temp vars

hardfist:
**Description:**

When compressing arrow IIFEs, the minifier can synthesize temporary variable declarations for extracted parameters. That path always emitted `let`, even when `compress.ecma` was ES5.

This updates the synthesized declaration kind to use `var` below ES2015 and keep `let` for ES2015+, then adds an ES5 fixture covering the arrow IIFE temp-var path. Related snapshots were updated for the same behavior change.

**BREAKING CHANGE:**

None.

N/A

**Testing:**

- `cargo fmt --all`

### DIFF
--- a/.changeset/respect-ecma-iife-temp-vars.md
+++ b/.changeset/respect-ecma-iife-temp-vars.md
@@ -1,0 +1,7 @@
+---
+swc: patch
+swc_core: patch
+swc_ecma_minifier: patch
+---
+
+fix(es/minifier): respect ecma when emitting IIFE temp vars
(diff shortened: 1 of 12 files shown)


Austaras:
I believe the current behavior is fine as there're very few browsers that supports arrow but not let/const.

hardfist:
> I believe the current behavior is fine as there're very few browsers that supports arrow but not let/const.

its actually about performance let and const is much slower in some js engine(like quickjs and safari
Related to https://esbuild.github.io/faq/#:~:text=bundling%20is%20active.-,For%20performance

let | const is about 10x+ slower than var during parsing in [primjs](https://github.com/lynx-family/primjs) engine(which is a modified version of quickjs), when I replace let | const to var in our code, the whole parse time reduced from 453ms to 264ms for our code

kdy1 (dismissed):
Sounds good, thanks!

I'll merge after adding changesets

PR #11894, opened 2026-05-24
Optimize parser token and lexer hot paths

hardfist:
## Summary

- Inline hot lexer word/identifier/keyword token paths.
- Make the lexer keyword callback generic over `FnOnce` instead of passing a `fn` pointer.
- Inline hot parser token-buffer bump helpers.
- Pre-size the parser's local atom store for larger inputs to reduce hstr table rehashing during parse.

## Why

Valgrind/Callgrind on the `typescript.js` parse benchmark showed repeated overhead at the lexer/parser token boundary and a smaller but measurable hstr atom-store cost from string/identifier interning. These changes keep the same parser behavior while giving LLVM better visibility into hot token advancement paths and avoiding avoidable atom-store growth work on large bundles.

Measured locally with `scripts/valgrind-callgrind.sh swc parse 1` from the companion benchmark harness:

| Step | Callgrind Ir |
| --- | ---: |
| SWC baseline before lazy source hash | 690,111,316 |
| SWC after lazy source hash | 643,569,691 |
| After lexer inline + generic keyword callback | 628,040,751 |
| After parser buffer inline | 612,471,050 |
| After atom-store capacity hint | 611,103,226 |

Inline-related delta from the lazy-hash baseline before the atom-store change: `-31,098,641 Ir` (`-4.83%`).

Atom-store capacity hint delta from the previous local step: `-1,367,824 Ir` (`-0.22%`). In the same Callgrind profile, `RawTable<(hstr::dynamic::Item, ())>::reserve_rehash` self cost dropped from `1,523,682 Ir` to `506,780 Ir`.

Total local delta from the lazy-hash baseline: `-32,466,465 Ir` (`-5.04%`). Total local delta from the original SWC baseline: `-79,008,090 Ir` (`-11.45%`).

A broader `skip_space` rewrite was also tested and reverted because it regressed Callgrind Ir, so this PR keeps the optimization targeted to measured hot boundaries and atom-store growth.

## Checks

- `cargo test -p swc_ecma_parser --lib`
- Companion benchmark harness: `cargo test`
- Companion benchmark harness: `scripts/valgrind-callgrind.sh swc parse 1`

### DIFF
--- a/crates/hstr/src/dynamic.rs
+++ b/crates/hstr/src/dynamic.rs
@@ -62,13 +62,23 @@ pub struct AtomStore {
 
 impl Default for AtomStore {
     fn default() -> Self {
-        Self {
-            data: hashbrown::HashMap::with_capacity_and_hasher(64, BuildEntryHasher::default()),
-        }
+        Self::with_capacity(64)
     }
 }
 
 impl AtomStore {
+    /// Create an atom store with enough capacity for `capacity` non-inline
+    /// atoms before rehashing.
+    #[inline]
+    pub fn with_capacity(capacity: usize) -> Self {
+        Self {
+            data: hashbrown::HashMap::with_capacity_and_hasher(
+                capacity,
+                BuildEntryHasher::default(),
+            ),
+        }
+    }
+
     #[inline(always)]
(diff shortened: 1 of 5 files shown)


hardfist:
@codspeedbot fix this regression

PR #11904, opened 2026-05-28
fix(swc): preserve plugin error context

hardfist:
**Description:**

Preserve the full `anyhow` context chain when reporting SWC Wasm plugin transform failures.

Previously plugin errors were reported with `err.to_string()`, which only retained the outermost context. This made ABI/runtime failures hard to diagnose because the diagnostic that reaches consumers such as Rspack could stop at `failed to invoke plugin on ...`.

This now reports plugin errors with `format!("{err:#}")`, so the compact context chain is preserved.

**Error report comparison:**

For a plugin instantiation failure caused by a missing `env.__free` import:

Before:

```text
failed to invoke plugin on 'Some("src/index.jsx")'
```

After:

```text
failed to invoke plugin on 'Some("src/index.jsx")': failed to invoke `@swc/plugin-remove-console` as js transform plugin at @swc/plugin-remove-console: unknown import: `env::__free` has not been defined
```

**BREAKING CHANGE:**

None.

### DIFF
--- a/.changeset/plugin-error-context.md
+++ b/.changeset/plugin-error-context.md
@@ -1,5 +1,6 @@
 ---
 swc: patch
+swc_core: patch
 ---
 
 fix(swc): preserve plugin error context in diagnostics
--- a/.changeset/plugin-error-context.md
+++ b/.changeset/plugin-error-context.md
@@ -1,0 +1,5 @@
+---
+swc: patch
+---
+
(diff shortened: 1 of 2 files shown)


hardfist:
@codspeedbot explain why this is faster

kdy1 (approved):
Thanks!

PR #11803, opened 2026-04-17
fix(es/ast)!: fix panic on JSX surrogate entities

cuyl:
**Description:**
    
This commit fixes a panic (RuntimeError: unreachable) that occurred when the JSX parser
encountered numeric entities representing UTF-16 surrogates (e.g., &amp;#xD83E;&amp;#xDD80;).

The previous implementation used `char::from_u32(...).expect(...)`, which panics if the
value is a surrogate or exceeds the Unicode range. This is now replaced with a safer
`parse_from_code` helper that returns an error instead of panicking.

`swc_ecma_parser` now supports merging consecutive JSX entities that form a
UTF-16 surrogate pair into a single Unicode character. This matches the behavior of
other JSX parsers and ensures that characters outside the Basic Multilingual Plane
are correctly decoded when represented as surrogate pairs.

Changes:
- Replaced panicking `char::from_u32` with safe numeric entity parsing.
- Updated `read_jsx_entity` to return `Wtf8Atom` to support surrogate pairs.
- Implemented merging logic for UTF-16 surrogate pairs across consecutive entities.
- Added test cases covering surrogate pairs in both JSX attributes and text.

**BREAKING CHANGE:**

- Changed JSXText::value to Wtf8Atom for surrogate pairs support.

**Known issue:**

-  `Wtf8Atom`’s generic serde implementation does not emit lone surrogates as JSON ` \uXXXX` escapes. To do that, we need a serde_json-specific serializer path, such as one based on RawValue.

Fixes #11802.

PR #11905, opened 2026-05-28
fix(encoding)!: Fix incorrect fields count

quininer:
**Description:**

Currently, our `encoding` mod does not correctly handle the `ignore` attr when get fields count.

**BREAKING CHANGE:**

This doesn't cause a problem now because both encode and decode use wrong count. However, it's important to note that this fix will result in a breaking change.

### DIFF
--- a/.changeset/fix-encoding-ignored-field-count.md
+++ b/.changeset/fix-encoding-ignored-field-count.md
@@ -1,0 +1,6 @@
+---
+ast_node: major
+swc_core: major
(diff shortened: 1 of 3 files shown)


kdy1 (dismissed):
I'll merge this along with https://github.com/swc-project/swc/pull/12065 and https://github.com/swc-project/swc/pull/11803

kdy1 (approved):

PR #11920, opened 2026-06-04
fix(bindings/node): Preserve source context for AST transforms

kdy1:
**Description:**

This fixes the Node `@swc/core` AST handoff path so source maps do not depend on a process-global `Compiler`/`SourceMap` when JavaScript receives a parsed `Program` and later passes it back to `transform` or `print`.

The native parser now returns an internal envelope with the parsed `Program` plus the original source context. The public JS wrapper unwraps it back to the normal AST shape and stores the source context as non-enumerable metadata. When the AST is passed back to native code, the wrapper sends the envelope; native code restores the source file in a fresh compiler and rebases AST spans into that new source map. Legacy raw `Program` JSON input still falls back to the old behavior.

This also stops parser calls from writing comments into the global compiler comment store, and adds regression coverage for the MUI box-drawing comment case, `print(parse(...))`, and concurrent plugin transforms.

Validation:

- `cargo test -p binding_core_node`
- `pnpm --dir packages/core build:dev`
- `pnpm --dir packages/core exec rstest __tests__/transform/sourcemap_test.js`
- `pnpm --dir packages/core test`
- `cargo fmt --all`

**BREAKING CHANGE:**

None.

Follow-up to #11919. Root cause for the Next.js MUI sourcemap panic originally observed while investigating #11918.

### DIFF
--- a/.changeset/fix-node-ast-source-context.md
+++ b/.changeset/fix-node-ast-source-context.md
@@ -1,0 +1,5 @@
+---
+"@swc/core": patch
+---
+
+fix(core): Preserve source context for AST transform sourcemaps
--- a/.changeset/fix-node-ast-source-context.md
+++ b/.changeset/fix-node-ast-source-context.md
@@ -1,5 +1,0 @@
----
-"@swc/core": patch
----
-
-fix(core): Preserve source context for AST transform sourcemaps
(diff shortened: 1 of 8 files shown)


kdy1 (commented):
I verified that normal `transform()` path is not affected

PR #11926, opened 2026-06-09
fix(es/parser): handle Flow async generic arrows

### LINKED ISSUE #11925: Flow parser rejects valid syntax: `async <T>() =>` arrows and anonymous function-type params in `declare class` methods
### Describe the bug

When parsing with `jsc.parser.syntax: "flow"`, SWC rejects two constructs that are valid Flow and are accepted by the reference Flow parser (`hermes-parser`, also used by `flow-bin`, `babel-plugin-syntax-hermes-parser`, and `prettier-plugin-hermes-parser`):

1. **`async` arrow functions with type parameters** — `async <T>(x: T): Promise<T> => x`
2. **Anonymous function-type parameters in a `declare class` method signature** — `m(a: string, (e: E) => void): void`

Both surfaced migrating a large Flow codebase's Vite build from `flow-remove-types` to `@swc/core` (5,162 files; these were the only two constructs SWC could not parse).

### Input code

```typescript
const { transformSync } = require('@swc/core');
const opts = { filename: 'input.js', jsc: { parser: { syntax: 'flow' } } };

// Case 1 — async arrow with type parameters → THROWS "Expected ',', got ':'"
transformSync('const f = async <T>(x: T): Promise<T> => x;', opts);

// Case 2 — anonymous function-type param in a declare class method
//          → THROWS "Unexpected token `(`. Expected yield, an identifier, [ or {"
transformSync('declare class C { m(a: string, (e: mixed) => void): void; }', opts);


These minimal variants all parse correctly, which isolates each bug:


transformSync('const f = <T>(x: T): T => x;', opts);                       // non-async generic arrow — OK
transformSync('const f = async (x: number): number => x;', opts);         // async, no type params — OK
transformSync('declare class C { m(a: string, cb: (e: mixed) => void): void; }', opts); // named param — OK
```

### Config

```json
{ "jsc": { "parser": { "syntax": "flow" } } }
```

I cant seem setup flow in the playground

### SWC Info output

    Operating System:
        Platform: darwin
        Arch: arm64
        Machine Type: arm64
        Version: Darwin Kernel Version 25.5.0: Mon Apr 27 20:41:15 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6041
        CPU: (16 cores)
            Models: Apple M4 Max

    Binaries:
        Node: 24.15.0
        npm: 11.12.1
        Yarn: N/A
        pnpm: 11.1.0

    Relevant Packages:
        @swc/core: 1.15.40
        @swc/helpers: N/A
        @swc/types: 0.1.26
        typescript: 5.8.3

    SWC Config:
        output: N/A
        .swcrc path: N/A

    Next.js info:
        output: N/A

### Expected behavior

Both inputs parse successfully (they are valid Flow — `hermes-parser` / `@babel/preset-flow` + `babel-plugin-syntax-hermes-parser` accept and transform both).

### Actual behavior

- Case 1: `x Expected ',', got ':'` (at the first typed parameter)
- Case 2: `` x Unexpected token `(`. Expected yield, an identifier, [ or { ``

### Version

`@swc/core` **1.15.40** (also reproduces on 1.15.33). Independent of `jsx`, `components`, `patternMatching`, and `requireDirective`.

### Additional context

Case 1 looks like the Flow-parser counterpart of #9698 (the TypeScript async-arrow-with-type-params bug, since fixed) — the same gap appears to remain on the `syntax: "flow"` path, and triggers even for single-line type parameter lists.

kdy1:
**Description:**

Fixes two Flow parser gaps reported in #11925:

- Parse async generic arrows with typed params and explicit return annotations, such as `async <T>(x: T): Promise<T> => x`.
- Parse anonymous function-type parameters in `declare class` method signatures, preserving the existing `__flow_anon_param_N` synthetic parameter convention.
- Adds focused Flow parser fixtures for both cases.

Validated with:

- `cargo fmt --all`
- `cargo test -p swc_ecma_parser --features flow --test flow -- --ignored`
- `cargo test -p swc_ecma_transforms_typescript flow_strip_reparses_as_javascript -- --ignored`
- `cargo test -p swc_ecma_parser --features flow --test flow_hermes`

**BREAKING CHANGE:**

No.

Fixes #11925

### DIFF
--- a/.changeset/fix-flow-parser-11925.md
+++ b/.changeset/fix-flow-parser-11925.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_parser: patch
+---
+
+fix(es/parser): Parse Flow async generic arrows and declare class anonymous function params
--- a/crates/swc_ecma_parser/src/parser/pat.rs
+++ b/crates/swc_ecma_parser/src/parser/pat.rs
@@ -783,7 +783,9 @@ impl<I: Tokens> Parser<I> {
             let decorators = self.parse_decorators(false)?;
             let pat_start = self.cur_pos();
 
-            let pat = if self.input_mut().eat(Token::DotDotDot) {
+            let pat = if let Some(pat) = self.try_parse_flow_anon_formal_param(params.len())? {
+                pat
+            } else if self.input_mut().eat(Token::DotDotDot) {
                 let dot3_token = self.span(pat_start);
 
                 let mut pat = self.parse_binding_pat_or_ident(false)?;
@@ -827,6 +829,9 @@ impl<I: Tokens> Parser<I> {
                 self.parse_formal_param_pat()?
             };
             let is_rest = matches!(pat, Pat::Rest(_));
+            if is_rest {
+                rest_span = pat.span();
+            }
 
             params.push(Param {
                 span: self.span(param_start),
--- a/crates/swc_ecma_parser/src/parser/typescript.rs
+++ b/crates/swc_ecma_parser/src/parser/typescript.rs
@@ -3214,6 +3214,8 @@ impl<I: Tokens> Parser<I> {
             TsFnParam::Array(param) => Pat::Array(param),
             TsFnParam::Rest(param) => Pat::Rest(param),
             TsFnParam::Object(param) => Pat::Object(param),
+            #[cfg(swc_ast_unknown)]
+            _ => unreachable!(),
         }
     }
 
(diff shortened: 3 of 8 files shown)


matclayton:
Thanks for this, we've run it on a ~1M line flow codebase, and these were the only issues we saw btw. We dont use the modern react syntax, but probably have most other permutations in our codebase!

PR #11940, opened 2026-06-15
fix(es/react-compiler): skip TypeScript `this` pseudo-params in scope collector

wbinnssmith:
TypeScript allows `function f(this: T) {}` to declare the type of this inside a function body.

To swc, this looks like a real parameter. The scope collector was registering it as a binding, causing the compiler to emit an unexpected "reserved word" diagnostic for any hook containing a function with a TS-annotated `this` param.

This returns early specifically when the ident name is "this".


kdy1 (dismissed):

kdy1:
I'll merge after adding changesets

PR #11948, opened 2026-06-19
fix(react-compiler): preserve class syntax context

kdy1:
**Description:**
Preserve the original SWC `Class.ctxt` while converting React Compiler AST back into SWC AST. The reverse converter now looks up the preserved class shell by span and uses its syntax context before loading the full preserved class, falling back to an empty context only when no preserved class exists.

This fixes a hygiene fidelity issue where class syntax context could be lost during round-trip conversion, while keeping the existing class shell restoration behavior for decorators, body, type metadata, and super-class handling. Added a regression test with a non-empty `SyntaxContext` to lock the behavior.

Validation:
- `cargo fmt --all`

**BREAKING CHANGE:**

None.

N/A

### DIFF
--- a/crates/swc_ecma_react_compiler/src/preserved_ast.rs
+++ b/crates/swc_ecma_react_compiler/src/preserved_ast.rs
@@ -1,5 +1,5 @@
 use rustc_hash::FxHashMap;
-use swc_common::{source_map::SmallPos, util::take::Take, Span};
+use swc_common::{source_map::SmallPos, util::take::Take, Span, SyntaxContext};
 use swc_ecma_ast::*;
 
 #[derive(Default, Clone)]
@@ -224,6 +224,15 @@ impl PreservedAst {
             class.span.lo.to_u32(),
             PreservedNode::Class(Box::new(class.clone())),
         );
+    }
+
+    pub fn class_ctxt_for_span(&self, span: Span) -> Option<SyntaxContext> {
+        let key = span.lo.to_u32();
+        let Some(PreservedNode::Class(snapshot)) = self.nodes.get(&key) else {
+            return None;
+        };
+
+        Some(snapshot.ctxt)
     }
 
     pub fn load_class(&mut self, class: &mut Class) -> bool {
--- a/.changeset/preserve-class-syntax-context.md
+++ b/.changeset/preserve-class-syntax-context.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_react_compiler: patch
+---
+
+fix(es/react-compiler): Preserve class syntax context during reverse conversion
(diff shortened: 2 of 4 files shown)


magic-akari:
I checked the standard swc transform pipeline.

In the normal `swc` config path, React Compiler runs before both `resolver` and the final `hygiene` pass:

1. The file is parsed into an AST.
https://github.com/swc-project/swc/blob/6fbe188741cebee363e7afeec4b0c4aa96144070/crates/swc/src/config/mod.rs#L339
2. `swc_ecma_react_compiler::transform(...)` runs.
https://github.com/swc-project/swc/blob/6fbe188741cebee363e7afeec4b0c4aa96144070/crates/swc/src/config/mod.rs#L344
3. `resolver(...)` runs afterwards.
https://github.com/swc-project/swc/blob/6fbe188741cebee363e7afeec4b0c4aa96144070/crates/swc/src/config/mod.rs#L395
4. `hygiene_with_config(...)` runs near the end, before `fixer`.
https://github.com/swc-project/swc/blob/6fbe188741cebee363e7afeec4b0c4aa96144070/crates/swc/src/config/mod.rs#L792

So in the normal swc pipeline, React Compiler does not see hygiene-generated syntax contexts. Also, the parser creates `Class` nodes with the default/empty context, so `Class.ctxt` should normally be empty when React Compiler runs.

Because of that, if this PR is meant to fix loss of hygiene-generated `Class.ctxt` in the standard swc React Compiler pipeline, I think the fix may be unnecessary.

kdy1:
It's about next.js, not @swc/core. 

@wbinnssmith Can you check if this branch fixes the problem?

crates/swc_ecma_react_compiler/src/preserved_ast.rs
```diff
@@ -226,6 +226,15 @@ impl PreservedAst {
         );
     }
 
+    pub fn class_ctxt_for_span(&self, span: Span) -> Option<SyntaxContext> {
```
magic-akari, line 229: Currently, all fields except ⁠`super_class`⁠ are restored in ⁠`load_class`⁠, including ⁠`ctxt⁠`. Therefore, ⁠`class_ctxt_for_span`⁠ might be redundant.

magic-akari (changes requested):

crates/swc_ecma_react_compiler/src/preserved_ast.rs
magic-akari: If ⁠`load_class⁠` fails, ⁠`class_ctxt_for_span⁠` will fail as well, since they share the same key.

kdy1:
Closing as I think this PR was a wrong approach

PR #11967, opened 2026-06-26
test(react-compiler): add build-pass fixtures for wrapped assignment targets

CPunisher:
**Description:**

This updates the React Compiler fixture harness to separate fixture expectations into:

- `compile-pass`: existing fixtures that must produce compiled output and match snapshots
- `build-pass`: fixtures that only assert the transform pipeline does not panic, regardless of whether React Compiler emits a compiled program

It also adds a regression fixture for wrapped assignment targets covered by #11952, including parenthesized targets, TS assertions, `satisfies`, non-null assertions, TS instantiation, update expressions, `for-in` targets, and `super` property assignment/update cases.

- Follow-up to #11952


kdy1 (approved):
Thank you!

PR #11968, opened 2026-06-27
fix(es/fixer): normalize for-head ident patterns

magic-akari:
**Description:**

Normalize paren remover output for parenthesized identifiers in `for-in` / `for-of` heads.

The parser can represent cases like `for ((x) of y) {}` and `for ((x) in y) {}` as `ForHead::Pat(Pat::Expr(...))`. After redundant parentheses are removed, the AST could still contain `Pat::Expr(Expr::Ident(x))`, leaving an expression-shaped identifier in a pattern position.

This PR normalizes that final shape to `Pat::Ident(x)` from `visit_mut_for_head`, so it matches the AST shape produced by `for (x of y) {}` / `for (x in y) {}`. It also adds AST-shape coverage for nested parentheses such as `for ((((x))) of y) {}`.

Verification:
- `cargo fmt --all`
- `cargo test -p swc_ecma_transforms_base paren_remover_normalizes_ident_for_head_pat_exprs`
- `cargo test -p swc_ecma_transforms_base`

N/A

### DIFF
--- a/.changeset/normalize-for-head-ident-patterns.md
+++ b/.changeset/normalize-for-head-ident-patterns.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_transforms_base: patch
+---
+
+fix(es/fixer): Normalize for-head identifier patterns after removing parentheses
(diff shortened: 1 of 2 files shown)


kdy1 (dismissed):

kdy1:
I'll merge after adding the changeset

PR #11971, opened 2026-06-29
fix(es/minifier): preserve switch fallthrough termination

kdy1:
**Description:**

Fixes switch termination detection so fallthrough from matched cases into later cases is accounted for before dead-code elimination. This preserves code after switches like `default: return 1; case "0":`, where the later empty case can still complete normally.

This PR updates `StmtExt::terminates`, adds focused switch termination tests, adds issue #11970 minifier fixture coverage, and adds an exec regression proving `classify("0")` still reaches the post-switch `return 2`.

**Testing:**

- `git submodule update --init --recursive`
- `UPDATE=1 cargo test -p swc_ecma_minifier`
- `cargo test -p swc_ecma_utils`
- `cargo test -p swc_ecma_minifier`
- `(cd crates/swc_ecma_minifier && ./scripts/exec.sh issue_11970)`
- `cargo fmt --all`
- `cargo clippy --all --all-targets -- -D warnings`

**Related issue (if exists):**

Closes #11970

PR #11972, opened 2026-06-29
fix(es/minifier): Check last case

Austaras:
**Description:**
#11971 doesn't seem correct. If a case termiante check to be `Ok(false)` it would rely on its previous case.

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_utils/src/lib.rs
+++ b/crates/swc_ecma_utils/src/lib.rs
@@ -499,6 +499,7 @@ pub trait StmtExt {
             allow_break: bool,
             allow_throw: bool,
         ) -> Result<bool, ()> {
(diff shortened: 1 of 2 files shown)


crates/swc_ecma_utils/src/lib.rs
```diff
@@ -3926,7 +3934,7 @@ switch (foo) {
 
     #[test]
     fn switch_non_terminating_case_falls_through_to_terminating_case() {
-        assert!(stmt_in_function_terminates(
+        assert!(!stmt_in_function_terminates(
```
Austaras, line 3937: This doesn't terminate at all.

kdy1 (approved):

kdy1:
Nice catch, thanks!

PR #11973, opened 2026-06-29
feat: expose ast utilities in typescript_wasm

avivkeller:
Currently, https://github.com/nodejs/node relies on two separate AST parsers, as SWC does not expose it's AST bindings to JS-land. This PR changes SWC to expose those bindings, allowing for consumers, like node core, to parse AST using SWC.

See: https://github.com/nodejs/amaro/pull/435

### DIFF
--- a/bindings/binding_typescript_wasm/Cargo.toml
+++ b/bindings/binding_typescript_wasm/Cargo.toml
@@ -23,6 +23,12 @@ owo-colors = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 serde-wasm-bindgen = { workspace = true }
 swc_common = { path = "../../crates/swc_common" }
+swc_ecma_ast = { path = "../../crates/swc_ecma_ast", features = [
+  "serde-impl",
+] }
+swc_ecma_parser = { path = "../../crates/swc_ecma_parser", default-features = false, features = [
+  "typescript",
+] }
 swc_error_reporters = { path = "../../crates/swc_error_reporters" }
 swc_ts_fast_strip = { path = "../../crates/swc_ts_fast_strip", features = [
   "wasm-bindgen",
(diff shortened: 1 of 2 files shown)


kdy1 (changes requested):
Can we do rust-side transform/analysis code instead of js side code?
Asking because it would be really slow to pass AST from Rust to js

kdy1:
If it's really inevitable I can merge this PR, but I think it's not a good idea for performance

avivkeller:
Post https://github.com/nodejs/node/pull/64034 (which removes our dependency on AST parsing for part of the REPL), we use AST parsing to:
- Get the first expression in a code string based on a column (used for determining where an assertion error came from)
- Transform static imports to dynamic ones
- Determine if JS syntax is unterminated, and therefor recoverable
- Determine if JS syntax is valid

We can probably determine if JS syntax is valid + recoverable (in the REPL) via CDP's `Runtime.compileScript` (which'll determine the validity of syntax). (See chrome's implementation of `IsExpressionComplete`)

For me, it feels weird to expose Node.js-internal-use-only Rust bindings (like `getFirstExpression` or `transformImports`), but WDYT? I can also get some Node.js-ers opinions.

kdy1:
I made https://github.com/swc-project/swc/pull/11975 based on the requirements of https://github.com/nodejs/node/pull/64034

Note: I made it with Codex, so I need to verify it. I'm busy atm, so I can verify later this week

kdy1:
@avivkeller Does the API in https://github.com/swc-project/swc/pull/11975 looks good to you?

kdy1:
@avivkeller As an alternative, I can create a nodejs-only binding wasm. What do you think about it?

avivkeller:
Sorry for the delay here! Been extremely swamped, thanks for the second ping. I'd love for some Node.js-ers to give their opinions before we proceed with anything, cc @marco-ippolito @jakobjingleheimer

marco-ippolito:
I dont mind the experiment we can release a nightly of amaro

avivkeller:
@kdy1 both ideas look good to me, so which ever you prefer

PR #11978, opened 2026-07-02
fix(es/minifier): always run hygiene before the mangler

### LINKED ISSUE #11977: Minifier emits duplicate top-level `let x` / `var x` ("Identifier 'x' has already been declared") with compress + mangle
## Describe the bug

With `compress` and `mangle` both enabled, the minifier can emit a program with two top-level declarations of the same identifier (`let x;` plus `var x = …`) in the same scope, which is a `SyntaxError: Identifier 'x' has already been declared`. The **input is valid JS; the minified output is not.**

The compressor inlines an IIFE whose parameter is `x`, hoisting a top-level `let x;` for the parameter. Independently, the mangler renames another top-level binding to `x`. Neither pass reconciles with the other, so the two collide at the top level.

Downstream report (this breaks Turbopack / Next.js production builds): https://github.com/vercel/next.js/issues/95372

## Input code

```js
var Subscription = function() {
  function Subscription1(listeners, listener) {}
  Subscription1.prototype.add = function(subscription) {
    if (this.unsubscribed) {}
  };
}();
try {
  serverOnlyRequire = eval('require');
} catch (err) {}
var __require = /* @__PURE__ */ ((x)=>("TURBOPACK compile-time truthy", 1) ? __turbopack_context__.z : "TURBOPACK unreachable")(function(x) {
})(ErrorType || {});
var isRequestError = (error)=>{
  let u = i.getKey ?? ((p, s)=>`x`), f = async ()=>{
    try {} catch (s) {}
  };
}
var x = class extends d {};
```

## Config

```json
{ "compress": {}, "mangle": {} }
```

## Reproduce (`@swc/core@1.15.43`)

Save the input above as `input.js`, then:

```js
import { minifySync } from '@swc/core';
import { readFileSync } from 'node:fs';
import vm from 'node:vm';

const code = readFileSync('input.js', 'utf8');
const { code: out } = minifySync(code, { compress: {}, mangle: {} });
console.log(out);
new vm.Script(out); // throws
```

```
$ node repro.mjs
let x;var Subscription=function(){(function(r,e){}).prototype.add=function(r){this.unsubscribed}}();try{serverOnlyRequire=eval("require")}catch(err){}var __require=(x=function(r){},__turbopack_context__.z)(ErrorType||{}),isRequestError=r=>{i.getKey},x=class extends d{};

SyntaxError: Identifier 'x' has already been declared
```

Note the two top-level declarations in the output: `let x;` (hoisted from the IIFE parameter by the compressor) and `var x=class extends d{}` (the mangler renamed this binding to `x`).

## Expected behavior

Minified output should always be valid JS. The compressor-hoisted binding and the mangler-renamed binding must not collide — e.g. one should be renamed to `x1`.

## Notes

- **Both passes are required.** With `mangle` disabled (`{ compress: {}, mangle: false }`) the compressor already renames its hoisted binding to `x1` and the output is valid. The duplicate is introduced only by the **interaction** of compress + mangle.
- The bug depends on the surrounding top-level declarations — they steer the mangler's frequency-based name allocation to reuse `x`. Reducing the case further can make it stop reproducing, so the snippet above is close to minimal.

## Version

- Reproduces on `@swc/core@1.15.43` (latest at time of filing).
- Also reproduces via the Rust crates shipped by Turbopack: `swc_core` `69.0.0` / `swc_ecma_minifier` `55.0.2`.

lukesandberg:
## Description

Fixes a `compress` + `mangle` interaction that produces invalid output:

```
SyntaxError: Identifier 'x' has already been declared
```

### Root cause

The minify pipeline skipped `hygiene()` whenever the mangler was enabled — `if !is_mangler_enabled { hygiene() }` — on the assumption that the mangler renames everything hygiene would. **That assumption is false.** The mangler preserves names it is not allowed to rename, most importantly top-level bindings, which with the default `mangle.top_level = false` are never renamed at all. It therefore cannot deconflict a synthesized binding that collides with a preserved one.

IIFE inlining hoists a parameter (or body-level `var`) binding into the enclosing scope, reusing its original nested syntax context. When that lands at the top level next to a user binding of the same name, the two collide. `hygiene()` would rename one of them (e.g. to `x1`), but it was being skipped in exactly the configuration that needs it (compress + mangle — the default `@swc/core` script setup).

**Two conditions must align, which is why it's rare:**
1. The unused IIFE parameter is normally removed by dead-code elimination *before* it can be hoisted. It survives here only because the module contains a top-level `eval` (`serverOnlyRequire = eval('require')`), which sets `HAS_EVAL_CALL` on the top-level scope and disables that elimination. This `eval('require')` shim is Turbopack runtime boilerplate — hence the downstream Turbopack/Next.js breakage.
2. SWC's parser does not enforce lexical redeclaration, so the invalid output shipped silently; Node rejects it only at load time.

### Fix

Run `hygiene()` unconditionally after `optimize()` in every minify entry point:
- `Compiler::minify` (`crates/swc/src/lib.rs`)
- the `transform()` minifier pass (`crates/swc/src/config/mod.rs`)
- the node minifier binding (`bindings/binding_minifier_node/src/minify.rs`)

and mirror it in the minifier test harness. Because hygiene runs *after* all compressor passes, it does not desync usage data or regress optimization (an earlier attempt that remapped the binding mid-pipeline defeated `reduce_vars`/`evaluate`; this does not). The wasm minifier binding inherits the fix through `Compiler::minify`.

This addresses the whole class of bug — any synthesized binding that collides with a preserved name — not just IIFE inlining.

### Fixture churn

A handful of fixtures update because hygiene now disambiguates shadowed names — e.g. a nested `function n()` the mangler had left shadowing an outer `n`, or a `catch (a) { var a }` redeclaration. These were latent bugs the same misconfiguration produced without an error. All updated outputs are byte-verified valid via `node --check`.

## Related Issue

Closes #11977
Downstream: vercel/next.js#95372

## Tests

- Repro fixture `crates/swc_ecma_minifier/tests/fixture/issues/vercel/issue-11977/` (compress `{defaults:true}` + mangle `{}`). The first commit captures the broken output as a baseline; the fix commit updates it to valid JS.
- `swc_ecma_minifier`: compress / mangle / eval / format / exec / terser_exec all green.
- `swc`: full suite green (two pre-existing local stacktrace-format failures, unrelated, confirmed identical on the base commit).
- `cargo fmt`; `clippy -D warnings` clean on all touched crates.


kdy1 (changes requested):
`hygiene` should never be run when the mangling is enabled.
Also, the perf regression should be fixed

PR #11980, opened 2026-07-03
fix(es/minifier): avoid hoisted binding mangle collisions

### LINKED ISSUE #11977: Minifier emits duplicate top-level `let x` / `var x` ("Identifier 'x' has already been declared") with compress + mangle

kdy1:
**Description:**

Fixes a minifier interaction where the compressor could hoist an IIFE parameter into a top-level lexical declaration and the name mangler could later reuse the same printed name for another top-level binding. The rename analyzer now distinguishes true source top-level bindings from nested-context bindings that were physically hoisted to the program root, so hoisted bindings can still be renamed to avoid invalid output.

Added a regression fixture for issue 11977 covering `compress` + `mangle` with the reported eval/top-level binding collision.

**BREAKING CHANGE:**

None.

Fixes #11977

**Validation:**

- `UPDATE=1 cargo test -p swc_ecma_minifier --test compress 11977 -- --ignored --nocapture`
- `cargo test -p swc_ecma_minifier --test compress 11977 -- --ignored --nocapture`
- `node --check crates/swc_ecma_minifier/tests/fixture/issues/11977/output.js`
- `./scripts/exec.sh`
- `cargo fmt --all`


Austaras:
This doesn't really solve the underlying problem consider
```js
g.foo = function() {
var Subscription = function() {
  function Subscription1(listeners, listener) {}
  Subscription1.prototype.add = function(subscription) {
    if (this.unsubscribed) {}
  };
}();
try {
  serverOnlyRequire = eval('require');
} catch (err) {}
var __require = /* @__PURE__ */ ((x)=>("TURBOPACK compile-time truthy", 1) ? __turbopack_context__.z : "TURBOPACK unreachable")(function(x) {
})(ErrorType || {});
var isRequestError = (error)=>{
  let u = i.getKey ?? ((p, s)=>\`x\`), f = async ()=>{
    try {} catch (s) {}
  };
}
var x = class extends d {};
console.log(x, x)
}
```

kdy1:
Closing as this PR was a wrong approach

PR #11984, opened 2026-07-04
fix(es/minifier): Disable IIFE invoke when there's eval

Austaras:
**Description:**
See comment.

And I still don't really understand why this works in `transform` but not `minify`

**BREAKING CHANGE:**

#11977


kdy1 (approved):
Thank you!

PR #11987, opened 2026-07-05
fix(es/minifier): Invoke IIFE when has eval

Austaras:
**Description:**
This PR changes invoke IIFE strategy into two rules:
1. abandon invoke if inserted scope contains eval and invoke IIFE would add ident(may_add_ident)
2. or the IIFE itself has eval

**BREAKING CHANGE:**
Now every arrow function has its own syntax context marked. It might break something


kdy1 (approved):
Wow. Nice work!

PR #11992, opened 2026-07-05
refactor(es/module): Align module transform records with spec terms

magic-akari:
**Description:**

This PR refactors the shared CommonJS/AMD/UMD module-lowering collection layer to use terminology and boundaries closer to ECMAScript module records.

The old `ModuleDeclStrip` abstraction mixed declaration stripping, module request collection, export collection, and emitter-facing naming. This change renames and reshapes that layer into `module_record.rs` with clearer concepts:

- `ModuleRecordCollector`
- `RequestedModules` / `RequestedModule`
- `ModuleRecordEntry`
- `ModuleRequestUsage`
- `LocalExportEntries`
- `ModuleRecordEntryReducer`
- `ExportBinding`

CJS, AMD, and UMD now consume the same module-record-oriented data model while preserving their existing emit behavior. The format-specific passes still own wrapper shape, dependency emission, interop helper application, `export =` handling, and import-reference rewriting.

This is intended to be a behavior-preserving refactor. The main goal is to make the transform pipeline easier to reason about and maintain before future module interop work.

The README was also expanded with diagrams and a short walkthrough of the transform flow.

Verification:

- `cargo fmt --all`

**BREAKING CHANGE:**

None.

N/A

### DIFF
--- a/.changeset/align-module-records.md
+++ b/.changeset/align-module-records.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_transforms_module: patch
+---
+
+refactor(es/module): Align module transform records with spec terms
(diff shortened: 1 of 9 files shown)


kdy1 (approved):
Thanks!

PR #11998, opened 2026-07-07
fix(es/minifier): Make Infect Collect collect every used ident

### LINKED ISSUE #11983: minifier drops a top-level const that a live (inlined) function body still references → ReferenceError at runtime
### Describe the bug

The minifier removes a top-level `const` declaration while an inlined copy of a
function body that reads it stays alive, leaving a free identifier in the
output. Calling the surviving function throws `ReferenceError: A is not
defined`.

It looks like the usage analysis miscounts: the const's only *init-time* read
sits in a call that gets constant-folded, so the binding is treated as dead —
but the same function is also called dynamically, and its body (which still
references the const) is inlined at that second call site after the fold.

This appears related to the known-issues meta #2934 (*"reduce var declaration
if a previous var is only used in its initialization"*) and to PR #9107
(closed, unmerged), but it reproduces on the current release and is distinct
from #8825/#8826 (fixed in 2024).

### Input code

```js
const N = 260;
const A = 40 + N;
function f(w) { return A / w; }
const S = f(621);
export function dyn(w) { return f(w) + S; }
```

Reproduce via the JS API:

```js
const { code } = await swc.minify(src, {
  compress: true,
  mangle: true, // also reproduces with mangle: false
  module: true,
});
```

### Actual output

```js
let n=300/621;export function dyn(t){return A/t+n}
```

`dyn(100)` → `ReferenceError: A is not defined`. Note `A` survives un-mangled —
the mangler no longer owns the name because the declaration was eliminated.

### Expected output

`A`'s binding must be kept, or its value inlined into the surviving body:

```js
let n=300/621;export function dyn(t){return 300/t+n}
```

### Ingredients (from reducing a real-world production crash)

- `A`'s initializer must reference another binding: `const A = 40 + N`
  triggers the bug; a literal `const A = 300` or a plain alias `const A = N`
  compile correctly (the value is inlined into `f` instead).
- The folded call's result must land in a **non-exported** binding: making it
  `export const S = f(621)` makes the bug disappear.
- The second call to `f` must be live at runtime (any dynamic argument).

### Version matrix

Byte-identical broken output on every version tested, spanning Jan 2024 →
today: `@swc/core` 1.3.107, 1.4.17, 1.5.29, 1.6.13, 1.7.42, 1.9.3, 1.10.18,
1.11.31, 1.12.14, 1.13.5, 1.14.0, **1.15.43 (latest)**.

### Real-world impact (Next.js / Turbopack)

This miscompiled a module-scope layout constant in Langfuse's production
Next.js 16.2.9 bundle (Turbopack vendors `swc_ecma_minifier` 45.0.0 /
`swc_core` 57.0.0 in `@next/swc`): opening a side panel crashed the page with
`ReferenceError: COLLAPSED_RAIL_BOUNDARY_PX is not defined` — production
builds only, `next dev` is unminified. Revert: langfuse/langfuse#14736;
workaround (function-local const): langfuse/langfuse#14735.

<details>
<summary>3-file Next.js 16.2.9 repro (crash on page load)</summary>

`package.json`

```json
{
  "private": true,
  "scripts": { "build": "next build", "start": "next start" },
  "dependencies": { "next": "16.2.9", "react": "19.2.4", "react-dom": "19.2.4" }
}
```

`lib/repro.js`

```js
const REPRO_BASE_PX = 260;
const REPRO_BOUNDARY_PX = 40 + REPRO_BASE_PX;
function reproShareFor(w) {
  return REPRO_BOUNDARY_PX / w;
}
const REPRO_STATIC_SHARE = reproShareFor(621);
export function reproDyn(w) {
  return reproShareFor(w) + REPRO_STATIC_SHARE;
}
```

`pages/index.js`

```js
import { useEffect, useState } from "react";
import { reproDyn } from "../lib/repro";

export default function Home() {
  const [v, setV] = useState(null);
  useEffect(() => {
    setV(reproDyn(window.innerWidth));
  }, []);
  return <p>{v === null ? "pending" : String(v)}</p>;
}
```

`npm run build && npm run start` → open the page →
`ReferenceError: REPRO_BOUNDARY_PX is not defined`, "Application error: a
client-side exception has occurred". The emitted chunk contains
`REPRO_BOUNDARY_PX/window.innerWidth+.4830917874396135` with no declaration of
`REPRO_BOUNDARY_PX` anywhere in the bundle.

</details>

### Config

```json
{
  "compress": true,
  "mangle": true,
  "module": true
}
```

### Version

@swc/core 1.15.43 (macOS arm64) via `swc.minify()`; also every version in the
matrix above, and `swc_ecma_minifier` 45.0.0 as vendored by Next.js 16.2.9
Turbopack.

Austaras:
**Description:**

**BREAKING CHANGE:**

Closes #11983


crates/swc_ecma_minifier/src/compress/optimize/sequences.rs
```diff
@@ -1102,56 +1099,7 @@ impl Optimizer<'_> {
...
-                };
-
-                if deps.iter().any(|(id, kind)| {
-                    id == &e_id && matches!(kind, AccessKind::Reference | AccessKind::Call)
-                }) {
-                    return false;
-                }
-            }
-
-            if !self.assignee_skippable_for_seq(a, &e_id) {
```
Austaras, line 1154: This was already made useless by #7804

crates/swc_ecma_minifier/src/compress/optimize/inline.rs
```diff
@@ -437,39 +437,7 @@ impl Optimizer<'_> {
...
-                            if let Some(v_usage) = self.data.vars.get(&id) {
-                                if v_usage.flags.contains(VarUsageInfoFlags::REASSIGNED) {
-                                    return;
-                                }
-                            } else {
-                                return;
-                            }
-                        }
-                    }
+                    Expr::Fn(_) | Expr::Arrow(_) => {}
```
Austaras, line 441: These doesn't seem to be useful at all

kdy1 (dismissed):

PR #12007, opened 2026-07-09
fix(react-compiler): Remove React-like prefilter

imjordanxd:
**Description:**

Remove SWC's React-like function prefilter and defer compilation eligibility to React Compiler.

The prefilter duplicates React Compiler policy and rejects source before the compiler can evaluate it. For example:

```jsx
function createFoo() {
  return function Foo() {
    "use memo";
    return <div>hello, world</div>;
  };
}
```

React Compiler's Babel integration compiles the annotated inner function, but SWC previously returned the source unchanged because `createFoo` did not look React-like and the prefilter intentionally did not inspect nested function bodies.

This pattern is not necessarily a React-blessed use of the directive nor is it conventional React. Regardless, the SWC adapter should not impose a stricter eligibility policy or limit valid uses of the React Compiler API. React Compiler should remain responsible for deciding what to compile.

This PR removes the React-like heuristic while retaining the separate guard for unsupported resource-management declarations. It also adds a fixture covering the nested directive case.

N/A

**Verification:**

- `cargo fmt --all`

---

_Prepared with assistance from OpenAI Codex 🤖._


magic-akari:
I like this direction and think it's the right call. My only concern: are we diverging from React Compiler's Babel plugin behavior?

imjordanxd:
@magic-akari In my opinion, no—I think this brings SWC closer to the Babel plugin's behavior.

Babel traverses all function declarations and expressions, only skipping a function body after selecting that function for compilation. Since `createFoo` is not selected, traversal continues into its body, where the nested `Foo` is opted in by `"use memo"`.

The Rust compiler entrypoint already follows that discovery behavior. SWC's file-level React-like prefilter was an additional early check that prevented the compiler from seeing this case. I've added a fixture to cover it.

Thanks for raising this, though—please let me know if there's another Babel-side filter I've overlooked.

_Sent from OpenAI Codex 🤖._

magic-akari (approved):

Issue #11684, opened 2026-03-15
`es/minifier`: useless arguments aren't removed from new expr


Le0Developer:
### Describe the feature

- `n=function(){throw 1}("test");` -> `n=function(){throw 1}();` "test" is removed
- `n=new function(){throw 1}("test");` -> `n=new function(){throw 1}("test");` "test" is kept

(The `throw 1` is there to prevent swc from removing the function entirely)

### Babel plugin or link to the feature description

https://play.swc.rs/?version=1.15.18&code=H4sIAAAAAAAAA8uzzUstV0grzUsuyczP09CsLskoyi9XMKzVUCpJLS5R0rTmyrPFJw0AoWDVmUEAAAA%3D&config=H4sIAAAAAAAAA42VzW7bMAyA732KwOceumIrhj3AbnsGQbEoR5ksGSLlxijy7qVlu3MbKtgliPmJ4j%2F19nA4NGdsm1%2BHN%2F7LH4NOCOnjmyU4BdIXljTQ9hrb5AZqHjd6xhlZ7RGK6LqQhnTqgIoWPj89f181Gh8jwqaxynoXnJ32NtvYDwkQdzKW8pW5h0CzmFKGxz1K8VWQH2P0oMMdojQqFwg6SPjZr9UT7%2FWAoEadhFtmR3VyGCUTM8wERg0pDiIPxpGLgW3eUgPaqDYaEJBL0JIbQVJjW6wWkMMT4inYwDF3XSnzF20Ytc%2BaBJtwKRVhb2%2FZKTokZXOQMrjASgoWuOb2q6azKgHlJFg8RxcqJfkLwAnwGjHoHqR7ywnL3VTTtnc1XbDcsDQJnLtbijJAxzlVzlkhsXNmIJGTipnA5BbmzLZ1KqcBnQEF1nKjSPTVUXuSTNI0QLQC4Opqyf9FrmoTuOJ5Fur0NwdIYmutB3pNpyrEqT9GaYQ2ZaBTNHXOFaBYpYlXw2Wo4hwMcD%2BAkU5klOTzdJKiqHzZkDf9wCPBF6rOx6O4GQbu7VK6H6vs%2BrGOex26Mve7tUlx8DCCr%2FXzf8zKXcoOj3N7byN%2BM8bwU5By8nhxfnv69HpwKA%2Fbbwmq6aPJu4BKey6vykvz79D2gGyJahz%2B2RSX26%2FvIzH0OOsGAAA%3D

[...]

PR #12012, opened 2026-07-10
fix(es/minifier): drop pure extra args of new function expr

skku970412:
## Summary

This fixes the direct function-constructor case from #11684.

The minifier can already remove useless pure arguments from a direct function-call IIFE. This PR adds a small corresponding cleanup for a known-safe `NewExpr` shape:

```js
new function() {
    throw 1;
}("test")
```

## Root cause

`CallExpr` and `NewExpr` are optimized through separate visitor paths. The existing call-IIFE argument cleanup does not run for direct function-constructor expressions.

## Design

After maintainer feedback, this PR intentionally keeps the optimization narrow and avoids introducing new visitors.

It only handles direct anonymous, parameterless function constructors with a trivially safe body, currently an empty body or a single literal `throw` statement.

It removes only side-effect-free arguments and skips the entire cleanup if any spread argument is present.

More complex constructor bodies are intentionally skipped instead of being analyzed.

## Scope and non-goals

This PR does not optimize:
- named function expressions,
- functions with parameters,
- default/rest parameters,
- non-trivial constructor bodies,
- `this` / `arguments` / `new.target` observability cases,
- class computed keys,
- class heritage expressions,
- arbitrary constructors,
- spread-containing argument lists,
- side-effectful arguments.

## Fixture coverage

Added SWC-owned fixture coverage under:

```text
crates/swc_ecma_minifier/tests/fixture/issues/11684
```

Covered:
- reported pure string argument removal,
- side-effectful argument preservation,
- spread-containing argument-list preservation,
- functions with parameters being skipped,
- non-trivial constructor bodies being skipped,
- class computed key bodies being skipped as non-trivial,
- arbitrary constructor preservation.

## Commands run

```text
cargo fmt --all
UPDATE=1 cargo test -p swc_ecma_minifier --test compress custom_fixture_tests__fixture__issues__11684__input_js
cargo test -p swc_ecma_minifier --test compress custom_fixture_tests__fixture__issues__11684__input_js
cargo test -p swc_ecma_minifier --test exec issue_11684_new_function_expr
cargo test -p swc_ecma_minifier
cargo clippy -p swc_ecma_minifier --all-targets -- -D warnings
./scripts/test.sh
./scripts/exec.sh
UPDATE=1 ./scripts/test.sh
git diff --check
```

## Risk notes

The change is intentionally conservative. It only removes arguments from direct anonymous, parameterless function constructors with a trivially safe body and only when the arguments are side-effect-free.

More complex cases are skipped rather than analyzed.

Issue: https://github.com/swc-project/swc/issues/11684


crates/swc_ecma_minifier/src/compress/optimize/iife.rs
```diff
@@ -500,6 +500,66 @@ impl Optimizer<'_> {
...
+                        .iter()
+                        .any(|param| matches!(&param.pat, Pat::Rest(_))),
+                )
+            }
+            _ => return,
+        };
+
+        if let Some(scope) = self.data.get_scope(ctxt) {
+            if scope.intersects(ScopeData::USED_ARGUMENTS.union(ScopeData::HAS_EVAL_CALL)) {
+                return;
```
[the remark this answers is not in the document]
skku970412: Thanks, fixed in b9599903. I tightened this from a lexical `arguments` check into a constructor-argument observability guard.

The new-function cleanup now bails out when the constructor body may reach the current constructor function object or dynamically observe invocation arguments, including lexical `arguments`, eval, with, `this`, referenced named function-expression identifiers, `new.target`, rest params, and spread argument lists.

I also added fixture/exec coverage for `this.constructor.arguments`, `F.arguments`, `new.target.arguments`, and escape paths like `leak(this)` / `leak(F)`.

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
```diff
@@ -500,6 +500,95 @@ impl Optimizer<'_> {
...
+            return true;
+        }
+
+        if let Some(ident) = ident {
+            if contains_ident_ref(&function.body, ident) {
+                return true;
+            }
+        }
+
+        contains_new_target(body)
```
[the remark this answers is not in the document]
skku970412: Thanks, also fixed in b9599903. Default parameter initializers now participate in the same observability checks before any trailing argument is removed.

Specifically, `function.params` is checked for `this`, referenced named function-expression identifiers, and `new.target`, so cases such as `new function(a = console.log(new.target.arguments.length)){}(undefined, "x")`, `this.constructor.arguments`, and `F.arguments` in parameter defaults are preserved. Added fixture and exec coverage for these cases.

skku970412:
@codex review

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
```diff
@@ -500,6 +500,95 @@ impl Optimizer<'_> {
...
+        ident: Option<&Ident>,
+        function: &Function,
+    ) -> bool {
+        let Some(body) = function.body.as_ref() else {
+            return true;
+        };
+
+        if let Some(scope) = self.data.get_scope(function.ctxt) {
+            if scope.intersects(
+                ScopeData::USED_ARGUMENTS
```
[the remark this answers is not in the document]
skku970412: Thanks, good catch. I added an explicit computed-class-key observability guard instead of relying only on `ScopeData::USED_ARGUMENTS`.

The new-function argument cleanup now bails out when class computed keys or class heritage expressions evaluated in the constructor context may observe the invocation argument list through `arguments`, `this`, a referenced named function expression, `new.target`, or direct eval.

I also added fixture and exec coverage for the exact computed-key case, named function variants, `new.target`, default parameter class expressions, and class heritage expressions.

skku970412:
@codex review

skku970412:
Thanks for the review. I appreciate the careful checks on the constructor argument observability cases.

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
```diff
@@ -1737,3 +1832,146 @@ impl Optimizer<'_> {
         None
     }
 }
+
+struct ContainsNewTarget {
```
kdy1: Can you remove all new visitors?

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
```diff
@@ -1737,3 +1832,146 @@ impl Optimizer<'_> {
...
+fn contains_new_target<N>(n: &N) -> bool
+where
+    N: VisitWith<ContainsNewTarget>,
+{
+    let mut v = ContainsNewTarget { found: false };
+    n.visit_with(&mut v);
+    v.found
+}
+
+struct ConstructorArgObserver<'a> {
```
kdy1: This one, too

kdy1 (changes requested):
I think this optimization is not big enough to justify introduction of new visitors

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
skku970412: Thanks, that makes sense. I removed the new visitor-based observability checks and narrowed the optimization instead.

The PR now keeps only a small known-safe NewExpr fast path: direct anonymous, parameterless function constructors with a trivially safe body. More complex constructor bodies are skipped instead of being analyzed.

This preserves the reported case while avoiding new visitors for constructor argument observability.

crates/swc_ecma_minifier/src/compress/optimize/iife.rs
skku970412: Removed this visitor as well in e4e8879. The cleanup now relies on the narrowed fast path instead of recursive observability visitors.

skku970412:
@codex review

skku970412:
@kdy1 Thanks for the follow-up test coverage. The visitor-based implementation has been removed, the optimization is now limited to the known-safe fast path, and all checks are passing. Could you please take another look when you have time?

kdy1:
Sorry. It seems like https://github.com/swc-project/swc/pull/12017 (merged today) supersedes this PR. I'll close this PR

PR #12014, opened 2026-07-11
refactor(es/parser): remove legacy lexer crate

kdy1:
**Description:**

Remove the unmaintained `swc_ecma_lexer` compatibility crate now that `swc_ecma_parser` contains the maintained lexer and parser implementation.

- Delete the `swc_ecma_lexer` crate and its lockfile entry.
- Remove the legacy token compatibility API from `swc_ecma_parser`.
- Remove the parser's now-unused `phf` dependency.
- Update the unstable lexer documentation and the remaining JSX source reference.
- Add a changeset for the affected published crates.

**BREAKING CHANGE:**

The `swc_ecma_lexer` crate and the legacy `swc_ecma_parser::token` API are removed. Consumers that need direct lexer token access should use `swc_ecma_parser` with its `unstable` feature.

**Validation:**

- `cargo metadata --no-deps`
- `cargo test -p swc_ecma_transforms_react`
- `cargo fmt --all`

N/A


kdy1 (commented):
I think it's now good to go, as the main user of this crate was Deno

kdy1 (commented):
I was wrong - Deno is still using SWC, so I'll postpone this PR

kdy1:
Status update:

Deno is still actively using `swc_ecma_lexer` and the legacy `swc_ecma_parser::token` API through `deno_ast` and `dprint-swc-ext`.

Merging this PR would not immediately break Deno because it pins existing published SWC versions. However, it would prevent Deno from upgrading to newer SWC versions until those dependencies are migrated to `swc_ecma_parser`’s unstable lexer API.

That migration is not entirely mechanical yet:

- `dprint-swc-ext` relies on traits such as `Spanned` and `PartialEq` for `TokenAndSpan`, which the unstable type does not currently implement.
- Deno uses token payloads, including identifiers and lexer errors, while the unstable `TokenAndSpan` only contains the token kind and span.
- There is no active migration PR in either `deno_ast` or `dprint-swc-ext`.
- The open `deno_ast` OXC port is experimental, unreviewed, and currently conflicting, so it does not provide a reliable timeline.

Before moving this PR forward, we should either:

1. Improve the unstable lexer API and migrate `dprint-swc-ext` and `deno_ast`, or
2. Explicitly accept that Deno will remain pinned to the final published legacy lexer version.

The PR also needs to be rebased, and the remaining fixture lockfiles need to be regenerated.

PR #12020, opened 2026-07-13
fix(es): preserve PURE annotation ownership

### LINKED ISSUE #12019: SWC removes parentheses around PURE-annotated receiver, changing annotation range
### Describe the bug

SWC re-emits a parenthesized PURE-annotated receiver expression without the parentheses:

```js
(/* @__PURE__ */ new Set(x)).forEach(cb);
```

becomes:

```js
/* @__PURE__ */ new Set(x).forEach(cb);
```

These two forms are not equivalent for tools that understand `/* @__PURE__ */`.

In the input, the annotation applies only to the `new Set(x)` receiver expression. In the output, the annotation appears to apply to the whole chained call `new Set(x).forEach(cb)`.

That can cause a later minify/DCE pass to remove the entire `.forEach(cb)` call, including side effects inside `cb`.

This is especially risky when a previous build step emits code like:

```js
(/* @__PURE__ */ new Set([...Object.keys(currentProps), ...Object.keys(targetProps)])).forEach((key) => {
  changedProps[key] = targetProps[key];
});
```

After SWC re-emits it without preserving the parentheses, a downstream minifier can interpret the PURE annotation as applying to the entire chained call and drop the callback mutation.

### Input code

```typescript
(/* @__PURE__ */ new Set(x)).forEach(cb);


Slightly more concrete example:


const changed = {};

(/* @__PURE__ */ new Set(["a"])).forEach((key) => {
  changed[key] = 1;
});

console.log(changed);
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript"
    },
    "target": "es2022"
  },
  "minify": false
}
```

https://github.com/kimjong95/rolldown-pure-set-foreach-repro

### SWC Info output

```txt
Operating System:
    Platform: darwin
    Arch: arm64
    Machine Type: arm64
    Version: Darwin Kernel Version 25.5.0: Mon Apr 27 20:41:15 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6041
    CPU: (12 cores)
        Models: Apple M4 Pro

Binaries:
    Node: 25.8.0
    npm: 11.11.0
    Yarn: 1.22.22
    pnpm: 11.0.7

Relevant Packages:
    @swc/core: 1.15.43
    @swc/helpers: N/A
    @swc/types: 0.1.27
    next: 15.5.20

SWC Config:
    output: N/A
    .swcrc path: N/A

Next.js info:
    output: N/A
```

### Expected behavior

SWC should preserve the effective range of the `/* @__PURE__ */` annotation.

Expected output:

```js
(/* @__PURE__ */ new Set(x)).forEach(cb);
```

or another output form that keeps the annotation attached only to the `new Set(x)` receiver expression, not to the whole `new Set(x).forEach(cb)` call.

### Actual behavior

With `@swc/core@1.15.43`, this input:

```js
(/* @__PURE__ */ new Set(x)).forEach(cb);
```

is emitted as:

```js
/* @__PURE__ */ new Set(x).forEach(cb);
```

Local reproduction:

```js

const input = `(/* @__PURE__ */ new Set(x)).forEach(cb);`;

swc.transform(input, {
  jsc: {
    parser: { syntax: "ecmascript" },
    target: "es2022",
  },
  minify: false,
}).then((out) => {
  console.log(out.code);
});
```

Output:

```js
/* @__PURE__ */ new Set(x).forEach(cb);
```

This changes how PURE-aware tools can interpret the annotation.

### Version

@swc/core@1.15.43

### Additional context

This was found while investigating a downstream production regression where a library build emitted:

```js
(/* @__PURE__ */ new Set([...Object.keys(currentProps), ...Object.keys(targetProps)])).forEach((key) => {
  if (key === "text") return;
  if (!isEqual(currentProps[key], targetProps[key])) {
    changedProps[key] = targetProps[key];
  }
});
```

After SWC re-emitted the expression as:

```js
/* @__PURE__ */ new Set([...Object.keys(currentProps), ...Object.keys(targetProps)]).forEach((key) => {
  if (key === "text") return;
  if (!isEqual(currentProps[key], targetProps[key])) {
    changedProps[key] = targetProps[key];
  }
});
```

a later minify/DCE pass could treat the whole chained call as pure and remove it, reducing the restore result to `{}`.

In our reproduction, three equivalent restore implementations behave differently after this PURE annotation range changes:

| Case | Result |
|---|---|
| `Set + forEach`, after PURE range is widened | callback side effect can be dropped, result becomes `{}` |
| `Set + for...of` | preserved |
| `Array + forEach` | preserved |

Related Rolldown issue:

https://github.com/rolldown/rolldown/issues/10247

The Rolldown maintainer also noted there that the parenthesis should not be removed in this case because it changes the `/* @__PURE__ */` range.

skku970412:
**Description:**

Parentheses can delimit the effective range of a PURE annotation around a receiver. Removing them could widen the annotation to a following member/call chain and allow a later DCE pass to remove callback side effects.

This PR:

- preserves the boundary when a PURE-annotated call, `new`, tagged template, or optional call is the head of an expression continuation;
- retains all nested-parenthesis provenance so comments on intermediate parentheses move to the exact annotation owner in source order;
- records and restores annotation ownership across minifier rewrites with a dedicated mark, including compress-disabled and mangle-only paths;
- moves comments from erased TypeScript expression wrappers to their runtime expressions while leaving ordinary and non-owner parenthesis folding unchanged.

Regression coverage includes compiler, fixer, minifier, TypeScript, React, optional-chain, nested/optimizer, negative non-owner, mangle-only, and emit/reparse/recompress paths.

Validation:

- `cargo test -p swc_ecma_transforms_base --lib`
- `cargo test -p swc_ecma_transforms_base --test fixture`
- `cargo test -p swc_common -p swc_ecma_minifier --lib`
- `cargo test -p swc_ecma_transforms_typescript --test strip comments_on_erased_expression_wrappers -- --nocapture`
- `cargo test -p swc --test projects 12019`
- `cargo test -p swc_ecma_minifier --test compress 12019`
- `cargo test -p swc_ecma_minifier --test exec issue_12019_pure_annotation_boundary_survives_ -- --nocapture`

Closes #12019


kdy1:
Sorry for closing, but I feel like this PR is a real AI slop. I'm open to AI PRs, but that doesn't mean I'm fine with any AI PRs.
It changes too many contracts without any prior discussion, so I'll close this PR

PR #12026, opened 2026-07-16
fix(es/minifier): Measure number length precisely

### LINKED ISSUE #12008: es/minifier: constant evaluation doesn't always select the shortest result
When evaluating constant calculations, the evaluated result isn't always the shortest.

Input:
```typescript
console.log(0.1 + (0.2 - 0.1) * 0.5);
```

Output:
```typescript
console.log(.15000000000000002);
```

If I manually remove the spaces, then the minification is better:

Input:
```typescript
console.log(0.1+(0.2-0.1)*0.5);
```

Output:
```typescript
console.log(.1+.05);
```

Technically `.15000000000000002` is the correct value, but I think for most use cases, it could be replaced with `.15` (maybe a new option could be added to control this?).
In any case, the expression evaluation should select the shortest result:

`.1+(.2-.1)*.5`
`.1+.1*.5`
`.1+.05` ← shortest
`.15000000000000002`

---

A similar issue is that some expressions are not evaluated, even if the final result would actually be shorter:

Input:
```typescript
console.log(0.7 + (0.9 - 0.7) * 0.5);
```

Output:
```typescript
console.log(.7+(.9-.7)*.5);
```

But the minified result could be `console.log(.8);`.

Note that during the evaluation, the second expression is longer than the first, but fully evaluating it actually produces shorter code:
`.7+(.9-.7)*.5`
`.7+.20000000000000007*.5`
`.7+.10000000000000003`
`.8` ← shortest

---

Config:
```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript"
    },
    "minify": {
      "compress": true
    }
  },
  "minify": true
}
```
(Just need to enable `minify` and `compress`)

[Playground link](https://play.swc.rs/?version=1.15.43&code=H4sIAAAAAAAAA0vOzyvOz0nVy8lP1zDQM1TQVgBSRgq6CkCOpoIWkDLVtOblSkZVpg1SpAtSooVVgTnEHEuwOeZwcwBVyTeobQAAAA%3D%3D&config=H4sIAAAAAAAAA32US3LbMAyG9zmFR%2Bts20UPkF3PwKFJUKbDh4YAHWsyvnshSrLdGNJOwocfIAEQ32%2BHQ3dG0%2F05fPMn%2Fwy6IJT7P1twTKSvbOnARI2m%2BIG695WecUJOB4Rmus2kI116oKbCX4t7F3JGWN0XW%2FTJu%2FE5oclxKID4ZGMrx6sREuH%2F%2BoWV%2FDUBKvXZfsw5gE47RGlUPhH0UKTAJoegBwR10UWIMp1UF49ZSjHBSmDVUPIg8mQ9%2BZw45yu1oK0y2YKAfAFD%2FgKSjHOxLCFfT7hPwxaOte9bk3%2Bo4aJD1STkhGtrCZ9WiHrKHkm5mqQSznCjBjNcivtT6Z0qQLWkV905%2B7TRk08ArkDQiElHkOI2D8fztKV2u0qfHI8sjQLn%2BZZumaDnoirvnVDZqTJQyEvdLGCrgamyRjrOgjfKh96CAud4VoTQ%2BOXJnKSkNA6QnQC4v9pJUzUDdX%2BFG3x6EDv4g29J8oAtHlHTaZviGI857CSIQKdsdxy4FZS3ceEtcR22eU0WeDTAii4VG3hdAvwAKKvQluXLbPDz4IiqD%2Fn4WBOLw%2B2%2BhKNO%2FeO9z3v4bXHoYra1wWXDT%2F2d9%2FLv7uG0buH1BJ3Hv6uw5bz9A1EN%2Bh8sBgAA)

SWC version: `1.15.43`

Austaras:
**Description:**

**BREAKING CHANGE:**

Closes #12008


crates/swc_ecma_utils/src/number.rs
```diff
@@ -188,6 +193,78 @@ impl std::ops::Not for JsNumber {
...
+            + 3;
+
+        num.truncate(num.len() - cnt);
+        num.push('e');
+        num.push_str(&cnt.to_string());
+    }
+
+    num
+}
+
```
Austaras, line 262: Surely the performance of this function could be improved.

kdy1 (approved):
Thanks!

kdy1 (approved):
I verified the deps to prevent cycle

---

https://github.com/Austaras/swc/blob/ef8fc8c963867a81cef761eff5bd5d5459129859/crates/swc_ecma_utils/Cargo.toml#L44

 - `swc_ecma_utils` depends on the parser (dev)

PR #12029, opened 2026-07-16
fix(es/minifier): Preserve top-level declarations referenced only by direct eval

### LINKED ISSUE #12028: Minifier drops top-level declarations referenced only by direct `eval`
### Describe the bug

The minifier's DCE removes top-level declarations that are referenced only through a direct `eval`, corrupting code that relies on them at runtime — the `eval` then observes a missing binding and throws `ReferenceError`.

Direct-`eval` protection currently applies only within function scopes (added in #4693, closing #2728). `found_direct_eval` is propagated up to the root scope, but the module/script root never consumes it, so top-level bindings reachable by a direct `eval` are dropped.

### Input code

```js
var x = 1;
eval("x");
```

Also reproduces with a cyclic class pair reachable only via `eval`, and with a top-level binding referenced by an `eval` nested inside a function (the flag propagates from the function scope to the root):

```js
class A { m() { return B; } }
class B extends A { m() { return A; } }
eval("A");

var x = 1;
function f(s) { return eval(s); }
f("x");
```

### Config

```json
{
  "jsc": { "minify": { "compress": { "unused": true, "toplevel": true }, "mangle": false } },
  "minify": true
}
```

Equivalently, any minification with top-level DCE enabled (e.g. module output). Playground: https://play.swc.rs/ (paste the input above with top-level DCE on).

### Expected behavior

`x` (and the classes) are preserved and the `eval` resolves them at runtime — matching how a direct `eval` **inside a function** is already handled.

### Actual behavior

The top-level declarations are removed and the `eval` throws `ReferenceError`.

### Version

Reproduces on current `main`, and on released versions since #4693 (which fixed function scopes but not the root scope).

baltasarblanco:
## Description

The DCE tree-shaker protects declarations reachable by a direct `eval` only within function scopes (added in #4693). `found_direct_eval` is propagated to the root scope, but `visit_mut_module` / `visit_mut_script` never consumed it, so top-level declarations referenced only through a direct `eval` were dropped — the `eval` then observes a missing binding at runtime.

This consumes `found_direct_eval` at the root: when a direct `eval` reaches the root scope, the root-scope declarations are pinned so DCE cannot remove them. "Root-scope" means the module/script's own bindings — immediate top-level declarations (var/fn/class + imports) plus `var`s hoisted out of nested top-level statements — collected **without** descending into nested function scopes, so a top-level `eval` doesn't pin bindings it can't resolve. Each is pinned both as a graph entry (so `subtract_cycles` never subtracts it across Repeat passes — this covers reachable class cycles) and marked used (this covers lone bindings that have no in-graph references).

Localized to the two root entry points; the added work only runs when a direct `eval` reaches the root scope, so it should be performance-neutral.

Regression tests added to `simplify_dce` (each fails without the fix): a top-level `var`, a top-level class cycle, a top-level binding referenced by a nested `eval`, and a `var` hoisted out of a top-level block.

Scope note: an `eval` inside an unreachable (dead) nested function still pins the enclosing top-level declarations, matching the existing function-scope behavior since #4693, which likewise doesn't gate on reachability. This only ever over-preserves, which is safe.

## Related issue

Fixes #12028


baltasarblanco:
Rewrote the description to match the final implementation. All the review feedback is addressed as of 6c8d4f9:

- **Scope of collection:** root-scope declarations only, without descending into function bodies.
- **Cycle re-subtraction:** eval-preserved ids are pinned as graph `entries`, so `subtract_cycles` doesn't drop them across Repeat passes.
- **Var hoisting:** the root now also collects `var`s hoisted out of nested top-level statements.
- **Unreachable `eval`:** kept consistent with the existing function-scope behavior (over-preserves only, which is safe).

kdy1 (approved):
I _think_ this PR is good to go, mainly because this is a _fix_ rather than a rule addition. 
Thank you!

PR #12031, opened 2026-07-17
refactor(es/minifier): Remove ProgramData.top

Austaras:
**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/optimize/unused.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/unused.rs
@@ -707,7 +707,8 @@ impl Optimizer<'_> {
 
         if self
             .data
-            .top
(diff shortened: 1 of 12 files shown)


crates/swc_ecma_minifier/src/compress/optimize/unused.rs
```diff
@@ -707,7 +707,8 @@ impl Optimizer<'_> {
 
         if self
             .data
-            .top
+            .get_scope(self.ctx.var_scope)
```
Austaras, line 710: It's possible to make this more fine-grained by checking the ident is block scoped or function scoped. However this doesn't seem very useful.

kdy1 (approved):
Nice work! Thanks!

PR #12037, opened 2026-07-19
fix(es/minifier): preserve argument aliases in nested scopes

### LINKED ISSUE #12032: Unused ignore arguments usage in outer scope
### Describe the bug

A long-standing issue(that nobody cares) revealed by Codex.

### Input code

```typescript
function run(f) {
    f()
}

function foo(b) {
    run(() => {
        b = 2
    })

    console.log(arguments[0])
}

foo(1)
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2015",
    "loose": false,
    "minify": {
      "compress": {
        "defaults": true
      }
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

https://play.swc.rs/?version=1.15.43&code=H4sIAAAAAAAAA0srzUsuyczPUygqzdNI01So5lIAgjQNTa5aLq40mGRafr5GEkwSpFJDU8HWDsoHgSQFWwUjMK9WkwtMJ%2BfnFefnpOrl5KdrJBall%2Bam5pUURxvEQgwGmmeoCQCs42gffAAAAA%3D%3D&config=H4sIAAAAAAAAA02PQQrDIBBF9zmFzLqLttAueoceQsxYLBrFGaEScvdqokk24rzvn%2F%2BdByHgSwpeYi7XMgQZCeM%2BF0J5YvkrBDgHJBVNYLh0lalKWlrCFS2bAizjB7m6kO7X26M5wHpP2B2NOTMZnc%2BZyrsQkejECh1Ry2S5Uo4Jm7AM%2FVyzwfkxWTx%2BVFtvPZ5wPOqRew8w9O7OdfvyB8dntFUdAQAA

### SWC Info output

_No response_

### Expected behavior

log 2

### Actual behavior

log 1

### Version

latest

### Additional context

_No response_

cyphercodes:
**Description:**

Assignments to a function parameter can occur inside a nested closure. The unused-assignment optimizer was checking `arguments` usage in that nested scope, which allowed it to remove the assignment even when the parameter is aliased by the outer non-strict function's `arguments` object.

This changes the lookup to use the parameter's declaring syntax context and adds a regression fixture for the nested-assignment case.

Verification:
- `cargo test -p swc_ecma_minifier --test compress 12032 -- --nocapture`
- `cargo test -p swc_ecma_minifier` (2,128 passed, 1 ignored)
- `cargo clippy -p swc_ecma_minifier --all-targets -- -D warnings`
- Node execution of both fixture input and optimized output (`2` in both cases)

Fixes #12032

**BREAKING CHANGE:**

None.

#12032

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/optimize/unused.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/unused.rs
@@ -787,7 +787,9 @@ impl Optimizer<'_> {
                 ) && var.usage_count == 0
                     && var.flags.contains(VarUsageInfoFlags::DECLARED)
                     && (!var.flags.contains(VarUsageInfoFlags::DECLARED_AS_FN_PARAM)
-                        || !self.data.used_arguments(self.ctx.scope)
+                        // The assignment can be inside a nested function, so query the
+                        // parameter's declaring scope instead of the current scope.
+                        || !self.data.used_arguments(i.id.ctxt)
                         || self.ctx.expr_ctx.in_strict)
                 {
                     report_change!(
--- a/crates/swc_ecma_minifier/tests/fixture/issues/12032/config.json
+++ b/crates/swc_ecma_minifier/tests/fixture/issues/12032/config.json
@@ -1,0 +1,5 @@
+{
+    "defaults": true,
+    "toplevel": false,
+    "passes": 0
+}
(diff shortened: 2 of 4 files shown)


Austaras:
This would not be enough. In inline pass swc would change ident ctxt(to avoid name collision) which would result in non exist scope data.

PR #12041, opened 2026-07-20
perf(es/renamer): Skip identity rename mappings on apply

hamidrezahanafi:
**Description:**

Hygiene often builds a rename map that is almost entirely identity entries (`foo` → `foo`). The apply Operator already no-ops those lookups, but still walks the full AST.

This change:
- Keeps identity entries in the map **during** analysis so later scopes can still skip via `to.get(id)` (required when hygiene runs without resolver; shared Ids would otherwise blow up the reverse map).
- Strips identity entries with `map.retain` **before** apply when `keep_class_names` is off.
- Leaves identity entries in place when `keep_class_names` is on, so Operator can still rewrite `class Foo` into `let Foo = class Foo` (existing behavior relied on a non-empty identity map). Minifier `keep_classnames` preserves class Ids so they never enter the map and still skip apply.

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_transforms_base/src/rename/mod.rs
+++ b/crates/swc_ecma_transforms_base/src/rename/mod.rs
@@ -318,7 +318,16 @@ where
         // later scopes can skip via `to.get(id)`. Without that (e.g. hygiene
         // alone, no resolver), the same Id is reprocessed in every scope and
         // the reverse map blows up.
-        map.retain(|id, v| v.atom() != &id.0);
+        //
+        // Keep identities when `keep_class_names` is set: Operator uses that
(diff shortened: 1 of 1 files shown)


kdy1 (commented):
I like the idea. Can you update the test fixtures to make CI pass?

hamidrezahanafi:
> I like the idea. Can you update the test fixtures to make CI pass?

Updated, It should pass all the tests now!

Austaras (approved):
I like this idea too.

kdy1 (commented):
Thank you for the clever trick! Nice optimizaiton!

PR #12042, opened 2026-07-20
refactor(es/ast): Remove Hash from ECMA AST nodes

magic-akari:
**Description:**

Remove `Hash` implementations from ECMA AST nodes and adapt downstream users.

Maintaining a `Hash` contract constrains how special `f64` values are represented. Removing it unblocks representing `NaN` and `Infinity` as `Lit::Num` again.

**BREAKING CHANGE:**

ECMA AST nodes and dependent public types no longer implement `Hash`.

### DIFF
--- a/.changeset/remove-ecma-ast-hash.md
+++ b/.changeset/remove-ecma-ast-hash.md
@@ -1,0 +1,10 @@
+---
+swc_bundler: major
+swc_core: major
+swc_ecma_ast: major
+swc_ecma_lints: major
+swc_ecma_minifier: major
+swc_ecma_parser: major
+---
+
+refactor(es/ast): Remove `Hash` implementations from ECMA AST nodes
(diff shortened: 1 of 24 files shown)


Austaras:
This makes no sense. Since we impl Eq for Number(which f64 does not), we could impl Hash for Number too. [ordered-float](https://docs.rs/ordered-float/latest/src/ordered_float/lib.rs.html#2186) could be taken as a reference.

magic-akari:
After reconsidering this, I think we should retain `Hash` for now to avoid breaking the existing public API. I’ll use `ordered-float` as a reference to make `Number`’s `Eq` and `Hash` implementations consistent.

That said, we do not need to treat `Hash` on AST nodes as a long-term API guarantee. We may choose not to implement it in SWC v2.

I’m closing this PR in favor of a more focused change.

Austaras:
Not impl Hash provides no benefit at the moment. But I agree that may change.

PR #12043, opened 2026-07-20
fix(es/ast): support hashing non-finite numbers

magic-akari:
**Description:**

Canonicalize non-finite numeric values in `Lit::Num` while preserving the existing ECMA AST `Hash` API.

SWC currently represents `Infinity` and `NaN` in several forms. `eval_global_vars` rewrites `Infinity` to `1 / 0`, while `eval_numbers` can rewrite division by zero back to an `Infinity` identifier. Other constant-evaluation paths produce `Lit::Num(INFINITY)`, which codegen prints as `Infinity`. This churn makes optimization passes repeatedly recognize equivalent values in different AST forms.

Follow esbuild’s approach of using semantic constants during optimization. Allow `Lit::Num` to store valid JavaScript non-finite values as the canonical internal representation, and leave syntax selection to codegen. Codegen can emit readable `Infinity` and `NaN` when safe or shorter `1 / 0` and `0 / 0` when minifying.

**BREAKING CHANGE:**

`Number` now treats all NaN values as equal and distinguishes `-0.0` from `+0.0`.

- **#12043** (current)
  - #12047
    - #12050
      - #12048
        - #12051
      - #12052
        - #12053
      - #12054
      - #12055
      - #12056
      - #12057
    - #12058
    - #12049
      - #12059
        - #12060

### DIFF
--- a/.changeset/fix-number-hash.md
+++ b/.changeset/fix-number-hash.md
@@ -1,0 +1,7 @@
+---
+ast_node: minor
+swc_core: major
+swc_ecma_ast: major
+---
+
+fix(es/ast): Support NaN and infinity in Number while preserving Hash.
--- a/crates/ast_node/src/ast_node_macro.rs
+++ b/crates/ast_node/src/ast_node_macro.rs
@@ -7,11 +7,29 @@ use syn::{
 #[derive(Clone)]
 pub struct Args {
     pub ty: Literal,
+    pub no_partial_eq: bool,
 }
 
 impl Parse for Args {
     fn parse(i: ParseStream<'_>) -> syn::Result<Self> {
-        Ok(Args { ty: i.parse()? })
+        let ty = i.parse()?;
+        let mut no_partial_eq = false;
+
+        if i.parse::<Option<Token![,]>>()?.is_some() {
+            let option: Ident = i.parse()?;
+
+            if option == "no_partial_eq" {
+                no_partial_eq = true;
+            } else {
+                return Err(Error::new(option.span(), "unknown ast_node option"));
+            }
+        }
+
+        if !i.is_empty() {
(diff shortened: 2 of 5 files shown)


kdy1:
Why do we need this?

magic-akari:
> Why do we need this?

My motivation comes from a technique used by esbuild. In [internal/config/globals.go](https://github.com/evanw/esbuild/blob/6ff1d8b0d8c134e867a397eef39702a223ebef9e/internal/config/globals.go#L1013-L1021), its define phase replaces the global `undefined`, `NaN`, and `Infinity` identifiers with semantic constants. Later transform and minification passes operate on these internal values instead of preserving their identifier-based AST representation.

I traced SWC’s current compressor and found that we already mix several representations. `eval_global_vars` converts `Infinity` to `1 / 0`, while `eval_numbers` can convert another `x / 0` back to an `Infinity` identifier and even has explicit branches saying “Prefer `1/0` over Infinity” and “Prefer `0/0` over NaN.” Other constant-evaluation paths produce `Lit::Num(INFINITY)`, which codegen prints as `Infinity`. This is not an infinite loop within one compression pass, but it creates real representation churn across optimization, codegen, and reparsing. The same value must also be recognized repeatedly in different AST forms.

Using `Lit::Num` as the internal working representation gives these transformations a canonical endpoint. Codegen can then choose the appropriate syntax at the output boundary—for example, readable `Infinity`/`NaN` in normal output when safe, or shorter `1/0`/`0/0` when minifying.

This PR removes the artificial restriction that prevents `Lit::Num` from representing these valid JavaScript numeric values, while preserving the existing `Hash` API.

kdy1 (approved):
Thank you for the explanation! I'll merge this PR after adding the context to the PR body (and the commit message)

PR #12047, opened 2026-07-20
fix(es/codegen): emit non-finite numeric literals

magic-akari:
**Description:**

Emit non-finite `Lit::Num` values safely in every expression context. Readable output preserves an existing `NaN` or `Infinity` raw value, while generated and minified values use capture-free numeric expressions. Hygiene prevents preserved identifier forms from being shadowed.

- #12043
  - **#12047** (current)
    - #12050
      - #12048
        - #12051
      - #12052
        - #12053
      - #12054
      - #12055
      - #12056
      - #12057
    - #12058
    - #12049
      - #12059
        - #12060


kdy1 (approved):
👍

PR #12053, opened 2026-07-21
fix(es/minifier): bound arguments parameter injection

magic-akari:
**Description:**

Replacing an `arguments` access with a named parameter can require injecting every missing parameter up to that index. A large index can therefore generate an unbounded number of parameters and substantially increase output size.

This PR follows Terser's fixed bound: an access is replaced only when it reuses an existing parameter or requires injecting at most five additional parameters. Larger gaps remain as `arguments` accesses, with fixture coverage for the five/six boundary.

**Related issue (if exists):**

- Close: #12108

---

- #12043
  - #12047
    - #12050
      - #12048
        - #12051
      - #12052
        - **#12053** (current)
      - #12054
      - #12055
      - #12056
      - #12057
    - #12058
    - #12049
      - #12059
        - #12060

PR #12054, opened 2026-07-21
fix(es/minifier): index strings by UTF-16 code unit

magic-akari:
**Description:**

JavaScript string property indexing uses UTF-16 code units. Existing constant folding checked string-literal indices against WTF-8 byte length, which can treat out-of-bounds indices as valid for non-ASCII BMP characters, astral characters, and lone surrogates. In the expression simplifier, that mismatch could also reach an `unreachable!` branch.

This PR validates and reads string-literal indices using UTF-16 code units in both the minifier and expression simplifier. It preserves the WTF-8 byte-length check as a fast upper-bound rejection, folds valid indices to the correct code unit, and handles UTF-16 out-of-bounds indices according to each pass's existing semantics.

Coverage includes BMP non-ASCII characters, astral surrogate pairs, and lone surrogates. The change was verified with:

- `cargo test -p swc_ecma_minifier -p swc_ecma_transforms_optimization`
- `cargo fmt --all`

- #12043
  - #12047
    - #12050
      - #12048
        - #12051
      - #12052
        - #12053
      - **#12054** (current)
      - #12055
      - #12056
      - #12057
    - #12058
    - #12049
      - #12059
        - #12060


kdy1 (dismissed):

kdy1 (dismissed):

magic-akari:
This PR unintentionally exposes some pre-existing bugs on `main` by broadening the constant-folding scope. I’ll revert that part of the change and keep this PR focused on UTF-16 string indexing. The underlying coercion issues can be fixed separately.

kdy1 (approved):

PR #12057, opened 2026-07-21
fix(es/minifier): avoid inexact number radix folding

magic-akari:
**Description:**

`Number#toString(radix)` folding currently converts integer values through `u128`. Unsafe integers can already have rounded representations, so this conversion can produce digits that do not match JavaScript for non-power-of-two radices.

This PR restricts radix folding to safe integers, with a separate exact path for supported power-of-two radices below the `u128` bound. Values outside those limits remain for runtime evaluation.

- #12043
  - #12047
    - #12050
      - #12048
        - #12051
      - #12052
        - #12053
      - #12054
      - #12055
      - #12056
      - **#12057** (current)
    - #12058
    - #12049
      - #12059
        - #12060


kdy1 (approved):
👍

PR #12060, opened 2026-07-21
refactor(es/typescript): track enum evaluation context

magic-akari:
**Description:**

The existing enum evaluator keeps only a per-declaration map of previously evaluated members. That state is too narrow to match TypeScript behavior for duplicate or merged declarations, forward and self references, external constants, and references that shadow global names.

This PR introduces source-order-aware enum evaluation state in both the transform and fast-DTS paths. It records member names, bindings, external constants, and reference provenance across declarations, while the transform-side name collection uses the same stack-safe expression traversal as the main analyzer.

- #12043
  - #12047
    - #12050
      - #12048
        - #12051
      - #12052
        - #12053
      - #12054
      - #12055
      - #12056
      - #12057
    - #12058
    - #12049
      - #12059
        - **#12060** (current)

### DIFF
--- a/.changeset/fix-enum-template-cooked.md
+++ b/.changeset/fix-enum-template-cooked.md
@@ -1,0 +1,7 @@
+---
+swc_core: patch
+swc_ecma_transforms_typescript: patch
+swc_typescript: patch
+---
+
+fix(es/typescript): Evaluate enum templates from cooked values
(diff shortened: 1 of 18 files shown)


magic-akari:
Closing this draft. The enum evaluation inconsistencies are real, but this approach is too broad and introduces a reported performance regression. It is not a good tradeoff in its current form.

PR #12068, opened 2026-07-22
feat(es/minifier): expand support for trivial spreads

lukesandberg:
**Description:**

Support trivial object spreads of pure expression of primitive type

`{...null}` was already supported, this extends that to support `{...undefined}` `{...2}` etc.

[Motivating Example](https://play.swc.rs/?version=1.15.46&code=H4sIAAAAAAAAA0srzUsuyczPU0jLz9fQVKjmUlAoSi0pLcpTMNQ25KrlSs7PK87PSdXLyU%2FXqE60MtRR0NPT0yjLz0yB6NCs1eQCAHICWctEAAAA&config=H4sIAAAAAAAAA41VwW7bMAy99ysCn3focir2AbvtGwTFohxlsmiIlJugyL%2BPVuw0bWhjlyDm4yNF8lH6eNntmhO1za%2Fdh%2FyVj8Fmgnz%2FFgtdEtuzWBpoe0ttDgM3Pxb0RBPkbSSopusNadjmDriyaP%2B638%2BMJiISLIzZ1ocU%2FOUxZ4v9kIHowSZWCVl6SExf%2BTOW8X0COJdH%2BwExgk0biLFkQmLoIGuBW4zRDgRmtFmJMp3U5kCopZjAwuDMkHFQ8eQCB0yS8xl1YJ1p0YEChQwthxE0muQSWiIpT6mnwg4OpevqnL%2BxYbSxWFZywrmORE6rRD1iIDa%2BJK2FN3ClBzdwbu53ZvAmA5ecnnknDGllJn8BpAPREiXbgxa3enjR0xrbbzJD8iJZvii46FurMkEnTTUheKWzU2cgc9CmmcGVFqbOtuvoSvcoODDgvUhF4dJ74Pao5eTLAOgVQMZrvSaqG2DuS7iCT%2FuwAf%2BWIlnX1%2BzRWz6uo3TpDxg3EvTAR3QbDjIJxnU4yyVxHtbxkhyIMsCpLoUq8HwHiP4ZTazX5ZM0ZDskoukiHj5vidnher%2BGe5u6uu4P1yXjEGGEuKbi%2F9iQTVTONk6iXhb7aXnhTROllb0LP1%2B%2FvBpSysvyW4tqenTloaCqyek1EQn1mE7UfHour8fSuibQn4Vdm3X9B%2FL9yjXnBgAA)

Turbopack is adding support for tree shaking CJS exports, it would be convenient to rewrite

`module.exports = {unused: foo() }` as `module.exports = {...(void foo())}` when we determine that the property was unused.  This could then get appropriately folded by SWC.

This pattern presumably comes up in normal code too

`someFunction({prop1, ...(process.env.NODE_ENV == "development" ? {extra debugging props} : undefined)})`

and similar patterns involving null safe access `someFunction({...do?.you?.exist()})` if SWC can fold the condition it should also be able to drop the spread.

**BREAKING CHANGE:**

None.

### DIFF
--- a/.changeset/drop-trivial-object-spread.md
+++ b/.changeset/drop-trivial-object-spread.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_minifier: patch
+---
+
+feat(es/minifier): drop trivial object spread of primitives
--- a/crates/swc_ecma_minifier/src/compress/pure/misc.rs
+++ b/crates/swc_ecma_minifier/src/compress/pure/misc.rs
@@ -292,13 +292,32 @@ impl Pure<'_> {
             }
         }
 
+        // Spreading a primitive of these types copies zero own enumerable
+        // properties (`ToObject` on them yields a wrapper whose properties are
+        // all non-enumerable), so `{ ...x }` where `x` is one of them
+        // contributes nothing.
+        //
+        // Strings are intentionally EXCLUDED: they expose indexed own
+        // enumerable properties (`{ ..."ab" }` === `{ 0: "a", 1: "b" }`).
+        // Objects, arrays, functions, etc. are also excluded as they carry
+        // real properties.
+        fn spreads_no_props(expr: &Expr, expr_ctx: ExprCtx) -> bool {
+            matches!(
+                expr.get_type(expr_ctx),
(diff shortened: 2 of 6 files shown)


kdy1 (approved):
Good idea! Thanks!

PR #12069, opened 2026-07-23
perf(es/minifier): reuse ProgramData across compress iterations

hamidrezahanafi:
## Summary
- Reuse a single `ProgramData` on the compressor across hoist and fixed-point iterations instead of allocating a fresh analysis every pass.
- Add `ProgramData::analyze` / `clear_retaining_capacity` so top-level `vars`, `initialized_vars`, and `scopes` capacity is retained while facts are cleared correctly (scopes must be cleared, not zeroed in place).

This is a low-risk foundation for later coarse re-analyze gating. Local wall-clock gains on CodSpeed-style lib benches were small / within noise; peak RSS was essentially unchanged.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo check --all --all-targets`
- [x] `cargo test -p swc_ecma_minifier --features concurrent --features par-core/chili`
- [ ] CodSpeed CI on this PR

Made with [Cursor](https://cursor.com)

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/mod.rs
+++ b/crates/swc_ecma_minifier/src/compress/mod.rs
@@ -25,7 +25,7 @@ use crate::{
     compress::hoist_decls::decl_hoister,
     mode::Mode,
     option::{CompressOptions, MangleOptions},
-    program_data::analyze,
+    program_data::ProgramData,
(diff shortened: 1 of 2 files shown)


hamidrezahanafi:
closing as it didn't move the performance

Issue #11517, opened 2026-01-29
SWC minifier incorrectly merges conditions (semantic change)


lerrybe:
### Describe the bug

Thank you for creating such a great tool.

While testing, I noticed that the runtime behavior of the code built with SWC differs from the behavior in the development environment, and during this process I discovered the following issue.

In the SWC Playground, the following behavior also can be observed.

When two conditional branches have similar structures and differ only by string literals, the minifier appears to treat them as equivalent and merges them into a single branch. 

### Input code

```typescript
export const buildErrorLog = ({
  errorType,
  mode,
}: {
  errorType?: string;
  mode: 'modeA' | 'modeB';
}) => {
  const isModeA = mode === 'modeA';
  const isModeB = mode === 'modeB';
  
  if (errorType === 'A_ERROR') {
    const message = 'A error occured';
    return { fieldX: true, fieldY: true, message };
  }
  if (errorType === 'B_ERROR') {
    const message = 'B error occured';
    return { fieldX: true, fieldY: true, message };
  }
  return { fieldX: true, fieldY: true, message: 'Invalid configuration' };
};
```

### Config

```json
This behavior occurs in the SWC Playground when compiling TypeScript targeting ES2015 or later with minify, compress, and mangle enabled.

(A SWC Playground link is attached.)
```

### Link to the code that reproduces this issue

[...]

PR #12079, opened 2026-07-28
perf(es/minifier): Apply pending substitutions in batch for if-dense statement lists

hamidrezahanafi:
**Description:**

Profiling the minifier on large inputs showed that `inline_with_multi_replacer` is invoked tens of thousands of times per `optimize()` call (each call = one `Finalizer` walk + one `NormalMultiReplacer` walk over a small subtree), with hit rates of only 0.4–1.7% (replacer) and 0.006–0.04% (finalizer) — i.e. ~99% of the walks change nothing. About half of them came from `merge_similar_ifs`, which flushed pending substitutions into **each `if` statement individually** before merging (34k double walks per run on typescript.js alone).

This PR applies pending substitutions to the whole statement list in a **single batch walk** when the list is `if`-dense (`>= 16` ifs), and keeps the per-`if` loop for sparse lists. The two forms are semantically equivalent: the substitution maps do not change during application except for consumption of applied entries, which happens in the same statement order either way. The density gate preserves the win where batching pays off (Σ(if subtrees) ≈ Σ(list)) without regressing `if`-sparse inputs, where per-`if` walks are already the cheaper strategy.

**Output:** byte-identical on the entire fixture corpus — the full suite passes with **zero fixture updates required** (including the big projects/bench lib fixtures).

**Test plan:**
- [x] `cargo fmt --all`
- [x] `cargo clippy --all --all-targets -- -D warnings`
- [x] `cargo test -p swc_ecma_minifier` (fully green, no fixture updates)
- [x] `./scripts/exec.sh` (2128/2128 pass)

**BREAKING CHANGE:**

None.

https://github.com/swc-project/swc/issues/11517 (the pre-existing workaround this change preserves, but cheaper)

### DIFF
--- a/.changeset/batch-substitutions-if-dense.md
+++ b/.changeset/batch-substitutions-if-dense.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_minifier: patch
+---
+
+perf(es/minifier): Apply pending substitutions in batch for if-dense statement lists
--- a/crates/swc_ecma_minifier/src/compress/optimize/conditionals.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/conditionals.rs
@@ -251,13 +251,10 @@ impl Optimizer<'_> {
             return;
         }
 
-        // we must inline first to avoid https://github.com/swc-project/swc/issues/11517
-        stmts
-            .iter_mut()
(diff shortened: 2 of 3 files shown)


hamidrezahanafi:
@kdy1 I am testing a few optimizations with codespeed that actually move the needle when LTO is enabled. should there be someway to have LTO enabled for the benchmark? like adding a label to check LTO baseline?

PR #12093, opened 2026-07-31
fix(es/parser): reject unicode escapes in RegExp flags

Ayush7614:
**Description:**

ECMA-262 requires a Syntax Error when IdentifierPart in RegExp flags contains a Unicode escape sequence. SWC's lexer had an explicit TODO for this and discarded the `has_escape` flag from `read_word_as_str_with`, so inputs like `/a/g\u0069` parsed successfully even though engines reject them.

This change:

- emits `UnicodeEscapeInRegExpFlags` when flags contain `\u` / `\u{...}` escapes
- allows flags that start with `\` so fully-escaped flags such as `/\u0067` are scanned and rejected correctly

Verified against Node and with fixtures under `tests/errors/regexp-flags-unicode-escape/**`.

N/A (direct PR; no existing tracked issue)


kdy1 (approved):
Thanks!

PR #12094, opened 2026-07-31
fix(es/module): strip import attributes from CJS dynamic import

Ayush7614:
**Description:**

`cjs_dynamic_import` had a `TODO: handle import assert`. With import attributes, CommonJS lowering was incorrect:

- `import("./data.json", { with: { type: "json" } })` became `require("./data.json", { with: ... })` — invalid `require` arity/options
- `import(path, { with: ... })` became `Promise.resolve(path, opts).then(...)` — attributes silently discarded by `Promise.resolve`

`require` cannot express import attributes, so this change truncates dynamic import arguments to the module specifier only (matching the existing AMD path that already used `args[..1]`).

Fixtures cover literal paths, computed paths, legacy `assert`, and mixed cases under `tests/fixture/common/dynamic-import-attributes/**`.

N/A (direct PR; no existing tracked issue)

### DIFF
--- a/.changeset/cjs-dynamic-import-attributes.md
+++ b/.changeset/cjs-dynamic-import-attributes.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_transforms_module: patch
+---
+
+fix(es/module): strip import attributes from CommonJS dynamic import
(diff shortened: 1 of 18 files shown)


magic-akari:
I don’t think we should do this. In general, import attributes have no well-defined CommonJS equivalent. Stripping them would silently change the program’s semantics, so explicitly special-casing them here actually makes the behavior worse.

I would prefer to leave the current lowering unchanged. When users choose CommonJS output, we should trust that their input does not contain import attributes. We should neither detect nor transform this unsupported case, nor make every user pay additional complexity or runtime overhead for it.

PR #12104, opened 2026-08-06
fix(es/react-compiler): recover source map spans from locations

mattcosta7:
React Compiler reverse conversion discarded source locations when Babel `BaseNode.start/end` were absent, producing valid but severely sparse source maps. `BaseNode.loc` retains the necessary offsets.

- **Span recovery**
  - Preserve existing absolute `start/end` behavior.
  - Fall back to `loc.start.index` / `loc.end.index`.
  - Collapse a missing end index to a zero-width span.
  - Keep `DUMMY_SP` only when neither position representation exists.

```rust
let start = loc.start.index.map(|index| source_file_start_pos + BytePos(index));
let end = loc.end.index.or(loc.start.index)
    .map(|index| source_file_start_pos + BytePos(index));
```

- **Correct offset basis**
  - Forward conversion stores `start/end` as absolute SWC `BytePos` values.
  - `loc.index` is file-relative, so the original source-file start position is threaded into `ReverseCtx` before rebuilding spans. This avoids incorrect mappings for later files in a shared `SourceMap`.

- **Regression coverage**
  - Cover all `start/end` and `loc` fallback branches.
  - Add a `useMemo`/`useCallback`-heavy source-map density test that asserts mapping and generated-line counts.

Nodes synthesized by React Compiler with neither `start/end` nor `loc` still receive `DUMMY_SP`; mapping those requires separate source-span inheritance work.

### DIFF
--- a/crates/swc_ecma_react_compiler/src/convert_ast.rs
+++ b/crates/swc_ecma_react_compiler/src/convert_ast.rs
@@ -28,6 +28,7 @@ use crate::preserved_ast::PreservedAst;
 pub struct ConvertResult {
     pub file: File,
     pub preserved_ast: PreservedAst,
+    pub source_file_start_pos: swc_common::BytePos,
 }
 
 /// Converts an SWC AST to the React compiler's Babel-compatible AST.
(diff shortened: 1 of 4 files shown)


magic-akari:
duplicate #11979

PR #12105, opened 2026-08-06
fix(react-compiler): make fast check conservative

marcoshernanz:
## Description

This fixes false negatives in the existing React Compiler `fast_check::is_required` predicate and documents its conservative skip-gate contract.

Before this PR, `is_required` was a low-false-positive heuristic, but it was not safe for callers that skip React Compiler when it returns `false`. On the real v0 corpus it missed anonymous `forwardRef` callbacks and a module using member hook calls. This PR improves the existing method rather than adding a second public predicate, following maintainer feedback.

The new implementation allows false positives but not false negatives: `false` means compilation cannot change the program. It scans every function context for JSX, actual hook calls matching React Compiler HIR's own hook-name rule, member hook calls, and opt-in/dynamic-gating directives in real directive prologues.

The Node `isReactCompilerRequired{,Sync}` APIs continue to call `is_required`. Fatal and recovered parse errors return `true`, because an unsuccessful fixed-syntax parse cannot prove that compilation is unnecessary.

The package test command enumerates both test files explicitly so it remains compatible with the existing Windows + Node 18 CI job.

## v0 correctness corpus

I benchmarked a fixed v0 checkout (`1ab042a47c053bf5ddad5ffb9c6af153ce8e5b56`) on a 16-vCPU / 32-GB Vercel DevBox:

- 1,816 tracked JS/TS source files
- 10,464,109 source bytes
- zero parser failures
- 257 modules actually transformed by the Rust React Compiler
- 67 compiler diagnostics

The harness parses the corpus once, runs detectors repeatedly, and uses the actual compiler result as ground truth. It asserts that gated and ungated transformed paths are identical.

| Detector | Selected files | False negatives |
| --- | ---: | ---: |
| `is_required` before this PR | 279 | 10 / 257 (3.89%) |
| Current Next.js conservative visitor | 315 | 0 / 257 |
| This PR: improved `is_required` | **302** | **0 / 257** |

The old implementation missed nine real `React.forwardRef` UI modules and one hook module using `React.useState` / `React.useEffect`. The improved implementation retains all 257 transforms while rejecting 13 more false positives than the current Next.js visitor.

## End-to-end compiler effect

A fresh optimized release run after applying all review feedback used five interleaved gated/ungated samples on the exact corpus:

| Pipeline | Median | Transformed modules |
| --- | ---: | ---: |
| No gate | 2,371.956 ms | 257 |
| Improved `is_required` gate | **1,906.180 ms** | 257 |

That removes **465.776 ms / 19.64%** from the React Compiler pipeline on the real v0 corpus without changing compiler output. The detector itself took 20.624 ms for the full 10.46-MB corpus (mean of 100 passes).

The temporary synthetic CodSpeed benchmark was removed at review request. The performance evidence remains the real v0 corpus with actual compiler output as the correctness oracle.

## Review feedback incorporated

- Improved the existing `is_required` API instead of introducing `may_require`.
- Detect hooks only at call sites and reuse React Compiler's `is_hook_name` rule (`use` plus uppercase ASCII or a digit).
- Restrict opt-in strings to leading directive prologues using `Stmt::can_precede_directive`.
- Treat both fatal and recovered parser errors conservatively in the Node binding.
- Added actual-compiler regressions proving name-only and JSX-free `React.createElement` wrapper cases do not compile.
- Removed the synthetic benchmark and changed the changeset from a minor feature to a patch fix.

## Validation

- `cargo test -p swc_ecma_react_compiler` (107 unit tests + 15 fixtures)
- `cargo test -p binding_react_compiler_node` (14 passed)
- `cargo clippy -p swc_ecma_react_compiler -p binding_react_compiler_node --all-targets -- -D warnings`
- `pnpm --filter @swc/react-compiler build:dev`
- `pnpm --filter @swc/react-compiler test` (8/8)
- Earlier full-PR validation: `cargo clippy --all --all-targets -- -D warnings`; `packages/core: pnpm build:dev && pnpm test` (117 passed, 3 skipped)
- Final GPT-5.6 Sol xhigh + Claude Opus 5 xhigh autoreview panel: zero findings, “patch is correct” (0.99 confidence)

Follow-up to [vercel/next.js#96820](https://github.com/vercel/next.js/pull/96820).


marcoshernanz:
@kdy1, could you review the API and semantics of `may_require`?

Next.js wants to use this as a conservative skip gate before invoking React Compiler. On the tracked v0 corpus (1,816 JS/TS modules, 10.46 MB):

- The existing `is_required` missed 10 of the 257 modules that actually transformed (nine `React.forwardRef` modules and one member-hook module).
- `may_require` had zero false negatives and selected the same 315 modules as the conservative Next.js implementation.
- The upstream detector was 1.52% faster than the Next.js-local implementation.
- Applying the safe gate reduced the measured full compiler pipeline from 2,312 ms to 1,817 ms (-21.4%).

This preserves `is_required` unchanged; the narrow review decision is whether the conservative predicate belongs in this crate and its Node binding, and whether `may_require` is the right contract/name. The PR includes focused false-negative tests, fixture coverage, a CodSpeed benchmark, and a changeset.

crates/swc_ecma_react_compiler/src/fast_check.rs
```diff
@@ -1,14 +1,33 @@
...
+///
+/// False positives only add compiler work, while false negatives could change
+/// output. This scans every function context for JSX, plausible hook calls, and
+/// opt-in directives, including dynamic gating directives.
+pub fn may_require(program: &Program) -> bool {
+    let mut finder = PotentialFinder::default();
+    finder.visit_program(program);
+    finder.found
+}
+
```
lukesandberg, line 20: this is perhaps a question for maintainers, but this really makes me think that only `may_require` should exist and the docs should say 'there may be some false positives but no false negatives'

crates/swc_ecma_react_compiler/src/fast_check.rs
```diff
@@ -18,6 +37,80 @@ struct Finder {
...
 }
 
+#[derive(Default)]
+struct PotentialFinder {
+    found: bool,
+}
+
+fn is_hook_callee(expr: &Expr) -> bool {
+    match expr {
+        Expr::Ident(ident) => ident.sym.starts_with("use"),
```
lukesandberg: should we restrict this test to only function calls `useEffect`  not anything like `const user = getUser()`

we could also enhance it to be `use` + an upper case ascii char

crates/swc_ecma_react_compiler/src/fast_check.rs
```diff
@@ -18,6 +37,80 @@ struct Finder {
...
+            return;
+        }
+        node.visit_children_with(self);
+    }
+
+    fn visit_str(&mut self, node: &Str) {
+        if self.found {
+            return;
+        }
+        self.found = is_opt_in_directive(node);
```
lukesandberg: a directive is only a directive if it is the leading statement of a function/module

lukesandberg: i think there is an existing function for this in swc 'may_lead_directive' or something

crates/swc_ecma_react_compiler/Cargo.toml
```diff
@@ -30,8 +30,13 @@ swc_ecma_parser    = { version = "43.0.0", path = "../swc_ecma_parser", default-
...
 [dev-dependencies]
-swc_common      = { version = "24.0.0", path = "../swc_common" }
+codspeed-criterion-compat = { workspace = true }
+swc_common                  = { version = "24.0.0", path = "../swc_common" }
 swc_ecma_parser = { version = "43.0.0", path = "../swc_ecma_parser", default-features = false, features = [
   "typescript",
 ] }
 testing         = { version = "25.0.0", path = "../testing" }
+
+[[bench]]
```
lukesandberg: im not sure the benchmark is very important, drop it?

crates/swc_ecma_react_compiler/src/fast_check.rs
```diff
@@ -1,14 +1,33 @@
...
+/// Conservatively determines whether infer or annotation mode could compile
+/// anything.
+///
+/// False positives only add compiler work, while false negatives could change
+/// output. This scans every function context for JSX, plausible hook calls, and
+/// opt-in directives, including dynamic gating directives.
+pub fn may_require(program: &Program) -> bool {
+    let mut finder = PotentialFinder::default();
+    finder.visit_program(program);
+    finder.found
```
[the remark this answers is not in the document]
marcoshernanz: I verified this against the compiler implementation rather than widening the gate. Infer-mode name eligibility still requires `calls_hooks_or_creates_jsx`; a PascalCase/hook-like name alone does not compile. The new regression runs the actual transform and proves that `App` returning `React.createElement(...)` and a constant-returning `useComputed` produce no compiler output, while `may_require` safely returns false.

crates/swc_ecma_react_compiler/src/fast_check.rs
marcoshernanz: I kept `is_required` for source compatibility because it has been public since #10422; removing it here would be a breaking Rust API change. I tightened the docs as suggested: `may_require` now explicitly says false positives are possible but false negatives are not, and `is_required` directs skip-gate callers to `may_require`. The Node binding uses only `may_require`. Leaving this thread open for a maintainer to decide whether the old public API should eventually be deprecated or removed in a breaking release.

bindings/binding_react_compiler_node/src/support.rs
```diff
@@ -39,10 +39,10 @@ fn is_react_compiler_required_inner(code: &str) -> bool {
     );
 
     let Ok(program) = program else {
-        return false;
+        return true;
```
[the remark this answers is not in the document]
marcoshernanz: Fixed in `8cacc00ba3`: the binding now retains `recovered_errors` and returns `true` when the vector is nonempty. The Rust regression uses top-level `return 1;` and asserts that parsing succeeds with a recovered error before checking the conservative result.

crates/swc_ecma_react_compiler/src/fast_check.rs
marcoshernanz: Done in `8cacc00ba3`. Hook detection now happens only in `visit_call_expr`, so identifiers such as `const user = getUser()` are ignored. I reused `react_compiler_hir::environment::is_hook_name` instead of duplicating the compiler rule; it accepts `use` followed by an ASCII uppercase character or digit (`use3rdParty` is valid). Added positive and negative tests. On v0 this tightened selection from 315 to 302 files while retaining all 257 transforms.

crates/swc_ecma_react_compiler/src/fast_check.rs
marcoshernanz: Done in `8cacc00ba3`. The implementation uses SWC's `Stmt::can_precede_directive` and scans only the directive prologue of modules, scripts, functions, and block-bodied arrows. Tests cover leading directives after `use strict`, plus non-leading strings, variable string values, and strings after imports.

crates/swc_ecma_react_compiler/Cargo.toml
marcoshernanz: Dropped in `8cacc00ba3`, including the benchmark file, Cargo manifest entry, dev dependency, and lockfile change. The PR body now keeps the real-v0 corpus results as the performance evidence.

kdy1 (dismissed):

kdy1 (commented):
One question. Why do we need an extra method? Can't we just modify `is_required`?

bgw:
> One question. Why do we need an extra method? Can't we just modify is_required?

@marcoshernanz put this in the description:

> Removing it here would also be a breaking Rust API change, so this PR preserves and documents it for compatibility.

But I think you're asking why we don't just change the behavior of the existing API, and not the API itself?

lukesandberg:
yeah this seems like it is just an improvement to is_required, fix false negatives, reduce false positives, and improve docs

crates/swc_ecma_react_compiler/src/fast_check.rs
marcoshernanz: Resolved in signed commit `6f72747ac0`: I removed `may_require` and moved the reviewed conservative implementation and contract onto the existing `is_required` API. Its docs now state that false positives are possible but false negatives are not, and the Node binding calls that single method. I also changed both changesets from minor features to patch fixes. On the pinned v0 corpus it selects 302 modules, retains all 257 actual transforms, and has zero false negatives.

crates/swc_ecma_react_compiler/src/fast_check.rs
```diff
@@ -18,6 +40,160 @@ struct Finder {
...
+        node.visit_children_with(self);
+    }
+
+    fn visit_call_expr(&mut self, node: &CallExpr) {
+        if self.found {
+            return;
+        }
+        if let Callee::Expr(expr) = &node.callee {
+            if is_hook_callee(expr) {
+                self.found = true;
```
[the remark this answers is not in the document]
marcoshernanz: I verified this exact wrapper path against the actual compiler fork and added a regression rather than widening the gate. `get_component_or_hook_like` still requires `calls_hooks_or_creates_jsx` after `forwardRef` / `memo` inference. For the JSX-free `React.createElement` example, the test now asserts all three facts: `is_required == false`, the real compiler returns no program, and it emits no diagnostics. This wrapper therefore does not compile; treating it as required would only add a false positive.

marcoshernanz:
Ready for re-review, @kdy1 and @magic-akari.

The final API concern is addressed in signed commit [`6f72747ac0`](https://github.com/marcoshernanz/swc/commit/6f72747ac0e91c182f64cae2cff666696194c702):

- Removed the extra `may_require` method and made the existing `fast_check::is_required` conservative.
- Preserved the earlier hook-call, member-hook, directive-prologue, and recovered-parser-error fixes.
- Added an actual-compiler regression for the JSX-free `forwardRef(...React.createElement...)` case.
- Changed the changeset from a minor feature to a patch fix.
- Resolved all seven review threads.

On the pinned v0 corpus, the final predicate selects 302 modules, retains all 257 compiler transformations with zero false negatives, and reduces the compiler pipeline from 2,371.956 ms to 1,906.180 ms (-19.64%). The final GPT-5.6 Sol xhigh + Claude Opus 5 xhigh review panel reported zero findings, and the relevant Linux/macOS/Windows CI jobs are green.

Would appreciate another pass when you have time.

kdy1 (approved):
Thank you!

PR #12106, opened 2026-08-07
fix(es/minifier): remove unused variable initializer cycles

SyMind:
**Description:**

DCE previously visited variable initializers without associating deferred references with the initialized binding. References inside function, arrow, and eligible class expressions were therefore treated as top-level graph entries, which prevented otherwise unreachable dependency cycles from being removed.

For example:

```js
var AsyncParsePluginContext = class {
    parse(current) {
        return parseAsync(current);
    }
};

async function parseAsync() {
    return new AsyncParsePluginContext();
}
```

The reference to `parseAsync` was incorrectly treated as an entry:

```text
entry -> parseAsync -> AsyncParsePluginContext
```

This PR attributes references from direct deferred-value initializers to their variable binding:

```text
AsyncParsePluginContext <-> parseAsync
```

With no external entry into this strongly connected component, DCE can remove the entire unused cycle.

The optimization applies to identifier bindings initialized directly with:

- Function expressions
- Arrow functions
- Class expressions without heritage, decorators, computed keys, static initialization, or other observable definition-time work

Named function and class expressions are also handled using their local expression bindings, preventing self-references from being mistaken for references to outer declarations with the same symbol.

The conservative behavior is preserved for eager or potentially observable initializers, including:

- Destructuring bindings
- Class heritage
- Effectful class definitions
- Static initialization
- Eager `let` and `const` dependency cycles

The new DCE fixture covers:

- Removable `var`, `let`, and `const` function and class expression cycles
- Cycles retained through an external reference
- Named function and class expression bindings
- Class heritage and effectful static initialization
- Eager `let` and `const` cycles

The minifier size snapshot is also updated to reflect the resulting `echarts.js` gzip reduction from `314.10 KiB` to `314.09 KiB`.

**Testing:**

- `cargo test -p swc_ecma_transforms_optimization`
- `crates/swc_ecma_minifier/scripts/exec.sh`
- `crates/swc_ecma_minifier/scripts/test.sh`
- `cargo fmt --all`

### DIFF
--- a/.changeset/fix-minifier-unused-initializer-cycles.md
+++ b/.changeset/fix-minifier-unused-initializer-cycles.md
@@ -1,0 +1,7 @@
+---
+swc_core: patch
+swc_ecma_minifier: patch
+swc_ecma_transforms_optimization: patch
+---
+
+fix(es/minifier): Remove unused variable initializer cycles
--- a/crates/swc_ecma_minifier/tests/libs-size.snapshot.md
+++ b/crates/swc_ecma_minifier/tests/libs-size.snapshot.md
@@ -2,7 +2,7 @@
 | --- | --- | --- | --- |
 | antd.js | 6.38 MiB | 2.06 MiB | 445.44 KiB |
 | d3.js | 542.74 KiB | 261.26 KiB | 85.54 KiB |
-| echarts.js | 3.41 MiB | 977.04 KiB | 314.10 KiB |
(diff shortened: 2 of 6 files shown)


SyMind:
@codex review

SyMind:
@kdy1 please review

kdy1 (dismissed):
Thank you! I'll merge after adding changeset

PR #12110, opened 2026-08-10
fix(es): avoid incorrect `instanceof` optimizations

### LINKED ISSUE #12111: Optimizer can miscompile or discard `instanceof` evaluation
### Describe the bug

SWC can incorrectly fold some `instanceof` expressions based on operand shape.

For example:

```js
console.log(({ __proto__: null }) instanceof Object);
```

evaluates to `false`, but can be optimized to:

```js
console.log(true);
```

### Expected behavior

Avoid folding `instanceof` unless the result can be established correctly.

Fix: #12110

teamleaderleo:
**Description:**

Closes #12111.

Avoid folding `instanceof` from operand shape alone. The operator may invoke `Symbol.hasInstance`, and an invalid right-hand operand may throw, so folds such as object-like values to `true` or primitive-like values to `false` are not generally valid.

For example:

```js
({ __proto__: null }) instanceof Object
```

evaluates to `false`, but could previously fold to `true`.

This keeps SWC's existing Terser-compatible treatment of discarded `instanceof` expressions unchanged. Tests under `tests/terser` are unchanged.

The change removes the operand-shape-based `instanceof` folds in the minifier and expression simplifier and adds SWC-owned regressions for cases where the result is used.

`in` / `pure_getters` are unchanged.

**BREAKING CHANGE:**

None.

**Validation:**

```text
cargo fmt --all -- --check
cargo clippy -p swc_ecma_transforms_optimization --all-targets -- -D warnings
cargo clippy -p swc_ecma_minifier --all-targets -- -D warnings
focused instanceof used-result minifier fixture
```


teamleaderleo:
Updated the draft.

Austaras (changes requested):
As I stated in related issue

teamleaderleo:
Ok, so we keep the existing Terser assumptions for the unused `instanceof` case, but the other changes are acceptable?

teamleaderleo:
I updated this PR. I kept the existing discarded-result behavior and reduced the PR to the incorrect `instanceof` constant folding.

crates/swc_ecma_minifier/tests/terser/compress/pure_funcs/relational/output.js
```diff
@@ -1,4 +1,4 @@
-foo() instanceof bar();
```
Austaras: This seems wrong

crates/swc_ecma_minifier/src/compress/optimize/mod.rs
```diff
@@ -639,6 +639,17 @@ impl Optimizer<'_> {
         )
     }
 
+    fn is_pure_call_value(&self, e: &Expr) -> bool {
+        match e {
+            Expr::Call(CallExpr { ctxt, .. }) => ctxt.has_mark(self.marks.pure),
```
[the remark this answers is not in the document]
teamleaderleo: I have extended the guard to pure-marked `NewExpr` and `TaggedTpl` values too, and added fixture coverage.

crates/swc_ecma_minifier/tests/terser/compress/pure_funcs/relational/output.js
teamleaderleo: Okay, my interpretation: keep the existing discarded `instanceof` behavior generally,

but leave the full expression intact when one of its operands comes from an expression SWC already marked pure.

I updated the PR that way.

Austaras: Sorry, I haven't seen `foo` has been marked as pure. `foo()` should be removed in this case.

The general idea is not to change any test case under `tests/terser`.

teamleaderleo: Thank you for clarifying. I restored the `tests/terser` expectations and removed the pure-marked `instanceof` exception. The PR is back to only fixing the incorrect constant folding.

Austaras: Despite being useless, could you restore some harmless correct constant folding like `1 instanceof Object`?

PR #12115, opened 2026-08-12
fix(react-compiler): preserve TypeScript function overload signatures

CPunisher:
**Description:**

Preserve whether a function originally had an implementation body when converting through the React Compiler AST.

React Compiler's `FunctionDeclaration` representation always contains a block body. The SWC converter therefore synthesized an empty block for TypeScript overload signatures, but the reverse conversion did not restore the original missing body. The TypeScript transform then treated every overload signature as a concrete runtime function, producing duplicate default exports.

This change records the original body presence in `PreservedAst` and restores `None` for overload signatures. The integration test forces React Compiler output with `compilationMode: "all"` and verifies that the overloads remain bodyless while the implementation retains its body.

Validation:

- `cargo fmt --all`

**BREAKING CHANGE:**

None.

https://github.com/web-infra-dev/rspack/issues/15158


magic-akari:
My concern is that `had_body` exists only in SWC’s `PreservedAst`. React Compiler never sees it.

```ts
// Original TypeScript
export default function foo(x: string): string;
export default function foo(x: number): number;
export default function foo(x: string | number) {
  return x;
}
```

React Compiler receives an AST semantically equivalent to:

```ts
export default function foo(x: string): string {}
export default function foo(x: number): number {}
export default function foo(x: string | number) {
  return x;
}
```

Instead of:

```text
ExportDefaultDeclaration(TSDeclareFunction)
ExportDefaultDeclaration(TSDeclareFunction)
ExportDefaultDeclaration(FunctionDeclaration)
```

At the React Compiler boundary, two type-only signatures have already been misrepresented as runtime default-exported functions. That input has the wrong semantics, so its behavior is unpredictable, even if SWC can restore the missing bodies afterward.

kdy1 (approved):

PR #12122, opened 2026-08-17
fix(es/modules): preserve destructuring assignment targets in SystemJS

SyMind:
**Description:**

Fix the SystemJS transform producing an invalid assignment target for ordinary destructuring assignments.

The export binding rewriter temporarily takes an `AssignTargetPat` while looking for exported bindings. It previously restored the pattern only when an exported binding was found, leaving non-exported patterns as `Invalid` and causing code generation to emit `<invalid> = []`.

This change always restores the assignment pattern before recursively visiting it. It also adds a regression fixture for `[a, b] = []`.

**Validation:**


SyMind:
@kdy1 please review

magic-akari (approved):

PR #12131, opened 2026-08-20
fix(es/minifier): Mark for update and test as executed multiple time

### LINKED ISSUE #12128: Minifier changes closure semantics when inlining a helper called from a for-loop test
### Describe the bug

When a helper function called from a for loop test declares a local variable that is captured by a nested closure, the minifier may move that local variable outside the helper function.

The helper is invoked multiple times, so the local variable should have a separate binding for every invocation. After minification, all iterations share one binding, which changes the value observed by the closure.

This requires inline, reduce_vars, and toplevel in the reproduction below.

### Input code

~~~typescript
var callback;
var _loop = function () {
    var value = callback;
    return value ? "break" : callback = function () {
        return value;
    };
};

for (; _loop() !== "break";);
console.log(callback());
~~~

### Config

~~~json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript"
    },
    "target": "es5",
    "minify": {
      "compress": {
        "inline": true,
        "reduce_vars": true,
        "toplevel": true
      },
      "mangle": false
    }
  },
  "minify": true,
  "isModule": false
}
~~~

[SWC Playground reproduction](https://play.swc.rs/?version=1.16.1&code=H4sIAAAAAAAAA22OwQrCMAyG73mKuFN78QEMxUeRrHQiC43EtRfx3dcxpzAWcglfvj%2BpbBhZpOc4EtQ23UT1iQGHkuP00IzO4xuw1UIrS0mN%2Fp2FWJqK5S%2B8Ytdb4rHDy2%2FtMG9vrlkfgtYwqKGj9ZtmnELYYskTRM0vlXQWvbvthvOeZgf83ZTPAAAA&config=H4sIAAAAAAAAA2WOSw6DMAwF95wi8pptN71Dz1BFwaBU%2BclOUBHK3etCKEjd2TPPT147peDFBu5qlVGWpImRfrsQXkLWbyGAxms2ZFOGTdZ%2Bv8maJsxbgm%2FQoLfBjsu1yESfCJkvTKgNzgYUlqlgf3LCoRh8zvLPv8wxOZzRNdNEPRLgdZjct3TUjndduxY4PztKwfIjDuVyUD%2F30bFkGQEAAA%3D%3D)

### SWC Info output

Reproduced in SWC Playground using version 1.16.1.

### Expected behavior

The minified code should preserve a separate value binding for every invocation of _loop.

Executing both the input and the minified output should print:

~~~text
undefined
~~~

It is valid to inline _loop, but the function scope containing the captured value binding must be preserved.

For example, output with equivalent semantics could retain an IIFE:

~~~js
var callback;
for (; "break" !== function () {
    var value = callback;
    return value ? "break" : callback = function () {
        return value;
    };
}(););
console.log(callback());
~~~

### Actual behavior

SWC produces:

~~~js
for(var callback,value;"break"!==((value=callback)?"break":callback=function(){return value}););console.log(callback());
~~~

The local value declaration has been moved outside the helper function.

During the first iteration, the created closure captures the shared value variable. During the second iteration, that same variable is overwritten with callback. As a result, invoking callback() returns the callback function instead of the first invocation's undefined value.

Therefore, the minified output is not semantically equivalent to the input.

### Version

1.16.1

Austaras:
**Description:**

**BREAKING CHANGE:**

Closes #12128

### DIFF
--- a/crates/swc_ecma_minifier/tests/terser/compress/issue_t120/issue_t120_1/output.js
+++ b/crates/swc_ecma_minifier/tests/terser/compress/issue_t120/issue_t120_1/output.js
@@ -1,5 +1,8 @@
 function foo(node) {
-    for(var i; (i = node.data) && i.a != i.b;)node = node.data;
+    for(; function(obj) {
+        var i = obj.data;
+        return i && i.a != i.b;
+    }(node);)node = node.data;
     return node;
 }
 var x = {
--- a/.changeset/preserve-loop-execution-context.md
+++ b/.changeset/preserve-loop-execution-context.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_minifier: patch
+---
+
+fix(es/minifier): Preserve closure bindings in repeatedly executed loop expressions
--- a/crates/swc_ecma_minifier/src/compress/optimize/mod.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/mod.rs
@@ -1898,10 +1898,8 @@ impl VisitMut for Optimizer<'_> {
     }
 
     fn visit_mut_do_while_stmt(&mut self, n: &mut DoWhileStmt) {
-        {
-            let ctx = self.ctx.clone().with(BitCtx::ExecutedMultipleTime, true);
-            n.visit_mut_children_with(&mut *self.with_ctx(ctx));
-        }
+        let ctx = self.ctx.clone().with(BitCtx::ExecutedMultipleTime, true);
+        n.visit_mut_children_with(&mut *self.with_ctx(ctx));
     }
 
     #[cfg_attr(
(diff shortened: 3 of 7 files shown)


crates/swc_ecma_minifier/tests/terser/compress/issue_t120/issue_t120_1/output.js
```diff
@@ -1,5 +1,8 @@
 function foo(node) {
-    for(var i; (i = node.data) && i.a != i.b;)node = node.data;
+    for(; function(obj) {
+        var i = obj.data;
+        return i && i.a != i.b;
+    }(node);)node = node.data;
```
Austaras, line 5: [Terser](https://github.com/terser/terser/commit/f5b720c6b9322ba7e40ae39e5897f9aa2c19eab7) itself abandons invoke IIFE in for loop so I guess its fine.

kdy1 (dismissed):

Issue #12130, opened 2026-08-20
block_scoping: const reassignment succeeds when transpiled to ES5


yeonhj0507:
### Describe the bug

`const` declared a read-only binding, but swc simply transpiles it to `var`, without the reassingment check.

### Input code

```typescript
const x = 0;
try {
  x = 42;
  console.log(x);
} catch (e) {
  console.log(e);
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false
    },
    "target": "es5",
    "loose": false,
    "minify": {
      "compress": false,
      "mangle": false
    }
  },
  "module": {
    "type": "commonjs"
  },
  "minify": false,
  "isModule": false
}
```

### Link to the code that reproduces this issue

https://play.swc.rs/?version=1.16.1&code=H4sIAAAAAAAAA0vOzysuUahQsFUwsOYqKapUqOZSAHNNjKyBrGSgdH5Oql5OfrpGhaY1V61CcmJJcoaCRqomWCWyfCpIngsAz9n%2FWFEAAAA%3D&config=H4sIAAAAAAAAA1WPSw7DIBBD9zkFmnW33fQOPQSik4iInzCViqLcvUAgbXbYfmOLbRKCVih6iK08iwgyguOpi4PskvwUh1hZCRV1SHQb6YoazdKAm7UfCSUZF07tCveOk%2FEePPDuWe30nP8HlbchMnAFKyrdYvg6N%2FVJsv71bmH%2FSMqhqtpmvVtBP3Isnu2k8RznR%2Ff%2BBaNtxkwaAQAA

### Expected behavior

```js
function _readOnlyError(r) { throw new TypeError('"' + r + '" is read-only'); }
var x = 0;
try {
  42, _readOnlyError("x");
  console.log(x);
} catch (e) {
  console.log(e);
}
```

### Actual behavior

```js
var x = 0;
try {
    x = 42;
    console.log(x);
} catch (e) {
    console.log(e);
}
```

### Version

1.16.1

### Additional context

swc defines the helper function named `_read_only_error(name) {...}` that throws a TypeError when a binding is read-only. However, swc doesn't call this helper when transpiling a reassignment to a `const` binding.

magic-akari:
This is a long-standing known limitation, not a regression. We have been aware that lowering a `const` binding to `var` does not preserve the runtime error caused by writing to that binding, and the decision not to emulate this behavior was intentional.

[...]

PR #12132, opened 2026-08-20
fix(es2015): preserve const reassignment errors in block scoping

Alb3e3:
## Summary
- preserve read-only errors when `const` bindings are lowered by the ES2015 block scoping transform
- add a regression snapshot for `const x = 0; x = 42;`

## Details
`block_scoping` converted `const` declarations to `var`, but did not remember that the original binding was read-only. That meant reassignments like `x = 42` were emitted unchanged instead of calling `_read_only_error("x")`.

This patch collects `const` bindings before mutation and rewrites simple identifier assignments targeting those bindings so the RHS is still evaluated, but the transformed output ends with `_read_only_error(...)` instead of performing the write.

## Testing
- `cargo test -p swc_ecma_transforms_compat --test es2015_block_scoping issue_12130 -- --exact`
- `cargo test -p swc_ecma_transforms_compat --test es2015_block_scoping issue_2027_1 -- --exact`
- `cargo test -p swc_ecma_transforms_compat --test es2015_block_scoping for_loop -- --exact`

A broader run of `cargo test -p swc_ecma_transforms_compat --test es2015_block_scoping` still hits pre-existing local environment failures for mocha-backed exec tests, so I kept verification focused on the changed contract.

## AI disclosure
This patch was drafted with AI assistance and then manually reviewed and verified with the targeted SWC tests above.

### DIFF
--- a/crates/swc_ecma_compat_es2015/src/block_scoping/mod.rs
+++ b/crates/swc_ecma_compat_es2015/src/block_scoping/mod.rs
@@ -4,7 +4,7 @@ use indexmap::IndexMap;
 use rustc_hash::{FxHashMap, FxHashSet};
 use smallvec::SmallVec;
 use swc_atoms::{atom, Atom};
-use swc_common::{util::take::Take, Mark, Spanned, SyntaxContext, DUMMY_SP};
+use swc_common::{util::take::Take, Mark, Span, Spanned, SyntaxContext, DUMMY_SP};
 use swc_ecma_ast::*;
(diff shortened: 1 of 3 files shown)


magic-akari:
As explained in #12130, this is a long-standing known limitation and the current behavior is intentional. We don't plan to add the additional transform complexity and generated code needed to preserve this runtime error.

I'm closing this PR as not planned. Please see #12130 for the full rationale.

PR #12138, opened 2026-08-22
fix(es/minifier): Don't replace value-used console.*.bind() calls with undefined

striedinger:
**Description:**

With `drop_console`, every call rooted at `console` is replaced with `undefined`, including calls whose result is used:

```js
// input
const err = console.error.bind(console);
err("boom");

// output with drop_console: true
const err = void 0;
err("boom"); // TypeError: err is not a function
```

This breaks real-world code at runtime, e.g. Emscripten prologues (`var err = console.error.bind(console)`) and `@sqlite.org/sqlite-wasm`.

Now `bind`/`toString`/`valueOf` (written as `.bind` or `["bind"]`) invoked directly on a known console method keep their result by replacing only the console method, the same way terser does since terser/terser#323: `console.error.bind(console)` -> `(()=>{}).bind()`. With `ecma` >= 2015 the noop is an arrow, which - like the native console methods - is not a constructor; ES5 keeps `function () {}`.

A `?.` in the callee keeps its per-hop short-circuiting:

- `console.debug?.bind(x)` -> `(console.debug && noop)?.bind()`: a nullish method still yields `undefined`.
- `console?.error.bind(x)` -> `console == null ? void 0 : (console.error && noop).bind()`: a nullish `console` still yields `undefined`, while a nullish `console.error` still throws at the `.bind` access.

Everything else is unchanged and still collapses to `undefined`: direct calls, `.call`/`.apply` (whose results are `undefined` anyway), custom properties on `console` or its methods (`console.state.valueOf(...)`, `console.log.capture(...)`), and deeper chains like `console.a.b.c(...)`.

Also, only the unresolved global `console` is matched now, so a local binding that happens to be named `console` is no longer dropped (matches terser's `is_undeclared_ref` guard).

Notes:

- A chained call in statement position (`console.error.bind(console);`) now leaves `(()=>{}).bind();` instead of being erased, since value usage isn't known at this site and the substitution must be safe in any position.
- Local invariants beyond the documented assumptions, each with a code comment per the AGENTS.md guidance: under ES5, constructing a dropped bound method succeeds instead of throwing (non-constructibility is not expressible in ES5); own overrides of `bind`/`toString`/`valueOf` on a console method are not preserved (names are matched statically, as terser does); and the hoisted `console == null` check reads the global twice, assuming a stable binding rather than a side-effectful accessor.

All scenarios are covered by executing fixture suites (`tests/fixture/issues/drop-console-*`) with checked-in `expected.stdout`.


kdy1:
There's another way to fix this. https://github.com/swc-project/swc/pull/12139
[I asked Codex](https://github.com/swc-project/swc/pull/12139#pullrequestreview-4998668495) how does this compare

kdy1 (approved):
Thank you!

PR #12141, opened 2026-08-22
fix(es/minifier): preserve bindings in copied inline arrows

marcoshernanz:
**Description:**

Cheap arrow functions stored in `Finalizer::lits` can be cloned into multiple sibling scopes while retaining the same resolved binding IDs. A later compression pass can turn each copy into an IIFE with local declarations. With mangling enabled, the shared ID can receive a short name that collides with another binding in one sibling scope, producing invalid JavaScript.

This change freshens binding IDs when `Finalizer` clones an inline literal. It applies at both clone sites (identifier replacement and object-shorthand expansion), which fixes the ownership boundary instead of coupling hygiene to a later IIFE optimization. Non-binding literals keep the existing fast path. The helper also replaces the equivalent existing remapping code for cloned simple callees.

The regression fixture is the exact bundle shape reported by Next.js. Its empty `expected.stdout` intentionally makes the fixture execute the generated code, so the pre-fix duplicate declaration fails with a parse error.

**Validation:**

- `cargo fmt --all`
- `cargo clippy -p swc_ecma_minifier --lib -- -D warnings`
- `cargo test -p swc_ecma_minifier --no-fail-fast`
- `crates/swc_ecma_minifier/scripts/exec.sh` (484 minifier fixtures and 2,128 execution fixtures passed)
- rebuilt the historical SWC native binding consumed by the affected Next.js release; the reporter's Next.js 16.3.1 app builds, starts, and renders correctly
- verified passes 1/2/3, mangle on/off, toplevel on/off, object shorthand, and same-spelling collision variants
- interleaved benchmark over 23 reporter server chunks: invalid outputs 1 → 0; aggregate output 569,415 → 569,403 bytes; candidate time +1.28% and +1.56% in two runs (A/A noise +0.44% and +0.28%)

The workspace-wide all-target clippy command was also attempted; it is currently blocked by unrelated pre-existing unknown-lint errors in `swc_ecma_lints` and `swc_ecma_transforms_react`. The touched minifier crate's library clippy is clean.

**Related issue:**

- https://github.com/vercel/next.js/issues/97517
- Regression became reachable after https://github.com/swc-project/swc/pull/11021


kdy1 (approved):
Thanks! I added some small tests

PR #12158, opened 2026-08-25
fix(es/compat): preserve for-of var binding scope

### LINKED ISSUE #12153: Object-rest transform changes for-of var bindings into block-scoped bindings
### Describe the bug

When lowering object rest for an ES2015 target, SWC transforms a `var` destructuring declaration in a `for-of` head into declarations using `let` inside the loop body.
The bound names of:
```js
for (var { a, ...b } of iterable) {}
```
belong to the surrounding variable environment and must remain accessible after the loop. 
However, moving them into let declarations inside the loop body changes their scope and lifetime.

### Input code

```typescript
for (var { a, ...b } of [{ x: 1, y: 2 }]) {}
console.log(a, b);
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript"
    },
    "minify": {
      "compress": false,
      "mangle": false
    },
    "target": "es2015"
  },
  "minify": false
}
```

https://play.swc.rs/?version=1.16.1&code=H4sIAAAAAAAAA0vLL1LQKEssUqhWSNRR0NPTS1KoVchPU4iuVqiwUjDUUai0UjBSqI3VVKiu5UrOzyvOz0nVy8lP1wCqTtK0BgCoRhSNPwAAAA%3D%3D&config=H4sIAAAAAAAAA1WNvQqAMBCDd5%2BidHZQwcW3OcopFVvLXQdL8d09K%2F5tyReS5EopPbPRg8oixQQgRnq8EE4%2BwiZEo3HAhmyIuoR7fXWc9XZM345ZXSBkFjbCwljfgQM%2FLXjj30oEmjCWH%2B6atj8%2FSvbuX6X9AERNOxe2AAAA

### SWC Info output

_No response_

### Expected behavior

Both `a` and `b` are declared using `var`, so they should remain accessible after the loop.
Expected output:
`undefined {x:1,y:2}`

Using var declaration inside the loop body would also preserve the required variable scope:
```js
...
for (var _ref of [
    {
        x: 1,
        y: 2
    }
]){
    var { a } = _ref, b = _object_without_properties(_ref, [
        "a"
    ]);
}
console.log(a, b);
```

### Actual behavior

```js
function _object_without_properties(source, excluded) {
    if (source == null) return {};
    var target = {}, sourceKeys, key, i;
    if (typeof Reflect !== "undefined" && Reflect.ownKeys) {
        sourceKeys = Reflect.ownKeys(Object(source));
        for(i = 0; i < sourceKeys.length; i++){
            key = sourceKeys[i];
            if (excluded.indexOf(key) >= 0) continue;
            if (!Object.prototype.propertyIsEnumerable.call(source, key)) continue;
            target[key] = source[key];
        }
        return target;
    }
    target = _object_without_properties_loose(source, excluded);
    if (Object.getOwnPropertySymbols) {
        sourceKeys = Object.getOwnPropertySymbols(source);
        for(i = 0; i < sourceKeys.length; i++){
            key = sourceKeys[i];
            if (excluded.indexOf(key) >= 0) continue;
            if (!Object.prototype.propertyIsEnumerable.call(source, key)) continue;
            target[key] = source[key];
        }
    }
    return target;
}
function _object_without_properties_loose(source, excluded) {
    if (source == null) return {};
    var target = {}, sourceKeys = Object.getOwnPropertyNames(source), key, i;
    for(i = 0; i < sourceKeys.length; i++){
        key = sourceKeys[i];
        if (excluded.indexOf(key) >= 0) continue;
        if (!Object.prototype.propertyIsEnumerable.call(source, key)) continue;
        target[key] = source[key];
    }
    return target;
}
for (var _ref of [
    {
        x: 1,
        y: 2
    }
]){
    let { a } = _ref, b = _object_without_properties(_ref, [
        "a"
    ]);
}
console.log(a, b); // reference error
```

### Version

1.16.1

### Additional context

The closest previous fix appears to be [#7437](https://github.com/swc-project/swc/pull/7437). That change used `let` for declarations generated by the object-rest transform in order to preserve per-iteration lexical bindings. However, the transform currently appears to use `VarDeclKind::Let` unconditionally.
The transform likely needs to distinguish the original declaration kind:
- An original `var` declaration must preserve function/global variable scope.
- Original `let` and `const` declarations must preserve per-iteration lexical bindings.

Simply changing all generated declarations back to var would likely regress #6988, so the transformation should branch based on the declaration kind or hoist var bindings and emit assignments inside each iteration.

mameikagou:
**Description:**

The object-rest transform always emitted `let` for declarations moved from a `for-of` head into the loop body. That preserves lexical bindings, but changes an original `var` binding from function scope to block scope and makes it unavailable after the loop.

This preserves the original declaration kind, adds fixture coverage for `var`/`let`/`const`, and adds a runtime regression that checks both post-loop `var` access and per-iteration lexical bindings.

Verified with the focused object-rest fixture and runtime tests, `cargo fmt`, and Clippy for the affected crates.

Fixes #12153

### DIFF
--- a/.changeset/preserve-for-of-var-scope.md
+++ b/.changeset/preserve-for-of-var-scope.md
@@ -1,0 +1,7 @@
+---
+swc_core: patch
+swc_ecma_transformer: patch
+swc_ecma_transforms_compat: patch
+---
+
+fix(es/compat): preserve for-of object-rest binding scope
--- a/crates/swc_ecma_transformer/src/es2018/object_rest_spread.rs
+++ b/crates/swc_ecma_transformer/src/es2018/object_rest_spread.rs
@@ -519,6 +519,10 @@ impl ObjectRestSpreadPass {
 
         match left {
             ForHead::VarDecl(var_decl) => {
+                // Keep the original declaration kind: `var` bindings are
+                // function-scoped, while `let` and `const` need per-iteration
+                // lexical bindings.
+                let kind = var_decl.kind;
                 let ref_ident = private_ident!("_ref");
                 let pat = var_decl.decls[0].name.take();
                 var_decl.decls[0].name = ref_ident.clone().into();
@@ -527,7 +531,7 @@ impl ObjectRestSpreadPass {
                 lowerer.visit(pat, Box::new(ref_ident.into()));
 
                 let stmt: Stmt = VarDecl {
-                    kind: VarDeclKind::Let,
+                    kind,
                     decls: lowerer.out.into_decls(),
                     ..Default::default()
                 }
--- a/crates/swc_ecma_transforms_compat/tests/__swc_snapshots__/tests/es2018_object_rest_spread.rs/rest_for_x.js
+++ b/crates/swc_ecma_transforms_compat/tests/__swc_snapshots__/tests/es2018_object_rest_spread.rs/rest_for_x.js
@@ -1,7 +1,7 @@
 var _ref;
 // ForXStatement
 for (var _ref1 of []){
-    let { a } = _ref1, b = _object_without_properties(_ref1, [
+    var { a } = _ref1, b = _object_without_properties(_ref1, [
         "a"
     ]);
 }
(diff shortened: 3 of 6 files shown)


kdy1 (approved):

kdy1:
Thanks!

PR #12160, opened 2026-08-26
fix(es/minifier): preserve do-while control-flow targets

### LINKED ISSUE #12159: Minifier retargets break and continue when unwrapping do...while(false)
### Describe the bug

The `loops` compressor replaces a `do...while(false)` statement with its body even when that body contains a direct `break` or `continue`.

Those abrupt completions target the `do` loop. Once the loop is removed, they can target an enclosing loop and silently change program behavior. Without an enclosing target, the optimization can instead emit an illegal jump statement.

The bug does not require source labels. In the downstream Kotlin/JS case, SWC first normalizes a labeled jump such as `break inner` to an unlabeled `break` while the loop still exists; a later compression pass then removes that target loop. The unlabeled reproduction below demonstrates the same underlying issue directly.

### Input code

```typescript
function run() {
    const trace = [];

    for (let i = 0; i < 2; i++) {
        do {
            trace.push("body" + i);
            break;
        } while (false);

        trace.push("tail" + i);
    }

    return trace.join(",");
}

console.log(run());
```

### Config

```json
// @swc/core minifySync options
{
    "compress": {
        "dead_code": true,
        "evaluate": true,
        "loops": true,
        "passes": 2
    },
    "mangle": false
}
```

https://github.com/August-Z/swc/tree/aca52cb768faf2f649e0cc076b6a053cebc4b6b1/crates/swc_ecma_minifier/tests/fixture/issues/do-while-false-control-flow

### SWC Info output

Operating System:
    Platform: darwin
    Arch: arm64
    Machine Type: arm64
    Version: Darwin Kernel Version 25.0.0
    CPU: Apple M3 Pro (12 cores)

Binaries:
    Node: 26.3.0
    npm: 11.16.0
    Yarn: 1.22.22
    pnpm: 11.19.0

Relevant Packages:
    @swc/core: 1.16.1 (verified in an isolated npm install)
    @swc/helpers: N/A
    @swc/types: N/A

SWC Config:
    API: minifySync
    .swcrc path: N/A

### Expected behavior

The `break` must continue to target the inner `do...while` loop. Executing the transformed code should print:

```text
body0,tail0,body1,tail1
```

### Actual behavior

The inner loop is removed and its `break` is retargeted to the outer `for` loop. With `@swc/core@1.16.1`, the output is:

```js
function run(){let trace=[];for(let i=0;i<2;i++){trace.push("body"+i);break}return trace.join(",")}console.log(run());
```

It prints only `body0`.

### Version

@swc/core 1.16.1; also reproduced on swc main at 5b8c98d909b4c5db2730877f0be09182ee470a26

### Additional context

This was first observed downstream in web-infra-dev/rspack#15335. Kotlin/JS 1.7.20 emits a labeled single-iteration `do...while(false)` inside a suspend-function state-machine loop. After compression, the inner jump exits the outer state-machine loop, so later code is removed and the function returns `undefined`.

The behavior is independently reproducible with `@swc/core` and does not require Rspack, labels, or mangling. A direct `continue` has the analogous control-target problem. Without an enclosing loop, `function f(){do{break;}while(false);return 42}` can be reduced to invalid code equivalent to `function f(){break}`.

On current `main` (commit `5b8c98d909b4c5db2730877f0be09182ee470a26`, `swc_ecma_minifier 61.0.3`), the affected path is `optimize_loops_with_constant_condition` in `crates/swc_ecma_minifier/src/compress/pure/loops.rs`. It calls `should_not_inline_loop_body(&stmt.body, true)`, allowing the direct jump while removing its target.

Disabling the `loops` compressor option is the narrow reliable workaround.

SWC previously added a conservative guard for the separate `simplify` transform in #1618. This report concerns the minifier compressor's independent constant-loop optimization.

August-Z:
**Description:**

The minifier's `loops` optimization replaces a constant-false `do...while` statement with its body. It previously allowed a direct `break` or `continue` while doing so, even though those statements target the loop being removed.

After unwrapping, the abrupt completion could target a surrounding loop and silently change program behavior, or become an illegal statement when no enclosing target existed.

This change passes `false` for `allow_break_continue` when checking a constant-false `do...while` body. SWC now retains the loop when a direct `break` or `continue` depends on it, while other safe constant-false loop bodies remain eligible for unwrapping.

The regression fixture covers:

- unlabeled and labeled `break` statements nested in an outer loop;
- unlabeled and labeled `continue` statements nested in an outer loop;
- declaration hoisting after an early `break`.

Labels are not required to reproduce the bug. The labeled cases are included because they match the Kotlin/JS output reported in web-infra-dev/rspack#15335; after label normalization they expose the same direct control-flow dependency.

A similar semantic guard exists in the separate `simplify` transform following #1618. This change applies that constraint to the minifier compressor's independent constant-loop optimization.

**Testing:**

- `UPDATE=1 cargo test -p swc_ecma_minifier --test compress --features concurrent do_while_false_control_flow -- --nocapture`
- `cargo test -p swc_ecma_minifier --test compress --features concurrent do_while_false_control_flow -- --nocapture`
- `cargo fmt --all`
- `cargo clippy -p swc_ecma_minifier --all-targets --features concurrent -- -D warnings`

**BREAKING CHANGE:**

None.

Fixes #12159.

Related to web-infra-dev/rspack#15335.

See also #1618.

### DIFF
--- a/.changeset/preserve-do-while-control-flow.md
+++ b/.changeset/preserve-do-while-control-flow.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_minifier: patch
+---
+
+fix(es/minifier): Preserve do-while control-flow targets
--- a/crates/swc_ecma_minifier/src/compress/pure/loops.rs
+++ b/crates/swc_ecma_minifier/src/compress/pure/loops.rs
@@ -264,7 +264,9 @@ impl Pure<'_> {
             Stmt::DoWhile(stmt) => {
                 let val = stmt.test.as_pure_bool(self.expr_ctx);
                 if let Value::Known(false) = val {
-                    if should_not_inline_loop_body(&stmt.body, true) {
+                    // A direct break or continue targets this loop and cannot survive
+                    // unwrapping.
+                    if should_not_inline_loop_body(&stmt.body, false) {
                         return;
                     }
 
--- a/crates/swc_ecma_minifier/tests/fixture/issues/do-while-false-control-flow/config.json
+++ b/crates/swc_ecma_minifier/tests/fixture/issues/do-while-false-control-flow/config.json
@@ -1,0 +1,6 @@
+{
+    "dead_code": true,
+    "evaluate": true,
+    "loops": true,
+    "passes": 2
+}
--- a/crates/swc_ecma_minifier/tests/fixture/issues/do-while-false-control-flow/expected.stdout
+++ b/crates/swc_ecma_minifier/tests/fixture/issues/do-while-false-control-flow/expected.stdout
@@ -1,0 +1,5 @@
+unlabeled break body0,tail0,body1,tail1
+labeled break body0,tail0,body1,tail1
+unlabeled continue body0,tail0,body1,tail1
+labeled continue body0,tail0,body1,tail1
+hoist after break undefined
--- a/crates/swc_ecma_minifier/tests/fixture/issues/do-while-false-control-flow/input.js
+++ b/crates/swc_ecma_minifier/tests/fixture/issues/do-while-false-control-flow/input.js
@@ -1,0 +1,61 @@
+function unlabeledBreak() {
+    var result = [];
+    for (var i = 0; i < 2; i++) {
+        do {
+            result.push("body" + i);
+            break;
+        } while (false);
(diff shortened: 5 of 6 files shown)


kdy1 (approved):
Thank you!

PR #12164, opened 2026-08-27
fix(es/react-compiler): preserve compiled body source maps

LingyuCoder:
**Description:**

React Compiler rebuilt nodes retain Babel `loc` information but often omit `start` and `end`. The reverse AST converter previously turned those nodes into `DUMMY_SP`, so codegen omitted source-map mappings for compiled component and hook bodies.

This change preserves the source file start position across the AST round trip, converts file-relative `loc.index` values back to global SWC `BytePos` values, and lets rebuilt expression statements inherit their retained statement location when the root expression has no span. Recovered spans also allow preserved TypeScript JSX type arguments to be restored.

The source-map fixture preloads another source file to cover non-initial `SourceMap` ranges, then verifies that `Promise.resolve`, `setValue(next)`, and JSX body output resolve to the correct generated line, source file, and original line.

- https://github.com/web-infra-dev/rspack/issues/15350

### DIFF
--- a/crates/swc_ecma_react_compiler/src/lib.rs
+++ b/crates/swc_ecma_react_compiler/src/lib.rs
@@ -97,6 +97,7 @@ pub fn transform(
     let ConvertResult {
         file,
         preserved_ast,
+        source_start,
     } = convert_program(program, source_text, comments);
     let emit_success_error_diagnostics = options.no_emit;
     let result =
@@ -117,7 +118,7 @@ pub fn transform(
 
     let rename_plan = build_rename_plan(&scope_info, &renames);
     let program = file.map(|file: react_compiler_ast::File| {
-        let mut compiled = convert_program_to_swc(&file, preserved_ast);
+        let mut compiled = convert_program_to_swc(&file, preserved_ast, source_start);
         apply_renames(&mut compiled, &rename_plan);
         compiled
     });
--- a/.changeset/preserve-react-compiler-source-maps.md
+++ b/.changeset/preserve-react-compiler-source-maps.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_react_compiler: patch
+---
+
+fix(es/react-compiler): Preserve source maps for compiled function bodies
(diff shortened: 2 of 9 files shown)


crates/swc_ecma_react_compiler/src/lib.rs
```diff
@@ -117,7 +118,7 @@ pub fn transform(
 
     let rename_plan = build_rename_plan(&scope_info, &renames);
     let program = file.map(|file: react_compiler_ast::File| {
-        let mut compiled = convert_program_to_swc(&file, preserved_ast);
+        let mut compiled = convert_program_to_swc(&file, preserved_ast, source_start);
         apply_renames(&mut compiled, &rename_plan);
```
magic-akari, line 122: This also appears to reintroduce the rename collision found in #11979. `apply_renames` still matches identifiers across the entire converted program by `span.lo`, so assigning recovered spans to compiler-generated identifiers can make them collide with original reference positions and be renamed incorrectly.

magic-akari (changes requested):
Thanks for investigating this and for adding the source-map regression coverage.

This seems to duplicate #11979, where the same `loc`-to-`BytePos` approach and its caveats were already discussed. Rebasing with `source_start` does address the `SourceMap`-global versus file-relative coordinate issue raised there, but I still don't think reconstructing SWC spans from `loc.index` is the right fix.

`loc` is Babel-compatible, and its offsets are expected to use JavaScript string indexing, i.e. UTF-16 code units, whereas SWC's `BytePos` is UTF-8 byte-based. The bridge previously put byte offsets into `loc` only because the reverse conversion did not consume them. Reading those values back makes that non-standard representation part of the contract and reintroduces the UTF-8/UTF-16 problem I mentioned in #11979.

The new regression fixture does not cover this distinction either: its input is ASCII-only, and it checks the original line but not the original column.

The root cause is that React Compiler drops the node-level `start` / `end` offsets when rebuilding AST nodes. I think the proper fix is for React Compiler to preserve those offsets, as proposed in react/react#36921, rather than for each downstream tool to reconstruct them from `loc`.

That upstream PR has been waiting for review from the React Compiler team for some time. Could we focus on moving it forward instead? The Rspack reproduction and the source-map regression coverage from this PR should provide useful additional context and coverage there.

LingyuCoder:
Thanks for the suggestion and for pointing me to #11979 and react/react#36921.

Since there is already ongoing work addressing the same root cause, I agree that this PR is redundant and can be closed.

One caveat from verification: the current head of #11979 does not appear to handle a shared `SourceMap` correctly. In the source-map regression based on web-infra-dev/rspack#15350, when the input is not the first `SourceFile`, `Promise.resolve` is mapped to original line 1 instead of line 6.

This may need to be considered as part of #11979, or addressed in a follow-up after the upstream React Compiler fix lands.

PR #12168, opened 2026-08-31
refactor(react-compiler): Use official packages

wbinnssmith:
**Description:**

Replace the unmaintained `forked_react_compiler` packages published by the Oxc team with the official `react_compiler` packages. The Oxc team now maintains its own separate fork.

None.


kdy1 (approved):
Thanks!

Issue #12102, opened 2026-08-05
TypeScript enum transform does not fold namespace member access (`N.foo`, `N.M.foo`)


baltasarblanco:
### Describe the bug

When an enum member is initialized from a constant exported by a `namespace`, swc emits a
reverse mapping instead of folding the value. `tsc` folds it and emits a plain string enum.

For `declare namespace`, the consequence goes beyond a shape difference: the namespace is
erased from the output, but the member access survives in the initializer, so the emitted
module throws a `ReferenceError` when that enum is initialized.

This is not a checker-dependent difference: `tsc`'s own `transpileModule` — single file, no
type resolution, the same constraint swc operates under — already folds the one-level form,
the arbitrarily nested form, and the ambient form.

The result is also inconsistent inside a single enum: in `enum E { A = N.foo, B = "b" }`,
`B` is emitted as a string enum member while `A` is reverse-mapped, so the object ends up
with a spurious `E["s"] = "A"` property.

This is the follow-up to the limitation noted in #12101, which fixes the plain `const`
variable case (#11715) but leaves member access untouched.

### Input code

```typescript
namespace N { export const foo = "s"; }
enum E { A = N.foo, B = "b" }

namespace M { export namespace Inner { export const bar = "t"; } }
enum F { A = M.Inner.bar, B = "b" }

declare namespace D { enum A { X = 1 } }
enum G { P = D.A.X, Q }
```

### Config

```json
{
  "jsc": {
    "parser": { "syntax": "typescript", "tsx": false },
    "target": "es2022",
    "loose": false,
    "minify": { "compress": false, "mangle": false }
  },
  "module": { "type": "es6" },
  "minify": false,
  "isModule": true
}
```

### Expected behavior

`tsc` 5.9.3, `transpileModule`, `target: ES2022`, `module: ESNext` (namespace emit elided):

```js
var E;
(function (E) {
    E["A"] = "s";
    E["B"] = "b";
})(E || (E = {}));

var F;
(function (F) {
    F["A"] = "t";
    F["B"] = "b";
})(F || (F = {}));

var G;
(function (G) {
    G[G["P"] = 1] = "P";
    G[G["Q"] = 2] = "Q";
})(G || (G = {}));
```

Evaluating that output and reading the three objects:

```
{"E":{"A":"s","B":"b"},"F":{"A":"t","B":"b"},"G":{"1":"P","2":"Q","P":1,"Q":2}}
```

### Actual behavior

[...]

PR #12181, opened 2026-09-04
fix(es/typescript): Preserve asserted enum member semantics

### LINKED ISSUE #12150: TypeScript enum transform treats type assertions in enum initializers as transparent, diverging from tsc
**Describe the bug**

`tsc` treats a type assertion in an enum member initializer as removing constness — the member is emitted with a reverse mapping, keeping the asserted expression's runtime shape. SWC strips the assertion and classifies the member as a plain string constant, emitting a string-enum shape with no reverse mapping.

**Input code**

```ts
enum E { A = "s" as string, B = "t" as any, C = "u" as const }
```

**Actual behavior** (`@swc/core` 1.16.1)

```js
var E = /*#__PURE__*/ function(E) {
    E["A"] = "s";
    E["B"] = "t";
    E["C"] = "u";
    return E;
}(E || {});
```

**Expected behavior** (`tsc` 5.9.3, `transpileModule`)

```js
var E;
(function (E) {
    E[E["A"] = "s"] = "A";
    E[E["B"] = "t"] = "B";
    E[E["C"] = "u"] = "C";
})(E || (E = {}));
```

Observable difference: `E["s"]` is `"A"` under `tsc` and `undefined` under SWC.

The syntactic rule — any type syntax on the initializer removes constness — is the one confirmed as intentional in microsoft/TypeScript#63275 and described in evanw/esbuild#4387. Assertions on a *reference* (`enum E { A = foo as string }`) already follow it after #12101; assertions on a *literal* do not.

Note for whoever picks this up: the transparent-assertion behavior was introduced in #11769 for #11761, and `ts_enum_with_type_assertion` pins it — a fix here needs to account for both.

Noted as out of scope in #12101, reporting separately as promised there.

**Version**: `@swc/core` 1.16.1

GtechGovind:
**Description:**

Type assertions make TypeScript enum member initializers syntactically non-constant. Preserve that semantic classification after the assertion wrapper is stripped instead of replacing it with a recomputed literal value.

This makes asserted string initializers emit runtime reverse mappings consistently with `tsc`, while ordinary constant enum members remain unchanged. Regression coverage includes `as string`, `as any`, `as const`, and legacy angle-bracket assertions.

Closes #12150

**Testing:**

- `cargo test -p swc_ecma_transforms_typescript --test strip ts_enum_with_type_assertion -- --exact`
- `cargo test -p swc_ecma_transforms_typescript` (5,290 tests passed)
- `cargo test -p swc --test projects fixture_tests__fixture__issues_11xxx__11761__input -- --exact`


baltasarblanco:
Heads-up so it doesn't surprise you later: I'm starting on #12102, which also lands in `compute_member` — namespace objects rather than assertions, behind the same `const_enum_only` gate yours sits next to. Yours is ahead of mine, so I'll rebase on top of it; nothing needed on your side.

kdy1:
> I'm sorry, but I don't think we need tsc parity here

https://github.com/swc-project/swc/issues/12150#issuecomment-5646418940

PR #12285, opened 2026-09-07
fix(es/minifier): shorten loose undefined comparisons to null

SyMind:
**Description:**

Optimize loose `==` and `!=` comparisons by replacing side-effect-free `undefined` expressions with `null`, saving two bytes per replacement.

```js
// Input
console.log(it == undefined);

// Before
console.log(void 0==it);

// After
console.log(null==it);
```

Applies to either operand when `comparisons` is enabled. Preserves strict comparisons, shadowed `undefined` bindings, side effects, and exceptions.

Adds five regression fixtures and updates affected output snapshots.

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc/tests/tsc-references/parserindenter.2.minified.js
+++ b/crates/swc/tests/tsc-references/parserindenter.2.minified.js
@@ -34,7 +34,7 @@ import { _ as _class_call_check } from "@swc/helpers/_/_class_call_check";
         var commentLastLineNumber = this.snapshot.GetLineNumberFromPosition(token.Span.endPosition());
         if (token.lineNumber() == commentLastLineNumber) return result;
         var commentFirstLineIndentationDelta = this.GetIndentationDelta(token.Span.startPosition(), null);
-        if (void 0 != commentFirstLineIndentationDelta) for(var line = token.lineNumber() + 1; line <= commentLastLineNumber; line++){
+        if (null != commentFirstLineIndentationDelta) for(var line = token.lineNumber() + 1; line <= commentLastLineNumber; line++){
             var lineStartPosition = this.snapshot.GetLineFromLineNumber(line).startPosition(), lineIndent = this.GetLineIndentationForOffset(lineStartPosition), commentIndentationInfo = this.ApplyIndentationDelta2(lineIndent, commentFirstLineIndentationDelta);
             if (null != commentIndentationInfo) {
                 var tokenStartPosition = lineStartPosition + lineIndent.length, commentIndentationEdit = this.GetIndentEdit(commentIndentationInfo, tokenStartPosition, !1);
--- a/crates/swc_ecma_minifier/src/compress/optimize/ops.rs
+++ b/crates/swc_ecma_minifier/src/compress/optimize/ops.rs
@@ -4,11 +4,15 @@ use swc_ecma_utils::{ExprExt, Type, Value};
 use Value::Known;
 
 use super::{BitCtx, Optimizer};
-use crate::{compress::util::negate, util::make_bool};
+use crate::{
+    compress::util::{is_pure_undefined, negate},
+    util::make_bool,
+};
 
 impl Optimizer<'_> {
     ///
     /// - `'12' === `foo` => '12' == 'foo'`
+    /// - `x == undefined` => `x == null` (also for `!=`)
     pub(super) fn optimize_bin_equal(&mut self, e: &mut BinExpr) {
         if !self.options.comparisons {
             return;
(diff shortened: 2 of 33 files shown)


Austaras:
This PR is fine, but as almost all modern websites run on gzip, it wouldn't have any observable effect.

kdy1 (commented):
I checked whether this provides a meaningful runtime optimization in modern browsers. My conclusion is that this is a valid small minification improvement, but there is no demonstrated runtime performance benefit.

For `function f(x) { return x == void 0; }` and the equivalent `x == null`, local Node.js v20.20.2/V8 produced identical four-byte bytecode: `Ldar a0`, `TestUndetectable`, `Return`. Current V8 source also recognizes both forms and routes loose equality through the same `CompareUndetectable()` path, so this example is already equivalent before optimizing JIT compilation:
- https://github.com/v8/v8/blob/main/src/interpreter/bytecode-generator.cc
- https://github.com/v8/v8/blob/main/src/interpreter/bytecode-array-builder.cc#L644-L655

I also compared the eight existing JavaScript snapshots modified by this PR at its base and head revisions. Across 8,095,631 original bytes, the change saves 88 uncompressed bytes (about 0.0011%). Compressing each file separately with gzip level 9 and summing their sizes gives a net saving of 9 bytes; a few individual files become slightly larger after compression.

These size measurements are for the checked-in snapshots, not production bundles. I did not measure Brotli, browser loading time, or Safari/Firefox runtime performance, and the bytecode inspection is not a timing benchmark.

This seems reasonable to merge as a small output-size improvement. I would not describe it as a meaningful browser execution-speed optimization. The PR description already frames it appropriately as saving two bytes per replacement.

kdy1 (commented):
Follow-up: I also measured all 12 libraries in `crates/swc_ecma_minifier/benches/full`. **This corpus shows no output-size improvement: every minified output is byte-for-byte identical before and after this PR.**

Base: `a5bd5924c14e702b995cdbb6581a9ecf4098df5a`

Head: `949ba8d13a5e0e38378f63af94ff49b61f46f174`

Inputs: all 12 JavaScript files in `crates/swc_ecma_minifier/benches/full`. Both revisions were built locally and executed using a copy of the existing `tests/size.rs` pipeline (default compression, top-level mangling, minified code generation, no emitted comments). gzip uses the harness's flate2 `Compression::best()`. Brotli uses Node.js v20.20.2 / Brotli 1.1.0, quality 11, text mode. Each file is compressed independently. All numbers are bytes; delta = head - base, so negative means smaller. This measures output sizes, not execution speed.

| Library | Minified base | Minified head | Δ | gzip base | gzip head | Δ | Brotli base | Brotli head | Δ |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| antd.js | 2,155,682 | 2,155,682 | 0 | 456,133 | 456,133 | 0 | 356,641 | 356,641 | 0 |
| d3.js | 267,525 | 267,525 | 0 | 87,588 | 87,588 | 0 | 73,598 | 73,598 | 0 |
| echarts.js | 1,000,485 | 1,000,485 | 0 | 321,628 | 321,628 | 0 | 259,437 | 259,437 | 0 |
| jquery.js | 89,894 | 89,894 | 0 | 30,927 | 30,927 | 0 | 28,177 | 28,177 | 0 |
| lodash.js | 70,578 | 70,578 | 0 | 25,217 | 25,217 | 0 | 22,379 | 22,379 | 0 |
| moment.js | 58,705 | 58,705 | 0 | 18,687 | 18,687 | 0 | 16,921 | 16,921 | 0 |
| react.js | 23,026 | 23,026 | 0 | 8,260 | 8,260 | 0 | 7,282 | 7,282 | 0 |
| terser.js | 457,348 | 457,348 | 0 | 123,380 | 123,380 | 0 | 106,080 | 106,080 | 0 |
| three.js | 645,600 | 645,600 | 0 | 158,464 | 158,464 | 0 | 128,819 | 128,819 | 0 |
| typescript.js | 3,328,841 | 3,328,841 | 0 | 860,752 | 860,752 | 0 | 643,981 | 643,981 | 0 |
| victory.js | 710,660 | 710,660 | 0 | 157,833 | 157,833 | 0 | 123,106 | 123,106 | 0 |
| vue.js | 116,285 | 116,285 | 0 | 42,804 | 42,804 | 0 | 38,107 | 38,107 | 0 |
| TOTAL | 8,924,629 | 8,924,629 | 0 | 2,291,673 | 2,291,673 | 0 | 1,804,528 | 1,804,528 | 0 |

All 12 minified outputs are byte-for-byte identical between revisions. As a positive control, the same pipeline compiled `console.log(it == undefined);` to `console.log(void 0==it);` (24 bytes) on base and `console.log(null==it);` (22 bytes) on head. This confirms that the compared builds distinguish the PR transformation. No output-size benefit was observed for this library corpus.

PR #12287, opened 2026-09-07
mod(es/minifier): Check Map/Set param is array

Austaras:
**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_utils/src/lib.rs
+++ b/crates/swc_ecma_utils/src/lib.rs
@@ -737,8 +737,9 @@ pub trait ExprExt {
     }
 
     #[inline(always)]
+    #[deprecated(note = "please use `is_array` in `Expr` directly")]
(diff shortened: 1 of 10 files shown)


crates/swc_ecma_utils/src/lib.rs
```diff
@@ -737,8 +737,8 @@ pub trait ExprExt {
     }
 
     #[inline(always)]
-    fn is_array_lit(&self) -> bool {
```
kdy1, line 740: Can we preserve the previous method? I don't want to make this PR a breaking change

Austaras: Ok.

kdy1 (dismissed):
Thank you!

@claude Run $add-changset

kdy1:
@codex Add changeset

kdy1 (approved):

PR #12293, opened 2026-09-08
refactor(es/minifier): simplify constant-false loop guards

August-Z:
**Description:**

The constant-false `do...while` guard has an `allow_break_continue` parameter whose value is `false` at every reachable call. Remove the unused parameter and simplify the recursive checks while preserving the guard's existing acceptance and rejection behavior.

Add regression coverage for control-flow targets, hoisted bindings, condition side effects and exceptions, directives, parent rewrites, and compression option boundaries. These fixtures protect the existing conservative behavior when later transformations revisit the loop or its ancestors.

The new opt-in Script checks execute the original source and each of two successive minifications against independent expectations, using the same compression and mangling options. They check stdout or typed completion values, with dedicated cases validating distinctions such as `undefined` versus a string and positive versus negative zero.

**Validation:**

Local validation on macOS aarch64 with Node v26.3.0:

- `cargo fmt --all`
- `cargo test -p swc_ecma_minifier`: 7,977 passed, 48 ignored.
- `cargo test -p swc_ecma_minifier --test compress`: independently rerun; 3,074 passed, 27 ignored.
- `cargo test -p swc_ecma_minifier --features concurrent,debug --test compress do_while_false`: 58 passed.
- `(cd crates/swc_ecma_minifier && ./scripts/exec.sh)`: 487 exec tests and 2,128 terser_exec tests passed.
- `cargo clippy -p swc_ecma_minifier --all-targets -- -D warnings`

The code and tests are identical to those used for these checks. The local `cargo clippy --all --all-targets -- -D warnings` run failed on macOS because of the unchanged `noexec_mount_in` dead-code warning in `crates/swc_native_addon/src/platform/unix.rs:117`. Its production caller is Linux-only. GitHub Actions has independently passed the [CI run](https://github.com/swc-project/swc/actions/runs/34212998202) for the current PR head, including the minifier crate tests on the PR merge commit and the configured Linux, macOS, Windows, and Wasm checks.

The [CodSpeed report for the current PR head](https://github.com/swc-project/swc/pull/12293#issuecomment-5583250640) reports no significant performance changes across 200 compared benchmarks in CPU Simulation. It uses fallback main commit `d602f58`; the 61 skipped entries are historical benchmark identifiers already unregistered at that baseline. The current SWC and minifier benchmark targets completed their measurements and uploaded their data.

Follow-up to #12160.


kdy1 (approved):
Sounds good, thanks!

kdy1:
I may invert the test fixture changes in the future, but I'll do that as a separate PR

August-Z:
Sounds good, thanks for reviewing!

PR #12336, opened 2026-09-09
fix(es/minifier): Check class side effect

### LINKED ISSUE #12311: es/minifier: sequence sinking crosses effectful class evaluation
### Describe the bug

is_skippable_for_seq returns may_have_side_effects directly for a class expression. The polarity is inverted: an effectful class is declared safe to cross. This lets the earlier foo() initializer sink past a static bar() call.

### Input code

```js
function f(){let a=foo();return [class {static x=bar()},a]}function foo(){console.log("foo");return 1}function bar(){console.log("bar")}f()
```

### Config

Options used by the exact-source reproduction:

```json
{
  "compress": {
    "defaults": false,
    "collapse_vars": true,
    "sequences": true,
    "unused": true
  },
  "mangle": false,
  "module": false
}
```

[Portable input and settings in SWC Playground](https://play.swc.rs/?version=1.16.2&code=H4sIAAAAAAAC%2F1WNQQqAIBBFrzK4UoigdXiSaDGJhiAO6AiBePeshdH28d7%2FrkTDniI4qWqwDKgdkVRrslxShM0EzBlqZmRv4NIHJqnahHtzI32CaihmCnYOdErRkRgby6e%2B9V%2FtSKjW729g38TgiwAAAA%3D%3D&config=H4sIAAAAAAAC%2F02NXQrDMAyDr1L8vBPsDjvDMIlaMvK32BktpXdfElrom%2FRZsnb6iKHntFPmIihDyhaV1yYJJrCY4rLS8ZhIuSzQcZCIVamx4KKbt9EzKeQCkWEsZq5eu5nZC1rUJO85C96%2FttW4ltqx4FsRDW6oxiqwp%2B%2FLgePicf06jvvw2XHySrbeQn8RWMHU3AAAAA%3D%3D)

### SWC Info output

Source build of `swc_ecma_minifier 61.0.5`, commit `246e6156ca14eaa07243ac8d9aa994c91c95cf3d`, macOS arm64, Rust `1.96.0-nightly (f5eca4fcf 2026-04-09)`, Node `v20.20.2`.

The observed checkout and freshly fetched authoritative `origin/main` are the same commit. Fetch: `git fetch --no-tags origin main`, `2026-09-09T03:15:52Z`, repository `swc-project/swc`. Reproduction was performed in a separate checkout of that fetched commit. Thus the cause remains on the default branch, not only in an older release.

For minify cases, the source harness follows the repository example: parse, resolve bindings, convert Terser compressor options, call optimize, apply fixer, emit, and execute original/output separately. Script-only behavior uses Script parsing. The playground carries portable source/settings; its published version is not evidence that the release matches the pinned source build. No implementation change was applied.

### Expected behavior

Exit 0; stdout:

```text
foo
bar
```

### Actual behavior

```js
function f(){return[class{static x=bar()},foo()]}function foo(){return console.log("foo"),1}function bar(){console.log("bar")}f();
```

Exit 0; stdout:

```text
bar
foo
```

### Version

`swc_ecma_minifier 61.0.5`, `246e6156ca14eaa07243ac8d9aa994c91c95cf3d` (also freshly fetched `main`).

### Additional context

Responsible code: [compress/optimize/sequences.rs:1354](https://github.com/swc-project/swc/blob/246e6156ca14eaa07243ac8d9aa994c91c95cf3d/crates/swc_ecma_minifier/src/compress/optimize/sequences.rs#L1354).

#### Proposed Scope

Correct the class skippability predicate so only an effect-free class evaluation may be crossed. Preserve ordering for computed keys, static fields, and static blocks; retain safe sinking across classes whose evaluation has no effects.

Limit the change to this semantic boundary and its regression coverage. No API expansion, dependency update, migration, deployment, or feature-flag rollout is required. Preserve the documented minifier assumptions; the reproduction does not require overriding builtin implementations.

#### Acceptance Criteria and Test Scenarios

- The primary input must log foo before bar.
- Cover a computed class key and a static block with the same order-sensitive arrangement.
- A class with only unevaluated instance methods must not introduce extra calls or block otherwise safe behavior.
- Add regression coverage in the existing SWC-owned fixture/execution suites under `crates/swc_ecma_minifier/tests/fixture/issues`, or the existing `tests/eval.rs` API harness for the Evaluator case. Check exact execution/return values and valid generated syntax, not only snapshots.
- Initialize submodules, update fixture expectations only as needed, and rerun `cargo test -p swc_ecma_minifier` without UPDATE. Preserve passing controls above. Keep regression comments in English.

#### Duplicate search

No same-root open duplicate was found in bounded searches including `"class" "evaluation order"` and related minifier terms. Similar issues were evaluated by root cause and transformation boundary.

#### Out of Scope

Unrelated rewrites, new user options, implementing other audit findings, and changing semantic assumptions.

---

_This is a message for readers, not the author of this issue._

Please read [no +1](https://github.com/golang/go/wiki/NoPlusOne) before leaving a comment.

Austaras:
**Description:**

**BREAKING CHANGE:**

Closes #12311, Closes #12317


crates/swc_ecma_minifier/src/compress/optimize/sequences.rs
```diff
@@ -1351,7 +1351,7 @@ impl Optimizer<'_> {
 
             Expr::Update(..) => false,
             Expr::SuperProp(..) => false,
-            Expr::Class(_) => e.may_have_side_effects(self.ctx.expr_ctx),
+            Expr::Class(_) => !e.may_have_side_effects(self.ctx.expr_ctx),
```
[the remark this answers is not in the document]
Austaras: This is fine since accessor is a ts syntax. Minifier should not accept ts

crates/swc_ecma_minifier/tests/fixture/issues/9460/auto-accessor/input.js
kdy1, line 1: Please restore these test cases. We need to support these because minfier can be invoked separately

Austaras: We don't. Just as we have minmial support for ts and jsx in minifier that it merely doesn't panic on such input, we should not consider anything decorator related in minifier because no runtime support it.

kdy1 (dismissed):
Good point, thanks!

I'll merge this after adding the changeset

PR #12343, opened 2026-09-10
refactor(es/minifier): Remove useless helpers

Austaras:
**Description:**

**BREAKING CHANGE:**

### DIFF
--- a/crates/swc_ecma_minifier/src/compress/pure/dead_code.rs
+++ b/crates/swc_ecma_minifier/src/compress/pure/dead_code.rs
@@ -1,7 +1,7 @@
 use par_iter::prelude::*;
 use swc_common::{util::take::Take, EqIgnoreSpan, Spanned, DUMMY_SP};
 use swc_ecma_ast::*;
-use swc_ecma_utils::{extract_var_ids, ExprCtx, ExprExt, StmtExt, StmtLike, Value};
+use swc_ecma_utils::{extract_var_ids, ExprCtx, ExprExt, StmtExt, StmtLike, Type, Value};
 use swc_ecma_visit::{noop_visit_type, Visit, VisitWith};
 
 use super::Pure;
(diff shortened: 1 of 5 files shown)


Austaras:
Let's see if performance has any changes.

crates/swc_ecma_utils/src/lib.rs
```diff
@@ -3346,9 +3346,11 @@ fn get_type(expr: &Expr, ctx: ExprCtx) -> Value<Type> {
...
         }) if &**length == "length" => match &**obj {
             Expr::Array(ArrayLit { .. }) | Expr::Lit(Lit::Str(..)) => Known(Type::Num),
-            Expr::Ident(Ident { sym: arguments, .. }) if &**arguments == "arguments" => {
-                Known(Type::Num)
-            }
+            Expr::Ident(Ident {
+                sym: arguments,
+                ctxt,
+                ..
+            }) if &**arguments == "arguments" && *ctxt == ctx.unresolved_ctxt => Known(Type::Num),
```
kdy1, line 3353: [P1] Do not treat implicit `arguments.length` as permanently numeric

An Arguments object's own `length` property is writable. In a non-strict function, `arguments.length = null; ({} = arguments.length)` must throw `TypeError`; this type proof makes the new empty-object-pattern simplifiers replace the assignment with its RHS and suppress that exception. Return `Unknown` here unless mutation is tracked.

crates/swc_ecma_minifier/src/compress/pure/dead_code.rs
```diff
@@ -96,7 +40,11 @@ impl Pure<'_> {
                 ..
             }) if match &*left {
                 AssignTargetPat::Object(obj) => {
-                    obj.props.is_empty() && is_definitely_non_nullish(right)
+                    obj.props.is_empty()
+                        && !matches!(
+                            right.get_type(self.expr_ctx),
```
kdy1, line 45: [P2] Preserve the syntactic non-nullish proof for empty patterns

`get_type` is deliberately less complete than the removed predicate: for example, it returns `Unknown` for an arrow expression, so `({} = () => {})` is no longer eliminated even though the RHS cannot be nullish. Classes, BigInts, and mixed-type logical branches have the same regression. This leaves avoidable bytes in minified output (also mirrored by the branch remover); retain an equivalent proof or extend the analysis before using it here.

crates/swc_ecma_minifier/src/compress/pure/misc.rs
```diff
@@ -2339,8 +2301,10 @@ impl Pure<'_> {
                         .and_then(|arg| arg.first())
                         .map(|arg| {
                             arg.spread.is_none()
-                                && (eval_to_nullish(self.expr_ctx, &arg.expr)
-                                    || is_valid_map_set_init(&arg.expr, self.expr_ctx, callee))
+                                && (matches!(
+                                    &arg.expr.get_type(self.expr_ctx,),
```
kdy1, line 2305: [P2] Keep recognizing conditionals whose branches are both nullish

For `new Map(flag ? null : void 0)`, the removed helper proved that the first argument is nullish and could drop the unused construction while preserving `flag`. `get_type` merges `Null` and `Undefined` to `Unknown`, so this condition no longer fires. Keep the nullish-result predicate for this path (the same applies to `Set`).

crates/swc_ecma_minifier/src/compress/pure/dead_code.rs
kdy1, line 45: [P1] Do not rely on unresolved-global type proofs in a direct-eval scope

Direct eval can introduce a function-local binding that the resolver cannot see. For example, after `eval("var Infinity = null")`, `({} = Infinity)` must throw, but `get_type` still reports the unresolved reference as numeric and this rewrite removes the exception. Keep the former syntactic conservatism, or guard this optimization in eval-containing scopes.

crates/swc_ecma_transforms_optimization/src/simplify/branch/mod.rs
```diff
@@ -189,7 +138,11 @@ impl VisitMut for Remover {
                 ..
             }) if match &*left {
                 AssignTargetPat::Object(obj) => {
-                    obj.props.is_empty() && is_definitely_non_nullish(right)
+                    obj.props.is_empty()
+                        && !matches!(
+                            right.get_type(self.expr_ctx),
```
kdy1, line 143: [P1] Skip type-based destructuring removal inside `with` bodies

`Remover` currently descends into a `with` body, where the resolver leaves `Infinity` unresolved even though it can be dynamically shadowed. With `with ({ Infinity: null }) { ({} = Infinity) }`, this condition rewrites away the required `TypeError`. Other simplifier passes visit only the `with` object; apply the same boundary here or avoid this inference in dynamic scope.

kdy1 (commented):
Review Full results for base `32c9894cff0ce90cd757566b06d7f6e93c9640de` through head `102148daa81d1c466156df0f19489339c9ff34cc`.

- Baseline reviews: 3 successful
- Follow-up passes: 4 completed
- Reviewer retirements: all 3 (reviewers 1 and 2 after pass 3; reviewer 3 after pass 4)
- Worker attempts: 3; retries: 0
- Findings: 5 total (P1 × 3, P2 × 2)

Material verification limitation: reviewers inspected clean pinned checkouts and ran the required CLI review, but no project test suite was run. The CLI used its configured default model and reasoning settings.

crates/swc_ecma_utils/src/lib.rs
kdy1: I don't think this worths a fix

crates/swc_ecma_minifier/src/compress/pure/dead_code.rs
Austaras: I would fix this later

crates/swc_ecma_minifier/src/compress/pure/misc.rs
Austaras: I don't believe anybody would write code like this, and such rare hardly deserves a special handle.

kdy1 (approved):
Thank you! I'll add some AGENTS.md rules to prevent pedantic reviews like the previous one

PR #12345, opened 2026-09-11
fix(es/minifier): Preserve let initialization when dropping unused variables

CPunisher:
**Description:**

Fix a temporal dead zone (TDZ) violation introduced when `unused` removes a declarator from a `let` declaration. A retained binding without an initializer must be initialized before side effects from later declarators can access it.

**Minimal reproduction**

Compressor options:

```json
{
    "defaults": false,
    "unused": true
}
```

Input:

```js
function test(read, props) {
    let localCountry,
        unused = (localCountry = read()),
        { value } = props;
    return [localCountry, value];
}

console.log(JSON.stringify(test(() => "CN", { value: 1 })));
```

The input prints `["CN",1]`. Before this fix, the compressor produces:

```js
function test(read, props) {
    localCountry = read();
    let localCountry, { value } = props;
    return [localCountry, value];
}
```

Calling this function throws `ReferenceError: Cannot access 'localCountry' before initialization`.

This reproduction requires the options above. The `compress` example currently uses default compressor options, which optimize this input differently and do not reproduce the error.

**Cause and fix**

1. The original declaration initializes its bindings from left to right. Although `localCountry` has no initializer expression, evaluating its declarator initializes the binding to `undefined`.
2. `unused` removes the unused binding but retains the assignment in its initializer as a side effect.
3. `visit_mut_var_declarators` previously skipped retained declarators without initializers without clearing `can_prepend`.
4. On reaching the destructuring declarator, it therefore moved the collected assignment before the entire `let` declaration, placing the assignment inside `localCountry`'s TDZ.

Clear `can_prepend` when a retained, initializer-free declarator belongs to a `let` declaration. The side effect stays after that binding's initialization. Existing handling of `var`, which is already initialized before execution reaches the declaration, remains applicable.

The fixed output is:

```js
function test(read, props) {
    let localCountry, { value } = (localCountry = read(), props);
    return [localCountry, value];
}
```

Adds one fixture covering both a string result and an `undefined` result from `read()`, plus patch changesets for `swc_ecma_minifier` and `swc_core`.

**Validation:**

- Verified that the fixture fails before the fix and passes with the fix.
- Executed the input and generated output separately with Node.js and compared their stdout.
- `cargo test -p swc_ecma_minifier --test compress unused --quiet` — 155 tests passed.
- `cargo fmt -p swc_ecma_minifier -- --check`
- `cargo clippy -p swc_ecma_minifier --lib -- -D warnings`

None.


CPunisher:
A more accurate way can be checking whether the variable that comes later references the variable that comes before it

kdy1 (commented):
Review Full completed for base `32c9894cff0ce90cd757566b06d7f6e93c9640de` through head `2894c3fcf7236f980f460f9f88b7df0fd95ae663`.

- Successful baseline reviews: 3/3.
- Completed follow-up passes: 3; all three reviewers retired after three consecutive passes with no new validated root cause.
- Worker attempts: 3; retry attempts: 0.
- Findings: no reportable issues found in the reviewed scope. This does not prove correctness or security.
- Verification and limitations: three independent read-only `codex review --base 32c9894c` runs, source/diff/fixture-contract inspection, and one targeted `swc_ecma_minifier` fixture test. The full Rust fixture suite was not run. The CLI used its configured model and reasoning defaults; no model override was supplied.

kdy1 (approved):

PR #12347, opened 2026-09-11
fix(es): Decode template raw values when cooked is missing

### LINKED ISSUE #12346: Template literal lowering changes the string value when cooked is absent
## Description

Lowering an ordinary template literal with `cooked: None` changes its string value when the raw text contains an escape sequence.

For a newline escape, the expected result is one newline character (byte `10`). The actual result is a literal backslash followed by `n` (bytes `92, 110`).

## Minimal Rust reproduction

```rust
use swc_common::DUMMY_SP;
use swc_ecma_ast::{Expr, ExprStmt, Lit, Pass, Program, Script, Stmt, Tpl, TplElement};
use swc_ecma_compat_es2015::template_literal;

fn main() {
    // Equivalent to the ordinary JavaScript template `\n`, with cooked omitted.
    let template = Expr::Tpl(Tpl {
        span: DUMMY_SP,
        exprs: vec![],
        quasis: vec![TplElement {
            span: DUMMY_SP,
            tail: true,
            cooked: None,
            raw: r"\n".into(),
        }],
    });
    let mut program = Program::Script(Script {
        body: vec![Stmt::Expr(ExprStmt {
            span: DUMMY_SP,
            expr: Box::new(template),
        })],
        ..Default::default()
    });

    template_literal(Default::default()).process(&mut program);

    let Program::Script(script) = program else {
        panic!("expected script")
    };
    let Stmt::Expr(stmt) = &script.body[0] else {
        panic!("expected expression")
    };
    let Expr::Lit(Lit::Str(value)) = &*stmt.expr else {
        panic!("expected string literal")
    };
    let actual = value.value.to_string_lossy();
    println!("Expected: {:?}, bytes: {:?}", "\n", b"\n");
    println!("Actual:   {:?}, bytes: {:?}", actual, actual.as_bytes());
    assert_eq!(actual, "\n", "template lowering changed the string value");
}
```

## Expected and actual behavior

The assertion should pass: the resulting string should contain one newline character.

Instead, it fails with:

```text
Expected: "\n", bytes: [10]
Actual:   "\\n", bytes: [92, 110]

assertion `left == right` failed: template lowering changed the string value
  left: "\\n"
 right: "\n"
```

## Original issue in a styled-components build

This issue was observed in a styled-components build with the styled-components SWC plugin's CSS minification enabled, followed by SWC template literal lowering. The Rust example above is a minimal reproduction of that issue. The affected code contains a template like this:

```js
import styled from 'styled-components';

const Button = styled.button`
  &::before {
    content: '\\2b';
  }
`;
```

The pseudo-element should display `+`. In the affected build, it displays the literal text `\2b` instead. The CSS value received at runtime contains two backslashes instead of one.

magic-akari:
**Description:**

Use `Str::from_tpl_raw` when `cooked` is missing in template literal lowering and Fast DTS, preserving escape sequences' string values. This fixes doubled backslashes in CSS produced by the styled-components plugin.

Also correct line-ending normalization and Unicode escape handling in the decoder, with regression tests for lowering and declaration generation.

- Closes #12346.

### DIFF
--- a/crates/swc_typescript/src/fast_dts/types.rs
+++ b/crates/swc_typescript/src/fast_dts/types.rs
@@ -13,7 +13,7 @@ use super::{
     inferrer::ReturnTypeInferrer,
     type_ann,
     util::{
-        ast_ext::{ExprExit, PatExt, StaticProp},
+        ast_ext::{tpl_element_value, ExprExit, PatExt, StaticProp},
         types::{ts_keyword_type, ts_lit_type},
     },
     FastDts,
@@ -401,25 +401,9 @@ impl FastDts {
                     .into();
                 }
 
-                if let Expr::Tpl(Tpl {
-                    span,
-                    exprs,
-                    quasis,
-                }) = computed.expr.as_ref()
-                {
-                    if exprs.is_empty() {
-                        let str_prop = quasis
-                            .first()
-                            .and_then(|el| el.cooked.as_ref())
-                            .unwrap()
-                            .clone();
-
-                        let str_prop = Str {
-                            span: *span,
-                            value: str_prop,
-                            raw: None,
-                        };
-
+                if let Expr::Tpl(tpl) = computed.expr.as_ref() {
+                    if let Some(mut str_prop) = self.tpl_to_string(tpl) {
+                        str_prop.span = tpl.span;
                         return str_prop.into();
                     }
                 }
@@ -483,10 +467,7 @@ impl FastDts {
 
         tpl.quasis.first().map(|element| Str {
             span: DUMMY_SP,
-            value: element
-                .cooked
-                .clone()
-                .unwrap_or_else(|| element.raw.clone().into()),
+            value: tpl_element_value(element).into_owned(),
             raw: None,
         })
     }
--- a/.changeset/fast-dts-template-missing-cooked.md
+++ b/.changeset/fast-dts-template-missing-cooked.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_typescript: patch
+---
+
+fix(typescript): Decode raw-only templates in Fast DTS values, computed property names, accessor matching, and enum evaluation.
--- a/.changeset/template-literal-missing-cooked.md
+++ b/.changeset/template-literal-missing-cooked.md
@@ -1,0 +1,7 @@
+---
+swc_core: patch
+swc_ecma_compat_es2015: patch
+swc_ecma_transforms_compat: patch
+---
+
+fix(es/compat): Decode raw escapes when lowering ordinary templates whose cooked values are missing.
--- a/.changeset/template-raw-decoding-boundaries.md
+++ b/.changeset/template-raw-decoding-boundaries.md
@@ -1,0 +1,6 @@
+---
+swc_core: patch
+swc_ecma_ast: patch
+---
+
+fix(es/ast): Preserve line endings, Unicode escapes, and unpaired surrogates when decoding template raw values.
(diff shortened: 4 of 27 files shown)


crates/swc_typescript/src/fast_dts/types.rs
```diff
@@ -486,7 +486,7 @@ impl FastDts {
             value: element
                 .cooked
                 .clone()
-                .unwrap_or_else(|| element.raw.clone().into()),
+                .unwrap_or_else(|| Str::from_tpl_raw(element)),
```
kdy1: [P1] Decode raw-only computed template keys

The new fallback only protects `tpl_to_string`. An expression-free computed template key takes `transform_property_name_to_expr` instead, where lines 404–415 still unwrap `cooked`. If a transform supplies the raw-only template AST this PR now supports, e.g. `export const x = { [`\x41`]: 1 } as const;`, Fast DTS panics instead of emitting a declaration. Decode that quasi with `Str::from_tpl_raw` too and cover a computed-key fixture.

kdy1 (commented):
Review Full completed for base `32c9894cff0ce90cd757566b06d7f6e93c9640de` and head `a53eb7b42ebe567734d922a2583374f0b153f337`.

Three independent baseline reviews completed successfully. Four follow-up passes retired all reviewers (13 worker attempts, 0 retries).

Findings:

- [P1] Raw-only computed template property keys still panic in Fast DTS (inline comment below).
- [P1] [Enum template evaluation still requires `cooked`](https://github.com/swc-project/swc/blob/a53eb7b42ebe567734d922a2583374f0b153f337/crates/swc_typescript/src/fast_dts/enum.rs#L101): `transform_enum` returns no value for a raw-only quasi at lines 101/107, so valid template enum initializers/interpolations emit an enum-initializer diagnostic and lose their initializer.
- [P1] [Raw-only template member keys cannot resolve enum members](https://github.com/swc-project/swc/blob/a53eb7b42ebe567734d922a2583374f0b153f337/crates/swc_typescript/src/fast_dts/util/ast_ext.rs#L247): static-name lookup returns `None` for a computed template key whose `cooked` value is absent, so a raw-only template member reference fails enum evaluation.

Verification: `cargo test -p swc_ecma_ast --test lit` passed (20 tests); `cargo test -p swc_typescript --test typescript` passed (66 tests); the new missing-cooked ES2015 fixtures passed. The full ES2015 template-literals target could not complete 13 unrelated exec tests because `mocha` was unavailable.

This review found reportable issues in the reviewed scope; it does not establish overall correctness or security.

crates/swc_typescript/src/fast_dts/types.rs
magic-akari: Fixed.
Computed template keys now reuse `tpl_to_string`, which decodes `raw` when `cooked` is missing. 
Added a fixture that reproduced the panic and now verifies identical declaration output with and without cooked values.

kdy1:
I'll remove the noise with https://github.com/swc-project/swc/pull/12350

kdy1 (commented):
## Review Full

Reviewed base `f37ce705d2dce729fe4448dded8ce9a1f700ad3c` against head `5a6b883380b744324559b9d75ea3107fd030722f`.

- Three of three baseline reviews completed successfully.
- Five follow-up passes completed; all three reviewers retired after the required three consecutive no-new passes.
- Total worker attempts: 14; retry attempts: 0.
- No reportable issues were found in the reviewed scope.

### Verification limitations

Static review and targeted fixture/compile checks were completed. Full runtime validation was limited by the environment: some test/build attempts were blocked by read-only `sccache`/target creation, and broader compatibility execution tests could not run because `mocha` is unavailable. This review does not establish that the PR is otherwise correct or bug-free.

kdy1 (commented):
Sorry for the late review. I was struggling to verify if it's not a major bump, because of the trait api changes. Changesets are good, so I'll review this soon

kdy1 (approved):

PR #12348, opened 2026-09-11
perf(es/transforms): Skip script traversal in export-default-from

### LINKED ISSUE #9703: Optimize passes like `export_default_from` to use iteration instead of visitor

pederbe:
**Description:**

**Speeds up `export_default_from` on scripts by eliminating unnecessary AST traversal.** Local CodSpeed simulation and walltime benchmarks both confirmed the improvement for 16- and 1,024-statement scripts; all four module benchmarks were classified as unchanged.

The transform now implements `Pass` directly, iterating over the module body and returning immediately for scripts. Existing export rewriting and namespace-pass composition are preserved.

Add export coverage, a CodSpeed benchmark for scripts and modules, and a patch changeset.

**Testing:**

- Linux: `cargo test -p swc_ecma_transforms_proposal` (739 passed, 9 ignored).
- `cargo fmt --all`.
- Linux: `cargo clippy --all --all-targets -- -D warnings`.
- Local CodSpeed simulation and walltime comparisons: both script sizes improved; all four module cases classified as unchanged.

Closes #9703.


kdy1 (commented):
Review Full completed against base `32c9894cff0ce90cd757566b06d7f6e93c9640de` and head `e7f58fbaa31f89eac29b9a4168edef484c29b445`.

Three independent baseline reviews completed successfully. Three follow-up passes then completed; reviewers 1, 2, and 3 each retired after three consecutive passes without a newly validated finding. Worker attempts: 3; retry attempts: 0.

No reportable PR-introduced issues were found in the reviewed scope. This does not establish correctness or security.

Verification included clean pinned checkouts and a clean diff. Follow-up reviewers also reported successful targeted compilation of the proposal benchmark and export test target (14 tests). Material limitations: no full workspace test suite was run; one baseline review could not run its targeted test in the read-only review environment, and the other baseline reviews did not independently run tests.

kdy1 (approved):
Thanks! I was thinking if this worths a dedicated benchmark, but I concluded it's good because many `export default from`s are common in barren files

PR #12353, opened 2026-09-13
fix(es/parser): Respect line breaks after `declare`

Dev-next-gen:
**Description:**

I noticed that SWC treats `declare` as a modifier even when a line break separates it from the declaration that follows. TypeScript applies ASI there, so this input:

```ts
declare
class C {}
```

is the expression statement `declare;` followed by an ordinary class. SWC parsed it as an ambient class instead, which the TypeScript strip then removes from the output, so `C` silently disappears at runtime. With `function` it fails outright: `declare\nfunction f() {}` is rejected with "An implementation cannot be declared in ambient contexts" although it is valid TypeScript.

The fix returns early from the `declare` arm of `parse_ts_expr_stmt` when the next token is on a new line, which is what the `module` arm of `parse_ts_decl` already does (`declare\nmodule M {}` was the one case that parsed correctly). Babel's `tsTryParseDeclare` has the same `isLineTerminator()` check. I left Flow alone.

I compared SWC's statement kinds against `ts.createSourceFile` (TypeScript 6.0.3) for `declare` followed by a newline and `class`, `function`, `const`, `let`, `var`, `enum`, `const enum`, `namespace`, `module`, `type`, `interface`, `abstract class`, `global` and a multiline comment. Before the change only `module` matched, and after it all of them do. `declare enum E {}` and the other same-line forms are unchanged, and so is the `export declare` path.

On whether this is safe: the only inputs that change are the ones where a line break follows `declare`, and for those the old AST contradicted what `tsc` produces. The repository already carries TypeScript's own conformance test `tests/tsc/asiPreventsParsingAsInterface04.ts`, whose comments say `declare`, `interface` and `I` should each be identifiers followed by a block, but its snapshot recorded an ambient interface. I regenerated that snapshot, which now matches those comments. I also added a `declare-line-break` fixture. It fails without the change and passes with it.

`cargo test -p swc_ecma_parser --test typescript` passes apart from `type_only_import_specifier/invalid_type_only`, which fails the same way on `main` in my environment.

Same family as #12341 (line breaks before enum names).

AI tools used


Dev-next-gen:
The red `Test - swc` jobs were my fault. The swc crate keeps its own snapshots of `asiPreventsParsingAsInterface04.ts` under `crates/swc/tests/tsc-references`, and I had only regenerated the parser one. With `declare` no longer swallowing the next line, `interface` is parsed as an expression and those snapshots gain an "Expression expected" diagnostic on line 5. I pushed f2358558a1 with the updated `.1.normal.js` and `.2.minified.js`, and `cargo test -p swc --test tsc` now passes locally (4579 passed, 0 failed).

The Benchmark failures are unrelated. `Benchmark swc_ecma_lints` died on `curl: (35) OpenSSL SSL_connect: Connection reset by peer` while downloading the CodSpeed runner from release-assets.githubusercontent.com, and the other benchmark jobs were cancelled as a consequence.

kdy1 (approved):

PR #12358, opened 2026-09-13
fix(es/parser): allow escaped keywords as property names

magic-akari:
**Description:**

Accept escaped reserved words in property names and import/export names, such as `({ br\u0065ak: 42 }).br\u0065ak`. The lexer previously rejected these valid names before the parser could determine their role.

Preserve decoded keyword names and defer escape validation until parsing distinguishes names from keywords and bindings. Continue rejecting escaped keywords in shorthand properties, imports without aliases, and keyword syntax, including `static` modifiers and static blocks.


kdy1 (approved):
Thank you!

PR #12368, opened 2026-09-14
fix(es/parser): recognize named optional and rest tuple elements

magic-akari:
**Description:**

Accept TypeScript tuples with unnamed optional elements followed by named optional or rest elements, such as `[number?, value?: string]` and `[number?, ...rest: string[]]`. Previously, named optional and rest elements were mistaken for required elements and rejected.


crates/swc_ecma_lexer/tests/typescript.rs
```diff
@@ -0,0 +1,58 @@
...
+use swc_common::{sync::Lrc, SourceFile, SourceMap};
+use swc_ecma_lexer::{error::SyntaxError, lexer, Lexer, Parser, StringInput, Syntax};
+
+fn load_file(path: &Path) -> Lrc<SourceFile> {
+    let cm = SourceMap::default();
+    cm.load_file(path).expect("failed to load fixture")
+}
+
+// Reuse the AST parser's fixtures to keep both public parser paths in sync.
+#[testing::fixture("../swc_ecma_parser/tests/typescript/types/tuple-optional/input.ts")]
```
kdy1: I don't like this way of testing. Can we reuse the existing `#[testing::fixture]`?

magic-akari: Since `swc_ecma_lexer` is going to be removed, would manual verification be sufficient here? We already have regression coverage in the main parser’s fixture suite. I’d suggest removing the new lexer tests and keeping the main parser fixtures. What do you think?

kdy1 (approved):

PR #12372, opened 2026-09-14
fix(es/parser): preserve type references named asserts before line breaks

magic-akari:
**Description:**

Fix valid TypeScript being rejected when a return type named `asserts` is followed by another member on the next line:

```typescript
type asserts = number;
interface I {
  f(): asserts
  p: number;
}
```

The parser now treats `asserts` as the return type and keeps `p` as a separate property. This also prevents class fields from being silently removed from the AST and generated JavaScript. The fix applies to both parser implementations.


crates/swc_ecma_lexer/src/parser/typescript.rs
```diff
@@ -62,4 +66,25 @@ mod tests {
         })
         .unwrap();
     }
+
+    // Share inputs with the main parser's AST snapshot tests.
+    #[testing::fixture("../swc_ecma_parser/tests/typescript/function/predicate-types/*.ts")]
```
kdy1: Can you convert this to an inline test?

magic-akari: Since `swc_ecma_lexer` is a legacy crate, I'll remove the test case I added for it and only keep the one for the main parser crate.

kdy1 (approved):

PR #12384, opened 2026-09-16
feat(es/minifier): honor `/*#__PURE__*/` on property reads and destructuring patterns

lukesandberg:
SWC supports a `pure_getters` compress option, but it is a whole-program promise that *no* property read anywhere has side effects.

This adds a per-site opt-in — `/*#__PURE__*/` on an individual property read and destructuring pattern. Rollup and esbuild read the annotation the same way.

[Playground (before)](https://play.swc.rs/?version=1.15.46&code=H4sIAAAAAAAC%2F1WQMU%2FEMAyF9%2FyKp7LAiWt3TgwM7Ahxc%2BU2bgmkceQkIHTiv5OoQojR1vP3PtltUTTjApne8I1FZUPn3dSdjBkGPIQgmTLbO1BKZWOLLJgYhOQsH3lZeM7HRZmhTPYWSeAyXIJViZFtb4bD1Tg%2BnZ8fx%2FEwtJ6efql7yTnQX01UTqwfXFETz1RS61o5Z1bMUryFlgDSyWUl%2Fao7y71p1OjJhZ348lqvdqaTmvbV6lP0PaFNsJyyljkXdWFFpAYPvZklpIz%2FuhdQFamfuW%2Fmld5C4rn3sl53VgJ3NyfzA4fr7vBGAQAA)

```js
import { obj } from "lib";

/*#__PURE__*/ obj.annotated;   // dropped
obj.plain;                     // preserved
const /*#__PURE__*/ { a, b } = obj;  // dropped
```

---

### 1. `feat(es/minifier): drop ignored property accesses under pure_getters`

`pure_getters` was consulted in exactly one place — `take_pat_if_unused`, which only decides whether unused bindings can be dropped out of a destructuring pattern. A property access in statement position was *always* preserved, even with `pure_getters: true`, because `ignore_return_value` had no `Expr::Member` arm. So the option bought considerably less than its name suggests.

This adds that arm. When the option allows it, the access is replaced by the side effects of its operands: `x().y` becomes `x()`, and `({p: 1}).p` disappears. The object is evaluated before a computed key, preserving evaluation order (`x()[y()]` → `x(), y()`).

`can_assume_pure_getter` implements all three variants of the option rather than only `Bool(true)`:

- `Str` consults the allow list, including string-literal computed keys.
- `Strict` stays **disabled** — in terser it relaxes nullish checks, not getter effects.

Writes are untouched: assignment, compound assignment, update and `delete` can all run a setter, and the write is the point of the statement.

This unblocks four terser fixtures (`issue_2313_6`, `issue_2938_3`, `issue_2938_4`, `impure_getter_2`), which now match terser's output. `issue_2838` stays postponed — it needs dead *assignment* removal, a separate optimization.

### 2. `feat(es/minifier): honor /*#__PURE__*/ on property reads and destructuring patterns`

Pure marks normally ride on a node's `SyntaxContext`, but `MemberExpr`, `ObjectPat` and `ArrayPat` have none, so annotations on them go in a side table keyed by source position. It is collected by `info_marker` during the traversal it already performs, and read by `pure_optimizer` and `take_pat_if_unused`.

#### Why a side table rather than `ctxt` on `MemberExpr`

Very feasible but API breaking for the JS-visible AST structures, hence the side-table.

**Testing:**

New `tests/fixture/pure-getters/` suite covering the three option variants, evaluation order, annotation ownership, nullish assertion, survival through inlining, and the boundaries where the optimization must not fire (writes, optional chaining, `strict`).

- `cargo test -p swc_ecma_minifier` — green, no snapshot churn
- `cargo test -p swc` — green
- `cargo check --workspace --all-targets` — clean
- `cargo clippy -p swc_ecma_minifier --all-targets -- -D warnings` — clean
- `cargo fmt --all` — applied

**BREAKING CHANGE:**

None. Both changes are opt-in: the first only activates under an explicit `pure_getters` setting, the second only where the user wrote an annotation.

None.


crates/swc_ecma_minifier/src/compress/optimize/unused.rs
```diff
@@ -441,12 +441,24 @@ impl Optimizer<'_> {
...
-            Some(Expr::TaggedTpl(t)) => t.ctxt.has_mark(pure_mark),
+        // An annotation on the pattern itself, as in
+        // `const /*#__PURE__*/ { a } = obj`. Unlike an annotation on a
+        // call-shaped initializer, this one is authored *about the
+        // destructuring*: it asserts both that the property reads it performs
+        // have no side effects and that the initializer is not nullish, so
+        // dropping the pattern cannot swallow a `TypeError`.
+        let has_pat_pure_ann = match &*name {
+            Pat::Object(p) => self.pure_annotations.has_pattern(p.span.lo),
+            Pat::Array(p) => self.pure_annotations.has_pattern(p.span.lo),
```
lukesandberg: merge the branches?

crates/swc_ecma_minifier/src/compress/optimize/unused.rs
```diff
@@ -441,12 +441,24 @@ impl Optimizer<'_> {
...
-            Some(Expr::Call(c)) => c.ctxt.has_mark(pure_mark),
-            Some(Expr::New(n)) => n.ctxt.has_mark(pure_mark),
-            Some(Expr::TaggedTpl(t)) => t.ctxt.has_mark(pure_mark),
+        // An annotation on the pattern itself, as in
+        // `const /*#__PURE__*/ { a } = obj`. Unlike an annotation on a
+        // call-shaped initializer, this one is authored *about the
+        // destructuring*: it asserts both that the property reads it performs
+        // have no side effects and that the initializer is not nullish, so
+        // dropping the pattern cannot swallow a `TypeError`.
+        let has_pat_pure_ann = match &*name {
```
lukesandberg:
```suggestion
        // Respect annotations on the pattern itself.
        //  e.g. `const /*#__PURE__*/ { a } = obj`.
        // Asserting that the implicit property accesses are pure.
```

crates/swc_ecma_minifier/src/compress/pure/misc.rs
```diff
@@ -107,6 +108,54 @@ enum GroupType<'a> {
...
+    /// `pure_getters` is an unsound-by-design option: the user promises that
+    /// property reads are free of side effects. We therefore only consult it,
+    /// and never try to prove the claim ourselves.
+    ///
+    /// - [PureGetterOption::Bool(true)] assumes every property read is pure.
+    /// - [PureGetterOption::Str] assumes only the listed property names are.
+    /// - [PureGetterOption::Bool(false)] and [PureGetterOption::Strict] make no
+    ///   such promise. `Strict` only relaxes *nullish* checks in terser, not
+    ///   getter effects, so it must not enable this optimization.
+    fn can_assume_pure_getter(&self, prop: &MemberProp) -> bool {
```
lukesandberg, line 114: simplify comment
```suggestion
    /// Returns `true` if reading `prop` off an arbitrary object can be assumed
    /// not to not have observable side effects,  according to the `pure_getters`
    /// compress option.
    fn can_assume_pure_getter(&self, prop: &MemberProp) -> bool {
```

crates/swc_ecma_minifier/src/compress/pure/misc.rs
```diff
@@ -107,6 +108,54 @@ enum GroupType<'a> {
...
+        match &self.options.pure_getters {
+            PureGetterOption::Bool(true) => true,
+            PureGetterOption::Bool(false) | PureGetterOption::Strict => false,
+            PureGetterOption::Str(allowed) => match prop {
+                MemberProp::Ident(i) => allowed.contains(&i.sym),
+                // A private name can only resolve to a member of the
+                // lexically enclosing class, never to a listed public
+                // property name.
+                MemberProp::PrivateName(..) => false,
```
lukesandberg, line 122:
```suggestion
            PureGetterOption::Str(allowed) => match prop {
                MemberProp::Ident(i) => allowed.contains(&i.sym),
                // A private names are not supported
                MemberProp::PrivateName(..) => false,
```

crates/swc_ecma_minifier/src/compress/pure/misc.rs
```diff
@@ -107,6 +108,54 @@ enum GroupType<'a> {
...
+    }
+
+    /// Returns `true` if reading `member` can be assumed free of getter side
+    /// effects, either because `pure_getters` says so globally or because this
+    /// specific access is annotated with `/*#__PURE__*/`.
+    ///
+    /// The annotation is scoped to the single access it precedes, which lets a
+    /// library opt into the optimization at a known-safe call site without
+    /// enabling the unsound global option for the whole program.
+    fn is_pure_member_access(&self, member: &MemberExpr) -> bool {
```
lukesandberg, line 139:
```suggestion
    /// Returns `true` if reading `member` can be assumed free of getter side
    /// effects, either because `pure_getters` says so globally or because this
    /// specific access is annotated with `/*#__PURE__*/`.
    fn is_pure_member_access(&self, member: &MemberExpr) -> bool {
```

crates/swc_ecma_minifier/src/compress/pure/misc.rs
lukesandberg: do we still need this branch?

crates/swc_ecma_minifier/src/compress/pure/misc.rs
```diff
@@ -1995,6 +2044,38 @@ impl Pure<'_> {
...
+                // With `pure_getters`, or a `/*#__PURE__*/` annotation on
+                // this access, reading a property is assumed to be free of
+                // side effects, so the access itself can be dropped.
+                // The object and a computed key still have to be evaluated,
+                // because they can run arbitrary code (`x().y` must keep
+                // `x()`).
+                //
+                // `super.foo` is deliberately not handled here: it is a
+                // `SuperPropExpr`, and `super` is not an expression we can
+                // evaluate on its own.
```
lukesandberg: simplify this comment

something like:

// unuused member accesses can be dropped only if they are pure

crates/swc_ecma_minifier/src/compress/pure/mod.rs
```diff
@@ -65,6 +67,11 @@ struct Pure<'a> {
     options: &'a CompressOptions,
     config: PureOptimizerConfig,
     marks: Marks,
+
+    /// Pure annotations on nodes without a `SyntaxContext` to carry them.
+    ///
+    /// A shared reference so that [`Parallel::create`] stays cheap.
```
lukesandberg:
```suggestion
```

crates/swc_ecma_minifier/src/compress/mod.rs
```diff
@@ -59,6 +62,9 @@ struct Compressor<'a> {
     marks: Marks,
     options: &'a CompressOptions,
     mangle_options: Option<&'a MangleOptions>,
+
+    /// Pure annotations on nodes without a `SyntaxContext` to carry them.
```
lukesandberg: dropp comment

crates/swc_ecma_minifier/src/metadata/mod.rs
```diff
@@ -161,6 +173,51 @@ impl VisitMut for InfoMarker<'_> {
...
+        // A member expression starts at the same position as its own object,
+        // so a comment before `x()` in `/*#__PURE__*/ x().y` is found by a
+        // lookup on either node. That annotation belongs to the call, which
+        // already consumes it as a pure call, so only claim it here when the
+        // object cannot have taken it: the annotation has to sit between the
+        // start of the expression and the property being read.
+        //
+        // This keeps `/*#__PURE__*/ a.b` (object is a bare identifier)
+        // working while leaving `/*#__PURE__*/ x().y` to the call.
+        if n.obj.span().lo == n.span.lo && !matches!(&*n.obj, Expr::Ident(..) | Expr::This(..)) {
```
lukesandberg: so how do we mark the property access of `x().y ` as pure

`/*#__PURE__*/(x()).y`?

crates/swc_ecma_minifier/src/metadata/mod.rs
```diff
@@ -161,6 +173,51 @@ impl VisitMut for InfoMarker<'_> {
         n.visit_mut_children_with(self);
     }
 
+    /// Records `/*#__PURE__*/ obj.prop`.
+    ///
+    /// The annotation asserts that reading this property does not invoke a
+    /// getter with side effects. It says nothing about the object expression,
+    /// which is still evaluated.
```
lukesandberg: drop this comment

crates/swc_ecma_minifier/src/metadata/mod.rs
```diff
@@ -161,6 +173,51 @@ impl VisitMut for InfoMarker<'_> {
...
+        if has_pure(self.comments, n.span) && !n.span.is_dummy_ignoring_cmt() {
+            self.annotations.insert_member(n.span.lo);
+        }
+    }
+
+    /// Records `const /*#__PURE__*/ { a } = obj`.
+    ///
+    /// On a destructuring pattern the annotation asserts both that the reads
+    /// it performs are free of side effects and that the initializer is not
+    /// nullish, so dropping the pattern cannot swallow a `TypeError`.
```
lukesandberg:
```suggestion
    /// Records `const /*#__PURE__*/ { a } = obj`.
```

crates/swc_ecma_minifier/src/metadata/pure_annotations.rs
```diff
@@ -0,0 +1,68 @@
...
+/// Cheap to share: the minifier's passes hold a `&PureAnnotations`, including
+/// across the parallel `Pure` visitor.
+#[derive(Debug, Default)]
+pub(crate) struct PureAnnotations {
+    /// `lo` of each annotated [`swc_ecma_ast::MemberExpr`].
+    members: FxHashSet<BytePos>,
+
+    /// `lo` of each annotated destructuring pattern.
+    patterns: FxHashSet<BytePos>,
+}
```
lukesandberg: is it important to distinguish the two cases?  should this just be a single FxHashSet?

Also does swc have an AutoSet implementation that stors small sets as vectors?

crates/swc_ecma_minifier/src/eval.rs
```diff
@@ -96,6 +96,10 @@ impl Evaluator {
                     ..Default::default()
                 },
                 None,
+                // The evaluator runs on code it synthesizes or on snippets
+                // whose annotations were already consumed, so there is nothing
+                // to look up here.
```
lukesandberg: drop these comments in the tests

crates/swc_ecma_minifier/src/compress/optimize/unused.rs
lukesandberg: Done — merged into the existing `has_pure_ann` match.

crates/swc_ecma_minifier/src/compress/optimize/unused.rs
lukesandberg: Applied.

crates/swc_ecma_minifier/src/compress/pure/misc.rs
lukesandberg: Applied.

crates/swc_ecma_minifier/src/compress/pure/misc.rs
lukesandberg: Applied. Also moved the `strict` rationale to a one-liner on that arm, since it's the non-obvious one (in terser `strict` relaxes nullish checks, not getter effects, so it must not enable this).

crates/swc_ecma_minifier/src/compress/pure/misc.rs
lukesandberg: Applied.

crates/swc_ecma_minifier/src/compress/pure/mod.rs
lukesandberg: Dropped.

crates/swc_ecma_minifier/src/compress/mod.rs
lukesandberg: Dropped.

crates/swc_ecma_minifier/src/metadata/mod.rs
lukesandberg: Applied.

crates/swc_ecma_minifier/src/eval.rs
lukesandberg: Dropped, all four call sites.

crates/swc_ecma_minifier/src/compress/pure/misc.rs
lukesandberg: Good catch — no. It fell through identically to `_ => {}` below it, and the new pure-access arm above now handles the case it used to sit in front of. Removed; full suite stays green without it.

crates/swc_ecma_minifier/src/compress/pure/misc.rs
lukesandberg: Applied, using your wording. Kept a short second sentence noting that the object and computed key are still evaluated, since that is the part that is easy to get wrong when reading the arm.

crates/swc_ecma_minifier/src/metadata/pure_annotations.rs
lukesandberg: Merged into a single `FxHashSet`. You are right that the split bought nothing: a `BytePos` denotes one syntactic node, so a member expression and a pattern can never collide on the same `lo`.

No `AutoSet` in this repo — I grepped, and the only small-collection deps are `smallvec` and `indexmap` (I think you are thinking of the Turbopack type). Happy to swap in something vector-backed if you want; the set is empty for the overwhelmingly common case of source with no such annotations, so I left it as a plain `FxHashSet` for now.

crates/swc_ecma_minifier/src/metadata/mod.rs
lukesandberg: Currently you cannot, and `/*#__PURE__*/(x()).y` does not work either — `paren_remover` relocates the comment onto the inner call, so it stays the call's annotation. I tested all the placements I could think of (`/*#__PURE__*/ x().y`, `/*#__PURE__*/(x()).y`, `(x())/*#__PURE__*/.y`) and none reach the outer read.

The only way to express it today is to split the object out, which does work:

```js
const t = x();
/*#__PURE__*/ t.y;   // => x(), i.e. the read is dropped, the call is kept
```

This is a real limitation of keying on `lo`, and it is the strongest argument for putting a real `ctxt` on `MemberExpr` instead — with a mark on the node there is no ambiguity about which node the annotation attached to, and `/*#__PURE__*/ x().y` could mean the read. I measured that change (see the PR description: free in memory, 15 call sites) and the blocker is purely that `ctxt` is serialized into the JS-visible AST.

I think the `lo` version is still worth landing: the common case is a read off a plain identifier or an import, which works. But if you would rather have the general form, I am happy to do the `ctxt` version for a major release instead — your call.

crates/swc_ecma_minifier/src/metadata/mod.rs
lukesandberg: Shortened to three lines. Kept a brief note since the guard looks arbitrary otherwise — dropping it entirely makes `/*#__PURE__*/ x().y` silently drop `x()`, which is a miscompile I hit while building this.

lukesandberg:
@codspeedbot fix this regression

crates/swc_ecma_minifier/src/metadata/pure_annotations.rs
```diff
@@ -0,0 +1,40 @@
...
+use rustc_hash::FxHashSet;
+use swc_common::BytePos;
+
+/// Source positions of nodes annotated with `/*#__PURE__*/` that cannot carry
+/// the annotation on their own [`swc_common::SyntaxContext`].
+///
+/// Cheap to share: the minifier's passes hold a `&PureAnnotations`, including
+/// across the parallel `Pure` visitor.
+#[derive(Debug, Default)]
+pub(crate) struct PureAnnotations {
```
kdy1, line 21: I love this idea

kdy1 (approved):
Thank you! I think this approach is pretty clever

Austaras:
> This adds a per-site opt-in — `/*#__PURE__*/` on an individual property read and destructuring pattern. Rollup and esbuild read the annotation the same way.

[esbuild](https://esbuild.github.io/try/#dAAwLjI4LjIALS1taW5pZnkAaW1wb3J0IHsgb2JqIH0gZnJvbSAibGliIjsKCi8qI19fUFVSRV9fKi8gb2JqLmFubm90YXRlZDsgICAvLyBkcm9wcGVkCm9iai5wbGFpbjsgICAgICAgICAgICAgICAgICAgICAvLyBwcmVzZXJ2ZWQKY29uc3QgLyojX19QVVJFX18qLyB7IGEsIGIgfSA9IG9iajsgIC8vIGRyb3BwZWQ) doesn't. I don't know much about rollup but it is not a minifier. And terser does not support this too. I don't really think SWC should. @kdy1 what's your opinion?

lukesandberg:
Sorry i got that wrong.

I figure if we allow pure annotations to allow us to trim unused function calls, it makes sense to apply it to property access, since the only reason we might not drop it when unused is because it _might_ actually be a function call (getter).

Austaras:
I'm not aginast this idea, it's not bad. But I don't SWC should be the one and only tool support this feature.

Austaras:
Also https://github.com/javascript-compiler-hints/compiler-notations-spec/blob/main/pure-notation/pure-notation-spec.md

PR #12385, opened 2026-09-17
perf(swc_core): Deduplicate ICU dependencies

chenjiahan:
**Description:**

`@swc/core` currently includes both ICU 1.x and 2.x. Update `idna_adapter` from 1.2.0 to 1.2.1 in `Cargo.lock` so URL handling shares ICU 2.x with Unicode regular expression transforms, removing the ICU 1.x dependency chain.

Measured on macOS arm64:

| Artifact | Before (kB) | After (kB) | Reduction (kB) | Reduction (%) |
| --- | ---: | ---: | ---: | ---: |
| `swc.darwin-arm64.node` | 26,695.152 | 26,645.584 | 49.568 | 0.186% |


kdy1 (approved):
Thanks

PR #12390, opened 2026-09-17
fix(es/compat): Preserve class name TDZ in computed method keys

### LINKED ISSUE #6098: Regression: ReferenceError is not thrown when Javascript symbols are not initialized
### Describe the bug

In v.1.2.189 and earlier this feature was previously working, but versions after that do not have the expected behavior. It was tested in the SWC playground with default settings. 

When creating a class and making a method with a symbol that has the same name as the class, such as `[Foo](){},` it previously threw a reference error. In newer versions it does not throw any ReferenceErrors.

### Input code

```typescript
var B = class Foo { [Foo](){} }
```

### Config

_No response_

### Playground link

https://play.swc.rs/?version=1.2.189&code=H4sIAAAAAAAAAytLLFJwUrBVSM5JLC5WcPb2VKhWiAZSsRqa1bUKtQAf53TWHwAAAA%3D%3D&config=H4sIAAAAAAAAAy3LQQrDMAxE0bvM2ttufIccQqiiJNSx0ciQEHz3xKWrYXj8CxsV%2BUKIfyyQYXwhoYnTfALPPeSYoEWovrbAGAmlvvvXfunZnoXWUuq%2BEQ%2BydldbpBE5vFvCyuUfzD9uqCJXNngAAAA%3D

### Expected behavior

A reference error should be thrown

### Actual behavior

No reference error is thrown and the class is allowed to be created

### Version

v1.2.196 and after

### Additional context

_No response_

pederbe:
**Description:**

ES2015 class lowering currently evaluates direct computed method name reads after the generated constructor binding exists, so `[Foo]` does not observe the named class binding's temporal dead zone.

Rewrite eager reads of the resolved inner class name with the existing TDZ helper before method-key extraction. This preserves computed-key ordering without affecting same-spelling outer bindings or deferred function bodies.

Fixes #6098


magic-akari:
Thanks for the contribution and the test coverage. I don't think we should merge this change, for the same general reasons discussed in https://github.com/swc-project/swc/issues/12130#issuecomment-5358067032.

The example in #6098 is expected to throw a `ReferenceError`. The question is not whether that behavior is required by the specification, but whether reproducing it in the downlevel class transform justifies the additional complexity and maintenance cost.

This change restores a runtime error for code that accesses the class's inner binding before initialization. The issue demonstrates an observable conformance difference, but no concrete application use case has been presented that depends on preserving this particular error. Babel's class transform also does not preserve this specific case, although that is supporting context rather than the basis for this decision.

Adding a check for a direct read such as `[Foo]` is relatively straightforward. Preserving class-name TDZ behavior consistently is more involved: a closure may read the binding while a computed key is being evaluated, or it may be retained and called after class initialization, when the same read must succeed. Assignments and updates also require careful handling of evaluation order. The patch explicitly leaves some of these cases outside its scope.

I appreciate that the implementation is deliberately narrow and reuses an existing helper. However, accepting it would still add special-case transformation logic and an ongoing maintenance obligation without providing consistent TDZ preservation. I don't think the practical benefit demonstrated here justifies that tradeoff.

Our compatibility transforms aim to support practical application code on older targets. That does not require implementing every observable error-path distinction regardless of its cost. For this case, we prefer to retain the existing behavior.

This is a scope decision, not a request to expand the patch to cover the remaining cases. For these reasons, I'm closing this PR.

pederbe:
I understand, no worries. The reasoning makes sense. Cheers.

Issue #11084, opened 2025-09-12
Left-hand side with object attributes substituted by value in array destructuring


0xc22b:
### Describe the bug

Using array destructuring with object attributes on the left-hand side:

```
const bin = {
  hasMore: false, hasDisorder: false,
};
[bin.hasMore, bin.hasDisorder] = [true, true];
```

It gets compiled into:

```
  [!1, !1] = [
    !0,
    !0
  ];
```

which is incorrect: SyntaxError: Invalid destructuring assignment target

### Input code

```typescript
export function test() {
  const bin = {
    hasMore: false, hasDisorder: false,
  };
  [bin.hasMore, bin.hasDisorder] = [true, true];
}
```

### Config

```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "tsx": false
    },
    "target": "es2017",
    "loose": false,
    "minify": {
      "compress": {
        "arguments": false,
        "arrows": true,
        "booleans": true,
        "booleans_as_integers": false,
        "collapse_vars": true,
        "comparisons": true,
        "computed_props": true,
        "conditionals": true,
        "dead_code": true,
        "directives": true,
        "drop_console": false,
        "drop_debugger": true,
        "evaluate": true,
        "expression": false,
        "hoist_funs": false,
        "hoist_props": true,
        "hoist_vars": false,
        "if_return": true,
        "join_vars": true,
        "keep_classnames": false,
        "keep_fargs": true,
        "keep_fnames": false,
        "keep_infinity": false,
        "loops": true,
        "negate_iife": true,
        "properties": true,
        "reduce_funcs": false,
        "reduce_vars": false,
        "side_effects": true,
        "switches": true,
        "typeofs": true,
        "unsafe": false,
        "unsafe_arrows": false,
        "unsafe_comps": false,
        "unsafe_Function": false,
        "unsafe_math": false,
        "unsafe_symbols": false,
        "unsafe_methods": false,
        "unsafe_proto": false,
        "unsafe_regexp": false,
        "unsafe_undefined": false,
        "unused": true,
        "const_to_let": true,
        "pristine_globals": true
      },
      "mangle": false
    }
  },
  "module": {
    "type": "es6"
  },
  "minify": false,
  "isModule": true
}
```

### Link to the code that reproduces this issue

[...]

PR #12398, opened 2026-09-21
fix(es/minifier): mark properties assigned through destructuring targets as mutated

### LINKED ISSUE #12397: es/minifier: `hoist_props` inlines properties written through destructuring assignment targets (incomplete fix of #11084)
**Describe the bug**

With `hoist_props` enabled (the default), a property that is only written through a destructuring assignment pattern is treated as never written. The property is hoisted and its initial value is inlined everywhere, including into the assignment target, which produces invalid JavaScript.

**Input code**

```js
const o = { n: 1 };
[o.n] = [2];
console.log(o.n);
```

**Config**

Rust API, `swc_ecma_minifier::optimize` on a `Script`:

```rust
MinifyOptions {
    compress: Some(CompressOptions {
        top_level: Some(TopLevelOptions { functions: true }),
        ..Default::default()   // hoist_props: true
    }),
    mangle: Some(MangleOptions { top_level: Some(true), ..Default::default() }),
    ..Default::default()
}
```

Setting only `hoist_props: false` makes every variant below correct. The bug also reproduces inside a function body, so `top_level` is not required.

**Expected behavior**

Prints `2`.

**Actual behavior**

```js
[1]=[2],console.log(1);
```

`SyntaxError: Invalid destructuring assignment target`

**Scope**

Reproduces with:
- object patterns: `({ x: o.n } = { x: 2 })`
- nested members: `[obj.a.x] = [10]`
- `var` and `const` bindings, number and boolean values
- inside a function, not only at top level

A plain assignment `o.n = 2` is handled correctly, so the gap is specific to member expressions inside assignment patterns.

**Relation to #11084**

#11084 was closed by #11221. Its regression fixture `tests/fixture/issues/11084` runs with `{ booleans, properties, evaluate }`, which leaves `hoist_props` off. Re-running that same input with `hoist_props` enabled still reproduces the original report: case 1 emits an invalid assignment target, and case 3 hoists `obj.a.x` / `obj.b.y` into `const` bindings that the destructuring then assigns to.

**Version**

`swc_ecma_minifier` 61.0.8 (crates.io) and `main` @ 4b42cd59d7.

Found with a differential tester that runs the original and minified output under Node and compares stdout and exit status.

baltasarblanco:
**Description:**

With `hoist_props`, a property that is only written through a destructuring assignment
target (`[o.n] = [2]`, `({ x: o.n } = obj)`) was treated as never written, so it was
hoisted and its initial value inlined into the assignment target itself, producing invalid
JavaScript (`[1]=[2]`).

The usage analyzer already marks `o` as property-mutated for `o.n = v` via
`mark_mutation_if_member`, but the `AssignTarget::Pat` branch of `visit_assign_expr` and
the for-in/for-of heads never did. This PR walks the pattern and marks every member target
at any depth (array, object, default, rest) the same way. It only ever over-preserves.

The extra work runs only for assignments with a pattern on the left and for-in/for-of
heads, as a single walk of the pattern.

The existing `issues/11084` fixture runs without `hoist_props`, which is why the original
report still reproduced. With `@swc/core` 1.16.2 default options it breaks inside any
function body (e.g. an IIFE), and at the top level once `toplevel` is enabled.

This also updates `issues/9739`: its snapshot had frozen the same bug through variable
inlining (`[arr[0], arr[1]] = …` became writes to fresh array literals, so `arr` was never
swapped). The new output keeps the swap.

**Testing:**

- `issue_12397_hoist_props_destructuring_member_targets` (exec test, 9 variants) fails
  without the change and passes with it.
- `cargo test -p swc_ecma_minifier --no-fail-fast`: 0 failures across all test binaries.
- `cargo clippy -p swc_ecma_minifier --all-targets -- -D warnings`: clean.

Fixes #12397. Follow-up of #11084 / #11221.


kdy1 (approved):
Thank you!

